PR #685, opened 2019-02-01
TypeScript migration

pronebird:

Git checklist:

1. This PR migrates the entire app from `Flow` to `TypeScript`. While being seemingly large PR, the majority of changes only swap a few characters or add missing type information.

1. Since TypeScript does not copy assets during transpilation (what a word!), I had to add a few rudimental scripts to cover that. It's not perfect and changes to assets won't be picked up automatically, as it used to be but we'll fix this at some point later.

1. TypeScript typings are a part of NPM ecosystem and installed as dependencies (see `@types/*`) packages. The typings for the missing dependencies can be added at `gui/types`.

1. The main TypeScript compiler config is located under `gui/tsconfig.json`, same goes for the linter config which can be found at `gui/tslint.json`.

1. The `assets` folder was moved from `packages/desktop/src/assets` to `packages/desktop/assets`. The paths were prepended with the extra `../` accordingly when needed.

1. We use a lot of disjoint unions that don't really tag the variants, instead we simply use unique keys to detect the variants, i.e:
  
    ```ts
    type T = { a: string } | { b: string };

    var a: T = ...
    // then we go
    if (a.a) {
      // variant a
    } else if (a.b) {
      // variant b
    }
    ```
    
    That doesn't work this way in TypeScript, they suggest using tagged unions, i.e:

    ```ts
    type T = { type: "a", value: string } | { type: "b", value: string };

    var a: T = ...
    switch (a.type) {
      case 'a': ...
      case 'b': ...
    }
    ```
    
    To avoid messing around with our codebase too much I used `in` operator instead which works:
    
    ```ts
    if ('a' in a) {
     // do something with a.a
    } else if ('b' in a) {
     // do something with a.b
    }
    ```
    
    But I guess it can happen that in future we'll have to tag unions to follow the best practices.

1. Note that the tests are currently broken and CI should ignore them, same goes for any linter issues in desktop package. All of this will be addressed in the follow up PRs.

Please pull the PR and run it on your computer because it's a big change.

`yarn workspace @mullvad/components build` - to compile TS in components package
`yarn workspace desktop develop` - to compile and run the desktop app

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/685)


jvff:
___
*[gui/packages/desktop/package.json, line 80 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/685#-LXs8Kby2szpXmNuZtgL:-LXsDFKl6f-a5-Y077wZ:bcr5s2o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1e1e05bbb49196039d615de15fc523b42250192e/gui/packages/desktop/package.json#L80)):*

As I've stated in the description to this PR:

> Note that the tests are currently broken and CI should ignore them, same goes for any linter issues in desktop package. All of this will be addressed in the follow up PRs.

If I went on fixing tests and lint issues the `git diff` would be way harder to follow for many reasons.
</blockquote></details>

Got it, thanks!

jvff:
___
*[gui/packages/desktop/src/main/index.js, line 726 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/685#-LXsDpmw7owpKMGyimKQ:-LXsDpmw7owpKMGyimKR:bvydn65) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f97c5621fad66ae9768dad8cc422bc3a61f76ef6/gui/packages/desktop/src/main/index.js#L726)):*
> ```JavaScript
> 
>       default:
>         throw new Error(`Invalid tunnel state: ${(tunnelState.state: empty)}`);
> ```

Quick question: does Typescript check for `switch` exhaustiveness?

jvff:
___
*[gui/packages/desktop/src/main/keyframe-animation.ts, line 31 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/685#-LXsEm6V1KPXO4jQoArZ:-LXsK33J6g4QCBDgXxiu:bsdf1is) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1e1e05bbb49196039d615de15fc523b42250192e/gui/packages/desktop/src/main/keyframe-animation.ts#L31)):*

`newValue?: onFinishFn` must be equivalent to `newValue: onFinishFn | undefined`. 

Unfortunately TypeScript throws an error saying that setter and getter must have the same type when a setter method is defined with an optional argument, i.e: `set onFinish(newValue?: T)`.
</blockquote></details>

Interesting! Thanks :)

jvff:
___
*[gui/packages/desktop/src/main/index.js, line 726 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/685#-LXsDpmw7owpKMGyimKQ:-LXsNqu35VIcf8nFjqKS:b903klo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f97c5621fad66ae9768dad8cc422bc3a61f76ef6/gui/packages/desktop/src/main/index.js#L726)):*

There is a PR already to allow exhaustive :)

https://github.com/palantir/tslint/pull/4469
</blockquote></details>

Awesome :D

PR #706, opened 2019-02-22
Add gettext support

pronebird:

Git checklist:

### What

1. Gettext integration into the source code. I used `node-gettext` which unfortunately cannot consume `PO` directly so I had to use the `gettext-parser` for parsing. In the future we could improve that by converting `PO` to `JSON` and then simply shipping `JSON` only with the app. I think we'll be able to sort that out using gulp or similar build systems down the road.

1.  Translations extraction from the source code and catalogue updates. Unfortunately `xgettext` didn't quite work with JSX so I had to use `gettext-extractor`. You will find the corresponding script under `scripts/extract-translations.js`

1. Updating translations. This is done using gettext tools and a bash script located under `scripts/update-translations.sh`.

### Updating translations

Simply run:

```
yarn workspace desktop update-translations
```

This will update the `locales/messages.pot` template file and update the existing language catalogues under `locales/{locale-id}/messages.po`

#### Example

```ts
import { gettext, pgettext } from '../shared/gettext';

const translatedMessage = gettext(
  'quit-app' // message ID
);

// TRANSLATORS: this comment will be extracted and passed to translators
const translatedMessageWithContext = gettext(
  'notifications', // context 
  'update-available' // message ID
);
```

#### Contexts

Assume we have two buttons using the same message ID:

```ts
// fileA.ts
<Button>{gettext("quit-button")}</Button>
```

```ts
// fileB.ts
<Button>{gettext("quit-button")}</Button>
```

Now the translators will only be able to translate the `quit-button` message once and the same message will be used in both `fileA` and `fileB` regardless the fact that both buttons can be used in entirely different parts of the app and carry different meaning.

But what if one button quits the app and the other quits the chat room for example?

Then we need to use contexts to distinguish between them and enable translators to translate each case individually:

```ts
// fileA.ts
<Button>{ pgettext("chat-room", "quit-button") }</Button>
```

```ts
// fileB.ts
<Button>{ pgettext("settings", "quit-button") }</Button>
```

#### Translator notes

Developers can leave comments to translators to give them some insight into what they translate by using the `// TRANSLATORS: ` comment before the invocation to `gettext` or `pgettext`. 

Some messages also contain placeholders for variables plugged in runtime, it's good to outline those in the translator notes.

```ts
// TRANSLATORS: Account expiration message. Example: Your account expires on %(expiry)s
gettext("account-expiration-message")
```

_Note: Unfortunately for the time being, our gettext parser only supports single line comments, so please do not write poems in comments just yet or do a one-liner._

#### Formatting localised messages

Using `sprintf-js` we can format the localised messages in two different fashions. Which one is better is still an open question, however I've inclined towards old python way because it seems to be more intuitive in my opinion. See below.

##### Classic sprintf

```ts
// TRANSLATORS: Example: Set OpenVPN MSS value. Valid range: %d - %d.
sprintf( gettext('mssfix-footer-msg'), MIN_MSSFIX_VALUE, MAX_MSSFIX_VALUE )
```

##### Old Python way

```ts
// TRANSLATORS: Example: Set OpenVPN MSS value. Valid range: %(min)d - %(max)d.
sprintf( gettext('mssfix-footer-msg'), {
  min: MIN_MSSFIX_VALUE,
  max: MAX_MSSFIX_VALUE,
} )
```

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/706)


gui/packages/desktop/src/renderer/components/Settings.tsx
```diff
@@ -85,33 +95,37 @@ export default class Settings extends Component<IProps> {
     const isOutOfTime = expiry ? expiry.hasExpired() : false;
     const formattedExpiry = expiry ? expiry.remainingTime().toUpperCase() : '';
 
+    // TRANSLATORS: OUT OF TIME
+    const outOfTimeMessage = pgettext('settings-view', 'account-cell-out-of-time-subtext');
```
faern: So with this solution the developers need to come up with a unique identifier for each string? For the website I know the original text (English) is used as the actual lookup key for the translation. Is that not possible using this tool?

jvff:
___
*[gui/packages/desktop/src/renderer/components/Settings.tsx, line 99 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/706#-LZL7HOkvLU5p6vIb-Hg-r2-99:-LZdEkB1-mhlca6DyC_I:bghal0f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/435d77a8390043d44ef9864b4cc426f784b3659d/gui/packages/desktop/src/renderer/components/Settings.tsx#L99)):*

After looking at Crowdin I realised they do not support the "no source language" configuration so it seems that I'll be moving English over to the source code.
</blockquote></details>

BTW, on Android the translation strings *must* have an identifier. Not sure if it affects us in this case though.

PR #718, opened 2019-03-04
New version check

faern:
Unify the version check into one RPC call. Should cut time to do it in half (since we don't have any HTTP keep-alive and thus do the full handshake for each request)

This also changes the struct layout, since we merged it into one struct.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/718)

### DIFF
--- a/gui/src/renderer/redux/version/reducers.ts
+++ b/gui/src/renderer/redux/version/reducers.ts
@@ -25,14 +25,11 @@ export default function(
   action: ReduxAction,
 ): IVersionReduxState {
   switch (action.type) {
-    case 'UPDATE_LATEST': {
-      const { latest, ...other } = action.latestInfo;
+    case 'UPDATE_LATEST':
       return {
         ...state,
-        ...other,
-        ...latest,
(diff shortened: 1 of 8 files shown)


gui/src/renderer/redux/version/reducers.ts
```diff
@@ -26,11 +26,10 @@ export default function(
 ): IVersionReduxState {
   switch (action.type) {
     case 'UPDATE_LATEST': {
-      const { latest, ...other } = action.latestInfo;
+      const { ...other } = action.latestInfo;
```
pronebird: Just remove that line and return the latestInfo itself.

```js
return {
  ...state,
  ...action.latestInfo,
}
```

pronebird: Btw there is no need to scope the case given that we don't need to define any variables inside of case, so the entire case should look like that:

```js
case 'UPDATE_LATEST':
      return {
        ...state,
        ...action.latestInfo,
      };
```

PR #797, opened 2019-04-04
Improve firewall rules persistence on Windows

mvd-ows:
Lots of changes:

* Firewall rules are installed to stick until the next reboot, or until the daemon removes them.

* We now maintain a registry with GUIDs of all the objects we might install into WFP, so we can purge them all will.

* Initialization updated to purge firewall before establishing baseline.

* Cannot install WFP objects that are not in the registry to prevent issues down the road.

* Initialization extended to atomically enter the blocked state during init, if applicable.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/797)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ Line wrap the file at 100 chars.                                              Th
 ### Fixed
 - Reset the tray icon padlock to the unsecured state, when losing connectivity with the daemon.
 
+### Changed
+#### Windows
+- Make the firewall rules permanent until reboot, or until the daemon removes them.
+
 ## [2019.3] - 2019-04-02
 ### Fixed
 #### Windows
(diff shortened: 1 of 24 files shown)


mvd-ows:
___
*[talpid-core/src/tunnel_state_machine/mod.rs, line 195 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/797#-Lbdo5EB5FpmCdavBO7W:-Lbdo5EB5FpmCdavBO7X:b6ed0y0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/eca93f8828df00e6e34d7de3d50e6f324f42d408/talpid-core/src/tunnel_state_machine/mod.rs#L195)):*
> ```Rust
>         let args = if block_when_disconnected {
>             FirewallArguments {
> ```

I figured this would be better than hiding it inside `FiewallArguments::new()`, especially since we're only creating the firewall at one location in the code.

PR #808, opened 2019-04-10
Map an error instead of panicing when starting tokio runtime

faern:
If the thread we start that runs `tokio::run(fut)` stops before a message is sent on the `start` channel, then the channel is cancelled. This for example happens if `tokio::run` fails to create the runtime. This leads to the `expect("server panicked")` panicking the current thread and later the main thread of the daemon panics because of this.

With this tiny fix the panic is turned into an error. So a proper `Err` is returned from `IpcServer::start_with_metadata` instead. This does not solve the root problem. But at least the daemon can log the error chain and handle everything much better.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/808)

### DIFF
--- a/talpid-ipc/src/lib.rs
+++ b/talpid-ipc/src/lib.rs
@@ -25,6 +25,9 @@ pub enum Error {
     #[error(display = "Unable to start IPC server")]
     StartServerError(#[error(cause)] io::Error),
 
+    #[error(display = "IPC server thread panicked and never returned a start result")]
+    ServerThreadPanicError,
+
     #[error(display = "Error in IPC server")]
     IpcServerError(#[error(cause)] io::Error),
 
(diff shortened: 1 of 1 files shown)


talpid-ipc/src/lib.rs
```diff
@@ -65,7 +68,7 @@ impl IpcServer {
                 thread::spawn(move || tokio::run(fut));
                 start
                     .wait()
-                    .expect("server panicked")
+                    .map_err(|_cancelled| Error::ServerThreadPanicError)?
```
pronebird: Curious, can you squash these two into one statement?

```rust
start
  .wait()
  .map_err(|_cancelled| Error::ServerThreadPanicError)? // <-- 
  .map(Err) // <--
```

PR #859, opened 2019-05-17
Add better offline detection on Windows.

pinkisemils:
@mvd-ows added code to enumerate network adapters to check if the host is connected to a network. This code is then used in the offline checking code in the daemon, in combination with the existing code that's used to detect when the host is entering sleep. This greatly improves offline detection on Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/859)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,9 @@ Line wrap the file at 100 chars.                                              Th
 #### macOS
 - Raise max number of open files for the daemon to 1024. Should prevent threads from panicking.
 
+#### Windows
+- Add better offline detection
+
 
 ## [2019.4] - 2019-05-08
 This release is identical to 2019.4-beta1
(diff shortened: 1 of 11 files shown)


mvd-ows:
___
*[windows/winnet/src/extras/loader/loader.cpp, line 12 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/859#-Lf4bl81-wcK5AVJRzE-:-Lf4dgNu8X36TvDUsr3v:btcybtb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/99382ff26a6d95477f43690e46b11fb489664b7a/windows/winnet/src/extras/loader/loader.cpp#L12)):*

I'm not a fan of commented out code as a way to keep version history. If this is not intended to be used now, can we just delete it?
</blockquote></details>

It's a stand-alone testing tool. Rather than keeping all the code, and making a nice CLI this was deemed to be the quickest path forward. Now when I need the commented code (perhaps tomorrow) I don't have to go looking for it in Git.

PR #860, opened 2019-05-17
Windows firewall updates (mDNS, NDP, etc)

mvd-ows:
As above.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/860)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,10 +32,10 @@ Line wrap the file at 100 chars.                                              Th
 - Add GUI translations for Italian, Japanese, Dutch, Portugese, Russian and Turkish.
 - Add missing GUI translations for Czech Republic, USA and UK in the select location view.
 - Add translations for the current location displayed on the main screen in the GUI.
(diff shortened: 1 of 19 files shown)


mvd-ows:
___
*[windows/winfw/src/winfw/rules/permitdhcpserver.cpp, line 86 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/860#-LfMCcOV50NykRua4Y97:-LfOERE-9im3pgtz-9PU:bfcodgz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8ee0f2265770690c6624522d59d82a46e90f10d/windows/winfw/src/winfw/rules/permitdhcpserver.cpp#L86)):*

Shouldn't this also specify the multicast address? 
</blockquote></details>

I looked it up and apparently the `DHCPOFFER`message can be sent as either unicast or multicast (not sure how unicast would work this early though). Then as the address negotiation completes the server sends a `DHCPACK` message which is definitely unicast.

Since we can't express the unicast rule we'll just use the common subset and define a single (weaker) rule to cover all cases.

PR #886, opened 2019-05-29
Add account expiry refresh on iOS

pronebird:

Git checklist:

- Add account expiry refresh which starts as soon as the first request comes in to refresh the account expiry and then repeats every 60 seconds until there are no observers left. Currently we only register observers in Settings and Account views. Observers are weak references and automatically get dropped as soon as the aforementioned controllers deallocate.
- Tweak the header bar margins to match the desktop app
- Add `WeakBox<T>` which enables us holding the weak references in arrays.
- Add helpers to `StaticTableViewDataSource` to be able to reload the table view rows.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/886)


jvff:
___
*[ios/MullvadVPN/AccountExpiryRefresh.swift, line 49 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/886#-Lg2dzHUDkcrwmywXkEQ:-Lg2ewBs0TuOAkEsXPiC:bynphew) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2dd2547453c8e79fed28a3b99cd1efb8f02a2321/ios/MullvadVPN/AccountExpiryRefresh.swift#L49)):*

Good point. The script throws `find: illegal option -- n`... what..
</blockquote></details>

That's weird.  Did you run the command manually?

PR #933, opened 2019-06-19
Fix in and out label width in connection info panel

pronebird:

Git checklist:

This PR refactors the connection info panel and removes the fixed width from the In and Out labels which was causing layout issues in different languages.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/933)


gui/src/renderer/components/ConnectionInfo.tsx
```diff
@@ -16,7 +17,6 @@ const styles = {
     fontWeight: '600',
     color: 'rgb(255, 255, 255)',
     flex: 0,
-    flexBasis: 30,
```
pronebird: FYI removed the fixed width

PR #964, opened 2019-07-04
Make symlinks in /usr/local/bin

faern:
Support asked if we can make the CLI easier to find. So not every guide have to start with how to find/run it. This PR solves that for macOS specifically. We create a symlink to our two CLI tools in `/usr/local/bin`. From what I have gathered online this path should be in all users default `PATH` *but* the directory itself might not exist. So we should hopefully be good if we try to create it and then add the symlinks.

I'll make sure to test this on an untouched macOS computer before merging. But can have it reviewed now.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/964)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ Line wrap the file at 100 chars.                                              Th
 #### Windows
 - Add migration logic to restore lost settings after major Windows update.
 
+#### macOS
+- Add the Mullvad CLI frontend and problem report CLI tool to the PATH, so it can be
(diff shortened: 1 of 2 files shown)


faern:
I tried this on a clean macOS 10.14.5 computer. `/usr/local/bin` was indeed in the user `PATH` but the directory did not exist. This PR correctly created the dir and added the symlink, making them directly available in any already open terminal without sourcing any environment file or anything.

PR #965, opened 2019-07-04
Add GUI page for managing WireGuard key

pinkisemils:
This PR adds a new settings page for managing WireGuard keys. This allows the user to manually generate and verify their keys with our API. The user will be able to verify if previously generated key is valid, and re-generate one if it isn't. If key generation fails because their account has too many keys already, the user will be shown a button that takes him to the website. If the app window is closed when automatic key generation fails, a notification will be shown. The notification support is a bit crude as they won't take the user to the appropriate key management screen.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/965)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Line wrap the file at 100 chars.                                              Th
 ## [Unreleased]
 ### Added
 - Add a switch to turn off system notifications under Preferences in the GUI.
+- Add new settings page for generating and verifying wireguard keys.
 
 #### Windows
 - Add migration logic to restore lost settings after major Windows update.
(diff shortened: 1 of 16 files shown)


jvff:
___
*[mullvad-daemon/src/lib.rs, line 1187 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/965#-LiyG8LF0RH8XrBC-EFn:-LiyG8LG6D43KcnnRVmv:bu59f7c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7f4fc1479e4f765bc948c186bb809dc54eb0c7bd/mullvad-daemon/src/lib.rs#L1187)):*
> ```Rust
>         match result() {
>             Ok(key_event) => {
>                 self.
> ```

This looks like it's incomplete?

PR #981, opened 2019-07-18
Create build script for Android release builds

jvff:
This PR creates a script to help build the Android APKs. The script will automatically build `mullvad-jni` for all supported ABIs, and copy the resulting shared library along with `libwg.so` to a directory where Gradle can pack them into the APK. The final APK is then copied to the `dist` directory with the version name.

The `README` was updated to describe how to set-up the environment for all supported ABIs and to use the script.

The script supports building development builds using `--dev-build`. These are built faster and are automatically signed with a debug key, making the development process simpler. The generated APKs have a `-debug` suffix.

In order to build a "release" APK, it has to be signed. This is done automatically, as long as the signing keys are configured according to the README.

The PR also includes some ProGuard rules to not obfuscate classes that `mullvad-jni` uses, so that in release builds they can still be found.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **No user-visible changes included.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/981)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,5 +17,6 @@
 /dist-assets/openvpn
 /dist-assets/openvpn.exe
 /windows/**/bin/
+/android/local.properties
 **/.vs/
 *.bak
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 /dist-assets/openvpn
 /dist-assets/openvpn.exe
 /windows/**/bin/
+/android/keystore.properties
 /android/local.properties
 **/.vs/
 *.bak
--- a/README.md
+++ b/README.md
@@ -96,12 +96,30 @@ unzip android-ndk-r20-linux-x86_64.zip
   --platform=android-21 \
   --arch=arm64 \
   --install-dir=$PWD/toolchains/android21-aarch64
+./android-ndk-r20/build/tools/make-standalone-toolchain.sh \
+  --platform=android-21 \
+  --arch=arm \
+  --install-dir=$PWD/toolchains/android21-armv7
+./android-ndk-r20/build/tools/make-standalone-toolchain.sh \
+  --platform=android-21 \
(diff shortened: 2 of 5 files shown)


jvff:
___
*[android/build.sh, line 5 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/981#-Lk4OIkzEeBauyVbUAYR:-Lk4PEA1B1cmlIuzrWzL:b-v0emyp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2535252a0f64332a148344d60ec19705ca5ac78b/android/build.sh#L5)):*

We have been through a couple of solutions for this problem before. And what we have in `build.sh` now seems to be the most portable and cross platform compatible. So can't we just re-use what we have learned before?:

```bash
SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
cd "$SCRIPT_DIR"
```
</blockquote></details>

Can I just use `cd "$(dirname "${BASH_SOURCE[0]}")"`? I don't think it makes sense to `cd` twice if I'm not storing the path anywhere.

jvff:
___
*[android/build.sh, line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/981#-Lk4Or48BiuGrYGJUdOC:-Lk4QOXx1aSXTnXjYr4Y:b3xwt5c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2535252a0f64332a148344d60ec19705ca5ac78b/android/build.sh#L27)):*

We could potentially do the refactoring later. Just saying we might want to consider it. And if we keep the style the same now already that will be a lot easier when we do it.
</blockquote></details>

So, should I change all variables to be upper case?

PR #991, opened 2019-07-26
Automatically retry key generation when Android app resumes

jvff:
This PR is a minor UX improvement. If the user has too many keys registered to his account, he currently has to go to the web-site to revoke some keys. After doing so, when he returns the app it would previously still think there are too many keys, so the user would have to log out and log in again for the changes to take effect.

This PR implements automatic retries to generate the WireGuard key if it had previously failed because there were too many keys. The retries occur every time the app is resumed, so this aligns with the user returning to the app.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/991)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/MainActivity.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/MainActivity.kt
@@ -87,6 +87,11 @@ class MainActivity : FragmentActivity() {
         }
     }
 
+    override fun onResume() {
+        super.onResume()
+        keyStatusListener.onResume()
+    }
+
     override fun onStop() {
         unbindService(serviceConnection)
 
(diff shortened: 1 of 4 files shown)


jvff:
___
*[mullvad-jni/src/lib.rs, line 255 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/991#-LkiYidBACYaoU4LfyTl:-Lkwyq2O3AC5neh3doK-:bgpvyen) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8497d6a6ceb1b6085be1855b0d61840045afa649/mullvad-jni/src/lib.rs#L255)):*

I think calling `.expect()` here isn't ideal either :/ You should still log this, my original issue was that the log message could be more specific, not that this should be handled differently. I'll be more explicit the next time around.
</blockquote></details>

But the internal RPC should never fail unless the daemon thread stops, right? If that happens, we might just abort anyway.

Or do you think there's a way for the call to return an error that's still recoverable from?

PR #1058, opened 2019-08-26
Drop packets when in the blocking state

jvff:
This PR implements a simple firewall on Android using the platform VPN service API. Basically it opens a fake VPN tunnel that will intercept all packets and drop them, so that effectively no packets are sent from the device to the internet.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1058)


jvff:
___
*[mullvad-jni/src/vpn_service_tun_provider.rs, line 92 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1058#-LoQN6gI-dmAi3GFrn3W:-LoQPJAh5L57053xFJIT:bavelro) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/176c448949a2a167ea9bcaf1bddca2c220e39e7c/mullvad-jni/src/vpn_service_tun_provider.rs#L92)):*

`tun` here is a `File`, so it's dropped when returning from `get_tun`. I can't see you pass the ownership anywhere. Seems like the rest of this method only need the raw fd. Could not `prepare_tun` just return the raw fd directly so we don't have to worry about that?
</blockquote></details>

It's not actually, dropped. `prepare_tun` returns `&File`.

jvff:
___
*[mullvad-jni/src/vpn_service_tun_provider.rs, line 124 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1058#-LoQPHNzEt6lA75QXFOY:-LoQPidk9TTLED1uvXaR:bmxpwty) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/176c448949a2a167ea9bcaf1bddca2c220e39e7c/mullvad-jni/src/vpn_service_tun_provider.rs#L124)):*

Should this not be named `create_tun` to be consistent with the others? Same as for the logging errors, "tun" her refers to the tun adapter, not the tunnel, right?
</blockquote></details>

The issue is that the trait already has a `create_tun` method :/

PR #1062, opened 2019-08-28
Switch build pipeline to Gulp

pronebird:

Git checklist:

This PR migrates the project to a better build pipeline that we can customise and adjust for our needs.

I've looked at both Webpack and Gulp once more and decided to go with Gulp for its simplicity and minimalism.

1. Add tasks for building and running GUI in development
1. Add `gulp-envify` plugin to turn all references to `process.NODE_ENV` into `"production"`. This is useful because some users have `NODE_ENV` set in their environment, which causes the app to crash because it attempts to load the extensions used in development only, which are not bundled with the production app. As of now this would result in a dead code in production, but if we use some sort of JS minifier, we could perhaps shake it off. But that's for later.
1. Remove the use of `process.env.BROWSER_SYNC_CLIENT_URL` in `index.html` file to load hot-reload script in a browser process. HTML pre-processing is used instead to inject the script when in development.
1. Add `Content-Security-Policy` meta tag in `index.html` to whitelist the scripts that browser is allowed to execute. Unfortunately `unsafe-inline` still has to be used because Electron injects some scripts itself to throw security warnings. `unsafe-inline` allows the use of `<script>some-code</script>` inside of HTML documents.
1. Explicitly enabled `nodeIntegration` for the browser window to fix the warning displayed in console. Electron 5 and onwards have node integration turned off by default.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1062)


gui/tasks/electron.js
```diff
@@ -0,0 +1,23 @@
...
+
+function startElectron(done) {
+  subprocess = spawn(electron, ['.'], {
+    env: { ...process.env, NODE_ENV: 'development' },
+    stdio: 'inherit',
+  });
+  done();
+}
+
+function stopElectron() {
```
pronebird: FYI Gulp tasks automatically handle streams, subprocesses and promises when they are returned back form tasks. i.e it waits for async task to complete. The other way to handle this is to accept a callback argument and call it later.

gui/tasks/scripts.js
```diff
@@ -0,0 +1,32 @@
...
+    watch.start('--noClear', '--sourceMap', '--project', '.');
+    return watch.tsc;
+  };
+  compileScripts.displayName = 'compile-scripts-watch';
+
+  return compileScripts;
+}
+
+function compileScripts() {
+  const tsProject = ts.createProject('tsconfig.json');
```
pronebird: FYI Too bad `gulp-typescript` does not support incremental compilation for TS, so I had to stick with `TscWatchClient` for development builds, however we still can use `gulp-typescript` for making release builds.

gui/tsconfig.json
```diff
@@ -28,6 +28,7 @@
   ],
   "include": [
     "src/**/*.ts",
-    "src/**/*.tsx"
+    "src/**/*.tsx",
+    "assets/geo/*.json"
```
pronebird: FYI Had to mention JSON files because `gulp-typescript` couldn't see them. When using `tsc` direction that wasn't a problem. Either way...

PR #1100, opened 2019-09-13
Add bridge location reset

pronebird:

Git checklist:

1. Reset the bridge constraints to `any` when bridge state is changed to `auto` or `off`
1.  Hack together the `Closest to exit server` option at the top of the list of bridge locations. The equivalent for that is setting the bridge constraints to `any`.

### DIFF
--- a/gui/src/shared/relay-location-builder.ts
+++ b/gui/src/shared/relay-location-builder.ts
@@ -1,0 +1,41 @@
+import { Constraint, LiftedConstraint, RelayLocation } from './daemon-rpc-types';
+
+export interface ILocationBuilder<Self> {
+  country(country: string): Self;
+  city(country: string, city: string): Self;
+  hostname(country: string, city: string, hostname: string): Self;
+  any(): Self;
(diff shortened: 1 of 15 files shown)


gui/src/shared/relay-location-builder.ts
```diff
@@ -0,0 +1,50 @@
...
+    hostname: (country: string, city: string, hostname: string) => {
+      receiver({ only: { hostname: [country, city, hostname] } });
+      return context;
+    },
+    any: () => {
+      receiver('any');
+      return context;
+    },
+    fromRaw(location: LiftedConstraint<RelayLocation>) {
+      if (location === 'any') {
```
jvff: Can't this be simplified using the new type information? I.e., could it be simplified to do something like:

```
if (location === 'any') {
    return this.any();
} else {
    receiver({ only: location });
    return context;
}
```

pronebird: Ah sure! Done!

PR #1102, opened 2019-09-13
Retry to create tunnel on bad file descriptor

jvff:
The Android `VpnService` sometimes provides a tun device with a bad file descriptor. This was previously handled correctly in `wireguard-go`, where a special error code was returned and the tunnel state machine's `ConnectingState` would automatically retry to connect.

When the Android blocking state was implemented, the tun file descriptor is duplicated in order to keep the tunnel open. However, calling `dup` on a bad file descriptor leads to an error and the connection fails, entering the blocked state.

Unfortunately, since the tunnel creation code is handled by a trait object (`&mut dyn TunProvider`), only a `BoxedError` instance can be returned to the `ConnectingState`. Therefore, returning a special error code to indicate that the connection should be reattempted would require changing the error type returned by `TunProvider::create_tun`.

A different solution was used for this PR. The `VpnServiceTunProvider` was updated to try to create the tun device for a specified number of attempts before failing. This means that opening a tun device for either a new tunnel connection or for entering the blocking state should tolerate occasional bad tun device file descriptors provided, and simply return an actual working file descriptor to a tun device or fail after the specified number of attempts.

The number of attempts is initially set to `4`, which should be safe enough for the error to appear only under extremely rare scenarios.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1102)

### DIFF
--- a/mullvad-jni/src/vpn_service_tun_provider.rs
+++ b/mullvad-jni/src/vpn_service_tun_provider.rs
@@ -95,7 +95,7 @@ impl VpnServiceTunProvider {
     }
 
     fn get_tun_fd(&mut self, config: TunConfig) -> Result<RawFd, Error> {
-        for _ in 0..MAX_PREPARE_TUN_ATTEMPTS {
+        for retry in 1..=MAX_PREPARE_TUN_ATTEMPTS {
             let tun = self.prepare_tun(config.clone())?;
 
             match Self::duplicate_tun(tun) {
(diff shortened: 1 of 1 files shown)


jvff:
*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1102#-LojbTpa-rRbVe1TCQFe:-LojbTpcBbyMBq4_bZ8t:bmytj7j) (no related file):*
Left a test device running in an infinite loop connecting and disconnecting every 10 seconds, and ran into the error again. It did retry successfully, but it still got a bad file descriptor on all retries. Will try to add a small delay to see if it fixes it.

---

jvff:
*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1102#-LojbTpa-rRbVe1TCQFe:-LopLAheByjtMhUCXGqm:b-wm5scd) (no related file):*

Is this not because the subsequent calls to `prepare_tun` won't actually create a new tunnel? Since there is one active and the config is identical.
</blockquote></details>

Yeah, I think that's the case. Nice catch, thanks! I'll leave it running again to test.

---

PR #1104, opened 2019-09-16
Update how the UI responds to predicted and real tunnel state changes

jvff:
This PR changes which parts of the UI change on predicted tunnel state changes, and which parts change on real tunnel state change events. A quick summary is:

1. Predicted tunnel state
- Action button
- Switch location button text

2. Real tunnel state
- Notification banner (showing "Blocking connections" when in Connecting state)
- Location information
- In/Out IP address
- Connection status message
- Header bar color
- Notification (both button and text)

As part of this PR, the race condition prevention code was slightly refactored. The code exists so that fetching the tunnel state for the first time doesn't overwrite any tunnel state that was received by an event from the daemon. Previously the real state would be set to `null` until an event changed it or the initial fetch successfully set it. Now, it's set to a fixed object reference representing an initial disconnected state, and while the `state` properties references the same initial object (compared using the `===` operator) it's considered as not initially set. Any change to this state sets it to a new object received externally, so the reference equality will fail after the state is set.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1104)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ConnectFragment.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ConnectFragment.kt
@@ -162,14 +162,14 @@ class ConnectFragment : Fragment() {
     private fun updateTunnelState(uiState: TunnelState, realState: TunnelState) =
         GlobalScope.launch(Dispatchers.Main)
     {
+        notificationBanner.tunnelState = realState
         locationInfoCache.state = realState
         locationInfo.state = realState
         headerBar.setState(realState)
+        status.setState(realState)
 
         actionButton.tunnelState = uiState
         switchLocationButton.state = uiState
-        notificationBanner.tunnelState = uiState
-        status.setState(uiState)
     }
 
     private fun updateKeyStatus(keyStatus: KeygenEvent?) = GlobalScope.launch(Dispatchers.Main) {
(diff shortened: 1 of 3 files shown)


jvff:
___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/dataproxy/ConnectionProxy.kt, line 135 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1104#-Lost5tJB71fGiz6YFXo:-LotfuSs8oiJBfCFNmIo:b4w92tr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b1e3965ac30c73abd880955dcc5e20a3430c30be/android/src/main/kotlin/net/mullvad/mullvadvpn/dataproxy/ConnectionProxy.kt#L135)):*

Ok. So part of my concern is not an issue. But the part about having a state which we don't know if it's the one the daemon has is still possibly there I guess?

It seems like we start a background job to get the state from the daemon. But nothing makes sure it actually responds/completes. I don't know what the UI looks like in this intermediate situation. But I assume it's going to go to the main view?

Would probably be good if we could stay on the "Connecting to Mullvad system service" until we established connectivity with the daemon and got all initial state data, so we know what we render from there on is accurate. But that might of course be way out of scope for this PR. Just want to bring up the topic.
</blockquote></details>

Sure. We can do that in a separate PR.

PR #1106, opened 2019-09-16
Reset state if anticipated state doesn't arrive

jvff:
The UI will anticipate certain tunnel states (Connecting and Disconnecting) when certain UI actions are performed in order to provide feedback as fast as possible. However, if any issue occurs when sending the command or when the daemon is handling the command, the anticipated state will never arrive. Therefore, this PR adds a timeout after setting the UI state to an anticipated state. If the time expires without receiving a tunnel state event from the daemon, the UI state is restored to the previous known tunnel state.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1106)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/dataproxy/ConnectionProxy.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/dataproxy/ConnectionProxy.kt
@@ -32,7 +32,7 @@ class ConnectionProxy(val context: Context, val daemon: Deferred<MullvadDaemon>)
     private val initialState: TunnelState = TunnelState.Disconnected()
 
     var state = initialState
-        set(value) {
+        private set(value) {
             field = value
             resetAnticipatedStateJob?.cancel()
             onStateChange.notify(value)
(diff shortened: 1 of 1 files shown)


jvff:
___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/dataproxy/ConnectionProxy.kt, line 118 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1106#-Lou0rNM8tDAiE54lXxb:-Lou35uZ29p8QznuCHIR:bqx6sej) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c56624ec145d8bd0fb230819b3fbdfbcdc00d7bc/android/src/main/kotlin/net/mullvad/mullvadvpn/dataproxy/ConnectionProxy.kt#L118)):*

Ok, cool! Just wanted to make sure!
</blockquote></details>

The reset runs on the main thread, but the tunnel state events are received on a background thread.

However, we're only updating the `uiState`, which is either updated here or updated on a new tunnel state event _after_ this `resetAnticipatedJob` is canceled. Worst case is that it cancels right after the delay (the only place where it will cancel and stop execution) and the `uiState` gets set to the `state`, which was already updated before calling `cancel` and which will be propagated to `uiState` again afterwards. At least AFAICT, the worst case scenario is that `uiState` is set twice to the same value.

PR #1117, opened 2019-09-18
Windows: Fix installer registry paths

mvd-ows:
As above.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1117)

### DIFF
--- a/dist-assets/windows/installer.nsh
+++ b/dist-assets/windows/installer.nsh
@@ -471,21 +471,33 @@
 !macro customInit
 
 	Push $0
-	
-	registry::MoveKey "HKLM\SOFTWARE\8fa2c331-e09e-5709-bc74-c59df61f0c7e" "HKLM\SOFTWARE\Mullvad VPN"
+
+	# Application settings key
+	# Migrate 2018.(x<6) to current
+	registry::MoveKey "HKLM\SOFTWARE\8fa2c331-e09e-5709-bc74-c59df61f0c7e" "HKLM\SOFTWARE\${PRODUCT_NAME}"
 
 	# Discard return value
 	Pop $0
 	Pop $0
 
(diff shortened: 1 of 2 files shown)


mvd-ows:
___
*[dist-assets/windows/installer.nsh, line 493 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1117#-Lp3DWwh8JcpV03y3dDM:-Lp3DWwh8JcpV03y3dDN:b-nm3oh0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96e74c09aff0aa4129c5badcf86a239be2a508d3/dist-assets/windows/installer.nsh#L493)):*
> ```NSIS
> 	# Application uninstall key
> 	# Migrate 2018.6 through 2019.7 to current
> 	registry::MoveKey "HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Mullvad VPN" "HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{${APP_GUID}}"
> ```

There is a different variable that would be more suitable than using `APP_GUID`. The variable I'm thinking of has the full path we need, but it's being defined/assigned in NSIS as opposed to before NSIS is started. So there's the question of which happens first, the variable assignment or our migration logic being activated. This felt safer.

mvd-ows:
___
*[gui/tasks/distribution.js, line 63 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1117#-Lp3E-O4DZWLtTRre0iq:-Lp3E-O4DZWLtTRre0ir:b-8svfab) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96e74c09aff0aa4129c5badcf86a239be2a508d3/gui/tasks/distribution.js#L63)):*
> ```JavaScript
> 
>   nsis: {
>     guid: '2A356FD4-03B7-4F45-99B4-737BE580DC82',
> ```

This GUID can be anything and there is no obvious advantage to using the one `electron-builder` generates for us. If anything, we could end up in a similar situation as the one we're trying to solve, if ever they decide to generate it differently.

PR #1124, opened 2019-09-19
Add RelayCache

pronebird:

Git checklist:

This PR adds a relay cache that stores relays on disk that maintains a `relays.json` in a shared container on iOS. 

The `relays.json` file is automatically updated every hour upon the read access, it felt like the easiest way to keep it up to date so far because both processes (the tunnel and the GUI app) will be accessing it so there is really no central place to keep it up to date.

A shared container is supposed to be a sandbox extension for the file system and can be accessed by both processes (Tunnel and the GUI app). I utilise a file coordinator to make sure that all reads and writes are atomic across processes.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1124)

### DIFF
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -42,6 +42,11 @@
 		58BFA5C022A7C8A900A6173D /* MullvadAPI.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58ADDB3D227B1CD900FAFEA7 /* MullvadAPI.swift */; };
 		58BFA5C122A7C92400A6173D /* JsonRequestProcedure.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58F19E30228B2AEB00C7710B /* JsonRequestProcedure.swift */; };
 		58BFA5C222A7C92900A6173D /* JsonRpc.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58ADDB3B227B1BD200FAFEA7 /* JsonRpc.swift */; };
+		58BFA5C322A7C93400A6173D /* RelayList.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5888AD88227B18C40051EB06 /* RelayList.swift */; };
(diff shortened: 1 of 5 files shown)


ios/MullvadVPN.xcodeproj/project.pbxproj
```diff
@@ -800,6 +813,7 @@
 				CODE_SIGN_ENTITLEMENTS = PacketTunnel/PacketTunnel.entitlements;
 				CODE_SIGN_STYLE = Automatic;
 				DEVELOPMENT_TEAM = G7CDBEG477;
+				ENABLE_BITCODE = NO;
```
pronebird: FYI had to disable bitcode because I had some issues with Go and compiling it with bitcode on.

ios/MullvadVPN/ApplicationConfiguration.swift
```diff
@@ -0,0 +1,18 @@
...
+//  ApplicationConfiguration.swift
+//  MullvadVPN
+//
+//  Created by pronebird on 05/06/2019.
+//  Copyright © 2019 Amagicom AB. All rights reserved.
+//
+
+import Foundation
+
+class ApplicationConfiguration {
```
pronebird: FYI Identifiers here are hardcoded and there is no any other way to access them at runtime.

- `securityGroupIdentifier` is used for accessing a shared container. The FileManager on iOS returns a container path for the given security group identifier. Both tunnel and the GUI app are in this security group so they both can access a shared container on the filesystem, but they need this identifier in our to fetch the right location on disk.

- `packetTunnelExtensionIdentifier` will be needed later to configure tunnels.

ios/MullvadVPN/RelayList.swift
```diff
@@ -8,27 +8,40 @@
 //
 
 import Foundation
+import Network
 
-struct RelayList: Decodable {
-    struct Country: Decodable {
+struct RelayList: Codable {
```
pronebird: FYI turned this into `Codable` because it has to be cached on disk, hence must be serialisable.

PR #1161, opened 2019-10-02
Use authentication tokens for account and wireguard keys pages.

pinkisemils:
This PR changes the GUI so that the URLs for the account and wireguard keys pages will be opened with the `token` URL parameter populated with a token received from the daemon. The net result is that the user will already be logged in when they will open these links.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1161)

### DIFF
--- a/gui/src/main/daemon-rpc.ts
+++ b/gui/src/main/daemon-rpc.ts
@@ -428,6 +428,15 @@ export class DaemonRpc {
     }
   }
 
+  public async getWwwAuthToken(): Promise<string> {
+    const response = await this.transport.send('get_www_auth_token');
+    try {
+      return validate(string, response);
+    } catch (error) {
+      throw new ResponseParseError('Invalid response from get_www_auth_token', error);
+    }
+  }
+
   public async getRelayLocations(): Promise<IRelayList> {
     const response = await this.transport.send('get_relay_locations');
     try {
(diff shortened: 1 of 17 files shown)


gui/src/main/daemon-rpc.ts
```diff
@@ -428,6 +428,15 @@ export class DaemonRpc {
     }
   }
 
+  public async getWwwAuthToken(): Promise<string> {
+    const response = await this.transport.send('get_www_auth_token');
+    try {
+      return validate(string, response);
+    } catch (error) {
+      throw new ResponseParseError('Invalid response from get_account_data', error);
```
pronebird: Typo: `get_account_data` in the comment

gui/src/renderer/app.tsx
```diff
@@ -285,6 +285,16 @@ export default class AppRenderer {
     return IpcRendererEventChannel.accountHistory.removeItem(accountToken);
   }
 
+  public async openLinkWithAuth(link: string) {
+    let token;
+    try {
+      token = await IpcRendererEventChannel.account.getWwwAuthToken();
+    } catch (e) {
+      token = '';
```
pronebird: Do we need to log errors here?

```
log.error(`Failed to get the WWW auth token: ${e.message}`)
```

PR #1183, opened 2019-10-08
Install nseventmonitor from NPM

pronebird:

Git checklist:

Since @DMarby published nseventmonitor for us, we can actually now reference it directly from npm which should enable us to use `npm ci` to install deps on CI.

https://www.npmjs.com/package/nseventmonitor

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1183)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,15 @@ matrix:
   include:
     # GUI - Electron
     - language: node_js
-      node_js: '8'
+      node_js: '12'
       cache: npm
       services:
         - xvfb
+      before_install:
+        - npm install -g npm
       install:
         - cd gui
-        - npm install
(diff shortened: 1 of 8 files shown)


pronebird:
@faern I think `npm` is being silly. Let me remove the dependency and reinstall it.

pronebird:
I've updated the PR to use `npm ci` in place of `npm install` in build scripts and CI. The difference is that `npm ci` installs all dependencies at once – it always wipes the existing `node_modules/` before installation. It will fail if `package-lock.json` is missing.

pronebird:
Lock file has been updated too, looks correct. Good spot @faern !

PR #1217, opened 2019-10-24
Perform installation of Rust in one place

faern:
While I was tinkering with #1216 I also thought it was quite inconsistent to install rustup so far from each other depending on platform. When we did it directly in `.travis.yml` for Windows that was still going to be done in a git bash shell, so it could just as well be moved to our CI script to make things more even.

The `-c rustfmt` part is moved from the rustup install to the `rustup toolchain install` step. Apparently this is where the flag can automatically backtrack to last known good toolchain with said component. It did not do that if provided directly to the rustup install script I learned from some IRC conversations.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1217)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -127,11 +127,6 @@ matrix:
       name: Daemon, Windows - stable Rust
       env: RUST_VERSION=stable
       language: bash
-      install: &rust_windows_install
-        - curl -sSf -o rustup-init.exe https://win.rustup.rs/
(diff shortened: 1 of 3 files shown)


faern:
I tried locally and I can reproduce the odd behavior. `--default-toolchain none` seems to be what causes it. I reported a bug: https://github.com/rust-lang/rustup.rs/issues/2082

PR #1247, opened 2019-11-19
Idiomatic duration constants

faern:
This fixes the formatting error on https://travis-ci.com/mullvad/mullvadvpn-app/jobs/258138656#L230

And it also tries to be a bit more idiomatic on storing times as structured types instead of integers. That way the variable name can reflect what the time is supposed to represent, instead of the time unit and how long time it is. I'm not sure if the current names are perfect. Because I did not know exactly what they represented, which is exactly the problem I'm trying to address here :)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1247)


faern:
The formatting error has already been fixed in another PR. Anyhow, the delays could still make use of descriptive names, so the PR still has value.

PR #1283, opened 2019-11-29
Windows: Unify logging

mvd-ows:
As above.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1283)


mvd-ows:
___
*[mullvad-daemon/Cargo.toml, line 58 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1283#-LurKAEK9Bw92jFgZB1N:-LusnY3UEf-f7RcPdhm-:b-x2wt4b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/263e03953a644b2c8cf06849c92d22218e1dd060/mullvad-daemon/Cargo.toml#L58)):*

There's no reason to define the full version, if you just specify `0.2` then we'll be able to seamlessly use newer versions when we manually update them. And since this dependency is already listed for _unix_ targets, we might as well define it once for all targets.
</blockquote></details>

OK, makes sense. `0.2.20` was a copy-paste from some other `Cargo.toml`we have.

PR #1300, opened 2019-12-04
Windows: Register NSI as a service dependency

mvd-ows:
Mark NSI as a dependency of the Daemon service. The WireGuard project suggested this, and after looking into what NSI does ("This service delivers network notifications (e.g. interface addition/deleting etc)."), it makes sense to add it.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1300)


mvd-ows:
Including this fix in another larger PR.

PR #1303, opened 2019-12-04
Swap blocking tunnel state with error state

pinkisemils:
To be more explicit about what state the daemon currently resides when the tunnel state machine encounters an unrecoverable error, the blocking state has been swapped out for an error state which contains a flag indicating whether traffic is being blocked or not. This will allow the daemon's clients (our GUI and Android app) to better signal the security guarantees the app is or is not providing at any given moment.
The current changes made to the android code as far as user visible strings go are just placeholders and should be revised when the UI will be adjusted to show the error state appropriately. It should be incredibly obvious that the app is not blocking any traffic when this occurs.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1303)


jvff:
___
*[talpid-types/src/tunnel.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1303#-LvIRvPk5pnRnhtJGVUR:-LvKfAZMA4ubVIkczAl6:bemomd3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b1a3f196e4a300547911c4ac5a36d1aca6e494c/talpid-types/src/tunnel.rs#L78)):*

I'm thinking about it too, but I don't think `ErrorState` is a good name for this either - it's not an error that we're blocking because the host is offline.
</blockquote></details>

Well, you could kind of interpret that trying to create a tunnel on a host without internet is an error, but I agree it is stretching the meaning a bit.

PR #1315, opened 2019-12-06
Stop CVE-2019-14899 by dropping packets to tunnel IP

faern:
Stops an attacker on the same network from discovering the tunnel IP of the device running this app.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1315)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Line wrap the file at 100 chars.                                              Th
 ### Fixed
 - Fix improved WireGuard port selection
 
+### Security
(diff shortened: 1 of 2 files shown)


faern:
It's worth mentioning that our app is **only vulnerable to stage 1** of the attack: Figuring out the tunnel IP. It has never been possible to carry out the remaining steps, to find open TCP connections, sequence numbers or inject traffic. And this attack was **only possible towards our app if "Local network sharing" was enabled**, which it's not by default.

Issue #1281, opened 2019-11-29
Mullvad CLI Windows 10


rdebruin:
Mullvad CLI cannot be started from a DOS box. Reason: C:\Program Files\Mullvad VPN\resources is not added to the PATH.

C:\Users\ruudd>mullvad
'mullvad' is not recognized as an internal or external command,
operable program or batch file.

Current PATH with Mullvad Release version 2019.9 version installed:

C:\Users\ruudd>path
PATH=C:\Program Files (x86)\Intel\Intel(R) Management Engine Components\iCLS\;C:\Program Files\Intel\Intel(R) Management Engine Components\iCLS\;C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\WINDOWS\System32\OpenSSH\;C:\Program Files\Intel\WiFi\bin\;C:\Program Files\Common Files\Intel\WirelessCommon\;C:\Program Files (x86)\Intel\Intel(R) Management Engine Components\DAL;C:\Program Files\Intel\Intel(R) Management Engine Components\DAL;C:\Program Files (x86)\Intel\Intel(R) Management Engine Components\IPT;C:\Program Files\Intel\Intel(R) Management Engine Components\IPT;C:\Users\ruudd\AppData\Local\Microsoft\WindowsApps

Using Windows 10 Pro (latest version)

faern:
When you say dos box, do you mean the "command prompt" (`cmd.exe`)?

Works for me here. So I can't reproduce the issue. Please go via support for help using the app. You can send a problem report from within the app (settings cog wheel icon -> Report a problem), that way support will have access to the installation log and might be able to diagnose the problem.

rdebruin:
Yes, the "command prompt". Installed it multiple times with the same result
from scratch. I have send a problem report.

Op vr 29 nov. 2019 om 09:59 schreef Linus Färnstrand <
notifications@github.com>:

rdebruin:
I have send the report two times, but no answer yet. Installed Mullvad on my "work laptop" and on another laptop with a fresh copy of Windows 10 Pro installed. No other programs involved. In both cases Mullvad is not added to the PATH so I can't run Mullvad CLI from a DOS command prompt. Of course, it runs fine when starting it from the Mullvad directory itself,

faern:
[...]

PR #1401, opened 2020-01-17
Fix system PATH update

dlon:
On some systems, our changes to the system PATH are not reflected in `cmd` when it is run from Win+R. In this case, PATH is inherited from `explorer.exe` rather than read directly from the registry. We already notify `explorer.exe` of the change, which works for some users but not others.
The main reason seems to be that the installer would sometimes restart `explorer.exe` before the PATH was updated, which could result in a race condition. There may also have been a problem relating to Unicode.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1401)

### DIFF
--- a/dist-assets/windows/installer.nsh
+++ b/dist-assets/windows/installer.nsh
@@ -850,10 +850,9 @@
 		${BreakInstallation}
 		Abort
 	${EndIf}
-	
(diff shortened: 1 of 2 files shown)


faern:
Fixes #1281

PR #1499, opened 2020-02-16
Fit long location name

raksooo:
This PR changes the city/country name display to one line for country and one for city. If the name does not fit it slowly scrolls from left to right with a marquee effect.

I'm not sure why but the gif below is a little bit slower and more laggy than the actual effect.

output

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1499)


gui/src/renderer/components/Marquee.tsx
```diff
@@ -0,0 +1,63 @@
...
+          ref={this.textRef}
+          style={[styles.text, this.textAnimation, this.props.style]}>
+          {this.props.children}
+        </Animated.Text>
+      </View>
+    );
+  }
+
+  private async startAnimation() {
+    setTimeout(async () => {
```
pronebird: Save the timeout and clear it in `componentWillUnmount()` to make sure we don't run anything after the component is gone.

gui/src/renderer/components/Marquee.tsx
```diff
@@ -0,0 +1,63 @@
...
+        const textLayout = await UserInterface.measureLayoutRelativeToWindow(this.textRef.current);
+        const viewLayout = await UserInterface.measureLayoutRelativeToWindow(this);
+        this.startAnimationImpl(textLayout.width - viewLayout.width, false);
+      }
+    }, 1000);
+  }
+
+  private startAnimationImpl(length: number, reverse: boolean) {
+    if (length >= 0) {
+      Animated.timing(this.initialLeft, {
```
pronebird: Same here, store the returned `Types.Animated.CompositeAnimation` and call `anim.stop()` from `componentWillUnmount()`. This will prevent the whole thing from crashing on `UserInterface.measureLayoutRelativeToWindow`

gui/src/renderer/components/Marquee.tsx
```diff
@@ -0,0 +1,63 @@
...
+        toValue: reverse ? 0.0 : -length,
+        duration: length * 80,
+        delay: 2000,
+        easing: Animated.Easing.Linear(),
+      }).start(() => {
```
pronebird: `start` passes the options dictionary `{ finished: bool }` as the first argument to the closure that. you pass to `.start(closure)`. You can use `finished` to avoid repeating animation when the component is about to go away, otherwise it's likely to cause an infinite loop?

pronebird: Alternatively it's possible to use a flag to detect if the component is still mounted, i.e:

```
componentDidMount() {
  this.isMounted = true
}

componentWillUnmount() {
  this.isMounted = false
}
```

PR #1518, opened 2020-02-20
Improve prepare release

faern:
Slightly improve the `prepare_release.sh` script. This script is used to bump the version in all relevant files and add a signed git tag.

The `cargo build` part was only needed because it indirectly updates `Cargo.lock` with the new versions. But we did not actually need to build anything. So having it behind a timer solves the issue of having to wait a long time. Ugly, but I don't have a better solution.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1518)

### DIFF
--- a/prepare_release.sh
+++ b/prepare_release.sh
@@ -28,7 +28,13 @@ echo "Updating version in metadata files..."
 
 echo "Syncing Cargo.lock with new version numbers"
 source env.sh ""
-cargo +stable build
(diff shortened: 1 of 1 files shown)


dlon:
Isn't it possible for `cargo build` to fail before it's generated Cargo.lock? It might make more sense to check if the status code was 124?

PR #1519, opened 2020-02-20
Add promise rules to ESLint

raksooo:
This PR adds eslint rules `require-await` and `@typescript-eslint/no-floating-promises`. In addition it removes the `import` and `promise` plugins.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1519)


gui/src/main/index.ts
```diff
@@ -582,10 +582,10 @@ class ApplicationMain {
     }
   }
 
-  private setTunnelState(newState: TunnelState) {
+  private async setTunnelState(newState: TunnelState) {
     this.tunnelState = newState;
     this.updateTrayIcon(newState, this.settings.blockWhenDisconnected);
-    this.updateLocation();
+    await this.updateLocation();
```
pronebird: This call may take up to 10 seconds to complete as it calls `daemonRpc.getLocation()`. We should never truly wait for it to complete. It may as well fail.

gui/src/main/daemon-rpc.ts
```diff
@@ -551,12 +551,14 @@ export class DaemonRpc {
...
 
-  public async unsubscribeDaemonEventListener(
+  public unsubscribeDaemonEventListener(
     listener: SubscriptionListener<DaemonEvent>,
   ): Promise<void> {
     if (listener.subscriptionId) {
       return this.transport.unsubscribe('daemon_event', listener.subscriptionId);
     }
+
+    return Promise.resolve();
```
pronebird: FYI this is exactly what `async` keyword does. It also unwraps the `Promise` if you return one from `async` function, i.e if we were to translate the effect of `async` keyword into code I think it looks something like this:

```
function somethingAsync() {
  return new Promise((resolve, reject) => {
      const maybePromise = gutsOfYourAsyncFunc()

      if (isPromise(maybePromise)) {
          maybePromise.then(resolve).catch(reject)
     } else {
          resolve(maybePromise)
     }
  })
}
```

gui/src/renderer/components/NotificationBanner.tsx
```diff
@@ -213,6 +213,7 @@ export class NotificationBanner extends Component<
     if (prevProps.visible !== this.props.visible) {
       // enable drawer-like animation when changing banner's visibility
       this.setState({ contentPinnedToBottom: true }, () => {
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
         this.animateHeightChanges();
```
pronebird: You `await` in here but don't `await` below. 

Maybe just wrap this call into something like:

```
animateHeightChangesHelper() {
   this.animateHeightChanges().then(() => {})
      .catch((_error) => {})
}
```

raksooo: There's also an eslint rule for not having empty functions which would complain about this.

PR #1533, opened 2020-02-23
Minor change in README.md

sm0rux:
As Mullvad app for Android now handles local network sharing I think another checkbox should be marked in README.md.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1533)


faern:
Thanks for the PR! Yes it's correct that this checkbox should be ticked now.

I had to merge via github and then amend and sign the commit. Since you did not have a fork it was not possible for me to get the commit from anywhere :thinking: and we manually merge and sign everything. But it was doable.

PR #1586, opened 2020-03-20
Fix paths to script

dlon:
Some scripts were not updated when a file was renamed.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1586)


faern:
This was fixed by #1584

PR #1594, opened 2020-03-26
Render ModalAlert in portal

raksooo:
This PR adds a container for modals to be rendered in and renders them in that element using portals.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1594)

### DIFF
--- a/gui/src/renderer/components/Modal.tsx
+++ b/gui/src/renderer/components/Modal.tsx
@@ -1,7 +1,10 @@
 import * as React from 'react';
+import ReactDOM from 'react-dom';
 import { Component, Styles, Text, View } from 'reactxp';
 import { colors } from '../../config.json';
 import ImageView from './ImageView';
+
+const MODAL_CONTAINER_ID = 'modalContainer';
 
 const styles = {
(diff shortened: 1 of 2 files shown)


gui/src/renderer/components/Modal.tsx
```diff
@@ -98,6 +106,16 @@ interface IModalAlertProps {
 
 export class ModalAlert extends Component<IModalAlertProps> {
   public render() {
+    const modalContainer = document.getElementById(MODAL_CONTAINER_ID);
+    if (modalContainer !== null) {
+      return ReactDOM.createPortal(this.renderModal(), modalContainer);
+    } else {
+      log.error('Modal container not found when rendering modal');
```
pronebird: You can `throw Error("Modal container not found when rendering modal")`, we usually catch those in the error boundary and log them automatically. Also, this is likely a critical error which should not happen.

raksooo: Okay :+1: Should critical errors be handled differently? Should I skip the `modalContainer !== null` check?

PR #1609, opened 2020-03-27
Start with correct menubar icon

raksooo:
If the app is started when the daemon is already connected, the menubar icon animates from disconnected to connected. This gives the impression that the user wasn't already connected. This PR solves that by starting the app with the correct menubar icon.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1609)


raksooo:
New, better solution to the same problem in #1613

PR #1662, opened 2020-04-16
Fix crash when running from SD card or from container

jvff:
The app would previously crash on Android if it was installed on an SD card or on a container (using for example Android For Word). This happened because the app relied on hard-coded paths for certain things like the settings file, the relay list file, and the log files.

This PR fixes the issue by using the app's cache directory and the app's files directory (as determined by the Android system).

As part of the change, the `Settings` type was refactored so that the file operations were moved from `mullvad-types` to `mullvad-daemon`, and the settings file path can now be specified through a parameter.

If a settings file is found in the old location, it is moved to the new location before the daemon is started.

Still to be done before sending the PR to review:
- [x] Update the problem report handling to use the correct new log directories
- [x] Delete the old file paths when starting to avoid having duplicates

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1662)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ Line wrap the file at 100 chars.                                              Th
 
 #### Android
 - Change button colors on problem report no email confirmation dialog to match the desktop version.
+- Fix crash when attempting to run app from the non-default location, such as the SD card or from a
+  different user profile.
 
 ### Security
 #### macOS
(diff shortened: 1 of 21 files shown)


jvff:
___
*[mullvad-problem-report/src/lib.rs, line 103 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1662#-M56vLqD-eFCQQbtZ1Hi:-M57OhUY8vwe7A5oD14E:buh4335) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/10630ccfc0632bc0f674eb1326c1066b9765dcfe/mullvad-problem-report/src/lib.rs#L103)):*

No of course not. The binary would have to get all the dependencies for `collect_report` first and handle any errors in that before calling it.
</blockquote></details>

The problem is in how to handle the error. Currently, either the file is added to the problem report or the error to load it is (which includes the error on getting the file path). How would we include the error in the problem report if it's done out of this function?

PR #1680, opened 2020-04-24
Fix traffic leak when the app starts with "Auto-Connect" enabled

jvff:
Sometimes, when the app started with the "Auto-Connect" setting enabled, it would fail to create the tunnel device and enter the non-blocking error state. This happened because the `VpnService.prepare()` method wasn't called before trying to create the tunnel.

This is usually not an issue because the `MainActivity` will use the `ConnectionProxy` to connect, which does call `VpnService.prepare()` before sending a `connect` command to the daemon. However, if the daemon decides to connect by itself (like when the "Auto-Connect" setting is enabled), the method isn't called.

This PR fixes the issue by always calling `VpnService.prepare` when connecting. This means it does get called twice when connecting, but if the permission is already granted the calls just return `null`. However, if the permission isn't granted, then the code will now return a new error to indicate that the permission was denied. This new error is used to show a more specific error message in the notification banner.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1680)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,10 @@ Line wrap the file at 100 chars.                                              Th
   checks delayed app startup when "block when disconnected" was enabled and performed system network
   requests to Apple.
 
+#### Android
+- Fix failure to create tunnel when app is started with auto-connect enabled. This would sometimes
+  lead to a traffic leak.
+
 
 ## [2020.4-beta2] - 2020-04-08
 ### Added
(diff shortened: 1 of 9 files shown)


jvff:
___
*[android/src/main/res/values/strings.xml, line 124 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1680#-M5vkSKrAhF7WYRe_9JL:-M5vl5GLCBiQ4D26uU5d:bd7z1kp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9c352b1de65ff227cd6470ca075d8569d204a0ed/android/src/main/res/values/strings.xml#L124)):*

It should be noted that in this state, no blocking can take place.Saw
</blockquote></details>

Yep. The title still is the `not_blocking_internet` string.

PR #1707, opened 2020-05-04
Add ko translation

HyungJu:
This PR adds Korean Translation. Mullvad VPN has only few languages support and it could be disadvantage for korean users. This commits will solve that problem. 

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1707)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Line wrap the file at 100 chars.                                              Th
 
 ## [Unreleased]
 ### Added
+- Add Korean translation.
 #### Android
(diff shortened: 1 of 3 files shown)


faern:
Thanks for this hard work and for caring about our app and fellow Korean users. Sadly we can't accept translation submissions from third parties. We have no way of verifying what it says is correct. We do all app translating via professional translators and proof readers.

We have spoken internally about potential languages to add. The result is that we have now added Korean to the list of languages we will include next time we translate the app into new languages. So Korean support will come in a few releases!

PR #1720, opened 2020-05-07
Allow redeeming vouchers in the Welcome screen

jvff:
Previously the Welcome screen would only show a button to add more time that goes to the website. This PR also adds a button to redeem vouchers from within the app. When the button is pressed, a dialog is opened for the user to type in the voucher code.

The voucher input area is formatted so that hypens separate every four voucher code characters, like it is presented in the physical vouchers. This is done with a `SegmentedInputFormatter` class, which also handles the way Android notifies text changes and allows the input to be further changed without resetting the selection (or the cursor position).

One edge case is also handled: if text is removed and the selection bounds is after a separator (or, more commonly, the cursor is after a separator) then the bounds are changed to make sure it they are before the separator. This avoids the situation where a separator is removed from the middle of the text, and the formatter just adds it again moving the cursor to the original position and preventing the user from deleting more than one segment. When that happens, the separator is added again, but the cursor (or selection bound) is moved back to before the separator so further deletes can continue normally.

The formatter helper class may also be reused in the future to format the account input in the Login screen.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1720)


jvff:
___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/ui/RedeemVoucherDialogFragment.kt, line 134 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1720#-M6odOtb4YdsvePgMiTB:-M6ofCK0BT__bHt3Xr6O:bcv97zf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aae43cbcd79324a068b25a6c86fd718e99bfeb20/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/RedeemVoucherDialogFragment.kt#L134)):*

Shouldn't a lock be acquired here in case the view is being torn down just as the result has been received?
</blockquote></details>

Shouldn't be needed. The `submitVoucher` runs on the UI thread, and is managed by the `jobTracker`. If the view is destroyed, the job tracker will cancel all jobs, so it doesn't execute anything after the background thread submitting the voucher finishes.

PR #1759, opened 2020-05-15
Switch to using DispatchWallTime and DispatchSourceTimer in key rotation

pronebird:

Git checklist:

After doing some testing we figured that `DispatchTime` relies on device uptime which doesn't tick when the device is hibernated causing timers to fire much later than expected. With `DispatchWallTime` the timers that are due would fire immediately on wake up.

1. `DispatchTime` is relative to system clock that ticks since boot
1. `DispatchWallTime` - An absolute point in time according to the wall clock, with microsecond precision.

I also decided to go with `DispatchSourceTimer` instead of `DispatchQueue.asyncAfter` because I am not entirely sure that the timer created internally by `DispatchQueue.asyncAfter` is cancelled upon the work item (`DispatchWorkItem`) cancellation.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1759)

### DIFF
--- a/ios/MullvadVPN/AutomaticKeyRotationManager.swift
+++ b/ios/MullvadVPN/AutomaticKeyRotationManager.swift
@@ -50,22 +50,31 @@ class AutomaticKeyRotationManager {
     private let persistentKeychainReference: Data
     private var rotateKeySubscriber: AnyCancellable?
 
+    /// A dispatch queue used for synchronization
     private let dispatchQueue = DispatchQueue(label: "net.mullvad.vpn.key-manager", qos: .background)
-    private var retryWorkItem: DispatchWorkItem?
 
+    /// A timer source used to schedule a delayed key rotation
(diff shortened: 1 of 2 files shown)


ios/MullvadVPN/AutomaticKeyRotationManager.swift
```diff
@@ -130,26 +139,26 @@ class AutomaticKeyRotationManager {
...
+        timerSource.setEventHandler { [weak self] in
             self?.performKeyRotation()
         }
-        retryWorkItem = workItem
-        dispatchQueue.asyncAfter(deadline: deadline, execute: workItem)
+
+        timerSource.schedule(wallDeadline: wallDeadline)
+        timerSource.activate()
+
+        self.timerSource = timerSource
```
pronebird, line 161: FYI The timer seems to the properly cancelled on deallocation.

PR #1761, opened 2020-05-18
Update Android build scripts to add build for F-Droid

jvff:
This PR updates the Android build scripts so that the app can be built for distribution on F-Droid.

This required a few changes to the overall build process:

1. Remove the dependency on node-js to determine the Android version
2. Created a separate build type for F-Droid. This is a release build that is not signed, and can be built using `./build-apk.sh --fdroid`.
3. Checks if the `gradlew` script is present before using it, and falls back the system global `gradle` if it is installed.
4. Changed the WireGuard-Go build so that it can be executed without using the Docker image.
5. Modified the `version-metadata.sh` to skip the desktop-specific directories when building for Android.

Git checklist:

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -45,6 +45,14 @@ android {
         }
     }
 
+    buildTypes {
+        fdroid {
+            initWith release
+            minifyEnabled false
+            signingConfig null
+        }
+    }
+
     sourceSets {
         main {
             assets {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -83,7 +83,7 @@ buildscript {
         google()
 
         maven {
-            url = uri("https://plugins.gradle.org/m2/")
+            url "https://plugins.gradle.org/m2/"
         }
     }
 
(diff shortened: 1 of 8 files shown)


jvff:
___
*[build-apk.sh, line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1761#-M7bp6dl8bNqXgx-rUaz:-M7c-3m98TlEYUonAkn5:b-ph56on) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4df590698b216d6d3db5c1211b04921c8bd37476/build-apk.sh#L53)):*

There is no `gradlew` in the repo root. Do I miss a `cd` somewhere or does it exist in the root when building for F-Droid?
</blockquote></details>

No, this is a bug :/

Should I check for `"$SCRIPT_DIR/android/gradlew"` and use that as the command? Or should I just check for `"android/gradlew"` and keep using `/gradlew` as the command?

PR #1808, opened 2020-06-02
Switch to using ephemeral URLSession configuration

pronebird:

Git checklist:

Disable `URLSession` cache by switching to ephemeral configuration. As per docs the main advantage of using the ephemeral configuration is privacy to avoid any API responses to be cached on disk.

>An ephemeral session configuration object is similar to a default session configuration (see default), except that the corresponding session object doesn’t store caches, credential stores, or any session-related data to disk. Instead, session-related data is stored in RAM. The only time an ephemeral session writes data to disk is when you tell it to write the contents of a URL to a file.

https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1410529-ephemeral 

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1808)

### DIFF
--- a/ios/CHANGELOG.md
+++ b/ios/CHANGELOG.md
@@ -35,6 +35,7 @@ Line wrap the file at 100 chars.                                              Th
   and 90 days to always be displayed in days quantity.
 - Fix a number of errors in DNS64 resolution and IPv6 support.
 - Update the tunnel state when the app returns from suspended state.
+- Disable `URLSession` cache.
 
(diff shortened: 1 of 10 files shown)


faern:
This fixed audit finding [*MUL-02-001 iOS: Lack of filesystem protections (Info)*](https://github.com/mullvad/mullvadvpn-app/blob/master/audits/2020-06-12-cure53.md#miscellaneous-issues)

PR #1817, opened 2020-06-04
Fix routing on MacOS

pinkisemils:
I've changed the MacOS routing code to not use old futures and rely on tokio 0.2 and also fixed an issue with default route handling.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1817)


pinkisemils:
Submodules are hard.

PR #1819, opened 2020-06-04
Improve firewall rules on Linux

dlon:
Restrict the rule allowing traffic to the relay server to match on uid 0, since the daemon and its subprocesses run as root.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1819)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1682,19 +1682,18 @@ dependencies = [
 
 [[package]]
 name = "nftnl"
-version = "0.4.0"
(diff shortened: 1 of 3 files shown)


faern:
This fixed audit finding [MUL-02-002 WP2: Firewall allows deanonymization by eavesdropper (Medium)](https://github.com/mullvad/mullvadvpn-app/blob/master/audits/2020-06-12-cure53.md#identified-vulnerabilities)

PR #1822, opened 2020-06-05
Secure screens with potentially sensitive data

jvff:
Some screens show sensitive information such as the account number, the WireGuard key or the user's reply-to email address after sending a problem report. This information might be accidentally recorded by a screenshot or shown in the recent apps view. This PR makes sure to configure the `FLAG_SECURE` window flag when such information might be shown, to prevent this information to be recorded somewhere.

The Problem Report screen will only set the flag if the user's reply-to email is shown in the screen after the problem report has been sent successfully. The WireGuard Key screen and the Account screen are both marked with flag every time they are visible.

In order to make it simpler and less error-prone to mark screens as secure, the `MainActivity` holds a counter of currently visible secure screens. The counter is only incremented by an `enterSecureScreen()` method and is only decremented by a `leaveSecureScreen()` method. The `ServiceAwareFragment` (which is a super-class to most of the screens) contains a helper `isSecureScreen` read-only property that sub-classes can override and set to `true`. When it is set to `true`, the fragment will call `enterSecureScreen()` as soon as it is added to the activity (and before it is made visible) and will call `leaveSecureScreen()` just before it is removed from the activity (and after its views have been destroyed).

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1822)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Line wrap the file at 100 chars.                                              Th
 - Make all screens scrollable to better handle small screens and split-screen mode.
 - Ignore touch events when another view is shown on top of the app in order to prevent tapjacking
   attacks.
+- Prevent screens showing potentially sensitive data from being recorded.
 
 #### Linux
 - Send an ICMP reject message or TCP reset packet when blocking outgoing packets to prevent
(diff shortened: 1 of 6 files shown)


faern:
This fixed audit finding [MUL-02-003 WP1: General hardening recommendations for Android app (Info)](https://github.com/mullvad/mullvadvpn-app/blob/master/audits/2020-06-12-cure53.md#miscellaneous-issues)

PR #1830, opened 2020-06-09
Limit security identifiers with access to the management interface pipe

dlon:
In some limited circumstances, daemon commands could be sent to a host in a local network, if a user was authenticated or if authentication had been disabled, and if "Allow LAN" was enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1830)


faern:
This fixed audit finding [MUL-02-007 WP2: Named Pipe exposed via SMB accessible to everyone (Medium)](https://github.com/mullvad/mullvadvpn-app/blob/master/audits/2020-06-12-cure53.md#identified-vulnerabilities)

PR #1869, opened 2020-06-24
Install zsh shell completions to /usr/share instead of /usr/local/share

### LINKED ISSUE #1861: Files in /usr/local cause issues in Fedora Silverblue
## Issue report

Operating system: Fedora Silverblue 32

App version: 2020.4 (during installation)

### Issue description

When attempting to install the latest version of the app, it appears there are zsh-autocompletion files being added to /usr/local/share instead of /usr/share. This causes an issue in Silverblue and other rpm-ostree based distros due to projectatomic/rpm-ostree#233.

The issue could potentially be fixed by putting the auto completion files in /usr/share instead of /usr/local/share. I'm pretty sure this would be the convention for pre-packaged apps anyways, since /usr/local/ is usually reserved for apps not managed by the distro's package manager.

From what I can tell this is a separate issue from #1570. That issue was with files being added to /opt, which from what I understand did not exist in older versions of silverblue. However, /opt is now a symlink to /var/opt, so I didn't need to do the workaround mentioned there.

faern:
Tries to fix #1861. Also it seems more correct to not use `/usr/local` for this.

I tried it only on Fedora 31 so far but the completions work after this move.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1869)

### DIFF
--- a/gui/tasks/distribution.js
+++ b/gui/tasks/distribution.js
@@ -154,7 +154,7 @@ const config = {
       distAssets('linux/problem-report-link') + '=/usr/bin/mullvad-problem-report',
       distAssets('shell-completions/mullvad.bash') +
         '=/usr/share/bash-completion/completions/mullvad',
-      distAssets('shell-completions/_mullvad') + '=/usr/local/share/zsh/site-functions/_mullvad',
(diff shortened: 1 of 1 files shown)


faern:
Reverted the change in the deb file since it did not work on Ubuntu when the completions were placed in `/usr/share`.

PR #1907, opened 2020-07-08
Uncombine tunnel management

pronebird:

Git checklist:

That's a big one and it was almost impossible to decouple the changes as everything is intertwined.

1. Remove "Combine" from TunnelManager and PacketTunnelProvider
1. Refresh relays when GUI app starts. All other updates are covered by PacketTunnelProvider, which runs the periodic updates via `RelayCache`
1. Rename `TunnelConfiguration` to `TunnelSettings` - shorter and nicer. Same goes for `TunnelConfigurationManager` -> `TunnelSettingsManager`
1. Update account expiry each time visiting Settings navigation controller.
1. Add `AlertPresenter` which enables us to queue up alerts since only one can be displayed at a time. This is especially relevant when errors pile in and we need to display them to users.
1. Bring back Curve25519 implementation in C from Jason since. It's later needed for iOS 12 support since CryptoKit is iOS 13+.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1907)


ios/MullvadVPN/Account.swift
```diff
@@ -99,34 +21,64 @@ private enum UserDefaultsKeys: String {
...
+            return UserDefaults.standard.object(forKey: UserDefaultsKeys.accountExpiry.rawValue) as? Date
+        }
     }
 
+    private enum ExclusivityCategory {
+        case exclusive
+    }
+
+    private let rpc = MullvadRpc.withEphemeralURLSession()
+    private var updateExpiryTask: URLSessionTask?
```
pronebird, line 78: TODO: remove unused variable.

ios/MullvadVPN/AlertPresenter.swift
```diff
@@ -0,0 +1,97 @@
...
+
+            self.presentingController.present(self.alertController, animated: true, completion: self.presentCompletion)
+        }
+    }
+}
+
+/// A helper struct that swizzles `viewDidDisappear` on `UIAlertController` in order to be able to
+/// detect when the controller disappears.
+/// The event is broadcasted via `kUIAlertControllerDidDissmissNotification` notification.
+private struct AlertPresenterUIKitHooks {
```
pronebird, line 70: This is a patch on UIKit to be able to intercept the `viewDidDisapper` on `UIAlertController` and submit `Notification` to tell `AlertPresenter` when `UIAlertController` disappears. This is needed because only one `UIAlertController` can be presented at a time and there is no easy way to know when it's being dismissed. This is nasty but hopefully shouldn't cause much trouble.

ios/MullvadVPN/AppDelegate.swift
```diff
@@ -21,21 +20,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
         #endif
 
         let accountToken = Account.shared.token
 
-        loadTunnelSubscriber = TunnelManager.shared.loadTunnel(accountToken: accountToken)
-            .receive(on: DispatchQueue.main)
-            .sink(receiveCompletion: { (completion) in
-                if case .failure(let error) = completion {
-                    fatalError("Failed to restore the account: \(error.localizedDescription)")
+        RelayCache.shared.updateRelays()
```
pronebird, line 30: Update relays each time the GUI app starts in case if the tunnel doesn't run and can't provide updates on time. Otherwise the tunnel runs periodic updates.

ios/MullvadVPN/PacketTunnelIpc.swift
```diff
@@ -6,43 +6,38 @@
...
 
-import Combine
 import Foundation
 import NetworkExtension
 
 /// A enum describing the kinds of requests that `PacketTunnelProvider` handles
-enum PacketTunnelRequest: Int, Codable {
-    /// Request the tunnel to reload the configuration
-    case reloadConfiguration
+enum PacketTunnelRequest: Int, Codable, RawRepresentable {
```
pronebird, line 13: On iOS 13, JSONEncoder is able to serialise a single primitive into JSON, however on iOS 12 it throws an error saying that it needs a top level object to encode it as JSON. So basically I changed this `enum` to encode itself into the dictionary of following shape:

```
{ type: Int }
```

ios/MullvadVPN/WireguardPrivateKey.swift
```diff
@@ -6,7 +6,6 @@
 //  Copyright © 2019 Mullvad VPN AB. All rights reserved.
 //
 
-import CryptoKit
```
pronebird, line 9: CryptoKit is iOS 13 only and since we're rolling out iOS 12 support, I had to switch back to Jason's implementation in C.

ios/PacketTunnel/PacketTunnelSettingsGenerator.swift
```diff
@@ -13,7 +13,7 @@ import os
 
 struct PacketTunnelSettingsGenerator {
     let mullvadEndpoint: MullvadEndpoint
-    let tunnelConfiguration: TunnelConfiguration
+    let tunnelConfiguration: TunnelSettings
```
pronebird, line 16: TODO: rename to `tunnelSettings`

ios/PacketTunnel/WireguardDevice.swift
```diff
@@ -318,11 +295,10 @@ class WireguardDevice {
 
             // Re-resolve endpoints on network changes
             if let currentConfiguration = self.configuration {
-                let newResolvedConfiguration = Self.resolveConfiguration(currentConfiguration)
-                let commands = newResolvedConfiguration.endpointUapiConfiguration()
+                let resolvedConfiguration = Self.resolveConfiguration(currentConfiguration)
+                let commands = resolvedConfiguration.endpointUapiConfiguration()
 
                 Self.setWireguardConfig(handle: handle, commands: commands)
-                self.resolvedConfiguration = newResolvedConfiguration
```
pronebird, line 325: We never really used `resolvedConfiguration` for anything so I decided to drop it.

PR #1918, opened 2020-07-13
Implement UI for split tunnelling on Android

jvff:
This PR is the first step towards split tunneling on Android. It just creates the initial UI for the feature, as a new screen inside the "Advanced settings" menu. It contains a toggle for the feature itself as well as list of apps that can be toggled to be inside or outside of the tunnel. The UI itself can be tweaked later and should be considered somewhat experimental.

The UI doesn't actually change anything regarding the tunnel. Actually getting the feature working will be done in a separate PR.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Changelog entry will be added when the feature is functional.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1918)


jvff:
___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/applist/AppListHolder.kt, line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1918#-MCDu4E5-zLBvLSZ4EM5:-MCE0Rbk1-ciw0Pz8Kao:bj8hao4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3aa4fb38caefb8cad00329ca3b550edb6be4141d/android/src/main/kotlin/net/mullvad/mullvadvpn/applist/AppListHolder.kt#L14)):*

Is this an app _list_ or just an app element that is to be used in a list?
</blockquote></details>

It is an app element that is to be used in a list. I guess the name `AppInfoHolder` would be a better fit? Or maybe `AppListItemHolder`? :thinking:

PR #1944, opened 2020-07-21
Remove unused dependencies found with cargo-udeps

faern:
I used the third party `cargo-udeps` tool on our repository. Turns out we had quite a few unused dependencies. Most common `mullvad-paths` and `talpid-ipc` that were left on crates that no longer used them.

The real world difference here is very small since the total dependency tree only got rid of `tokio-service`. But crates should not depend on crates they don't need anyway so...

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/1944)


faern:
Currently on hold since it breaks the lockfile in this sad and magical way again. This work can be done when migration away from json-rpc is done.

PR #2004, opened 2020-08-20
Don't show out-of-time notification for new accounts

jvff:
Previously, the app would always show a notification when the account time ran out or was close to running out. The previous version (2020.6-beta1) changed that to not show the notification for newly created accounts in order to improve the user experience. However, that change had a bug in it. The account would be marked as not new as soon as the service could read the account expiry time. Since the "Welcome" screen checks for a new expiry every 15 seconds, the notification would reappear after 15 seconds or so.

This PR fixes the issue by only marking the account as not new after the expiry changes from the initial value. In order to do this, the first valid expiry value for the newly created account is stored and used to compare future account expiry values. This "cached" value is reset whenever the new account flag is changed.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2004)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ Line wrap the file at 100 chars.                                              Th
 - Show "Exclude applications" header if needed when entering the "Split tunneling" screen.
 - Fix check for update versions and check for support for current version.
 - Fix crash that could happen when leaving the Select Location screen.
+- Don't show out-of-time notification for newly created accounts.
 
 
 ## [2020.6-beta1] - 2020-08-20
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt
@@ -31,6 +31,7 @@ class AccountCache(val daemon: MullvadDaemon, val settingsListener: SettingsList
     private var accountNumber by onAccountNumberChange.notifiable()
     private var accountExpiry by onAccountExpiryChange.notifiable()
 
+    private var createdAccountExpiry: DateTime? = null
     private var oldAccountExpiry: DateTime? = null
 
     init {
@@ -41,13 +42,14 @@ class AccountCache(val daemon: MullvadDaemon, val settingsListener: SettingsList
 
     fun createNewAccount(): String? {
         newlyCreatedAccount = true
+        createdAccountExpiry = null
 
         return daemon.createNewAccount()
     }
 
     fun login(account: String) {
         if (account != accountNumber) {
-            newlyCreatedAccount = false
+            markAccountAsNotNew()
             daemon.setAccount(account)
         }
     }
@@ -91,6 +93,11 @@ class AccountCache(val daemon: MullvadDaemon, val settingsListener: SettingsList
         jobTracker.cancelAllJobs()
     }
 
+    private fun markAccountAsNotNew() {
+        newlyCreatedAccount = false
+        createdAccountExpiry = null
+    }
+
     private fun handleNewAccountNumber(newAccountNumber: String?) {
         synchronized(this) {
             accountExpiry = null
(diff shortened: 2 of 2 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2004)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt, line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2004#-MFFjh1x1BjSptGhmWRA:-MFUq1hb83VdVqlbmRkZ:bfbkr0z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7e16a8c92ffc80a5c31fc0646a16ae9793cfc0bf/android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt#L28)):*

Yep, precisely :D
</blockquote></details>

So really, it's value only ever changes once (even though it's applied multiple times). I'd leave it as a boolean rather than an observer of a related value, but that's just my opinion.

jvff:
___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt, line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2004#-MFFjh1x1BjSptGhmWRA:-MFWVVck4BzavLXQur56:bw8zbbn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7e16a8c92ffc80a5c31fc0646a16ae9793cfc0bf/android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt#L28)):*

So really, it's value only ever changes once (even though it's applied multiple times). I'd leave it as a boolean rather than an observer of a related value, but that's just my opinion.
</blockquote></details>

I'm not sure I understand. This field is a `boolean`, but setting it clears the `createdAccountExpiry` value. :/

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2004)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt, line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2004#-MFFjh1x1BjSptGhmWRA:-MF_FDMD6mOxpIkM0NRF:b6youpw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7e16a8c92ffc80a5c31fc0646a16ae9793cfc0bf/android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt#L28)):*

I'm not sure I understand. This field is a `boolean`, but setting it clears the `createdAccountExpiry` value. :/
</blockquote></details>

So regardless of the value being set, the `createdAccountExpiry` is always cleared. Why not do that explicitly?

jvff:
___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt, line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2004#-MFFjh1x1BjSptGhmWRA:-MF_FT0s4SfSrJXtNhSf:br770bo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7e16a8c92ffc80a5c31fc0646a16ae9793cfc0bf/android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt#L28)):*

So regardless of the value being set, the `createdAccountExpiry` is always cleared. Why not do that explicitly?
</blockquote></details>

We could, but I felt that it would be safer to do it here in case we forget to reset it when changing this value in the future. Currently we would have to clear it in two places, when a new expiration date is received and when logging in. If we miss any of these places (or any new places) in future refactors it might not show the notification for a legitimate case :/

jvff:
___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt, line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2004#-MFFjh1x1BjSptGhmWRA:-MF_GOWV44M34rnw39AT:bco62or) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7e16a8c92ffc80a5c31fc0646a16ae9793cfc0bf/android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt#L28)):*

We could, but I felt that it would be safer to do it here in case we forget to reset it when changing this value in the future. Currently we would have to clear it in two places, when a new expiration date is received and when logging in. If we miss any of these places (or any new places) in future refactors it might not show the notification for a legitimate case :/ 
</blockquote></details>

Do you think the issue is that setting the property hides the fact that another property is affected and that this isn't immediately obvious?

If that's the case, maybe we can refactor the code to use a new helper method, something like `markAccountAsNotNew`? What do you think?

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2004)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2033, opened 2020-08-26
Build all Rust code on nighly, since the blocker (jsonrpc) is gone

faern:
The jsonrpc stuff is finally gone. According to this fixme comment that was the only thing stopping us from building on nightly. And I tested now, it works to build on nightly. So let's run the full CI there as before, fewer special cases.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2033)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2033)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2036, opened 2020-08-26
Run mullvad-setup prepare-restart in preinstall script instead of postinstall

faern:
Fixes a leak during upgrade due to mullvad-setup no longer being compatible with old daemon

Not mentioned in changelog as a fix, since upgrading between any existing versions does not have this leak. It's only from any existing release to anything after the gRPC migration that causes the leak.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2036)


pinkisemils (approved):
:lgtm:

> Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2036#-:-MFfKO4zBiUEyAQHbIKG:bg7igk9)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2043, opened 2020-08-28
Change account input border rendering

jvff:
The specific border behavior that the desktop app has for the account input component isn't simple to replicate on Android. The first attempt used a foreground drawable, but that was later replaced with the approach that used an overlay and a clip outline. The problem with this approach is that it requires the views beneath the overlay to be aware of the clip outline and to not draw close to it to avoid rendering issues. Still, those issues would still appear depending on how the animation of the account history is performed.

In order to solve these issues, this PR implements the border behavior in a third implementation that is more of simpler things. In summary, it removes the use of an overlay drawable and a clip outline with eight image views to draw the border. Four of them draw the corners, and four of them draw the straight lines. It's probably slightly less efficient, since it increases GPU overdraw and initial layout calculations, but it shouldn't be noticeable at all.

The advantage of this approach is that it is more portable since it uses simple base features (image views and vector drawables) and it should always work without the child views having to know about any borders and clip limitations.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Internal refactor, not user visible.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2043)


pinkisemils (approved):
:lgtm:

> Reviewed 14 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2043#-:-MG5AEKB22F_3AegqMiz:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2046, opened 2020-08-28
Remove useless unwrap in relays.rs

NilsIrl:
Removes an unused unwrap.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2046)

### DIFF
--- a/mullvad-daemon/src/relays.rs
+++ b/mullvad-daemon/src/relays.rs
@@ -597,15 +597,13 @@ impl RelaySelector {
         } else {
             // Pick a random number in the range 0 - total_weight. This choses the relay.
             let mut i: u64 = self.rng.gen_range(0, total_weight + 1);
-            Some(
-                relays
-                    .iter()
-                    .find(|relay| {
-                        i = i.saturating_sub(relay.weight);
-                        i == 0
-                    })
-                    .unwrap(),
-            )
+            relays
+                .iter()
+                .find(|relay| {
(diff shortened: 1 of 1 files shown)


pronebird:
I think the `unwrap()` is there on purpose to crash the process if the weight based algorithm has a mistake. It acts as assertion.

NilsIrl:
That's what I thought, however, what kind of mistake can occur? It seems to me that nothing can occur that would cause find to return `None`. It's up to you.

pronebird:
@NilsIrl I think that `find()` can technically return `None` if it doesn't match any element, which should never happen. So again I believe it's an assertion from the developer to make sure that anyone who attempts to modify that function will hit the panic, although that doesn't stop anyone from removing that `unwrap()` either. So I somewhat agree that this `unwrap()` is superficial and subjective.

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2046)* status: :shipit: complete! all files reviewed, all discussions resolved

faern:
This branch does not compile. The expected return type is `Option<&Relay>` but this PR makes it into `&Relay`.

Yes I think we want this code as is. So the panic is triggered if a bug is introduced in the logic. It's quite easy to accidentally introduce an off by one error or something like that. We don't want that to silently cause `None` to be returned, we want that to shout loud and clear that there is a bug.

PR #2058, opened 2020-09-02
Start tunnel state machine in appropriate state.

pinkisemils:
Previously, after upgrading, there always existed a small leak window where the firewall was reset to a non-blocking state for a very brief period of time. There were two causes for this:
- the firewall was initialized to a non-blocking state unless the daemon was configured to block when disconnected
- the tunnel state machine would always start from a disconnected state which would always reset the firewall

The first problem is resolved by having the daemon specify the firewall arguments, using the cached target state and the _block when disconnected_ setting to deduce whether the firewall should be initialized to a blocked state.
The second is resolved by allowing to specify whether the tunnel state machine should start in the disconnected or connecting state.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Line wrap the file at 100 chars.                                              Th
 - Add support for WireGuard's kernel module if it's loaded.
 
 ### Fixed
+- Stop resetting the firewall after an upgrade to not leak after an upgrade.
 
 #### Windows
 - Remove firewall filters (unblock internet access) when "Always require VPN" is enabled and the app
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -558,26 +558,7 @@ where
         let tunnel_parameters_generator = MullvadTunnelParametersGenerator {
             tx: internal_event_tx.clone(),
         };
-        let tunnel_command_tx = tunnel_state_machine::spawn(
-            settings.allow_lan,
-            settings.block_when_disconnected,
-            tunnel_parameters_generator,
-            log_dir,
-            resource_dir,
-            cache_dir.clone(),
-            internal_event_tx.to_specialized_sender(),
-            tunnel_state_machine_shutdown_tx,
-            #[cfg(target_os = "android")]
-            android_context,
-        )
-        .await
-        .map_err(Error::TunnelError)?;
 
-        let wireguard_key_manager =
-            wireguard::KeyManager::new(internal_event_tx.clone(), rpc_handle.clone());
-
-        // Attempt to download a fresh relay list
-        relay_selector.update().await;
 
         let initial_target_state = if settings.get_account_token().is_some() {
             if settings.auto_connect {
@@ -590,6 +571,29 @@ where
         } else {
             TargetState::Unsecured
         };
+
+
+        let tunnel_command_tx = tunnel_state_machine::spawn(
+            settings.allow_lan,
+            settings.block_when_disconnected,
+            tunnel_parameters_generator,
+            log_dir,
+            resource_dir,
+            cache_dir.clone(),
+            internal_event_tx.to_specialized_sender(),
+            tunnel_state_machine_shutdown_tx,
+            initial_target_state != TargetState::Secured,
+            #[cfg(target_os = "android")]
+            android_context,
+        )
+        .await
+        .map_err(Error::TunnelError)?;
+
+        let wireguard_key_manager =
+            wireguard::KeyManager::new(internal_event_tx.clone(), rpc_handle.clone());
+
+        // Attempt to download a fresh relay list
+        relay_selector.update().await;
 
         let mut daemon = Daemon {
             tunnel_command_tx,
--- a/mullvad-setup/src/main.rs
+++ b/mullvad-setup/src/main.rs
@@ -72,7 +72,7 @@ async fn reset_firewall() -> Result<(), Error> {
 
     let mut firewall = Firewall::new(FirewallArguments {
         initialize_blocked: false,
-        allow_lan: None,
+        allow_lan: true,
     })
     .map_err(Error::FirewallError)?;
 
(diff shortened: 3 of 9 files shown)


dlon (approved):
> Reviewed 4 of 4 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058)* status: all files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[CHANGELOG.md, line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058#-MGDbg6n78iVGLKMQUfC:-MGDbg6n78iVGLKMQUfD:bpnkosq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4297fb82a31e4580e7a9e4fcf6d19fde3c6fd5f1/CHANGELOG.md#L28)):*
> ```Markdown
> 
> ### Fixed
> - Stop resetting the firewall after an upgrade to not leak after an upgrade.
> ```

I think this could be more leniently worded, given that the prior changes *nearly* eliminated leaks. I wasn't able to produce any, at least. E.g., "[...] completely eliminate the possibility of leaks".
___
*[mullvad-daemon/src/lib.rs, line 568 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058#-MGDnbFIET6pD05PFgSD:-MGDnbFIET6pD05PFgSE:b-y98x9o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/91f62acd73b12c162b908186087d93d4ac6dd719/mullvad-daemon/src/lib.rs#L568)):*
> ```Rust
>             .unwrap_or(settings.block_when_disconnected);
> 
>         let firewall_args = FirewallArguments {
> ```

You can infer whether to block initially and what firewall arguments to use from `block_when_disconnected` and `connect_immediately` in `tunnel_state_machine::spawn` instead of here. This is assuming that it is fine to initialize blocked when using `auto_connect`, and I don't see why it would hurt, given that it will block as soon as it enters the connecting state anyhow.

Currently, you can pass `firewall_args.initialize_blocked=false` and `settings.block_when_disconnected=true` at the same time to `spawn`, which is a bit weird.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058)* status: all files reviewed, 3 unresolved discussions (waiting on @pinkisemils)

___
*[mullvad-types/src/states.rs, line 23 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058#-MGE2RpBDFHWcQ_bfClI:-MGE2RpCBUCpTZEInjmN:b99of6t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/91f62acd73b12c162b908186087d93d4ac6dd719/mullvad-types/src/states.rs#L23)):*
> ```Rust
> impl TargetState {
>     /// Returns true if the target state is `Secured`.
>     pub fn is_secured(&self) -> bool {
> ```

Is this helper really needed? would not just `state == TargetState::Secured` work equally well? Not much longer and fewer layers in between.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058)* status: 1 of 9 files reviewed, 2 unresolved discussions (waiting on @dlon, @faern, and @jvff)

___
*[CHANGELOG.md, line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058#-MGDbg6n78iVGLKMQUfC:-MGEe4bw1h0iUd4-j050:bjqqyom) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4297fb82a31e4580e7a9e4fcf6d19fde3c6fd5f1/CHANGELOG.md#L28)):*

I think this could be more leniently worded, given that the prior changes *nearly* eliminated leaks. I wasn't able to produce any, at least. E.g., "[...] completely eliminate the possibility of leaks".
</blockquote></details>

Given that leaks definitely occurred previously, I think it makes sense to leave this as is.
___
*[mullvad-types/src/states.rs, line 23 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058#-MGE2RpBDFHWcQ_bfClI:-MGEnwll5cimltKgEi_O:b-yu7q59) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/91f62acd73b12c162b908186087d93d4ac6dd719/mullvad-types/src/states.rs#L23)):*

Is this helper really needed? would not just `state == TargetState::Secured` work equally well? Not much longer and fewer layers in between.
</blockquote></details>

When accessing values of options, `map(TargetState::is_secured)` feels a lot less cumbersome than `map(|state| state == TargetState::Secured)`. But I agree that it doesn't help with clarity. And in the end, the target state ends up not being optional anyway, so these changes are superfluous.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058)* status: all files reviewed, 2 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[talpid-core/src/tunnel_state_machine/mod.rs, line 87 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058#-MGI3YA1DLYe4FL3CPmq:-MGI3YA25tMbkPXziumX:bw0xb57) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42c7e0c8b1129ba67c17fdb1a556d5b950441869/talpid-core/src/tunnel_state_machine/mod.rs#L87)):*
> ```Rust
>     state_change_listener: impl Sender<TunnelStateTransition> + Send + 'static,
>     shutdown_tx: oneshot::Sender<()>,
>     connect_immediately: bool,
> ```

This parameter should probably be renamed now.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

jvff (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

faern (approved):
> Reviewed 8 of 9 files at r3, 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2058)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2061, opened 2020-09-02
Fix NM for IPv6 with OpenVPN

pinkisemils:
The changes made to NM to get it to work with the kernel WireGuard implementation broke it for OpenVPN. If the address that's already set for an interface is local to the link, then it shouldn't be reset by NM - it will complain and return an error. So I changed the code to only append the IPv6 addresses to the config if the link does not have link local addresses. I did not bump the changelog because this bug hasn't been released yet.

Makes me think whether it'd be better to set the IP address as link local for WireGuard too.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2061)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2061)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2062, opened 2020-09-03
Create `SwitchLocationButton` in order to fix a rare crash

jvff:
For a long time, there has been a crash caused by an `ArrayIndexOutOfBoundsException`. However, since this occurred inside a `TextView` while laying out the UI, I was unable to find on which View this happened. I never managed to reproduce the error until today, so it was just left in the queue.

I stumbled upon a [StackOverflow question](https://stackoverflow.com/questions/17047172/android-java-lang-arrayindexoutofboundsexception-length-10-index-10) that seemed to have a similar issue. The cause of that issue was that `maxLines` was set to `-1`, an invalid value. When I grepped the code for `maxLines`, there was an instance in a controller class called `SelectLocationButton`. Changing that to the correct value (`Int.MAX_VALUE`) fixed the crash, but the button would sometimes have the wrong height. This was probably due to the hacky height calculation code in the controller.

To properly fix both issues, a new `SwitchLocationButton` widget was created in this PR. The custom widget is an attempt to simplify the original button. It removes all of the custom height changing and margin setting code. Instead, two buttons are placed in a `FrameLayout`, so that they are laid out once, and the larger button can reserve the needed space. The button now can have as many lines as needed, which better matches the desktop app.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2062)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2062)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2062)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2063, opened 2020-09-03
Update UI even when window doesn't have focus

jvff:
I've been overly conservative regarding the fragment lifecycles. They always wait for the window to gain focus before they start updating. This works for most cases, but it fails to work on split-screen mode when the user is using another app (i.e., our app's window loses focus). This becomes more apparent with configuration changes (like changing the text size in the accessibility options), where the app would return to the launch screen and wait until it receives focus again.

This PR changes all fragments to use `onStart`/`onStop` instead of `onResume`/`onPause`. This improves the situation in split screen mode with little impact on normal usage.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2063)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Line wrap the file at 100 chars.                                              Th
 #### Android
 - Fix rare crash that could happen when starting the background service.
 - Fix rare crash that happened with large text sizes and long location names on the main screen.
+- Fix UI not updating in split screen mode when the window is unfocused.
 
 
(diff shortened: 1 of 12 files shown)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2063#-:-MGIRhA83wFnDaW9F0h8:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2065, opened 2020-09-03
Allow more systemd-resovled config paths

pinkisemils:
In #1968, I removed the non-stub systemd-resolved config paths, as there can be issues if there are more than 3 DNS server entries set. This isn't actually ideal, as then the daemon will fall back to using resolvconf, and our DNS servers will not be used at all. Since it takes active effort on the users part to add extra DNS entries to their config to break the behavior I'm introducing in this PR, but a default install will fail on master, I'm adding back the stub config paths back. I have confirmed that this fix works on Debian testing.


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2065)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2078, opened 2020-09-04
Change animation to show and hide account history

jvff:
This PR tweaks the account history widget to become more similar to the desktop. The border is now drawn around the whole widget, including the history list, and the history now expands and collapses using an animation to change the container height. The bottom margin is also changed with the height so that the widget remains in the same position.

It is now also possible to touch outside the account history widget in order to collapse it. This will also hide the on-screen keyboard.

There is also a small workaround to make sure the widget is positioned correctly as soon as possible. Otherwise, it will wait until the animation starts to relayout and ends up jumping vertically a bit when the user touches the account input area.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **UI tweak, no changelog entry necessary**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2078)


pinkisemils (approved):
> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2078)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2080, opened 2020-09-07
Remove dependency on validated

raksooo:
This PR removes the dependency on `validated`. Most of the usages were removed when switching to gRPC but it was still used for validating the `gui-settings.json` file. Luckily the validation of this file is simple to implement ourselves. The reason why we want to get rid of validated is that it lacks typescript types and is poorly maintained.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080)


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[gui/src/main/gui-settings.ts, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080#-MGghdDY6fULV6t8LzWW:-MGghdDY6fULV6t8LzWX:blc606s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3bbb58e35b2aed143376174cebb602c37257d79/gui/src/main/gui-settings.ts#L80)):*
> ```XML
>       this.stateValue = {
>         ...defaultSettings,
>         ...this.validateSettings(rawJson),
> ```

`validateSettings` produces the exact shape object and not `Partial`, so adding new fields to schema will result in validation error when loading the settings file from the previous version of the app.

pronebird (commented):
> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[gui/src/main/gui-settings.ts, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080#-MGghdDY6fULV6t8LzWW:-MGgpmqG41JZiOR_inRP:bpnjyd9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3bbb58e35b2aed143376174cebb602c37257d79/gui/src/main/gui-settings.ts#L80)):*

`validateSettings` produces the exact shape object and not `Partial`, so adding new fields to schema will result in validation error when loading the settings file from the previous version of the app.
</blockquote></details>

I think this works the same as before but I agree that this is an issue. How about allowing fields to be `undefined`?

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080)* status: all files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[gui/src/main/gui-settings.ts, line 101 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080#-MGgqBPk7mCWMI9hfpMs:-MGgqBPl2RSoNL2ndPml:b89ez90) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3bbb58e35b2aed143376174cebb602c37257d79/gui/src/main/gui-settings.ts#L101)):*
> ```XML
>     Object.entries(settingsSchema).forEach(([key, expectedType]) => {
>       const actualType = typeof settings[key];
>       if (actualType !== expectedType) {
> ```

Allowing fields to be `undefined` would change this line to
```
if (actualType !== 'undefined' && actualType !== expectedType) {
```

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080)* status: all files reviewed, 2 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/gui-settings.ts, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080#-MGghdDY6fULV6t8LzWW:-MGlhLB_4NB4w2Y1ePAL:b-kis5p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3bbb58e35b2aed143376174cebb602c37257d79/gui/src/main/gui-settings.ts#L80)):*

I think this works the same as before but I agree that this is an issue. How about allowing fields to be `undefined`?
</blockquote></details>

Currently the validation function throws an error if any of the fields are missing in the JSON saved on disk as the `typeof settings[key]` for the non-existent field would be `undefined`. 

Therefore the `...defaultSettings` spread doesn't make any sense as this function would always exit on error.

`Partial<ObjectType>` means that we have a shape but all fields are optional, i.e:

```
interface Props {
  barks: boolean;
  sparks: boolean;
}
```

With `Partial<Props>` applied, this becomes equivalent to:

```
interface PartialProps {
  barks?: boolean;
  sparks?: boolean;
}
```

So to fix that I think `validateSettings` should allow the missing keys, however throw an error if any of the keys matching the schema have a type mismatch.
___
*[gui/src/main/gui-settings.ts, line 101 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080#-MGgqBPk7mCWMI9hfpMs:-MGljNXQ9p3wsU6O3-fx:b8qzl5z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3bbb58e35b2aed143376174cebb602c37257d79/gui/src/main/gui-settings.ts#L101)):*
Yeah, that should work but be careful with `undefined` fields, i.e:

```
> { ...{ a: 'test' }, ...{ a: undefined } }
{ a: undefined }
```

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080)* status: all files reviewed, 2 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/gui-settings.ts, line 101 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080#-MGgqBPk7mCWMI9hfpMs:-MGlpd2XCNLdg2OJinP7:b-ix3ju0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3bbb58e35b2aed143376174cebb602c37257d79/gui/src/main/gui-settings.ts#L101)):*

Yeah, that should work but be careful with `undefined` fields, i.e:

```
> { ...{ a: 'test' }, ...{ a: undefined } }
{ a: undefined }
```
</blockquote></details>

I think it may be prudent to `delete` the field if it's `undefined`, i.e:

```
if typeof settings[key] === undefined {
  delete settings[key]
}
```

and if you want to avoid mutating the input, then you could construct a new object instead and then return it.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[gui/src/main/gui-settings.ts, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080#-MGghdDY6fULV6t8LzWW:-MGmC___BFyctrX_F3AJ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3bbb58e35b2aed143376174cebb602c37257d79/gui/src/main/gui-settings.ts#L80)):*

Currently the validation function throws an error if any of the fields are missing in the JSON saved on disk as the `typeof settings[key]` for the non-existent field would be `undefined`. 

Therefore the `...defaultSettings` spread doesn't make any sense as this function would always exit on error.

`Partial<ObjectType>` means that we have a shape but all fields are optional, i.e:

```
interface Props {
  barks: boolean;
  sparks: boolean;
}
```

With `Partial<Props>` applied, this becomes equivalent to:

```
interface PartialProps {
  barks?: boolean;
  sparks?: boolean;
}
```

So to fix that I think `validateSettings` should allow the missing keys, however throw an error if any of the keys matching the schema have a type mismatch.
</blockquote></details>

Done.
___
*[gui/src/main/gui-settings.ts, line 101 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080#-MGgqBPk7mCWMI9hfpMs:-MGmC8jBE2KVz6ngnouB:b-ykprk7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3bbb58e35b2aed143376174cebb602c37257d79/gui/src/main/gui-settings.ts#L101)):*

I think it may be prudent to `delete` the field if it's `undefined`, i.e:

```
if typeof settings[key] === undefined {
  delete settings[key]
}
```

and if you want to avoid mutating the input, then you could construct a new object instead and then return it.
</blockquote></details>

There's no `undefined` type in JSON so that shouldn't be an issue. Either way the `in` operator should handle this as expected :) Done.

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2080#-:-MGnE6GE-4SXbf5p1lz1:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

PR #2081, opened 2020-09-07
Fix deadlock in the problem report tool, and futures updates

dlon:
The runtime is prevented from doing any work due to a `future.wait()` in the same thread, causing a deadlock. This was fixed by returning a new-style future from `ProblemReportProxy` and letting the runtime poll it instead of blocking the thread.

The remaining uses of `futures 0.1` were also removed from `mullvad-daemon` and `mullvad-rpc`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2081)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1241,7 +1241,6 @@ dependencies = [
  "chrono 0.4.15 (registry+https://github.com/rust-lang/crates.io-index)",
  "err-derive 0.2.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "filetime 0.2.12 (registry+https://github.com/rust-lang/crates.io-index)",
- "futures 0.1.29 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "http 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "hyper 0.13.7 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/mullvad-rpc/Cargo.toml
+++ b/mullvad-rpc/Cargo.toml
@@ -10,8 +10,7 @@ publish = false
 [dependencies]
 chrono = { version = "0.4", features = ["serde"] }
 err-derive = "0.2.1"
-futures = { version = "0.3", features = [ "default", "compat" ] }
-futures01 = { package = "futures", version = "0.1" }
+futures = "0.3"
 http = "0.2"
 hyper = "0.13"
 ipnetwork = "0.16"
--- a/mullvad-rpc/src/rest.rs
+++ b/mullvad-rpc/src/rest.rs
@@ -5,7 +5,6 @@ use futures::{
     stream::StreamExt,
     TryFutureExt,
 };
-use futures01::Future as OldFuture;
 use hyper::{
     client::{connect::Connect, Client},
     header::{self, HeaderValue},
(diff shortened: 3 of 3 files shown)


faern (changes requested):
> Reviewed 8 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2081)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[mullvad-daemon/src/lib.rs, line 1237 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2081#-MGdNTGFEyu8XY_HyZ-P:-MGdNTGFEyu8XY_HyZ-Q:bxhe1wu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f05c347a9d07911ffa4643f2f50cace0c331500/mullvad-daemon/src/lib.rs#L1237)):*
> ```Rust
>     }
> 
>     async fn on_create_new_account(
> ```

It does not look like this method does anything async (There is not a single `.await` inside it). It didn't before this rewrite either, so nothing really changed. Question is if it would make sense to not use `tokio::spawn`, but rather just run the async operation directly in this task? Or make it a non-async fn?

I see this is true for many `on_*` methods in this file.

faern (approved):
> Reviewed 1 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2081)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[mullvad-daemon/src/lib.rs, line 1237 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2081#-MGdNTGFEyu8XY_HyZ-P:-MGdOA-A37kH0rqGpR1G:b-vrxg6h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f05c347a9d07911ffa4643f2f50cace0c331500/mullvad-daemon/src/lib.rs#L1237)):*

It does not look like this method does anything async (There is not a single `.await` inside it). It didn't before this rewrite either, so nothing really changed. Question is if it would make sense to not use `tokio::spawn`, but rather just run the async operation directly in this task? Or make it a non-async fn?

I see this is true for many `on_*` methods in this file.
</blockquote></details>

Not marking this as a blocker. Since this PR does not introduce this oddity. But worth discussing.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2081)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[mullvad-daemon/src/lib.rs, line 1237 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2081#-MGdNTGFEyu8XY_HyZ-P:-MGdlDZHEOqkV8lrE0qq:b5jjl8f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f05c347a9d07911ffa4643f2f50cace0c331500/mullvad-daemon/src/lib.rs#L1237)):*

Not marking this as a blocker. Since this PR does not introduce this oddity. But worth discussing.
</blockquote></details>

The rationale for its being `async` is that it uses `tokio::spawn`. Spawning outside of a tokio runtime causes a panic, but since it is async it is harder to do so by accident. I agree that this is somewhat strange, and you might have better suggestions for how to deal with this.

As for whether to use `spawn` at all, the reason for doing so is that it allows more events to be processed while this has not yet finished. Otherwise `handle_event` will wait here for the future to resolve before moving on to the next event (line 652 in `lib.rs`), and this might take seconds:

```rust
while let Some(event) = self.rx.next().await {
    self.handle_event(event).await;
}
```
Though it might make more sense to spawn inside this loop instead of in the `on_*` methods.

PR #2084, opened 2020-09-08
Log better wg impl selection

faern:
We want to make it simple to see from the logs which WireGuard implementation is actually used. So I add logging even in the case where the kernel one is used instead of staying silent. This makes it more explicit.

Also replaced some ` - ` delimiters and made them `: ` to be more consistent as well as used our `display_chain_with_msg` to print the entire chain and not just the top error.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2084)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2084)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2085, opened 2020-09-08
Fix build.sh for android tags

faern:
`git describe` did not work very well any more when we might have two tags on the same commit. It could return the wrong one and the build scripts failed to realize it was a release build.

My change involves using `$PRODUCT_VERSION` (the version we think we are building) and checking what commit hash the corresponding release tag is on. We then get the commit hash for `HEAD`. If they are equal, it means this should be a release build for that `$PRODUCT_VERSION`.

Git checklist:

* [ ] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header.
* [ ] Check that commits follow the [Mullvad coding guidelines](https://github.com/mullvad/coding-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2085)

PR #2088, opened 2020-09-08
Allow build[-apk].sh to work even when release tag does not exist

faern:
Follow up on #2085. Currently it crashes when trying to build and the version metadata files has a version that there is no tag for. this should not really ever happen in the real world. But during this transition period that will be true. And also, better to cover an edge case than crash.

`git rev-parse HEAD` should never fail. `HEAD` does always exist. But some arbitrary tag might not exist.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2088)

### DIFF
--- a/build-apk.sh
+++ b/build-apk.sh
@@ -47,7 +47,7 @@ if [[ "$GRADLE_BUILD_TYPE" == "release" ]]; then
     fi
 fi
 
-product_version_commit_hash=$(git rev-parse android/$PRODUCT_VERSION^{commit})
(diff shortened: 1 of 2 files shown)


jvff (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2088#-:-MGhphV47gUEKk87SlTx:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2092, opened 2020-09-08
Fix inconsistent quick settings tile behavior

jvff:
Previously, the quick settings tile would have an inconsistent behavior when there was no account set. If the service was already running, pressing the tile would make the daemon enter the blocked state. If it wasn't running, it would be started but the UI would be opened so that the user could log in first.

This PR makes the behavior consistent by always opening the UI if the user is logged out.

Unfortunately, on my test device there is still a small difference. For some unknown reason, Android will take a few seconds (3-5) to start the UI if the service is already running. I couldn't figure out what could cause this. The closest I came was with removing the `singleTask` configuration on the activity, which reduced the lag to less than one second. However, that would reintroduce the failure to quit bug, so I decided to leave it as is.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2092)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@ Line wrap the file at 100 chars.                                              Th
 - Fix UI not updating in split screen mode when the window is unfocused.
 - Fix split tunneling not being correctly configured after restarting the app.
 - Fix app reopening after pressing the Quit button because app was running multiple tasks.
+- Fix inconsistent behavior of the quick-settings tile when logged out. It would sometimes enter the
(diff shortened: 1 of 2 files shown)


faern (approved):
:lgtm: I can't evaluate the code on a very technical level. But the description seems sound and the code seem to match :)

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2092#-:-MGm2hYtE1i8tAUyHYSz:bintwqf)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2104, opened 2020-09-10
Mark VPN connection as not metered

jvff:
Previously, the app didn't mark if the VPN connection was metered or not (i.e., charged by the amount of data transmitted). On recent Android versions, the default value is now set to `true`, which means some apps will behave differently while connected to the VPN server, in an attempt to reduce data charges.

This PR changes that so that the connection is marked as not metered. Since Mullvad does not charge per amount of data, it makes sense to mark the VPN connection as not metered. This has the effect of letting the underlying network decide if the connection is metered. If any of the transport layers is marked as metered, then the whole connection is marked as metered, otherwise it remains as not metered, and the behavior is as expected for the user.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2104)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ Line wrap the file at 100 chars.                                              Th
 - Fix app reopening after pressing the Quit button because app was running multiple tasks.
 - Fix inconsistent behavior of the quick-settings tile when logged out. It would sometimes enter the
   blocking state and sometimes open the UI for the user to login. Now it always opens the UI.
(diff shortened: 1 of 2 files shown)


faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2104#-:-MGtqrJO6Ug7as0GP5G1:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2107, opened 2020-09-11
Import plural translations into Android

jvff:
Previously, the translation converter helper tool would only load Android plural resources and append them as missing translations to the template file (`gui/locales/messages.pot`). This PR updates the tool so that it can properly read the plural messages from each translation file and then convert them and write them out as Android plural resources XML files.

Since different locales have different pluralization rules, this is a bit tricky. On Android, the options available to translators is rather limited, since the messages must specify the plural quantity based on five options (`zero`, `one`, `few`, `many` and `other`) while gettext allows a general formula than can produce an unlimited number of strings to use. To make the conversion possible, only known formulas are parsed from the translation file header. They are then used to map to Android's plural scheme.

The manually implemented parser in the `Translation::from_file` method became more complex, since now it must load not only message entries, but also the plural form header and the plural entries. It might be interesting to rewrite the parser using some library in the future, to make it more maintainable. But that should probably be left for a separate PR if needed.

The normal messages have a simple pattern of a `msgid` line followed by `msgstr` line. However, the header has the same pattern, but with an empty ID and value, and then a list of "header" strings follow. The parser keeps track if it entered a header area to see if it can find the plural form header. The plural entries are similar to the normal message entries, but the ID is followed by a `msgid_plural` line, and then a series of `msgstr[X]` lines to create an array with the plural items. These array values are collected into a `BTreeMap`, so that they can be sorted and then are later easily zipped together with their respective Android plural quantity.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Internal tool update, no user visible changes.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2107)

### DIFF
--- a/android/translations-converter/src/android.rs
+++ b/android/translations-converter/src/android.rs
@@ -234,3 +234,17 @@ impl PluralResources {
         }
     }
 }
+
+impl PluralResource {
+    /// Create a plural resource representation.
+    ///
+    /// The resource has a name, used as the identifier, and a list of items. Each item contains
+    /// the message and the quantity it should be used for.
+    pub fn new(name: String, values: impl Iterator<Item = (PluralQuantity, String)>) -> Self {
+        let items = values
+            .map(|(quantity, string)| PluralVariant { quantity, string })
+            .collect();
+
+        PluralResource { name, items }
+    }
+}
--- a/android/translations-converter/src/android.rs
+++ b/android/translations-converter/src/android.rs
@@ -225,3 +225,12 @@ impl IntoIterator for PluralResources {
         self.entries.into_iter()
     }
 }
+
+impl PluralResources {
+    /// Create an empty list of plural resources.
+    pub fn new() -> Self {
+        PluralResources {
+            entries: Vec::new(),
+        }
+    }
+}
--- a/android/translations-converter/src/android.rs
+++ b/android/translations-converter/src/android.rs
@@ -193,7 +193,7 @@ pub struct PluralVariant {
 }
 
 /// A valid quantity for a plural variant.
-#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize)]
 #[serde(rename_all = "snake_case")]
 pub enum PluralQuantity {
     Zero,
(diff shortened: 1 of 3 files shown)


pinkisemils (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2107)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[android/translations-converter/src/gettext.rs, line 75 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2107#-MOCkL6O9G_jTHPcD3o_:-MOCkL6O9G_jTHPcD3oa:b-j4jst4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6947abd3c545285fd516514924c6abf636886594/android/translations-converter/src/gettext.rs#L75)):*
> ```Rust
>     /// that the header value is one of some hard-coded values, so if new languages that have new
>     /// plurals are added, the code will have to be updated.
>     pub fn from_file(file_path: impl AsRef<Path>) -> Self {
> ```

This would benefit massively from having a sample from a translation file so it'd be obvious what the expected input looks like.

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2107)* status: 2 of 3 files reviewed, all discussions resolved (waiting on @pinkisemils)

___
*[android/translations-converter/src/gettext.rs, line 75 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2107#-MOCkL6O9G_jTHPcD3o_:-MOCutOZE90tOPFJfmYp:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6947abd3c545285fd516514924c6abf636886594/android/translations-converter/src/gettext.rs#L75)):*

This would benefit massively from having a sample from a translation file so it'd be obvious what the expected input looks like.
</blockquote></details>

Done.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2107)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2121, opened 2020-09-16
Fix conditions to anticipate reconnecting state

jvff:
When a reconnect is requested in the Kotlin side, the `ConnectionProxy` will attempt to "anticipate" the reconnecting state. That means that it will notify any UI state listeners to update to the reconnecting state _before_ the actual reconnecting state is received from the Rust side. This is done to improve the responsiveness of the UI so that the user feels a reaction to the command as soon as possible.

Previously, the reconnecting state would be incorrectly anticipated in a few cases. Most notably, when the tunnel was disconnected. This became apparent with the introduction of the split tunneling feature, because every change to the split tunneling configuration requested a reconnect, and the Rust side would correctly ignore the request if the target state was unsecured. This meant that when the split tunneling configuration changed (when for example it was first loaded from persistent storage), it would reconnect the tunnel and the `ConnectionProxy` would incorrectly anticipate the reconnect if the tunnel was disconnected, so parts of the UI would update incorrectly, and then return to the correct state after a few seconds.

This PR updates the conditions used to determine when the reconnect state should be anticipated to prevent incorrect UI updates.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2121)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,8 @@ Line wrap the file at 100 chars.                                              Th
   blocking state and sometimes open the UI for the user to login. Now it always opens the UI.
 - Mark the VPN connection as not metered, so that Android properly reports if the connection is or
   isn't metered based solely on the underlying network, and not on the VPN connection.
+- Fix connect action button sometimes showing itself as "Cancel" instead of "Secure my connection"
+  for a few seconds.
 
 #### Linux
 - Fix split tunneling rules preventing `systemd-resolved` from performing DNS lookups for excluded
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/ConnectionProxy.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/ConnectionProxy.kt
@@ -109,15 +109,26 @@ class ConnectionProxy(val context: Context, val daemon: MullvadDaemon) {
         synchronized(this) {
             val currentState = uiState
 
-            if (currentState is TunnelState.Disconnecting &&
-                currentState.actionAfterDisconnect == ActionAfterDisconnect.Reconnect
-            ) {
-                return false
-            } else {
+            val willReconnect = when (currentState) {
+                is TunnelState.Disconnected -> false
+                is TunnelState.Disconnecting -> {
+                    when (currentState.actionAfterDisconnect) {
+                        ActionAfterDisconnect.Nothing -> false
+                        ActionAfterDisconnect.Reconnect -> false
+                        ActionAfterDisconnect.Block -> true
+                    }
+                }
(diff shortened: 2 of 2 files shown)


pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2121)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @jvff)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/ConnectionProxy.kt, line 117 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2121#-MHl4oTRBdUJDidRFOvl:-MHl4oTRBdUJDidRFOvm:b1ag3aj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ca21e771714e879fdc55d8b9a735f1165e8ce1c8/android/src/main/kotlin/net/mullvad/mullvadvpn/service/ConnectionProxy.kt#L117)):*
> ```Kotlin
>                     when (currentState.actionAfterDisconnect) {
>                         ActionAfterDisconnect.Nothing -> false
>                         ActionAfterDisconnect.Reconnect -> false
> ```

Shouldn't this be `true`?

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2121)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/ConnectionProxy.kt, line 117 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2121#-MHl4oTRBdUJDidRFOvl:-MHl65Ql25ip5CFx5S7p:b-drrwr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ca21e771714e879fdc55d8b9a735f1165e8ce1c8/android/src/main/kotlin/net/mullvad/mullvadvpn/service/ConnectionProxy.kt#L117)):*

Shouldn't this be `true`? 
</blockquote></details>

It shouldn't actually matter, since the UI state would change into the same value, so I just left it as to not change it.

pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2121#-:-MHl7Bxw8e6GoOtlQ4i1:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2121)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2128, opened 2020-09-18
Remove verdict from masquerade rule

dlon:
The `masquerade` rule is slightly off. It has an "accept" verdict, but this makes no sense. If you attempt to add the same rule using `nft`, you will see that it rejects this, because `masquerade` is supposed to be used in place of a verdict. It already functioned normally, so this fix is technically correct but it does not seem to matter in practice.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2128)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2128#-:-MHjv3Wf6p2T0ohJEs6V:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2136, opened 2020-09-21
Fix crash if service connection is lost when opening the Split Tunneling settings screen

jvff:
A crash report indicated that the `splitTunneling` property was used before it was initialized in the `SplitTunnelingFragment`. The stack trace showed that this happened inside the `onAttach` method. The property should be initialized in the call to `super.onAttach`, _as long as the service connection is already available_. If it's not available, it doesn't get initialized, and that shouldn't be a problem since the super `ServiceDependentFragment` class should handle it gracefully by showing a temporary screen and then returning to the launch screen. However, the `SplitTunnelingFragment` tries to use the property out of one of the "safe" methods, so the super-class handling of the situation doesn't get a chance to execute because the code crashes before that happens.

This PR fixes that by delaying the initialization of the adapter so that it is only initialized inside a "safe" method, ensuring the `splitTunneling` property is initialized by then.

The PR also adds a missing `private` modifier to the property's setter for consistency with the other properties.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2136)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ Line wrap the file at 100 chars.                                              Th
   for a few seconds.
 - Fix the notification sometimes leaving the foreground and becoming dismissable even if the UI was
   still visible.
+- Fix crash if connection to service is lost while opening the Split Tunneling settings screen.
 
 #### Linux
 - Fix split tunneling rules preventing `systemd-resolved` from performing DNS lookups for excluded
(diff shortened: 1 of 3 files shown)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2136#-:-MHpIBZP7bZvlHvQeQAV:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2140, opened 2020-09-23
Fix `NullPointerException` inside `MullvadTileService`

jvff:
There was a single crash report of a `NullPointerException` inside the `MullvadTileService`. This is pretty easy to avoid, and we simply add a `null` check in `qsTile?.apply`. However, the reason that `qsTile` was (or rather, `getQsTile()` returned) `null` is unknown.

The documentation only says that the returned tile is only valid between the `onStartListening()` and the `onStopListening()` callbacks. I changed the `super` call in `onStopListening()` to be in the end of the callback, just to make sure we use a more cautious interpretation of the documentation by not calling `getQsTile()` after calling `super.onStopListening()`. Still, looking at the code in AOSP it seems that changing the order shouldn't change anything.

Looking at the `TileService` code in AOSP, I couldn't find a reason for the `mTile` field to be `null`, except if there's an issue in Android and when binding to the internal service it fails to get the tile instance, or if the `getQsTile()` method is called too soon, which doesn't seem to be the case here.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2140)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@ Line wrap the file at 100 chars.                                              Th
 - Fix the notification sometimes leaving the foreground and becoming dismissable even if the UI was
   still visible.
 - Fix crash if connection to service is lost while opening the Split Tunneling settings screen.
+- Fix rare crash that could occur when the tunnel state changes when showing or hiding the quick
+  settings tile.
 
 #### Linux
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2140)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2142, opened 2020-09-23
Allow expanding locations with one relay on Android

jvff:
Previously the Select Location screen would not show child items for locations that had only one relay. However, that might lead to some UX issues, where the user can't see what city is available in a country, and can't select a single server if more are added later without having to change the constraint again.

This PR changes that so that all locations with at least one relay can be fully expanded.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Minor UI tweak, but if a changelog entry is necessary, we can add one for all platforms later when they are updated.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2142)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/relaylist/RelayCity.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/relaylist/RelayCity.kt
@@ -53,6 +53,4 @@ class RelayCity(
             return 1
         }
     }
-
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2142)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2147, opened 2020-09-24
Allow removal of accounts from the account history

jvff:
This is the final PR to change the account history in order to add a removal button. This PR implements sending the remove account from history request to the daemon, as well as the UI parts to trigger the request.

The list is only updated when a new history is received from the daemon. That means that it is not simple to add an animation for when an entry is removed. Since it requires some planning as to how to change the data flow in order to detect which entry is removed, this can be done in the future in a separate PR if needed.

One issue with the update to a new history with one entry removed is that the account login widget would jump a bit vertically with the update. That happened because the widget would calculate it's maximum size based on the number of entries. The PR also works around that by changing the calculation to use the maximum number of account history entries.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2147)


pinkisemils (approved):
> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2147)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2147)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2148, opened 2020-09-25
Reset focus on navigation and other accessibility improvements

raksooo:
This PR:
* Resets focus on navigation to either the first header or some fallback element
* Announces navigation with an invisible `div` that has the `aria-live` argument set
* Restores focus when closing a modal
* Makes notification area polite since that seems more correct
* Change `role` to `status` for `NotificationArea` and `SecuredLabel` since `alert` should be used for errors.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148)

### DIFF
--- a/gui/src/renderer/components/Connect.tsx
+++ b/gui/src/renderer/components/Connect.tsx
@@ -6,6 +6,7 @@ import NotificationArea from '../components/NotificationArea';
 import { AuthFailureKind, parseAuthFailure } from '../../shared/auth-failure';
 import { LoginState } from '../redux/account/reducers';
 import { IConnectionReduxState } from '../redux/connection/reducers';
+import { FocusFallback } from './Focus';
 import { Brand, HeaderBarStyle, HeaderBarSettingsButton } from './HeaderBar';
 import ImageView from './ImageView';
 import { Container, Header, Layout } from './Layout';
@@ -89,7 +90,9 @@ export default class Connect extends React.Component<IProps, IState> {
       <ModalContainer>
         <Layout>
           <Header barStyle={this.headerBarStyle()}>
-            <Brand />
+            <FocusFallback>
+              <Brand />
+            </FocusFallback>
             <HeaderBarSettingsButton />
           </Header>
           <StyledContainer>
--- a/gui/src/renderer/components/Focus.tsx
+++ b/gui/src/renderer/components/Focus.tsx
@@ -1,0 +1,63 @@
+import path from 'path';
+import React, { useImperativeHandle, useState } from 'react';
+import { useLocation } from 'react-router';
+import { sprintf } from 'sprintf-js';
+import styled from 'styled-components';
+import { messages } from '../../shared/gettext';
+
+const PageChangeAnnouncer = styled.div({
+  width: 0,
+  height: 0,
+  overflow: 'hidden',
+});
+
+export interface IFocusHandle {
+  resetFocus(): void;
+}
+
+interface IFocusProps {
+  children?: React.ReactElement;
+}
+
+function Focus(props: IFocusProps, ref: React.Ref<IFocusHandle>) {
+  const location = useLocation();
+  const [title, setTitle] = useState<string>();
+
+  useImperativeHandle(
+    ref,
+    () => ({
+      resetFocus: () => {
+        const pageName = path.basename(location.pathname);
+        const titleElement = document.getElementsByTagName('h1')[0];
+        const titleContent = titleElement?.textContent ?? pageName;
+        setTitle(titleContent);
+
+        const focusElement = titleElement ?? document.getElementsByTagName('header')[0];
+        if (focusElement) {
+          focusElement.setAttribute('tabindex', '-1');
+          focusElement.focus();
+        }
+      },
+    }),
+    [location.pathname],
+  );
+
+  return (
+    <>
+      {title && (
+        <PageChangeAnnouncer aria-live="polite">
+          {
+            // TRANSLATORS: This string is used to notify users of screenreaders that the view has
+            // TRANSLATORS: changed, usually as a result of pressing a navigation button.
+            // TRANSLATORS: Available placeholders:
+            // TRANSLATORS: %(title)s - page title
+            sprintf(messages.pgettext('accessibility', '%(title)s, View loaded'), { title })
+          }
+        </PageChangeAnnouncer>
+      )}
+      {props.children}
+    </>
+  );
+}
+
+export default React.memo(React.forwardRef(Focus));
--- a/gui/src/renderer/components/Focus.tsx
+++ b/gui/src/renderer/components/Focus.tsx
@@ -5,6 +5,8 @@ import { sprintf } from 'sprintf-js';
 import styled from 'styled-components';
 import { messages } from '../../shared/gettext';
 
+const FOCUS_FALLBACK_CLASS = 'focus-fallback';
+
 const PageChangeAnnouncer = styled.div({
   width: 0,
   height: 0,
@@ -32,7 +34,8 @@ function Focus(props: IFocusProps, ref: React.Ref<IFocusHandle>) {
         const titleContent = titleElement?.textContent ?? pageName;
         setTitle(titleContent);
 
-        const focusElement = titleElement ?? document.getElementsByTagName('header')[0];
+        const focusElement =
+          titleElement ?? document.getElementsByClassName(FOCUS_FALLBACK_CLASS)[0];
         if (focusElement) {
           focusElement.setAttribute('tabindex', '-1');
           focusElement.focus();
@@ -61,3 +64,13 @@ function Focus(props: IFocusProps, ref: React.Ref<IFocusHandle>) {
 }
 
 export default React.memo(React.forwardRef(Focus));
+
+interface IFocusFallbackProps {
+  children: React.ReactElement;
+}
+
+export function FocusFallback(props: IFocusFallbackProps) {
+  return React.cloneElement(props.children, {
+    className: `${props.children.props.className} ${FOCUS_FALLBACK_CLASS}`,
+  });
+}
--- a/gui/src/renderer/components/HeaderBar.tsx
+++ b/gui/src/renderer/components/HeaderBar.tsx
@@ -66,9 +66,9 @@ const Logo = styled(ImageView)({
   margin: '4px 0 3px',
 });
 
-export function Brand() {
+export function Brand(props: React.HTMLAttributes<HTMLDivElement>) {
   return (
-    <BrandContainer>
+    <BrandContainer {...props}>
       <Logo width={44} height={44} source="logo-icon" />
       <Title>{messages.pgettext('generic', 'MULLVAD VPN')}</Title>
     </BrandContainer>
--- a/gui/src/renderer/components/ImageView.tsx
+++ b/gui/src/renderer/components/ImageView.tsx
@@ -53,9 +53,10 @@ export default function ImageView(props: IImageViewProps) {
       </ImageMask>
     );
   } else {
+    const { source: _source, width, height, ...otherProps } = props;
     return (
-      <Wrapper onClick={props.onClick} className={props.className}>
-        <img src={url} width={props.width} height={props.height} aria-hidden={true} />
+      <Wrapper {...otherProps}>
+        <img src={url} width={width} height={height} aria-hidden={true} />
       </Wrapper>
     );
   }
--- a/gui/src/renderer/components/Modal.tsx
+++ b/gui/src/renderer/components/Modal.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useRef, useState } from 'react';
+import React, { useContext, useEffect, useMemo, useRef, useState } from 'react';
 import ReactDOM from 'react-dom';
 import styled from 'styled-components';
 import { colors } from '../../config.json';
@@ -41,6 +41,7 @@ interface IModalContext {
   activeModal: boolean;
   setActiveModal: (value: boolean) => void;
   modalContainerRef: React.RefObject<HTMLDivElement>;
+  previousActiveElement: React.MutableRefObject<HTMLElement | undefined>;
 }
 
 const noActiveModalContextError = new Error('ActiveModalContext.Provider missing');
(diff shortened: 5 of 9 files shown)


pronebird (changes requested):
> Reviewed 4 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148)* status: 4 of 8 files reviewed, 2 unresolved discussions (waiting on @raksooo)

___
*[gui/src/renderer/components/Focus.tsx, line 56 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148#-MI3acYV3UUKicwv_sNc:-MI3acYW-FqGCAbM_i-c:b-tr3jwb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7c96ffc4bbff6dced8dfa208b9c7a8833af3a84/gui/src/renderer/components/Focus.tsx#L56)):*
> ```XML
>           // TRANSLATORS: Available placeholders:
>           // TRANSLATORS: %(title)s - page title
>           sprintf(messages.pgettext('accessibility', '%(title)s, View loaded'), { title })
> ```

Maybe "view appeared" instead?
___
*[gui/src/renderer/components/HeaderBar.tsx, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148#-MI3atL85tduLW86pKd1:-MI3atL85tduLW86pKd2:b89rxw8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7c96ffc4bbff6dced8dfa208b9c7a8833af3a84/gui/src/renderer/components/HeaderBar.tsx#L80)):*
> ```XML
>           aria-label={
>             // TRANSLATORS: This communicates the logo image to users of screen readers
>             messages.pgettext('accessibility', 'Mullvad logo')
> ```

Is there any way we can simply exclude the useless announcements like "this is a logo"? If we have a blind user this is absolutely irrelevant to announce as they won't see the logo anyway.

pronebird (commented):
> Reviewed 4 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148)* status: all files reviewed, 2 unresolved discussions (waiting on @raksooo)

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148)* status: all files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[gui/src/renderer/components/Focus.tsx, line 56 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148#-MI3acYV3UUKicwv_sNc:-MI3fe558yfiYGnNu4hD:bfgken2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7c96ffc4bbff6dced8dfa208b9c7a8833af3a84/gui/src/renderer/components/Focus.tsx#L56)):*

Maybe "view appeared" instead?
</blockquote></details>

I first used the string "View changed" but thought "loaded" made more sense but I'm not sure which one is best.
___
*[gui/src/renderer/components/HeaderBar.tsx, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148#-MI3atL85tduLW86pKd1:-MI3ceDIDpnCM67xDU3h:b-lpgnfp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7c96ffc4bbff6dced8dfa208b9c7a8833af3a84/gui/src/renderer/components/HeaderBar.tsx#L80)):*

Is there any way we can simply exclude the useless announcements like "this is a logo"? If we have a blind user this is absolutely irrelevant to announce as they won't see the logo anyway.
</blockquote></details>

That was my idea at first as well, but then I changed my mind since the GUI might be described with phrases like "To the right of the logo". Now when reading a bit more about it, all I can find is examples where the logo also is a link to the start page, and one common label is "<Company name>" which would be redundant with the same text next to it. I'm fine with either hiding it to screen readers or keeping it like it is.

Do you have a suggestion for which element we should focus when the view is loaded instead? Maybe `SecuredLabel`?

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[gui/src/renderer/components/HeaderBar.tsx, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148#-MI3atL85tduLW86pKd1:-MI3i8qFCCM0uxkZ2dkp:bs1yl2q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7c96ffc4bbff6dced8dfa208b9c7a8833af3a84/gui/src/renderer/components/HeaderBar.tsx#L80)):*
As I had mentioned before, when doing accessibility on iOS, I would attempt at grouping elements to avoid the unnecessary complexity for the blind users, to make it easier for them to navigate through elements & interact with the UI with ease. 

If I were to close my eyes and navigate through the app with screen reader, I would be mildly infuriated to stumble upon the logo (which I wouldn't be able to see) each time I need to change the server.

> Do you have a suggestion for which element we should focus when the view is loaded instead? Maybe SecuredLabel?

Ideally we should focus at the top of the view, at the heading, i.e Mullvad VPN, the next element should be settings button, then the banner with the status message & so forth. 

This aligns with the behaviour of navigation controllers on iOS. Just enable the voice over on your iPhone accessibility settings & try walking the system Settings app, you'll see what I mean. You can also assign the Accessibility Shortcut to Voice Over & triple press the home button to turn it on & off in any app or context.

I think iPhones have the best in class accessibility & an infinite source of knowledge in that regard which can be applied on desktop too.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[gui/src/renderer/components/HeaderBar.tsx, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148#-MI3atL85tduLW86pKd1:-MI3sbo64tc5A1RPffML:b-29ic9w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7c96ffc4bbff6dced8dfa208b9c7a8833af3a84/gui/src/renderer/components/HeaderBar.tsx#L80)):*

As I had mentioned before, when doing accessibility on iOS, I would attempt at grouping elements to avoid the unnecessary complexity for the blind users, to make it easier for them to navigate through elements & interact with the UI with ease. 

If I were to close my eyes and navigate through the app with screen reader, I would be mildly infuriated to stumble upon the logo (which I wouldn't be able to see) each time I need to change the server.

> Do you have a suggestion for which element we should focus when the view is loaded instead? Maybe SecuredLabel?

Ideally we should focus at the top of the view, at the heading, i.e Mullvad VPN, the next element should be settings button, then the banner with the status message & so forth. 

This aligns with the behaviour of navigation controllers on iOS. Just enable the voice over on your iPhone accessibility settings & try walking the system Settings app, you'll see what I mean. You can also assign the Accessibility Shortcut to Voice Over & triple press the home button to turn it on & off in any app or context.

I think iPhones have the best in class accessibility & an infinite source of knowledge in that regard which can be applied on desktop too.
</blockquote></details>

I don't know if I agree. When using screen readers on computers it's common to use commands that jumps to next input/button/heading etc. and if you're familiar with the app you'll just use the command that goes to the next button until you're on the "Select location" button.

I'm fine with hiding the logo from screen readers but not because it makes navigation more difficult. The reason I see is that it might be unnecessary information, but when reading about it I find opinions pointing both ways.

I looked around a bit on both macOS and iOS and found these two examples where in both cases it's possible to focus the logo and it says something like "Siri, image" and "Artwork, Mullvad VPN, image". Although another example that's more aligned with what you're saying is the "Find my iPhone"-app which only contains a logo and a login form for me and in that view it isn't possible to focus the logo.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[gui/src/renderer/components/HeaderBar.tsx, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148#-MI3atL85tduLW86pKd1:-MI3wAa55_9OLv8Ns27g:b2vpb8n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7c96ffc4bbff6dced8dfa208b9c7a8833af3a84/gui/src/renderer/components/HeaderBar.tsx#L80)):*

I don't know if I agree. When using screen readers on computers it's common to use commands that jumps to next input/button/heading etc. and if you're familiar with the app you'll just use the command that goes to the next button until you're on the "Select location" button.

I'm fine with hiding the logo from screen readers but not because it makes navigation more difficult. The reason I see is that it might be unnecessary information, but when reading about it I find opinions pointing both ways.

I looked around a bit on both macOS and iOS and found these two examples where in both cases it's possible to focus the logo and it says something like "Siri, image" and "Artwork, Mullvad VPN, image". Although another example that's more aligned with what you're saying is the "Find my iPhone"-app which only contains a logo and a login form for me and in that view it isn't possible to focus the logo.


</blockquote></details>

The black box around the Siri logo didn't make it to the screenshot though :o

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[gui/src/renderer/components/HeaderBar.tsx, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148#-MI3atL85tduLW86pKd1:-MI4-12x-zPIdzM3JUpX:b-klhiql) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7c96ffc4bbff6dced8dfa208b9c7a8833af3a84/gui/src/renderer/components/HeaderBar.tsx#L80)):*

The black box around the Siri logo didn't make it to the screenshot though :o
</blockquote></details>

Another example from Apple's "Tips" app, it says something like "The iOS12 logo, image"

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[gui/src/renderer/components/HeaderBar.tsx, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148#-MI3atL85tduLW86pKd1:-MI3wk1JAxLsJp5N7GVS:b-78nrmn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7c96ffc4bbff6dced8dfa208b9c7a8833af3a84/gui/src/renderer/components/HeaderBar.tsx#L80)):*

> I don't know if I agree. When using screen readers on computers it's common to use commands that jumps to next input/button/heading etc. and if you're familiar with the app you'll just use the command that goes to the next button until you're on the "Select location" button.

It's similar on mobile devices, you have what is called a rotor control where you can navigate different layers of the app. It's certainly less efficient than hitting a hot key on the keyboard, but still.

> I looked around a bit on both macOS and iOS and found these two examples where in both cases it's possible to focus the logo and it says something like "Siri, image" and "Artwork, Mullvad VPN, image". Although another example that's more aligned with what you're saying is the "Find my iPhone"-app which only contains a logo and a login form for me and in that view it isn't possible to focus the logo.

I suppose that each department at Apple implements accessibility in their own way. My opinion that a blind user wouldn't benefit from knowing that the view has an icon that he or she can't see. Maybe a color blind user would benefit from that information, but then what use does it have if it says "An artwork..".

Although a different context, you can see that in action if you walk over iPhone Settings app with voice over, where each cell basically a single "button" & not a collection of labels & images.

> Another example from Apple's "Tips" app, it says something like "The iOS12 logo, image"

I don't think it's an example of a good accessibility. In that case the image acts as a heading. They could have said "What's new in iOS 12" in the header instead. 

Seems like it's fixed in the iOS 14 tips app which says "What's new in iOS 14" which I think makes more sense in terms of accessibility. However the navigation title, which is hidden on the screenshot & located next to the back button in the navigation bar, still says "What's new" when voice over is on.



Last but not least. I'd suggest to skim through a couple of videos from WWDC on accessibility. I don't remember which ones I looked at back in 2016 but some of those were really good. You can ignore the technical part & just focus on the interaction. They bring folks with disabilities on stage to demo how they make their lives easier. Maybe they have something for macOS/desktop too?

https://developer.apple.com/videos/frameworks/accessibility/

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148)* status: 6 of 9 files reviewed, all discussions resolved (waiting on @pronebird)

___
*[gui/src/renderer/components/HeaderBar.tsx, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2148#-MI3atL85tduLW86pKd1:-MI4NGkB5pkIOmJFQyIY:b-eunece) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7c96ffc4bbff6dced8dfa208b9c7a8833af3a84/gui/src/renderer/components/HeaderBar.tsx#L80)):*


> I don't know if I agree. When using screen readers on computers it's common to use commands that jumps to next input/button/heading etc. and if you're familiar with the app you'll just use the command that goes to the next button until you're on the "Select location" button.

It's similar on mobile devices, you have what is called a rotor control where you can navigate different layers of the app. It's certainly less efficient than hitting a hot key on the keyboard, but still.

> I looked around a bit on both macOS and iOS and found these two examples where in both cases it's possible to focus the logo and it says something like "Siri, image" and "Artwork, Mullvad VPN, image". Although another example that's more aligned with what you're saying is the "Find my iPhone"-app which only contains a logo and a login form for me and in that view it isn't possible to focus the logo.

I suppose that each department at Apple implements accessibility in their own way. My opinion that a blind user wouldn't benefit from knowing that the view has an icon that he or she can't see. Maybe a color blind user would benefit from that information, but then what use does it have if it says "An artwork..".

Although a different context, you can see that in action if you walk over iPhone Settings app with voice over, where each cell basically a single "button" & not a collection of labels & images.

> Another example from Apple's "Tips" app, it says something like "The iOS12 logo, image"

I don't think it's an example of a good accessibility. In that case the image acts as a heading. They could have said "What's new in iOS 12" in the header instead. 

Seems like it's fixed in the iOS 14 tips app which says "What's new in iOS 14" which I think makes more sense in terms of accessibility. However the navigation title, which is hidden on the screenshot & located next to the back button in the navigation bar, still says "What's new" when voice over is on.



Last but not least. I'd suggest to skim through a couple of videos from WWDC on accessibility. I don't remember which ones I looked at back in 2016 but some of those were really good. You can ignore the technical part & just focus on the interaction. They bring folks with disabilities on stage to demo how they make their lives easier. Maybe they have something for macOS/desktop too?

https://developer.apple.com/videos/frameworks/accessibility/
</blockquote></details>

Logo has now been removed.

raksooo:
I merged this since there were minimal changes since the review was completed previously, and those changes were agreed upon.

PR #2149, opened 2020-09-25
Start key rotation after initial key is created.

pinkisemils:
When a new account is set, and a key is generated, the daemon previously didn't start the key rotation timer. The effect is that if a new user never reboots, the key would not be rotated indefinitely. I've changed the daemon so that whenever a new key event is received and the current account has no key, the key rotation would be started. The reason for not starting the key rotation whenever a new key is created is because the key rotation will almost always be started already.

The last commit here is a small refactoring to simplify the code for key rotation/generation. This is purely optional and I can drop it, since we're so close to a release.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ Line wrap the file at 100 chars.                                              Th
 
 ### Fixed
 - Stop resetting the firewall after an upgrade to not leak after an upgrade.
+- Start key rotation when WireGuard key is first created.
 
 #### Windows
 - Remove firewall filters (unblock internet access) when "Always require VPN" is enabled and the app
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -623,22 +623,6 @@ where
 
         daemon.ensure_wireguard_keys_for_current_account().await;
 
-        if let Some(token) = daemon.settings.get_account_token() {
-            daemon
-                .wireguard_key_manager
-                .set_rotation_interval(
-                    &mut daemon.account_history,
-                    token,
-                    daemon
-                        .settings
-                        .tunnel_options
-                        .wireguard
-                        .automatic_rotation
-                        .map(|hours| Duration::from_secs(60u64 * 60u64 * hours as u64)),
-                )
-                .await;
-        }
-
         Ok(daemon)
     }
 
@@ -1084,6 +1068,8 @@ where
                         account: account.clone(),
                         wireguard: None,
                     });
+                // if no key existed before
+                let first_key_for_account_on_host = account_entry.wireguard.is_none();
                 account_entry.wireguard = Some(data);
                 match self.account_history.insert(account_entry).await {
                     Ok(_) => {
@@ -1091,7 +1077,10 @@ where
                             self.schedule_reconnect(WG_RECONNECT_DELAY).await;
                         }
                         self.event_listener
-                            .notify_key_event(KeygenEvent::NewKey(public_key))
+                            .notify_key_event(KeygenEvent::NewKey(public_key));
+                        if first_key_for_account_on_host {
+                            self.ensure_key_rotation().await;
+                        }
                     }
                     Err(e) => {
                         log::error!(
@@ -1117,6 +1106,21 @@ where
                 self.event_listener
                     .notify_key_event(KeygenEvent::GenerationFailure);
             }
+        }
+    }
+
+    async fn ensure_key_rotation(&mut self) {
+        if let Some(token) = self.settings.get_account_token() {
+            let rotation_interval = self
+                .settings
+                .tunnel_options
+                .wireguard
+                .automatic_rotation
+                .map(|hours| Duration::from_secs(60u64 * 60u64 * hours as u64));
+
+            self.wireguard_key_manager
+                .set_rotation_interval(&mut self.account_history, token, rotation_interval)
+                .await;
         }
     }
 
@@ -1342,13 +1346,6 @@ where
             }
 
             self.ensure_wireguard_keys_for_current_account().await;
-
-            if let Some(token) = account_token {
-                // update automatic rotation
-                self.wireguard_key_manager
-                    .reset_rotation(&mut self.account_history, token)
-                    .await;
-            }
         }
         Ok(account_changed)
     }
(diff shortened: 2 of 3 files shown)


dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149)* status: all files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[mullvad-daemon/src/wireguard.rs, line 157 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149#-MI4KzVQ53BnZmhLVyRm:-MI4KzVR1yn3MenYCZ7s:bfmnlu9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e3894cb3e8eee20fd615665ddc454c7eaf68a81/mullvad-daemon/src/wireguard.rs#L157)):*
> ```Rust
> 
>     /// Generate a new private key asynchronously. The new keys will be sent to the daemon channel.
>     pub async fn generate_key_async(&mut self, account: AccountToken, timeout: Option<Duration>) {
> ```

I see. `generate_key_async` does not wait until the key has been received and the settings/account updated. The daemon called `set_rotation` immediately after `generate_key_async`, so the account was likely to have no associated key?
___
*[mullvad-daemon/src/wireguard.rs, line 220 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149#-MI4TslC6Z2f1yuyXsnz:-MI4LBVCEs9znl2j96bD:b-ic65aq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e3894cb3e8eee20fd615665ddc454c7eaf68a81/mullvad-daemon/src/wireguard.rs#L220)):*
> ```Rust
>                 Ok(wireguard_data) => {
>                     let public_key = wireguard_data.get_public_key();
>                     let _ = daemon_tx.send(InternalDaemonEvent::WgKeyEvent((
> ```

Should it be verified that the settings are actually updated, and not start the rotation task otherwise? `Daemon::handle_wireguard_key_event` can fail if the active account has changed or if the account history cannot be saved, both of which seem possible.
___
*[mullvad-daemon/src/wireguard.rs, line 427 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149#-MI4TDNT7F5WQwCCL_H4:-MI4TDNT7F5WQwCCL_H5:b-orwgy5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e3894cb3e8eee20fd615665ddc454c7eaf68a81/mullvad-daemon/src/wireguard.rs#L427)):*
> ```Rust
> 
>     fn stop_automatic_rotation(&mut self) {
>         if let Some(abort_handle) = self.abort_scheduler_tx.take() {
> ```

Won't this fail to cancel the key rotation task started by `generate_key_async`?

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[mullvad-daemon/src/wireguard.rs, line 157 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149#-MI4KzVQ53BnZmhLVyRm:-MI4dggg9RS9YNiPgrnM:bb77ucv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e3894cb3e8eee20fd615665ddc454c7eaf68a81/mullvad-daemon/src/wireguard.rs#L157)):*

I see. `generate_key_async` does not wait until the key has been received and the settings/account updated. The daemon called `set_rotation` immediately after `generate_key_async`, so the account was likely to have no associated key?
</blockquote></details>

Yep.
___
*[mullvad-daemon/src/wireguard.rs, line 220 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149#-MI4TslC6Z2f1yuyXsnz:-MI4dr1s32uSnEs7de-s:bxtd8bs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e3894cb3e8eee20fd615665ddc454c7eaf68a81/mullvad-daemon/src/wireguard.rs#L220)):*

Should it be verified that the settings are actually updated, and not start the rotation task otherwise? `Daemon::handle_wireguard_key_event` can fail if the active account has changed or if the account history cannot be saved, both of which seem possible.
</blockquote></details>

In the case that the account has been changed, this future will be cancelled anyway.
___
*[mullvad-daemon/src/wireguard.rs, line 427 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149#-MI4TDNT7F5WQwCCL_H4:-MI4dm8fCSt8OeokEPcD:b-vg120d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e3894cb3e8eee20fd615665ddc454c7eaf68a81/mullvad-daemon/src/wireguard.rs#L427)):*

Won't this fail to cancel the key rotation task started by `generate_key_async`?
</blockquote></details>

Oh, it will. I was under the false impression that `self.current_job` would hold both key generation and key rotation futures. I guess I'll just drop the last commit as it introduces more trouble than it seems to be worth.

dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149)* status: all files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[mullvad-daemon/src/lib.rs, line 625 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149#-MIJ31yt233reMxmh_HP:-MIJ31yt233reMxmh_HQ:bxnwsvo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2948623d95f1793b4eb544fa5b9c20a9d2490340/mullvad-daemon/src/lib.rs#L625)):*
> ```Rust
> 
>         daemon.ensure_wireguard_keys_for_current_account().await;
>         daemon.ensure_key_rotation().await;
> ```

This should probably still only be called in the case where no key will be generated
___
*[mullvad-daemon/src/wireguard.rs, line 220 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149#-MI4TslC6Z2f1yuyXsnz:-MIJ57V48hD6YgQqgtcm:b942g1l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e3894cb3e8eee20fd615665ddc454c7eaf68a81/mullvad-daemon/src/wireguard.rs#L220)):*

In the case that the account has been changed, this future will be cancelled anyway.
</blockquote></details>

I still think that it's somewhat difficult to follow all of the logic related to key rotation.

`generate_key_async` is a somewhat confusing name now, since it isn't *obvious* that it uses `tokio::spawn` and returns before its' done when `await`ed. Renaming it to `spawn_key_generate_task` or something might help.

I wonder if it could be less tightly coupled to `Daemon`. E.g., I haven't thought about it much, but maybe sending `InternalDaemonEvent`s directly is bad.

That's for some other time, though.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2149)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

PR #2156, opened 2020-10-01
Add option to make the CLI wait for the target state

raksooo:
This PR adds an argument (`--wait`) to the `connect`, `disconnect` and `reconnect` CLI commands which keeps the CLI running until the target state is reached. This will make it possible to run commands in sequence in a bash script, useful for automated testing for example.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ Line wrap the file at 100 chars.                                              Th
 ## [Unreleased]
 ### Added
 - Improve accessibility in the desktop app.
+- Add `--wait` flag to `connect`, `disconnect` and `reconnect` CLI subcommands to make the CLI wait
+  for the target state to be reached before exiting.
 
 ### Changed
 - Use the API to fetch API IP addresses instead of DNS.
--- a/mullvad-cli/src/cmds/connect.rs
+++ b/mullvad-cli/src/cmds/connect.rs
@@ -1,5 +1,6 @@
-use crate::{new_rpc_client, Command, Result};
-use talpid_types::ErrorExt;
+use crate::{format, new_rpc_client, state, Command, Error, Result};
+use futures::StreamExt;
+use mullvad_management_interface::types::tunnel_state::State;
 
 pub struct Connect;
 
@@ -12,13 +13,38 @@ impl Command for Connect {
     fn clap_subcommand(&self) -> clap::App<'static, 'static> {
         clap::SubCommand::with_name(self.name())
             .about("Command the client to start establishing a VPN tunnel")
+            .arg(
+                clap::Arg::with_name("wait")
+                    .long("wait")
+                    .short("w")
+                    .help("Wait until connected before exiting"),
+            )
     }
 
-    async fn run(&self, _: &clap::ArgMatches<'_>) -> Result<()> {
+    async fn run(&self, matches: &clap::ArgMatches<'_>) -> Result<()> {
         let mut rpc = new_rpc_client().await?;
-        if let Err(e) = rpc.connect_tunnel(()).await {
-            eprintln!("{}", e.display_chain());
+
+        let receiver_option = if matches.is_present("wait") {
+            Some(state::state_listen(rpc.clone()))
+        } else {
+            None
+        };
+
+        if rpc.connect_tunnel(()).await?.into_inner() {
+            if let Some(mut receiver) = receiver_option {
+                while let Some(state) = receiver.next().await {
+                    let state = state?;
+                    format::print_state(&state);
+                    match state.state.unwrap() {
+                        State::Connected(_) => return Ok(()),
+                        State::Error(_) => return Err(Error::CommandFailed("connect")),
+                        _ => {}
+                    }
+                }
+                return Err(Error::StatusListenerFailed);
+            }
         }
+
         Ok(())
     }
 }
--- a/mullvad-cli/src/cmds/disconnect.rs
+++ b/mullvad-cli/src/cmds/disconnect.rs
@@ -1,4 +1,6 @@
-use crate::{new_rpc_client, Command, Result};
+use crate::{format, new_rpc_client, state, Command, Error, Result};
+use futures::StreamExt;
+use mullvad_management_interface::types::tunnel_state::State::Disconnected;
 
 pub struct Disconnect;
 
@@ -11,11 +13,37 @@ impl Command for Disconnect {
     fn clap_subcommand(&self) -> clap::App<'static, 'static> {
         clap::SubCommand::with_name(self.name())
             .about("Command the client to disconnect the VPN tunnel")
+            .arg(
+                clap::Arg::with_name("wait")
+                    .long("wait")
+                    .short("w")
+                    .help("Wait until disconnected before exiting"),
+            )
     }
 
-    async fn run(&self, _: &clap::ArgMatches<'_>) -> Result<()> {
+    async fn run(&self, matches: &clap::ArgMatches<'_>) -> Result<()> {
         let mut rpc = new_rpc_client().await?;
-        rpc.disconnect_tunnel(()).await?;
+
+        let receiver_option = if matches.is_present("wait") {
+            Some(state::state_listen(rpc.clone()))
+        } else {
+            None
+        };
+
+        if rpc.disconnect_tunnel(()).await?.into_inner() {
+            if let Some(mut receiver) = receiver_option {
+                while let Some(state) = receiver.next().await {
+                    let state = state?;
+                    format::print_state(&state);
+                    match state.state.unwrap() {
+                        Disconnected(_) => return Ok(()),
+                        _ => {}
+                    }
+                }
+                return Err(Error::StatusListenerFailed);
+            }
+        }
+
         Ok(())
     }
 }
--- a/mullvad-cli/src/cmds/reconnect.rs
+++ b/mullvad-cli/src/cmds/reconnect.rs
@@ -1,5 +1,6 @@
-use crate::{new_rpc_client, Command, Result};
-use talpid_types::ErrorExt;
+use crate::{format, new_rpc_client, state, Command, Error, Result};
+use futures::StreamExt;
+use mullvad_management_interface::types::tunnel_state::State;
 
 pub struct Reconnect;
 
@@ -10,14 +11,40 @@ impl Command for Reconnect {
     }
 
     fn clap_subcommand(&self) -> clap::App<'static, 'static> {
-        clap::SubCommand::with_name(self.name()).about("Command the client to reconnect")
+        clap::SubCommand::with_name(self.name())
+            .about("Command the client to reconnect")
+            .arg(
+                clap::Arg::with_name("wait")
+                    .long("wait")
+                    .short("w")
+                    .help("Wait until reconnected before exiting"),
+            )
     }
 
-    async fn run(&self, _: &clap::ArgMatches<'_>) -> Result<()> {
+    async fn run(&self, matches: &clap::ArgMatches<'_>) -> Result<()> {
         let mut rpc = new_rpc_client().await?;
-        if let Err(e) = rpc.reconnect_tunnel(()).await {
-            eprintln!("{}", e.display_chain());
+
+        let receiver_option = if matches.is_present("wait") {
+            Some(state::state_listen(rpc.clone()))
+        } else {
+            None
+        };
+
+        if rpc.reconnect_tunnel(()).await?.into_inner() {
+            if let Some(mut receiver) = receiver_option {
+                while let Some(state) = receiver.next().await {
+                    let state = state?;
+                    format::print_state(&state);
+                    match state.state.unwrap() {
+                        State::Connected(_) => return Ok(()),
+                        State::Error(_) => return Err(Error::CommandFailed("reconnect")),
+                        _ => {}
+                    }
+                }
+                return Err(Error::StatusListenerFailed);
+            }
         }
+
         Ok(())
     }
 }
--- a/mullvad-cli/src/cmds/status.rs
+++ b/mullvad-cli/src/cmds/status.rs
@@ -1,17 +1,7 @@
-use crate::{format::print_keygen_event, new_rpc_client, Command, Error, Result};
+use crate::{format, format::print_keygen_event, new_rpc_client, Command, Error, Result};
 use mullvad_management_interface::{
-    types::{
-        daemon_event::Event as EventType,
-        error_state::{
-            firewall_policy_error::ErrorType as FirewallPolicyErrorType, Cause as ErrorStateCause,
-            FirewallPolicyError, GenerationError,
-        },
-        ErrorState, ProxyType, TransportProtocol, TunnelEndpoint, TunnelState, TunnelType,
-    },
-    ManagementServiceClient,
+    types::daemon_event::Event as EventType, ManagementServiceClient,
 };
-use mullvad_types::auth_failed::AuthFailed;
-use std::fmt::Write;
 
 pub struct Status;
 
@@ -45,7 +35,7 @@ impl Command for Status {
         let mut rpc = new_rpc_client().await?;
         let state = rpc.get_tunnel_state(()).await?.into_inner();
 
-        print_state(&state);
+        format::print_state(&state);
         if matches.is_present("location") {
             print_location(&mut rpc).await?;
         }
@@ -58,7 +48,7 @@ impl Command for Status {
             while let Some(event) = events.message().await? {
                 match event.event.unwrap() {
                     EventType::TunnelState(new_state) => {
-                        print_state(&new_state);
+                        format::print_state(&new_state);
                         use mullvad_management_interface::types::tunnel_state::State::*;
                         match new_state.state.unwrap() {
                             Connected(..) | Disconnected(..) => {
@@ -98,151 +88,6 @@ impl Command for Status {
     }
 }
 
-fn print_state(state: &TunnelState) {
-    use mullvad_management_interface::types::{tunnel_state, tunnel_state::State::*};
-
-    print!("Tunnel status: ");
-    match state.state.as_ref().unwrap() {
-        Error(error) => print_error_state(error.error_state.as_ref().unwrap()),
-        Connected(tunnel_state::Connected { relay_info }) => {
-            let endpoint = relay_info
-                .as_ref()
-                .unwrap()
-                .tunnel_endpoint
-                .as_ref()
-                .unwrap();
-            println!("Connected to {}", format_endpoint(&endpoint));
-        }
-        Connecting(tunnel_state::Connecting { relay_info }) => {
-            let endpoint = relay_info
-                .as_ref()
-                .unwrap()
-                .tunnel_endpoint
-                .as_ref()
-                .unwrap();
-            println!("Connecting to {}...", format_endpoint(&endpoint));
-        }
-        Disconnected(_) => println!("Disconnected"),
-        Disconnecting(_) => println!("Disconnecting..."),
-    }
-}
-
-fn format_endpoint(endpoint: &TunnelEndpoint) -> String {
-    let mut out = format!(
-        "{} {} over {}",
-        match TunnelType::from_i32(endpoint.tunnel_type).expect("unknown tunnel protocol") {
-            TunnelType::Wireguard => "WireGuard",
-            TunnelType::Openvpn => "OpenVPN",
-        },
-        endpoint.address,
-        format_protocol(
-            TransportProtocol::from_i32(endpoint.protocol).expect("unknown transport protocol")
-        ),
-    );
-
-    if let Some(ref proxy) = endpoint.proxy {
-        write!(
-            &mut out,
-            " via {} {} over {}",
-            match ProxyType::from_i32(proxy.proxy_type).expect("unknown proxy type") {
-                ProxyType::Shadowsocks => "Shadowsocks",
-                ProxyType::Custom => "custom bridge",
-            },
-            proxy.address,
-            format_protocol(
-                TransportProtocol::from_i32(proxy.protocol).expect("unknown transport protocol")
-            ),
-        )
-        .unwrap();
-    }
-
-    out
-}
-
-fn print_error_state(error_state: &ErrorState) {
-    if error_state.blocking_error.is_some() {
-        eprintln!("Mullvad daemon failed to setup firewall rules!");
-        eprintln!("Deamon cannot block traffic from flowing, non-local traffic will leak");
-    }
-
-    match ErrorStateCause::from_i32(error_state.cause) {
-        Some(ErrorStateCause::AuthFailed) => {
-            println!(
-                "Blocked: {}",
-                AuthFailed::from(error_state.auth_fail_reason.as_ref())
-            );
-        }
-        #[cfg(target_os = "linux")]
-        Some(ErrorStateCause::SetFirewallPolicyError) => {
-            println!("Blocked: {}", error_state_to_string(error_state));
-            println!("Your kernel might be terribly out of date or missing nftables");
-        }
-        _ => println!("Blocked: {}", error_state_to_string(error_state)),
-    }
-}
-
-fn error_state_to_string(error_state: &ErrorState) -> String {
-    use ErrorStateCause::*;
-
-    let error_str = match ErrorStateCause::from_i32(error_state.cause).expect("unknown error cause")
-    {
-        AuthFailed => {
-            return if error_state.auth_fail_reason.is_empty() {
-                "Authentication with remote server failed".to_string()
-            } else {
-                format!(
-                    "Authentication with remote server failed: {}",
-                    error_state.auth_fail_reason
-                )
-            };
-        }
-        Ipv6Unavailable => "Failed to configure IPv6 because it's disabled in the platform",
-        SetFirewallPolicyError => {
-            return policy_error_to_string(error_state.policy_error.as_ref().unwrap())
-        }
-        SetDnsError => "Failed to set system DNS server",
-        StartTunnelError => "Failed to start connection to remote server",
-        TunnelParameterError => {
-            return format!(
-                "Failure to generate tunnel parameters: {}",
-                tunnel_parameter_error_to_string(error_state.parameter_error)
-            );
-        }
-        IsOffline => "This device is offline, no tunnels can be established",
-        TapAdapterProblem => "A problem with the TAP adapter has been detected",
-        #[cfg(target_os = "android")]
-        VpnPermissionDenied => "The Android VPN permission was denied when creating the tunnel",
-        #[cfg(not(target_os = "android"))]
-        _ => unreachable!("unknown error cause"),
-    };
-
-    error_str.to_string()
-}
-
-fn tunnel_parameter_error_to_string(parameter_error: i32) -> &'static str {
-    match GenerationError::from_i32(parameter_error).expect("unknown generation error") {
-        GenerationError::NoMatchingRelay => "Failure to select a matching tunnel relay",
-        GenerationError::NoMatchingBridgeRelay => "Failure to select a matching bridge relay",
-        GenerationError::NoWireguardKey => "No wireguard key available",
-        GenerationError::CustomTunnelHostResolutionError => {
-            "Can't resolve hostname for custom tunnel host"
-        }
-    }
-}
-
-fn policy_error_to_string(policy_error: &FirewallPolicyError) -> String {
-    let cause = match FirewallPolicyErrorType::from_i32(policy_error.r#type)
-        .expect("unknown policy error")
-    {
-        FirewallPolicyErrorType::Generic => return "Failed to set firewall policy".to_string(),
-        FirewallPolicyErrorType::Locked => format!(
-            "An application prevented the firewall policy from being set: {} (pid {})",
-            policy_error.lock_name, policy_error.lock_pid
-        ),
-    };
-    format!("Failed to set firewall policy: {}", cause)
-}
-
 async fn print_location(rpc: &mut ManagementServiceClient) -> Result<()> {
     let location = rpc.get_current_location(()).await;
     let location = match location {
@@ -277,11 +122,4 @@ async fn print_location(rpc: &mut ManagementServiceClient) -> Result<()> {
         location.latitude, location.longitude
     );
     Ok(())
-}
-
-fn format_protocol(protocol: TransportProtocol) -> &'static str {
-    match protocol {
-        TransportProtocol::Udp => "UDP",
-        TransportProtocol::Tcp => "TCP",
-    }
 }
--- a/mullvad-cli/src/format.rs
+++ b/mullvad-cli/src/format.rs
@@ -1,4 +1,14 @@
-use mullvad_management_interface::types::KeygenEvent;
+use mullvad_management_interface::types::{
+    error_state::{
+        firewall_policy_error::ErrorType as FirewallPolicyErrorType, Cause as ErrorStateCause,
+        FirewallPolicyError, GenerationError,
+    },
+    tunnel_state,
+    tunnel_state::State::*,
+    ErrorState, KeygenEvent, ProxyType, TransportProtocol, TunnelEndpoint, TunnelState, TunnelType,
+};
+use mullvad_types::auth_failed::AuthFailed;
+use std::fmt::Write;
 
 pub fn print_keygen_event(key_event: &KeygenEvent) {
     use mullvad_management_interface::types::keygen_event::KeygenEvent as EventType;
@@ -16,5 +26,155 @@ pub fn print_keygen_event(key_event: &KeygenEvent) {
         EventType::GenerationFailure => {
             println!("Failed to generate new WireGuard key");
         }
+    }
+}
+
+pub fn print_state(state: &TunnelState) {
+    print!("Tunnel status: ");
+    match state.state.as_ref().unwrap() {
+        Error(error) => print_error_state(error.error_state.as_ref().unwrap()),
+        Connected(tunnel_state::Connected { relay_info }) => {
+            let endpoint = relay_info
+                .as_ref()
+                .unwrap()
+                .tunnel_endpoint
+                .as_ref()
+                .unwrap();
+            println!("Connected to {}", format_endpoint(&endpoint));
+        }
+        Connecting(tunnel_state::Connecting { relay_info }) => {
+            let endpoint = relay_info
+                .as_ref()
+                .unwrap()
+                .tunnel_endpoint
+                .as_ref()
+                .unwrap();
+            println!("Connecting to {}...", format_endpoint(&endpoint));
+        }
+        Disconnected(_) => println!("Disconnected"),
+        Disconnecting(_) => println!("Disconnecting..."),
+    }
+}
+
+fn format_endpoint(endpoint: &TunnelEndpoint) -> String {
+    let mut out = format!(
+        "{} {} over {}",
+        match TunnelType::from_i32(endpoint.tunnel_type).expect("unknown tunnel protocol") {
+            TunnelType::Wireguard => "WireGuard",
+            TunnelType::Openvpn => "OpenVPN",
+        },
+        endpoint.address,
+        format_protocol(
+            TransportProtocol::from_i32(endpoint.protocol).expect("unknown transport protocol")
+        ),
+    );
+
+    if let Some(ref proxy) = endpoint.proxy {
+        write!(
+            &mut out,
+            " via {} {} over {}",
+            match ProxyType::from_i32(proxy.proxy_type).expect("unknown proxy type") {
+                ProxyType::Shadowsocks => "Shadowsocks",
+                ProxyType::Custom => "custom bridge",
+            },
+            proxy.address,
+            format_protocol(
+                TransportProtocol::from_i32(proxy.protocol).expect("unknown transport protocol")
+            ),
+        )
+        .unwrap();
+    }
+
+    out
+}
+
+fn print_error_state(error_state: &ErrorState) {
+    if error_state.blocking_error.is_some() {
+        eprintln!("Mullvad daemon failed to setup firewall rules!");
+        eprintln!("Deamon cannot block traffic from flowing, non-local traffic will leak");
+    }
+
+    match ErrorStateCause::from_i32(error_state.cause) {
+        Some(ErrorStateCause::AuthFailed) => {
+            println!(
+                "Blocked: {}",
+                AuthFailed::from(error_state.auth_fail_reason.as_ref())
+            );
+        }
+        #[cfg(target_os = "linux")]
+        Some(ErrorStateCause::SetFirewallPolicyError) => {
+            println!("Blocked: {}", error_state_to_string(error_state));
+            println!("Your kernel might be terribly out of date or missing nftables");
+        }
+        _ => println!("Blocked: {}", error_state_to_string(error_state)),
+    }
+}
+
+fn error_state_to_string(error_state: &ErrorState) -> String {
+    use ErrorStateCause::*;
+
+    let error_str = match ErrorStateCause::from_i32(error_state.cause).expect("unknown error cause")
+    {
+        AuthFailed => {
+            return if error_state.auth_fail_reason.is_empty() {
+                "Authentication with remote server failed".to_string()
+            } else {
+                format!(
+                    "Authentication with remote server failed: {}",
+                    error_state.auth_fail_reason
+                )
+            };
+        }
+        Ipv6Unavailable => "Failed to configure IPv6 because it's disabled in the platform",
+        SetFirewallPolicyError => {
+            return policy_error_to_string(error_state.policy_error.as_ref().unwrap())
+        }
+        SetDnsError => "Failed to set system DNS server",
+        StartTunnelError => "Failed to start connection to remote server",
+        TunnelParameterError => {
+            return format!(
+                "Failure to generate tunnel parameters: {}",
+                tunnel_parameter_error_to_string(error_state.parameter_error)
+            );
+        }
+        IsOffline => "This device is offline, no tunnels can be established",
+        TapAdapterProblem => "A problem with the TAP adapter has been detected",
+        #[cfg(target_os = "android")]
+        VpnPermissionDenied => "The Android VPN permission was denied when creating the tunnel",
+        #[cfg(not(target_os = "android"))]
+        _ => unreachable!("unknown error cause"),
+    };
+
+    error_str.to_string()
+}
+
+fn tunnel_parameter_error_to_string(parameter_error: i32) -> &'static str {
+    match GenerationError::from_i32(parameter_error).expect("unknown generation error") {
+        GenerationError::NoMatchingRelay => "Failure to select a matching tunnel relay",
+        GenerationError::NoMatchingBridgeRelay => "Failure to select a matching bridge relay",
+        GenerationError::NoWireguardKey => "No wireguard key available",
+        GenerationError::CustomTunnelHostResolutionError => {
+            "Can't resolve hostname for custom tunnel host"
+        }
+    }
+}
+
+fn policy_error_to_string(policy_error: &FirewallPolicyError) -> String {
+    let cause = match FirewallPolicyErrorType::from_i32(policy_error.r#type)
+        .expect("unknown policy error")
+    {
+        FirewallPolicyErrorType::Generic => return "Failed to set firewall policy".to_string(),
+        FirewallPolicyErrorType::Locked => format!(
+            "An application prevented the firewall policy from being set: {} (pid {})",
+            policy_error.lock_name, policy_error.lock_pid
+        ),
+    };
+    format!("Failed to set firewall policy: {}", cause)
+}
+
+fn format_protocol(protocol: TransportProtocol) -> &'static str {
+    match protocol {
+        TransportProtocol::Udp => "UDP",
+        TransportProtocol::Tcp => "TCP",
     }
 }
--- a/mullvad-cli/src/main.rs
+++ b/mullvad-cli/src/main.rs
@@ -10,6 +10,7 @@ pub use mullvad_management_interface::{self, new_rpc_client};
 mod cmds;
 mod format;
 mod location;
+mod state;
 
 pub const BIN_NAME: &str = "mullvad";
 pub const PRODUCT_VERSION: &str = include_str!(concat!(env!("OUT_DIR"), "/product-version.txt"));
@@ -33,6 +34,9 @@ pub enum Error {
 
     #[error(display = "Command failed: {}", _0)]
     CommandFailed(&'static str),
+
+    #[error(display = "Failed to listen for status updates")]
+    StatusListenerFailed,
 }
 
 #[tokio::main]
--- a/mullvad-cli/src/main.rs
+++ b/mullvad-cli/src/main.rs
@@ -30,6 +30,9 @@ pub enum Error {
     /// The given command is not correct in some way
     #[error(display = "Invalid command: {}", _0)]
     InvalidCommand(&'static str),
+
+    #[error(display = "Command failed: {}", _0)]
+    CommandFailed(&'static str),
 }
 
 #[tokio::main]
--- a/mullvad-cli/src/state.rs
+++ b/mullvad-cli/src/state.rs
@@ -1,0 +1,41 @@
+use crate::{Error, Result};
+use futures::{
+    channel::{mpsc, mpsc::Receiver},
+    SinkExt,
+};
+use mullvad_management_interface::{
+    types::{daemon_event::Event as EventType, TunnelState},
+    ManagementServiceClient,
+};
+
+// Spawns a new task that listens for tunnel state changes and forwards it through the returned
+// channel. Panics if called from outside of the Tokio runtime.
+pub fn state_listen(mut rpc: ManagementServiceClient) -> Receiver<Result<TunnelState>> {
+    let (mut sender, receiver) = mpsc::channel::<Result<TunnelState>>(1);
+    tokio::spawn(async move {
+        match rpc.events_listen(()).await {
+            Ok(events) => {
+                let mut events = events.into_inner();
+                loop {
+                    let forward = match events.message().await {
+                        Ok(Some(event)) => match event.event.unwrap() {
+                            EventType::TunnelState(new_state) => Ok(new_state),
+                            _ => continue,
+                        },
+                        Ok(None) => break,
+                        Err(status) => Err(Error::GrpcClientError(status)),
+                    };
+
+                    if let Err(_) = sender.send(forward).await {
+                        break;
+                    }
+                }
+            }
+            Err(status) => {
+                let _ = sender.send(Err(Error::GrpcClientError(status))).await;
+            }
+        }
+    });
+
+    receiver
+}
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -141,9 +141,9 @@ pub enum Error {
 /// Enum representing commands that can be sent to the daemon.
 pub enum DaemonCommand {
     /// Set target state. Does nothing if the daemon already has the state that is being set.
-    SetTargetState(oneshot::Sender<()>, TargetState),
+    SetTargetState(oneshot::Sender<bool>, TargetState),
     /// Reconnect the tunnel, if one is connecting/connected.
-    Reconnect,
+    Reconnect(oneshot::Sender<bool>),
     /// Request the current state.
     GetState(oneshot::Sender<TunnelState>),
     /// Get the current geographical location.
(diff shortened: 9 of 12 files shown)


raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 0 of 7 files reviewed, 2 unresolved discussions

___
*[mullvad-cli/src/state.rs, line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIYw5yu79DPh7hLL97t:-MIYw5yu79DPh7hLL97u:bssqlxy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94f6e2ca7a7e492316f015da82c8140f04ae64f7/mullvad-cli/src/state.rs#L11)):*
> ```Rust
>     rpc: &mut ManagementServiceClient,
>     begin_condition: impl Fn(&State) -> bool,
>     continue_condition: Box<dyn (Fn(&State) -> bool) + Send>,
> ```

Is there a better way to type this argument?
___
*[mullvad-cli/src/cmds/reconnect.rs, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIYvWSD8cSdz-FlaVwl:-MIYvWSD8cSdz-FlaVwm:b-o8zls7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94f6e2ca7a7e492316f015da82c8140f04ae64f7/mullvad-cli/src/cmds/reconnect.rs#L42)):*
> ```Rust
>                                 ..
>                             }),
>                     }) => false,
> ```

Is this correct? I found that the condition in the daemon was
```
if self.target_state == TargetState::Secured || self.tunnel_state.is_in_error_state() {
```
and figured it should be the same as this, since I don't have access to the current `TargetState`. Can this be expressed better?

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on @raksooo)

___
*[mullvad-cli/src/state.rs, line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIYw5yu79DPh7hLL97t:-MIYzFtADdfXnKo5tcix:b-11n20s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94f6e2ca7a7e492316f015da82c8140f04ae64f7/mullvad-cli/src/state.rs#L11)):*

Is there a better way to type this argument?
</blockquote></details>

Does not `impl Fn(&State) -> bool` work? If that does not work then maybe
```rust
pub async fn state_listen<C: Fn(&State) -> bool + Send + 'static>(continue_condition: C)
```
___
*[mullvad-cli/src/cmds/connect.rs, line 50 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIYzedeD50UGgx0X-An:-MIYzedeD50UGgx0X-Ao:bze90d8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/cmds/connect.rs#L50)):*
> ```Rust
>         }
>         if let Some(handle) = status_listen_handle {
>             handle.await.expect("Failed to listen to status updates");
> ```

Is it a bug if the await fails? Otherwise it would be great if it was handled as an error and not panic.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @raksooo)

___
*[mullvad-cli/src/cmds/reconnect.rs, line 52 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZ-9Q9DHZPXoI8gszl:-MIZ-9Q9DHZPXoI8gszm:b-rzhhu4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/cmds/reconnect.rs#L52)):*
> ```Rust
>                     _ => true,
>                 }),
>                 Box::new(format::print_state),
> ```

All invocations of `state::state_listen` use the exact same argument for `on_new_state`. So does it really have to take this as an argument? Can't it just call `format::print_state` directly?

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @raksooo)

___
*[mullvad-cli/src/state.rs, line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZB-UD-7VBQK_0BNKM:-MIZB-UD-7VBQK_0BNKN:b77r3eh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L14)):*
> ```Rust
>     on_new_state: Box<dyn Fn(&TunnelState) + Send>,
> ) -> Result<Option<JoinHandle<()>>> {
>     let state = rpc.get_tunnel_state(()).await?.into_inner();
> ```

This is the only awaiting that is done in this async function. And then it conditionally does something depending on the result. The `JoinHandle` is just a future, so essentially this function is `async fn -> Result<Option<Future>>`. Looks slightly off to me :thinking: 

Can't this function be focused only on the event listener stream part. Fetching and evaluating the initial state can be done somewhere else. Possibly directly in the connect/disconnect/reconnect subcommands themselves. A little bit of code duplication is not inherently evil. At least not when the alternative is to make the functions more complex to be generic enough.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @raksooo)

___
*[mullvad-cli/src/cmds/connect.rs, line 50 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIYzedeD50UGgx0X-An:-MIZCkRt9-HWNiHdybx4:b-9phb65) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/cmds/connect.rs#L50)):*

Is it a bug if the await fails? Otherwise it would be great if it was handled as an error and not panic.
</blockquote></details>

By bad. The `await` is on the `JoinHandle`. Which will only fail if the future panics. Which indeed would be a bug.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @raksooo)

___
*[mullvad-cli/src/state.rs, line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZB-UD-7VBQK_0BNKM:-MIZCzSS4Ohl103kPKeM:b-6k31zj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L14)):*

This is the only awaiting that is done in this async function. And then it conditionally does something depending on the result. The `JoinHandle` is just a future, so essentially this function is `async fn -> Result<Option<Future>>`. Looks slightly off to me :thinking: 

Can't this function be focused only on the event listener stream part. Fetching and evaluating the initial state can be done somewhere else. Possibly directly in the connect/disconnect/reconnect subcommands themselves. A little bit of code duplication is not inherently evil. At least not when the alternative is to make the functions more complex to be generic enough.
</blockquote></details>

It would also get rid of one of the closure arguments. Which is another thing that would make this function and all callsites easier to read. A function that takes two closures as arguments and both return just `false/true` is a bit hard to understand what they do without reading a lot of the implementation IMO.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @raksooo)

___
*[mullvad-cli/src/state.rs, line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZB-UD-7VBQK_0BNKM:-MIZNn1l4vcKkJZJXgc3:bdwdfn6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L14)):*

It would also get rid of one of the closure arguments. Which is another thing that would make this function and all callsites easier to read. A function that takes two closures as arguments and both return just `false/true` is a bit hard to understand what they do without reading a lot of the implementation IMO.
</blockquote></details>

Sure, sounds good :+1: I'll try it out!
___
*[mullvad-cli/src/cmds/reconnect.rs, line 52 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZ-9Q9DHZPXoI8gszl:-MIZO7yBCJi40pLIqHED:bl9mn5k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/cmds/reconnect.rs#L52)):*

All invocations of `state::state_listen` use the exact same argument for `on_new_state`. So does it really have to take this as an argument? Can't it just call `format::print_state` directly?
</blockquote></details>

Sounds reasonable. My first intention was to make use of `state_listen` in the `status` subcommand as well, but then realized it would complicate things a bit too much.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @raksooo)

___
*[mullvad-cli/src/state.rs, line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZB-UD-7VBQK_0BNKM:-MIZQtcHBqizjkiGSYeC:b-i5jb92) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L14)):*

Sure, sounds good :+1: I'll try it out!
</blockquote></details>

Would also not hurt with a line or two of docs to explain what the function does :)

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 0 of 7 files reviewed, 4 unresolved discussions (waiting on @raksooo)

___
*[mullvad-cli/src/cmds/reconnect.rs, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIYvWSD8cSdz-FlaVwl:-MIZQzLw7ZTJiPkkETBs:b-7lxzj4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94f6e2ca7a7e492316f015da82c8140f04ae64f7/mullvad-cli/src/cmds/reconnect.rs#L42)):*

Is this correct? I found that the condition in the daemon was
```
if self.target_state == TargetState::Secured || self.tunnel_state.is_in_error_state() {
```
and figured it should be the same as this, since I don't have access to the current `TargetState`. Can this be expressed better?
</blockquote></details>

Another solution to this would be to let the daemon include whether or not it actually initiates a reconnect.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[mullvad-cli/src/state.rs, line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIYw5yu79DPh7hLL97t:-MIZVDr89TSl1XDJohis:b8z50h5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94f6e2ca7a7e492316f015da82c8140f04ae64f7/mullvad-cli/src/state.rs#L11)):*

Does not `impl Fn(&State) -> bool` work? If that does not work then maybe
```rust
pub async fn state_listen<C: Fn(&State) -> bool + Send + 'static>(continue_condition: C)
```
</blockquote></details>

Works great! I thought I tried this but I must have made a mistake or might have been confused by some other error. Done.
___
*[mullvad-cli/src/state.rs, line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZB-UD-7VBQK_0BNKM:-MIZWDjg8EwWaZQI8kWO:brz0qox) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L14)):*

Would also not hurt with a line or two of docs to explain what the function does :)
</blockquote></details>

There's one more `await` a few lines down, when assigning `events`. I moved the evaluation of the initial state to connect, disconnect and reconnect as you suggested. That still leaves me with `Result<JoinHandle>`.
___
*[mullvad-cli/src/cmds/reconnect.rs, line 52 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZ-9Q9DHZPXoI8gszl:-MIZVSuACqTB79AvVdS7:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/cmds/reconnect.rs#L52)):*

Sounds reasonable. My first intention was to make use of `state_listen` in the `status` subcommand as well, but then realized it would complicate things a bit too much.
</blockquote></details>

Done.

pinkisemils (changes requested):
> Reviewed 2 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 2 of 7 files reviewed, 5 unresolved discussions (waiting on @faern and @raksooo)

___
*[mullvad-cli/src/state.rs, line 8 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_2p618VItIu4oa1pE:-MI_2p618VItIu4oa1pF:b-rn3p8x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L8)):*
> ```Rust
> use tokio::task::JoinHandle;
> 
> pub async fn state_listen(
> ```

Rather than have a function that takes two closures, would it not be easier to have a function that returns a stream that is constructed with the current state and a `rpc.events_listen()` ? Possibly the stream could be altered to only issue new items for the tunnel state events.

The consumers can drop the stream, thus closing the subscriber via something like 
```
let first_event = tunnel_events.next_item().await;
if let TunnelState::Connected == first_event { break };
while let Some(event) = tunnel_events.next_item().await {
     if let TunnelState::Connected = event { break };
}

```
Maybe the _current_ state doesn't even need to be a part of the same tunnel event stream. 
___
*[mullvad-cli/src/state.rs, line 20 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_5RTy1ORTao9PdllO:-MI_5RTy1ORTao9PdllP:bkcyell) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L20)):*
> ```Rust
>         Some(inner_state) if begin_condition(&inner_state) => {
>             let mut events = rpc.events_listen(()).await?.into_inner();
>             let join_handle = tokio::spawn(async move {
> ```

Does this really need to spawn a whole new future? It could just return either a stream or a future that resolves when the correct state is achieved.

pinkisemils (commented):
> Reviewed 1 of 7 files at r1, 2 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 5 of 7 files reviewed, 5 unresolved discussions (waiting on @faern and @raksooo)

___
*[mullvad-cli/src/cmds/reconnect.rs, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIYvWSD8cSdz-FlaVwl:-MIcA8HdE6DSH7XaRrlF:b-emfxn3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94f6e2ca7a7e492316f015da82c8140f04ae64f7/mullvad-cli/src/cmds/reconnect.rs#L42)):*

Another solution to this would be to let the daemon include whether or not it actually initiates a reconnect.
</blockquote></details>

This is not correct, as there's no reason to not reconnect when there's a blocking error. But I also like the idea of returning a boolean to indicate whether a reconnect was issued.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 5 of 7 files reviewed, 5 unresolved discussions (waiting on @faern and @raksooo)

___
*[mullvad-cli/src/state.rs, line 8 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_2p618VItIu4oa1pE:-MIcQniA51XlG1P7qz5p:b-ylyfoi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L8)):*

Rather than have a function that takes two closures, would it not be easier to have a function that returns a stream that is constructed with the current state and a `rpc.events_listen()` ? Possibly the stream could be altered to only issue new items for the tunnel state events.

The consumers can drop the stream, thus closing the subscriber via something like 
```
let first_event = tunnel_events.next_item().await;
if let TunnelState::Connected == first_event { break };
while let Some(event) = tunnel_events.next_item().await {
     if let TunnelState::Connected = event { break };
}

```
Maybe the _current_ state doesn't even need to be a part of the same tunnel event stream. 
</blockquote></details>

Sounds like a good idea, I'll try it out!
___
*[mullvad-cli/src/state.rs, line 20 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_5RTy1ORTao9PdllO:-MIcQvyKDWVc-Zb0pgkY:bhi72tl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L20)):*

Does this really need to spawn a whole new future? It could just return either a stream or a future that resolves when the correct state is achieved.
</blockquote></details>

The reason why I did this was to ensure that the we start listening for changes before running `connect`, but as long as `events_listen` is called before we should be fine I guess. I'll see what I can do!
___
*[mullvad-cli/src/cmds/reconnect.rs, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIYvWSD8cSdz-FlaVwl:-MIcQE-i2wCIwYkzkGDN:bga7sec) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94f6e2ca7a7e492316f015da82c8140f04ae64f7/mullvad-cli/src/cmds/reconnect.rs#L42)):*

This is not correct, as there's no reason to not reconnect when there's a blocking error. But I also like the idea of returning a boolean to indicate whether a reconnect was issued.
</blockquote></details>

I'll try to make the daemon return whether or not it issues the reconnect.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 5 of 7 files reviewed, 6 unresolved discussions (waiting on @faern and @raksooo)

___
*[mullvad-cli/src/cmds/connect.rs, line 36 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIcpmqWBq7Pzqx1UgXQ:-MIcpmqWBq7Pzqx1UgXR:b9omaua) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4e4ee10d80e90c3dc4f6ddfef2fd35b642fc4c0f/mullvad-cli/src/cmds/connect.rs#L36)):*
> ```Rust
>                     state::state_listen(&mut rpc, |state| match state {
>                         Connected(_) => false,
>                         Error(_) => false,
> ```

Maybe it would be good if `mullvad connect` exited with a non-zero exit code in case the daemon ends up in an error state? Would make it easier to detect issues when scripting using the CLI.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 5 of 7 files reviewed, 6 unresolved discussions (waiting on @faern and @raksooo)

___
*[mullvad-cli/src/state.rs, line 20 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_5RTy1ORTao9PdllO:-MIciQQS5EUOnkUIbC5u:bpwd0r9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L20)):*

The reason why I did this was to ensure that the we start listening for changes before running `connect`, but as long as `events_listen` is called before we should be fine I guess. I'll see what I can do!
</blockquote></details>

It seems like it's actually impossible to initiate a subscription without waiting for the events to arrive. This seems like a small inefficiency on the part of the GRPC library that we can't initiate a subscription without waiting to receive all of the messages. So I think to preserve the order of operations (listening for the state before connecting), it's best to leave the `tokio::spawn` as is.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 2 of 10 files reviewed, 2 unresolved discussions (waiting on @faern, @pinkisemils, and @raksooo)

___
*[mullvad-cli/src/state.rs, line 8 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_2p618VItIu4oa1pE:-MIdnjez30JY9S0OKizF:b-ckk1pw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L8)):*

Sounds like a good idea, I'll try it out!
</blockquote></details>

Do you still think a stream would be a better idea if we have to use `tokio::spawn` anyway?
___
*[mullvad-cli/src/cmds/connect.rs, line 36 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIcpmqWBq7Pzqx1UgXQ:-MIdnZOs4e-Vymx4Ck9i:bz5py0o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4e4ee10d80e90c3dc4f6ddfef2fd35b642fc4c0f/mullvad-cli/src/cmds/connect.rs#L36)):*

Maybe it would be good if `mullvad connect` exited with a non-zero exit code in case the daemon ends up in an error state? Would make it easier to detect issues when scripting using the CLI.
</blockquote></details>

Good idea! Done.
___
*[mullvad-cli/src/cmds/reconnect.rs, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIYvWSD8cSdz-FlaVwl:-MIdnduk7AnFYKoELsIl:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94f6e2ca7a7e492316f015da82c8140f04ae64f7/mullvad-cli/src/cmds/reconnect.rs#L42)):*

I'll try to make the daemon return whether or not it issues the reconnect.
</blockquote></details>

Done.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 2 of 10 files reviewed, 2 unresolved discussions (waiting on @pinkisemils and @raksooo)

___
*[mullvad-cli/src/state.rs, line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZB-UD-7VBQK_0BNKM:-MIdq20c5eqswEN3eJDP:b-fhicvf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L14)):*

There's one more `await` a few lines down, when assigning `events`. I moved the evaluation of the initial state to connect, disconnect and reconnect as you suggested. That still leaves me with `Result<JoinHandle>`.
</blockquote></details>

It's still an async function returning a future. How about moving the initial `rpc.events_listen().await` into the async block that you spawn and making the function not an async one? :thinking:

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 2 of 10 files reviewed, 2 unresolved discussions (waiting on @faern, @pinkisemils, and @raksooo)

___
*[mullvad-cli/src/state.rs, line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZB-UD-7VBQK_0BNKM:-MIdx44F9KoWrwcYugln:bx1xuee) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L14)):*

It's still an async function returning a future. How about moving the initial `rpc.events_listen().await` into the async block that you spawn and making the function not an async one? :thinking: 
</blockquote></details>

Then `rpc` is moved into the async block and can't be used for calling e.g. `rpc.connect_tunnel()` in `connect.rs`. I messed around a lot with this and the only other solution I've come up with is to use multiple rpc clients, but that doesn't feel like a good idea. Are there any other solutions?

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 2 of 10 files reviewed, 2 unresolved discussions (waiting on @faern, @pinkisemils, and @raksooo)

___
*[mullvad-cli/src/state.rs, line 8 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_2p618VItIu4oa1pE:-MIdt00kELCQvQKwmE1o:b-9vlc5y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L8)):*

Do you still think a stream would be a better idea if we have to use `tokio::spawn` anyway?
</blockquote></details>

I think the code might be a bit clearer. IMO, this  function could just spawn the daemont event subscriber and forward the events via a channel. The caller can filter out and do with them what they please.

pinkisemils (changes requested):
> Reviewed 7 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 9 of 10 files reviewed, 3 unresolved discussions (waiting on @faern and @raksooo)

___
*[mullvad-cli/src/state.rs, line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZB-UD-7VBQK_0BNKM:-MIs3Adl9s2494FZguv9:b5uk7y3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L14)):*

Then `rpc` is moved into the async block and can't be used for calling e.g. `rpc.connect_tunnel()` in `connect.rs`. I messed around a lot with this and the only other solution I've come up with is to use multiple rpc clients, but that doesn't feel like a good idea. Are there any other solutions?
</blockquote></details>

You could clone it.
___
*[mullvad-daemon/src/lib.rs, line 1892 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIs9oXI-BQ3vxeBpwcE:-MIs9oXI-BQ3vxeBpwcF:bezmnbo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5ec93e80b980c927db006d8fe0fc3ec22185064f/mullvad-daemon/src/lib.rs#L1892)):*
> ```Rust
>     /// progress towards that state.
>     /// Returns a bool representing whether or not a state change was initiated or an error if
>     /// trying to set secured state, but no account token is present.
> ```

This comment was wrong, but it still is wrong. The function doesn't check if an account token is set.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 9 of 10 files reviewed, 3 unresolved discussions (waiting on @pinkisemils and @raksooo)

___
*[mullvad-cli/src/state.rs, line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZB-UD-7VBQK_0BNKM:-MIsUp4k2TXTxdiwBTMR:byjoe50) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L14)):*

You could clone it.
</blockquote></details>

Is it cloneable? Does that not create multiple RPC connections?

Yes if we need to listen to event changes before we send the connect command, and gRPC does not allow us to issue two calls concurrently then we need two connections. But it really should allow concurrent calls though.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 9 of 10 files reviewed, 3 unresolved discussions (waiting on @faern and @raksooo)

___
*[mullvad-cli/src/state.rs, line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIZB-UD-7VBQK_0BNKM:-MIsXfUo6q_8j-odPZOS:bbojzwd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L14)):*

Is it cloneable? Does that not create multiple RPC connections?

Yes if we need to listen to event changes before we send the connect command, and gRPC does not allow us to issue two calls concurrently then we need two connections. But it really should allow concurrent calls though.
</blockquote></details>

It is cloneable.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 9 of 10 files reviewed, 4 unresolved discussions (waiting on @faern and @raksooo)

___
*[mullvad-cli/src/state.rs, line 22 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIt0j3lDzyUpyr_X4Mq:-MIt0j3p9YFlZoIlAaLg:b-rfltyv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5ec93e80b980c927db006d8fe0fc3ec22185064f/mullvad-cli/src/state.rs#L22)):*
> ```Rust
>                     match continue_condition(&new_state.state.unwrap()) {
>                         Ok(false) => break Ok(()),
>                         Err(e) => break Err(e),
> ```

Maybe it would be nicer to express the `continue_condition` closure signature as `Fn(&State) -> Option<Result<()>>`. `None` means continue and `Some(x)` means abort where `x` is simply the value we break on. Meaning `Some(Ok(()))` is "abort, everything went well" and `Some(Err(...))` "Abort, there was an unexpected error". Seems simpler to reason about, at least for my head. And this match would be simpler as well.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 5 of 10 files reviewed, 4 unresolved discussions (waiting on @faern, @pinkisemils, and @raksooo)

___
*[mullvad-cli/src/state.rs, line 8 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_2p618VItIu4oa1pE:-MIt2fBQBPAImjfO2vSh:b9robwn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L8)):*

I think the code might be a bit clearer. IMO, this  function could just spawn the daemont event subscriber and forward the events via a channel. The caller can filter out and do with them what they please.
</blockquote></details>

I tried this out. What do you think? I might have to update it to send errors through the channel as well.
___
*[mullvad-cli/src/state.rs, line 22 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIt0j3lDzyUpyr_X4Mq:-MIt2sPy7Ju_Y6l_IptM:b-we9lu1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5ec93e80b980c927db006d8fe0fc3ec22185064f/mullvad-cli/src/state.rs#L22)):*

Maybe it would be nicer to express the `continue_condition` closure signature as `Fn(&State) -> Option<Result<()>>`. `None` means continue and `Some(x)` means abort where `x` is simply the value we break on. Meaning `Some(Ok(()))` is "abort, everything went well" and `Some(Err(...))` "Abort, there was an unexpected error". Seems simpler to reason about, at least for my head. And this match would be simpler as well.
</blockquote></details>

I switched to using a channel instead as Emils suggested. If we agree that that is a good solution this might not be needed.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 5 of 10 files reviewed, 4 unresolved discussions (waiting on @faern, @pinkisemils, and @raksooo)

___
*[mullvad-daemon/src/lib.rs, line 1892 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIs9oXI-BQ3vxeBpwcE:-MIt4MDL5-8u8_RyexfY:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5ec93e80b980c927db006d8fe0fc3ec22185064f/mullvad-daemon/src/lib.rs#L1892)):*

This comment was wrong, but it still is wrong. The function doesn't check if an account token is set.
</blockquote></details>

Done.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 5 of 10 files reviewed, 6 unresolved discussions (waiting on @faern, @pinkisemils, and @raksooo)

___
*[mullvad-cli/src/cmds/connect.rs, line 37 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIt9-av6nuusPNZzcl6:-MIt9-aw-KG9cvcXWwvc:bmorx9u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ff7492569945a641bb1e557f6d2be23a096ffd8/mullvad-cli/src/cmds/connect.rs#L37)):*
> ```Rust
>                 while let Some(state) = receiver.next().await {
>                     format::print_state(&state);
>                     match state.state.unwrap() {
> ```

Will the state always exist?
___
*[mullvad-cli/src/cmds/connect.rs, line 39 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIt96Ta7ig01JUe9qRt:-MIt96Ta7ig01JUe9qRu:baalgpz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ff7492569945a641bb1e557f6d2be23a096ffd8/mullvad-cli/src/cmds/connect.rs#L39)):*
> ```Rust
>                     match state.state.unwrap() {
>                         State::Connected(_) => break,
>                         State::Error(_) => Err(Error::CommandFailed("connect"))?,
> ```

I'm not sure how idiomatic `Err(_)?` is considered. I think `return Err(_)` is more clear.

pinkisemils (changes requested):
> Reviewed 4 of 5 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 9 of 10 files reviewed, 6 unresolved discussions (waiting on @faern, @pinkisemils, and @raksooo)

___
*[mullvad-cli/src/state.rs, line 14 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MItT5Wn2uGn60osAuG8:-MItT5Wn2uGn60osAuG9:b-16odk5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ff7492569945a641bb1e557f6d2be23a096ffd8/mullvad-cli/src/state.rs#L14)):*
> ```Rust
> pub fn state_listen(mut rpc: ManagementServiceClient) -> Receiver<TunnelState> {
>     let (mut sender, receiver) = mpsc::channel::<TunnelState>(100);
>     tokio::spawn(async move {
> ```

As it stands, this function depends on being executed from within a `tokio` runtime, otherwise it will panic. This might be a good thing to document.
___
*[mullvad-cli/src/cmds/disconnect.rs, line 41 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MItf0Tv3TfyAreeV0_t:-MItf0Tv3TfyAreeV0_u:b-mej9qm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ff7492569945a641bb1e557f6d2be23a096ffd8/mullvad-cli/src/cmds/disconnect.rs#L41)):*
> ```Rust
>                         _ => {}
>                     }
>                 }
> ```

I think that it's an error to not receive the event the CLI's waiting for.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 9 of 10 files reviewed, 6 unresolved discussions (waiting on @faern, @pinkisemils, and @raksooo)

___
*[mullvad-cli/src/state.rs, line 8 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_2p618VItIu4oa1pE:-MIth7x77Us2igPA0B-4:b-1qom05) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L8)):*

I tried this out. What do you think? I might have to update it to send errors through the channel as well.
</blockquote></details>

It depends - if the reason the subscription failed is relevant to us, then yes. Otherwise, we can just infer that some kind of an error happened because the subscription was closed (`events.next().await` returns `None`).

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 9 of 10 files reviewed, 7 unresolved discussions (waiting on @faern, @pinkisemils, and @raksooo)

___
*[mullvad-cli/src/state.rs, line 18 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIxAaaj3HBkDkvPgE6D:-MIxAaakB5Lv-EJnWHy-:b-hlasf0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ff7492569945a641bb1e557f6d2be23a096ffd8/mullvad-cli/src/state.rs#L18)):*
> ```Rust
>         while let Ok(Some(event)) = events.message().await {
>             if let EventType::TunnelState(new_state) = event.event.unwrap() {
>                 sender.send(new_state).await.unwrap();
> ```

Can you really `.unwrap()` the send? That will panic once the receiver is dropped, no? Maybe we don't actually drop it until program termination in practice. But it would be way nicer if this task detected when the receiver was no longer interested in the events and then aborted the loop cleanly.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 5 of 11 files reviewed, 4 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-cli/src/state.rs, line 8 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_2p618VItIu4oa1pE:-MJ3LCb4BDNAmiNc9Kt1:b-brz86w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L8)):*

It depends - if the reason the subscription failed is relevant to us, then yes. Otherwise, we can just infer that some kind of an error happened because the subscription was closed (`events.next().await` returns `None`).
</blockquote></details>

I went with sending a result through the channel. It feels like a good idea to be able to forward Grpc errors even though it complicates the code a bit.
___
*[mullvad-cli/src/state.rs, line 14 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MItT5Wn2uGn60osAuG8:-MJ3KoCbB3MvA_3tXzng:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ff7492569945a641bb1e557f6d2be23a096ffd8/mullvad-cli/src/state.rs#L14)):*

As it stands, this function depends on being executed from within a `tokio` runtime, otherwise it will panic. This might be a good thing to document.
</blockquote></details>

Done.
___
*[mullvad-cli/src/state.rs, line 18 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIxAaaj3HBkDkvPgE6D:-MJ3KpjG3bAcnA1DnKbF:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ff7492569945a641bb1e557f6d2be23a096ffd8/mullvad-cli/src/state.rs#L18)):*

Can you really `.unwrap()` the send? That will panic once the receiver is dropped, no? Maybe we don't actually drop it until program termination in practice. But it would be way nicer if this task detected when the receiver was no longer interested in the events and then aborted the loop cleanly.
</blockquote></details>

Done.
___
*[mullvad-cli/src/cmds/connect.rs, line 37 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIt9-av6nuusPNZzcl6:-MJ3Ks4T6jFKP_Kt_0n5:brasx8n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ff7492569945a641bb1e557f6d2be23a096ffd8/mullvad-cli/src/cmds/connect.rs#L37)):*

Will the state always exist?
</blockquote></details>

I just copied this from `mullvad-cli/src/cmds/status.rs` and assume it's correct.
___
*[mullvad-cli/src/cmds/connect.rs, line 39 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIt96Ta7ig01JUe9qRt:-MJ3L5c9CmiiYBhotpQ0:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ff7492569945a641bb1e557f6d2be23a096ffd8/mullvad-cli/src/cmds/connect.rs#L39)):*

I'm not sure how idiomatic `Err(_)?` is considered. I think `return Err(_)` is more clear.
</blockquote></details>

Done.
___
*[mullvad-cli/src/cmds/disconnect.rs, line 41 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MItf0Tv3TfyAreeV0_t:-MJ3L7Qq8j3yVaAVkUYG:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ff7492569945a641bb1e557f6d2be23a096ffd8/mullvad-cli/src/cmds/disconnect.rs#L41)):*

I think that it's an error to not receive the event the CLI's waiting for. 
</blockquote></details>

Done.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 5 of 11 files reviewed, 4 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-cli/src/state.rs, line 8 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_2p618VItIu4oa1pE:-MJ5vJ8e6k_sfeWrrzVj:b-lu2f2j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L8)):*

I went with sending a result through the channel. It feels like a good idea to be able to forward Grpc errors even though it complicates the code a bit.
</blockquote></details>

This code became a lot more complicated and I'm not sure if I prefer this over the solution without the channel. What do you think? Is there any way to simplify the current code?

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 5 of 11 files reviewed, 4 unresolved discussions (waiting on @faern, @pinkisemils, and @raksooo)

___
*[mullvad-cli/src/state.rs, line 23 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MJ73lux3Ows7jPPVrd4:-MJ73lux3Ows7jPPVrd5:bmk8oxk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/90d8676f49a280d3372b7053e142f13e2c6cd950/mullvad-cli/src/state.rs#L23)):*
> ```Rust
>                         Ok(Some(event)) => match event.event.unwrap() {
>                             EventType::TunnelState(new_state) => Some(Ok(new_state)),
>                             _ => None,
> ```

Nit. But maybe this could be `continue` and then the type of the variable that is currently called `forward` could be simplified to `Result<_, _>`? The `None` value would not exist.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 5 of 11 files reviewed, 3 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-cli/src/state.rs, line 23 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MJ73lux3Ows7jPPVrd4:-MJ7HPOpDLDpInc-eblG:bz5py0o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/90d8676f49a280d3372b7053e142f13e2c6cd950/mullvad-cli/src/state.rs#L23)):*

Nit. But maybe this could be `continue` and then the type of the variable that is currently called `forward` could be simplified to `Result<_, _>`? The `None` value would not exist.
</blockquote></details>

Good idea! Done.

faern (approved):
> Reviewed 5 of 6 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: all files reviewed, 3 unresolved discussions (waiting on @pinkisemils and @raksooo)

___
*[mullvad-cli/src/cmds/connect.rs, line 36 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MJ7KF5oD3jLBcPszEsh:-MJ7KF5pA9T_fx_wr-pU:bj9r6jp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/90d8676f49a280d3372b7053e142f13e2c6cd950/mullvad-cli/src/cmds/connect.rs#L36)):*
> ```Rust
>             if let Some(mut receiver) = receiver_option {
>                 loop {
>                     if let Some(state) = receiver.next().await {
> ```

Can't the `loop {} + if let Some(state)` be simplified as `while let Some(state)`? The error handling would be slightly different, so maybe not worth it, not sure. But it's an idea.

pinkisemils (approved):
> Reviewed 1 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[mullvad-cli/src/state.rs, line 8 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_2p618VItIu4oa1pE:-MJ82IKXDsIXXrv4vYXo:bgkp53z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L8)):*

This code became a lot more complicated and I'm not sure if I prefer this over the solution without the channel. What do you think? Is there any way to simplify the current code?
</blockquote></details>

The code could be simplified a bit. I'd argue that even if you were to change this function back to it's initial state where it took two closures and returned only when the correct/incorrect state is received, the function would be as complex as it is now, given that you'd have to filter our the failures of the GRPC subscriber, end of the stream and the tunnel state changes anyway.  But if you feel like that would improve readability, go ahead and change it :)
___
*[mullvad-cli/src/cmds/connect.rs, line 37 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MIt9-av6nuusPNZzcl6:-MJ3qyQCC9NMMVwJFqOh:bdggu8k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ff7492569945a641bb1e557f6d2be23a096ffd8/mullvad-cli/src/cmds/connect.rs#L37)):*

I just copied this from `mullvad-cli/src/cmds/status.rs` and assume it's correct.
</blockquote></details>

The `unwrap()` is there because the type is auto-generated from our proto files and in Protobuf, all fields are optional.

pinkisemils (approved):
> Reviewed 1 of 6 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: all files reviewed, 2 unresolved discussions (waiting on @faern and @raksooo)

___
*[mullvad-cli/src/cmds/connect.rs, line 36 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MJ7KF5oD3jLBcPszEsh:-MJ85oaR6bylaxG-USpW:b-qhxqai) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/90d8676f49a280d3372b7053e142f13e2c6cd950/mullvad-cli/src/cmds/connect.rs#L36)):*

Can't the `loop {} + if let Some(state)` be simplified as `while let Some(state)`? The error handling would be slightly different, so maybe not worth it, not sure. But it's an idea.
</blockquote></details>

I agree, and this could be extended to other functions here as well.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[mullvad-cli/src/state.rs, line 8 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_2p618VItIu4oa1pE:-MJBG5FY9UC331cAj1z9:b-fwt7m6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L8)):*

The code could be simplified a bit. I'd argue that even if you were to change this function back to it's initial state where it took two closures and returned only when the correct/incorrect state is received, the function would be as complex as it is now, given that you'd have to filter our the failures of the GRPC subscriber, end of the stream and the tunnel state changes anyway.  But if you feel like that would improve readability, go ahead and change it :)
</blockquote></details>

Sorry for being unclear, I was referring to revision 4 which only has the `continue_callback`. That revision could of course be improved with some of the things I've done later such as moving the `rpc.events_listen` into the tokio task, and letting the daemon provide whether or not a state change was issued. But I like this solution as well. It's probably not worth the time to change approach again since this works well :+1:
___
*[mullvad-cli/src/cmds/connect.rs, line 36 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MJ7KF5oD3jLBcPszEsh:-MJBGmmz4e9SXdThZqj3:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/90d8676f49a280d3372b7053e142f13e2c6cd950/mullvad-cli/src/cmds/connect.rs#L36)):*

I agree, and this could be extended to other functions here as well.
</blockquote></details>

Done.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: 8 of 11 files reviewed, all discussions resolved (waiting on @faern)

___
*[mullvad-cli/src/state.rs, line 8 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-MI_2p618VItIu4oa1pE:-MJC0cCuE0Gs4SzIsZvq:bosyu0d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ef584b6c0c9b8ff5558eccbd4c58a000318cd28/mullvad-cli/src/state.rs#L8)):*

Sorry for being unclear, I was referring to revision 4 which only has the `continue_callback`. That revision could of course be improved with some of the things I've done later such as moving the `rpc.events_listen` into the tokio task, and letting the daemon provide whether or not a state change was issued. But I like this solution as well. It's probably not worth the time to change approach again since this works well :+1:
</blockquote></details>

I like this version. Short and readable. No strange input arguments.

faern (approved):
> Reviewed 3 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
:lgtm:

> Reviewed 3 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156#-:-MJC6OEeDN5rAFOy9GCw:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> Reviewed 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2156)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2162, opened 2020-10-03
Refactor payment buttons on Android

jvff:
The Account settings screen, the Welcome screen and the Out Of Time screen all have payment buttons. The code to handle them is the same, but repeated for each screen. Instead of adding repeated code for a new in-app purchase button, this PR streamlines the payment buttons to group shared code and reduce repeated code.

Two new widgets were created, one is the `RedeemVoucherButton` and the other is the `SitePaymentButton`. The former has the common code to open the Redeem Voucher dialog, and the second isn't very useful yet, since it mostly just configures the URL. However, this will make it easier to add the in-app purchase button later, since there will be different labels for the site payment button based on the product flavor.

The `SitePaymentButton` also has a small quirk. It has a `newAccount` property that should be set to configure the label, so that it's either "Buy credit" or "Buy more credit".

The layout code for the buttons is now in a separate XML layout file which is included by the layouts for the screens that use them. However, this did require some margin and padding changes so that the UI positions wouldn't change and so that common margin parameters could be used with the shared buttons.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2162)


pinkisemils (approved):
> Reviewed 11 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2162)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2164, opened 2020-10-03
Quit when notification is dismissed

jvff:
There have been a few reports of the app starting by itself (and connecting to the tunnel) when it's not supposed to (for example, after disconnecting while having the auto-connect setting enabled). The cause is still unknown, but this PR is an attempt to prevent this from happening by stopping the background service when the notification is dismissed. By stopping the service, there should be no way for it to start unless the user opens the app again.

Dismissing the notification is only possible if the notification is not on the foreground and the screen is unlocked, which means that it can only be dismissed while disconnected, the screen is unlocked and the tunnel is disconnected.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2164)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ Line wrap the file at 100 chars.                                              Th
 - Fix crash if connection to service is lost while opening the Split Tunneling settings screen.
 - Fix rare crash that could occur when the tunnel state changes when showing or hiding the quick
   settings tile.
+- Fix app starting by itself sometimes.
 
 #### Linux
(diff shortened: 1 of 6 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2164)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2164)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2168, opened 2020-10-06
Add persistent firewall filters on Windows

dlon:
These changes cause the WinFw module to add persistent WFP filters when deinitialized with the "continue blocking" cleanup policy, and the daemon exits in a blocking state. This happens if "Always require VPN" is enabled, or if auto-connect is enabled.

Two sets of filters are required: Boot-time filters are used for blocking traffic before the BFE service has started. Persistent filters are used to block traffic after the BFE service has started, but before mullvad-daemon has started.

This prevents traffic leaks during boot when either "Always require VPN" or auto-connect is enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,11 @@ Line wrap the file at 100 chars.                                              Th
   settings tile.
 - Fix app starting by itself sometimes.
 
+### Security
+#### Windows
+- Block all traffic received or sent before the BFE service and daemon service have started during
+  boot, if "Always require VPN" or auto-connect is enabled.
+
 
 ## [2020.6-beta3] - 2020-10-06
 This release is for desktop only.
--- a/dist-assets/windows/installer.nsh
+++ b/dist-assets/windows/installer.nsh
@@ -40,6 +40,7 @@
 !define INSTALL_REGISTRY_KEY "Software\${PRODUCT_NAME}"
 
 !define BLOCK_OUTBOUND_IPV4_FILTER_GUID "{a81c5411-0fd0-43a9-a9be-313f299de64f}"
+!define PERSISTENT_BLOCK_OUTBOUND_IPV4_FILTER_GUID "{79860c64-9a5e-48a3-b5f3-d64b41659aa5}"
 
 #
 # ExtractTapDriver
@@ -645,6 +646,9 @@
 #
 !macro customInstall
 
+	Var /GLOBAL BlockFilterResult
+	Var /GLOBAL PersistentBlockFilterResult
+
 	Push $R0
 
 	log::Initialize ${LOG_FILE}
@@ -703,10 +707,15 @@
 	Delete "$INSTDIR\mullvad vpn.exe"
 
 	nsExec::ExecToStack '"$SYSDIR\netsh.exe" wfp show security FILTER ${BLOCK_OUTBOUND_IPV4_FILTER_GUID}'
-	Pop $0
+	Pop $BlockFilterResult
 	Pop $1
 
-	${If} $0 == 0
+	nsExec::ExecToStack '"$SYSDIR\netsh.exe" wfp show security FILTER ${PERSISTENT_BLOCK_OUTBOUND_IPV4_FILTER_GUID}'
+	Pop $PersistentBlockFilterResult
+	Pop $1
+
+	${If} $BlockFilterResult == 0
+	${OrIf} $PersistentBlockFilterResult == 0
 		MessageBox MB_ICONEXCLAMATION|MB_YESNO "Do you wish to unblock your internet access? Doing so will leave you with an unsecure connection." IDNO customInstall_abortInstallation_skip_firewall_revert
 		${ClearFirewallRules}
 	${EndIf}
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -640,6 +640,13 @@ where
             }
         }
 
+        // If auto-connect is enabled, block all traffic before shutting down to ensure
+        // that no traffic can leak during boot.
+        #[cfg(windows)]
+        if self.settings.auto_connect {
+            self.send_tunnel_command(TunnelCommand::BlockWhenDisconnected(true));
+        }
+
         self.finalize().await;
         Ok(())
     }
--- a/talpid-core/src/firewall/windows.rs
+++ b/talpid-core/src/firewall/windows.rs
@@ -319,11 +319,11 @@ mod winfw {
     }
 
     #[allow(dead_code)]
-    #[repr(u8)]
+    #[repr(u32)]
     #[derive(Clone, Copy)]
     pub enum WinFwCleanupPolicy {
-        ContinueBlocking = 0u8,
-        ResetFirewall = 1u8,
+        ContinueBlocking = 0,
+        ResetFirewall = 1,
     }
 
     ffi_error!(InitializationResult, Error::Initialization);
--- a/windows/winfw/src/winfw/mullvadguids.cpp
+++ b/windows/winfw/src/winfw/mullvadguids.cpp
@@ -100,7 +100,7 @@ MullvadGuids::DetailedIdentityRegistry MullvadGuids::DetailedRegistry(IdentityQu
 {
 	std::multimap<WfpObjectType, GUID> registry;
 
-	if (IdentityQualifier::IncludeDeprecated == qualifier)
+	if (IdentityQualifier::IncludeDeprecated == (qualifier & IdentityQualifier::IncludeDeprecated))
 	{
 		registry = DeprecatedIdentities();
 	}
@@ -147,6 +147,22 @@ MullvadGuids::DetailedIdentityRegistry MullvadGuids::DetailedRegistry(IdentityQu
 	registry.insert(std::make_pair(WfpObjectType::Filter, Filter_Dns_PermitTunnel_Outbound_Ipv4()));
 	registry.insert(std::make_pair(WfpObjectType::Filter, Filter_Dns_PermitTunnel_Outbound_Ipv6()));
 
+	if (IdentityQualifier::IncludePersistent == (qualifier & IdentityQualifier::IncludePersistent))
+	{
+		registry.insert(std::make_pair(WfpObjectType::Provider, ProviderPersistent()));
+		registry.insert(std::make_pair(WfpObjectType::Sublayer, SublayerPersistent()));
+
+		registry.insert(std::make_pair(WfpObjectType::Filter, Filter_Boottime_BlockAll_Inbound_Ipv4()));
+		registry.insert(std::make_pair(WfpObjectType::Filter, Filter_Boottime_BlockAll_Outbound_Ipv4()));
+		registry.insert(std::make_pair(WfpObjectType::Filter, Filter_Boottime_BlockAll_Inbound_Ipv6()));
+		registry.insert(std::make_pair(WfpObjectType::Filter, Filter_Boottime_BlockAll_Outbound_Ipv6()));
+
+		registry.insert(std::make_pair(WfpObjectType::Filter, Filter_Persistent_BlockAll_Inbound_Ipv4()));
+		registry.insert(std::make_pair(WfpObjectType::Filter, Filter_Persistent_BlockAll_Outbound_Ipv4()));
+		registry.insert(std::make_pair(WfpObjectType::Filter, Filter_Persistent_BlockAll_Inbound_Ipv6()));
+		registry.insert(std::make_pair(WfpObjectType::Filter, Filter_Persistent_BlockAll_Outbound_Ipv6()));
+	}
+
 	return registry;
 }
 
@@ -159,6 +175,20 @@ const GUID &MullvadGuids::Provider()
 		0xb9db,
 		0x43c0,
 		{ 0xb3, 0x43, 0xeb, 0x93, 0x65, 0xc7, 0xbd, 0xd2 }
+	};
+
+	return g;
+}
+
+//static
+const GUID &MullvadGuids::ProviderPersistent()
+{
+	static const GUID g =
+	{
+		0x2bc5bc63,
+		0x80b0,
+		0x4119,
+		{ 0x86, 0xd3, 0x6a, 0xfe, 0x0d, 0xff, 0x2a, 0x26 }
 	};
 
 	return g;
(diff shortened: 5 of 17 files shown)


mvd-ows (changes requested):
> Reviewed 25 of 25 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[dist-assets/windows/installer.nsh, line 719 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ0vzJ34CXVLwLkFu7y:-MJ0vzJ34CXVLwLkFu7z:b-b8y2jf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81ec0d8d8299542e1f4432424ee12a47db41d3af/dist-assets/windows/installer.nsh#L719)):*
> ```NSIS
> 	${If} $BlockFilterResult == 0
> 	${OrIf} $PersistentBlockFilterResult == 0
> 		MessageBox MB_ICONEXCLAMATION|MB_YESNO "Do you wish to unblock your internet access? Doing so will leave you with an unsecure connection." IDNO customInstall_abortInstallation_skip_firewall_revert
> ```

Maybe use a capital `I` on `Internet`? Different sources have different opinions on this. Not sure what we usually do.
___
*[windows/winfw/src/winfw/mullvadguids.cpp, line 103 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ1CDQd3x5rHGZIITdB:-MJ1CDQd3x5rHGZIITdC:bkdfdw8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81ec0d8d8299542e1f4432424ee12a47db41d3af/windows/winfw/src/winfw/mullvadguids.cpp#L103)):*
> ```cpp
> 	std::multimap<WfpObjectType, GUID> registry;
> 
> 	if (IdentityQualifier::OnlyCurrent != (qualifier & IdentityQualifier::IncludeDeprecated))
> ```

This is quite difficult to read. It depends both on the and-overload as well as the exact definition of `OnlyCurrent`.

Is it possible to use a `flag GetFlag(composite, flag)` or `bool CheckFlag(composite, flag), but using a function name more suited to the context?
___
*[windows/winfw/src/winfw/mullvadobjects.cpp, line 69 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ1DrK_2DQOdWbM-Bv1:-MJ1DrK_2DQOdWbM-Bv2:b-w5a25w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81ec0d8d8299542e1f4432424ee12a47db41d3af/windows/winfw/src/winfw/mullvadobjects.cpp#L69)):*
> ```cpp
> 	(*builder)
> 		.name(L"Mullvad VPN filters that survive across BFE stop/start")
> 		.description(L"Filters that restrict traffic before the has daemon started")
> ```

Grammar.

mvd-ows (commented):
> Reviewed 4 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: 13 of 17 files reviewed, 3 unresolved discussions (waiting on @mvd-ows)

___
*[dist-assets/windows/installer.nsh, line 719 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ0vzJ34CXVLwLkFu7y:-MJ2IIveDTxkUk4jktvh:b-t7ivg5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81ec0d8d8299542e1f4432424ee12a47db41d3af/dist-assets/windows/installer.nsh#L719)):*

Maybe use a capital `I` on `Internet`? Different sources have different opinions on this. Not sure what we usually do.
</blockquote></details>

I checked this with Kat, so it should be good.
___
*[windows/winfw/src/winfw/mullvadguids.cpp, line 103 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ1CDQd3x5rHGZIITdB:-MJ2I6gVBaQrVls3u6gO:bbdob0p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81ec0d8d8299542e1f4432424ee12a47db41d3af/windows/winfw/src/winfw/mullvadguids.cpp#L103)):*

This is quite difficult to read. It depends both on the and-overload as well as the exact definition of `OnlyCurrent`.

Is it possible to use a `flag GetFlag(composite, flag)` or `bool CheckFlag(composite, flag), but using a function name more suited to the context?
</blockquote></details>

I've simplified it.
___
*[windows/winfw/src/winfw/mullvadobjects.cpp, line 69 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ1DrK_2DQOdWbM-Bv1:-MJ2ICkM9p57N4pSkJM3:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81ec0d8d8299542e1f4432424ee12a47db41d3af/windows/winfw/src/winfw/mullvadobjects.cpp#L69)):*

Grammar.
</blockquote></details>

Done.

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[mullvad-daemon/src/lib.rs, line 644 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ2KClqDFPdmsf3OBe1:-MJ2KClqDFPdmsf3OBe2:bl7lwt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/635fb3855265ff00fb1f4ef2698b58fc56f0514e/mullvad-daemon/src/lib.rs#L644)):*
> ```Rust
> 
>         #[cfg(windows)]
>         if self.settings.auto_connect {
> ```

Can we get a comment here? Explaining that on daemon shut down we block the firewall if `auto-connect` is enabled, in order to make it properly block all traffic during boot until the daemon can take over again.
___
*[talpid-core/src/firewall/windows.rs, line 322 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ2LvVGC1bQ_8S7cTTD:-MJ2LvVGC1bQ_8S7cTTE:bjmg523) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/635fb3855265ff00fb1f4ef2698b58fc56f0514e/talpid-core/src/firewall/windows.rs#L322)):*
> ```Rust
> 
>     #[allow(dead_code)]
>     #[repr(u32)]
> ```

Is this representation currently wrong on master or does this PR also change the C++ version? I only see the `WINFW_CLEANUP_POLICY` enum, but it does not specify type as far as I can see :thinking:

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: 15 of 17 files reviewed, 2 unresolved discussions (waiting on @faern and @mvd-ows)

___
*[mullvad-daemon/src/lib.rs, line 644 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ2KClqDFPdmsf3OBe1:-MJ2RfRJ8Xd_lz0FeJQt:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/635fb3855265ff00fb1f4ef2698b58fc56f0514e/mullvad-daemon/src/lib.rs#L644)):*

Can we get a comment here? Explaining that on daemon shut down we block the firewall if `auto-connect` is enabled, in order to make it properly block all traffic during boot until the daemon can take over again.
</blockquote></details>

Done.
___
*[talpid-core/src/firewall/windows.rs, line 322 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ2LvVGC1bQ_8S7cTTD:-MJ2RjOf9o5-eoSey-ty:byfiw39) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/635fb3855265ff00fb1f4ef2698b58fc56f0514e/talpid-core/src/firewall/windows.rs#L322)):*

Is this representation currently wrong on master or does this PR also change the C++ version? I only see the `WINFW_CLEANUP_POLICY` enum, but it does not specify type as far as I can see :thinking: 
</blockquote></details>

It is wrong on master. The underlying type here will be [`int`](https://en.cppreference.com/w/cpp/types/underlying_type), since none is specified, i.e. 32 bits in this case. I've changed it to `uint32_t` for clarity now (and because `int` is not fixed).

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: 15 of 17 files reviewed, 2 unresolved discussions (waiting on @dlon, @faern, and @mvd-ows)

___
*[talpid-core/src/firewall/windows.rs, line 322 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ2LvVGC1bQ_8S7cTTD:-MJ6JGGc0wK8aSUguOiZ:b-grk8if) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/635fb3855265ff00fb1f4ef2698b58fc56f0514e/talpid-core/src/firewall/windows.rs#L322)):*

It is wrong on master. The underlying type here will be [`int`](https://en.cppreference.com/w/cpp/types/underlying_type), since none is specified, i.e. 32 bits in this case. I've changed it to `uint32_t` for clarity now (and because `int` is not fixed).
</blockquote></details>

:O So the ABI for the function wall was wrong then. How did this not cause issues? If it's guaranteed to not cause call issues on current `master` this is fine to just change without much fanfare. But otherwise I would like this to be a `### Fixed` changelog entry maybe?
Do we have any idea if this could have caused issues?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: 15 of 17 files reviewed, 2 unresolved discussions (waiting on @faern and @mvd-ows)

___
*[talpid-core/src/firewall/windows.rs, line 322 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ2LvVGC1bQ_8S7cTTD:-MJ6hLOn1IcVJYpvYVLE:bvf8m8i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/635fb3855265ff00fb1f4ef2698b58fc56f0514e/talpid-core/src/firewall/windows.rs#L322)):*

:O So the ABI for the function wall was wrong then. How did this not cause issues? If it's guaranteed to not cause call issues on current `master` this is fine to just change without much fanfare. But otherwise I would like this to be a `### Fixed` changelog entry maybe?
Do we have any idea if this could have caused issues?
</blockquote></details>

I investigated this. The short answer is that it was not a problem in practice.

Here is a small C program containing a function with a signature equivalent to that of `WinFw_Deinitialize`:

```
extern "C"
__declspec(dllexport)
__declspec(noinline)
void __stdcall testfunc(int a)
{
	std::cout << a << std::endl;
}

typedef void(*wrongfuncsig)(unsigned char);
```

A normal call to the function `testfunc` in C produces the following code:

```
	testfunc(1);
mov         ecx,1
call        testfunc (07FF6B2D31000h) 
```

And using a function pointer (with the incorrect signature):

```
	wrongfuncsig func = (wrongfuncsig)testfunc;
	func(1);
mov         cl,1  
call        testfunc (07FF6B2D31000h)  
```

As you can tell, only the lowest fourth of the `ecx` register is set when using the wrong signature. I could show this on the Rust side:

```
#![feature(asm)]

extern "system" {
    #[link_name = "testfunc"]
    pub fn testfunc(a: u8);
}

fn main() {
    unsafe {
        asm!("mov ecx, 0xffffffff");
        testfunc(1);
    }
}
```

This outputs `-255` (i.e., `ffffff01`).

However, calling `testfunc` with the argument 0 works as expected. The assembly code reveals why:

```
	movl	$4294967295, %ecx
	xorl	%ecx, %ecx
	popq	%rax
	jmp	testfunc
```

When `testfunc(0)` is called, the entire `ecx` register is zeroed.

Conclusion:

`WinFw_Deinitialize(WINFW_CLEANUP_POLICY_RESET_FIREWALL)` on master is not likely to produce the right value for `cleanupPolicy`.

However, the code only ever passed the argument `WINFW_CLEANUP_POLICY_CONTINUE_BLOCKING`, whose value is 0, to the function. This zeroes the entire registry that represents the argument.

So in practice, this bug was never an issue.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: 15 of 17 files reviewed, all discussions resolved (waiting on @mvd-ows)

___
*[talpid-core/src/firewall/windows.rs, line 322 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ2LvVGC1bQ_8S7cTTD:-MJ71zmZBywDqJwrCFN_:ba39s9m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/635fb3855265ff00fb1f4ef2698b58fc56f0514e/talpid-core/src/firewall/windows.rs#L322)):*

I investigated this. The short answer is that it was not a problem in practice.

Here is a small C program containing a function with a signature equivalent to that of `WinFw_Deinitialize`:

```
extern "C"
__declspec(dllexport)
__declspec(noinline)
void __stdcall testfunc(int a)
{
	std::cout << a << std::endl;
}

typedef void(*wrongfuncsig)(unsigned char);
```

A normal call to the function `testfunc` in C produces the following code:

```
	testfunc(1);
mov         ecx,1
call        testfunc (07FF6B2D31000h) 
```

And using a function pointer (with the incorrect signature):

```
	wrongfuncsig func = (wrongfuncsig)testfunc;
	func(1);
mov         cl,1  
call        testfunc (07FF6B2D31000h)  
```

As you can tell, only the lowest fourth of the `ecx` register is set when using the wrong signature. I could show this on the Rust side:

```
#![feature(asm)]

extern "system" {
    #[link_name = "testfunc"]
    pub fn testfunc(a: u8);
}

fn main() {
    unsafe {
        asm!("mov ecx, 0xffffffff");
        testfunc(1);
    }
}
```

This outputs `-255` (i.e., `ffffff01`).

However, calling `testfunc` with the argument 0 works as expected. The assembly code reveals why:

```
	movl	$4294967295, %ecx
	xorl	%ecx, %ecx
	popq	%rax
	jmp	testfunc
```

When `testfunc(0)` is called, the entire `ecx` register is zeroed.

Conclusion:

`WinFw_Deinitialize(WINFW_CLEANUP_POLICY_RESET_FIREWALL)` on master is not likely to produce the right value for `cleanupPolicy`.

However, the code only ever passed the argument `WINFW_CLEANUP_POLICY_CONTINUE_BLOCKING`, whose value is 0, to the function. This zeroes the entire registry that represented the argument.

So in practice, this bug was never an issue.
</blockquote></details>

Great research and explanation! So all current releases are safe in practice then. Good. Great it's fixed however!

Any way we could/should adopt our FFI or Rust/C++ guidelines to prevent such issues in the future? We should probably always explicitly state the integer representation type on enums on both sides as a first step.

faern (commented):
> Reviewed 1 of 12 files at r2, 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: 16 of 17 files reviewed, all discussions resolved (waiting on @mvd-ows)

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon:
> 
> 
> > _[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)_ status: 15 of 17 files reviewed, all discussions resolved (waiting on @mvd-ows)
> 
> _[talpid-core/src/firewall/windows.rs, line 322 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168#-MJ2LvVGC1bQ_8S7cTTD:-MJ71zmZBywDqJwrCFN_:ba39s9m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/635fb3855265ff00fb1f4ef2698b58fc56f0514e/talpid-core/src/firewall/windows.rs#L322)):_
> Previously, dlon (David Lönnhager) wrote…
> 
> Great research and explanation! So all current releases are safe in practice then. Good. Great it's fixed however!
> 
> Any way we could/should adopt our FFI or Rust/C++ guidelines to prevent such issues in the future? We should probably always explicitly state the integer representation type on enums on both sides as a first step.

Reviewable broke for me. Anyway, that seems like a good idea.

dlon (commented):
> Reviewed 10 of 25 files at r1, 1 of 12 files at r2, 4 of 4 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2168)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2177, opened 2020-10-09
Improve keyboard navigation on Android

jvff:
Some Android devices do not have touch screens (like TVs). On these devices, the app experience was subpar. This PR attempts to improve that. While it doesn't make it great (a second pass to tweak the UI colors, for example, is still needed later), it does make it usable for most cases.

There are different aspects that this PR touches. On some screens the only changes need is to make a few views focusable. In some cases we need to configure custom widgets so that they are marked as focusable. The custom buttons also needed to have their background drawables changed so that they used a different color to show that they were focused.

The first larger refactor was on the Select Location screen. The location row views had to be redesigned. Instead of setting a custom background color like other views, it overlays the views with a transparent clickable view. This overlay view has the default selectable background from Android. This means that it has the "ripple effect" when clicked, and that it is highlighted when focused. The chevron already used that background, so it didn't need any changes.

It's probably noteworthy that I attempted to use a flat hierarchy first, replacing the `LinearLayout` with a single `RelativeLayout`. However, there seems to be some issue that causes the chevron `ImageButton` widget to not fully match the parent's height in that configuration. So I just kept the `LinearLayout` there so that it can still contain the chevron and keep it working as expected.

The second large refactor was on the Login screen. [Previously][change-account-history-animation], the account login widget was made to collapse the account history when the account input area lost focus. This made things a bit complicated, because as it was the history would collapse as soon as the user navigated to an account history entry. However, removing the code would prevent the history from collapsing automatically.

The solution used in this PR was to listen for focus events inside the account history child views, which include each row's label and remove button. However, this means that a sequence of "focus lost -> focus gained -> focus lost -> focus gained -> ..." events would be sent while the user was navigating inside the `AccountLogin` widget. To prevent starting and stopping the animation every time, the PR also adds a `Debouncer` helper class to only notify of the last event once Android has finished sending the focus changed events.

The final issue on the Login screen was that selecting an account history entry with the keyboard didn't clear the focus for that view. This meant that in that case the history wouldn't collapse while logging in. The fix for this was to force the background to gain focus when starting to log in.

Finally, the `InformationView` was refactored a bit to use the `observable` property delegate to make the code a bit more compact. It then uses the delegate for the `onClick` listener property so that the widget is made focusable if it can be clicked.

Git checklist:

[change-account-history-animation]: https://github.com/mullvad/mullvadvpn-app/pull/2078

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2177)


pinkisemils (commented):
> Reviewed 7 of 27 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2177)* status: 7 of 27 files reviewed, all discussions resolved

___
*[CHANGELOG.md, line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2177#-MJD6u-f4sl5U6XdQjXg:-MJD6u-f4sl5U6XdQjXh:b-kwsgrp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c1320db3f4e3dfc7ef7fb5e2777412646b3ec631/CHANGELOG.md#L37)):*
> ```Markdown
> - Remove the Quit button.
> - Add button to remove account and WireGuard key from history in the login screen.
> - Improve navigation in the app using a keyboard, so that touchless devices (lik TVs) can be used
> ```

A small wee typo here - ` (lik TVs)`

pinkisemils (changes requested):
> Reviewed 3 of 27 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2177)* status: 10 of 27 files reviewed, 1 unresolved discussion (waiting on @jvff)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/util/Debouncer.kt, line 9 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2177#-MJDBGyE7syqSuI4MRNF:-MJDBGyF8GpzED56QIzt:b-p25kvk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c1320db3f4e3dfc7ef7fb5e2777412646b3ec631/android/src/main/kotlin/net/mullvad/mullvadvpn/util/Debouncer.kt#L9)):*
> ```Kotlin
> // started. If the events are coming from the UI thread, this means that this class will only send
> // the last event received before the UI thread finishes its current task.
> class Debouncer<T>(initialValue: T, val intervalInMs: Long = 0) {
> ```

This docstring tells me an implementation detail but does not clarify the purpose of this object.

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2177)* status: 10 of 27 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[CHANGELOG.md, line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2177#-MJD6u-f4sl5U6XdQjXg:-MJDabHG20TcfFqU8aw7:b-dtkds3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c1320db3f4e3dfc7ef7fb5e2777412646b3ec631/CHANGELOG.md#L37)):*

A small wee typo here - ` (lik TVs)`
</blockquote></details>

Done, thanks!
___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/util/Debouncer.kt, line 9 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2177#-MJDBGyE7syqSuI4MRNF:-MJDack9-EGTb7e7hQ9_:bnyvdge) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c1320db3f4e3dfc7ef7fb5e2777412646b3ec631/android/src/main/kotlin/net/mullvad/mullvadvpn/util/Debouncer.kt#L9)):*

This docstring tells me an implementation detail but does not clarify the purpose of this object.
</blockquote></details>

Done, hopefully this is better :)

pinkisemils (approved):
> Reviewed 15 of 27 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2177)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/util/Debouncer.kt, line 9 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2177#-MJDBGyE7syqSuI4MRNF:-MJR7K78647xcXycXZif:bszm2f7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c1320db3f4e3dfc7ef7fb5e2777412646b3ec631/android/src/main/kotlin/net/mullvad/mullvadvpn/util/Debouncer.kt#L9)):*

Done, hopefully this is better :)
</blockquote></details>

Much better!

PR #2200, opened 2020-10-21
Detect and log if multiple daemon instances are running

jvff:
There have been a few problem reports with logs that seem a bit odd. It appears that multiple daemon instances could be running, which would cause problems. In order to try to figure out if that is what is happening, this PR adds an atomic counter to the `mullvad-jni` crate that is incremented whenever a new daemon instance is created and decremented whenever an instance finishes. In theory it should always be zero when the daemon is starting, and the PR will log an error message if the counter is not zero.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2200)

### DIFF
--- a/mullvad-jni/src/lib.rs
+++ b/mullvad-jni/src/lib.rs
@@ -26,13 +26,17 @@ use std::{
     io,
     path::{Path, PathBuf},
     ptr,
-    sync::{mpsc, Arc, Once},
+    sync::{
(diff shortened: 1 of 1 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2200)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2202, opened 2020-10-21
Add escape hatch

raksooo:
This PR adds an escape hatch which navigates back to the main view when escape is pressed.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Line wrap the file at 100 chars.                                              Th
 - Improve accessibility in the desktop app.
 - Add `--wait` flag to `connect`, `disconnect` and `reconnect` CLI subcommands to make the CLI wait
   for the target state to be reached before exiting.
+- Navigate back to the main view when escape is pressed.
 
 #### Windows
 - Add support for custom DNS resolvers (CLI only).
--- a/gui/src/renderer/components/KeyboardNavigation.tsx
+++ b/gui/src/renderer/components/KeyboardNavigation.tsx
@@ -1,0 +1,27 @@
+import React, { useCallback, useEffect } from 'react';
+import { useHistory } from 'react-router';
+import History from '../lib/history';
+
+interface IKeyboardNavigationProps {
+  children: React.ReactElement;
+}
+
+export default function KeyboardNavigation(props: IKeyboardNavigationProps) {
+  const history = useHistory() as History;
+
+  const handleKeyDown = useCallback(
+    (event: KeyboardEvent) => {
+      if (event.key === 'Escape') {
+        history.reset();
+      }
+    },
+    [history.reset],
+  );
+
+  useEffect(() => {
+    document.addEventListener('keydown', handleKeyDown);
+    return () => document.removeEventListener('keydown', handleKeyDown);
+  }, [handleKeyDown]);
+
+  return props.children;
+}
--- a/gui/src/renderer/components/Modal.tsx
+++ b/gui/src/renderer/components/Modal.tsx
@@ -152,7 +152,9 @@ class ModalAlertWithContext extends React.Component<IModalAlertProps & IModalCon
 
   public componentDidMount() {
     this.props.setActiveModal(true);
-    document.addEventListener('keydown', this.handleKeyPress);
+    // The `true` argument specifies that the event should be dispatched in the capture phase. This
+    // makes sure that this component catches the event before the escape hatch.
+    document.addEventListener('keydown', this.handleKeyPress, true);
 
     const modalContainer = this.props.modalContainerRef.current;
     if (modalContainer) {
@@ -170,7 +172,7 @@ class ModalAlertWithContext extends React.Component<IModalAlertProps & IModalCon
 
   public componentWillUnmount() {
     this.props.setActiveModal(false);
-    document.removeEventListener('keydown', this.handleKeyPress);
+    document.removeEventListener('keydown', this.handleKeyPress, true);
 
     this.appendScheduler.cancel();
     this.props.modalContainerRef.current?.removeChild(this.element);
@@ -219,6 +221,7 @@ class ModalAlertWithContext extends React.Component<IModalAlertProps & IModalCon
 
   private handleKeyPress = (event: KeyboardEvent) => {
     if (event.key === 'Escape') {
+      event.stopPropagation();
       this.props.close?.();
     }
   };
--- a/gui/src/renderer/lib/history.ts
+++ b/gui/src/renderer/lib/history.ts
@@ -1,4 +1,10 @@
-import { Location, Action, LocationListener, LocationDescriptor } from 'history';
+import { Location, Action, LocationDescriptor } from 'history';
+
+type LocationListener<S = unknown> = (
+  location: Location<S>,
+  action: Action,
+  entries: Location<S>[],
+) => void;
 
 // It currently isn't possible to implement this correctly with support for a generic state. State
 // can be added as a generic type (<S = unknown>) after this issue has been resolved:
@@ -27,24 +33,32 @@ export default class History {
   }
 
   public push = (nextLocation: LocationDescriptor<S>, nextState?: S) => {
+    const affectedEntries = [this.entries[this.index]];
     const location = this.createLocation(nextLocation, nextState);
     this.lastAction = 'PUSH';
     this.index += 1;
     this.entries.splice(this.index, this.entries.length - this.index, location);
-    this.notify();
+    this.notify(affectedEntries);
   };
 
   public replace = (nextLocation: LocationDescriptor<S>, nextState?: S) => {
+    const affectedEntries = [this.entries[this.index]];
     this.entries[this.index] = this.createLocation(nextLocation, nextState);
     this.lastAction = 'REPLACE';
-    this.notify();
+    this.notify(affectedEntries);
   };
 
   public go = (n: number) => {
     if (this.canGo(n)) {
-      this.index += n;
+      const nextIndex = this.index + n;
+      const affectedEntries =
+        this.index < nextIndex
+          ? this.entries.slice(this.index, nextIndex)
+          : this.entries.slice(nextIndex + 1, this.index + 1);
+
+      this.index = nextIndex;
       this.lastAction = 'POP';
-      this.notify();
+      this.notify(affectedEntries);
     }
   };
 
@@ -52,16 +66,18 @@ export default class History {
   public goForward = () => this.go(1);
 
   public reset = () => {
+    const affectedEntries = this.entries.slice(1);
     this.lastAction = 'POP';
     this.index = 0;
-    this.notify();
+    this.notify(affectedEntries);
   };
 
   public resetWith = (nextLocation: LocationDescriptor<S>, nextState?: S) => {
+    const affectedEntries = [...this.entries];
     this.entries = [this.createLocation(nextLocation, nextState)];
     this.lastAction = 'REPLACE';
     this.index = 0;
-    this.notify();
+    this.notify(affectedEntries);
   };
 
   public canGo(n: number) {
@@ -82,8 +98,8 @@ export default class History {
     throw Error('Not implemented');
   }
 
-  private notify() {
-    this.listeners.forEach((listener) => listener(this.location, this.action));
+  private notify(affectedEntries: Location<S>[]) {
+    this.listeners.forEach((listener) => listener(this.location, this.action, affectedEntries));
   }
 
   private createLocation(location: LocationDescriptor<S>, state?: S): Location<S> {
--- a/gui/src/renderer/routes.tsx
+++ b/gui/src/renderer/routes.tsx
@@ -15,6 +15,7 @@ import SelectLocationPage from './containers/SelectLocationPage';
 import SettingsPage from './containers/SettingsPage';
 import SupportPage from './containers/SupportPage';
 import WireguardKeysPage from './containers/WireguardKeysPage';
+import History from './lib/history';
 import { getTransitionProps } from './transitions';
 
 interface IAppRoutesState {
@@ -38,12 +39,14 @@ class AppRoutes extends React.Component<RouteComponentProps, IAppRoutesState> {
   public componentDidMount() {
     // React throttles updates, so it's impossible to capture the intermediate navigation without
     // listening to the history directly.
-    this.unobserveHistory = this.props.history.listen((location) => {
-      this.setState((state) => ({
-        previousLocation: state.currentLocation,
-        currentLocation: location,
-      }));
-    });
+    this.unobserveHistory = (this.props.history as History).listen(
+      (location, _action, affectedEntries) => {
+        this.setState({
+          previousLocation: affectedEntries[0],
+          currentLocation: location,
+        });
+      },
+    );
   }
 
   public componentWillUnmount() {
--- a/gui/src/renderer/routes.tsx
+++ b/gui/src/renderer/routes.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { Route, RouteComponentProps, Switch, withRouter } from 'react-router';
 import Launch from './components/Launch';
+import KeyboardNavigation from './components/KeyboardNavigation';
 import Focus, { IFocusHandle } from './components/Focus';
 import LinuxSplitTunnelingSettings from './components/LinuxSplitTunnelingSettings';
 import TransitionContainer, { TransitionView } from './components/TransitionContainer';
@@ -64,34 +65,36 @@ class AppRoutes extends React.Component<RouteComponentProps, IAppRoutesState> {
 
     return (
       <PlatformWindowContainer>
-        <Focus ref={this.focusRef}>
-          <TransitionContainer onTransitionEnd={this.onNavigation} {...transitionProps}>
-            <TransitionView viewId={location.key || ''}>
-              <Switch key={location.key} location={location}>
-                <Route exact={true} path="/" component={Launch} />
-                <Route exact={true} path="/login" component={LoginPage} />
-                <Route exact={true} path="/connect" component={ConnectPage} />
-                <Route exact={true} path="/settings" component={SettingsPage} />
-                <Route exact={true} path="/settings/language" component={SelectLanguagePage} />
-                <Route exact={true} path="/settings/account" component={AccountPage} />
-                <Route exact={true} path="/settings/preferences" component={PreferencesPage} />
-                <Route exact={true} path="/settings/advanced" component={AdvancedSettingsPage} />
-                <Route
-                  exact={true}
-                  path="/settings/advanced/wireguard-keys"
-                  component={WireguardKeysPage}
-                />
-                <Route
-                  exact={true}
-                  path="/settings/advanced/linux-split-tunneling"
-                  component={LinuxSplitTunnelingSettings}
-                />
-                <Route exact={true} path="/settings/support" component={SupportPage} />
-                <Route exact={true} path="/select-location" component={SelectLocationPage} />
-              </Switch>
-            </TransitionView>
-          </TransitionContainer>
-        </Focus>
+        <KeyboardNavigation>
+          <Focus ref={this.focusRef}>
+            <TransitionContainer onTransitionEnd={this.onNavigation} {...transitionProps}>
+              <TransitionView viewId={location.key || ''}>
+                <Switch key={location.key} location={location}>
+                  <Route exact={true} path="/" component={Launch} />
+                  <Route exact={true} path="/login" component={LoginPage} />
+                  <Route exact={true} path="/connect" component={ConnectPage} />
+                  <Route exact={true} path="/settings" component={SettingsPage} />
+                  <Route exact={true} path="/settings/language" component={SelectLanguagePage} />
+                  <Route exact={true} path="/settings/account" component={AccountPage} />
+                  <Route exact={true} path="/settings/preferences" component={PreferencesPage} />
+                  <Route exact={true} path="/settings/advanced" component={AdvancedSettingsPage} />
+                  <Route
+                    exact={true}
+                    path="/settings/advanced/wireguard-keys"
+                    component={WireguardKeysPage}
+                  />
+                  <Route
+                    exact={true}
+                    path="/settings/advanced/linux-split-tunneling"
+                    component={LinuxSplitTunnelingSettings}
+                  />
+                  <Route exact={true} path="/settings/support" component={SupportPage} />
+                  <Route exact={true} path="/select-location" component={SelectLocationPage} />
+                </Switch>
+              </TransitionView>
+            </TransitionContainer>
+          </Focus>
+        </KeyboardNavigation>
       </PlatformWindowContainer>
     );
   }


raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: 0 of 9 files reviewed, 1 unresolved discussion

___
*[gui/src/renderer/transitions.ts, line 51 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKBhUpN6JeO1OZtuXV7:-MKBhUpOAUIixN2xqzY0:bmqpcoj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/transitions.ts#L51)):*
> ```XML
>   r('/settings', '/settings/support', transitions.push),
>   r(/.*/, /\/settings.*/, transitions.slide),
>   r(/.*/, '/select-location', transitions.slide),
> ```

I tried a few different ways to allow animating from everything in settings back to the main view and finally came to this solution with allowing regular expressions for the paths. The only negative aspect that I see is that all subviews in settings need to have a path beginning with `/settings/`.

Another solution could be to allow `to` to be `null` as well but that has some unwanted effects such as that the app would animate between `/connect` and `/` with a slide animation.

pronebird (changes requested):
> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 5 unresolved discussions (waiting on @raksooo)

___
*[CHANGELOG.md, line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKE_1Zy6WmgLlK1vdcB:-MKE_1Zy6WmgLlK1vdcC:b-jbem7m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/CHANGELOG.md#L30)):*
> ```Markdown
> - Add `--wait` flag to `connect`, `disconnect` and `reconnect` CLI subcommands to make the CLI wait
>   for the target state to be reached before exiting.
> - Add escape hatch that navigates back to the main view when escape is pressed in the desktop app.
> ```

Buzzwords in changelog. It would be much simpler to say that pressing Escape takes user to the main view.
___
*[gui/src/renderer/transitions.ts, line 51 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKBhUpN6JeO1OZtuXV7:-MKEPrFhEpWKiajMWiA_:bi2tbvj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/transitions.ts#L51)):*

I tried a few different ways to allow animating from everything in settings back to the main view and finally came to this solution with allowing regular expressions for the paths. The only negative aspect that I see is that all subviews in settings need to have a path beginning with `/settings/`.

Another solution could be to allow `to` to be `null` as well but that has some unwanted effects such as that the app would animate between `/connect` and `/` with a slide animation.
</blockquote></details>

This is not really needed in my opinion, as we never go from anywhere to sub-path under `/settings`.
___
*[gui/src/renderer/components/EscapeHatch.tsx, line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKET0p78e48XHpIPLqp:-MKET0p78e48XHpIPLqq:b-eqd4mf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/EscapeHatch.tsx#L10)):*
> ```XML
> }
> 
> export default function EscapeHatchT(props: IEscapeHatchProps) {
> ```

FYI I see where the name comes from, but really, why not `KeyboardNavigation`? This is all what it is, it's plain simple.
___
*[gui/src/renderer/components/EscapeHatch.tsx, line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKETQtm72gYIsuM0gWE:-MKETQtm72gYIsuM0gWF:b-h79n6j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/EscapeHatch.tsx#L24)):*
> ```XML
>           history.push('/');
>         } else if (loggedIn) {
>           history.push('/connect');
> ```

Here is where you have to be careful. The way it's organised, when you push an item to history, you make it possible to come back to the preceding item. `NavigationBar` back buttons & close button (indirectly) use `history.goBack()`.

Look at `app.tsx`. `memoryHistory.replace()` is used exclusively when navigating to things like `/` or `/login`. There is no path from `/connect` to `/login`. The `/login` page is only accessibly programatically and there is no way to go forward from `/login` page t `/connect` once you logged out.

What I would do instead is `history.goBack(-n)` where `n` is the number of components that we have to pop before coming back to `/connect`, `/login` or `/`, whatever. It may not work flawlessly when trying to pop from the sub-page under `/settings/<sub_page>` but then you'd have to see what the history listener gives to `TransitionContainer` & perhaps either let it queue up the transitions or skip the sub-transitions.

The simplest solution would be to pop one item off history entries if `history.length > 1`, this way you would have more generic solution.

Also it's worth to note that while browser's history does not expose `history.entries`, we use `MemoryHistory` which in fact does, but it's a private API, however you may find some use of it if needed.
___
*[gui/src/renderer/components/Modal.tsx, line 224 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKEZHqiC0UY2xD7HODv:-MKEZHqiC0UY2xD7HODw:b-8w78kb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/Modal.tsx#L224)):*
> ```XML
>   private handleKeyPress = (event: KeyboardEvent) => {
>     if (event.key === 'Escape') {
>       event.stopPropagation();
> ```

FYI one reason why we might consider upgrading modals to the routes to be able to participate in container transitions. But that's something for the future perhaps.
___
*[gui/src/renderer/lib/transition-rule.ts, line 21 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKEZgBZCWk7Dslf3Mt8:-MKEZgBZCWk7Dslf3Mt9:b-yczq7i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/lib/transition-rule.ts#L21)):*
> ```XML
> 
>   constructor(from: string | RegExp, to: string | RegExp, private fork: ITransitionFork) {
>     this.from = typeof from === 'string' ? new RegExp(from) : from;
> ```

A dangerous game of wrapping any random string into `RegExp`. What can go wrong?
___
*[gui/src/renderer/redux/userinterface/reducers.ts, line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKEbDhT1z_9elxk8GHI:-MKEbDhU75JWD1DJwrH3:b-2rv87n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/redux/userinterface/reducers.ts#L15)):*
> ```XML
>   windowFocused: boolean;
>   scrollPosition: Record<string, [number, number]>;
>   connectedToDaemon: boolean;
> ```

Not needed as the `EscapeHutch` should be able to navigation without much structural knowledge of the app.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 6 unresolved discussions (waiting on @pronebird and @raksooo)

___
*[gui/src/renderer/transitions.ts, line 51 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKBhUpN6JeO1OZtuXV7:-MKEig8fEpu8tQ52sGBE:bmukhmc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/transitions.ts#L51)):*

This is not really needed in my opinion, as we never go from anywhere to sub-path under `/settings`.
</blockquote></details>

We will probably need to go from a subpath under `/settings` to `/`/`/login`/`/connect`
___
*[gui/src/renderer/components/EscapeHatch.tsx, line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKET0p78e48XHpIPLqp:-MKEgN0F-Od1I921aYl-:b-34eurc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/EscapeHatch.tsx#L10)):*

FYI I see where the name comes from, but really, why not `KeyboardNavigation`? This is all what it is, it's plain simple.
</blockquote></details>

I thought the name was more widely used for this type of feature, just did a search and only found a few pages using this naming. I'm fine with `KeyboardNavigation` :+1:
___
*[gui/src/renderer/components/EscapeHatch.tsx, line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKETQtm72gYIsuM0gWE:-MKEgx758-MbQ5IeRlkM:bz36b42) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/EscapeHatch.tsx#L24)):*
I first tried this using `entries` but figured I had to do it some other way since it's not public on `history`. Nice to know that it's public on `memoryHistory`! How do you suggest using `memoryHistory`? Passing it as a prop? Or just casting `history` to `memoryHistory`? I agree that `history.go(-n)` would be a way better solution.

> The simplest solution would be to pop one item off history entries if history.length > 1, this way you would have more generic solution.

Could you elaborate? I don't understand what you're suggesting.
___
*[gui/src/renderer/components/Modal.tsx, line 224 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKEZHqiC0UY2xD7HODv:-MKEgmFj-bAPAbrkVckq:b-ebtrgl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/Modal.tsx#L224)):*

FYI one reason why we might consider upgrading modals to the routes to be able to participate in container transitions. But that's something for the future perhaps.
</blockquote></details>

What do you mean with "be able to participate in container transitions"?

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 6 unresolved discussions (waiting on @pronebird and @raksooo)

___
*[gui/src/renderer/components/EscapeHatch.tsx, line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKETQtm72gYIsuM0gWE:-MKEnCWD-Ej8iroQRBHQ:b-9pe4nf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/EscapeHatch.tsx#L24)):*
> I first tried this using entries but figured I had to do it some other way since it's not public on history. Nice to know that it's public on memoryHistory! How do you suggest using memoryHistory? Passing it as a prop? Or just casting history to memoryHistory? I agree that history.go(-n) would be a way better solution.

I merely highlighted that as a possibility. Doesn't mean you have to do that but in short yes `let memoryHistory = history as MemoryHistory`.

That's how it's initialised in `app.tsx`:

```
 private memoryHistory = createMemoryHistory();
 private reduxStore = configureStore(this.memoryHistory);
```

Technically your task can be solved with one line of code, i.e:

```
history.goBack(history.length - 1);
```

> Could you elaborate? I don't understand what you're suggesting.

What I suggested is to navigate one step back as opposed to trying to take the user to the main view.

```
history.goBack()
```
___
*[gui/src/renderer/components/Modal.tsx, line 224 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKEZHqiC0UY2xD7HODv:-MKEta7H9how1rw0705p:bl1qwt3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/Modal.tsx#L224)):*

What do you mean with "be able to participate in container transitions"?
</blockquote></details>

So that modals are animated as other containers within routes. This could also simplify the components that are currently holding those modals inside. 

There is a missing link between the static routes & dynamically displayed modals though. But with a bit of creativity, I think it's not impossible to make dynamic routes.

What else comes to my mind is that `History` API offers `state`, so it would technically be possible to display modal as a sub-route and give it some `state`, the `state` can be anything, including callbacks or whatever.

You get it?

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 6 unresolved discussions (waiting on @pronebird and @raksooo)

___
*[gui/src/renderer/components/EscapeHatch.tsx, line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKETQtm72gYIsuM0gWE:-MKEybfxDmAdZ8fIbLwM:btb866f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/EscapeHatch.tsx#L24)):*

> I first tried this using entries but figured I had to do it some other way since it's not public on history. Nice to know that it's public on memoryHistory! How do you suggest using memoryHistory? Passing it as a prop? Or just casting history to memoryHistory? I agree that history.go(-n) would be a way better solution.

I merely highlighted that as a possibility. Doesn't mean you have to do that but in short yes `let memoryHistory = history as MemoryHistory`.

That's how it's initialised in `app.tsx`:

```
 private memoryHistory = createMemoryHistory();
 private reduxStore = configureStore(this.memoryHistory);
```

Technically your task can be solved with one line of code, i.e:

```
history.goBack(history.length - 1);
```

> Could you elaborate? I don't understand what you're suggesting.

What I suggested is to navigate one step back as opposed to trying to take the user to the main view.

```
history.goBack()
```
</blockquote></details>

Edit: `history.goBack(1 - history.length);` the number has to be negative obviously.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 6 unresolved discussions (waiting on @pronebird and @raksooo)

___
*[gui/src/renderer/components/EscapeHatch.tsx, line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKETQtm72gYIsuM0gWE:-MKEvltx1YhgLoB3rjK1:boejp5q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/EscapeHatch.tsx#L24)):*
> Technically your task can be solved with one line of code, i.e:

Unfortunately it's not that simple. When we call `history.replace('/login')` for example, it just replaces the current entry. After going to the account view and logging out the list of entries looks something like this:
```
[
  {
    "pathname": "/connect",
    "search": "",
    "hash": "",
    "key": "myktmk"
  },
  {
    "pathname": "/settings",
    "search": "",
    "hash": "",
    "key": "d20ixt"
  },
  {
    "pathname": "/login",
    "search": "",
    "hash": "",
    "key": "pedn51"
  }
]
```

It's of course possible to go back to the previous entry that has the pathname `/`/`/login`/`/connect`. Or maybe make sure the list of entries is emptied when we replace a path.

> What I suggested is to navigate one step back as opposed to trying to take the user to the main view.

The aim is to go back to the main view. That's the conclusion Matilda came to and in my opinion that behavior makes the most sense.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 6 unresolved discussions (waiting on @pronebird and @raksooo)

___
*[gui/src/renderer/transitions.ts, line 51 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKBhUpN6JeO1OZtuXV7:-MKEynd_6VsaH6SXTTXT:b-4vzsf1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/transitions.ts#L51)):*

We will probably need to go from a subpath under `/settings` to `/`/`/login`/`/connect`
</blockquote></details>

I think that the history stack should contain all of the information needed to go back without introducing additional rules.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 6 unresolved discussions (waiting on @raksooo)

___
*[gui/src/renderer/components/EscapeHatch.tsx, line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKETQtm72gYIsuM0gWE:-MKF20KZ3TdX05pu1Mi_:b-hkjb0l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/EscapeHatch.tsx#L24)):*

> Technically your task can be solved with one line of code, i.e:

Unfortunately it's not that simple. When we call `history.replace('/login')` for example, it just replaces the current entry. After going to the account view and logging out the list of entries looks something like this:
```
[
  {
    "pathname": "/connect",
    "search": "",
    "hash": "",
    "key": "myktmk"
  },
  {
    "pathname": "/settings",
    "search": "",
    "hash": "",
    "key": "d20ixt"
  },
  {
    "pathname": "/login",
    "search": "",
    "hash": "",
    "key": "pedn51"
  }
]
```

It's of course possible to go back to the previous entry that has the pathname `/`/`/login`/`/connect`. Or maybe make sure the list of entries is emptied when we replace a path.

> What I suggested is to navigate one step back as opposed to trying to take the user to the main view.

The aim is to go back to the main view. That's the conclusion Matilda came to and in my opinion that behavior makes the most sense.
</blockquote></details>

Yeah that's unfortunate, I had impression that `replace` wiped the stack. Either way, we're not obliged to use `memoryHistory`, we can roll out our own, nothing wrong with managing the application history the way that suits us the best?

https://github.com/ReactTraining/history/blob/master/packages/history/index.ts#L996

I think that it would also be better if `history.go(-2)` gave the listener all entries that are being removed off the stack. This way it would be easy to pick the right transition from the last item without doing much heuristics or running regex or any of that.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 6 unresolved discussions (waiting on @raksooo)

___
*[gui/src/renderer/components/EscapeHatch.tsx, line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKET0p78e48XHpIPLqp:-MKFDBNx5Z-F7wfxUKEm:b-brhqmo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/EscapeHatch.tsx#L10)):*

I thought the name was more widely used for this type of feature, just did a search and only found a few pages using this naming. I'm fine with `KeyboardNavigation` :+1:
</blockquote></details>

I think it would make it easier for anyone joining the team to get familiar with the codebase if we used the names that better convey the meaning.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: 2 of 6 files reviewed, 5 unresolved discussions (waiting on @pronebird and @raksooo)

___
*[CHANGELOG.md, line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKE_1Zy6WmgLlK1vdcB:-ML79v-k85Hx-qQacLxs:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/CHANGELOG.md#L30)):*

Buzzwords in changelog. It would be much simpler to say that pressing Escape takes user to the main view.
</blockquote></details>

Done.
___
*[gui/src/renderer/transitions.ts, line 51 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKBhUpN6JeO1OZtuXV7:-ML7AUu16sF8pJhIP9iu:b-9ocec6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/transitions.ts#L51)):*

I think that the history stack should contain all of the information needed to go back without introducing additional rules.
</blockquote></details>

Now with the custom history implementation we still need to be able to navigate from any subpath under `/settings` to `/`/`/login`/`/connect`. I see a few alternatives:
1. The solution above
2. Allow `to` to be `null` and specifying the transitions as
```
r('/', null, transition.slide),
r('/connect', null, transition.slide),
r('/login', null, transition.slide),
```
3. Look at `history.entries` and use the location at `index + 1` or `index - 1` depending on the situation. This seems a lot more error prone though.

Do you have any other ideas? What do you think about these three?
___
*[gui/src/renderer/components/EscapeHatch.tsx, line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKETQtm72gYIsuM0gWE:-ML7A56TCOIr7BcFHDZV:b-myh2fe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/components/EscapeHatch.tsx#L24)):*

Yeah that's unfortunate, I had impression that `replace` wiped the stack. Either way, we're not obliged to use `memoryHistory`, we can roll out our own, nothing wrong with managing the application history the way that suits us the best?

https://github.com/ReactTraining/history/blob/master/packages/history/index.ts#L996

I think that it would also be better if `history.go(-2)` gave the listener all entries that are being removed off the stack. This way it would be easy to pick the right transition from the last item without doing much heuristics or running regex or any of that.
</blockquote></details>

It's now using our own history implementation
___
*[gui/src/renderer/redux/userinterface/reducers.ts, line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKEbDhT1z_9elxk8GHI:-ML7AJUMA38lQWbZfCRL:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/redux/userinterface/reducers.ts#L15)):*

Not needed as the `EscapeHutch` should be able to navigation without much structural knowledge of the app.
</blockquote></details>

Done.

pronebird (commented):
> Reviewed 5 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 2 unresolved discussions (waiting on @raksooo)

___
*[gui/src/renderer/transitions.ts, line 51 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-MKBhUpN6JeO1OZtuXV7:-ML7CPB0198lKyGvly-U:b-424vbv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e36471a3750e6738c5bf6e86151328fc8d7bca6/gui/src/renderer/transitions.ts#L51)):*

Now with the custom history implementation we still need to be able to navigate from any subpath under `/settings` to `/`/`/login`/`/connect`. I see a few alternatives:
1. The solution above
2. Allow `to` to be `null` and specifying the transitions as
```
r('/', null, transition.slide),
r('/connect', null, transition.slide),
r('/login', null, transition.slide),
```
3. Look at `history.entries` and use the location at `index + 1` or `index - 1` depending on the situation. This seems a lot more error prone though.

Do you have any other ideas? What do you think about these three?
</blockquote></details>

Suppose we're at `/settings/advanced` & we want to go back to `/connect`.

Navigation stack before the call to `history.reset()`:

- `/connect`
- `/settings`
- `/settings/advanced`

Navigation stack after calling to `history.reset()`:

- `/connect`
- ~`/settings`~
- ~`/settings/advanced`~

IMO reverse the transition from `/settings` to `/connect` and use it to animate from `/settings/advanced` to `/connect`.

pronebird (commented):
> Reviewed 1 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 2 unresolved discussions (waiting on @raksooo)

pronebird (approved):
> Reviewed 4 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[gui/src/renderer/components/KeyboardNavigation.tsx, line 5 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-ML8LG_n5an8tHuaChZE:-ML8LG_n5an8tHuaChZF:b-41ur2h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4619e4a5b8951d6ddc7a981efb47b007c0873782/gui/src/renderer/components/KeyboardNavigation.tsx#L5)):*
> ```XML
> import History from '../lib/history';
> 
> interface IEscapeHatchProps {
> ```

Minor naming convention. Seems like this was not renamed.

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @pronebird)

___
*[gui/src/renderer/components/KeyboardNavigation.tsx, line 5 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-ML8LG_n5an8tHuaChZE:-ML8QY-n3hSE6IwFIzXm:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4619e4a5b8951d6ddc7a981efb47b007c0873782/gui/src/renderer/components/KeyboardNavigation.tsx#L5)):*

Minor naming convention. Seems like this was not renamed.
</blockquote></details>

Done.

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2202#-:-ML8aT3DC3Zm91iThLo4:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2211, opened 2020-10-26
Linux - sort routes by prefix length before application

pinkisemils:
Up until now, when initializing the split tunneling routing table, the routes were being added in arbitrary order. Since routes can have dependencies, route application can fail if the routes are being set in the wrong order. This seems to happen often when the routes are set manually via `ip route add ...`. The solution is to sort the routes by prefix length so that the more specific routes are always applied first. 

In an unrelated note, I updated the changelog for both this issue and the issue with NM and periodic disconnection.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2211)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ Line wrap the file at 100 chars.                                              Th
 #### Windows
 - Add support for custom DNS resolvers (CLI only).
 
+#### Linux
+- Use NetworkManager to create a WireGuard interface.
+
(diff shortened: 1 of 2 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2211)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2210, opened 2020-10-23
Support custom DNS on Linux

dlon:
Allow custom DNS resolvers to be used on Linux. For addresses in private IP ranges, DNS requests are permitted by the firewall on all non-tunnel interfaces. For other addresses, requests are only allowed on the tunnel interface.

Related PRs: #2205, #2188

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2210)

PR #2214, opened 2020-10-27
Support custom DNS on macOS

dlon:
Support custom DNS on macOS. The details are essentially the same as in https://github.com/mullvad/mullvadvpn-app/pull/2210.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@ Line wrap the file at 100 chars.                                              Th
 - Use NetworkManager to create a WireGuard interface.
 - Add support for custom DNS resolvers (CLI only).
 
+#### macOS
+- Add support for custom DNS resolvers (CLI only).
+
 ### Changed
 - Use the API to fetch API IP addresses instead of DNS.
 - Remove WireGuard keys during uninstallation after the firewall is unlocked.
--- a/mullvad-cli/src/cmds/mod.rs
+++ b/mullvad-cli/src/cmds/mod.rs
@@ -63,7 +63,7 @@ pub fn get_commands() -> HashMap<&'static str, Box<dyn Command>> {
         Box::new(Disconnect),
         Box::new(Reconnect),
         Box::new(Lan),
-        #[cfg(any(windows, target_os = "linux"))]
+        #[cfg(not(target_os = "android"))]
         Box::new(CustomDns),
         Box::new(Relay),
         Box::new(Reset),
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -26,7 +26,7 @@ use futures::{
 };
 use log::{debug, error, info, warn};
 use mullvad_rpc::AccountsProxy;
-#[cfg(any(windows, target_os = "linux"))]
+#[cfg(not(target_os = "android"))]
 use mullvad_types::settings::DnsOptions;
 use mullvad_types::{
     account::{AccountData, AccountToken, VoucherSubmission},
@@ -43,7 +43,7 @@ use mullvad_types::{
     wireguard::KeygenEvent,
 };
 use settings::SettingsPersister;
-#[cfg(any(windows, target_os = "linux"))]
+#[cfg(not(target_os = "android"))]
 use std::net::IpAddr;
 #[cfg(not(target_os = "android"))]
 use std::path::Path;
@@ -197,7 +197,7 @@ pub enum DaemonCommand {
     /// Set if IPv6 should be enabled in the tunnel
     SetEnableIpv6(oneshot::Sender<()>, bool),
     /// Set custom DNS servers to use instead of passing requests to the gateway
-    #[cfg(any(windows, target_os = "linux"))]
+    #[cfg(not(target_os = "android"))]
     SetDnsOptions(oneshot::Sender<()>, DnsOptions),
     /// Set MTU for wireguard tunnels
     SetWireguardMtu(oneshot::Sender<()>, Option<u16>),
@@ -582,7 +582,7 @@ where
         let tunnel_command_tx = tunnel_state_machine::spawn(
             settings.allow_lan,
             settings.block_when_disconnected,
-            #[cfg(any(windows, target_os = "linux"))]
+            #[cfg(not(target_os = "android"))]
             Self::get_custom_resolvers(&settings.tunnel_options.dns_options),
             tunnel_parameters_generator,
             log_dir,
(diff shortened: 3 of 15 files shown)


pinkisemils (changes requested):
> Reviewed 13 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214)* status: 13 of 14 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/firewall/macos.rs, line 167 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214#-MKfHOhN1Iv8d23mqEOp:-MKfHOhN1Iv8d23mqEOq:bdx56al) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/313275939c99722758242d455e18abc5d647f622/talpid-core/src/firewall/macos.rs#L167)):*
> ```Rust
>         if is_local {
>             // Block requests on the tunnel interface
>             let block_tunnel_tcp = self
> ```

Is the block necessary? Sure, we don't want the DNS requests to seep into the tunnel, but if it's already routed through the tunnel interface, there's no way it'll reach the destination. And since the user has specified a local resolver, they probably need to ensure that it'll be routed outside the tunnel anyway. On macOS and Linux this could be tested before going into the connected state.
___
*[talpid-core/src/firewall/macos.rs, line 199 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214#-MKfJpbI8DVJuRbCxT8A:-MKfJpbJ6hcsGsoG4GgV:b-z1iw4x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/313275939c99722758242d455e18abc5d647f622/talpid-core/src/firewall/macos.rs#L199)):*
> ```Rust
>             let allow_nontunnel_udp = self
>                 .create_rule_builder(FilterRuleAction::Pass)
>                 .direction(pfctl::Direction::Out)
> ```

You shouldn't have to specify 2 rules each to allow traffic to flow both ways, you should be able to just not specify the direction, and make the rule keep state.
___
*[talpid-core/src/firewall/macos.rs, line 551 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214#-MKfGjhW8kGtdMDTzNtq:-MKfGjhW8kGtdMDTzNtr:bbaeq2f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/313275939c99722758242d455e18abc5d647f622/talpid-core/src/firewall/macos.rs#L551)):*
> ```Rust
>     for net in (&*super::ALLOWED_LAN_NETS)
>         .iter()
>         .chain(&*super::LOOPBACK_NETS)
> ```

I think this can be simplified to `(&*super::ALLOWED_LAN_NETS).iter().chain(&*super::LOOPBACK_NETS).any(|net| net.contains(*address))`. And I think something like this is implemented elswhere in `talpid-core`, maybe even in a firewall module for another platform. It'd be better not to repeat this code.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/firewall/macos.rs, line 167 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214#-MKfHOhN1Iv8d23mqEOp:-MKio3ZtCI-oGF2WQLPN:bn9kbk5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/313275939c99722758242d455e18abc5d647f622/talpid-core/src/firewall/macos.rs#L167)):*

Is the block necessary? Sure, we don't want the DNS requests to seep into the tunnel, but if it's already routed through the tunnel interface, there's no way it'll reach the destination. And since the user has specified a local resolver, they probably need to ensure that it'll be routed outside the tunnel anyway. On macOS and Linux this could be tested before going into the connected state.
</blockquote></details>

It's probably necessary to enforce that random LAN addresses don't leak into the tunnel and eventually get out to the internet.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214)* status: 11 of 16 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/firewall/macos.rs, line 199 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214#-MKfJpbI8DVJuRbCxT8A:-MKiuCf91SOFsGbpL0dt:bh7nx2c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/313275939c99722758242d455e18abc5d647f622/talpid-core/src/firewall/macos.rs#L199)):*

You shouldn't have to specify 2 rules each to allow traffic to flow both ways, you should be able to just not specify the direction, and make the rule keep state.
</blockquote></details>

Done. Although it may still specify the direction.
___
*[talpid-core/src/firewall/macos.rs, line 551 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214#-MKfGjhW8kGtdMDTzNtq:-MKiuQoYDx8sk0aptpQK:b-nliye9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/313275939c99722758242d455e18abc5d647f622/talpid-core/src/firewall/macos.rs#L551)):*

I think this can be simplified to `(&*super::ALLOWED_LAN_NETS).iter().chain(&*super::LOOPBACK_NETS).any(|net| net.contains(*address))`. And I think something like this is implemented elswhere in `talpid-core`, maybe even in a firewall module for another platform. It'd be better not to repeat this code.
</blockquote></details>

Done. I used the same code for nftables.

pinkisemils (approved):
> Reviewed 4 of 5 files at r2, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[talpid-core/src/routing/linux.rs, line 821 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214#-MKiw2Kd2EqhIAlDHGK5:-MKiw2Kd2EqhIAlDHGK6:b-mkvvkh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a60b383a64c9538c08b5d2f5f7d3dff79c3a78f3/talpid-core/src/routing/linux.rs#L821)):*
> ```Rust
> 
>     async fn delete_route_if_exists(&self, route: &Route) -> Result<()> {
>         if let Err(error) = self.delete_route(route).await {
> ```

Technically this could be improved to have less indentation by using a match statement with arms for`Err(Error::NetlinkError(rnetlink::Error::NetlinkError(msg))) if msg.code == -lib::ESRCH`, `Ok()` returning `Ok(())` and an arm for passing through the real error.

pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2214#-:-MKj12bwDViEUeN-pQcT:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2218, opened 2020-10-27
Ignore loopback routes on Linux.

pinkisemils:
On some older kernels, I've seen netlink report routes which seemingly route 0.0.0.0/0 through the loopback interface. This is bonkers and makes no sense. The changes I have here would just ignore such routes when parsing them. Another, not as destructive change, would be to parse them as is - if there is no destination NLA, just assume a default route and add it as such, and filter it out later.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2218)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,8 +46,12 @@ Line wrap the file at 100 chars.                                              Th
   more smoothly.
 - Run app in landscape mode on TVs.
 
+#### Linux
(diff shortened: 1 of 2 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2218)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2219, opened 2020-10-27
Make NM WireGuard tunnel work on older NM versions

pinkisemils:
Earlier, I made some changes to allow creating a WireGuard interface via NetworkManager. This code works, but improvements could be made to make it work with older NetworkManager versions - `AddConnection2` is only available as of NM 1.20, whereas WireGuard is supported as of version `1.16`.  So firstly, the NetworkManager WireGuard tunnel implementation checks if the NetworkManager that's available is new enough, then it attempts to apply the config, first via `AddConnection2`, and if that fails since that method doesn't exist, via `AddConnectionUnsaved`. There were also issues with IPv6 config when applying our DNS config via NM, which are resolved by checking what method is set for `ipv6` management.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2219)

### DIFF
--- a/talpid-core/src/dns/linux/network_manager.rs
+++ b/talpid-core/src/dns/linux/network_manager.rs
@@ -222,7 +222,11 @@ impl NetworkManager {
             // if the link contains link local addresses, addresses shouldn't be reset
             if ipv6_settings
                 .get("method")
-                .map(|method| method.as_str() != Some("link-local"))
+                .map(|method| {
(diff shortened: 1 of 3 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2219)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2224, opened 2020-10-28
Prevent duplicate fwmark rules

dlon:
`RouteManager::enable_exclusions_routes` is called when entering the connecting state, but `NLM_F_REPLACE` is ignored when the rule is added, causing duplicate rules to be added if it is called multiple times.

This may occur if connecting fails, for example, and the daemon attempts to reconnect. In this case, the disconnected state is not entered before the connecting state is entered again, and so `RouteManager::disable_exclusions_routes` is not called in between.

This should not cause any serious issues, but it does clutter the routing rules list seen with `ip rule`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2224)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ Line wrap the file at 100 chars.                                              Th
 - Handle statically added routes.
 - Stop reconnecting when using WireGuard and NetworkManager.
 - Apply DNS config quicker when managing DNS via NetworkManager.
+- Fix routing rules sometimes being duplicated.
 
 
 ### Security
--- a/talpid-core/src/routing/linux.rs
+++ b/talpid-core/src/routing/linux.rs
@@ -216,6 +216,10 @@ impl RouteManagerImpl {
         // TODO: IPv6
         use netlink_packet_route::constants::*;
 
+        if let Ok(true) = self.exclusions_rule_exists().await {
+            return Ok(());
+        }
+
         let mut req = NetlinkMessage::from(RtnlMessage::NewRule(self.fwmark_rule_message()));
         req.header.flags = NLM_F_REQUEST | NLM_F_ACK | NLM_F_CREATE | NLM_F_REPLACE;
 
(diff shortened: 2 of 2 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2224)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[talpid-core/src/routing/linux.rs, line 259 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2224#-MKkh_Py8HkL7mDzgo_E:-MKkh_Py8HkL7mDzgo_F:bcy9bm2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ba353a2f5b29cf5192bac421eb677d1609c29629/talpid-core/src/routing/linux.rs#L259)):*
> ```Rust
>     }
> 
>     async fn exclusions_rule_exists(&mut self) -> Result<bool> {
> ```

Would it be sufficient to only call this when the route manager is initialized? The problem is that the rule may be removed by mistake, and it will only be restored if the daemon is restarted. But maybe this is not something to worry about.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2224)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[talpid-core/src/routing/linux.rs, line 259 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2224#-MKkh_Py8HkL7mDzgo_E:-MKkiKHI9zH3BVAWc5o3:b-bis04p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ba353a2f5b29cf5192bac421eb677d1609c29629/talpid-core/src/routing/linux.rs#L259)):*

Would it be sufficient to only call this when the route manager is initialized? The problem is that it may be removed mistakenly, and it will only be restored if the daemon is restarted. But maybe this is not something to worry about.
</blockquote></details>

The above comment was actually referring to `enable_exclusions_routes`.

pinkisemils (approved):
Since this is a resource that's tied to a single instance of a tunnel, maybe this is something that should be cleaned up by a tunnel regardless of tunnel state? For now, we've been relying on the fact that the tunnel interface will be removed, and all the routes that use it will be dropped as a side effect. Maybe the tunnel should be more self-aware? The tunnel state machine could also clean up after itself - it does have to go from the connecting state back into itself somehow. But this approach also works well.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2224#-:-MKl7lgrEsEOUdw9sBb_:b3yjfjg)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/routing/linux.rs, line 259 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2224#-MKkh_Py8HkL7mDzgo_E:-MKl6bRtDkve3cDgEzoe:b-s4r6g6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ba353a2f5b29cf5192bac421eb677d1609c29629/talpid-core/src/routing/linux.rs#L259)):*

The above comment was actually referring to `enable_exclusions_routes`.
</blockquote></details>

Technically yes. I'm not aware of any piece of software that would play the role of a BOFH here. But due to similar reasons why we can't just read the `Mode` and `RcManager` fields of [DnsManager](https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.DnsManager.html) (users feel like NM is misbehaving, so they replace `/etc/resolv.conf` and make it immutable), I think it's more pragmatic to check if the split tunneling rules exist or not, and apply them again if needed. In general, I think it'd be cool if the daemon would deal with ephemeral and well-wishing wrong-doing user misconfigurations via reconnection.
But that's just my take, and I'm OK with being wrong about this.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2224)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

PR #2237, opened 2020-11-04
Don't manage DNS via systemd-resolved in NM module

dlon:
We generally fall back on NM if systemd-resolved cannot be used. In this case, if NM manages DNS via systemd-resolved, the configuration that is set on the connection is apparently ignored. Solve this by returning an error and falling back on the static resolver.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2237)

### DIFF
--- a/talpid-core/src/dns/linux/network_manager.rs
+++ b/talpid-core/src/dns/linux/network_manager.rs
@@ -40,7 +40,7 @@ pub enum Error {
     #[error(display = "Failed to match the returned D-Bus object with expected type")]
     MatchDBusTypeError(#[error(source)] dbus::arg::TypeMismatchError),
 
-    #[error(display = "DNS is managed by systemd-resolved - NM can't enforce DNS globally")]
+    #[error(display = "DNS is managed by systemd-resolved")]
     SystemdResolved,
 
     #[error(display = "Failed to find obtain devices from network manager")]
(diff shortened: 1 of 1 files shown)


pinkisemils (commented):
We should really investigate to see under what circumstances NM would be set to manage DNS via systemd-resolved yet the daemon is unable to do the same. The daemon does exactly the same thing NM would do, so I'd rather see that error resolved. And I'm increasingly leaning towards telling users that if they screw with their system, they can *deal with it*.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2237#-:-MLI1XH0DbXLCFL5Nxs-:b-il1naq)* status: 0 of 1 files reviewed, all discussions resolved

pinkisemils (commented):
Turns out, this seems to be an upstream bug, which should be fixed already. As such, I really don't believe we should be merging this.
https://bugs.launchpad.net/ubuntu/+source/network-manager/+bug/1847651

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2237#-:-MLI7Vy76mPjeusBAvx7:b-g8y6x)* status: 0 of 1 files reviewed, all discussions resolved

dlon:
Closing. A different approach will be used: #2241

PR #2241, opened 2020-11-04
Prefer systemd-resolved over NM if it's working

pinkisemils:
Originally, there were issues with NetworkManager not being able to set DNS servers per interface when it's managing DNS via systemd-resolved. Unfortunately, it fails silently and, worse still, there are users who have a config where `/etc/resolv.conf` is managed via NetworkManager but actual DNS management is done via systemd-resolved. In such a config, the daemon would fall back to using NetworkManager and it would fail silently. Whilst technically, NM is buggy here, the changes here would make the daemon prefer systemd-resolved over NM in a case where `/etc/resolv.conf` is still managed by NetworkManager yet DNS is actually managed via systemd-resolved's local resolver. This is achieved by verifying if NM is managing `/etc/resolv.conf` and relying on systemd-resolved when verifiying if the daemon should use systemd-resolved to apply our DNS config.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,8 @@ Line wrap the file at 100 chars.                                              Th
 - Stop reconnecting when using WireGuard and NetworkManager.
 - Apply DNS config quicker when managing DNS via NetworkManager.
 - Fix routing rules sometimes being duplicated.
+- When NetworkManager is managing /etc/resolv.conf but ultimately using systemd-resolved, use
+  systemd-resolved directly to manage DNS.
 
 
 ### Security
--- a/talpid-core/src/dns/linux/network_manager.rs
+++ b/talpid-core/src/dns/linux/network_manager.rs
@@ -127,10 +127,7 @@ impl NetworkManager {
             _ => (),
         };
 
-
-        let expected_resolv_conf = "/var/run/NetworkManager/resolv.conf";
-        let actual_resolv_conf = "/etc/resolv.conf";
-        if !eq_file_content(&expected_resolv_conf, &actual_resolv_conf) {
+        if !verify_etc_resolv_conf_contents() {
             log::debug!("/etc/resolv.conf differs from reference resolv.conf, therefore NM is not managing DNS");
             return Err(Error::NetworkManagerNotManagingDns);
         }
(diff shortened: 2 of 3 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/dns/linux/network_manager.rs, line 417 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241#-MLM_pIy7uuyDgEhKgfB:-MLM_pIz2xR1arp7bkhc:bz8owi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/470eb1825032a07bf23cd5f95692a6c3cbe5794e/talpid-core/src/dns/linux/network_manager.rs#L417)):*
> ```Rust
> pub fn is_nm_managing_via_resolved(connection: &Connection) -> bool {
>     let check_nm = || -> Result<bool> {
>         let management_mode: String = connection
> ```

It might be easier to follow if consistent variable names were used for `RC_MANAGEMENT_MODE_KEY` and `DNS_MODE_KEY`.
___
*[talpid-core/src/dns/linux/network_manager.rs, line 420 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241#-MLMZsgXCRVLsevIMCuL:-MLMZsgXCRVLsevIMCuM:b-oy2e1g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/470eb1825032a07bf23cd5f95692a6c3cbe5794e/talpid-core/src/dns/linux/network_manager.rs#L420)):*
> ```Rust
>             .with_path(NM_BUS, NM_DNS_MANAGER_PATH, RPC_TIMEOUT_MS)
>             .get(NM_DNS_MANAGER, DNS_MODE_KEY)
>             .map_err(Error::TooOldNetworkManager)?;
> ```

This error seems off. For example, NM might not exist.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241)* status: all files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-core/src/dns/linux/network_manager.rs, line 437 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241#-MLMdgaw0beLtKSdGTI3:-MLMdgax3p-UCzDtaVQP:bh4x3ev) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/470eb1825032a07bf23cd5f95692a6c3cbe5794e/talpid-core/src/dns/linux/network_manager.rs#L437)):*
Could we cover more cases? E.g. by calling `Reload()` in [`org.freedesktop.NetworkManager`](https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.html#gdbus-method-org-freedesktop-NetworkManager.Reload).

> (0x02) means to update DNS configuration, which usually involves writing /etc/resolv.conf anew.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/dns/linux/network_manager.rs, line 420 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241#-MLMZsgXCRVLsevIMCuL:-MLN2FMB9WO9Mv3tPz8k:bs436ai) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/470eb1825032a07bf23cd5f95692a6c3cbe5794e/talpid-core/src/dns/linux/network_manager.rs#L420)):*

This error seems off. For example, NM might not exist.
</blockquote></details>

I'll use a more generic one here then.
___
*[talpid-core/src/dns/linux/network_manager.rs, line 437 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241#-MLMdgaw0beLtKSdGTI3:-MLN0T-I1NZNrsn8fjxL:bbmi1y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/470eb1825032a07bf23cd5f95692a6c3cbe5794e/talpid-core/src/dns/linux/network_manager.rs#L437)):*

Could we cover more cases? E.g. by calling `Reload()` in [`org.freedesktop.NetworkManager`](https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.html#gdbus-method-org-freedesktop-NetworkManager.Reload).

> (0x02) means to update DNS configuration, which usually involves writing /etc/resolv.conf anew.
</blockquote></details>

Good point, I'll try that.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/dns/linux/network_manager.rs, line 437 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241#-MLMdgaw0beLtKSdGTI3:-MLNFNO99rdDcqpYZ0w0:b-1ryeoz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/470eb1825032a07bf23cd5f95692a6c3cbe5794e/talpid-core/src/dns/linux/network_manager.rs#L437)):*

Good point, I'll try that.
</blockquote></details>

Seems to work well!

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/dns/linux/network_manager.rs, line 437 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241#-MLMdgaw0beLtKSdGTI3:-MLNGYk07rUV6U7HFTs5:bifeohk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/470eb1825032a07bf23cd5f95692a6c3cbe5794e/talpid-core/src/dns/linux/network_manager.rs#L437)):*

Seems to work well!
</blockquote></details>

:lgtm:. It bothers me that `is_nm_managing_via_resolved` has the side effect of replacing resolv.conf, because the name suggests it has no side effects. Maybe it should be a separate function. But it's not a big deal.

I wonder if we could also do this instead of (only) calling `verify_etc_resolv_conf_contents`.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/dns/linux/network_manager.rs, line 437 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241#-MLMdgaw0beLtKSdGTI3:-MLNKYsE-6guo87e-Dec:b-jw1xcn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/470eb1825032a07bf23cd5f95692a6c3cbe5794e/talpid-core/src/dns/linux/network_manager.rs#L437)):*

:lgtm:. It bothers me that `is_nm_managing_via_resolved` has the side effect of replacing resolv.conf, because the name suggests it has no side effects. Maybe it should be a separate function. But it's not a big deal.

I wonder if we could also do this instead of (only) calling `verify_etc_resolv_conf_contents`.
</blockquote></details>

In `ensure_resolv_conf_is_managed`, that is.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2241)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2249, opened 2020-11-09
Disable NetworkManager's connectivity check.

pinkisemils:
It seems that NetworkManager can hang if it can't do it's connectivity check - which can often be induced by applying our firewall rules. As such, I've changed the tunnel state machine to disable the connectivity check before applying the firewall rules, and restoring it afterwards. No effort is made to track the flag that enables the connectivity check - the user may enable it after the rules have been applied - I don't believe we should do anything about that.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ Line wrap the file at 100 chars.                                              Th
 #### Linux
 - Optionally use NetworkManager to create WireGuard devices.
 - Add support for custom DNS resolvers (CLI only).
+- Disable NetworkManager's connectivity check before applying firewall rules to avoid triggerring
+  NetworkManager's [bug](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/312#note_453724)
 
 #### macOS
 - Add support for custom DNS resolvers (CLI only).
--- a/talpid-core/src/dns/linux/mod.rs
+++ b/talpid-core/src/dns/linux/mod.rs
@@ -64,6 +64,14 @@ impl super::DnsMonitorT for DnsMonitor {
         }
         Ok(())
     }
+
+    fn dbus_connection(&self) -> Option<&dbus::ffidisp::Connection> {
+        match &self.inner {
+            Some(DnsMonitorHolder::NetworkManager(nm)) => Some(&nm.dbus_connection),
+            Some(DnsMonitorHolder::SystemdResolved(sdr)) => Some(&sdr.dbus_connection),
+            _ => None,
+        }
+    }
 }
 
 pub enum DnsMonitorHolder {
--- a/talpid-core/src/dns/linux/network_manager.rs
+++ b/talpid-core/src/dns/linux/network_manager.rs
@@ -77,7 +77,7 @@ lazy_static! {
 }
 
 pub struct NetworkManager {
-    dbus_connection: Connection,
+    pub dbus_connection: Connection,
     device: Option<String>,
     settings_backup: Option<HashMap<String, HashMap<String, Variant<Box<dyn RefArg>>>>>,
 }
--- a/talpid-core/src/dns/linux/systemd_resolved.rs
+++ b/talpid-core/src/dns/linux/systemd_resolved.rs
@@ -79,7 +79,7 @@ lazy_static! {
 }
 
 pub struct SystemdResolved {
-    dbus_connection: Connection,
+    pub dbus_connection: Connection,
     interface_link: Option<(String, dbus::Path<'static>)>,
 }
 
--- a/talpid-core/src/dns/mod.rs
+++ b/talpid-core/src/dns/mod.rs
@@ -53,6 +53,11 @@ impl DnsMonitor {
         log::info!("Resetting DNS");
         self.inner.reset()
     }
+
+    #[cfg(target_os="linux")]
+    fn dbus_connection(&self) -> Option<&dbus::ffidisp::Connection> {
+        self.inner.dbus_connection()
+    }
 }
 
 trait DnsMonitorT: Sized {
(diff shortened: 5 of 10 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/tunnel_state_machine/mod.rs, line 340 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249#-MLgbN9C2jLV1t4GbPLj:-MLgbN9C2jLV1t4GbPLk:b-jmuics) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/360bdf8cad273f04c1422c3d4c37a3fb1bed5f99/talpid-core/src/tunnel_state_machine/mod.rs#L340)):*
> ```Rust
>     /// reset whenever the firewall is cleared.
>     #[cfg(target_os = "linux")]
>     pub fn disable_connectivity_check(&mut self) {
> ```

It would be good to hide some of this Linux-specific code instead of putting it directly in the tunnel state machine. I'm not sure what the best way to go about it would be. I would probably create a trait for handling different events, and store it as a list of observers.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/tunnel_state_machine/mod.rs, line 340 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249#-MLgbN9C2jLV1t4GbPLj:-MLgpmC6AJdeU1Nu82K2:b-qlkrq7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/360bdf8cad273f04c1422c3d4c37a3fb1bed5f99/talpid-core/src/tunnel_state_machine/mod.rs#L340)):*

It would be good to hide some of this Linux-specific code instead of putting it directly in the tunnel state machine. I'm not sure what the best way to go about it would be. I would probably create a trait for handling different events, and store it as a list of observers.
</blockquote></details>

Simpler to just create a separate module file and conditionally include it and conditionally call it
```rust
#[cfg(target_os = "linux")]
linux::disable_connectivity_check()
```
Much simpler than trying to express some generic trait that will only be used for a single thing anyway. It can be generalized later if it turns out more than one platform needs some custom setup code.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

___
*[talpid-core/src/tunnel_state_machine/mod.rs, line 340 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249#-MLgbN9C2jLV1t4GbPLj:-MLgxU-R3Ks3w5055LhP:bnpq09v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/360bdf8cad273f04c1422c3d4c37a3fb1bed5f99/talpid-core/src/tunnel_state_machine/mod.rs#L340)):*

Simpler to just create a separate module file and conditionally include it and conditionally call it
```rust
#[cfg(target_os = "linux")]
linux::disable_connectivity_check()
```
Much simpler than trying to express some generic trait that will only be used for a single thing anyway. It can be generalized later if it turns out more than one platform needs some custom setup code.
</blockquote></details>

 Since this function meddles with the system's state that needs to be restored later, I think it makes sense to have it be implemented on here, rather than be an opaque type.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon and @pinkisemils)

___
*[talpid-core/src/tunnel_state_machine/mod.rs, line 340 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249#-MLgbN9C2jLV1t4GbPLj:-MLhDSbsCeV6xUH14ROz:bce9adz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/360bdf8cad273f04c1422c3d4c37a3fb1bed5f99/talpid-core/src/tunnel_state_machine/mod.rs#L340)):*

 Since this function meddles with the system's state that needs to be restored later, I think it makes sense to have it be implemented on here, rather than be an opaque type.
</blockquote></details>

Yes. I agree a reader of the state machine should easily understand that on Linux we have to fiddle with NM first. That's why I did not want any generic trait. But the actual implementation (the bulk of the code) can still be shoved into a separate module without reducing the understandability of the code. As long as the call into that module is clear enough what it does.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

___
*[talpid-core/src/tunnel_state_machine/mod.rs, line 340 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249#-MLgbN9C2jLV1t4GbPLj:-MLhRFveEnPlxGUsFP3K:b1kg9kc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/360bdf8cad273f04c1422c3d4c37a3fb1bed5f99/talpid-core/src/tunnel_state_machine/mod.rs#L340)):*

Yes. I agree a reader of the state machine should easily understand that on Linux we have to fiddle with NM first. That's why I did not want any generic trait. But the actual implementation (the bulk of the code) can still be shoved into a separate module without reducing the understandability of the code. As long as the call into that module is clear enough what it does.
</blockquote></details>

The bulk of the code here tries to see if a DBus connection already exists and uses the existing one or creates a new one.

dlon (approved):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2249)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

PR #2259, opened 2020-11-11
Randomize API address cache

dlon:
This generates a list of socketaddrs and uses it initially instead of assuming that the main IP address will be reachable. The address cache is also shuffled after being loaded from disk or updated via the API.

The hard-coded address still exists, because it is used to generate the bundled relay and API caches.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)


faern (changes requested):
> Reviewed 12 of 21 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 12 of 21 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[mullvad-problem-report/src/lib.rs, line 274 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLvxmMa8R8oPCsC5iSf:-MLvxmMa8R8oPCsC5iSg:bohtdjl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-problem-report/src/lib.rs#L274)):*
> ```Rust
>         .map_err(Error::CreateRuntime)?;
> 
>     let cache_dir = mullvad_paths::cache_dir();
> ```

the problem report tool and the daemon run as different users. Will this return the correct cache directory for the normal user? Also about write-permissions. The normal user should not be able to write to the cache files used by the daemon, but it would be good if they could read them, so the problem report tool does not become crippled from not having access to the latest known API IPs.
___
*[mullvad-rpc/src/address_cache.rs, line 87 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLw1BBp2J1y-56F6lRs:-MLw1BBq74QSsleh2j_N:bnaei58) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-rpc/src/address_cache.rs#L87)):*
> ```Rust
>             addresses.sort();
>             inner.addresses.sort();
>             if addresses != inner.addresses {
> ```

Here the inner addresses are sorted. But if the comparison fails it's never shuffled again. Does that not leave `inner.addresses` in a sorted state after a call to `set_addresses` with the same list content it already had?

faern (changes requested):
> Reviewed 1 of 21 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 13 of 21 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[mullvad-rpc/src/address_cache.rs, line 164 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLw5PpNCeC7t0y4jSIX:-MLw5PpNCeC7t0y4jSIY:b-by66md) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-rpc/src/address_cache.rs#L164)):*
> ```Rust
>         let mut cache = Self {
>             addresses,
>             ..Default::default()
> ```

This is the only place where we instantiate a `AddressCacheInner`, no? Does the `Default` impl add any functionality we need? The way I read it it's currently just one more mental indirection to understand how a new instance is created. If it's created in more than one place, but we don't explicitly need the `Default` for anything I would have designed it with a `AddressCacheInner::new(addresses: Vec<...>)` instead IHMO.
___
*[mullvad-rpc/src/lib.rs, line 88 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLw6X523sD069csvvsH:-MLw6X523sD069csvvsI:biat7uw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-rpc/src/lib.rs#L88)):*
> ```Rust
>                     );
>                 }
>                 AddressCache::with_cache(resource_file).await?
> ```

Won't this make it try to save any updates to the resource file rather than the cache file? The resource file should never be changed. It must try to 1) load the cache file. 2) if (1) fails try to load the resource file. 3) On any future updates from getting new IPs from the API the cache file must be written.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 13 of 21 files reviewed, 4 unresolved discussions (waiting on @dlon)

___
*[mullvad-rpc/Cargo.toml, line 39 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLw77Xk-O0nQruz0hvY:-MLw77Xl2KZzXKW7co4I:b-r7vdxy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-rpc/Cargo.toml#L39)):*
> ```TOML
> 
> [[bin]]
> name = "address_cache"
> ```

I don't think these has to be named...? Cargo should automatically pick up anything in `src/bin/` and make it a program.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 13 of 21 files reviewed, 5 unresolved discussions (waiting on @dlon)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLwzicJ76SoQm4fy5W9:-MLwzicJ76SoQm4fy5WA:b-51kbrw) (no related file):*
@dlon Can we add logging that logs the new address every time the address cache skips to the next one? (And the first one used as well). Without this it will be really hard to see if the algorithm ever fails and becomes deterministic or whatever. Good for debugging to be able to see which IP it tries to use.

---

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 13 of 21 files reviewed, 5 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-problem-report/src/lib.rs, line 274 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLvxmMa8R8oPCsC5iSf:-MLweAHi2i0Wqi1v7Gb8:b2w3n9k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-problem-report/src/lib.rs#L274)):*

the problem report tool and the daemon run as different users. Will this return the correct cache directory for the normal user? Also about write-permissions. The normal user should not be able to write to the cache files used by the daemon, but it would be good if they could read them, so the problem report tool does not become crippled from not having access to the latest known API IPs.
</blockquote></details>

The directory will be different. `mullvad-setup` contains code for obtaining the path to the system user's app data directory. It can be reused. But unprivileged users still won't be able to access the cache directory. It could either ask for elevation or just fall back on using the resources directory in this case.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 13 of 21 files reviewed, 5 unresolved discussions (waiting on @dlon)

___
*[mullvad-problem-report/src/lib.rs, line 274 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLvxmMa8R8oPCsC5iSf:-MLxH6392D_I7Dekg-Xh:b-amyrtw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-problem-report/src/lib.rs#L274)):*

This is true. `mullvad-setup` contains code for obtaining the path to the system user's app data directory. It can be reused. But unprivileged users still won't be able to access the cache directory. It could either ask for elevation or just fall back on using the resources directory in this case.
</blockquote></details>

Oh. If the tool can't read the cache. Then the problem report tool is forever doomed to use the IPs the app was bundled with. Unless it tries to fetch the list of IPs and cache it under the cache directory for that user. But that's still fragile towards the problem report tool not being run until far later than installation time :/

Any reason not to allow unpriv users to read this cache file? As far as I know there is nothing secret there. The WireGuard keys that were secret have been moved from there out to the settings directory.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 13 of 21 files reviewed, 5 unresolved discussions (waiting on @dlon)

___
*[mullvad-rpc/src/address_cache.rs, line 87 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLw1BBp2J1y-56F6lRs:-MM01JSyBvdt71lfHNr9:b-2t1j1s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-rpc/src/address_cache.rs#L87)):*

Here the inner addresses are sorted. But if the comparison fails it's never shuffled again. Does that not leave `inner.addresses` in a sorted state after a call to `set_addresses` with the same list content it already had?
</blockquote></details>

As a follow up to this thread.. Just re-shuffling inner if it is not changed is also not ideal. Because then the list will get re-shuffled a lot even when the content does not change. Meaning it will not test the entire list. If we re-fetch the IPs from the API in a fixed interval, the practical result of that will be that the algorithm is essentially random, not shuffle.

I think the equality check has to be made on a separate instance of the list, not the shuffled one actually used for fetching the next IP to try.

dlon (commented):
There are still a few references to `FALLBACK_API_ADDRESS`. Those could be replaced with `unwrap`, if the code is correct, but I guess this doesn't hurt.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-:-MM0HOGaE9TULYUMYSyh:brype9g)* status: 10 of 21 files reviewed, 5 unresolved discussions (waiting on @faern)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLwzicJ76SoQm4fy5W9:-MM0FQQt9qw4RxAS7Hct:b-896fix) (no related file):*

@dlon Can we add logging that logs the new address every time the address cache skips to the next one? (And the first one used as well). Without this it will be really hard to see if the algorithm ever fails and becomes deterministic or whatever. Good for debugging to be able to see which IP it tries to use.
</blockquote></details>

Done.

---
___
*[mullvad-problem-report/src/lib.rs, line 274 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLvxmMa8R8oPCsC5iSf:-MM0FThN4_lPSOAa-uOj:b-gaj9qj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-problem-report/src/lib.rs#L274)):*

Oh. If the tool can't read the cache. Then the problem report tool is forever doomed to use the IPs the app was bundled with. Unless it tries to fetch the list of IPs and cache it under the cache directory for that user. But that's still fragile towards the problem report tool not being run until far later than installation time :/

Any reason not to allow unpriv users to read this cache file? As far as I know there is nothing secret there. The WireGuard keys that were secret have been moved from there out to the settings directory.
</blockquote></details>

For now it will use the resource directory. We can allow everyone to read the cache file but still can't obtain the path to the system user's AppData directory unless running with elevated privileges (on Windows).
___
*[mullvad-rpc/Cargo.toml, line 39 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLw77Xk-O0nQruz0hvY:-MM0G4oWDv4GGSM6MZXV:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-rpc/Cargo.toml#L39)):*

I don't think these has to be named...? Cargo should automatically pick up anything in `src/bin/` and make it a program.
</blockquote></details>

Done.
___
*[mullvad-rpc/src/address_cache.rs, line 87 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLw1BBp2J1y-56F6lRs:-MM0GE5wCv20OvEgrzPy:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-rpc/src/address_cache.rs#L87)):*

As a follow up to this thread.. Just re-shuffling inner if it is not changed is also not ideal. Because then the list will get re-shuffled a lot even when the content does not change. Meaning it will not test the entire list. If we re-fetch the IPs from the API in a fixed interval, the practical result of that will be that the algorithm is essentially random, not shuffle.

I think the equality check has to be made on a separate instance of the list, not the shuffled one actually used for fetching the next IP to try.
</blockquote></details>

Done.
___
*[mullvad-rpc/src/address_cache.rs, line 164 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLw5PpNCeC7t0y4jSIX:-MM0Gmu8E7cy88DJgiTB:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-rpc/src/address_cache.rs#L164)):*

This is the only place where we instantiate a `AddressCacheInner`, no? Does the `Default` impl add any functionality we need? The way I read it it's currently just one more mental indirection to understand how a new instance is created. If it's created in more than one place, but we don't explicitly need the `Default` for anything I would have designed it with a `AddressCacheInner::new(addresses: Vec<...>)` instead IHMO.
</blockquote></details>

Done.
___
*[mullvad-rpc/src/lib.rs, line 88 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLw6X523sD069csvvsH:-MM0Gv-b2WFCGZ9MFVZB:bt2vr1b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-rpc/src/lib.rs#L88)):*

Won't this make it try to save any updates to the resource file rather than the cache file? The resource file should never be changed. It must try to 1) load the cache file. 2) if (1) fails try to load the resource file. 3) On any future updates from getting new IPs from the API the cache file must be written.
</blockquote></details>

Fixed.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 10 of 23 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-rpc/src/lib.rs, line 62 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MM1-gnI9aI4-K5GxpNG:-MM1-gnJ4etjA9f-aIvN:b-j7ylz3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d02f0d56861a211b14f10ba9430b475e85ff9663/mullvad-rpc/src/lib.rs#L62)):*
> ```Rust
>             https_connector: HttpsConnectorWithSni::new(),
>             handle,
>             address_cache: AddressCache::new(vec![SocketAddr::new(API_IP, 443)], None)?,
> ```

The port is now duplicated. And it's a bit hard to follow when `API_IP` and `FALLBACK_API_ADDRESS` are split up between two files. Can we define the hardcoded single fallback socket addr in one single place maybe?

pinkisemils (commented):
> Reviewed 10 of 21 files at r1, 1 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 12 of 23 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-problem-report/src/lib.rs, line 274 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLvxmMa8R8oPCsC5iSf:-MM0TSr2Engol1TgBif5:b6vwrca) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-problem-report/src/lib.rs#L274)):*

For now it will use the resource directory. We can allow everyone to read the cache file but still can't obtain the path to the system user's AppData directory unless running with elevated privileges (on Windows).
</blockquote></details>

A dirty solution would be to save the cached addresses in both the resources directory and the cache directory.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 12 of 23 files reviewed, 2 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[mullvad-problem-report/src/lib.rs, line 274 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLvxmMa8R8oPCsC5iSf:-MMFDso23kjs4X3oCRdf:b87t9d0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-problem-report/src/lib.rs#L274)):*

A dirty solution would be to save the cached addresses in both the resources directory and the cache directory.
</blockquote></details>

Why both? I don't think we should start changing the resource directory, it should probably be read only for everyone. But if we were, why even bother with a cache and not store it only in the resource directory?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 10 of 21 files reviewed, 1 unresolved discussion (waiting on @faern and @pinkisemils)

___
*[mullvad-problem-report/src/lib.rs, line 274 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLvxmMa8R8oPCsC5iSf:-MMFOzNw9lbsAnRyT3ff:be9br76) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-problem-report/src/lib.rs#L274)):*

Why both? I don't think we should start changing the resource directory, it should probably be read only for everyone. But if we were, why even bother with a cache and not store it only in the resource directory?
</blockquote></details>

Addressing this might require migrating the cache directory. I think that that should be done separately, not in this PR. In master, only the main address is used for the problem report tool, so the fact that the bundled list is used should still be an improvement over what we have now.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 11 of 21 files reviewed, all discussions resolved (waiting on @faern and @pinkisemils)

___
*[mullvad-problem-report/src/lib.rs, line 274 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MLvxmMa8R8oPCsC5iSf:-MMGWWe44vkKRueZFOKY:b-oyqbj6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9137f8191f7a8ec6c936ed2bd3e54fabbab326f5/mullvad-problem-report/src/lib.rs#L274)):*

Addressing this might require migrating the cache directory. I think that that should be done separately, not in this PR. In master, only the main address is used for the problem report tool, so the fact that the bundled list is used should still be an improvement over what we have now.
</blockquote></details>

Yes. Let's not make this PR scope creep. It's fine if it uses the resource addresses right now!

pinkisemils (commented):
> Reviewed 2 of 6 files at r2, 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: 14 of 21 files reviewed, all discussions resolved (waiting on @faern)

pinkisemils (approved):
> Reviewed 5 of 21 files at r1, 1 of 6 files at r2, 2 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[mullvad-rpc/src/address_cache.rs, line 181 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259#-MMKx9z36JsbgQ8aDvjg:-MMKx9z36JsbgQ8aDvjh:b-em4a2v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bfc250aa0a5dfd6dcd793d7b29534c0b2c6b72b/mullvad-rpc/src/address_cache.rs#L181)):*
> ```Rust
>         .map_err(|error| Error::ReadAddressCache(error))?
>     {
>         // for line in lines.next_line() {
> ```

This commented out line should be removed.

pinkisemils (approved):
> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2259)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2223, opened 2020-10-28
Fix apps not being excluded sometimes when starting with Auto-Connect enabled

jvff:
Previously, sometimes the app would start with Auto-Connect enabled and connect automatically before the list of excluded apps was loaded from persistent storage. After loading is complete, the `SplitTunneling` class notifies the listener (`MullvadVpnService`) which then requests a reconnect.

However, if the reconnect is requested while the app is already reconnecting, the reconnect command wouldn't get sent to the daemon. That meant that the tunnel wouldn't be recreated to exclude the loaded apps, which led to the bug.

This PR fixes that by allowing the reconnect command to be sent to the daemon even if it is already reconnecting.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header.
* [x] Check that commits follow the [Mullvad coding guidelines](https://github.com/mullvad/coding-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2223)

PR #2261, opened 2020-11-11
Fix apps still not being excluded when starting with Auto-Connect enabled

jvff:
Sometimes the app would launch and not exclude any apps from the tunnel, which meant that the split-tunneling options weren't properly applied. A [previous PR](https://github.com/mullvad/mullvadvpn-app/pull/2223) attempted to fix the cause of that issue, where a race condition caused a reconnect to be requested while already reconnecting, which meant that it would discard the second reconnect request. This caused the issue because the second reconnect was necessary in order to recreate the tunnel with the exclusion list applied.

However, the issue was still happening sometimes. Upon further investigation, there appears to be another race condition. This time, the app starts the daemon and it connects successfully, but the "connecting" state event is received a little after the split tunneling list is loaded. This meant that the app still thought it was disconnected when it received the exclusion list, and decides to ignore the reconnect request.

This PR implements a more robust solution, where the `SplitTunneling` handler class is loaded once for the service lifetime (and not once for the daemon lifetime, as it was previously), and the daemon is only started after the class is created (which means that it waits for the exclusion list to be loaded).

The original race condition can still be issue in the future, because marking the tunnel as stale and requesting a reconnect might not actually recreate the tunnel. However, since these steps are only performed for split tunneling, and the code has a different solution now, it does not represent a risk for now. Still, a proper solution should be implemented in a separate PR to avoid any issues in the future.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **A changelog entry is already present from the previous PR.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2261)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadVpnService.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadVpnService.kt
@@ -9,6 +9,7 @@ import android.os.IBinder
 import android.util.Log
 import java.io.File
 import kotlin.properties.Delegates.observable
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.Job
@@ -43,6 +44,8 @@ class MullvadVpnService : TalpidVpnService() {
     private val binder = LocalBinder()
     private val serviceNotifier = EventNotifier<ServiceInstance?>(null)
 
+    private val splitTunneling = CompletableDeferred<SplitTunneling>()
+
     private var isStopping = false
     private var shouldStop = false
 
@@ -67,7 +70,6 @@ class MullvadVpnService : TalpidVpnService() {
 
     private lateinit var keyguardManager: KeyguardManager
     private lateinit var notificationManager: ForegroundNotificationManager
-    private lateinit var splitTunneling: SplitTunneling
     private lateinit var tunnelStateUpdater: TunnelStateUpdater
 
     private var pendingAction by observable<PendingAction?>(null) { _, _, action ->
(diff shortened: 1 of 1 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2261)* status: all files reviewed, 1 unresolved discussion (waiting on @jvff)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadVpnService.kt, line 254 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2261#-MLyPiPE7out4GNMKgsk:-MLyPiPE7out4GNMKgsl:b-hx8r05) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/700ce9a5bd0de46abc3d67a25131bfdbb874be1c/android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadVpnService.kt#L254)):*
> ```Kotlin
>             disallowedApps = excludedApps
>             markTunAsStale()
>             connectionProxy.reconnect()
> ```

What would need to change to have this not require a reconnect?

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2261)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadVpnService.kt, line 254 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2261#-MLyPiPE7out4GNMKgsk:-MM0f0fQ03NIaxDqPch_:b2x7bkl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/700ce9a5bd0de46abc3d67a25131bfdbb874be1c/android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadVpnService.kt#L254)):*

What would need to change to have this not require a reconnect?
</blockquote></details>

I don't think there is a way, we have to let the daemon know that it needs to reconnect :(

PR #2266, opened 2020-11-13
Fix apt downgrade warnings when going from beta to stable

raksooo:
This PR fixes the issue with downgrade warnings when installing a stable version over a beta with `apt`. The reason for the downgrade warnings is that the version strings are compared lexographically. The correct solution to this is to use a tilde between the different components in the version field in the .deb file, e.g. `2020.7.0~beta1~dev-111111` instead of `2020.7.0-beta1-dev-111111`.

In addition to changing the separator this PR also removes the `.0` from the version string, e.g. `2020.7~beta1~dev-111111` instead of `2020.7.0~beta1~dev-111111`.

https://www.debian.org/doc/debian-policy/ch-controlfields.html#version
```
The lexical comparison is a comparison of ASCII values modified so that all the letters sort earlier than all the non-letters and so that a tilde sorts before anything, even the end of a part.
```
```
One common use of ~ is for upstream pre-releases. For example, 1.0~beta1~svn1245 sorts earlier than 1.0~beta1, which sorts earlier than 1.0.
```

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2266)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,8 @@ Line wrap the file at 100 chars.                                              Th
 - Set search domains in NetworkManager's DNS configuration, resolving issues where NetworkManager
   is used to manage DNS via systemd-resolved.
 - Parse routes more permissively and log parsing errors less verbosely.
+- Fix incorrect version string in .deb installer causing downgrade warnings when upgrading from beta
+  to stable.
 
 ### Security
 - Restore the last target state if the daemon crashes. Previously, if auto-connect and
--- a/gui/package-lock.json
+++ b/gui/package-lock.json
@@ -3201,6 +3201,14 @@
         "semver": "^5.5.0",
         "shebang-command": "^1.2.0",
         "which": "^1.2.9"
+      },
+      "dependencies": {
+        "semver": {
+          "version": "5.7.1",
+          "resolved": "https://registry.npmjs.org/semver/-/semver-5.7.1.tgz",
+          "integrity": "sha512-sauaDf/PZdVgrLTNYHRtpXa1iRiKcaebiKQ1BJdpQlWH2lCvexQdX55snPFyK7QzpudqbCI0qXFfOasHdyNDGQ==",
+          "dev": true
+        }
       }
     },
     "crypto-random-string": {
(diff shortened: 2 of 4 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2266)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[gui/tasks/distribution.js, line 261 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2266#-MM1bolB8wuNGIdwi3VA:-MM1bolC62pvCtt6fjr9:b-ntkdwb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e727ac9731cc9477c9113ab872148c3f220b466f/gui/tasks/distribution.js#L261)):*
> ```JavaScript
>   if (prerelease[0]) {
>     // Replace first '-' with a '~' since the first one is the one between 'betaN' and 'dev-hash'.
>     versionParts.push(prerelease[0].replace('-', '~'));
> ```

:lgtm: . The only strange thing I see is that the hash will influence the version number. E.g., `2020.5~beta1~dev-1234` would be considered a downgrade from `2020.5~beta1~dev-4321`, which might not make sense. I don't consider this important as it only affects dev builds.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2266)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[gui/tasks/distribution.js, line 261 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2266#-MM1bolB8wuNGIdwi3VA:-MM1fIxn1bH6664zwzoC:b64tcc6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e727ac9731cc9477c9113ab872148c3f220b466f/gui/tasks/distribution.js#L261)):*

:lgtm: . The only strange thing I see is that the hash will influence the version number. E.g., `2020.5~beta1~dev-1234` would be considered a downgrade from `2020.5~beta1~dev-4321`, which might not make sense. I don't consider this important as it only affects dev builds.
</blockquote></details>

I thought about this but unfortunately there's no solution as far as I'm aware. The first thing that came to mind was to not include the hash but that would result in apt not installing the second one at all since it is the same version. I agree about it not being important since it only affect dev builds :+1:

PR #2283, opened 2020-11-19
Avoid race condition when updating notification

jvff:
Previously, the tunnel state notification could be updated by two threads: the UI thread or a background thread that dispatches tunnel state events. This meant that there was a risk of a race condition. I'm not sure if this has happened before, because I found this issue while investigating another, but I think I might have seen it when swiping away the notification while disconnecting and then the notification reappeared shortly afterwards showing that the daemon was in the disconnected state.

This PR fixes that by dispatching tunnel state events (and settings change events, which also affect the notification because it may show or hide the action button based on the login status) solely on the UI thread.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header.
* [x] Check that commits follow the [Mullvad coding guidelines](https://github.com/mullvad/coding-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2283)

PR #2286, opened 2020-11-21
Fix race condition that led to multiple daemon instances running at the same time

jvff:
Previously, it was possible to spawn multiple daemon instances in the right scenario because of a race condition. The condition happened when the service stopped right after it started, while the daemon was still starting. When the service stopped, the daemon had not finished setting up, so the stop handling code couldn't set a shutdown command to the daemon. Because Android may reuse the same process for the service, this meant that a second start would also try to start the daemon again, and the first instance would end up leaking and conflicting with the second instance (they would fight to set up the tunnel, for example).

This PR tries to fix the race condition by refactoring the daemon lifecycle handling code. It tries to use the same approach used [previously](https://github.com/mullvad/mullvadvpn-app/pull/2283) to fix a race condition in the notification, by using an actor model to handle the events. In this case, a new `DaemonInstance` helper actor class was created, and it will always start and stop the daemon serially.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **To the user, this looks like the same issue related to the app restarting by itself, so that changelog entry is being reused.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2286)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/DaemonInstance.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/DaemonInstance.kt
@@ -41,6 +41,7 @@ class DaemonInstance(val vpnService: MullvadVpnService, val listener: (MullvadDa
         var isRunning = true
 
         prepareFiles()
+        vpnService.splitTunneling.join()
 
         while (isRunning) {
             if (!waitForCommand(channel, Command.START)) {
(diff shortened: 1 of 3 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2286)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2286)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2286)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2288, opened 2020-11-21
Fix app showing it was blocking connections when it wasn't when the VPN permission was denied

jvff:
The app can fail creating the tun device when auto-connect is enabled but the VPN permission is not granted. When that happens, the tunnel state machine proceeds to enter the error state. When entering that state, it requests a tunnel device to be created. Previously, the app did not check if the tunnel device was created or not in the Java side. This meant that the tunnel state machine would think that the creation of the tun device succeeded, and assume it was in the blocking state. However, if the VPN permission isn't granted, it will never be able to create the tun device. That meant that the app was reporting that it was blocking connections even if it wasn't.

This PR fixes that by returning a boolean flag to indicate if the tunnel is up or not, which the Rust side can then convert into an error that the tunnel state machine will recognize as a failure to block connections. The app then correctly enters the critical error state.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2288)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ Line wrap the file at 100 chars.                                              Th
 - Fix apps not being excluded from the tunnel sometimes if auto-connect was enabled.
 - Fix crash that happened sometimes when closing the app or when requesting from the notification
   or the quick-settings tile for the app to connect or disconnect.
+- Fix app showing that it was blocking connections when it wasn't when VPN permission was denied.
 
 #### Windows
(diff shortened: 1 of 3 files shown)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2288#-:-MMouN2Q-BaodkIpGuXU:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2289, opened 2020-11-21
Always show error message when VPN permission is denied

jvff:
Previously, the VPN permission was handled in two different ways. If the user asked the app to connect but then denied the VPN permission, the app would just stay in the disconnected state. If the user enabled auto-connect and restarted the app, the app would show a message saying that the VPN permission was denied.

This PR makes sure that an error message is always shown if the VPN permission is denied. That means that if the user asks the app to connect and then denies the permission, the app will still try to connect so that it ends up in the error state and shows the error message that the VPN permission wasn't granted.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2289)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,8 @@ Line wrap the file at 100 chars.                                              Th
 - Improve navigation in the app using a keyboard, so that touchless devices (like TVs) can be used
   more smoothly.
 - Run app in landscape mode on TVs.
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2289)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2290, opened 2020-11-23
Fix downgrades on Windows

dlon:
Currently, not all block filters are removed by older daemons when the app is downgraded (since they did not exist). This is a problem when downgrading from `master` to `2020.7`. In general, downgrades are not supported, but not completely breaking things is probably a good idea. So this PR causes the uninstaller to fully uninstall the app when downgrading, which at least won't break things.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2290)


pinkisemils (approved):
:lgtm:

> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2290#-:-MMp1unvBJxvoa5h4NT5:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2294, opened 2020-11-23
Fix sending problem reports on Android

jvff:
A previous PR (#2259) changed a bit how the RPC client is configured when sending problem reports. It now requires the resource directory path so that the RPC client can load the API IP address to use for connecting to the API. However, on Android, the resource directory is determined at runtime because the user can install the app on an SD card. This means that the path can only be determined from the Android `Context` object, and that must be sent from the Java side to the Rust side. The original PR used the static path specified in `mullvad_paths`, which meant that on Android sending problem reports failed because it couldn't initialize the RPC client because the resource directory path was incorrect.

This PR makes sure that the resource directory path on Android is correctly sent to the RPC client when configuring it to send a problem report.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Fixes an issue not present in any released version, so no changelog entry is necessary.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2294)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/dataproxy/MullvadProblemReport.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/dataproxy/MullvadProblemReport.kt
@@ -13,6 +13,7 @@ const val PROBLEM_REPORT_FILE = "problem_report.txt"
 
 class MullvadProblemReport {
     val logDirectory = CompletableDeferred<File>()
+    val resourcesDirectory = CompletableDeferred<File>()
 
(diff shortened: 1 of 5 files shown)


dlon (approved):
:lgtm:

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2294#-:-MMr5qm40CWNjb6lPdVA:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2296, opened 2020-11-24
Stop using NM when it uses dnsmasq

pinkisemils:
NetworkManager will just not apply our DNS config when managing DNS via dnsmasq, so the correct thing to do is to fall back to `resolvconf` and using a static file.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2296)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2296)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2222, opened 2020-10-28
Fix crashes caused by a missing call to `Service.startForeground()`

jvff:
On Android 8.0 and up, apps can only start services from the background if they request the service to be on the foreground. However, that means that the service must call `startForeground` as soon as it starts or receives a start command after already having started. Otherwise, Android kills the service for bad behavior after a timeout.

Previously the app would mostly work, since the service would spend most of the time in the foreground. However, removing the Quit button and changing the dismiss behavior increased the rate of crashes.

This PR fixes the issue by placing the foreground on the foreground for a moment every time the service is created or requested to start. After placing the notification there it just restores the notification to the correct behavior (either on the foreground or not).

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header.
* [x] Check that commits follow the [Mullvad coding guidelines](https://github.com/mullvad/coding-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2222)

PR #2297, opened 2020-11-24
Fix regression that caused a missing call to `startForeground`

jvff:
There was a previous fix (#2222) for an issue where Android would kill the service due to a missing call to `startForeground`. However, the issue reappeared after a previous PR (#2283) refactored the notification handling code. Unfortunately, the new code introduces a delay when updating the notification and also fails to run the clean-up operation of removing the notification when the actor has finished handling messages. The former happens because even though the call to `startForeground` happens on the UI thread, it has to wait until the main operation of the thread has finished (and the `Handler` starts handling scheduled tasks, including the co-routine). The latter happens because when stopping, the service closes the actor's channel but finishes executing `onDestroy` before the actor has received the closed channel signal. And since after `onDestroy` it stops executing, the scheduled tasks (including the actor's coroutine job) doesn't execute.

This PR fixes the issues first by moving the code to be executed out of the actor's coroutine task. Since the calls still happen on the UI thread, no synchronization is needed. The acknowledgement of the `startForeground` call now is guaranteed to be executed inside the `onCreate` and `onStartCommand` methods, without delays. And the clean-up is now executed inside the `onDestroy` method, also without delay.

The `acknowledgeStartForegroundService` could have sent a message to update the notification to its required state. That wouldn't have caused any issues with Android's `startForeground` check. However, that would have caused issues with the `onForeground` flag, because it would have a delay removing the notification from the foreground (if needed) and during that delay the service would ignore the quit command and consequently it wouldn't be possible to swipe away the notification (it would just reappear).

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **This PR fixes an issue after a regression, so there's already a changelog entry for that issue.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2297)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/ForegroundNotificationManager.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/ForegroundNotificationManager.kt
@@ -111,7 +111,7 @@ class ForegroundNotificationManager(
         showOnForeground()
 
         // Restore the notification to its correct state.
-        updater.sendBlocking(UpdaterMessage.UpdateNotification())
+        updateNotification()
     }
 
     private fun runUpdater() = GlobalScope.actor<UpdaterMessage>(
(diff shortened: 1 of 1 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2297)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2297#-:-MMvFFlrBZ-f0bP0xa43:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2305, opened 2020-11-25
Fix daemon startup race condition

jvff:
A recent PR (#2286) fixed a race condition that ended up spawning multiple daemons. However, that surfaced an issue where Android kills the service prematurely and quickly after spawning it. When that happens, the daemon instance ends up shutting down before the service has fully initialized the classes that use the daemon. This ends up either setting the `instance` to a shut down daemon or having the dependent classes getting `null`s when requesting data from the daemon.

This PR ensures that the `instance` property is correctly set to `null` if Android kills it prematurely and prevents the race condition of the slow daemon startup to overwrite the `instance` property when the daemon has shut down.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2305)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadVpnService.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadVpnService.kt
@@ -174,6 +174,7 @@ class MullvadVpnService : TalpidVpnService() {
         state = State.Stopped
         notificationManager.onDestroy()
         daemonInstance.onDestroy()
+        instance = null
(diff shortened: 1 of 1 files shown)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2305#-:-MN3GgBsAw7xcSCIjSUO:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2306, opened 2020-11-25
Handle `null` values returned when daemon has stopped

jvff:
A previous issue surfaced a race condition when starting and stopping the daemon too quickly. The issue is fixed by a separate PR (#2305). However, one of the symptoms was a `NullPointerException` that was thrown when using a valued returned by the stopped daemon. The daemon proxy implementation returns `null` values when it has lost the connection to the daemon. In theory this shouldn't be an issue since when the daemon stops the UI returns to the launch screen and fetches are stopped. However, this may be an issue in some cases if some data is requested from the daemon right after it stopped and before the UI changes.

This PR improves the robustness of the daemon proxy by marking the returned values as nullable and handling `null` values where appropriate.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2306)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/AccountCache.kt
@@ -108,7 +108,9 @@ class AccountCache(val daemon: MullvadDaemon, val settingsListener: SettingsList
 
     private fun fetchAccountHistory() {
         jobTracker.newBackgroundJob("fetchHistory") {
-            accountHistory = daemon.getAccountHistory()
+            daemon.getAccountHistory()?.let { history ->
+                accountHistory = history
+            }
         }
     }
 
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/ConnectionProxy.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/ConnectionProxy.kt
@@ -194,7 +194,7 @@ class ConnectionProxy(val context: Context, val daemon: MullvadDaemon) {
         val currentState = daemon.getState()
 
         synchronized(this) {
-            if (state === initialState) {
+            if (state === initialState && currentState != null) {
                 state = currentState
             }
         }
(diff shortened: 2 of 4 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2306)* status: all files reviewed, 1 unresolved discussion (waiting on @jvff)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadDaemon.kt, line 66 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2306#-MN3GwY7556Rn3RB9TsJ:-MN3GwY7556Rn3RB9TsK:bt51t7d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8075aee453fcc3134410df7d2b6783c7fce10caa/android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadDaemon.kt#L66)):*
> ```Kotlin
> 
>     fun getCurrentVersion(): String {
>         return getCurrentVersion(daemonInterfaceAddress) ?: ""
> ```

Shouldn't this be nullable? I imagine that this would be a `Result<String>` in rust, and it would be handled a bit better on the UI side.

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2306)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadDaemon.kt, line 66 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2306#-MN3GwY7556Rn3RB9TsJ:-MN3Yzyr4NGk-JGku6G1:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8075aee453fcc3134410df7d2b6783c7fce10caa/android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadDaemon.kt#L66)):*

Shouldn't this be nullable? I imagine that this would be a `Result<String>` in rust, and it would be handled a bit better on the UI side.
</blockquote></details>

Done.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2306)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2311, opened 2020-11-27
Update wireguard-go

dlon:
wireguard-go can now rely directly on wintun.dll. This will eventually allow the app to not set up a Wintun interface (and indirectly install/update the driver) in the installer, which will also make it easier to share a single interface between OpenVPN and WireGuard.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2311)

PR #2323, opened 2020-12-01
Share a Wintun adapter between OpenVPN and WireGuard

dlon:
Currently, two separate Wintun interfaces are used for OpenVPN and WireGuard, the former being set up by the installer. Due to #2311, the driver no longer needs to be pre-installed before wireguard-go is used. So with the changes in this PR, the interface is created on demand for OpenVPN as well, and the same alias (Mullvad) is used with both tunnel protocols.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)


mvd-ows (commented):
> Reviewed 14 of 23 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)* status: 14 of 23 files reviewed, 9 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/winnet.rs, line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNUwVb5BLI3XP71y4UF:-MNUwVb5BLI3XP71y4UG:b-anywo7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/030dba0070fed0371c13023ba0474beaca810f74/talpid-core/src/winnet.rs#L28)):*
> ```Rust
>     EnableIpv6,
> 
>     /// Failed to enable IPv6 on the network interface.
> ```

Comment is out-of-date.
___
*[talpid-core/src/tunnel/mod.rs, line 384 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNUxrVj9c3uqizxwW2L:-MNUxrVk03c6sD7Sd_Rc:b-sk511v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/mod.rs#L384)):*
> ```Rust
>     }
> 
>     let guid = "{AFE43773-E1F8-4EBB-8536-576AB86AFE9A}";
> ```

One of the reasons one might want to use a fixed GUID is to prevent Windows from popping the stupid "configure this as a private or public network" dialog, every time the user reconnects. Is there a reason we don't use a fixed GUID also for OpenVPN?
___
*[talpid-core/src/tunnel/openvpn/windows.rs, line 149 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV3p0Y50Z1ag5PJkQj:-MNV3p0Y50Z1ag5PJkQk:brw75cb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/openvpn/windows.rs#L149)):*
> ```Rust
>                 wintun_dll.as_ptr(),
>                 ptr::null_mut(),
>                 LOAD_WITH_ALTERED_SEARCH_PATH,
> ```

This doesn't seem to be necessary since now `wintun.dll` is in the same directory as `mullvad-daemon.exe`. Default search rules would resolve dependencies successfully.
___
*[talpid-core/src/tunnel_state_machine/connecting_state.rs, line 162 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV7B2A-FCX4aZX1Q6x:-MNV7B2A-FCX4aZX1Q6y:bkl3cgy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/030dba0070fed0371c13023ba0474beaca810f74/talpid-core/src/tunnel_state_machine/connecting_state.rs#L162)):*
I'm unfamiliar with this syntax. Is the general error type supposed to be repeated?

> ```
> tunnel::Error::OpenVpnTunnelMonitoringError
> ```
___
*[wireguard/libwg/libwg_windows.go, line 21 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9Wfu0SRGppaOlgW_:-MNV9Wfu0SRGppaOlgWa:b-9bay59) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L21)):*
> ```Go
> 	"golang.zx2c4.com/wireguard/device"
> 	"golang.zx2c4.com/wireguard/tun"
> 	"golang.zx2c4.com/wireguard/tun/wintun"
> ```

Isn't there an associated lock-file that needs to be updated as well?
___
*[wireguard/libwg/libwg_windows.go, line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9pkNByVt-ruFk3vw:-MNV9pkNByVt-ruFk3vx:b-pbsrft) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L34)):*
> ```Go
> type LogContext = unsafe.Pointer
> 
> var MullvadPool *wintun.Pool
> ```

Where is this being released? What happens if it's not cleaned up?
___
*[wireguard/libwg/libwg_windows.go, line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNVA-Lq0x1U4E-1mFPQ:-MNVA-Lq0x1U4E-1mFPR:b4n0ghp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L94)):*
> ```Go
> 
> 	if tun.WintunPool != MullvadPool {
> 		tun.WintunPool = MullvadPool
> ```

This looks suspicious. Can you create an adapter into a default pool and assign it into another pool after the fact?
___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 58 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNUylcZ5tBgscx_X9Kg:-MNUylcZ5tBgscx_X9Kh:b-ukw8gp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/openvpn/mod.rs#L58)):*
> ```Rust
>     Data2: 0xE1F8,
>     Data3: 0x4EBB,
>     Data4: [0x85, 0x36, 0x57, 0x6A, 0xB8, 0x6A, 0xFE, 0x9A],
> ```

This is repeated elsewhere (well, not in the same format). Would it make sense to gather the definitions in a single location?
___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 428 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV1GprBfUPhsGXsj_T:-MNV1GprBfUPhsGXsj_U:bkxo388) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/openvpn/mod.rs#L428)):*
> ```Rust
>             .map_err(Error::WintunError)?;
> 
>             if reboot_required {
> ```

Is there a way to suppress this?

As an end-user, I would be furious if the app stopped working when I was in the middle of important work.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)* status: 14 of 23 files reviewed, 9 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/tunnel/mod.rs, line 384 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNUxrVj9c3uqizxwW2L:-MNXu6CXBhLSDvx7uRoF:b-x7zduy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/mod.rs#L384)):*

One of the reasons one might want to use a fixed GUID is to prevent Windows from popping the stupid "configure this as a private or public network" dialog, every time the user reconnects. Is there a reason we don't use a fixed GUID also for OpenVPN?
</blockquote></details>

We do, see below.
___
*[wireguard/libwg/libwg_windows.go, line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9pkNByVt-ruFk3vw:-MNXutJBDDxobxFbnYso:b1z6qca) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L34)):*

Where is this being released? What happens if it's not cleaned up?
</blockquote></details>

The GC should take care of this.
___
*[wireguard/libwg/libwg_windows.go, line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNVA-Lq0x1U4E-1mFPQ:-MNXvJIO-ghPs1Zxu3yf:b-v8ngub) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L94)):*

This looks suspicious. Can you create an adapter into a default pool and assign it into another pool after the fact?
</blockquote></details>

No, it must be destroyed before being recreated in the other pool.
___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 428 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV1GprBfUPhsGXsj_T:-MNXuDCX355YPBtwYG-V:b-amukeh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/openvpn/mod.rs#L428)):*

Is there a way to suppress this?

As an end-user, I would be furious if the app stopped working when I was in the middle of important work.
</blockquote></details>

I doubt that this should ever actually be necessary. But if it ever is, it will probably only occur if the driver needs to be installed or updated. So I think it's not a big deal.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)* status: 14 of 23 files reviewed, 7 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/tunnel/mod.rs, line 384 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNUxrVj9c3uqizxwW2L:-MNY2U9t3b3YjcsT3q4Q:b-nfkzw7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/mod.rs#L384)):*

We do, see below.
</blockquote></details>

OK, great!
___
*[wireguard/libwg/libwg_windows.go, line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9pkNByVt-ruFk3vw:-MNY2z6s9fQu57dgxv5v:b-so0o0l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L34)):*

The GC should take care of this.
</blockquote></details>

So auto-released as the process is shutting down?
___
*[wireguard/libwg/libwg_windows.go, line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNVA-Lq0x1U4E-1mFPQ:-MNY3Muv5VoKdqchOGxc:b-y3ed4o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L94)):*

No, it must be destroyed before being recreated in the other pool.
</blockquote></details>

Right. I was reading it wrong, thought `tun` was the virtual adapter. All good then.
___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 428 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV1GprBfUPhsGXsj_T:-MNY2crJ1zGMhHYeTxIv:bpnb6g0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/openvpn/mod.rs#L428)):*

I doubt that this should ever actually be necessary. But if it ever is, it will probably only occur if the driver needs to be installed or updated. So I think it's not a big deal.
</blockquote></details>

How would the driver know if it needs to be updated? Is there phone-home logic baked into the DLL?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)* status: 12 of 23 files reviewed, 7 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/winnet.rs, line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNUwVb5BLI3XP71y4UF:-MNY8vRF694WkL3KqCH8:b-at5685) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/030dba0070fed0371c13023ba0474beaca810f74/talpid-core/src/winnet.rs#L28)):*

Comment is out-of-date.
</blockquote></details>

Was this referring to something that's been removed?
___
*[talpid-core/src/tunnel/openvpn/windows.rs, line 149 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV3p0Y50Z1ag5PJkQj:-MNYAMS6CDut-OxdcTy2:bbumf3k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/openvpn/windows.rs#L149)):*

This doesn't seem to be necessary since now `wintun.dll` is in the same directory as `mullvad-daemon.exe`. Default search rules would resolve dependencies successfully.
</blockquote></details>

I used `resource_dir` because this is consistent with how other binaries are accessed. But it probably could be simply `wintun.dll`. Which seems best?
___
*[wireguard/libwg/libwg_windows.go, line 21 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9Wfu0SRGppaOlgW_:-MNYCzRT1u8zra26WSxk:bmb5vne) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L21)):*

Isn't there an associated lock-file that needs to be updated as well?
</blockquote></details>

It is up-to-date.
___
*[wireguard/libwg/libwg_windows.go, line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9pkNByVt-ruFk3vw:-MNYDB_zC1B6sowV5v9i:b-s9c723) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L34)):*

So auto-released as the process is shutting down?
</blockquote></details>

I haven't tested it specifically, but `wireguard-go` does nothing special to clean up its pool either, so I imagine it's correct. Also, the pool object itself is essentially just a string. It doesn't call any Windows API functions or anything like that when initialized or destroyed.
___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 428 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV1GprBfUPhsGXsj_T:-MNY9-VE81t6_BzO5rAZ:b89uosl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/openvpn/mod.rs#L428)):*

How would the driver know if it needs to be updated? Is there phone-home logic baked into the DLL?
</blockquote></details>

After `SetupDiCallClassInstaller(DIF_INSTALLDEVICE, ..)` is called by this DLL, it checks whether `DI_NEEDREBOOT` is set. I don't know when, if ever, this flag is actually set.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)* status: 13 of 23 files reviewed, 3 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/winnet.rs, line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNUwVb5BLI3XP71y4UF:-MNYRmKYD4HUxwViI5ko:b6b4wrz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/030dba0070fed0371c13023ba0474beaca810f74/talpid-core/src/winnet.rs#L28)):*

Was this referring to something that's been removed?
</blockquote></details>

Possibly. Or reviewable didn't move this box when code was most recently updated.

There was a comment that had been copy-pasted and didn't match the display message on the line following it.

Doesn't seem to be an issue now.
___
*[talpid-core/src/tunnel/openvpn/windows.rs, line 149 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV3p0Y50Z1ag5PJkQj:-MNYTg8b6H-lzv_kguoF:b-b3zn20) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/openvpn/windows.rs#L149)):*

I used `resource_dir` because this is consistent with how other binaries are accessed. But it probably could be simply `wintun.dll`. Which seems best?
</blockquote></details>

I think you're trying to address two issues at once.

My understanding is that you would use `LoadLibraryExW`/`LOAD_WITH_ALTERED_SEARCH_PATH` in case you're loading a DLL that is not in the same directory as your application. AND if the DLL has non-system dynamically linked dependencies that are loaded through the import table (not using LoadLibraryXxx).

E.g. if `wintun.dll` required `wintun-helper.dll` through its import table, and they were both in some other directory separate from `mullvad-daemon.exe`. This is not the case here so a regular `LoadLibraryW` would suffice.

The second issue is how the first/primary DLL is resolved. You can side-step all issues surrounding this by always using the full path.

In summary: Good that you started using the full path of the DLL. I don't think `LoadLibraryExW` adds anything but probably doesn't hurt anything, either.
___
*[wireguard/libwg/libwg_windows.go, line 21 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9Wfu0SRGppaOlgW_:-MNYaRO1AvsazPivR3w7:btgd69b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L21)):*

It is up-to-date.
</blockquote></details>

I'm not seeing it in the diff. How can it be up-to-date after you introduced a new dependency?
___
*[wireguard/libwg/libwg_windows.go, line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9pkNByVt-ruFk3vw:-MNYa9clAAwuaIhiFDKl:bm9cmnw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L34)):*

I haven't tested it specifically, but `wireguard-go` does nothing special to clean up its pool either, so I imagine it's correct. Also, the pool object itself is essentially just a string. It doesn't call any Windows API functions or anything like that when initialized or destroyed.
</blockquote></details>

OK, nice.
___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 428 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV1GprBfUPhsGXsj_T:-MNYTOsj56mmqsEmVg5O:b-j39bb6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/openvpn/mod.rs#L428)):*

After `SetupDiCallClassInstaller(DIF_INSTALLDEVICE, ..)` is called by this DLL, it checks whether `DI_NEEDREBOOT` is set. I don't know when, if ever, this flag is actually set.
</blockquote></details>

Ah, ok. For a virtual device that is trying to load the driver, this is probably never set.

pinkisemils (changes requested):
> Reviewed 15 of 23 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)* status: all files reviewed, 4 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/tunnel_state_machine/connecting_state.rs, line 162 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV7B2A-FCX4aZX1Q6x:-MNa7WHFER3YZaBO42hj:b2kf38s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/030dba0070fed0371c13023ba0474beaca810f74/talpid-core/src/tunnel_state_machine/connecting_state.rs#L162)):*

I'm unfamiliar with this syntax. Is the general error type supposed to be repeated?

> ```
> tunnel::Error::OpenVpnTunnelMonitoringError
> ```

</blockquote></details>

It's syntax for matching on a specific type of an enum whilst also binding the value to a variable.
There's two things going on, multiple match expressions are combined into one via `|` and each expression binds the value it matches to a name via `@`.  Here's a simpler [version](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=53d0e7a5328cf1d05b21e8ea41c1e76e) of the same thing.
It's a bit cumbersome, but it allows for using a single match arm to cover both virtual adapter error cases, since in both cases the same value has to be returned. It's the equivalent of 
```rust
match error {
    error @ tunnel::Error::OpenVpnTunnelMonitoringError(tunnel::openvpn::Error::DisabledVirtualAdapter) => warn!("{}",  error.display_chain_with_msg("Virtual adapter problem detected")),
    error @ tunnel::Error::OpenVpnTunnelMonitoringError(tunnel::openvpn::Error::MissingVirtualAdapter) => warn!("{}",  error.display_chain_with_msg("Virtual adapter problem detected")),
}
```
___
*[wireguard/libwg/libwg_windows.go, line 21 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9Wfu0SRGppaOlgW_:-MNaAT5z2UJAUb9wV91n:b-q8uv9i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L21)):*

I'm not seeing it in the diff. How can it be up-to-date after you introduced a new dependency?
</blockquote></details>

It's kind of like importing `talpid-core::net::wireguard` wouldn't necessitate a new entry in `Cargo.lock`. This is just importing a package from a module that's already listed as a dependency.
___
*[wireguard/libwg/libwg_windows.go, line 40 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNaAz354syIKImgcUPH:-MNaAz354syIKImgcUPI:b66411k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/14424d029483e4540de9f840d717225d37230315/wireguard/libwg/libwg_windows.go#L40)):*
> ```Go
> 	MullvadPool, err = wintun.MakePool("Mullvad")
> 	if err != nil {
> 		panic(fmt.Errorf("Failed to make pool: %w", err))
> ```

If this panics, the  whole daemon would be taken down, right? I don't think that's acceptable, this would result in the daemon thrashing.
___
*[talpid-core/src/tunnel/openvpn.rs, line 512 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNa77GACxwlewd6VrC8:-MNa77GACxwlewd6VrC9:b-8kdf2n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cba8a22891e6871cdaca92709fb652829020582/talpid-core/src/tunnel/openvpn.rs#L512)):*
> ```Rust
>             if let Some(log_path) = self.log_path.take() {
>                 if let Ok(log) = fs::read_to_string(log_path) {
>                     if log.contains("There are no TAP-Windows adapters on this system") {
> ```

Will we not be interested in WinTun specific log entries to see what happened? I guess we'll still have logs and it's too early to tell there will be any widespread and reoccuring WinTun issues.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon, @mvd-ows, and @pinkisemils)

___
*[talpid-core/src/tunnel/openvpn/windows.rs, line 149 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV3p0Y50Z1ag5PJkQj:-MNbzEHG4pYZKfzrqS-e:btdsfze) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/talpid-core/src/tunnel/openvpn/windows.rs#L149)):*

I think you're trying to address two issues at once.

My understanding is that you would use `LoadLibraryExW`/`LOAD_WITH_ALTERED_SEARCH_PATH` in case you're loading a DLL that is not in the same directory as your application. AND if the DLL has non-system dynamically linked dependencies that are loaded through the import table (not using LoadLibraryXxx).

E.g. if `wintun.dll` required `wintun-helper.dll` through its import table, and they were both in some other directory separate from `mullvad-daemon.exe`. This is not the case here so a regular `LoadLibraryW` would suffice.

The second issue is how the first/primary DLL is resolved. You can side-step all issues surrounding this by always using the full path.

In summary: Good that you started using the full path of the DLL. I don't think `LoadLibraryExW` adds anything but probably doesn't hurt anything, either.
</blockquote></details>

I see!
___
*[wireguard/libwg/libwg_windows.go, line 40 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNaAz354syIKImgcUPH:-MNbx5MW8ihgWcKZ2Okj:b-v0ye4l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/14424d029483e4540de9f840d717225d37230315/wireguard/libwg/libwg_windows.go#L40)):*

If this panics, the  whole daemon would be taken down, right? I don't think that's acceptable, this would result in the daemon thrashing.
</blockquote></details>

It takes down the daemon but the service will fail to start in the first place. It won't keep being restarted.

This really should never fail, though. It only fails if the name is too long, or if it cannot be converted to a UTF-16 string. And I'm not sure how one could properly handle those errors.

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[talpid-core/src/tunnel_state_machine/connecting_state.rs, line 162 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV7B2A-FCX4aZX1Q6x:-MNbz4QFDR6SLSRX10NH:b76sydm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/030dba0070fed0371c13023ba0474beaca810f74/talpid-core/src/tunnel_state_machine/connecting_state.rs#L162)):*

It's syntax for matching on a specific type of an enum whilst also binding the value to a variable.
There's two things going on, multiple match expressions are combined into one via `|` and each expression binds the value it matches to a name via `@`.  Here's a simpler [version](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=53d0e7a5328cf1d05b21e8ea41c1e76e) of the same thing.
It's a bit cumbersome, but it allows for using a single match arm to cover both virtual adapter error cases, since in both cases the same value has to be returned. It's the equivalent of 
```rust
match error {
    error @ tunnel::Error::OpenVpnTunnelMonitoringError(tunnel::openvpn::Error::DisabledVirtualAdapter) => warn!("{}",  error.display_chain_with_msg("Virtual adapter problem detected")),
    error @ tunnel::Error::OpenVpnTunnelMonitoringError(tunnel::openvpn::Error::MissingVirtualAdapter) => warn!("{}",  error.display_chain_with_msg("Virtual adapter problem detected")),
}
```

</blockquote></details>

Excellent, thank you.
___
*[wireguard/libwg/libwg_windows.go, line 21 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9Wfu0SRGppaOlgW_:-MNbzRFF0rc2y9AL0bpK:buacnax) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L21)):*

It's kind of like importing `talpid-core::net::wireguard` wouldn't necessitate a new entry in `Cargo.lock`. This is just importing a package from a module that's already listed as a dependency.
</blockquote></details>

Ah, right. Thanks. However, wouldn't the existing dependency still need to be updated? Since previously it didn't support adapter pools, and now it does.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[wireguard/libwg/libwg_windows.go, line 40 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNaAz354syIKImgcUPH:-MNc6NVW7MsNOiuZVGA4:bjvwj09) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/14424d029483e4540de9f840d717225d37230315/wireguard/libwg/libwg_windows.go#L40)):*

It takes down the daemon but the service will fail to start in the first place. It won't keep being restarted.

This really should never fail, though. It only fails if the name is too long, or if it cannot be converted to a UTF-16 string. And I'm not sure how one could properly handle those errors.
</blockquote></details>

Also, the default "WireGuard" pool is set up in the exact same way.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[wireguard/libwg/libwg_windows.go, line 21 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9Wfu0SRGppaOlgW_:-MNc6YLi3h0fhlb7grQR:bidfkp8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L21)):*

Ah, right. Thanks. However, wouldn't the existing dependency still need to be updated? Since previously it didn't support adapter pools, and now it does.
</blockquote></details>

Checking master, the dependencies were [updated](https://github.com/mullvad/mullvadvpn-app/blame/master/wireguard/libwg/go.mod) 6 days ago. And if the code required newer dependencies, it would fail in CI, since we do build this in CI.
___
*[wireguard/libwg/libwg_windows.go, line 40 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNaAz354syIKImgcUPH:-MNc5lbi86CtbDNCQi8h:bfuk46j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/14424d029483e4540de9f840d717225d37230315/wireguard/libwg/libwg_windows.go#L40)):*

Also, the default "WireGuard" pool is set up in the exact same way.
</blockquote></details>

Ah, yes. that's fair. Then this can be left as is.

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[wireguard/libwg/libwg_windows.go, line 21 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2323#-MNV9Wfu0SRGppaOlgW_:-MNc8WCWDdfecFqiNxbh:b-40vfjd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b260fbebb06f52ef8533321b30413da4c66e1e2f/wireguard/libwg/libwg_windows.go#L21)):*

Checking master, the dependencies were [updated](https://github.com/mullvad/mullvadvpn-app/blame/master/wireguard/libwg/go.mod) 6 days ago. And if the code required newer dependencies, it would fail in CI, since we do build this in CI.
</blockquote></details>

Great, that's the update I thought would be included here.

PR #2324, opened 2020-12-01
Refactor `LocationInfoCache` into actor

jvff:
Previously, the `LocationInfoCache` had some slightly complex logic to make sure only one fetch was running and to make sure that fetches were serialized. This PR simplifies that by refactoring the class to use an actor to fetch the location.

The fetches now happen entirely in the background thread, where previously the management part happened in the UI thread. Running parts of it in the UI thread was unnecessary, because the only receiver of location events (the `ConnectFragment`) was already spawning a UI thread coroutine to handle the location updates.

The logic for fetching should stay the same. The actor waits for a request to fetch, and then retries the fetch with an exponential back-off until a valid location is fetched. Cancelling the fetch was removed, but since cancelling is a cooperative action (i.e., suspend methods must check for a cancellation event and handle it accordingly) and the daemon side of the fetch never cancelled the underlying Rust-side future, the behavior still stays the same.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2324)

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -75,6 +75,7 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
         allWarningsAsErrors = true
 
         kotlinOptions.freeCompilerArgs += [
+            "-Xuse-experimental=kotlinx.coroutines.ExperimentalCoroutinesApi",
             "-Xuse-experimental=kotlinx.coroutines.ObsoleteCoroutinesApi",
         ]
     }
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2324)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2327, opened 2020-12-02
Upgrade futures-util and futures-task

faern:
These crates had security advisories posted against them for the versions we were using.

https://rustsec.org/advisories/RUSTSEC-2020-0059.html
https://rustsec.org/advisories/RUSTSEC-2020-0060.html

I have not actually tested this. Since they are patch upgrades and only a few crates I really hope nothing breaks. We will do more testing before releasing very soon anyway. And there are not many alternatives to merging this either. Since it's a potential security issue.

I don't know if we use the vulnerable code or not. But even if we don't some dependency of ours might and that's a huge amount of work to check. Better to just upgrade upgrade upgrade.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2327)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -643,9 +643,9 @@ dependencies = [
 
 [[package]]
 name = "futures-channel"
-version = "0.3.5"
+version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f366ad74c28cca6ba456d95e6422883cfb4b252a83bed929c83abfdbbf2967d5"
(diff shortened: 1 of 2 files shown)


pinkisemils (commented):
Those are some weird build errors, but :lgtm: 
We definitely don't use any of the affected types directly, but it's probably used by some of the libraries :/

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2327#-:-MNa4oea4sLPadSs7n3s:b-wt8laj)* status: 0 of 1 files reviewed, all discussions resolved

PR #2343, opened 2020-12-10
Warn if configured to use a public custom DNS server

jvff:
Our servers intercept and redirect DNS requests to prevent DNS leaks. This means that configuring a custom DNS server address that needs to be routed through the tunnel doesn't work as expected. Until this is fixed server side, the app will show a dialog warning the user about the issue.

This PR implements that dialog. It uses the `MainActivity` class to synchronize the status between the fragments, which is not ideal, but this code should be removed when things are updated in the servers.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Part of an unreleased feature, so this just reuses the same changelog entry.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/AdvancedFragment.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/AdvancedFragment.kt
@@ -5,6 +5,7 @@ import android.support.v7.widget.LinearLayoutManager
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import kotlinx.coroutines.CompletableDeferred
 import net.mullvad.mullvadvpn.R
 import net.mullvad.mullvadvpn.model.Settings
 import net.mullvad.mullvadvpn.ui.customdns.CustomDnsAdapter
@@ -115,6 +116,22 @@ class AdvancedFragment : ServiceDependentFragment(OnNoService.GoBack) {
             if (!wireguardMtuInput.hasFocus) {
                 wireguardMtuInput.value = settings.tunnelOptions.wireguard.mtu
             }
+        }
+    }
+
+    private fun showConfirmPublicDnsServerDialog(confirmation: CompletableDeferred<Boolean>) {
+        val transaction = fragmentManager?.beginTransaction()
+
+        detachBackButtonHandler()
+        transaction?.addToBackStack(null)
+
+        ConfirmPublicDnsDialogFragment()
+            .apply { confirmPublicDns = confirmation }
+            .show(transaction, null)
+
+        jobTracker.newUiJob("restoreBackButtonHandler") {
+            confirmation.await()
+            attachBackButtonHandler()
         }
     }
 
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/AdvancedFragment.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/AdvancedFragment.kt
@@ -36,7 +36,11 @@ class AdvancedFragment : ServiceDependentFragment(OnNoService.GoBack) {
 
         titleController = CollapsibleTitleController(view, R.id.contents)
 
-        customDnsAdapter = CustomDnsAdapter(customDns)
+        customDnsAdapter = CustomDnsAdapter(customDns).apply {
+            showPublicDnsAddressWarning = { confirmation ->
+                showConfirmPublicDnsServerDialog(confirmation)
+            }
+        }
 
         view.findViewById<CustomRecyclerView>(R.id.contents).apply {
             layoutManager = LinearLayoutManager(parentActivity)
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/ConfirmPublicDnsDialogFragment.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/ConfirmPublicDnsDialogFragment.kt
@@ -1,0 +1,66 @@
+package net.mullvad.mullvadvpn.ui
+
+import android.app.Dialog
+import android.content.DialogInterface
+import android.graphics.drawable.ColorDrawable
+import android.os.Bundle
+import android.support.v4.app.DialogFragment
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.view.ViewGroup.LayoutParams
+import android.widget.Button
+import kotlinx.coroutines.CompletableDeferred
+import net.mullvad.mullvadvpn.R
+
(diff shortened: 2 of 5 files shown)


pinkisemils (changes requested):
> Reviewed 3 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343)* status: 3 of 6 files reviewed, 1 unresolved discussion (waiting on @jvff)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt, line 82 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343#-MOCisYe-fcc2Hg6qUCJ:-MOCisYe-fcc2Hg6qUCK:b15qkbo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2c4322a9316c9819ed2931774c2e9ec47cd07bc/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt#L82)):*
> ```Kotlin
> 
>     var backButtonHandler: (() -> Boolean)? = null
>     var confirmPublicDnsServer: CompletableDeferred<Boolean>? = null
> ```

Does this field need to exist on the parent activity if it's only ever set here?

pinkisemils (commented):
> Reviewed 3 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343)* status: all files reviewed, 1 unresolved discussion (waiting on @jvff)

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt, line 82 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343#-MOCisYe-fcc2Hg6qUCJ:-MOCvE9tEzCnfZVFJ6WQ:bfvw79c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2c4322a9316c9819ed2931774c2e9ec47cd07bc/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt#L82)):*

Does this field need to exist on the parent activity if it's only ever set here? 
</blockquote></details>

The problem is communication between fragments :/ The dialog needs a way to send the result to the Advanced screen, and Android recommends using the owner activity for that. This is ugly indeed, but I'm strongly hoping this is temporary. If we still have this line in a few months I think we should definitely try to improve this, but we'd have to handle Android destroying/recreating the fragment, which might make it more complex :/

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343)* status: all files reviewed, 1 unresolved discussion (waiting on @jvff)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt, line 82 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343#-MOCisYe-fcc2Hg6qUCJ:-MOEB8DS7nWABUn89S0C:b4bvj8q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2c4322a9316c9819ed2931774c2e9ec47cd07bc/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt#L82)):*

The problem is communication between fragments :/ The dialog needs a way to send the result to the Advanced screen, and Android recommends using the owner activity for that. This is ugly indeed, but I'm strongly hoping this is temporary. If we still have this line in a few months I think we should definitely try to improve this, but we'd have to handle Android destroying/recreating the fragment, which might make it more complex :/
</blockquote></details>

Given that we often get to back to code that's written recently to fix bugs rather than design, I think we'd do good to at least get the design good for new code - it's highly unlikely we'll have plenty of spare time to redo this code.

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt, line 82 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343#-MOCisYe-fcc2Hg6qUCJ:-MOI-YfT9DIlr5_-S1GK:b-ow4mm3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2c4322a9316c9819ed2931774c2e9ec47cd07bc/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt#L82)):*

Given that we often get to back to code that's written recently to fix bugs rather than design, I think we'd do good to at least get the design good for new code - it's highly unlikely we'll have plenty of spare time to redo this code.
</blockquote></details>

Done.

Yeah, that makes sense. We don't know how long this code will stay there, and it's going to be unpleasant to work around it while it does.

I refactored to have the `AdvancedFragment` set the property in the fragment directly. The problem of this is that the property's value will be lost when the dialog fragment is destroy (for example when the configuration changes due to language change or starting split screen mode). The new code handles that by simply dismissing the dialog and returning to the previous screen, which will still be in the editing state. That means that the user can just try to add the DNS server again and the dialog will appear again.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt, line 82 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2343#-MOCisYe-fcc2Hg6qUCJ:-MOJHVoH-KAOkCQZ94OL:bz494pl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2c4322a9316c9819ed2931774c2e9ec47cd07bc/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt#L82)):*

Done.

Yeah, that makes sense. We don't know how long this code will stay there, and it's going to be unpleasant to work around it while it does.

I refactored to have the `AdvancedFragment` set the property in the fragment directly. The problem of this is that the property's value will be lost when the dialog fragment is destroy (for example when the configuration changes due to language change or starting split screen mode). The new code handles that by simply dismissing the dialog and returning to the previous screen, which will still be in the editing state. That means that the user can just try to add the DNS server again and the dialog will appear again.
</blockquote></details>

Great, this is a much nicer solution!

PR #2317, opened 2020-11-28
Implement initial View App Logs screen

jvff:
On Android, there is no easy way to share the collected logs included in the problem report. On desktop, the "View App Logs" button opens a text editor to view the collected logs. On Android, files can't be shared between apps easily, and require the implementation of a File Provider. Therefore, the Android app has been without the ability to view the collected logs until now.

Recently, debugging an issue turned out problematic because the issue caused problem reports to fail to send. Therefore, the process was more complicated than it needed to be, and a way for the reported to view the logs would have saved some time. Therefore, this PR implements a simple screen to load the collected logs and show them to the user. The text is selectable, so the user can copy it all (or parts of it) to send manually if sending a problem report through the app fails.

The screen in this initial version is pretty raw, and it could be improved to have an animation while it's loading, a button to share the text (or the file using a File Provider), etc.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header.
* [x] Check that commits follow the [Mullvad coding guidelines](https://github.com/mullvad/coding-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2317)

PR #2349, opened 2020-12-11
Refactor problem report into an actor

jvff:
A previous PR (#2317) implemented the ability to view the collected logs in a separate screen. As part of that PR, a line was added to set the `collectJob` to `null` when deleting the problem report file. However, the method to delete the file was called by the `collectJob`, which meant that the reference to the job would be overwritten while it was executing, preventing further usages from waiting for it to complete. This broke sending problem reports.

To fix this issue and to avoid similar issues in the future, the `MullvadProblemReport` class was refactor to remove juggling all the coroutine jobs and just have one that is implemented using the actor model. This simplifies the logic, makes the behavior more predictable and easier to maintain.

The PR also moves the native JNI method implementations in Rust to a new module, to make them easier to find and to reduce the size of the main `lib.rs` file.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Fixes a bug not present in any released version, no changelog entry necessary.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2349)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/dataproxy/MullvadProblemReport.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/dataproxy/MullvadProblemReport.kt
@@ -5,112 +5,126 @@ import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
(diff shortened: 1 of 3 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2349)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2353, opened 2020-12-14
Add new systemd-resolved resolv.conf path

pinkisemils:
`systemd-resolved` has a new stub file path - `/usr/lib/systemd/resolv.conf`. But this one is non-dynamic, it should never be written to. Since `/etc/resolv.conf` can be symlinked to it, it's vulnerable to being overwritten by other software. This means that just checking if `/etc/resolv.conf` is symlinked to it is not enough, the contents have to be verified too.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2353)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@ Line wrap the file at 100 chars.                                              Th
 #### Android
 - Allow to configure the tunnel to use custom DNS servers.
 
+#### Linux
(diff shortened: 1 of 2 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2353)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2389, opened 2021-01-13
Fix permissions on the OpenVPN plugin pipe

dlon:
#1830 fixed an issue where the management interface could sometimes be accessed anonymously in a LAN. The issue was fixed by explicitly setting the security descriptor on the pipe. `talpid-openvpn-plugin` had the same issue, but it wasn't considered critical. For instance, it should be possible to send a fake `RoutePredown` event, which will force the daemon to reconnect. But this is normally not possible unless passwordless sharing is enabled.

Updating `parity-tokio-ipc` to 0.8 fixes both issues, and also means that the code for changing the security identifiers can be removed from `mullvad-management-interface`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2389)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2389#-:-MR-WAFO4t2VT3B49iH_:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2410, opened 2021-01-22
Fix daemon cleanup on Android

jvff:
@axti noticed that the `MullvadDaemon.finalize` method was incorrect, and therefore the daemon command sender endpoint was never closed correctly. Fixing the finalizer method is possible, but it leads to linter warnings.

This PR takes a different approach to fix the issue. Since the `MullvadDaemon` instance is only handled by the `DaemonInstance` actor, it can call the destructor once the daemon stops. Therefore, a new `onDestory` destructor method is created and called from `DaemonInstance`. This new destructor also cleans up all callbacks that were configured on the daemon.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Fixes a bug that's very likely not visible to the user.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2410)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/DaemonInstance.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/DaemonInstance.kt
@@ -21,7 +21,8 @@ class DaemonInstance(val vpnService: MullvadVpnService, val listener: (MullvadDa
 
     private val commandChannel = spawnActor()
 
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2410)* status: :shipit: complete! all files reviewed, all discussions resolved

axti (approved):

PR #2418, opened 2021-01-26
Improve error handling in IPC and CLI

dlon:
Main changes:
* Return errors from `mullvad-daemon` commands whenever it makes sense. Currently, channels are usually dropped whenever an error occurs, so there is no way to tell what went wrong. A lot of cases are then simply mapped to "internal error" in gRPC.
* In the CLI, state the RPC that failed when more than one thing can go wrong. E.g., show "failed to fetch account data" instead of only some REST error.
* Improve presentation in CLI. The implementation of `Display` returns an unfriendly string for tonic statuses, and some messages are misleading or incorrect. (E.g., "Failed to communicate with mullvad-daemon over RPC" may be shown when the gRPC server returns an error.)
    For example, this was shown when `mullvad account get` is run and the API cannot be reached:
    ```
    mullvad account get
    Mullvad account: XXXXXXXXXXXXXXXXX
    Error: Failed to communicate with mullvad-daemon over RPC
    Caused by: status: Internal, message: "internal error", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Fri, 04 Dec 2020 09:18:25 GMT"} }
    ```
    Now this is displayed instead:
    ```
    mullvad account get
    Mullvad account: 8941815615653
    RPC failed: Failed to fetch account data
    Caused by: Unavailable: Cannot reach the API
    ```

Error messages are still user-facing. This is a drawback over using codes since it makes localization more difficult. However, the strings are only used by the CLI, which is not localized anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2418)

### DIFF
--- a/mullvad-cli/src/cmds/account.rs
+++ b/mullvad-cli/src/cmds/account.rs
@@ -91,7 +91,8 @@ impl Account {
             println!("Mullvad account: {}", settings.account_token);
             let expiry = rpc
                 .get_account_data(settings.account_token)
-                .await?
+                .await
+                .map_err(|error| Error::RpcFailed("Failed to fetch account data", error))?
                 .into_inner();
             println!(
                 "Expires at     : {}",
--- a/mullvad-cli/src/cmds/account.rs
+++ b/mullvad-cli/src/cmds/account.rs
@@ -92,7 +92,7 @@ impl Account {
             let expiry = rpc
                 .get_account_data(settings.account_token)
                 .await
-                .map_err(|error| Error::RpcFailed("Failed to fetch account data", error))?
+                .map_err(|error| Error::RpcFailedExt("Failed to fetch account data", error))?
                 .into_inner();
             println!(
                 "Expires at     : {}",
@@ -133,7 +133,7 @@ impl Account {
                     Code::NotFound | Code::ResourceExhausted => {
                         eprintln!("Failed to submit voucher: {}", err.message());
                     }
-                    _ => return Err(Error::GrpcClientError(err)),
+                    _ => return Err(Error::RpcFailed(err)),
                 }
                 std::process::exit(1);
             }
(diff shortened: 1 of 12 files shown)


pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2418)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2418)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[mullvad-daemon/src/lib.rs, line 201 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2418#-MS1zTc25Re8rHmgLozP:-MS2Kux64xAyX8rpX93n:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/73cb94c0600f1ffb1606b9d8c239289f51a99446/mullvad-daemon/src/lib.rs#L201)):*

Since this is slowly getting to be a handful, maybe we should have a type alias for the sender - `type ResponseTx<T, E> = oneshot::Sender<Result<T, E>>`. But maybe we're not there yet.
</blockquote></details>

Done.
___
*[mullvad-daemon/src/lib.rs, line 1432 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2418#-MS2-BxT1y4hV_ogCOoy:-MS2Ky0Q8n8Fboh4p0th:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/73cb94c0600f1ffb1606b9d8c239289f51a99446/mullvad-daemon/src/lib.rs#L1432)):*

Should be `submit_voucher response`. 
</blockquote></details>

Done.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2418)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2421, opened 2021-01-28
Use a shorter version string for Windows metadata

pinkisemils:
To have less variation in our submitted version strings, the way Windows versions are reported have been changed:
- the problem report will only include the major version (10, 7, 8)
- version check version header will include the major and minor versions and the build number like so `Windows 10.0 (18363)`
Maybe the minor version should be omitted, but then our ability to be more surprised about Windows 8 (as opposed to 8.1) users would be significantly impaired.

The logic to pull these numbers out of the OS already existed in the DNS code for Windows - to have the same code be used everywhere it was refactored to be put in the platform metadata crate. To allow the use of the platform metadata crate everywhere, it was renamed from `mullvad-platform-metadata` to `talpid-platform-metadata`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)


faern (changes requested):
> Reviewed 15 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: 15 of 16 files reviewed, 6 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-core/Cargo.toml, line 75 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBi9I80mbd-abygUpI:-MSBi9I93HEsuMZdvz0t:b-kumyeu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-core/Cargo.toml#L75)):*
> ```TOML
> widestring = "0.4"
> winreg = { version = "0.7", features = ["transactions"] }
> winapi = { version = "0.3.6", features = [ "combaseapi", "handleapi", "ifdef", "libloaderapi", "netioapi", "stringapiset", "synchapi", "winbase", "winuser"] }
> ```

Why does `talpid-core` need *more* features? Looks like we are just using less `winapi` code with this PR
___
*[talpid-platform-metadata/src/windows.rs, line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBjXoK3zZD1-ihj3cR:-MSBjXoK3zZD1-ihj3cS:b6eeiy6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L12)):*
> ```Rust
>         .unwrap_or_else(|_| ("N/A".to_string(), "N/A".to_string(), "N/A".to_string()));
> 
>     format!("Windows {}.{} ({})", major, minor, build)
> ```

Nit. The current problem report format is along these lines: `Windows 10 Home (10.0.18363 N/A Build 18363)`. With that in mind I think it would make sense to format this as either `{}.{}.{}` or `{}.{} Build {}`. as those seems to be slightly more Microsoft approved formats (their own tools print it) than the current proposal. The former is shorter and the latter is more readable.
___
*[talpid-platform-metadata/src/windows.rs, line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBnB4u-NK5HPF3QmvT:-MSBnB4u-NK5HPF3QmvU:bp6c1l1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L12)):*
> ```Rust
>         .unwrap_or_else(|_| ("N/A".to_string(), "N/A".to_string(), "N/A".to_string()));
> 
>     format!("Windows {}.{} ({})", major, minor, build)
> ```

I'm thinking it could be useful for debugging purposes to know if it's `Home` or `Pro`? Maybe not. Let's discuss it with support.
___
*[talpid-platform-metadata/src/windows.rs, line 19 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBj8BhBt5r1Gge5FT-:-MSBj8BhBt5r1Gge5FT0:bs9rzl8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L19)):*
> ```Rust
>         .map(|version| version.major_version().to_string())
>         .unwrap_or("N/A".into());
>     format!("Windows {}", major_version)
> ```

For the sake of being able to deprecate support for older OSes based on these stats, I think it sounds reasonable to be able to differentiate between 8 and 8.1 even in the short version? I'm not sure, but I think those two versions are different enough that we could theoretically deprecate support for 8 while still supporting 8.1?

Potentially we can special case it so only 8 has the minor included as it's not really used for 7 or 10 (yet?)
___
*[talpid-platform-metadata/src/windows.rs, line 52 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBkoY6B4txlSILj27o:-MSBkoY6B4txlSILj27p:b8vjca8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L52)):*
> ```Rust
>         }
> 
>         rtl_get_version = unsafe { mem::transmute(function_address) };
> ```

Does the variable really need to be declared at the function start and not here? I know the code is just copy pasted. This looks like something from C that does not need to be like this in Rust.
___
*[talpid-platform-metadata/src/windows.rs, line 54 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBlFex0YGqvUvEHeAm:-MSBlFex0YGqvUvEHeAn:b-k5kv2k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L54)):*
> ```Rust
>         rtl_get_version = unsafe { mem::transmute(function_address) };
> 
>         let mut version_info: RTL_OSVERSIONINFOW = unsafe { std::mem::zeroed() };
> ```

This probably is pretty safe in practice, but the more correct way to handle this is via `MaybeUninit`. To not let Rust treat it as an `RTL_OSVERSIONINFOW` until after it has been properly initialized. It would not hurt to move over to that.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: 15 of 16 files reviewed, 6 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-platform-metadata/src/windows.rs, line 52 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBkoY6B4txlSILj27o:-MSBsN333rrTqSyWIgon:b-h18r8g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L52)):*

Does the variable really need to be declared at the function start and not here? I know the code is just copy pasted. This looks like something from C that does not need to be like this in Rust.
</blockquote></details>

A `mem::transmute` that fully relies on type inference looks extremely scary. Transmutes are nasty to begin with, and when you can't see neither the input nor output types it's a bit too much for my taste :joy:.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: 15 of 16 files reviewed, 5 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-platform-metadata/src/windows.rs, line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBnB4u-NK5HPF3QmvT:-MSBygIo-Hv2aQj_u1pR:bk1qw6w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L12)):*

I'm thinking it could be useful for debugging purposes to know if it's `Home` or `Pro`? Maybe not. Let's discuss it with support.
</blockquote></details>

Resolved on slack. Closing. We don't need it.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: 15 of 16 files reviewed, 4 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-platform-metadata/src/windows.rs, line 19 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBj8BhBt5r1Gge5FT-:-MSC-8fP1BKzhuA5rhvA:btbsybh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L19)):*

For the sake of being able to deprecate support for older OSes based on these stats, I think it sounds reasonable to be able to differentiate between 8 and 8.1 even in the short version? I'm not sure, but I think those two versions are different enough that we could theoretically deprecate support for 8 while still supporting 8.1?

Potentially we can special case it so only 8 has the minor included as it's not really used for 7 or 10 (yet?)
</blockquote></details>

Again, sorted on slack. Maybe the minor is not that important to know after all :)

dlon (approved):
> Reviewed 1 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: all files reviewed, 5 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-platform-metadata/src/windows.rs, line 35 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSCWc3h7Qp2DtU1DwMn:-MSCWc3h7Qp2DtU1DwMo:b-vx3kr0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L35)):*
> ```Rust
>         let rtl_get_version: extern "stdcall" fn(*mut RTL_OSVERSIONINFOW);
> 
>         let module_name: Vec<u16> = OsString::from("ntdll")
> ```

This could be replaced with `OsStr::new("ntdll")`, right?

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: all files reviewed, 5 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-platform-metadata/src/windows.rs, line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSCfupM-GAnbxSVhyYk:-MSCfupNBCfcDvI54pgp:b-h2j0f4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L37)):*
> ```Rust
>         let module_name: Vec<u16> = OsString::from("ntdll")
>             .as_os_str()
>             .encode_wide()
> ```

There are some crates (e.g. [utf16_lit](https://docs.rs/utf16_lit/1.0.1/)) that provide macros for generating UTF-16 literals. Maybe this thing can be replaced with `utf16_null!("ntdll")` to avoid the runtime conversion.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: 13 of 16 files reviewed, 4 unresolved discussions (waiting on @faern)

___
*[talpid-core/Cargo.toml, line 75 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBi9I80mbd-abygUpI:-MSCk1y2E-VFdl4_xtdh:bd52pn9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-core/Cargo.toml#L75)):*

Why does `talpid-core` need *more* features? Looks like we are just using less `winapi` code with this PR
</blockquote></details>

It doesn't even build on `master` for me. Something else uses that feature down the line so the builds don't fail.
___
*[talpid-platform-metadata/src/windows.rs, line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBjXoK3zZD1-ihj3cR:-MSD0BuJ0p7xQERWv0EK:bi6no9c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L12)):*

Nit. The current problem report format is along these lines: `Windows 10 Home (10.0.18363 N/A Build 18363)`. With that in mind I think it would make sense to format this as either `{}.{}.{}` or `{}.{} Build {}`. as those seems to be slightly more Microsoft approved formats (their own tools print it) than the current proposal. The former is shorter and the latter is more readable.
</blockquote></details>

I've removed the minor version since we've concluded that it isn't useful at all.
___
*[talpid-platform-metadata/src/windows.rs, line 52 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBkoY6B4txlSILj27o:-MSCkwaCA16HQOTbp_gV:b-g8l73c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L52)):*

A `mem::transmute` that fully relies on type inference looks extremely scary. Transmutes are nasty to begin with, and when you can't see neither the input nor output types it's a bit too much for my taste :joy:.
</blockquote></details>

It's not type inference if the types are declared. I can move the declaration of that variable to here though. The transmutation is required because I didn't find a way to cast a void pointer to a function pointer.
___
*[talpid-platform-metadata/src/windows.rs, line 54 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBlFex0YGqvUvEHeAm:-MSCoANl7giV9aDiQp2t:bxzxwfe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L54)):*

This probably is pretty safe in practice, but the more correct way to handle this is via `MaybeUninit`. To not let Rust treat it as an `RTL_OSVERSIONINFOW` until after it has been properly initialized. It would not hurt to move over to that.
</blockquote></details>

The benefits are not really there given that there is no drop implementation for `RTL_OSVERSIONFOW`, and it's perfectly safe to access all of it's fields when un-initialized. In the end, more unsafe functions need to be used than in the non-MaybeUninit version.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: 13 of 16 files reviewed, 4 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[talpid-platform-metadata/src/windows.rs, line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSCfupM-GAnbxSVhyYk:-MSD593DEziF5mJGgIFe:b39o0yc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L37)):*

There are some crates (e.g. [utf16_lit](https://docs.rs/utf16_lit/1.0.1/)) that provide macros for generating UTF-16 literals. Maybe this thing can be replaced with `utf16_null!("ntdll")` to avoid the runtime conversion.
</blockquote></details>

I don't think we should do this for performance sake at runtime. That won't really be measurable. If anything, do it for code simplicity. But IMO only worth it if we can do it in many places. I would not pull in a dependency for replacing a single place.
___
*[talpid-platform-metadata/src/windows.rs, line 52 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBkoY6B4txlSILj27o:-MSD6a7GAUWXXJD8FbXz:bkq2714) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L52)):*

It's not type inference if the types are declared. I can move the declaration of that variable to here though. The transmutation is required because I didn't find a way to cast a void pointer to a function pointer.
</blockquote></details>

It's inferring the types of the `T` and `U` generics on `transmute`, no? Not very important. The point is to keep the type closely tied to the transmute so it's easier to see and harder to make a mistake.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: 13 of 16 files reviewed, 4 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[talpid-platform-metadata/src/windows.rs, line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBjXoK3zZD1-ihj3cR:-MSDBOZYC4Vtt1mGWcDr:b-lj0kmy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L12)):*

I've removed the minor version since we've concluded that it isn't useful at all.
</blockquote></details>

This can still be useful for problem reports. Please add it back as discussed on slack. Only exclude it from the short version.
___
*[talpid-platform-metadata/src/windows.rs, line 54 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSBlFex0YGqvUvEHeAm:-MSDAs1C8uOnLchiFNsp:bx714pc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L54)):*

The benefits are not really there given that there is no drop implementation for `RTL_OSVERSIONFOW`, and it's perfectly safe to access all of it's fields when un-initialized. In the end, more unsafe functions need to be used than in the non-MaybeUninit version.
</blockquote></details>

Yeah since you do `zeroed()` it should work fine. But that's different from uninitialized however. Not even integers are safe to handle uninitialized as the rules around them are not finalized and it should be considered UB. This code is fine.
___
*[talpid-platform-metadata/src/windows.rs, line 61 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSD9vvs4r7LEeo61vHi:-MSD9vvt6rN6E5-q4s7L:brczmsm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c628e5a35d2ac00ec3a79d21dd84b316f19e70f/talpid-platform-metadata/src/windows.rs#L61)):*
> ```Rust
>         unsafe {
>             (*version_info.as_mut_ptr()).dwOSVersionInfoSize =
>                 mem::size_of_val(&version_info) as u32;
> ```

Maybe a bit nitpicky. But I view `size_of_val` as harder to use correctly than `size_of`. Since If you did `&&version_info` or if `version_info` happened to be a reference already then this would return the size of a pointer without any compiler warnings. The docs of `size_of_val` states that it's mostly for the dynamic use case when `size_of` does not work.

dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: 13 of 16 files reviewed, 5 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[talpid-platform-metadata/src/windows.rs, line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSCfupM-GAnbxSVhyYk:-MSD9wwp23LypADegK8j:b2igsva) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L37)):*

I don't think we should do this for performance sake at runtime. That won't really be measurable. If anything, do it for code simplicity. But IMO only worth it if we can do it in many places. I would not pull in a dependency for replacing a single place.
</blockquote></details>

I agree. But in this case, the dependency should only contain macros, so it shouldn't add any bloat at least.
___
*[talpid-platform-metadata/src/windows.rs, line 42 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSDAJTDE4kcyTwY7_NI:-MSDAJTDE4kcyTwY7_NJ:b461hrg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c628e5a35d2ac00ec3a79d21dd84b316f19e70f/talpid-platform-metadata/src/windows.rs#L42)):*
> ```Rust
> impl WindowsVersion {
>     pub fn new() -> Result<WindowsVersion, io::Error> {
>         let module_name = utf16_lit::utf16!("ntdll");
> ```

If we do want to use this dependency, I think it needs to invoke `utf16_null!`. The string above is not null-terminated.

faern (commented):
> Reviewed 1 of 16 files at r1, 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: 14 of 16 files reviewed, 3 unresolved discussions (waiting on @faern and @pinkisemils)

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: 15 of 16 files reviewed, 3 unresolved discussions (waiting on @faern and @pinkisemils)

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: 13 of 16 files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

___
*[talpid-platform-metadata/src/windows.rs, line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSCfupM-GAnbxSVhyYk:-MSDIz5ECXCOokJv85Cy:b-e16pld) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9accd3529b289f79f73514055b25814f41dc462a/talpid-platform-metadata/src/windows.rs#L37)):*

I agree. But in this case, the dependency should only contain macros, so it shouldn't add any bloat at least.
</blockquote></details>

I think there might be more than a single place where this could be used, at least there'd be multiple places in `talpid-core`. But I'll revert this for now to lose the dependency.
___
*[talpid-platform-metadata/src/windows.rs, line 42 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421#-MSDAJTDE4kcyTwY7_NI:-MSD__UJ2ie5BM62ptrs:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c628e5a35d2ac00ec3a79d21dd84b316f19e70f/talpid-platform-metadata/src/windows.rs#L42)):*

If we do want to use this dependency, I think it needs to invoke `utf16_null!`. The string above is not null-terminated.
</blockquote></details>

Done.

dlon (approved):
> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2421)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2422, opened 2021-01-29
Add IP version constraint for WireGuard

dlon:
This adds a relay constraint for connecting to relays over IPv6. It can be enabled in the CLI:

```
mullvad relay set tunnel wireguard --ipv 6
```

If the constraint is set to either `4` or `Any`, then the relay selector will always prefer the IPv4 address.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Line wrap the file at 100 chars.                                              Th
 - Enable isolation of the Electron renderer process to protect against potentially malicious third
   party dependencies.
 - Add 51820 to list of WireGuard ports in app settings.
+- Add option to connect to WireGuard relays over IPv6.
 
 #### Android
 - Allow reaching the API server when connecting, disconnecting or in a blocked state.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -977,6 +977,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "itertools"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "37d572918e350e82412fe766d24b15e6682fb2ed2bbe018280caa810397cb319"
+dependencies = [
+ "either",
+]
+
+[[package]]
 name = "itoa"
 version = "0.4.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1231,6 +1240,7 @@ dependencies = [
  "env_logger 0.8.2",
  "err-derive 0.3.0",
  "futures",
+ "itertools 0.10.0",
  "mullvad-management-interface",
  "mullvad-paths",
  "mullvad-types",
@@ -1952,7 +1962,7 @@ checksum = "02b10678c913ecbd69350e8535c3aef91a8676c0773fc1d7b95cdd196d7f2f26"
 dependencies = [
  "bytes 0.5.6",
  "heck",
- "itertools",
+ "itertools 0.8.2",
  "log 0.4.14",
  "multimap",
  "petgraph",
@@ -1969,7 +1979,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "537aa19b95acde10a12fec4301466386f757403de4cd4e5b4fa78fb5ecb18f72"
 dependencies = [
  "anyhow",
- "itertools",
+ "itertools 0.8.2",
  "proc-macro2",
  "quote",
  "syn",
--- a/mullvad-cli/Cargo.toml
+++ b/mullvad-cli/Cargo.toml
@@ -20,6 +20,7 @@ env_logger = "0.8.2"
 futures = "0.3"
 natord = "1.0.9"
 serde = "1.0"
+itertools = "0.10"
 
 mullvad-types = { path = "../mullvad-types" }
 mullvad-paths = { path = "../mullvad-paths" }
--- a/mullvad-cli/src/cmds/relay.rs
+++ b/mullvad-cli/src/cmds/relay.rs
@@ -1,5 +1,6 @@
 use crate::{location, new_rpc_client, Command, Error, Result};
 use clap::{value_t, values_t};
+use itertools::Itertools;
 use std::{
     io::{self, BufRead},
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
@@ -643,9 +644,16 @@ impl Relay {
                         (false, true) => "WireGuard",
                         _ => unreachable!("Bug in relay filtering earlier on"),
                     };
+                    let mut addresses = vec![&relay.ipv4_addr_in];
+                    if !relay.ipv6_addr_in.is_empty() {
+                        addresses.push(&relay.ipv6_addr_in);
+                    }
                     println!(
                         "\t\t{} ({}) - {}, hosted by {}",
-                        relay.hostname, relay.ipv4_addr_in, support_msg, relay.provider
+                        relay.hostname,
+                        addresses.iter().join(", "),
+                        support_msg,
+                        relay.provider
                     );
                 }
             }
--- a/mullvad-cli/src/cmds/relay.rs
+++ b/mullvad-cli/src/cmds/relay.rs
@@ -157,8 +157,8 @@ impl Command for Relay {
                                     .possible_values(&["any", "udp", "tcp"]),
                             )
                             .arg(
-                                clap::Arg::with_name("ip protocol")
-                                    .long("ip")
+                                clap::Arg::with_name("ip version")
+                                    .long("ipv")
                                     .required(false)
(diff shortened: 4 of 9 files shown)


pinkisemils (commented):
> Reviewed 8 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422)* status: 8 of 9 files reviewed, all discussions resolved

___
*[mullvad-daemon/src/relays.rs, line 652 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422#-MSDb_HU91nIeAbbxfI6:-MSDb_HU91nIeAbbxfI7:bt6dahe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dba11614a88e82c11e1142a8751fbec7efc4613/mullvad-daemon/src/relays.rs#L652)):*
> ```Rust
>                             relay.ipv4_addr_in.into()
>                         }
>                         Constraint::Only(IpVersion::V6) => relay
> ```

Shouldn't relays without IPv6 entry addresses be filtered out at this point in time?
I think this would greatly benefit from having a unit test to see this does what it needs to do and then have a panic here IMO. But that might necessitate clients update their filters so that IPv6 and IPv4 relays are filtered properly. Or maybe this here is the expected behavior - maybe @faern can have a better idea as to what is supposed to happen here.
___
*[mullvad-daemon/src/relays.rs, line 665 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422#-MSDadhMCkL6DdbaYC5K:-MSDadhMCkL6DdbaYC5L:b-xotqbe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dba11614a88e82c11e1142a8751fbec7efc4613/mullvad-daemon/src/relays.rs#L665)):*
> ```Rust
>         };
> 
>         match constraints.tunnel_protocol {
> ```

I think to simplify this code, it'd be better to just have this match statement only be compiled on _not_ android, and have the android codepath just be a single expression (since it's always going to be WireGuard on Android).

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422)* status: 8 of 9 files reviewed, all discussions resolved

___
*[mullvad-daemon/src/relays.rs, line 652 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422#-MSDb_HU91nIeAbbxfI6:-MSDd7sI8k0LMTSOaWEY:bcz00s1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dba11614a88e82c11e1142a8751fbec7efc4613/mullvad-daemon/src/relays.rs#L652)):*

Shouldn't relays without IPv6 entry addresses be filtered out at this point in time?
I think this would greatly benefit from having a unit test to see this does what it needs to do and then have a panic here IMO. But that might necessitate clients update their filters so that IPv6 and IPv4 relays are filtered properly. Or maybe this here is the expected behavior - maybe @faern can have a better idea as to what is supposed to happen here.
</blockquote></details>

Probably, yes.

Looking at `mullvad_rpc::relay_list`, it looks as if it isn't really optional. Also, the API does currently return IPv6 addresses for all relays. Am I mistaken?

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[mullvad-cli/src/cmds/relay.rs, line 163 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422#-MSTO0nN8tosTD7XrXzG:-MSTO0nP5tNZLwvI_nqV:b-w0cav8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dba11614a88e82c11e1142a8751fbec7efc4613/mullvad-cli/src/cmds/relay.rs#L163)):*
> ```Rust
>                                     .required(false)
>                                     .default_value("any")
>                                     .possible_values(&["any", "4", "6"]),
> ```

An argument named `--ip` not taking an IP address is very much going against my instincts. Can we name the flag "ip version" instead of "ip protocol"? We have enough "protocol"s in this daemon already. It's transport protocols and tunnel protocols etc. `--ipv` might not be perfect, but it's at least not using a long name where everyone will assume it takes an IP address as argument.

We need to overhaul the UX of the CLI in general. But the more careful we are when adding new features now the less work will have to be done and fewer breaking changes when that overhaul comes.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422)* status: 3 of 9 files reviewed, 1 unresolved discussion (waiting on @faern and @pinkisemils)

___
*[mullvad-cli/src/cmds/relay.rs, line 163 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422#-MSTO0nN8tosTD7XrXzG:-MSXH5qO3swSec8s1PNI:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dba11614a88e82c11e1142a8751fbec7efc4613/mullvad-cli/src/cmds/relay.rs#L163)):*

An argument named `--ip` not taking an IP address is very much going against my instincts. Can we name the flag "ip version" instead of "ip protocol"? We have enough "protocol"s in this daemon already. It's transport protocols and tunnel protocols etc. `--ipv` might not be perfect, but it's at least not using a long name where everyone will assume it takes an IP address as argument.

We need to overhaul the UX of the CLI in general. But the more careful we are when adding new features now the less work will have to be done and fewer breaking changes when that overhaul comes.
</blockquote></details>

Done.
___
*[mullvad-daemon/src/relays.rs, line 665 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422#-MSDadhMCkL6DdbaYC5K:-MSXH9i283TG56RyowM-:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dba11614a88e82c11e1142a8751fbec7efc4613/mullvad-daemon/src/relays.rs#L665)):*

I think to simplify this code, it'd be better to just have this match statement only be compiled on _not_ android, and have the android codepath just be a single expression (since it's always going to be WireGuard on Android).
</blockquote></details>

Done.

pinkisemils (commented):
> Reviewed 2 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422)* status: 5 of 9 files reviewed, 1 unresolved discussion (waiting on @dlon and @pinkisemils)

___
*[mullvad-types/src/relay_constraints.rs, line 415 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422#-MSZrhvX-TKGKAaaVede:-MSZrhvX-TKGKAaaVedf:bwfimhr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78cd8b1d0b2120195338b0bd94b187ea5e6a2891/mullvad-types/src/relay_constraints.rs#L415)):*
> ```Rust
> 
> impl Match<WireguardEndpointData> for WireguardConstraints {
>     fn matches(&self, endpoint: &WireguardEndpointData) -> bool {
> ```

Shouldn't this function be updated too? Or does this implementation not matter because an IPv4 address be turned in an IPv6 one?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422)* status: 5 of 9 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[mullvad-types/src/relay_constraints.rs, line 415 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422#-MSZrhvX-TKGKAaaVede:-MSbcCiX2VsI3cZDADMx:b-z62sff) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78cd8b1d0b2120195338b0bd94b187ea5e6a2891/mullvad-types/src/relay_constraints.rs#L415)):*

Shouldn't this function be updated too? Or does this implementation not matter because an IPv4 address be turned in an IPv6 one?
</blockquote></details>

All WireGuard endpoints are currently assigned in-v6 addresses, which means that any constraint would match here.

pinkisemils (approved):
> Reviewed 4 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (commented):
> Reviewed 3 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2422)* status: 8 of 9 files reviewed, all discussions resolved (waiting on @pinkisemils)

PR #2378, opened 2021-01-07
Fall back on LLMNR/NetBIOS name resolution

dlon:
Support for custom DNS was included in the last beta release, and it involved some changes to how DNS is managed. This was because the interface metric was tied to DNS resolver preference in Windows 10, and this worked poorly with local resolvers.

This change caused problems for single-label names (e.g., NetBIOS names). If DNS resolution failed, it would not fall back on using LLMNR. Hence other hosts on a LAN were no longer being identified.

For whatever reason, the [least strict](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-gpnrpt/b5760f13-0a23-4e37-9253-ad12af60e8a6) setting must be used, i.e. "Always fall back to LLMNR and NetBIOS for any kind of name resolution error.". Otherwise, names do not resolve.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2378)

PR #2188, opened 2020-10-19
Support custom DNS servers on Windows

dlon:
This allows users to use DNS servers other than those provided by the relay server. One or more addresses can be set. For each address, the following applies:

* If an IP address is in a private network, non-tunnel interfaces are allowed by the firewall to send and receive DNS traffic to that address. However, the interfaces must be configured by the user to do so.

* For other IP addresses, the daemon configures the tunnel interface correctly to use the provided addresses. `WinFw` will allow traffic to pass to/from those addresses, and on that interface only. Due to hijacking, this currently has no effect.

Servers can be set or retrieved in the CLI as follows:
```
mullvad custom-dns set [addresses...]
mullvad custom-dns get
```

If `mullvad custom-dns set` is run without any arguments, the DNS server will be set to the gateway of the tunnel interface (i.e., the default setting).

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2188)

PR #2425, opened 2021-02-01
Use per-interface DNS config on Windows

dlon:
Previously, DNS servers were configured globally instead of per-interface (#2188). This is potentially better for a few reasons, but it sometimes broke network sharing and resolution of local hostnames. A previous fix (#2378) did not fully resolve this.

This PR reverts back to using per-interface settings, but also solves the problem of slow resolution for local resolvers. This is a safer default. Everything works the same as in the last stable build, unless custom DNS is enabled. The main drawbacks:
* The ordering of custom DNS resolvers is now irrelevant. Public DNS resolvers will be preferred since the tunnel interface has the lowest metric. Arguably, we never made any guarantees about the order anyhow.
* Setting custom local DNS resolvers does not set them on any interface. Currently, the assumption is that your "physical" interface is configured to use them already. This is likely usually the case (e.g., you'll want to set the resolver to something like `192.168.0.1`), but if it isn't, you'll need to manually configure the given network adapter. This could be changed by trying to infer which interface to use.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2425)

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,10 +382,9 @@ echo "org.gradle.jvmargs=-Xmx4608M" >> ~/.gradle/gradle.properties
    WireGuard on Linux.
 
 * `TALPID_DNS_CACHE_POLICY` - On Windows, this changes how DNS is configured:
-  * `1`: The default. This sets a global list of DNS servers that `dnscache` will use instead of
-         the servers specified on each interface.
-  * `0`: Only set DNS servers on the tunnel interface. This will misbehave if local custom DNS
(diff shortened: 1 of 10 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2425)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2433, opened 2021-02-02
Use path for app icon instead of data url

raksooo:
This PR changes the window icon on Linux to use a path instead of a data URL. The function returning the image was changed to returning the data URL instead of the path since the path isn't usable in the now sandboxed renderer process. But for the window icon a path is more suitable since it's not running within the sandbox and with the data URL the image wasn't displayed and also causes a segmentation fault in Electron 11.2.0+.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433)

### DIFF
--- a/gui/src/main/index.ts
+++ b/gui/src/main/index.ts
@@ -55,7 +55,7 @@ import { ConnectionObserver, DaemonRpc, SubscriptionListener } from './daemon-rp
 import { InvalidAccountError } from './errors';
 import Expectation from './expectation';
 import GuiSettings from './gui-settings';
-import { getAppIcon } from './linux-desktop-entry';
+import { findIconPath } from './linux-desktop-entry';
 import {
   backupLogFile,
   cleanUpLogDirectory,
@@ -1506,7 +1506,7 @@ class ApplicationMain {
       case 'linux':
         return new BrowserWindow({
           ...options,
-          icon: await getAppIcon('mullvad-vpn'),
+          icon: await findIconPath('mullvad-vpn'),
         });
 
       default: {
--- a/gui/src/main/linux-desktop-entry.ts
+++ b/gui/src/main/linux-desktop-entry.ts
@@ -39,14 +39,27 @@ export function shouldShowApplication(application: ILinuxApplication): boolean {
   );
 }
 
-export async function getAppIcon(name?: string) {
-  if (!name || path.isAbsolute(name)) {
-    return name;
-  }
+export async function getImageDataUrl(imagePath: string): Promise<string> {
+  if (imagePath && path.extname(imagePath) === '.svg') {
+    const contents = await fs.promises.readFile(imagePath);
+    return `data:image/svg+xml;base64,${contents.toString('base64')}`;
+  } else {
+    const image = nativeImage.createFromPath(imagePath);
 
+    if (image.isEmpty()) {
+      log.error(`Failed to load nativeImage: ${imagePath}`);
+      throw new Error(`Failed to load nativeImage: ${imagePath}`);
+    } else {
+      return image.toDataURL();
+    }
+  }
+}
+
+// Returns the path of the icon with the specified name. If none is found it returns undefined.
+export async function findIconPath(name: string): Promise<string | undefined> {
   // Chromium doesn't support .xpm files
   const extensions = ['svg', 'png'];
-  const iconPath = await findIcon(name, extensions, [
+  return findIcon(name, extensions, [
     getIconDirectories(),
     await getGtkThemeDirectories(),
     // Begin with preferred sized but if nothing matches other sizes should be considered as well.
@@ -54,19 +67,6 @@ export async function getAppIcon(name?: string) {
     // Search in all categories of icons.
     [/.*/],
   ]);
-
-  if (iconPath && path.extname(iconPath) === '.svg') {
-    try {
-      const contents = await fs.promises.readFile(iconPath);
-      return `data:image/svg+xml;base64,${contents.toString('base64')}`;
-    } catch (error) {
-      log.error(`Failed to read icon of application: ${name},`, error);
-    }
-  } else if (iconPath) {
-    return nativeImage.createFromPath(iconPath).toDataURL();
-  }
-
-  return undefined;
 }
 
 // Implemented according to freedesktop specification.
--- a/gui/src/main/linux-split-tunneling.ts
+++ b/gui/src/main/linux-split-tunneling.ts
@@ -4,7 +4,7 @@ import linuxAppList, { AppData } from 'linux-app-list';
 import path from 'path';
 import { pascalCaseToCamelCase } from './transform-object-keys';
 import { ILinuxSplitTunnelingApplication } from '../shared/application-types';
-import { getAppIcon, shouldShowApplication } from './linux-desktop-entry';
+import { findIconPath, getImageDataUrl, shouldShowApplication } from './linux-desktop-entry';
 
 const PROBLEMATIC_APPLICATIONS = {
   launchingInExistingProcess: [
@@ -42,9 +42,29 @@ export function getApplications(locale: string): Promise<ILinuxSplitTunnelingApp
     .filter(shouldShowApplication)
     .map(addApplicationWarnings)
     .sort((a, b) => a.name.localeCompare(b.name))
-    .map(async (app) => ({ ...app, icon: await getAppIcon(app.icon) }));
+    .map(replaceIconNameWithDataUrl);
 
   return Promise.all(applications);
+}
+
+async function replaceIconNameWithDataUrl(
+  app: ILinuxSplitTunnelingApplication,
+): Promise<ILinuxSplitTunnelingApplication> {
+  try {
+    // Either the app has no icon or it's already an absolute path.
+    if (app.icon === undefined || path.isAbsolute(app.icon)) {
+      return app;
+    }
+
+    const iconPath = await findIconPath(app.icon);
+    if (iconPath === undefined) {
+      return app;
+    }
+
+    return { ...app, icon: await getImageDataUrl(iconPath) };
+  } catch (e) {
+    return app;
+  }
 }
 
 function addApplicationWarnings(


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/linux-desktop-entry.ts, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433#-MSb4CWwBf2vtu6GQQah:-MSb4CWwBf2vtu6GQQai:bjod9m0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b92a52f107e90a59567a1935c9d945c60676f7b/gui/src/main/linux-desktop-entry.ts#L42)):*
> ```XML
> }
> 
> export async function getAppIcon(name?: string): Promise<string | undefined> {
> ```

Why not return `Promise<string>` instead? I think it's reasonable to throw an error if we can't read an image instead of using `Promise` that never returns any `Error` back.
___
*[gui/src/main/linux-desktop-entry.ts, line 50 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433#-MSb5qw68_w2vsKdRnMC:-MSb5qw68_w2vsKdRnMD:b-brhod5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b92a52f107e90a59567a1935c9d945c60676f7b/gui/src/main/linux-desktop-entry.ts#L50)):*
> ```XML
>       return `data:image/svg+xml;base64,${contents.toString('base64')}`;
>     } catch (error) {
>       log.error(`Failed to read icon of application: ${name},`, error);
> ```

I'd really just rethrow an error here instead of returning `undefined`.
___
*[gui/src/main/linux-desktop-entry.ts, line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433#-MSb4N2I6TlvekY9H0SB:-MSb4N2I6TlvekY9H0SC:bsibnpm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b92a52f107e90a59567a1935c9d945c60676f7b/gui/src/main/linux-desktop-entry.ts#L53)):*
> ```XML
>     }
>   } else if (iconPath) {
>     return nativeImage.createFromPath(iconPath).toDataURL();
> ```

It's a bit late but I looked at Electron docs today and found that `createFromPath()` returns an empty image if it cannot read it from the given path. 

We could treat it as an error, i.e:

```
const image = nativeImage.createFromPath(iconPath)

if (image.isEmpty()) {
  throw new Error('Native image failed to load image')
} else {
  return image.toDataURL()
}
```
___
*[gui/src/main/linux-desktop-entry.ts, line 59 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433#-MSb621O505cyNZ_mxPI:-MSb621O505cyNZ_mxPJ:b-dm6f2c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b92a52f107e90a59567a1935c9d945c60676f7b/gui/src/main/linux-desktop-entry.ts#L59)):*
> ```XML
> }
> 
> export async function getAppIconPath(name?: string): Promise<string | undefined> {
> ```

Same here, `Promise<string>` would be better IMO because then you have two code paths instead of three, i.e happy code path: `string | undefined`, unhappy code path: handle error.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433)* status: all files reviewed, 4 unresolved discussions (waiting on @raksooo)

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433)* status: all files reviewed, 4 unresolved discussions (waiting on @pronebird and @raksooo)

___
*[gui/src/main/linux-desktop-entry.ts, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433#-MSb4CWwBf2vtu6GQQah:-MSbCIIZ-757aBf7KNxO:bm2x0oe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b92a52f107e90a59567a1935c9d945c60676f7b/gui/src/main/linux-desktop-entry.ts#L42)):*

Why not return `Promise<string>` instead? I think it's reasonable to throw an error if we can't read an image instead of using `Promise` that never returns any `Error` back.
</blockquote></details>

This function receives the app name as argument and it's possible that the app doesn't have an icon. In that case we want to return `undefined`.

I agree that throwing an `Error` would be reasonable when we can't read a file. How would we handle this in the caller? Let's say an icon for an application in the split tunneling app list contains some kind of corruption or other error. I assume the best way would be to let the caller log the error and use `undefined` for the app icon?

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433)* status: all files reviewed, 4 unresolved discussions (waiting on @pronebird and @raksooo)

___
*[gui/src/main/linux-desktop-entry.ts, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433#-MSb4CWwBf2vtu6GQQah:-MSbEf4R1lcE7Eemrcwo:bj8zsmd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b92a52f107e90a59567a1935c9d945c60676f7b/gui/src/main/linux-desktop-entry.ts#L42)):*

This function receives the app name as argument and it's possible that the app doesn't have an icon. In that case we want to return `undefined`.

I agree that throwing an `Error` would be reasonable when we can't read a file. How would we handle this in the caller? Let's say an icon for an application in the split tunneling app list contains some kind of corruption or other error. I assume the best way would be to let the caller log the error and use `undefined` for the app icon?
</blockquote></details>

The functions could be changed to
```
export async function findAppIconPath(name: string): Promise<string | undefined>;
export async function getIconDataUrl(iconPath: string): Promise<string>;
```
which would make it easier to understand what is going on IMO. We would still need to return `undefined` if we don't find an icon for an app.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @pronebird and @raksooo)

___
*[gui/src/main/linux-desktop-entry.ts, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433#-MSb4CWwBf2vtu6GQQah:-MSbIyx43oOs5Gtc0sLd:btx28rq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b92a52f107e90a59567a1935c9d945c60676f7b/gui/src/main/linux-desktop-entry.ts#L42)):*

The functions could be changed to
```
export async function findAppIconPath(name: string): Promise<string | undefined>;
export async function getIconDataUrl(iconPath: string): Promise<string>;
```
which would make it easier to understand what is going on IMO. We would still need to return `undefined` if we don't find an icon for an app.
</blockquote></details>

How about something like this?
___
*[gui/src/main/linux-desktop-entry.ts, line 50 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433#-MSb5qw68_w2vsKdRnMC:-MSbJb458CLniDzbgl55:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b92a52f107e90a59567a1935c9d945c60676f7b/gui/src/main/linux-desktop-entry.ts#L50)):*

I'd really just rethrow an error here instead of returning `undefined`.
</blockquote></details>

Done.
___
*[gui/src/main/linux-desktop-entry.ts, line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433#-MSb4N2I6TlvekY9H0SB:-MSbJ3Mm9Wy3yeMxPcqi:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b92a52f107e90a59567a1935c9d945c60676f7b/gui/src/main/linux-desktop-entry.ts#L53)):*

It's a bit late but I looked at Electron docs today and found that `createFromPath()` returns an empty image if it cannot read it from the given path. 

We could treat it as an error, i.e:

```
const image = nativeImage.createFromPath(iconPath)

if (image.isEmpty()) {
  throw new Error('Native image failed to load image')
} else {
  return image.toDataURL()
}
```
</blockquote></details>

Done.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[gui/src/main/linux-desktop-entry.ts, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433#-MSb4CWwBf2vtu6GQQah:-MSbi7urEUF9VmB4ZsCh:b-s0u34h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b92a52f107e90a59567a1935c9d945c60676f7b/gui/src/main/linux-desktop-entry.ts#L42)):*
> I agree that throwing an Error would be reasonable when we can't read a file. How would we handle this in the caller? Let's say an icon for an application in the split tunneling app list contains some kind of corruption or other error. I assume the best way would be to let the caller log the error and use undefined for the app icon?

You do log errors right here so I guess the caller doesn't have to log the same error twice.

As I mentioned in our conversation on Slack, I think you could reduce the number of code paths by issuing an error when there is no icon. In my mind the caller does something like this:

```
try {
  let icon: string = await findAppIconPath(appName)
  let dataURL = await getIconDataUrl(icon)
  // set icon
} catch {
  // use generic icon
}
```

pronebird (approved):
:lgtm:

> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433#-:-MSbvfgG-tuBiffAoY6W:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2433)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2439, opened 2021-02-05
Implement Intermittent daemon

jvff:
There are certain parts of the app that can only work if a daemon instance is available. Most of the time there will be an instance available, but there are situations where it's not available:

- When the service (and/or the daemon) is still starting
- When the daemon has crashed and is restarting
- When the service was requested to stop but then requested to start shortly afterwards, so it must perform a "soft restart"

Previously, the code that depended on the daemon instance availability would operate through callbacks, so that they can correctly handle the events of an instance becoming available or unavailable. This added complexity to the code, which is exacerbated by the refactor to split the app in two processes.

This PR attempts to collect that complexity into a single helper class, called `Intermittent`. It is responsible for waiting for something (it has a generic type parameter) to become available, and suspends any coroutines that need that something until it becomes available. It still provides an `EventNotifier` endpoint for registering callbacks for the use cases that still need it.

The PR also performs a small change in `EventNotifier` to make it notify listeners in the order that they were registered, so that the creator of an `EventNotifier` can register specific callbacks to be executed before any other listener is notified.

The new `Intermittent` class is then used by the `DaemonInstance` class. The `DaemonInstance` class exposes an `Intermittent<MullvadDaemon>` and updates it internally whenever the instance becomes available or unavailable. The classes that still use `MullvadDaemon` directly will be changed by future PRs that incrementally remove the `ServiceInstance` callback channel.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2439)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/DaemonInstance.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/DaemonInstance.kt
@@ -14,7 +14,7 @@ import net.mullvad.mullvadvpn.util.Intermittent
 private const val API_IP_ADDRESS_FILE = "api-ip-address.txt"
 private const val RELAYS_FILE = "relays.json"
 
-class DaemonInstance(val vpnService: MullvadVpnService, val listener: (MullvadDaemon?) -> Unit) {
+class DaemonInstance(val vpnService: MullvadVpnService) {
     private enum class Command {
         START,
         STOP,
@@ -24,7 +24,6 @@ class DaemonInstance(val vpnService: MullvadVpnService, val listener: (MullvadDa
 
     private var daemon by observable<MullvadDaemon?>(null) { _, oldInstance, _ ->
         oldInstance?.onDestroy()
-        listener(newInstance)
     }
 
     val intermittentDaemon = Intermittent<MullvadDaemon>()
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/DaemonInstance.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/DaemonInstance.kt
@@ -9,6 +9,7 @@ import kotlinx.coroutines.channels.ClosedReceiveChannelException
 import kotlinx.coroutines.channels.ReceiveChannel
 import kotlinx.coroutines.channels.actor
 import kotlinx.coroutines.channels.sendBlocking
+import net.mullvad.mullvadvpn.util.Intermittent
 
 private const val API_IP_ADDRESS_FILE = "api-ip-address.txt"
 private const val RELAYS_FILE = "relays.json"
@@ -21,10 +22,12 @@ class DaemonInstance(val vpnService: MullvadVpnService, val listener: (MullvadDa
 
     private val commandChannel = spawnActor()
 
-    private var daemon by observable<MullvadDaemon?>(null) { _, oldInstance, newInstance ->
+    private var daemon by observable<MullvadDaemon?>(null) { _, oldInstance, _ ->
         oldInstance?.onDestroy()
         listener(newInstance)
     }
+
+    val intermittentDaemon = Intermittent<MullvadDaemon>()
 
     fun start() {
         commandChannel.sendBlocking(Command.START)
@@ -36,6 +39,7 @@ class DaemonInstance(val vpnService: MullvadVpnService, val listener: (MullvadDa
 
     fun onDestroy() {
         commandChannel.close()
+        intermittentDaemon.onDestroy()
     }
 
     private fun spawnActor() = GlobalScope.actor<Command>(Dispatchers.Default, Channel.UNLIMITED) {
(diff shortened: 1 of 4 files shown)


pinkisemils (changes requested):
> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2439)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @jvff)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/util/Intermittent.kt, line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2439#-MT39xMWDU7PLHUrKbOo:-MT396e63Dgk5NpkZJ64:bh0de0g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e19d347cc21fc8ad7133292e6131e7e02719254b/android/src/main/kotlin/net/mullvad/mullvadvpn/util/Intermittent.kt#L32)):*
> ```Kotlin
> class Intermittent<T> {
>     private val notifier = EventNotifier<T?>(null)
>     private val semaphore = Semaphore(1, 1)
> ```

Could the semaphore be replaced with a conditional variable? You could awake coroutines that are awaiting a value when it's being updated. A sempahore with a permit of 1 is a mutex anyway. I don't see a way that the semaphore might be bumped to more than 1 lease at a time - in which case I think using a regular mutex might be good enough?

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2439)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/util/Intermittent.kt, line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2439#-MT39xMWDU7PLHUrKbOo:-MT3JBwU9XhASFrPn7ik:b-dn8czm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e19d347cc21fc8ad7133292e6131e7e02719254b/android/src/main/kotlin/net/mullvad/mullvadvpn/util/Intermittent.kt#L32)):*

Could the semaphore be replaced with a conditional variable? You could awake coroutines that are awaiting a value when it's being updated. A sempahore with a permit of 1 is a mutex anyway. I don't see a way that the semaphore might be bumped to more than 1 lease at a time - in which case I think using a regular mutex might be good enough?
</blockquote></details>

I couldn't find a conditional variable implementation for coroutines. There's only Mutexes and Semaphores in the Kotlin standard coroutines library =/

While you're right that a mutex can be implemented by wrapping a semaphore with a single permit, the inverse is not true. A semaphore with a permit of one can have the permit released by any thread without prior notice, unblocking a single thread, while a mutex can only be unlocked/released by the first thread to lock it.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2439)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @jvff)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/util/Intermittent.kt, line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2439#-MT39xMWDU7PLHUrKbOo:-MT5LX8JC7xYmwQ5M043:b-e4o6ta) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e19d347cc21fc8ad7133292e6131e7e02719254b/android/src/main/kotlin/net/mullvad/mullvadvpn/util/Intermittent.kt#L32)):*

I couldn't find a conditional variable implementation for coroutines. There's only Mutexes and Semaphores in the Kotlin standard coroutines library =/

While you're right that a mutex can be implemented by wrapping a semaphore with a single permit, the inverse is not true. A semaphore with a permit of one can have the permit released by any thread without prior notice, unblocking a single thread, while a mutex can only be unlocked/released by the first thread to lock it.
</blockquote></details>

There is an Android specific [implementation](https://developer.android.com/reference/kotlin/android/os/ConditionVariable).

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2439)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/util/Intermittent.kt, line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2439#-MT39xMWDU7PLHUrKbOo:-MT5OrlD9t1FPn03iYj5:byrxjii) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e19d347cc21fc8ad7133292e6131e7e02719254b/android/src/main/kotlin/net/mullvad/mullvadvpn/util/Intermittent.kt#L32)):*

There is an Android specific [implementation](https://developer.android.com/reference/kotlin/android/os/ConditionVariable).
</blockquote></details>

Ah, but that is not coroutine friendly :/ I guess then the semaphore will be the best one can do here.

axti:
:lgtm:

axti:
:lgtm:

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2439)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2465, opened 2021-02-12
Android improve transition animation

axti:
This PR is changed z-order for fragments according to animation transitions dynamically instead of hardcoded z-order on 
layouts. From user perspective it will behave same, but from developers it will get more flexibility.
Also, animation duration moved to resources, so it will be consistent and easy to change in future.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2465)

### DIFF
--- a/android/src/main/res/anim/do_nothing.xml
+++ b/android/src/main/res/anim/do_nothing.xml
@@ -2,5 +2,5 @@
 <set xmlns:android="http://schemas.android.com/apk/res/android">
     <translate android:fromYDelta="0"
                android:toYDelta="0"
-               android:duration="450" />
+               android:duration="@integer/transition_animation_duration" />
 </set>
--- a/android/src/main/res/anim/fade_in.xml
+++ b/android/src/main/res/anim/fade_in.xml
@@ -2,5 +2,5 @@
 <set xmlns:android="http://schemas.android.com/apk/res/android">
     <alpha android:fromAlpha="0.0"
            android:toAlpha="1.0"
-           android:duration="450" />
+           android:duration="@integer/transition_animation_duration" />
 </set>
--- a/android/src/main/res/anim/fade_out.xml
+++ b/android/src/main/res/anim/fade_out.xml
@@ -2,5 +2,5 @@
 <set xmlns:android="http://schemas.android.com/apk/res/android">
     <alpha android:fromAlpha="1.0"
            android:toAlpha="0.0"
-           android:duration="450" />
+           android:duration="@integer/transition_animation_duration" />
 </set>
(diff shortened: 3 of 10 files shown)


jvff (approved):
> Reviewed 25 of 26 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2465)* status: 25 of 26 files reviewed, 1 unresolved discussion (waiting on @axti)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/ui/fragments/BaseFragment.kt, line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2465#-MTL3VTf5shCqESj4oMk:-MTL3VTgD1RIZC6K-s8t:b-nebi5n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3f709cb5e7fb23f126f8430aab77d0e36339dec3/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/fragments/BaseFragment.kt#L10)):*
> ```Kotlin
> abstract class BaseFragment : Fragment() {
>     override fun onCreateAnimation(transit: Int, enter: Boolean, nextAnim: Int): Animation? {
>         val zAdjustment = if (animationsToAdjustZorder.contains(nextAnim))
> ```

Wouldn't it be better if the `setTranslationZ` method is only called inside the `if` body? Or would that cause incorrect rendering later?

Also, could you add braces for the `if` and `else` bodies? In this case it's not a big problem, but there's always the risk of adding code in the future and forgetting to add the braces.

jvff (commented):
> Reviewed 1 of 26 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2465)* status: all files reviewed, 1 unresolved discussion (waiting on @axti)

axti:
___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/ui/fragments/BaseFragment.kt, line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2465#-MTL3VTf5shCqESj4oMk:-MTLDa1a04LmcM6p1vO0:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3f709cb5e7fb23f126f8430aab77d0e36339dec3/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/fragments/BaseFragment.kt#L10)):*

Wouldn't it be better if the `setTranslationZ` method is only called inside the `if` body? Or would that cause incorrect rendering later?

Also, could you add braces for the `if` and `else` bodies? In this case it's not a big problem, but there's always the risk of adding code in the future and forgetting to add the braces.
</blockquote></details>

Done.

jvff (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2465#-:-MTLJy5n2l6muCjzG7Ad:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2471, opened 2021-02-15
Stop applying IPv6 routes for OpenVPN

pinkisemils:
The OpenVPN event callback would panic when it fails to add an IPv6 route on machines that don't support IPv6. This is because the generic option for enabling IPv6 is not taken into account, and seemingly a considerable amount of users disable IPv6 in their kernels for _reasons_. As such, I've changed the callback to take the option into account.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2471)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ Line wrap the file at 100 chars.                                              Th
 ### Fixed
 - Fix delay in showing/hiding update notification when toggling beta program.
 
+#### Linux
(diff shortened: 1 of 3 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2471)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2479, opened 2021-02-16
MacOS - Drop the patch from the short version

pinkisemils:
Reporting the patch in the version of macOS in the version check is proving to be too granular, so I've changed the version check code to only report the full version in problem reports and only use the major and minor versions in the version check header.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2479)

### DIFF
--- a/talpid-platform-metadata/src/macos.rs
+++ b/talpid-platform-metadata/src/macos.rs
@@ -2,17 +2,31 @@ mod command;
 use command::command_stdout_lossy;
 
 pub fn version() -> String {
-    format!(
-        "macOS {}",
-        command_stdout_lossy("sw_vers", &["-productVersion"])
-            .unwrap_or(String::from("[Failed to detect version]"))
-    )
+    let version = run_sw_vers().unwrap_or(String::from("N/A"));
+    format!("macOS {}", version)
 }
 
+
 pub fn short_version() -> String {
-    version()
+    let version = run_sw_vers()
+        .and_then(parse_short_version_output)
+        .map(|(major, minor)| format!("{}.{}", major, minor))
+        .unwrap_or(String::from("N/A"));
+    format!("macOS {}", version)
 }
 
 pub fn extra_metadata() -> impl Iterator<Item = (String, String)> {
     std::iter::empty()
+}
+
+/// Outputs a string in a format `$major.$minor.$patch`, e.g. `11.0.1`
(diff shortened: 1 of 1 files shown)


faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2479)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[talpid-platform-metadata/src/macos.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2479#-MTeuUi38gCPVgoO3H2X:-MTeuUi38gCPVgoO3H2Y:b-n4rx4e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/21a3beb016b73c5fc7e3f3e5a58d6b6e5bff7df3/talpid-platform-metadata/src/macos.rs#L14)):*
> ```XML
>         .and_then(parse_short_version_output)
>         .map(|(major, minor)| format!("{}.{}", major, minor))
>         .unwrap_or(String::from("[Failed to detect version]"));
> ```

Nitpick. And so far we have never seen this in the reported stats, but just `N/A` would suffice, like for Windows.
___
*[talpid-platform-metadata/src/macos.rs, line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2479#-MTetvAN5HZ5u7nJnysV:-MTetvAN5HZ5u7nJnysW:b-pwi95w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/21a3beb016b73c5fc7e3f3e5a58d6b6e5bff7df3/talpid-platform-metadata/src/macos.rs#L30)):*
> ```XML
>     let mut parts = output.split('.');
>     let major = parts.next()?.parse().ok()?;
>     let minor = parts.next()?.parse().ok()?;
> ```

Looking at the current statistics this looks good enough. From what I can see the output always has at least the major and minor. The potential problem is if they ever report only `12` as the version, then this code will fall back to `[Failed to detect version]`.

It's fine for now since we don't know any case where this could happen. But worst case we have to revisit later and fix so it falls back to full `sw_vers` output if it does not understand the format or something.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2479)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2483, opened 2021-02-16
Fix log entry signaling use of wireguard-go

pinkisemils:
The daemon wasn't logging about falling back to WireGuard-go on Linux due to a typo of yours truly. This is a fix for that.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2483)


faern (approved):
Wow. Tricky error. The Rust compiler does not catch all errors :sweat_smile:

PR #2485, opened 2021-02-17
Use only TLS 1.3 when connecting to the API

pinkisemils:
Force rustls to only use TLS 1.3 when connecting to our API. 

For OpenVPN, we specify specific cyphersuits that we believe are not weak, should we do the same here?

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2485)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Line wrap the file at 100 chars.                                              Th
   security patches.
 - Allow provider constraint to specify multiple hosting providers.
 - Only download a new relay list if it has been modified.
+- Connect to the API only via TLS 1.3
 
 #### Android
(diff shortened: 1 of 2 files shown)


faern (approved):
TLS 1.3 does not support any weak cipher suites. So I don't think we have to overdo it. I think this is good enough.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2485#-:-MTkXac_CmIfcKldoWqs:bi1d06f)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2485)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2500, opened 2021-02-22
Add support for Apple ARM64 (aka Apple Silicon)

Ch00k:
This adds support for building the app on Apple Silicon.

The frontend change is the most annoying part here. Due to the fact that `grpc-tools` [refuses](https://github.com/grpc/grpc-node/issues/1497) to support ARM architecture the management interface proto files need to be pre-built on a non-ARM platform. Then these proto files can be used during the main build. I added a helper script `build_mi_proto.sh` to help in pre-building the proto files.

On the backend only bumping of a couple of versions is needed. For `pfctl-rs` I created an [issue](https://github.com/mullvad/pfctl-rs/issues/77) requesting a new release. For `rust-tun` I made a [PR](https://github.com/meh/rust-tun/pull/28) that bumps `ioctl-sys`. **UPD**: Both have been released on crates.io

~~Due to the nature of changes in `Cargo.toml` this PR should probably not be merged as is. Nevertheless I think it gives a good overview of the changes required to support Apple Silicon (and probably the rest of the ARM platforms, too).~~

I tested this by building a package on with `./build.sh --dev-build` on an M1 Mac Mini.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2500)


talpid-core/Cargo.toml
```diff
@@ -59,14 +59,14 @@ byteorder = "1"
 nftnl = { version = "0.6", features = ["nftnl-1-1-0"] }
 mnl = { version = "0.2.0", features = ["mnl-1-0-4"] }
 which = { version = "4.0", default-features = false }
-tun = "0.5"
+tun = { git = "https://github.com/Ch00k/rust-tun.git", branch = "bump-ioctl" }
```
faern, line 62: We want to avoid git dependencies as far as we can. Can you try getting this merged upstream and published?

faern (commented):
Thanks for this! We were actually looking at this internally, but had no code ready yet. We'll take a look at this and compare with what we had and see what fits together.

talpid-core/Cargo.toml
faern: Sorry. I see now that you linked to your PR. Yes. Let's hope @meh can merge and publish that :3

faern: This has been published :tada:

Ch00k: Fixed

faern:
Thanks a bunch for this! We will look at it during the week and come back with a more thorough review.

Do you know if this can be cross compiled from an Intel mac, or does it have to be built on an M1? Maybe the other way around works? The main question is whether or not we need two build servers or can produce all artefacts on a single machine.

Ch00k:
It is probably possible, but I have not looked into it unfortunately (my target was to build the app on Apple Silicon for Apple Silicon). Quick search shows that`electron-builder` seems to support universal builds as per https://github.com/electron-userland/electron-builder/pull/5481 and https://github.com/electron/universal.
Not sure if Rust and Go code can be cross-compiled though.

Ch00k:
Building amd64 on arm64 could theoretically be possible with Rosetta 🤔

build_mi_proto.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
raksooo, line 1: This script would fit better in `gui/scripts/`.

gui/package.json
```diff
@@ -68,7 +69,7 @@
     "chai-spies": "^1.0.0",
     "cross-env": "^5.1.3",
     "electron": "^11.2.3",
-    "electron-builder": "^22.8.0",
+    "electron-builder": "^22.10.4",
```
raksooo, line 72: Could you push the changes to `package-lock.json` as well?

README.md
```diff
@@ -297,6 +297,28 @@ This should produce an installer exe, pkg or rpm+deb file in the `dist/` directo
 
 Building this requires at least 1GB of memory.
 
+#### Apple ARM64 (aka Apple Silicon)
```
raksooo, line 300: The `prost` version we depend on doesn't come with the arm64 `protoc` binary bundled and unfortunately we can't update it yet since it's incompatible with some of our other dependencies. A simple workaround is to just install protobuf `brew install protobuf`. It would be nice if this workaround is added to the readme until we can update `prost`.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Line wrap the file at 100 chars.                                              Th
 - Enable isolation of the Electron renderer process to protect against potentially malicious third
   party dependencies.
 - Add 51820 to list of WireGuard ports in app settings.
+- Add support for Apple Silicon
```
raksooo, line 30: Since we won't release Apple Silicon builds until the remaining issues have been resolved we'll have to wait with including it in the changelog.

raksooo (changes requested):
Thanks, great work! The code looks nice and it seems to run pretty well. I've left a few comments that would be nice if they were fixed before merging.

This is a great step in the directions of releasing M1 native/universal builds but there are a few things that needs to be resolved before we can release it to customers. The remaining things are:
* OpenVPN and Shadowsocks x86 binaries are still bundled and those protocols can therefore not be used in the M1 build.
* We don't have a M1 build-server yet (only devices for testing). We either have to cross-compile it or set up a new build-server.
* We would probably like to produce a universal binary before releasing it to customers.

We'll continue working on those remaining issues!

build_mi_proto.sh
Ch00k: The script puts the proto files in `dist-assets/`. Would you also want to move them to somewhere in `gui/`?

faern:
@Ch00k Just to clarify the above. We don't expect you to solve OpenVPN or universal binaries etc in this PR. We can merge this after the smaller code attached comments have been fixed. Then we'll continue from there and integrate it into our real build process.

build_mi_proto.sh
raksooo: I don’t think we have a specific directory for content like this but yeah, moving the `mi_proto` directory to `gui/scripts/mi_proto` would make it consistent with some other scripts 👍

README.md
Ch00k: Indeed, I forgot that I had to install `protobuf`. Thanks for reminding me.

gui/package.json
Ch00k: Forgive my ignorance (my experience with node.js is quite limited), but how do I do that exactly? If I run `npm install` I get way more changes in `package-lock.json` than what I changed in `package.json` (and I also get some changes in `package.json` btw).

raksooo: Which version of `npm` are you using? It could be that your using 6.x? I’ll try it out myself in a moment and see whats going on.

README.md
Ch00k: Done

CHANGELOG.md
Ch00k: Done

gui/package.json
Ch00k: 7.5.0

README.md
```diff
@@ -297,6 +297,36 @@ This should produce an installer exe, pkg or rpm+deb file in the `dist/` directo
...
+[this](https://github.com/grpc/grpc-node/issues/1497) issue) the Apple ARM64 build must be done in 2 stages:
+
+1. Build management interface proto files on a non-ARM64 platform
+2. Use the built proto files during the main build by setting the `MI_PROTO_BUILD_DIR` environment variable
+   to the path the proto files
+
+To build the management interface proto files there is a script (execute it on a non-ARM64 platform):
+
+```bash
+./build_mi_proto.sh
```
raksooo, line 312: What does `mi` mean in the file/directory names?

Ch00k: Management Interface. Happy to rename it if you think it's not explicit enough or confusing :)

build_mi_proto.sh
Ch00k: `build_mi_proto.sh` now lives in `gui/scripts`, and it creates its artefacts directory alongside itself (`gui/scripts/mi_proto`)

README.md
raksooo: Ah I should have been able to guess that. No need to rename, `mi` works :)

raksooo: Okay, after some internal discussion we came to the conclusion that it would be a good idea to rename variables, scripts and dirs to not include `mi` since it's inconsistent with other shorter versions of `management-interface`. How about naming the script `prepare-proto.sh` or `export-proto.sh` to make it easier to understand how it differs from `build-proto.sh`, and maybe spell out `management interface` in variables and directory names?

gui/package.json
raksooo: Yeah there seems to be quite a lot of changes due to `electron-builder` having a few dependencies. Just leave it as it currently is and I'll fix it after this PR has been merged.

raksooo:
When testing this out I found that it fails to build the proto on Linux and Windows which makes the entire build fail. I've been looking into this and might know how to solve it but will have to do some more testing on an M1 mac as well, which I'll do tomorrow.

Ch00k:
@raksooo weird, I thought I tested the build on Linux, and it worked for me. I'll test it again. Can't test on Windows though because I don't have access to a Windows box.

README.md
Ch00k: Now that I look at both `build-proto.sh` and `build_mi_proto.sh`, I think that perhaps it does not make sense to split them. I could do the copying in the `else` clause of the `build-proto.sh` and get rid of `build_mi_proto.sh` completely. So the `else` clause would look like this:

```bash
  BUILD_DIR="$SCRIPT_DIR/management_interface_proto"
  mkdir -p "$BUILD_DIR"

  <run grpc_tools_node_protoc>

  cp $SCRIPT_DIR/../../gui/src/main/management_interface/* $BUILD_DIR
  cp $SCRIPT_DIR/../../gui/build/src/main/management_interface/* $BUILD_DIR

  echo "Management interface proto files built successfully."
  echo "Build directory: $BUILD_DIR"
```

The user would have to perform the build in two steps (after `cd gui/scripts`): 1) `npm ci`; 2) `./build-proto.sh`.
Yes, the copying part would be redundant during the "normal" build, but we will have only one script instead of two.
Another option is no copying at all, and ask the user to collect the proto files from the two `management_interface/` directories themselves.

What do you think?

raksooo: I think I prefer the last option, to not do any copying, as long as the locations are added to the readme. Hopefully `grpc-tools` will add support for the M1 macs in a not to distant future so hopefully this action won't be needed for too long.

raksooo:
@Ch00k Yeah, the comment makes it sound like the use of the `--no-optional` argument was a workaround for a bug in npm or something. I did some testing and it seems to work as expected without `--no-optional` on Linux and Windows. I'll have to do some more testing but I agree that removing it probably is the best approach.

raksooo:
@Ch00k Done testing `--no-optional`. Please go ahead and remove it 🙂

.gitignore
```diff
@@ -29,6 +29,7 @@
 /dist-assets/sslocal
 /dist-assets/sslocal.exe
 /dist-assets/shell-completions/
+/gui/scripts/mi_proto/
```
raksooo, line 32: I guess we could remove this as well now when the readme is referring to a tmp-directory?

Ch00k: Done

raksooo (approved):
Awesome! 🎉 Thanks for fixing the comments, and thanks for the contribution! I'll merge this later today.

Ch00k:
I just built the app end-to-end one more time, just to make sure everything still works. It does.

raksooo:
:+1: Thanks for rebasing. I've updated `package-lock.json` as well.

PR #2503, opened 2021-02-23
Fix unresponsive disconnecting state

dlon:
When `ConnectingState::enter` immediately fails and enters the disconnecting state, it sometimes passes a terminated (tunnel close event) future to `DisconnectingState::enter`. In `DisconnectingState::handle_event`, the "after-connect" action is immediately taken if the future is terminated, so as to not block until a command is sent.

The problem arises when the after-connect action is to reconnect, and the connecting state keeps failing in the same manner. The tunnel state machine becomes unresponsive to commands until the loop is broken. This can be verified by always returning `Err(RecoverableStartWireguardError);` from `WireguardMonitor::start`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2503)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Line wrap the file at 100 chars.                                              Th
 
 ### Fixed
 - Fix delay in showing/hiding update notification when toggling beta program.
+- Improve responsiveness when reconnecting after some failed connection attempts.
 
 #### Windows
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2503)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2415, opened 2021-01-25
Allow API calls in blocked state.

pinkisemils:
Currently on Android, the UI components will disable buttons that interact with the API in blocked states. I've changed the respective components to instead only disable said buttons if the device is offline.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2415)

PR #2532, opened 2021-03-04
Fix "Manage Keys" button in the WireGuard Key screen always being disabled

jvff:
A recent PR (#2415) improved the way the app accesses the API server, allowing it to be reached even when the app is in the blocked state. As part of the PR, the name of a field was changed in a way that inverted its semantics, but one of the usages still considered the old semantics. More specifically, the field was previously `hasConnectivity`, and the "Manage Keys" button was enabled using the value of that field. When the PR was merged, the field changed to `isOffline`, but the code to update the "Manage Keys" button still used the value directly.

This PR fixes that by using the correct semantics.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Fixes a bug that's not present in any publicly released version, no changelog entry necessary.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2532)


axti (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2532)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2534, opened 2021-03-04
Fix login input area disappearing sometimes

jvff:
Previously, it was possible for the account input area to disappear from the Login screen sometimes. This could be reproduced by logging out, pressing back on the login screen to leave the app, and then opening the app again. The cause of the issue was that the custom animation code would reset the height of the account input area to a calculated `collapsedHeight` value. That value was only calculated after a layout change, and when recreating the widget there would be no layout change so the listener wouldn't get called, meaning the value stayed at zero.

This PR fixes that by forcing the initial value to be always calculated during initialization, and updated on layout changes. The PR also changes stopping the animation by using `cancel()` instead of `end()` to prevent the animation from sending a new height value right before the height value is forcefully set to a known value, causing two successive height updates.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2534)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,9 @@ Line wrap the file at 100 chars.                                              Th
 #### Linux
 - Fix DNS issues where NM would overwrite Mullvad tunnel's DNS config in systemd-resolved.
 
+#### Android
+- Fix input area sometimes disappearing when returning to the Login screen.
+
 
 ## [2021.2] - 2021-02-18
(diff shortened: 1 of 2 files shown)


axti (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2534)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2538, opened 2021-03-05
Really fix disappearing account login input

jvff:
A previous PR (#2534) included a fix for the account input area sometimes disappearing from the Login screen. Unfortunately, the implementation wasn't correct. It was calculating the initial value using the configured height instead of the measured height. It worked with my internal testing because my debug build included some extra code that ended up not calling the `updateHeight` when it new that the height was zero, so that was what actually "fixed" the issue during testing.

After more research, I realized that using the measured height yielded bad aesthetics (because the measured height is the minimum height, and the actual height was configured by the container through `LayoutParams`). I also realized that the container was setting a fixed value `48dp` for the height, so there was actually no need to measure it.

This PR finally fixes the issue by using the fixed initial height value for the collapsed height when creating the `AccountLogin` widget. It creates a new dimension resource so that both the code and the layout XML file can reference the same value.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Previous PR already added the changelog entry for this fix.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2538)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/widget/AccountLogin.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/widget/AccountLogin.kt
@@ -6,7 +6,6 @@ import android.content.Context
 import android.util.AttributeSet
 import android.view.LayoutInflater
 import android.view.View
-import android.view.View.MeasureSpec
 import android.view.View.OnLayoutChangeListener
 import android.view.inputmethod.InputMethodManager
 import android.widget.RelativeLayout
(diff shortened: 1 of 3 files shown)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2538#-:-MV1zLXC1uraPrYtgck4:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2540, opened 2021-03-05
Remove old IPC socket during shutdown and startup of the daemon

pinkisemils:
The daemon can fail to start if a subsequently started daemon didn't clean up it's RPC socket. So I've changed the startup code and shutdown code to remove the RPC socket. This is not done on Windows because it's not actually possible to remove a named pipe AFAICT.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2540)

### DIFF
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -755,6 +755,13 @@ where
         mem::drop(event_listener);
         mem::drop(rpc_runtime);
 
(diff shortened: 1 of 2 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2540)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2106, opened 2020-09-11
Improve routing for mullvad-exclude on Linux

dlon:
These changes cause the `mullvad_exclusions` table to mirror `main`, except that routes (unless manually added) for tunnel devices are ignored. Currently, LAN traffic will generally operate as you'd expect, assuming that the other host is on the same network. But monitoring the traffic reveals that the destination MAC address belongs to the gateway rather than the host. This is because the second table only tracks the default route belonging to the physical interface. Including all routes fixes this.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2106)

PR #2551, opened 2021-03-09
Reduce superfluous openvpn logging

faern:
The `OpenVPN event {:?}` log has two problems:

1) It's outputting debug formatting on a log level lower than DEBUG. We should not do this unless well motivated.
2) It's a very very verbose log line and as far as I know we never use/read this line for debugging.

So I moved it to TRACE to get rid of it from problem reports.

I also noticed on a Windows log the following line:
```
[talpid_core::tunnel::openvpn][DEBUG] Ignoring OpenVpnEvent Up
```
And I was confused for a while. But I then realized we use `RouteUp` to detect the tunnel up stuff. `Up` is Linux specific but has been registered as an interesting event on all platforms anyway back in #2106 .

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2551)

### DIFF
--- a/talpid-core/src/tunnel/openvpn/mod.rs
+++ b/talpid-core/src/tunnel/openvpn/mod.rs
@@ -949,7 +949,7 @@ mod event_server {
             &self,
             request: Request<EventType>,
         ) -> std::result::Result<Response<()>, tonic::Status> {
-            log::info!("OpenVPN event {:?}", request);
+            log::trace!("OpenVPN event {:?}", request);
 
             let request = request.into_inner();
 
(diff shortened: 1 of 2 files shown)


dlon (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2551#-:-MVLMLHaEVrca3BxE5wx:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 327 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2551#-MVLKS61CeVSLwlX8D-A:-MVLLfuw0hCdqR2FvCpZ:bb77ucv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81d931b2ea75bec54f5878b216985a1610003a4d/talpid-core/src/tunnel/openvpn/mod.rs#L327)):*

As far as I can tell, this is superfluous. Since the entire block it's living inside has the same cfg applied to it.
</blockquote></details>

Yep.

PR #2557, opened 2021-03-09
Try to forcibly kill stuck service in Windows uninstaller

dlon:
Previously, the service was forcibly killed if it was "marked for deletion" while installing the app. This would occur if the daemon refused to stop, or for some other unknown reason. If the daemon was still running, then the attempt to stop it would eventually time out, `mullvad-daemon.exe` would never be replaced, and the user would be left with an old version of the daemon.

The changes in this PR only try to make sure that the service is not running before continuing to install the app. It doesn't address the issue of the installer silently continuing if files cannot be replaced, or of the uninstaller not aborting upon failure.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2557)

### DIFF
--- a/dist-assets/windows/installer.nsh
+++ b/dist-assets/windows/installer.nsh
@@ -37,6 +37,7 @@
 !define LOG_VOID 1
 
 # Windows error codes
+!define ERROR_SERVICE_DOES_NOT_EXIST 1060
 !define ERROR_SERVICE_MARKED_FOR_DELETE 1072
 !define ERROR_SERVICE_DEPENDENCY_DELETED 1075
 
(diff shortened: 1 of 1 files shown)


mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2557)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[dist-assets/windows/installer.nsh, line 844 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2557#-MVNVm3g5IqWZsMe1ZqG:-MVNVm3g5IqWZsMe1ZqH:bhma2h4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/901db2c06eabc9a467a7c6db68eebd25d3f02f8d/dist-assets/windows/installer.nsh#L844)):*
> ```NSIS
> 	${EndIf}
> 
> 	nsExec::ExecToStack '"$SYSDIR\net.exe" stop mullvadvpn'
> ```

Cut here and move everything to `StopAndDeleteService`?

The current function is already long as it is.

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2557)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[dist-assets/windows/installer.nsh, line 850 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2557#-MVNWiB-3-p_hRTkIx9Q:-MVNWiB-3-p_hRTkIx9R:ba22thy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/901db2c06eabc9a467a7c6db68eebd25d3f02f8d/dist-assets/windows/installer.nsh#L850)):*
> ```NSIS
> 	Pop $1
> 
> 	nsExec::ExecToStack '"$SYSDIR\sc.exe" delete mullvadvpn'
> ```

Also, the service name is repeated everywhere. Maybe you could make it a constant instead?

PR #2560, opened 2021-03-10
Clear editing position when disabling custom DNS

jvff:
Previously, the Advanced screen would break the back button behavior after enabling and disabling custom DNS (while having no custom DNS servers configured). The back button is overridden in order to cancel editing the custom DNS server address. The bug happens because disabling custom DNS does not mark internally that editing has stopped, so the back button handler still thinks it should try to stop editing.

This PR fixes that by marking the editing position as `null` when custom DNS is disabled.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Fixes a bug not present in any publicly released version.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2560)


axti (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2560)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2573, opened 2021-03-15
Preserve the daemon log from the old installation when upgrading

pinkisemils:
To aid with debugging the issues with failing to upgrade the daemon, the daemon log from the old instance should be preserved - otherwise, by the time the user sends a problem report, the old daemon logs have been overwritten so no useful information about what happened during the installation can be inferred.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2573)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2573)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2591, opened 2021-03-17
Improve Linux offline monitor

pinkisemils:
Previously, Linux connection check would request routes via NLM_F_REQUEST | NLM_F_DUMP, which would just return a list regardless of actual connectivity in some cases, sometimes having a loopback route, sometimes a real one. The end result was that the offline monitor was too permissive.  If this causes issues for users, they should probably use an env var to force the offline monitor off.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2591)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2591)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2599, opened 2021-03-19
Partially deserialize settings to reduce duplication in settings migration

dlon:
Previously, `Settings` was duplicated in the migration module. This would result in lots of duplicate types (even for minor changes), and it also risked breaking if any of the non-duplicated types changed. A `serde_json::Value` is now passed to the migration code instead, so that only relevant parts need to be deserialized or modified.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2599)

### DIFF
--- a/mullvad-types/src/settings/migrations/mod.rs
+++ b/mullvad-types/src/settings/migrations/mod.rs
@@ -1,14 +1,15 @@
-use super::{Error, Result, Settings};
+use super::{Error, Result};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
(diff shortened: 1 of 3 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2599)* status: :shipit: complete! all files reviewed, all discussions resolved

faern:
I don't have time to review this. If you feel it's a good improvement, then go :+1:

PR #2600, opened 2021-03-19
Improve prepare release

faern:
The main change here is to make desktop releases *not* bump the Android version. This is to avoid these errors in `build-apk.sh`:

```
fatal: ambiguous argument 'android/2021.2^{commit}': unknown revision or path not in the working tree.
```

Since `2021.2` is not a real Android release. The `build.gradle` file should likely never contain this version? But since Rust is used for both we always bump it. This is mostly so that `mullvad-problem-version` will contain the correct version string in Android builds. If we find some other way to inject versions into problem reports we can make the Rust crates only bump version on desktop releases.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600)* status: 0 of 5 files reviewed, all discussions resolved

___
*[env.sh, line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600#-MW9JWFp-zEhYwk3yvBm:-MW9JWFp-zEhYwk3yvBn:bwr4dit) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3d1cfb72565ebd89f45e33846960a30a3d0ca23a/env.sh#L1)):*
> ```Shell
> # shellcheck shell=bash
> ```

This script is not supposed to be ran directly. So no shebang. But shellcheck want this to know what shell to check it as.
___
*[version-metadata.sh, line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600#-MW9Je0091jucxWKCKAj:-MW9Je0091jucxWKCKAk:b-y53skf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3d1cfb72565ebd89f45e33846960a30a3d0ca23a/version-metadata.sh#L34)):*
> ```Shell
> 
>     local semver_version
>     semver_version=$(echo "$product_version" | sed -Ee 's/($|-.*)/.0\1/g')
> ```

These things are something recommended by shellcheck. Without this any failure in the computation of the value will go unnoticed. Even with `set -e` if you do `local foobar=$(false)` it does not exit in error. Separating the declaration and value computation like this will catch any error in the computation.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600)* status: 0 of 5 files reviewed, 1 unresolved discussion

___
*[version-metadata.sh, line 165 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600#-MW9JyTf1PUEugDiGP9Y:-MW9JyTg9IbWIDXrQ3x0:b-wyx1wk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3d1cfb72565ebd89f45e33846960a30a3d0ca23a/version-metadata.sh#L165)):*
> ```Shell
>             exit 1
>         fi
>         inject_version "$PRODUCT_VERSION"
> ```

It's a bit ugly that `ANDROID` and `DESKTOP` are treated as global variables, read directly from the functions. But `PRODUCT_VERSION` is not, it's passed as an argument. What do you think about this? Make all of them globals and not passed as arguments? I'm not aware of any nice and clean way to send all three as arguments. And we probably don't want to over-complicate things.

dlon (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[version-metadata.sh, line 165 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600#-MW9JyTf1PUEugDiGP9Y:-MW9VBNF5lRhPXtF2RDJ:b-6z5x97) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3d1cfb72565ebd89f45e33846960a30a3d0ca23a/version-metadata.sh#L165)):*

It's a bit ugly that `ANDROID` and `DESKTOP` are treated as global variables, read directly from the functions. But `PRODUCT_VERSION` is not, it's passed as an argument. What do you think about this? Make all of them globals and not passed as arguments? I'm not aware of any nice and clean way to send all three as arguments. And we probably don't want to over-complicate things.
</blockquote></details>

Couldn't they simply be passed as positional arguments, since they're always either `true` or `false`?

jvff (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[version-metadata.sh, line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600#-MW9Je0091jucxWKCKAj:-MWEa4S_6svBV0nKhRWp:b7b4pbv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3d1cfb72565ebd89f45e33846960a30a3d0ca23a/version-metadata.sh#L34)):*

These things are something recommended by shellcheck. Without this any failure in the computation of the value will go unnoticed. Even with `set -e` if you do `local foobar=$(false)` it does not exit in error. Separating the declaration and value computation like this will catch any error in the computation.
</blockquote></details>

Wow, I did not that. Very scary footgun... :cold_sweat: 
___
*[version-metadata.sh, line 162 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600#-MWEbadG1CumZMLWUIA0:-MWEbadG1CumZMLWUIA1:b-kafc4z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3d1cfb72565ebd89f45e33846960a30a3d0ca23a/version-metadata.sh#L162)):*
> ```Shell
>         if [ -z "$PRODUCT_VERSION" ]; then
>             echo "Please give the release version as an argument to this script."
>             echo "For example: '2018.1-beta3' for a beta release, or '2018.6' for a stable one."
> ```

Nit: Would it make sense to output this to `stderr` with `>&2` at the end of the lines? Obviously since this is an internal tool it shouldn't matter, but I just wanted to mention it in case it's something worthwhile doing.
___
*[version-metadata.sh, line 165 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600#-MW9JyTf1PUEugDiGP9Y:-MWEc2sB035NQ0FjLUTo:bjlg7qd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3d1cfb72565ebd89f45e33846960a30a3d0ca23a/version-metadata.sh#L165)):*

Couldn't they simply be passed as positional arguments, since they're always either `true` or `false`?
</blockquote></details>

I'm conflicted about what's best here. My "clean coder" side would prefer to have them as positional arguments, but my "anxiously afraid of future bugs" side would prefer to use global variables so that we don't make a mistake with the wrong order in the future :/

Rust shell scripts when? :joy: 

Anyway, I guess any of the three options is fine (third = leaving as is).

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon, @faern, and @jvff)

___
*[version-metadata.sh, line 165 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600#-MW9JyTf1PUEugDiGP9Y:-MWO7yPS8qNek0b84jgk:b-uhgd5k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3d1cfb72565ebd89f45e33846960a30a3d0ca23a/version-metadata.sh#L165)):*

I'm conflicted about what's best here. My "clean coder" side would prefer to have them as positional arguments, but my "anxiously afraid of future bugs" side would prefer to use global variables so that we don't make a mistake with the wrong order in the future :/

Rust shell scripts when? :joy: 

Anyway, I guess any of the three options is fine (third = leaving as is).
</blockquote></details>

"clean code" also usually does not recommend boolean arguments. Yeah I'm not a fan of sending all of them as positional arguments.

To at least make it symmetrical with `prepare-release.sh` I now moved argument parsing to the top, and consistently use only the "constants" instead of arguments. Is this fine? IMO it's the least bad compromise :)

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @jvff)

___
*[version-metadata.sh, line 162 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600#-MWEbadG1CumZMLWUIA0:-MWOCL4j4sp6eoTpkDDS:b9mjsqb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3d1cfb72565ebd89f45e33846960a30a3d0ca23a/version-metadata.sh#L162)):*

Nit: Would it make sense to output this to `stderr` with `>&2` at the end of the lines? Obviously since this is an internal tool it shouldn't matter, but I just wanted to mention it in case it's something worthwhile doing.
</blockquote></details>

Good idea. Pretty minor. But still nice! I also moved some of this stuff around for better symmetry.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @jvff)

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2600)* status: all files reviewed, 1 unresolved discussion (waiting on @jvff)

PR #2612, opened 2021-03-24
Change missing translation detection

jvff:
The translations converter tool had a few issues:

1. It would not load the last message of a gettext messages (or template) file
2. It would not unescape double-quotes in gettext messages
3. It would build a list of missing translations based only on translations, not on the translations template file

This PR fixes those issues by, respectively:

1. Adding an empty line to the list of parsed lines so that the "else" clause of the string matching loop is executed for the last iteration
2. Unescaping double-quotes in gettext messages when normalizing it. There's no need to escape it since when they are written the code uses `{:?}` which automatically escapes the double quotes
3. The list of missing translations (messages and plurals) is now built solely based on the messages template file, so that no duplicates are added and entries that are translated but not in the template file are also added.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Internal tool change, no changelog entry necessary.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2612)

### DIFF
--- a/android/translations-converter/src/gettext.rs
+++ b/android/translations-converter/src/gettext.rs
@@ -10,6 +10,7 @@ use std::{
 
 lazy_static! {
     static ref APOSTROPHE_VARIATION: Regex = Regex::new("’").unwrap();
+    static ref ESCAPED_DOUBLE_QUOTES: Regex = Regex::new(r#"\\""#).unwrap();
     static ref PARAMETERS: Regex = Regex::new(r"%\([^)]*\)").unwrap();
 }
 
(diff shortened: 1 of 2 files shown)


raksooo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2612#-:-MWZZWpMBD7gZm_4hppc:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2618, opened 2021-03-25
Fix deadlock in systemd-resolved DNS management

pinkisemils:
On some Ubuntu systems, it seems like the DNS watcher would deadlock - possibly because it's receiving too many DBus updates and the watcher thread is hogging the lock on the state. I've unloaded some of the responsibility from the _shared state_ so that no mutex is required to share the DNS config between threads and instead use an atomic bool to signal when the watcher thread should shut down. I'm still unsure if this whole cherade should be eschewed in favor of not reverting the config at all. I tested doing that, and it just works. My only concern would be that the watcher thread might not automatically stop. On the other hand, the watcher thread could always check if the link was removed - in which case it could stop.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618)

### DIFF
--- a/talpid-core/src/dns/linux/systemd_resolved.rs
+++ b/talpid-core/src/dns/linux/systemd_resolved.rs
@@ -1,13 +1,15 @@
 use crate::linux::{iface_index, IfaceIndexLookupError};
 use std::{
     net::IpAddr,
-    sync::{Arc, Mutex},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
     thread,
 };
 use talpid_dbus::systemd_resolved::{DnsState, SystemdResolved as DbusInterface};
 
 pub(crate) use talpid_dbus::systemd_resolved::Error as SystemdDbusError;
-use talpid_types::ErrorExt;
 
 pub type Result<T> = std::result::Result<T, Error>;
 
@@ -22,8 +24,13 @@ pub enum Error {
 
 pub struct SystemdResolved {
     pub dbus_interface: DbusInterface,
-    state: Option<Arc<Mutex<Option<DnsState>>>>,
-    watcher_thread: Option<thread::JoinHandle<()>>,
+    state: Option<SetConfigState>,
+}
+
+struct SetConfigState {
+    dns_config: Arc<DnsState>,
+    watcher_thread: thread::JoinHandle<()>,
+    watcher_should_shutdown: Arc<AtomicBool>,
 }
 
 
@@ -34,7 +41,6 @@ impl SystemdResolved {
         let systemd_resolved = SystemdResolved {
             dbus_interface,
             state: None,
-            watcher_thread: None,
         };
 
         Ok(systemd_resolved)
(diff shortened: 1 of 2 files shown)


pinkisemils:
I tested out a build where `SystemdResolved::reset()` wouldn't wait for the watcher thread to join, and it's faster of course.
But over the course of many hours of `while true; do mullvad disconnect; mullvad connect --wait; done;` the daemon kind of leaked 150MB of memory (and presumably threads). Once the thrashing is stopped, all the memory is reclaimed.  The code ends up being significantly simpler, but I'm not sure if we should allow this kind of behavior on customer machines. What are your thoughts?

dlon (commented):
> over the course of **many hours** of [...] the daemon kind of leaked 150MB of memory (and presumably threads)

Not too horrible. Is this still noticeably faster with the addition of `should_continue()` in the match callback?

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618#-:-MWi-0DxCPKvY7MKuBQ9:b-983d85)* status: all files reviewed, 3 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 93 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618#-MWhxjFx3uFB6ERRtIIU:-MWhxjFx3uFB6ERRtIIV:b-u4a5sa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7502bfb74ebb9947b3d5be2cc3652fbc4abeec2e/talpid-core/src/dns/linux/systemd_resolved.rs#L93)):*
> ```XML
>                     })();
>                 },
>                 move || !shutdown_now.load(Ordering::Acquire),
> ```

I think we can get away with `Ordering::Relaxed`. But I'm not certain.
___
*[talpid-dbus/src/systemd_resolved.rs, line 326 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618#-MWi1TulBKGXCE8x9qr_:-MWi1TulBKGXCE8x9qra:bd37076) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7502bfb74ebb9947b3d5be2cc3652fbc4abeec2e/talpid-dbus/src/systemd_resolved.rs#L326)):*
> ```XML
>                         }
>                     };
>                     should_continue()
> ```

I wonder if this should be checked first. Especially if the thread is joined *after* `should_shutdown = true` is set. Otherwise, might you not end up setting the config again after reverting it?
___
*[talpid-dbus/src/systemd_resolved.rs, line 340 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618#-MWhypSX0ksp_OizeLYE:-MWhypSX0ksp_OizeLYF:b-b6pxy3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7502bfb74ebb9947b3d5be2cc3652fbc4abeec2e/talpid-dbus/src/systemd_resolved.rs#L340)):*
> ```XML
>         match result {
>             Err(err) => {
>                 log::error!("Got error {:?}", err);
> ```

Isn't the error being logged twice now? Here and at the call site.

pinkisemils (commented):
It's a bit faster, but I haven't measured it. But it's all peanuts compared to not waiting for the thread at all.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618#-:-MWi8ouEA6w3QxLxKk86:b-1yckcc)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 93 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618#-MWhxjFx3uFB6ERRtIIU:-MWi91xW3Wor7Xc75QnE:bt1ai4a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7502bfb74ebb9947b3d5be2cc3652fbc4abeec2e/talpid-core/src/dns/linux/systemd_resolved.rs#L93)):*

I think we can get away with `Ordering::Relaxed`. But I'm not certain.
</blockquote></details>

`Relaxed` implies that there is no order. So loading in a relaxed fashion may well load a stale value, which is not what we want.
___
*[talpid-dbus/src/systemd_resolved.rs, line 326 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618#-MWi1TulBKGXCE8x9qr_:-MWi9MhG3M-66vFBp9Nb:bdeoa5d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7502bfb74ebb9947b3d5be2cc3652fbc4abeec2e/talpid-dbus/src/systemd_resolved.rs#L326)):*

I wonder if this should be checked first. Especially if the thread is joined *after* `should_shutdown = true` is set. Otherwise, might you not end up setting the config again after reverting it?
</blockquote></details>

A very good point, I'll fix the code. In the grander scheme of things, I don't really think it matters. 

Tangentially, I think the only purpose the `RevertLink` call now serves is to trigger the callback so that it gets removed quicker - would it not be better to have another watcher to listen for removals of an interface? Ultimately, if an interface gets removed, `systemd-resolved` will remove the config anyway, so us cleaning up really shouldn't matter.

dlon (approved):
I think I am fine with temporarily "leaking" resources while thrashing, if it makes resetting more responsive in the "nominal" case. But is there any way we might interrupt `dbus_connection.process` (which I assume is what is taking time) and close the thread sooner instead? Joining might not be necessary regardless.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618#-:-MWijPjFDUBZSU0P4RTO:b-vbxsls)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-dbus/src/systemd_resolved.rs, line 326 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618#-MWi1TulBKGXCE8x9qr_:-MWijsgA1PYbFQ_Kx9We:ba6c64u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7502bfb74ebb9947b3d5be2cc3652fbc4abeec2e/talpid-dbus/src/systemd_resolved.rs#L326)):*
Waiting for an interface to go down seems a bit wrong because a call to `DnsMonitor::reset` doesn't necessarily imply that a link will be destroyed, IMO. This is a doc comment for that method:

> Reset system DNS settings to what it was before being set by this instance.

pinkisemils (commented):
We could have another match that watches for deletions of the interface and once the interface is deleted, we could remove the match and stop watching. Then we don't need to join or revert the config.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618#-:-MWimcu_BXxPlNgmDIto:bx5o2qh)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

dlon (approved):
Sounds good.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2618#-:-MWinCA_C-BtCyOaYqa3:brxp9sq)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

PR #2620, opened 2021-03-26
Retry if sending problem reports fails

dlon:
The top address in `api-ip-address.txt` (ie the one last used by the service) is always preferred, and the problem report tool is not allowed to replace the cache. This meant that previously `mullvad-problem-report` would always use the same IP address to contact the API even if it failed using it before. Retrying a couple of times makes sure that two random addresses are used as well.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2620)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2620#-:-MWiGgX_EaeOzwLo0e4G:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2624, opened 2021-03-27
Add CI job for checking translations

jvff:
The process to handle adding translations to the app (desktop and Android) became more complex due to the requirement of keeping gettext files and Android resources synchronized. This PR aims to help with that problem by creating a new CI workflow that will fail in some situations where the translations aren't properly synchronized. More specifically, the workflow will fail in the following scenarios:

- adding a new message to the desktop GUI but not including that message in the template file (`messages.pot`)
- adding a new Android string resource that and either:
  - not adding the translations extracted from the desktop GUI locales for that message
  - not adding the message to the template file
- changing a message in the desktop GUI and not updating the template file or not updating the respective Android translations
- changing a string resource in the Android app and not updating its translations or its entry in the template file
- removing a message in the desktop GUI but not updating the template file or not updating the respective Android translations
- removing a string resource in the Android app and not removing its respective translations or its respective entry in the template file

If this job fails, the fix is to update the translations by running first `npm run update-translations` in the `gui` directory and then running the Android translation converter tool with `cargo run` in the `android/translations-converter` directory.

As part of the PR, the translations converter tool was updated to become more deterministic. The entries appended to the template file are now sorted so that they are always added in the same order. Also, the template file and the Android resources were updated.

A small bug in the tool was also fixed, where the single quote `'` was being incorrectly escaped in strings appended to the template file. The fix was to manually escape only the double quotes `"` instead of relying on `fmt::Debug` to do the escaping for us.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Internal tooling changes, no chaneglog entry necessary.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2624)

### DIFF
--- a/.github/workflows/daemon.yml
+++ b/.github/workflows/daemon.yml
@@ -5,6 +5,7 @@ on:
         paths-ignore:
             - '**/*.md'
             - .github/workflows/android*.yml
+            - .github/workflows/ios.yml
             - .github/workflows/rustfmt.yml
             - android/**
             - audits/**
--- a/.github/workflows/daemon.yml
+++ b/.github/workflows/daemon.yml
@@ -8,6 +8,7 @@ on:
             - .github/workflows/frontend.yml
             - .github/workflows/ios.yml
             - .github/workflows/rustfmt.yml
+            - .github/workflows/translations.yml
             - android/**
             - audits/**
             - ci/buildserver-*
--- a/.github/workflows/daemon.yml
+++ b/.github/workflows/daemon.yml
@@ -5,6 +5,7 @@ on:
         paths-ignore:
             - '**/*.md'
             - .github/workflows/android*.yml
+            - .github/workflows/frontend.yml
             - .github/workflows/ios.yml
             - .github/workflows/rustfmt.yml
             - android/**
--- a/.github/workflows/translations.yml
+++ b/.github/workflows/translations.yml
@@ -1,0 +1,50 @@
+name: Translation check CI
+on:
+    # Check whenever a file that affects Rust formatting is changed by a push
+    push:
+        paths:
+            - .github/workflows/translations.yml
+            - android/translations-converter/**
+            - android/src/**/plurals.xml
+            - android/src/**/strings.xml
+            - gui/**
+    # Check if requested manually from the Actions tab
+    workflow_dispatch:
+jobs:
+    check-translations:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v2
+
+            - name: Install Node.js
+              uses: actions/setup-node@v2
+              with:
+                node-version: '12'
+
+            - name: Update NPM
+              run: npm i -g npm
+
+            - name: Install and cache JS dependencies
+              uses: bahmutov/npm-install@v1
+              with:
+                  working-directory: gui
+                  install-command: npm ci
+
+            - name: Install nightly Rust
+              uses: ATiltedTree/setup-rust@v1.0.4
+              with:
+                rust-version: stable
+
+            - name: Extract messages from desktop GUI
+              working-directory: gui/locales
+              run: npm run update-translations
+
+            - name: Convert translations into Android resources
+              working-directory: android/translations-converter
+              run: cargo run
+
+            - name: Check if repository is up to date
+              run: |
+                git diff
+                ! git status -s | grep .
--- a/android/src/main/res/values-da/strings.xml
+++ b/android/src/main/res/values-da/strings.xml
@@ -115,7 +115,6 @@
     <string name="update_available">OPDATERING TILGÆNGELIG</string>
     <string name="update_available_footer">Opdatering tilgængelig, download den for at forblive sikker.</string>
     <string name="user_email_hint">Din e-mail (valgfrit)</string>
-    <string name="user_message_hint">Beskriv dit problem</string>
     <string name="view_logs">Se app-logfiler</string>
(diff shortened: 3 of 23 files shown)


jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2624)* status: 0 of 23 files reviewed, all discussions resolved

___
*[.github/workflows/daemon.yml, line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2624#-MWpCEb6AdTIy6WgMP03:-MWpCEb72HV-OuLS693Q:b-ro7lx1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2a4d0a1febd93c772f40a63ea3cf2256ab67bf94/.github/workflows/daemon.yml#L11)):*
> ```YAML
>             - .github/workflows/ios.yml
>             - .github/workflows/rustfmt.yml
>             - .github/workflows/translations.yml
> ```

FYI: The `daemon` workflow was being triggered to much, so I added some more paths for it to ignore.

raksooo (changes requested):
> Reviewed 23 of 23 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2624)* status: all files reviewed, 2 unresolved discussions (waiting on @jvff)

___
*[.github/workflows/translations.yml, line 9 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2624#-MWuNQdk2-Lbyhknpeoi:-MWuNQdk2-Lbyhknpeoj:b-ppuwnq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2a4d0a1febd93c772f40a63ea3cf2256ab67bf94/.github/workflows/translations.yml#L9)):*
> ```YAML
>             - android/translations-converter/**
>             - android/src/**/plurals.xml
>             - android/src/**/strings.xml
> ```

Are strings on Android manually added to `plurals.xml` or `strings.xml`? Or are these generated?
___
*[.github/workflows/translations.yml, line 50 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2624#-MWuIYhs9ei9LYUaxupg:-MWuIYhs9ei9LYUaxuph:bqtm3ff) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2a4d0a1febd93c772f40a63ea3cf2256ab67bf94/.github/workflows/translations.yml#L50)):*
> ```YAML
>               run: |
>                 git diff
>                 ! git status -s | grep .
> ```

Does `grep .` make any difference here? Shouldn't just `! git status -s` work?

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2624)* status: all files reviewed, 2 unresolved discussions (waiting on @raksooo)

___
*[.github/workflows/translations.yml, line 9 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2624#-MWuNQdk2-Lbyhknpeoi:-MWxq0LN4alFpM9CGLl3:b-3txcrk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2a4d0a1febd93c772f40a63ea3cf2256ab67bf94/.github/workflows/translations.yml#L9)):*

Are strings on Android manually added to `plurals.xml` or `strings.xml`? Or are these generated?
</blockquote></details>

Strings are manually added to `android/src/*/res/values/{strings,plurals}.xml`, but the translations that are in `android-src/*/res/values-<LOCALE>/{strings,plurals}.xml` are automatically generated by the tool.

I added both to the trigger since I want to run if a string is manually added and/or if the translations change.
___
*[.github/workflows/translations.yml, line 50 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2624#-MWuIYhs9ei9LYUaxupg:-MWxqh9eCG3sWASQv8pV:b274ypa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2a4d0a1febd93c772f40a63ea3cf2256ab67bf94/.github/workflows/translations.yml#L50)):*

Does `grep .` make any difference here? Shouldn't just `! git status -s` work?
</blockquote></details>

Unfortunately, `git status -s` doesn't return an error in the cases we need (AFAICT, it never returns an error unless the repo meta-data is corrupt I think). So I use `grep .` to get the error status.

What the command really does is `git status -s | ! grep .` but that's invalid syntax :laughing:

raksooo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2624#-:-MWy6sbA516712tP8uHg:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[.github/workflows/translations.yml, line 50 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2624#-MWuIYhs9ei9LYUaxupg:-MWy6pPm-g_fXiHe2ptK:b-cc7ba7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2a4d0a1febd93c772f40a63ea3cf2256ab67bf94/.github/workflows/translations.yml#L50)):*

Unfortunately, `git status -s` doesn't return an error in the cases we need (AFAICT, it never returns an error unless the repo meta-data is corrupt I think). So I use `grep .` to get the error status.

What the command really does is `git status -s | ! grep .` but that's invalid syntax :laughing: 
</blockquote></details>

Okay :+1:

PR #2627, opened 2021-03-29
Remove use_fwmark option

dlon:
If this setting is set to true, the firewall only permits packets with a particular fwmark to communicate outside the tunnel with the relay server. The alternative is that processes that run as root are allowed to talk to the relay.

It is no longer useful to set this to `false`, not even for bridges since we exclude `sslocal` from the tunnel. This change does mean that one must use `mullvad-exclude` (or, preferably, set the fwmark directly to `talpid_core::linux::TUNNEL_FW_MARK`) to run one's proxy server, or else the firewall will block it. But in practice, this is already the easiest way to go, because the routes are not automatically set up to deal with the other case.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2627)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,10 @@ Line wrap the file at 100 chars.                                              Th
 #### Windows
 - Upgrade Wintun from 0.9.2 to 0.10.1.
 
+#### Linux
+- Only allow packets with the mark set to `0x6d6f6c65` to communicate with the relay server.
+  Previously, bridges were expected to run as root instead.
+
 ### Fixed
 - Fix delay in showing/hiding update notification when toggling beta program.
 - Improve responsiveness when reconnecting after some failed connection attempts.
--- a/mullvad-cli/src/cmds/bridge.rs
+++ b/mullvad-cli/src/cmds/bridge.rs
@@ -66,31 +66,48 @@ fn create_bridge_set_subcommand() -> clap::App<'static, 'static> {
 
 
 fn create_set_custom_settings_subcommand() -> clap::App<'static, 'static> {
+    #[allow(unused_mut)]
+    let mut local_subcommand = clap::SubCommand::with_name("local")
+        .about("Registers a local SOCKS5 proxy")
+        .arg(
+            clap::Arg::with_name("local-port")
+                .help("Specifies the port the local proxy server is listening on")
+                .required(true)
+                .index(1),
+        )
+        .arg(
+            clap::Arg::with_name("remote-ip")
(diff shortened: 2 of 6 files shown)


pinkisemils (approved):
The changelog definitely needs an update. And maybe the README could have a section about this too. Maybe the _help_ section for applying a custom bridge in the CLI could mention the fwmark/exclusion requirement too.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2627#-:-MWyCREi2gkzMONUoR56:b25s7lv)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2627)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[CHANGELOG.md, line 64 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2627#-MX6swOL6V9aEWYlhzgX:-MX6swOM87qddPU21Xdv:b-3k83t8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/be2e7fd4fb140080840e156479cdd5f3fddc2d2c/CHANGELOG.md#L64)):*
> ```Markdown
> #### Linux
> - Only allow packets with the mark set to `0x6d6f6c65` to communicate with the relay server.
>   Previously, bridges were expected to run as root instead.
> ```

I think setting `SO_MARK` socket option requires root privileges too. Or at least `CAP_NET_ADMIN`, unless ran with our setuid tool.
___
*[mullvad-cli/src/cmds/bridge.rs, line 95 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2627#-MX6s0YW3xYCu5bOoCLk:-MX6s0YW3xYCu5bOoCLl:br6yj5h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/be2e7fd4fb140080840e156479cdd5f3fddc2d2c/mullvad-cli/src/cmds/bridge.rs#L95)):*
> ```XML
>         local_subcommand = local_subcommand.about(
>             "Registers a local SOCKS5 proxy. The server must be excluded using \
>            'mullvad-exclude', or its packet marks must be set to '0x6d6f6c65', in order \
> ```

I think it'd be clearer to refer to the socket option rather than just the packets. But that might be too specific again.

dlon:
> I think it'd be clearer to refer to the socket option rather than just the packets. But that might be too specific again.

Done.

PR #2630, opened 2021-03-30
Only keep a WireGuard key for the active account

dlon:
As above. This means that unsetting or logging out causes the key to be removed from the account.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Line wrap the file at 100 chars.                                              Th
 - Only download a new relay list if it has been modified.
 - Connect to the API only via TLS 1.3
 - Shrink account history capactity from 3 account entries to 1.
+- Remove WireGuard key from account history when logging out.
 - Allow whitespace in account token in CLI.
 - Read account token from standard input unless given as an argument in CLI.
 - Make WireGuard automatic key rotation interval mandatory and between 1 and 7 days.
--- a/mullvad-daemon/src/account_history.rs
+++ b/mullvad-daemon/src/account_history.rs
@@ -223,6 +223,20 @@ impl AccountHistory {
         self.save_to_disk().await
     }
 
+    /// Removes WireGuard keys associated with all accounts
+    pub async fn clear_keys(&mut self) -> Result<()> {
+        log::debug!("account_history::clear_keys");
+
+        for entry in self.accounts.lock().unwrap().iter_mut() {
+            if let Some(wg_data) = &entry.wireguard {
+                tokio::spawn(self.create_remove_wg_key_rpc(&entry.account, &wg_data));
+                entry.wireguard = None;
+            }
+        }
+
+        self.save_to_disk().await
+    }
+
     /// Remove account history
     pub async fn clear(&mut self) -> Result<()> {
         log::debug!("account_history::clear");
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -1471,14 +1471,20 @@ where
             self.event_listener
                 .notify_settings(self.settings.to_settings());
 
-            // Bump account history if a token was set
-            if let Some(token) = account_token.clone() {
-                if let Err(e) = self.account_history.bump_history(&token).await {
-                    log::error!("Failed to bump account history: {}", e);
-                }
+            if let Err(error) = self.account_history.clear_keys().await {
+                log::error!(
+                    "{}",
+                    error.display_chain_with_msg("Error while clearing account history keys")
+                );
             }
 
-            self.ensure_wireguard_keys_for_current_account().await;
+            if let Some(token) = &account_token {
+                // Bump account history if a token was set
+                if let Err(e) = self.account_history.bump_history(token).await {
+                    log::error!("Failed to bump account history: {}", e);
+                }
+                self.ensure_wireguard_keys_for_current_account().await;
+            }
         }
         Ok(account_changed)
     }


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[mullvad-daemon/src/account_history.rs, line 232 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630#-MX1qtImE9yB9DOL4140:-MX1qtImE9yB9DOL4141:bxoxj6f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0dda9c9fbae150888cab278020e5c181d7c82a27/mullvad-daemon/src/account_history.rs#L232)):*
> ```XML
>         for entry in self.accounts.lock().unwrap().iter_mut() {
>             if let Some(wg_data) = &entry.wireguard {
>                 tokio::spawn(self.create_remove_wg_key_rpc(&entry.account, &wg_data));
> ```

Spawning tasks and ignoring the outcomes is obviously faster than waiting on them. The annoying thing about it, though, is that the local history becomes out of sync with the real account history if any RPC fails.

This can probably happen pretty easily if the tunnel disconnects immediately afterwards (as it does), or randomly fails to reach the API.

I'm not sure which is preferable.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[mullvad-daemon/src/account_history.rs, line 232 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630#-MX1qtImE9yB9DOL4140:-MX1tfjTDTXEzI-LnGt2:b6lkh0e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0dda9c9fbae150888cab278020e5c181d7c82a27/mullvad-daemon/src/account_history.rs#L232)):*

Spawning tasks and ignoring the outcomes is obviously faster than waiting on them. The annoying thing about it, though, is that the local history becomes out of sync with the real account history if any RPC fails.

This can probably happen pretty easily if the tunnel disconnects immediately afterwards (as it does), or randomly fails to reach the API.

I'm not sure which is preferable.
</blockquote></details>

We should not over-complicate any solution in this area right now. Since we will re-work it quite a lot soon. I think this best effort approach is good enough for now.
___
*[mullvad-daemon/src/lib.rs, line 1477 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630#-MX1twEcD2BrIjtyt7O7:-MX1twEeBLS8p59tzsBV:b5kuypf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0dda9c9fbae150888cab278020e5c181d7c82a27/mullvad-daemon/src/lib.rs#L1477)):*
> ```XML
>                 log::error!(
>                     "{}",
>                     error.display_chain_with_msg("Error while setting account token")
> ```

Setting account token? The only error that can end up here is the `save_to_disk()`, no?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/lib.rs, line 1477 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630#-MX1twEcD2BrIjtyt7O7:-MX1vnfy7vsEg3n5lVGQ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0dda9c9fbae150888cab278020e5c181d7c82a27/mullvad-daemon/src/lib.rs#L1477)):*

Setting account token? The only error that can end up here is the `save_to_disk()`, no?
</blockquote></details>

Done.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/account_history.rs, line 232 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630#-MX1qtImE9yB9DOL4140:-MX25iFh7FSdHHpBHLXx:b-esw6q7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0dda9c9fbae150888cab278020e5c181d7c82a27/mullvad-daemon/src/account_history.rs#L232)):*

We should not over-complicate any solution in this area right now. Since we will re-work it quite a lot soon. I think this best effort approach is good enough for now.
</blockquote></details>

The risk here is that the user won't be able to connect when they run out of keys.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630)* status: 2 of 3 files reviewed, all discussions resolved

___
*[mullvad-daemon/src/account_history.rs, line 232 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630#-MX1qtImE9yB9DOL4140:-MX2ECIL8dYlBObkQkWr:b-8df2dq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0dda9c9fbae150888cab278020e5c181d7c82a27/mullvad-daemon/src/account_history.rs#L232)):*

The risk here is that the user won't be able to connect when they run out of keys. 
</blockquote></details>

When this is reworked, I think we'll want to want to write to the file only if the RPC actually succeeds. Probably by not spawning the RPC as a separate task.

If this is unacceptable to the frontends, the daemon should then be able to call `tokio::spawn(account_history.remove_account(...))`. I think we may actually want this to be synchronous in `mullvad-setup`, though, so that failures to remove keys while uninstalling can at least be logged.

Anyway, for now I followed the same pattern as `remove_account` and `insert`.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[mullvad-daemon/src/account_history.rs, line 232 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2630#-MX1qtImE9yB9DOL4140:-MX7Ly1aCj-pupxY6nBu:b35lige) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0dda9c9fbae150888cab278020e5c181d7c82a27/mullvad-daemon/src/account_history.rs#L232)):*

When this is reworked, I think we'll want to want to write to the file only if the RPC actually succeeds. Probably by not spawning the RPC as a separate task.

If this is unacceptable to the frontends, the daemon should then be able to call `tokio::spawn(account_history.remove_account(...))`. I think we may actually want this to be synchronous in `mullvad-setup`, though, so that failures to remove keys while uninstalling can at least be logged.

Anyway, for now I followed the same pattern as `remove_account`.
</blockquote></details>

There should at least be a way to signal to the frontend that keys may have not been removed. It doesn't have to block the logout forever.

dlon:
Superseded by #2801.

PR #2635, opened 2021-03-31
Stop logging ENETUNREACH errors when deducing geoip.

pinkisemils:
It's a burden to any consumers of the daemon logs to read through 4 lines of logging about how an IPv6 address was unreachable everytime the daemon connects and the current location and exit IPs are requested. As such, I've changed the logging code to not log `libc::ENETUNREACH` errors.

As an aside, I've also witnessed `ECONNREFUSED` when running the client from within a Linux VM. And whenever I initiated a reconnect on Windows via the GUI, both geoip requests would fail. But they would succeed later on, if the location is requested via `mullvad status --location`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2635)

### DIFF
--- a/mullvad-daemon/Cargo.toml
+++ b/mullvad-daemon/Cargo.toml
@@ -16,6 +16,7 @@ fern = { version = "0.6", features = ["colored"] }
 futures = "0.3"
 ipnetwork = "0.16"
 lazy_static = "1.0"
+libc = "0.2"
 log = "0.4"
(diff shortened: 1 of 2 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2635)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2656, opened 2021-04-12
Split `CustomDns` into service-side and UI-side classes

jvff:
This PR continues the work to split the Android app so that the service runs in a separate process. This PR focuses on the custom DNS functionality, splitting the `CustomDns` class in two, one for the service side and one for the UI side. The UI side class sends requests to add/remove/replace custom DNS servers, and to enable or disable the functionality. The service side then handles those requests.

The service side class does not need to send any events to the UI side `CustomDns` class because they are included in settings updates sent by the `SettingsListener`.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2656)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ipc/Request.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ipc/Request.kt
@@ -2,12 +2,16 @@ package net.mullvad.mullvadvpn.ipc
 
 import android.os.Message as RawMessage
 import android.os.Messenger
+import java.net.InetAddress
 import kotlinx.parcelize.Parcelize
 import org.joda.time.DateTime
 
 // Requests that the service can handle
 sealed class Request : Message.RequestMessage() {
     protected override val messageKey = MESSAGE_KEY
+
+    @Parcelize
+    data class AddCustomDnsServer(val address: InetAddress) : Request()
 
     @Parcelize
     object Connect : Request()
@@ -47,6 +51,18 @@ sealed class Request : Message.RequestMessage() {
 
     @Parcelize
     data class RemoveAccountFromHistory(val account: String?) : Request()
+
+    @Parcelize
+    data class RemoveCustomDnsServer(val address: InetAddress) : Request()
+
+    @Parcelize
+    data class ReplaceCustomDnsServer(
+        val oldAddress: InetAddress,
+        val newAddress: InetAddress
+    ) : Request()
+
+    @Parcelize
+    data class SetEnableCustomDns(val enable: Boolean) : Request()
 
     @Parcelize
     data class SetEnableSplitTunneling(val enable: Boolean) : Request()
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/CustomDns.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/CustomDns.kt
@@ -36,29 +36,13 @@ class CustomDns(private val endpoint: ServiceEndpoint) {
             }
 
             registerHandler(Request.SetEnableCustomDns::class) { request ->
-                if (request.enable) {
-                    enable()
-                } else {
-                    disable()
-                }
+                setEnabled(request.enable)
             }
         }
     }
 
     fun onDestroy() {
         endpoint.settingsListener.dnsOptionsNotifier.unsubscribe(this)
-    }
-
-    fun enable() {
-        synchronized(this) {
-            changeDnsOptions(true, dnsServers)
-        }
-    }
-
-    fun disable() {
-        synchronized(this) {
-            changeDnsOptions(false, dnsServers)
-        }
     }
 
     fun addDnsServer(server: InetAddress): Boolean {
@@ -100,6 +84,10 @@ class CustomDns(private val endpoint: ServiceEndpoint) {
                 changeDnsOptions(enabled, dnsServers)
             }
         }
+    }
+
+    fun setEnabled(enable: Boolean) {
+        changeDnsOptions(enable, dnsServers)
     }
 
     private fun changeDnsOptions(enable: Boolean, dnsServers: ArrayList<InetAddress>) {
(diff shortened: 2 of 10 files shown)


axti (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2656)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/endpoint/CustomDns.kt, line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2656#-MY5yIxO6O_QPoFHRdS3:-MY5yIxO6O_QPoFHRdS4:b-kps2ly) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/38bb3ef000f3aec89400e37aff74898082122530/android/src/main/kotlin/net/mullvad/mullvadvpn/service/endpoint/CustomDns.kt#L27)):*
> ```
> private var dnsServers = ArrayList<InetAddress>()
> ```
I think it is `val` actually and maybe use HashSet, what do you think?

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2656)* status: all files reviewed, 1 unresolved discussion (waiting on @jvff)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/endpoint/CustomDns.kt, line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2656#-MY5yIxO6O_QPoFHRdS3:-MY6HpnM8nyBycBTzSB3:btw45nb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/38bb3ef000f3aec89400e37aff74898082122530/android/src/main/kotlin/net/mullvad/mullvadvpn/service/endpoint/CustomDns.kt#L27)):*
<details><summary><i>Previously, axti (Alex Granin) wrote…</i></summary><blockquote>

> ```
> private var dnsServers = ArrayList<InetAddress>()
> ```
I think it is `val` actually and maybe use HashSet, what do you think? 
</blockquote></details>

Yeah, The problem with the `HashSet` is that it loses the insertion order, and I think that's important when passing the information to the tunnel. If I'm not mistaken, the device will try the DNS servers in order.

I'm not really sure what happens with repeated addresses. I'd expect it to be only used once, but it could retry it on failure. But that sounds very hacky, we should limit it to be more predictable, and not allow repeated addresses. Maybe `LinkedHashSet` could work here. I'll try that.

I made it `var` because it would change when a new list is received (i.e., when the `settingsListener` notifies of a new `DnsOptions`). But that seems like incorrect semantics, because we later change that list and other parts of the code might not be expecting that. I'll change it to be a `val` and clear it and add only the new elements when a new `DnsOptions` event is received.

Thanks for the heads-up! I'll see if I can fix the issues :+1:

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2656)* status: 9 of 10 files reviewed, all discussions resolved (waiting on @axti)

___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/endpoint/CustomDns.kt, line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2656#-MY5yIxO6O_QPoFHRdS3:-MY6UwYBEJCLaRZXn4mu:bhgs2bn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/38bb3ef000f3aec89400e37aff74898082122530/android/src/main/kotlin/net/mullvad/mullvadvpn/service/endpoint/CustomDns.kt#L27)):*

Yeah, The problem with the `HashSet` is that it loses the insertion order, and I think that's important when passing the information to the tunnel. If I'm not mistaken, the device will try the DNS servers in order.

I'm not really sure what happens with repeated addresses. I'd expect it to be only used once, but it could retry it on failure. But that sounds very hacky, we should limit it to be more predictable, and not allow repeated addresses. Maybe `LinkedHashSet` could work here. I'll try that.

I made it `var` because it would change when a new list is received (i.e., when the `settingsListener` notifies of a new `DnsOptions`). But that seems like incorrect semantics, because we later change that list and other parts of the code might not be expecting that. I'll change it to be a `val` and clear it and add only the new elements when a new `DnsOptions` event is received.

Thanks for the heads-up! I'll see if I can fix the issues :+1:
</blockquote></details>

Done. But I couldn't use a `LinkedHashSet`, because it doesn't provide an easy way to replace a custom DNS server in a specific position. This makes it harder for the `CustomDnsAdapter` to change an already added custom DNS server. Another issue I ran into was that the Rust side expects a Rust `Vec`, which is mapped to a Java `ArrayList`, so for the JNI layer it still needed to be an `ArrayList`. I figured it would be best to leave it as an `ArrayList` instead of working around these two issues.

But I think it still might make sense to take a look if the other platforms supported by Rust would benefit from using some sort of `HashSet` to avoid repeated custom DNS address entries.
___
*[android/src/main/kotlin/net/mullvad/mullvadvpn/service/endpoint/CustomDns.kt, line 108 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2656#-MY6Vnnl2V58mAVxAOg5:-MY6Vnnl2V58mAVxAOg6:b-2xc18g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/38bb3ef000f3aec89400e37aff74898082122530/android/src/main/kotlin/net/mullvad/mullvadvpn/service/endpoint/CustomDns.kt#L108)):*
> ```Kotlin
>     }
> 
>     private suspend fun changeDnsOptions(enable: Boolean, dnsServers: ArrayList<InetAddress>) {
> ```

FYI: I realized it didn't make much sense to have the second parameter, because all call sites were just setting it to the property value.

axti (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2656)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> Reviewed 9 of 10 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2656)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2673, opened 2021-04-15
Add support for building universal macOS app

raksooo:
This PR adds support for cross compiling the app for the M1 macs on x86 macs as well as support for making universal builds.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,8 @@
 /dist-assets/sslocal
 /dist-assets/sslocal.exe
 /dist-assets/shell-completions/
+/dist-assets/aarch64-apple-darwin/
+/dist-assets/x86_64-apple-darwin/
 /windows/**/bin/
 /windows/**/*.user
 /android/keystore.properties
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Line wrap the file at 100 chars.                                              Th
 - When `MULLVAD_MANAGEMENT_SOCKET_GROUP` is set, only allow the specified group to access the
   management interface UDS socket. This means that only users in that group can use the CLI and GUI.
 - Support WireGuard over TCP for custom VPN relays in the CLI.
+- Make app native on Apple Silicon.
 
 ### Changed
 - Upgrade OpenVPN from 2.5.0 to 2.5.1.
--- a/README.md
+++ b/README.md
@@ -296,7 +296,12 @@ This should produce an installer exe, pkg or rpm+deb file in the `dist/` directo
 
 Building this requires at least 1GB of memory.
 
-#### Apple ARM64 (aka Apple Silicon)
+#### macOS
+
+By default a universal app is built on macOS. To build specifically for x86_64 or ARM64 add
+`--target x86_64-apple-darwin` or `--target aarch64-apple-darwin`.
+
+##### Apple ARM64 (aka Apple Silicon)
 
 Due to inability to build the management interface proto files on ARM64 (see
 [this](https://github.com/grpc/grpc-node/issues/1497) issue) the Apple ARM64 build must be done in 2 stages:
--- a/build.sh
+++ b/build.sh
@@ -15,15 +15,34 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd "$SCRIPT_DIR"
 RUSTC_VERSION=$(rustc +stable --version)
 PRODUCT_VERSION=$(node -p "require('./gui/package.json').version" | sed -Ee 's/\.0//g')
-CARGO_TARGET_DIR=${CARGO_TARGET_DIR:-"$SCRIPT_DIR/target"}
+CARGO_TARGET_DIR=${CARGO_TARGET_DIR:-"target"}
 
 CARGO_ARGS=()
 NPM_PACK_ARGS=()
 
-source env.sh
+BUILD_MODE="release"
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --dev-build) BUILD_MODE="dev" ;;
+        --target)
+            TARGET=("$2")
+            shift
+            ;;
+        *)
+            echo "Unknown parameter: $1"
+            exit 1
+            ;;
+    esac
+    shift
+done
 
-if [[ "${1:-""}" != "--dev-build" ]]; then
-    BUILD_MODE="release"
+if [[ "$(uname -s)" == "Darwin" && -z ${TARGET:-""} ]]; then
+    echo "Defaulting to universal macOS target since no target was provided"
+    TARGET=(x86_64-apple-darwin aarch64-apple-darwin)
+    NPM_PACK_ARGS+=(--universal)
+fi
+
+if [[ "$BUILD_MODE" == "release" ]]; then
     if [[ $(git diff --shortstat 2> /dev/null | tail -n1) != "" ]]; then
         echo "Dirty working directory!"
         echo "You should only build releases in clean working directories in order to make it"
@@ -57,7 +76,6 @@ if [[ "${1:-""}" != "--dev-build" ]]; then
         export CSC_IDENTITY_AUTO_DISCOVERY=false
     fi
 else
-    BUILD_MODE="dev"
     NPM_PACK_ARGS+=(--no-compression)
     echo "!! Development build. Not for general distribution !!"
     unset CSC_LINK CSC_KEY_PASSWORD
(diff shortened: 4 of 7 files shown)


faern (changes requested):
> Reviewed 4 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673)* status: 4 of 7 files reviewed, 4 unresolved discussions (waiting on @raksooo)

___
*[README.md, line 301 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYL5ny_8HbxvViVDw8L:-MYL5ny_8HbxvViVDw8M:bvfeecc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/eddfff2e2e2ccf7091d9ea9ec98892d10e643854/README.md#L301)):*
> ```Markdown
> #### macOS
> 
> By default a universal app is build on macOS. To build specifically for x86_64 or ARM64 add
> ```

s/build/built
___
*[gui/tasks/distribution.js, line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYL3EE4BACTH3Rkaxsa:-MYL3EE594n5C0TrChcy:bv9n8wd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/eddfff2e2e2ccf7091d9ea9ec98892d10e643854/gui/tasks/distribution.js#L1)):*
> ```JavaScript
> const cp = require('child_process');
> ```

Is `cp` not very strongly associated with copying?
___
*[gui/tasks/distribution.js, line 284 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYL65r40aC5nFsPdxYH:-MYL65r40aC5nFsPdxYI:b-19sagw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/eddfff2e2e2ccf7091d9ea9ec98892d10e643854/gui/tasks/distribution.js#L284)):*
> ```JavaScript
> function getMacThirdPartyBinaryPath(relativePath) {
>   if (universal) {
>     return getMacBinaryPath(relativePath);
> ```

Why are the third party binaries under `dist-assets/$arch` when universal but under `dist-assets/binaries/$target_triple` when not universal? Should we not always get the third party binaries from the binaries submodule?
___
*[gui/tasks/distribution.js, line 286 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYL59Zl72b61HFkr5FD:-MYL59Zm1n82PURZStbM:bwxusjq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/eddfff2e2e2ccf7091d9ea9ec98892d10e643854/gui/tasks/distribution.js#L286)):*
> ```JavaScript
>     return getMacBinaryPath(relativePath);
>   } else {
>     const arch = process.arch === 'arm64' || arm64 ? 'aarch64' : 'x86_64';
> ```

Here we compute a local `arch`. But the function above seems to use a global `${arch}`. I find that confusing, but that's maybe just because I'm not that used to this file. Especially since they seem to have different types of values. The other one is one of `x64` or `arm64`.

Also, why do we need to check both the `arm64` variable *and* `process.arch`?

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673)* status: 2 of 7 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[gui/tasks/distribution.js, line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYL3EE4BACTH3Rkaxsa:-MYPFmkL2Nb-An37JZSo:b-p07wyv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/eddfff2e2e2ccf7091d9ea9ec98892d10e643854/gui/tasks/distribution.js#L1)):*

Is `cp` not very strongly associated with copying?
</blockquote></details>

It's commonly used as an abbreviation in JavaScript/TypeScript and is even used in the official documentation. But it turned out to be unused here so I've removed it.
___
*[gui/tasks/distribution.js, line 284 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYL65r40aC5nFsPdxYH:-MYPHaK_5t6KLRG_wB3t:b-vg3jkg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/eddfff2e2e2ccf7091d9ea9ec98892d10e643854/gui/tasks/distribution.js#L284)):*

Why are the third party binaries under `dist-assets/$arch` when universal but under `dist-assets/binaries/$target_triple` when not universal? Should we not always get the third party binaries from the binaries submodule?
</blockquote></details>

Described more thoroughly further down. In short it was due to the `${arch}` macro having the values `x64` and `arm64` which didn't match the target-triples.
___
*[gui/tasks/distribution.js, line 286 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYL59Zl72b61HFkr5FD:-MYPGDs5CkF-bjoQputy:be0hoz0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/eddfff2e2e2ccf7091d9ea9ec98892d10e643854/gui/tasks/distribution.js#L286)):*

Here we compute a local `arch`. But the function above seems to use a global `${arch}`. I find that confusing, but that's maybe just because I'm not that used to this file. Especially since they seem to have different types of values. The other one is one of `x64` or `arm64`.

Also, why do we need to check both the `arm64` variable *and* `process.arch`?
</blockquote></details>

This is/was a mess and is indeed confusing. To instruct electron bulider to find the correct binaries for each platform we have to use the file macros which are very limited. The `"${arch}"` macro is either `arm64` or `x64` which doesn't match our target-triples and I therefore had to move all binaries to a directories with those names. When making a non-universal build we can refer to the correct directory without using a file-macro since the arch is the same for everything.

The reason to why we needed both `arm64` and `process.arch` was that `arm64` initially wasn't set when building on the M1 macs.

I've now worked around this by setting an environment variable, containing the target triple, before packaging the binaries for each arch. This is still a bit hacky but a bit cleaner and hopefully much easier to understand.

faern (changes requested):
> Reviewed 1 of 7 files at r1, 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673)* status: 4 of 7 files reviewed, 1 unresolved discussion (waiting on @faern and @raksooo)

___
*[gui/tasks/distribution.js, line 289 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYQAgYp9HpfefRY1_nP:-MYQAgYq096iA9yc1p-Z:b-q7pnmd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/60103523917e1f6e098ffa305709d263500db67f/gui/tasks/distribution.js#L289)):*
> ```JavaScript
>     return ['arm64'];
>   } else {
>     return undefined;
> ```

When is it undefined? When not building universal on an Intel mac is my interpretation? Can we document this function a bit better to explain this to the user?

Also, the return value seems to be a list (but always a single entry). Can the documentation then also describe why it's a list please.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673)* status: 4 of 7 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[gui/tasks/distribution.js, line 289 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYQAgYp9HpfefRY1_nP:-MYdAr_U86AYk5N1OEQ0:b-3zcwgc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/60103523917e1f6e098ffa305709d263500db67f/gui/tasks/distribution.js#L289)):*

When is it undefined? When not building universal on an Intel mac is my interpretation? Can we document this function a bit better to explain this to the user?

Also, the return value seems to be a list (but always a single entry). Can the documentation then also describe why it's a list please.
</blockquote></details>

Done. It works with a string if there's only one value :+1:

faern (changes requested):
> Reviewed 2 of 4 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673)* status: all files reviewed, 2 unresolved discussions (waiting on @raksooo)

___
*[build.sh, line 172 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYjhGG681Pt4tWReBk_:-MYjhGG681Pt4tWReBka:bj829h7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/69a70d87df033da319f260849b1c789bc0b25404/build.sh#L172)):*
> ```Shell
> function build() {
>     local current_target=${1:-""}
>     local forTargetString=""
> ```

Inconsistent variable naming style. Please use snake_case.
___
*[build.sh, line 251 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYjehp33q7d0SPVmlqa:-MYjehp4-wnmFRH4QnAu:b-uddvx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/69a70d87df033da319f260849b1c789bc0b25404/build.sh#L251)):*
> ```Shell
> 
> if [[ "$(uname -s)" == "Darwin" || "$(uname -s)" == "Linux" ]]; then
>     mkdir -p "$SCRIPT_DIR/dist-assets/shell-completions"
> ```

Nitpick. Since we `cd` into the `$SCRIPT_DIR` at the very start of the script I think you can use only relative paths. For some places you need absolute paths maybe, but for running this command it should be fine I think. On the other hand it makes sense to be consistent with the cargo invocation below. Does that one need absolute path?

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673)* status: 5 of 7 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[build.sh, line 172 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYjhGG681Pt4tWReBk_:-MYnMbJw2xUoeszftPIo:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/69a70d87df033da319f260849b1c789bc0b25404/build.sh#L172)):*

Inconsistent variable naming style. Please use snake_case.
</blockquote></details>

Done.
___
*[build.sh, line 251 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-MYjehp33q7d0SPVmlqa:-MYnMooH8kadWNrs8waZ:b-x5dmbl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/69a70d87df033da319f260849b1c789bc0b25404/build.sh#L251)):*

Nitpick. Since we `cd` into the `$SCRIPT_DIR` at the very start of the script I think you can use only relative paths. For some places you need absolute paths maybe, but for running this command it should be fine I think. On the other hand it makes sense to be consistent with the cargo invocation below. Does that one need absolute path?
</blockquote></details>

Some of the `$SCRIPT_DIR` were here before I started working on this, but everything seems to work as expected without the `$SCRIPT_DIR`.

faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2673#-:-MYngfKG4JuhQpL4ZS9P:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2644, opened 2021-04-08
Split `ConnectionProxy` into a UI-side and a service-side class

jvff:
This PR continues the work to separate the Android app in order to run the service in a separate process. This PR focuses on the `ConnectionProxy`. There are now two `ConnectionProxy` classes, one used by the UI and one used by the service. The service-side class handles the actual communication with the daemon, while the UI side class sends requests and receives events, as well as handling a custom UI state that anticipates the real tunnel state.

As part of the changes, the `ConnectionProxy` is now decoupled from both the service and from the `MullvadVpnService`. Two new `VpnPermission` helper classes (one UI-side and one service-side) were created to help with this decoupling.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Internal refactor, no changelog entry necessary.**
* [x] Check that commits follow the [Mullvad coding guidelines](https://github.com/mullvad/coding-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2644)

PR #2692, opened 2021-04-21
Fix VPN permission request

jvff:
A recent refactoring (#2644) broke the logic to request the Android VPN permission. Previously, the `ConnectionProxy` would start the main activity and call a method to request the VPN permission directly. After the refactor, the service side `ConnectionProxy` would instead start the main activity with a request to connect, but that request would not end up requesting for the VPN permission.

While developing a fix, I realized that the intent mechanism wasn't very robust, because it wouldn't be delivered to the activity if the UI was already shown. Therefore, this PR fixes the issue in a different way, leveraging the new IPC channel between the UI and the service. Now, the `VpnPermission` class can send request events to the UI, which will then call the API to request the permission. If the UI is not visible, the permission request event is sent with the initial events when it connects to the service.

This makes the logic more robust, and also fixes a small issue (described in the changelog) where the dialog requesting for the permission wouldn't always show.

Git checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,11 @@ Line wrap the file at 100 chars.                                              Th
 - Set correct permissions for daemon's launch file in installer.
 - Fix downgrades on macOS silently keeping previous version.
 
+#### Android
+- Fix request to connect from notification or quick-settings tile not connecting if VPN permission
+  isn't granted to the app. The app will now show the UI to ask for the permission and correctly
+  connect after it is granted.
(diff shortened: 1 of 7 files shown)


axti (approved):
> Reviewed 7 of 7 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2692)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2702, opened 2021-04-28
Properly destroy `ServiceConnection` scope after unbinding from service

jvff:
Previously, it was possible to crash the app when leaving it then coming back to it. The cause was that the `ServiceConnection` scope was being recreated without first destroying the old scope. This only happened because the `serviceConnection` property was never set to `null` when intentionally unbinding from the service, because the `onServiceDisconnected` callback is only called if the service disconnects unintentionally.

This PR fixes that by setting the `serviceConnection` to `null` right after unbinding from the service. The call currently has to be done after calling `super.onStop` because the `ServiceDependentFragment` will only properly ignore the disconnection if it has detected that the state has changed to stopped.

The PR also removes a redundant `null` notification of the service connection.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Fixes a bug not present in any publicly released version.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2702)

### DIFF
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt
@@ -55,7 +55,6 @@ open class MainActivity : FragmentActivity() {
         override fun onServiceDisconnected(className: ComponentName) {
             android.util.Log.d("mullvad", "UI lost the connection to the service")
             serviceConnection = null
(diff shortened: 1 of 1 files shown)


axti (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2702)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2396, opened 2021-01-19
Run `MullvadVpnService` on a separate process on Android

jvff:
This is the final PR to run the `MullvadVpnService` in a separate process. It marks the service to run as a separate process in the manifest, and it removes the `LocalBinder` and the `ServiceInstance` that was used for communication. Now, all communication only happen through the IPC channel.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header.
* [x] Check that commits follow the [Mullvad coding guidelines](https://github.com/mullvad/coding-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2396)

PR #2703, opened 2021-04-29
Use IPC channel in quick-settings tile service

jvff:
This PR refactors the quick-settings tile service (`MullvadTileService`) to use the IPC mechanism that has been recently implemented (#2396). Previously, the quick-settings tile used shared preferences persistent storage as a communication mechanism, which sometimes led to issues where the persisted state was not up-to-date. This PR fixes that issue by reading the tunnel state directly through IPC.

As part of the PR, the tunnel state persistence management code was removed. However, some code was added to `MullvadVpnService` to clean up the old persisted value. This could should probably be removed in the future after some time.

This PR also marks the `MullvadTileService` to run on a separate PR. This was necessary to ensure there wasn't two `ServiceConnection` scopes active at the same time (in the same process). It should be possible to not need to run it on a separate process after the dependency injection code of the service connection is improved.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2703)


axti (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2703)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2706, opened 2021-04-29
Handle keyboard in formsheet presentation (Take 2)

pronebird:

Git checklist:

1. Fix the support for automatic adjustment of UIScrollView content insets to account for keyboard overlay when in form sheet presentation style on iOS 14. This kind of presentation style is iPad specific and behaves slightly different in different versions of iOS. 
  
    Most recent versions of iOS move the controller along with the keyboard which makes it impossible to determine the part of the view that's being covered by the keyboard at the time when keyboard is about to appear. The only solution to that problem that I could find was to track the container's `frame`. 
  
    This works fine, however it seems that on iOS 14 the `presentationController?.containerView` is `nil` at the time when keyboard appears so the only way to fix this was to traverse back up and fine the container view that's being moved along the keyboard manually. This solution was tested across iOS 12 - 14.

2. on iPad, the form sheet presentation (by default) prevents the keyboard dismissal, so hitting the dismiss button in the keyboard does nothing. To address this, the view controllers that present the keyboard should return `false` from `disablesAutomaticKeyboardDismissal`. Note that in all other presentation styles the default implementation of `disablesAutomaticKeyboardDismissal` returns false as per documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2706)

### DIFF
--- a/ios/MullvadVPN/AutomaticKeyboardResponder.swift
+++ b/ios/MullvadVPN/AutomaticKeyboardResponder.swift
@@ -88,20 +88,19 @@ class AutomaticKeyboardResponder {
     private var parentViewController: UIViewController? {
         var responder: UIResponder? = targetView
         let iterator = AnyIterator { () -> UIResponder? in
-            let next = responder?.next
-            responder = next
-            return next
+            responder = responder?.next
+            return responder
         }
 
         return iterator.first { $0 is UIViewController } as? UIViewController
     }
 
+    /// Returns the presentation container view that's moved along with the keyboard on iPad
     private var presentationContainerView: UIView? {
         var currentView = parentViewController?.view
         let iterator = AnyIterator { () -> UIView? in
-            let next = currentView?.superview
-            currentView = next
-            return next
+            currentView = currentView?.superview
+            return currentView
         }
 
         // Find the container view that private `_UIFormSheetPresentationController` moves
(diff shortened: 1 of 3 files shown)


pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2706)* status: 0 of 3 files reviewed, all discussions resolved

___
*[ios/MullvadVPN/AutomaticKeyboardResponder.swift, line 109 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2706#-MZT-HchBZHyKMJb52pX:-MZT-HchBZHyKMJb52pY:btd4wgd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b9d446b6a53f6f6ddfaa5558d27c706b9857d3b0/ios/MullvadVPN/AutomaticKeyboardResponder.swift#L109)):*
> ```Swift
>         // along with the keyboard.
>         return iterator.first { (view) -> Bool in
>             return view.description.starts(with: "<UIDropShadowView")
> ```

FYI I realise that this is too much reliance on private view hierarchy but believe me I have tried everything before going this route.

axti (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2706)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2720, opened 2021-05-04
Read application name from exe-file

raksooo:
This PR contains code for retrieving an applications name from the resource tree within the exe-file. This is used when displaying the name of excluded applications that has been added with the file-picker.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)

### DIFF
--- a/gui/src/main/index.ts
+++ b/gui/src/main/index.ts
@@ -1224,7 +1224,9 @@ class ApplicationMain {
         // that we want to add to the list of additional applications.
         if (typeof application === 'string') {
           this.guiSettings.addBrowsedForSplitTunnelingApplications(application);
-          const applicationPath = windowsSplitTunneling.addApplicationPathToCache(application);
+          const applicationPath = await windowsSplitTunneling.addApplicationPathToCache(
+            application,
+          );
           await this.daemonRpc.addSplitTunnelingApplication(applicationPath);
         } else {
           await this.daemonRpc.addSplitTunnelingApplication(application.absolutepath);
--- a/gui/src/main/windows-pe-parser.ts
+++ b/gui/src/main/windows-pe-parser.ts
@@ -196,6 +196,9 @@ export const ULONGLONG = {
     },
   },
 };
+export const WCHAR = {
+  primitive: { size: 2, reader: (buffer: Buffer) => buffer.readUInt16LE(0) },
+};
 export const WORD = {
   primitive: { size: 2, reader: (buffer: Buffer) => buffer.readUInt16LE(0) },
 };
@@ -371,7 +374,104 @@ export const IMAGE_IMPORT_MODULE_DIRECTORY = {
   ],
 };
 
+export const IMAGE_RESOURCE_DIRECTORY = {
+  struct: [
+    { name: 'Characteristics', datatype: DWORD },
+    { name: 'TimeDateStamp', datatype: DWORD },
+    { name: 'MajorVersion', datatype: WORD },
+    { name: 'MinorVersion', datatype: WORD },
+    { name: 'NumberOfNameEntries', datatype: WORD },
+    { name: 'NumberOfIdEntries', datatype: WORD },
+  ],
+};
+
+export const IMAGE_RESOURCE_DIRECTORY_ID_ENTRY = {
+  struct: [
+    { name: 'Id', datatype: DWORD },
+    { name: 'OffsetToData', datatype: DWORD },
+  ],
+};
+
+export const IMAGE_RESOURCE_DIRECTORY_DATA_ENTRY = {
+  struct: [
+    { name: 'DataRVA', datatype: DWORD },
+    { name: 'Size', datatype: DWORD },
+    { name: 'CodePage', datatype: DWORD },
+    { name: '_Reserved', datatype: DWORD },
+  ],
+};
+
+export const VS_FIXEDFILEINFO = {
+  struct: [
+    { name: 'dwSignature', datatype: DWORD },
+    { name: 'dwStrucVersion', datatype: DWORD },
+    { name: 'dwFileVersionMS', datatype: DWORD },
+    { name: 'dwFileVersionLS', datatype: DWORD },
+    { name: 'dwProductVersionMS', datatype: DWORD },
+    { name: 'dwProductVersionLS', datatype: DWORD },
+    { name: 'dwFileFlagsMask', datatype: DWORD },
+    { name: 'dwFileFlags', datatype: DWORD },
+    { name: 'dwFileOS', datatype: DWORD },
+    { name: 'dwFileType', datatype: DWORD },
+    { name: 'dwFileSubtype', datatype: DWORD },
+    { name: 'dwFileDateMS', datatype: DWORD },
+    { name: 'dwFileDateLS', datatype: DWORD },
+  ],
+};
+
+// This structure can't be expressed with this format since Padding1 and Padding2 has a variable
+// size.
+export const VS_VERSIONINFO = {
+  struct: [
+    { name: 'wLength', datatype: WORD },
+    { name: 'wValueLength', datatype: WORD },
+    { name: 'wType', datatype: WORD },
+    // { name: 'szKey', datatype: ARRAY(?, WCHAR) },
+    // { name: 'Padding1', datatype: WORD },
+    // { name: 'Value', datatype: VS_FIXEDFILEINFO },
+    // { name: 'Padding2', datatype: WORD },
+    // { name: 'Children', datatype: WORD },
+  ],
+};
+
+// This structure can't be expressed with this format since Padding has a variable size.
+export const STRING_FILE_INFO = {
+  struct: [
+    { name: 'wLength', datatype: WORD },
+    { name: 'wValueLength', datatype: WORD },
+    { name: 'wType', datatype: WORD },
+    // { name: 'szKey', datatype: ARRAY(?, WCHAR) },
+    // { name: 'Padding', datatype: WORD },
+    // { name: 'Children', datatype: WORD },
+  ],
+};
+
+// This structure can't be expressed with this format since Padding has a variable size.
+export const STRING_TABLE = {
+  struct: [
+    { name: 'wLength', datatype: WORD },
+    { name: 'wValueLength', datatype: WORD },
+    { name: 'wType', datatype: WORD },
+    // { name: 'szKey', datatype: ARRAY(?, WCHAR) },
+    // { name: 'Padding', datatype: WORD },
+    // { name: 'Children', datatype: WORD },
+  ],
+};
+
+// This structure can't be expressed with this format since Padding has a variable size.
+export const STRING_TABLE_STRING = {
+  struct: [
+    { name: 'wLength', datatype: WORD },
+    { name: 'wValueLength', datatype: WORD },
+    { name: 'wType', datatype: WORD },
+    // { name: 'szKey', datatype: ARRAY(?, WCHAR) },
+    // { name: 'Padding', datatype: WORD },
+    // { name: 'Value', datatype: WORD },
+  ],
+};
+
 export const IMAGE_DIRECTORY_ENTRY_IMPORT = 1;
+export const IMAGE_DIRECTORY_ENTRY_RESOURCE = 2;
 
 export type ImageNtHeadersUnion = typeof IMAGE_NT_HEADERS | typeof IMAGE_NT_HEADERS64;
 export type ImageOptionalHeaderUnion =
--- a/gui/src/main/windows-split-tunneling.ts
+++ b/gui/src/main/windows-split-tunneling.ts
@@ -16,10 +16,19 @@ import {
   IMAGE_OPTIONAL_HEADER32,
   ImageOptionalHeaderUnion,
   PrimitiveValue,
-  rvaToOffset,
+  rvaToOffset as rvaToOffsetImpl,
   StructValue,
   Value,
   DWORD,
+  IMAGE_DIRECTORY_ENTRY_RESOURCE,
+  IMAGE_RESOURCE_DIRECTORY,
+  IMAGE_RESOURCE_DIRECTORY_DATA_ENTRY,
+  VS_VERSIONINFO,
+  STRING_FILE_INFO,
+  STRING_TABLE,
+  StructWrapper,
+  STRING_TABLE_STRING,
+  IMAGE_RESOURCE_DIRECTORY_ID_ENTRY,
 } from './windows-pe-parser';
 
 interface ShortcutDetails {
@@ -27,6 +36,8 @@ interface ShortcutDetails {
   name: string;
   args?: string;
 }
+
+type RvaToOffset = (rva: number) => Promise<number>;
 
 // Applications are found by scanning the start menu directories
 const APPLICATION_PATHS = [
@@ -65,7 +76,9 @@ export async function getApplications(options: {
   }
 
   // Add excluded apps that are missing from the shortcut cache to it
-  options.applicationPaths?.forEach(addApplicationToAdditionalShortcuts);
+  if (options.applicationPaths) {
+    await Promise.all(options.applicationPaths.map(addApplicationToAdditionalShortcuts));
+  }
 
   await updateApplicationCache();
   // If applicationPaths is supplied the returnvalue should only contain the applications
@@ -88,14 +101,14 @@ export async function getApplications(options: {
 }
 
 // Adds either a shortcut or an executable to the additionalShortcuts list
-export function addApplicationPathToCache(applicationPath: string): string {
+export async function addApplicationPathToCache(applicationPath: string): Promise<string> {
   const parsedPath = path.parse(applicationPath);
   if (parsedPath.ext === '.lnk') {
     const shortcutDetiails = shell.readShortcutLink(path.resolve(applicationPath));
     additionalShortcuts.push({ ...shortcutDetiails, name: path.parse(applicationPath).name });
     return shortcutDetiails.target;
   } else {
-    addApplicationToAdditionalShortcuts(applicationPath);
+    await addApplicationToAdditionalShortcuts(applicationPath);
     return applicationPath;
   }
 }
@@ -135,7 +148,7 @@ async function updateApplicationCache(): Promise<void> {
 }
 
 // Add excluded apps that are missing from the shortcut cache to it
-function addApplicationToAdditionalShortcuts(applicationPath: string): void {
+async function addApplicationToAdditionalShortcuts(applicationPath: string): Promise<void> {
   if (
     shortcutCache[applicationPath.toLowerCase()] === undefined &&
     !additionalShortcuts.some(
@@ -144,7 +157,7 @@ function addApplicationToAdditionalShortcuts(applicationPath: string): void {
   ) {
     additionalShortcuts.push({
       target: applicationPath,
-      name: path.parse(applicationPath).name,
+      name: (await getProgramName(applicationPath)) ?? path.parse(applicationPath).name,
     });
   }
 }
@@ -173,18 +186,26 @@ function resolveLinks(linkPaths: string[]): ShortcutDetails[] {
           ...shell.readShortcutLink(path.resolve(link)),
           name: path.parse(link).name,
         };
-      } catch (_e) {
+      } catch {
         return null;
       }
     })
     .filter(
       (shortcut): shortcut is ShortcutDetails =>
         shortcut !== null &&
-        shortcut.name !== 'Mullvad VPN' &&
+        !shortcut.target.endsWith('Mullvad VPN.exe') &&
         shortcut.target.endsWith('.exe') &&
-        !shortcut.target.toLowerCase().includes('uninstall') &&
-        !shortcut.name.toLowerCase().includes('uninstall'),
+        !shortcut.target.toLowerCase().includes('install') && // Covers "uninstall" as well.
+        !shortcut.name.toLowerCase().includes('install'),
     );
+}
+
+async function getProgramName(exePath: string): Promise<string | undefined> {
+  try {
+    return await getProductName(exePath);
+  } catch {
+    return undefined;
+  }
 }
 
 // Removes all duplicate shortcuts.
@@ -229,7 +250,6 @@ async function importsDll(path: string, dllName: string): Promise<boolean> {
   try {
     fileHandle = await fs.promises.open(path, fs.constants.O_RDONLY);
   } catch (e) {
-    log.error('Failed to create file handle.', e);
     return false;
   }
 
@@ -240,42 +260,14 @@ async function importsDll(path: string, dllName: string): Promise<boolean> {
 
 async function getExeImports(fileHandle: fs.promises.FileHandle, path: string): Promise<string[]> {
   try {
-    const ntHeader = await getNtHeader(fileHandle);
-    const fileHeader = ntHeader.get<StructValue<typeof IMAGE_FILE_HEADER>>('FileHeader');
-    const optionalHeader = ntHeader.get<StructValue<ImageOptionalHeaderUnion>>('OptionalHeader');
-
-    const importTableRva = optionalHeader
-      .get<ArrayValue<typeof IMAGE_DATA_DIRECTORY>>('DataDirectory')
-      .nth(IMAGE_DIRECTORY_ENTRY_IMPORT)
-      .get('VirtualAddress')
-      .value();
-
-    if (importTableRva === 0x0) {
+    const tableOffsetResult = await getTableOffset(fileHandle, IMAGE_DIRECTORY_ENTRY_IMPORT);
+    if (tableOffsetResult) {
+      const { offset: importTableOffset, rvaToOffset } = tableOffsetResult;
+      const moduleNames = await getImportModuleNames(fileHandle, importTableOffset, rvaToOffset);
+      return moduleNames;
+    } else {
       return [];
     }
-
-    const numberOfSections = fileHeader.get<PrimitiveValue>('NumberOfSections').value<number>();
-    const ntHeaderEndOffset =
-      ntHeader.offset +
-      ntHeader.get<PrimitiveValue<typeof DWORD>>('Signature').size +
-      fileHeader.size +
-      fileHeader.get<PrimitiveValue>('SizeOfOptionalHeader').value();
-
-    const importTableOffset = await rvaToOffset(
-      fileHandle,
-      importTableRva,
-      numberOfSections,
-      ntHeaderEndOffset,
-    );
-
-    const moduleNames = await getImportModuleNames(
-      fileHandle,
-      importTableOffset,
-      ntHeader.endOffset,
-      numberOfSections,
-    );
-
-    return moduleNames;
   } catch (e) {
     log.error(`Failed to read .exe import table for ${path}.`, e);
     return [];
@@ -285,14 +277,28 @@ async function getExeImports(fileHandle: fs.promises.FileHandle, path: string): 
 async function readString(
   fileHandle: fs.promises.FileHandle,
   offset: number,
-  buffer = Buffer.alloc(1000),
-  index = 0,
-): Promise<string> {
-  await fileHandle.read(buffer, index, 1, offset + index);
-  if (buffer[index] === 0x0) {
-    return buffer.slice(0, index).toString('ascii');
+  encoding: 'ascii' | 'ucs2',
+): Promise<{ value: string; endOffset: number }> {
+  const characterSize = getCharacterSize(encoding);
+  const buffer = Buffer.alloc(characterSize);
+  await fileHandle.read(buffer, 0, characterSize, offset);
+
+  const nextOffset = offset + characterSize;
+  if (buffer.every((value) => value === 0)) {
+    return { value: '', endOffset: nextOffset };
   } else {
-    return readString(fileHandle, offset, buffer, index + 1);
+    const { value: nextValue, endOffset } = await readString(fileHandle, nextOffset, encoding);
+    const value = buffer.toString(encoding) + nextValue;
+    return { value, endOffset };
+  }
+}
+
+function getCharacterSize(encoding: 'ascii' | 'ucs2'): number {
+  switch (encoding) {
+    case 'ascii':
+      return 1;
+    case 'ucs2':
+      return 2;
   }
 }
 
@@ -302,12 +308,12 @@ async function getNtHeader(
 ): Promise<StructValue<ImageNtHeadersUnion>> {
   // Check whether or not the file follows the PE format.
   const dosHeader = await Value.fromFile(fileHandle, 0, DOS_HEADER);
-  const eMagic = dosHeader.get<PrimitiveValue>('e_magic').value<number>();
+  const eMagic = dosHeader.get<PrimitiveValue>('e_magic').value();
   if (eMagic !== 0x5a4d) {
     throw new Error('Not a PE file');
   }
 
-  const ntHeaderOffset = dosHeader.get<PrimitiveValue>('e_lfanew').value<number>();
+  const ntHeaderOffset = dosHeader.get<PrimitiveValue>('e_lfanew').value();
 
   // Check if this is a 32- or 64-bit exe-file and return the correct datatype.
   const ntHeader32 = await Value.fromFile(fileHandle, ntHeaderOffset, IMAGE_NT_HEADERS);
@@ -319,7 +325,7 @@ async function getNtHeader(
   const magic = ntHeader32
     .get<StructValue<typeof IMAGE_OPTIONAL_HEADER32>>('OptionalHeader')
     .get<PrimitiveValue>('Magic')
-    .value<number>();
+    .value();
 
   // magic is 0x20b for 64-bit executables.
   return magic === 0x20b
@@ -331,8 +337,7 @@ async function getNtHeader(
 async function getImportModuleNames(
   fileHandle: fs.promises.FileHandle,
   importTableOffset: number,
-  firstSectionHeaderOffset: number,
-  numberOfSections: number,
+  rvaToOffset: RvaToOffset,
 ): Promise<string[]> {
   const moduleNames: string[] = [];
   const entrySize = Value.sizeOf(IMAGE_IMPORT_MODULE_DIRECTORY);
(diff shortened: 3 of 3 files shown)


raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 0 of 3 files reviewed, all discussions resolved

___
*[gui/src/main/windows-split-tunneling.ts, line 531 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZrkb-yC3MhgaI_nUpN:-MZrkb-yC3MhgaI_nUpO:b-vm06t5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L531)):*
> ```XML
> 
>     const paddingOffset = szKeyOffset + szKey.length * 2;
>     const childrenOffset = (await findOffsetAfterPadding(fileHandle, paddingOffset, 2)) - 2;
> ```

I'm not sure why offsetting this by `2` is needed. But it works well. The documentation isn't very detailed.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 1 of 3 files reviewed, all discussions resolved

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 2 of 3 files reviewed, 7 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/windows-pe-parser.ts, line 391 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZswQql65z70Dtb4zl7:-MZswQqmDQYP-5GsYQYv:bd65efl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-pe-parser.ts#L391)):*
> ```XML
>   struct: [
>     { name: 'NameOrId', datatype: DWORD },
>     { name: 'OffsetToData', datatype: DWORD },
> ```

Is there no union-construct? The current name doesn't really convey what this field does.
___
*[gui/src/main/windows-pe-parser.ts, line 397 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZszBi91b2GfUTcYGN3:-MZszBi91b2GfUTcYGN4:b-njnh55) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-pe-parser.ts#L397)):*
> ```XML
> export const IMAGE_RESOURCE_DIRECTORY_STRING_U = {
>   struct: [
>     { name: 'Length', datatype: WORD },
> ```

This is the character length of the string, which for this particular string type is (string byte count / 2).

The total byte length for each string entry is (sizeof(WORD) + (2 * `Length`)), and the string is not null-terminated.
___
*[gui/src/main/windows-pe-parser.ts, line 398 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZt-IETA0f_TIJ8XiFM:-MZt-IEU9c3CJcWkIfyx:b-9jmxj0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-pe-parser.ts#L398)):*
> ```XML
>   struct: [
>     { name: 'Length', datatype: WORD },
>     { name: 'String', datatype: UTF8_STRING(0) },
> ```

This is not an UTF8 string. It's a non-null-terminated UCS2 string.
___
*[gui/src/main/windows-split-tunneling.ts, line 391 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZtEKYuDz7RvFoFyM09:-MZtEKYuDz7RvFoFyM0A:b9ibw9z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L391)):*
> ```XML
>   const optionalHeader = ntHeader.get<StructValue<ImageOptionalHeaderUnion>>('OptionalHeader');
> 
>   const importTableRva = optionalHeader
> ```

Variable name is wrong and misleading about 94% of the time.
___
*[gui/src/main/windows-split-tunneling.ts, line 434 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZtHqqd6pVClVar-wUQ:-MZtHqqd6pVClVar-wUR:bf7jro0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L434)):*
> ```XML
>   let offset = entryOffset + entry.size;
>   for (let i = 0; i < numberOfNameEntries + numberOfIdEntries; i++) {
>     const entry = await Value.fromFile(fileHandle, offset, IMAGE_RESOURCE_DIRECTORY_ENTRY);
> ```

This is shadowing the existing `entry` variable and makes the code harder to digest.
___
*[gui/src/main/windows-split-tunneling.ts, line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZtOR7i275SXEY1YDqw:-MZtJ_9b7ZPruchGXaPY:blgoibw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L439)):*
> ```XML
>     let nameOrId = Number(nameOffsetOrId).toString();
>     // If the first bit is 1 then the value is an offset to a string and otherwise it's an ID
>     if (nameOffsetOrId & 0x80000000) {
> ```

The first `NumberOfNameEntries` entries contain a string offset in this field. Consider that entries which use a numeric identifier could use a value of >= 0x80000000.
___
*[gui/src/main/windows-split-tunneling.ts, line 445 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZtOx-O1mf7CliSuW_4:-MZtOx-O1mf7CliSuW_5:b5tpteq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L445)):*
> ```XML
> 
>     offset += entry.size;
>     if (!ids.map((id) => `${id}`).includes(nameOrId)) {
> ```

I'm not sure exactly what this does. Are tree levels respected?

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 2 of 3 files reviewed, 8 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/windows-split-tunneling.ts, line 193 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvW8eu0NRXnbrQImtn:-MZvW8eu0NRXnbrQImto:b-fd2mng) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L193)):*
> ```XML
> async function getProgramName(exePath: string): Promise<string | undefined> {
>   try {
>     return getProductName(exePath);
> ```

missing `await`?

mvd-ows (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 2 of 3 files reviewed, 15 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/windows-split-tunneling.ts, line 358 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvZaUiBlNybFX4KDlH:-MZvZaUiBlNybFX4KDlI:bo7eybv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L358)):*
> ```XML
>         resourceTableOffset,
>         rvaToOffset,
>         [[16], [1], [0, 1033]],
> ```

You're looking for a specific resource. Why are multiple leafs considered?
___
*[gui/src/main/windows-split-tunneling.ts, line 378 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvYvPR0-QNAiWECxli:-MZvYvPR0-QNAiWECxlj:b-lilboj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L378)):*
> ```XML
>     }
>   } catch (e) {
>     await fileHandle.close();
> ```

If the `else` is taken above, the file handle will have already been closed. Maybe calling `close()` multiple times is fine?
___
*[gui/src/main/windows-split-tunneling.ts, line 456 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvNlDB3pxeixV_rcqz:-MZvNlDB3pxeixV_rcr-:b-4sn3pc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L456)):*
> ```XML
>     }
> 
>     if (leaf && path.length === 0) {
> ```

Can you clarify what the conditions are? `if leaf and something then`, what is `something`?
___
*[gui/src/main/windows-split-tunneling.ts, line 464 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvNCdv0ayDwrSIfNzQ:-MZvNCdv0ayDwrSIfNzR:b3p175e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L464)):*
> ```XML
> 
>       leaves.push(leafDataOffset);
>     } else {
> ```

Currently, the case `leaf && path.length !== 0` is handled here, which is wrong. You can't recurse into a leaf.
___
*[gui/src/main/windows-split-tunneling.ts, line 465 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvOhNI4iR33TL0WBrR:-MZvOhNI4iR33TL0WBrS:b-o6byg9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L465)):*
> ```XML
>       leaves.push(leafDataOffset);
>     } else {
>       return getResourceTreeLeafOffsets(
> ```

This looks suspicious. Shouldn't it be combining the return value with the `leaves` instance in the current stack frame?
___
*[gui/src/main/windows-split-tunneling.ts, line 555 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvcNx78NyJjOx34FFL:-MZvcNx78NyJjOx34FFM:b-2sl825) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L555)):*
> ```XML
>   const nameValue = await Value.fromFile(fileHandle, offset, IMAGE_RESOURCE_DIRECTORY_STRING_U);
>   const length = nameValue.get<PrimitiveValue>('Length');
>   const buffer = Buffer.alloc(length.value());
> ```

`Length` is a character count, not a byte count.
___
*[gui/src/main/windows-split-tunneling.ts, line 557 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvcAOg1VCCikA2qtCy:-MZvcAOg1VCCikA2qtCz:b-3ceum5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L557)):*
> ```XML
>   const buffer = Buffer.alloc(length.value());
>   await fileHandle.read(buffer, 0, buffer.length, offset + length.size);
>   return buffer.toString('utf-8');
> ```

Unsure what this does, but the source data is not UTF8-encoded. See previous comment.

mvd-ows (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 2 of 3 files reviewed, 17 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/windows-split-tunneling.ts, line 612 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvi96mCL1qAIlZVZrg:-MZvi96n8uXHLzqqA-w7:byc7614) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L612)):*
> ```XML
> ): Promise<number> {
>   const buffer = Buffer.alloc(bytes);
>   await fileHandle.read(buffer, 0, buffer.length, offset);
> ```

You're not supposed to read the padding, the value of padding bytes is undefined.
___
*[gui/src/main/windows-split-tunneling.ts, line 615 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvi_0b5QO6jBJIIvUg:-MZvi_0b5QO6jBJIIvUh:be9cpnr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L615)):*
> ```XML
>   if (buffer.readUInt8(0) === 0) {
>     if (offset % 4 === 0) {
>       return findOffsetAfterPadding(fileHandle, offset + 4);
> ```

You don't need recursion for this.

mvd-ows (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 2 of 3 files reviewed, 19 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/windows-pe-parser.ts, line 476 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvsFsQ3qfb6pU2pZFz:-MZvsFsR1kwLMF-sa068:b-k9fkx8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-pe-parser.ts#L476)):*
> ```XML
>     // { name: 'szKey', datatype: ARRAY(?, WCHAR) },
>     // { name: 'Padding', datatype: WORD },
>     // { name: 'Children', datatype: WORD },
> ```

This struct doesn't have any children. It represents a key-value pair and the final field is `Value`.
___
*[gui/src/main/windows-split-tunneling.ts, line 531 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZrkb-yC3MhgaI_nUpN:-MZvuG3C7kDC-KtKnJtx:b-gsjfkl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L531)):*

I'm not sure why offsetting this by `2` is needed. But it works well. The documentation isn't very detailed.
</blockquote></details>

`szKeyOffset` is correctly computed. This structure is a little retarded. There's no field that provides the length of the key.

You need to read a zero-terminated UCS2 string starting at `szKeyOffset`. This gives you the key-string and a new offset which "points" beyond the null-terminator. The new offset has to be aligned to 4 bytes, and this yields the offset of the value-string.

`valueLength`is the character-length of the value-string, including a null-terminator.
___
*[gui/src/main/windows-split-tunneling.ts, line 539 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvx2v51oPEbT9vcisl:-MZvx2v51oPEbT9vcism:b-glsrod) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L539)):*
> ```XML
>     strings.set(szKey, value);
> 
>     currentStringOffset = await findOffsetAfterPadding(
> ```

This should be `currentStringOffset += alignDword(stringValue.get<PrimitiveValue>('wLength').value());`

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: all files reviewed, 19 unresolved discussions (waiting on @raksooo)

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 1 of 3 files reviewed, 11 unresolved discussions (waiting on @mvd-ows and @pronebird)

___
*[gui/src/main/windows-pe-parser.ts, line 391 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZswQql65z70Dtb4zl7:-MZvam3PDzymiJv0FInq:b-ukz5no) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-pe-parser.ts#L391)):*

Is there no union-construct? The current name doesn't really convey what this field does.
</blockquote></details>

I haven't implemented support for a union-construct. How would this be expressed with a union? I don't think I've seen any documentation expressing this as a union.
___
*[gui/src/main/windows-pe-parser.ts, line 397 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZszBi91b2GfUTcYGN3:-M_-uNvz5myu_nY5aOVp:b-iv17n9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-pe-parser.ts#L397)):*

This is the character length of the string, which for this particular string type is (string byte count / 2).

The total byte length for each string entry is (sizeof(WORD) + (2 * `Length`)), and the string is not null-terminated.
</blockquote></details>

Okay :+1: Do you suggest any changes to this definition?
___
*[gui/src/main/windows-pe-parser.ts, line 398 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZt-IETA0f_TIJ8XiFM:-M_-uN3w7--cKygqMXrU:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-pe-parser.ts#L398)):*

This is not an UTF8 string. It's a non-null-terminated UCS2 string.
</blockquote></details>

Done.
___
*[gui/src/main/windows-split-tunneling.ts, line 193 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvW8eu0NRXnbrQImtn:-MZvb_H94Hn58nh5jOJ_:b-tu1aon) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L193)):*

missing `await`?
</blockquote></details>

It's not needed to await promises when returning, the result will be the same.
___
*[gui/src/main/windows-split-tunneling.ts, line 358 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvZaUiBlNybFX4KDlH:-MZvopDi2Bk7h3yco7xM:b-evxq5m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L358)):*

You're looking for a specific resource. Why are multiple leafs considered?
</blockquote></details>

I've noticed that VS_VERSIONINFO can either be placed at the leaf with id `0` or `1033`. If you know anywhere where this is documented I would like to read more about it :) 
___
*[gui/src/main/windows-split-tunneling.ts, line 378 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvYvPR0-QNAiWECxli:-MZvp9ru75sZPMyQoEnV:b-d9o6d1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L378)):*

If the `else` is taken above, the file handle will have already been closed. Maybe calling `close()` multiple times is fine?
</blockquote></details>

It works fine to call `.close()` twice, but the behavior isn't documented so I've changed it.
___
*[gui/src/main/windows-split-tunneling.ts, line 391 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZtEKYuDz7RvFoFyM09:-MZvbiqnAwUOtRcGzZsV:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L391)):*

Variable name is wrong and misleading about 94% of the time.
</blockquote></details>

Done.
___
*[gui/src/main/windows-split-tunneling.ts, line 434 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZtHqqd6pVClVar-wUQ:-MZvbkN3ADkLIx7drW3W:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L434)):*

This is shadowing the existing `entry` variable and makes the code harder to digest.
</blockquote></details>

Done.
___
*[gui/src/main/windows-split-tunneling.ts, line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZtOR7i275SXEY1YDqw:-MZvbp0t57O5HOK4vo3B:b13s1ei) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L439)):*

The first `NumberOfNameEntries` entries contain a string offset in this field. Consider that entries which use a numeric identifier could use a value of >= 0x80000000.
</blockquote></details>

`nameOffsetOrId & 0x80000000` will be truthy (!== 0) for values above `0x80000000` as well (as long as it's not greater than `0xffffffff`, but that shouldn't be possible).
___
*[gui/src/main/windows-split-tunneling.ts, line 445 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZtOx-O1mf7CliSuW_4:-MZvcWKv5d2JX15CfSaP:b6y9974) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L445)):*

I'm not sure exactly what this does. Are tree levels respected?
</blockquote></details>

Yes, `ids` only contains the valid IDs for the current tree level and if this condition is fulfilled then the remaining levels' IDs will be passed on in the recursive call further down.

The `map` call is only there to convert the ids to strings.
___
*[gui/src/main/windows-split-tunneling.ts, line 456 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvNlDB3pxeixV_rcqz:-MZvo2Gt-ML35a8ZO1rm:b-bg21ht) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L456)):*

Can you clarify what the conditions are? `if leaf and something then`, what is `something`?
</blockquote></details>

`path` contains the path through the tree and if it's 0, then the current entry is expected to be a leaf. So it's basically `if the current entry is a leaf and we expect it to be a leaf`.
___
*[gui/src/main/windows-split-tunneling.ts, line 464 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvNCdv0ayDwrSIfNzQ:-MZwJlYo33HK2Gl3rmt6:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L464)):*

Currently, the case `leaf && path.length !== 0` is handled here, which is wrong. You can't recurse into a leaf.
</blockquote></details>

Done.
___
*[gui/src/main/windows-split-tunneling.ts, line 465 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvOhNI4iR33TL0WBrR:-MZwK97c57edYYuTl8gJ:b-d2xsuy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L465)):*

This looks suspicious. Shouldn't it be combining the return value with the `leaves` instance in the current stack frame?
</blockquote></details>

Ah, forgot this while refactoring. I've been lucky that this hasn't caused any issues. Done.
___
*[gui/src/main/windows-split-tunneling.ts, line 539 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvx2v51oPEbT9vcisl:-M_0TfLXDLn_ykXZWiOt:b4855pb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L539)):*

This should be `currentStringOffset += alignDword(stringValue.get<PrimitiveValue>('wLength').value());`
</blockquote></details>

Thanks! Done.
___
*[gui/src/main/windows-split-tunneling.ts, line 555 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvcNx78NyJjOx34FFL:-M_0TmK36gtWdtt1ohoL:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L555)):*

`Length` is a character count, not a byte count.
</blockquote></details>

Done.
___
*[gui/src/main/windows-split-tunneling.ts, line 612 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvi96mCL1qAIlZVZrg:-MZvmnUQ-6LkGCV2mOGw:b-llsdn3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L612)):*

You're not supposed to read the padding, the value of padding bytes is undefined.
</blockquote></details>

Done. I read something about paddings larger than 32-byte but when reading more about this it looks like that's not the case.
___
*[gui/src/main/windows-split-tunneling.ts, line 615 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvi_0b5QO6jBJIIvUg:-MZvnbYl6jfef6K9DIN-:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L615)):*

You don't need recursion for this.
</blockquote></details>

Done.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 1 of 3 files reviewed, 11 unresolved discussions (waiting on @mvd-ows and @raksooo)

___
*[gui/src/main/windows-split-tunneling.ts, line 193 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvW8eu0NRXnbrQImtn:-M_0ZT-P2sphz7nDNHt4:b-nfyn3m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L193)):*

It's not needed to await promises when returning, the result will be the same.
</blockquote></details>

The try+catch block won't catch any errors unless you explicitly `await`, so if you want to return the Promise and filter out the errors the way you do, you have to:

```
return getProductName(exePath).catch(error => undefined)
```

But then in this case you don't need to mark this function as `async` since this is only a requirement for `await` calls.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 1 of 3 files reviewed, 11 unresolved discussions (waiting on @mvd-ows and @pronebird)

___
*[gui/src/main/windows-split-tunneling.ts, line 193 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvW8eu0NRXnbrQImtn:-M_0dyB09X6antRViGJc:bjrpk0o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L193)):*

The try+catch block won't catch any errors unless you explicitly `await`, so if you want to return the Promise and filter out the errors the way you do, you have to:

```
return getProductName(exePath).catch(error => undefined)
```

But then in this case you don't need to mark this function as `async` since this is only a requirement for `await` calls.
</blockquote></details>

Oh, didn't think about that. Thanks! Done.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 1 of 3 files reviewed, 11 unresolved discussions (waiting on @mvd-ows, @pronebird, and @raksooo)

___
*[gui/src/main/windows-pe-parser.ts, line 391 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZswQql65z70Dtb4zl7:-M_0yagu-Qkq9i6wrTkC:b-7a8iiz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-pe-parser.ts#L391)):*

I haven't implemented support for a union-construct. How would this be expressed with a union? I don't think I've seen any documentation expressing this as a union.
</blockquote></details>

A union in C/C++ is used to declare that a location in memory have multiple identities.

Observe the table at the end of the linked section, specifically the offsets of fields being listed: https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#resource-directory-entries

The structure has two unions, each containing two fields.

mvd-ows (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 2 of 3 files reviewed, 11 unresolved discussions (waiting on @mvd-ows, @pronebird, and @raksooo)

___
*[gui/src/main/windows-pe-parser.ts, line 397 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZszBi91b2GfUTcYGN3:-M_2B_Zf77wwM9kzC3a1:bgcbpxi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-pe-parser.ts#L397)):*

Okay :+1: Do you suggest any changes to this definition?
</blockquote></details>

The definition is fine. But the details mentioned above affect how parsing is implemented.
___
*[gui/src/main/windows-split-tunneling.ts, line 358 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvZaUiBlNybFX4KDlH:-M_2G4F961k2051bx7Xr:b-veczrt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L358)):*

I've noticed that VS_VERSIONINFO can either be placed at the leaf with id `0` or `1033`. If you know anywhere where this is documented I would like to read more about it :) 
</blockquote></details>

`1033` is the language code for US-english, IIRC. I'll look into it.
___
*[gui/src/main/windows-split-tunneling.ts, line 273 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-M_2DyRgDHDKx0XA90Eq:-M_2DyRgDHDKx0XA90Er:b-35gy8b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b59eb5b0edc98be4862619d522d3c16434aed6c1/gui/src/main/windows-split-tunneling.ts#L273)):*
> ```XML
> ): Promise<string> {
>   await fileHandle.read(buffer, index, bytes, offset + index);
>   if (buffer[index] === 0x0) {
> ```

All of the strings you're parsing out of the resource structures use UCS2-encoding. I.e. they will be stored like this in the binary: `s, 0x00, t, 0x00, r, 0x00, i, 0x00, n, 0x00, g, 0x00`.

You're supposed to read a word-sized item at a time and byte-swap it. So the first letter is 0x0073, etc.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 2 of 3 files reviewed, 9 unresolved discussions (waiting on @mvd-ows, @pronebird, and @raksooo)

___
*[gui/src/main/windows-split-tunneling.ts, line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZtOR7i275SXEY1YDqw:-M_2Ks1e3ssP935fQmgL:b-mjhphf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L439)):*

`nameOffsetOrId & 0x80000000` will be truthy (!== 0) for values above `0x80000000` as well (as long as it's not greater than `0xffffffff`, but that shouldn't be possible).
</blockquote></details>

Consider the following example:

```
IMAGE_RESOURCE_DIRECTORY
    NumberOfNameEntries = 3
    NumberOfIdEntries = 2
IMAGE_RESOURCE_DIRECTORY_ENTRY
    Using string identifier
IMAGE_RESOURCE_DIRECTORY_ENTRY
    Using string identifier
IMAGE_RESOURCE_DIRECTORY_ENTRY
    Using string identifier
IMAGE_RESOURCE_DIRECTORY_ENTRY
    Using numerical identifier
IMAGE_RESOURCE_DIRECTORY_ENTRY
    Using numerical identifier
```

The `IMAGE_RESOURCE_DIRECTORY` structure is followed by an array containing `NumberOfNameEntries` + `NumberOfIdEntries` entries of type `IMAGE_RESOURCE_DIRECTORY_ENTRY`.

The first `NumberOfNameEntries` entries use a string as identifier. The next `NumberOfIdEntries` use a number as identifier.

The current logic will attempt to parse a string identifier out of the last entries as well. This will turn sour if a specific entry uses an identifier with a value of >= 0x80000000.
___
*[gui/src/main/windows-split-tunneling.ts, line 445 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZtOx-O1mf7CliSuW_4:-M_2O1mx6YGxPA4Qw6hz:b9ldnc3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L445)):*

Yes, `ids` only contains the valid IDs for the current tree level and if this condition is fulfilled then the remaining levels' IDs will be passed on in the recursive call further down.

The `map` call is only there to convert the ids to strings.
</blockquote></details>

Cool, thanks.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 2 of 3 files reviewed, 6 unresolved discussions (waiting on @mvd-ows, @pronebird, and @raksooo)

___
*[gui/src/main/windows-split-tunneling.ts, line 456 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvNlDB3pxeixV_rcqz:-M_2OneQ78SGY9vBUSm2:b-ebb6zh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L456)):*

`path` contains the path through the tree and if it's 0, then the current entry is expected to be a leaf. So it's basically `if the current entry is a leaf and we expect it to be a leaf`.
</blockquote></details>

Cool, thanks.
___
*[gui/src/main/windows-split-tunneling.ts, line 531 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZrkb-yC3MhgaI_nUpN:-M_2UCdI7wKYe7tqQAEA:b-8pe1su) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L531)):*

`szKeyOffset` is correctly computed. This structure is a little retarded. There's no field that provides the length of the key.

You need to read a zero-terminated UCS2 string starting at `szKeyOffset`. This gives you the key-string and a new offset which "points" beyond the null-terminator. The new offset has to be aligned to 4 bytes, and this yields the offset of the value-string.

`valueLength`is the character-length of the value-string, including a null-terminator.


</blockquote></details>

Try using the attached logic.

First, you need:

`const paddingOffset = szKeyOffset + (szKey.length * 2) + 2;`

This is to account for the zero terminator.

Then you need (there is no `child` field in the current structure):

`const valueOffset = alignDword(paddingOffset);`

And finally you need:

`await fileHandle.read(buffer, 0, buffer.length, valueOffset);`

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: all files reviewed, 6 unresolved discussions (waiting on @pronebird and @raksooo)

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: all files reviewed, 5 unresolved discussions (waiting on @raksooo)

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on @mvd-ows and @raksooo)

___
*[gui/src/main/windows-split-tunneling.ts, line 531 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZrkb-yC3MhgaI_nUpN:-M_6ZoAj2wvAZBHygX9v:bd7gzsq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L531)):*

Try using the attached logic.

First, you need:

`const paddingOffset = szKeyOffset + (szKey.length * 2) + 2;`

This is to account for the zero terminator.

Then you need (there is no `child` field in the current structure):

`const valueOffset = alignDword(paddingOffset);`

And finally you need:

`await fileHandle.read(buffer, 0, buffer.length, valueOffset);`
</blockquote></details>

Workds great. I've changed `readString` to also return the offset after the string which simplifies this a bit.
___
*[gui/src/main/windows-split-tunneling.ts, line 273 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-M_2DyRgDHDKx0XA90Eq:-M_6ZqWW9jsPCQgKe8-z:b25j0fh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b59eb5b0edc98be4862619d522d3c16434aed6c1/gui/src/main/windows-split-tunneling.ts#L273)):*

All of the strings you're parsing out of the resource structures use UCS2-encoding. I.e. they will be stored like this in the binary: `s, 0x00, t, 0x00, r, 0x00, i, 0x00, n, 0x00, g, 0x00`.

You're supposed to read a word-sized item at a time and byte-swap it. So the first letter is 0x0073, etc.


</blockquote></details>

I've refactored `readString` a bit. For UCS2 it now reads 2 bytes at a time and then convert each couple to a string with `buffer.toString('ucs2')` until both bytes are `0`.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 1 of 3 files reviewed, 4 unresolved discussions (waiting on @mvd-ows and @raksooo)

___
*[gui/src/main/windows-split-tunneling.ts, line 358 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZvZaUiBlNybFX4KDlH:-M_KRuyZ3cbGwKOFcWih:bmt8juz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L358)):*

I can't find an information about id `0` but I can assure you that some programs on my Windows install have the `VS_VERSIONINFO` located there.
</blockquote></details>

It's difficult to find accurate and succinct information on this.

The value `16` comes from `winuser.h`:

`#define RT_VERSION      MAKEINTRESOURCE(16)`

The next level, using `1`, seems to be defined in `verrsrc.h`:

`#define VS_VERSION_INFO         1`

The leaf nodes are indeed identified by language IDs. The id`0` means `neutral`, and is prioritized above other leaf nodes.

Something in the back of my mind says the literal `RT_VERSION` name string is a valid identifier for the first level. I tried creating a file using this identifier, but either my patching was incomplete or my memory is wrong. This is something you may want to look into.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @mvd-ows and @raksooo)

___
*[gui/src/main/windows-split-tunneling.ts, line 273 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-M_2DyRgDHDKx0XA90Eq:-M_KfoDR7-NRGRYnsCFa:brup3q7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b59eb5b0edc98be4862619d522d3c16434aed6c1/gui/src/main/windows-split-tunneling.ts#L273)):*

I've refactored `readString` a bit. For UCS2 it now reads 2 bytes at a time and then convert each couple to a string with `buffer.toString('ucs2')` until both bytes are `0`. 
</blockquote></details>

That great :+1:

mvd-ows (changes requested):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @mvd-ows and @raksooo)

___
*[gui/src/main/windows-split-tunneling.ts, line 541 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-M_Kj_qp0-fOFTlgmSyS:-M_Kj_qp0-fOFTlgmSyT:b-yd46lf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/91af4c8015aff893c3a2def15f23afbfa03c0a63/gui/src/main/windows-split-tunneling.ts#L541)):*
> ```XML
> 
>     const valueOffset = alignDword(endOffset);
>     const { value } = await readString(fileHandle, valueOffset, 'ucs2');
> ```

Why are you using `readString()` to get the value-string? This is extremely inefficient.

The length of the string is already known so you can straight up read it.

mvd-ows (commented):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: all files reviewed, 4 unresolved discussions (waiting on @raksooo)

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: all files reviewed, 1 unresolved discussion (waiting on @mvd-ows)

___
*[gui/src/main/windows-split-tunneling.ts, line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-MZtOR7i275SXEY1YDqw:-M_ewZBN8TCXV1LAfQFJ:b-glmwcf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/08173b7fa1fc18de49c09ebb57d4145f8243c7f4/gui/src/main/windows-split-tunneling.ts#L439)):*

The part you're quoting is related to the structure as seen in isolation. Here's another excerpt from the article that speaks about the implementation:

```
...an array of IMAGE_RESOURCE_DIRECTORY_ENTRY structures that immediately follow the IMAGE_RESOURCE_DIRECTORY structure. The number of elements in the array is the sum of the NumberOfNamedEntries and NumberOfIdEntries fields. The directory entry elements that have name identifiers (rather than integer IDs) come first in the array.
```

Also, that article is meant as an aid in understanding the specification. The article has errors and omissions. The first version appeared in 1994, when the spec wasn't available.

If in doubt, refer documents in this order:

* SDK header files
* `pecoff.docx`, official specification
* Other sources
</blockquote></details>

Since I now skip all named entries this should work as expected.

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[gui/src/main/windows-split-tunneling.ts, line 442 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720#-M_gYzcYD_hT3oRoQupM:-M_gY9sQ62zVuFGGX3BB:bt1m3nh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b94244444f63ed8ca8ac0171aa52ccc8aca0b8e5/gui/src/main/windows-split-tunneling.ts#L442)):*
> ```XML
>   const leaves: number[] = [];
> 
>   let offset = tableOffset + table.size;
> ```

This can be

`let offset = tableOffset + sizeof(IMAGE_RESOURCE_DIRECTORY) + (numberOfNameEntries * sizeof(IMAGE_RESOURCE_DIRECTORY_ID_ENTRY));`

In order to remove n unnecessary disk reads and the same number of dummy iterations in the loop below.

Also adjust the looping condition.

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2720)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

PR #2738, opened 2021-05-11
Improve tunnel settings verification

pronebird:

Git checklist:

We've got a couple of problem reports with logs pointing at missing Keychain entries. It turned out that when settings up a new device from backup, the Keychain entries may not be restored. At the same time user preferences are usually restored from backup. 

Since we store the account token and expiry in user preferences, and VPN tunnel settings in Keychain, this would cause the inconsistent state inside of the app. Even better, the custom VPN configuration (tunnel) created by our app would be restored too but then contain the broken persistent keychain reference to tunnel settings and would ultimately fail to start the tunnel.

This PR improves the system VPN configuration (tunnel) and tunnel settings verification. In case when we have keychain entry, we can simply remove the VPN configuration (tunnel) from the system and then re-create it once user hits the button to connect the tunnel (this usually pops the system dialog to grant permissions to add VPN configuration). 

In all other cases we have to log out user since if we don't have Keychain entry with tunnel settings, it also means that we don't have a private key for the tunnel and we create it and push the public key to our servers during log-in to receive the WireGuard associated addresses. Doing it this way also avoids implicit networking.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2738)

### DIFF
--- a/ios/CHANGELOG.md
+++ b/ios/CHANGELOG.md
@@ -34,6 +34,8 @@ Line wrap the file at 100 chars.                                              Th
   disable on-demand when stopping the tunnel from within the app.
 - Fix bug that caused the app to skip tunnel settings migration from older versions of the app.
 - Localize some of well known StoreKit errors so that they look less cryptic when presented to user.
+- Improve tunnel settings verification to address issues with broken tunnel and missing Keychain
+  entries to tunnel settings in cases such as when setting up a new device from backup.
 
 
 ## [2021.1] - 2021-03-16


jvff (approved):
:lgtm:

> Reviewed 2 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2738#-:-M_S2qpbEmZ_zYrWNd72:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2745, opened 2021-05-17
iOS 12 fixes for iPad app

pronebird:

Git checklist:

- Since login controller is now being wrapped in `RootContainer`, we have to proxy `disablesAutomaticKeyboardDismissal` to the child controllers to make sure that user can dismiss the keyboard when in formsheet presentation on iPad.
- On iOS 12, the adaptive presentation call to `willPresentWithAdaptiveStyle` often receives `.none` as the adaptive presentation style, which means that no change to adaptive presentation should be made and we should carry on with the modal presentation style that was originally set on the presented view controller. This fix attempts to resolve the actual presentation style before deciding how to adapt the presentation. On iOS 13+ this call often receives the actual presentation style (i.e `.formSheet`), which either way should work fine if we handle it in the proposed way.
- On iOS 12, `modalPresentationStyle` is only consulted at the time of presentation delegate assignment. Set `modalPresentationStyle` first to fix that.
- Fix active modal presentation detection in `AutomaticKeyboardResponder` when parent controller is a child of another controller. Consult presenting controller to find the modally presented ascendant up the view controller hierarchy.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2745)

### DIFF
--- a/ios/MullvadVPN/AppDelegate.swift
+++ b/ios/MullvadVPN/AppDelegate.swift
@@ -633,17 +633,22 @@ extension AppDelegate: UIAdaptivePresentationControllerDelegate {
         }
     }
 
-    func presentationController(_ controller: UIPresentationController, viewControllerForAdaptivePresentationStyle style: UIModalPresentationStyle) -> UIViewController? {
-        return nil
-    }
-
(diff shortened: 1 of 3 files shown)


jvff (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2745#-:-M_uvuXo5JpXkfF87kjL:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2750, opened 2021-05-18
Remove 'blocking internet' notification shown while (re)connecting

dlon:
While connecting, the frontend shows a "blocking internet" notification. This is not necessarily true anymore. While packets outside the tunnel are dropped (except to the relay), traffic in the tunnel is permitted as soon as the tunnel state machine is told that the device is up. This can occur before entering the connected state.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2750)

### DIFF
--- a/gui/src/renderer/components/NotificationArea.tsx
+++ b/gui/src/renderer/components/NotificationArea.tsx
@@ -4,13 +4,11 @@ import log from '../../shared/logging';
 import {
   BlockWhenDisconnectedNotificationProvider,
   CloseToAccountExpiryNotificationProvider,
-  ConnectingNotificationProvider,
   ErrorNotificationProvider,
   InAppNotificationProvider,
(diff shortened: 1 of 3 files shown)


raksooo:
For how long can the device be up before entering the connected state?

dlon:
> For how long can the device be up before entering the connected state?

In practice, up to a second, maybe. More if latency or packet loss is high. That said, the connected state is entered as soon as any traffic is received on the tunnel interface, so that may not be a big deal. The main point is that the daemon does not explicitly block the internet for the most part, except when using OpenVPN.

raksooo:
IMO a one second error margin would be okay. This would probably make it more confusing when the app is having trouble to connect and is stuck in the connecting state for >30 seconds. Would it be possible to send an event to the frontend when the device is up, and at that point remove the notification?

PR #2753, opened 2021-05-19
Wait for IP interfaces to be attached before setting metrics

dlon:
`WinNet_EnsureBestMetric` sometimes fails due to `LUID [...] does not specify any IPv4 or IPv6 interfaces` (`windows/winnet/src/winnet/interfacePair.cpp:25`). The code in this PR waits for the interfaces to attach after creating the tunnel device. If the error continues to occur, it must mean that an incorrect LUID is used for setting the metric (unlikely), or that the interfaces are destroyed at some later point in time.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2753)


mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2753)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2756, opened 2021-05-20
Improve gettext message escaping

jvff:
My initial plan for this PR was to implement the full escape sequences supported by gettext to make this future proof, but as I dug into the details I realized that's more complicated than what initially expected. So this PR just improves the escaping to cover "most" cases.

Gettext messages follow the same escape rules as C strings. That leads to some weird situations, because C strings aren't exactly UTF-8, and things can get weird, like `"\x1FF"` is undefined behavior, `'\e'` is not standardized, and `"\u00C1"` can become one or two bytes, depending on the implementation. (Examples taken from [Wikipedia](https://en.wikipedia.org/wiki/Escape_sequences_in_C)).

Therefore, this PR only handles escaping some of the most common character sequences. The PR also improves handling single quotes, because they don't have to be escaped, but can. The `MsgString` therefore stores it as unescaped when escaping, but allows creating an instance where it is escaped and normalization will then remove any escaped single quotes to avoid any issues.

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header. **Internal tool changes, no user visible changes.**

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2756)

### DIFF
--- a/android/translations-converter/src/gettext/msg_string.rs
+++ b/android/translations-converter/src/gettext/msg_string.rs
@@ -17,9 +17,17 @@ impl MsgString {
 
     /// Create a new `MsgString` from string without any escaped characters.
     ///
-    /// This will ensure that the string has the double quotes characters properly escaped.
+    /// This will ensure that the string has common C escape sequences properly created for special
+    /// characters. It will not attempt to escape non-ASCII characters and will just keep them as
+    /// UTF-8 characters.
     pub fn from_unescaped(string: &str) -> Self {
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2756)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2773, opened 2021-05-27
Use port from URI instead of always using 443 when connecting to the API

pinkisemils:
The hostname part of a URI would always be parsed as a socket address, or when that inevitably fails, it'd be parsed into an IP address and then combined with port 443 to be turned into a socket address. URI is actually already properly parsed by hyper, thus the port is already separated and parsed, so that is now used. If a port isn't a part of the specified URI, only then is 443 used.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2773)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ Line wrap the file at 100 chars.                                              Th
 ### Fixed
 - Fix relay selection failing to pick a WireGuard relay when no tunnel protocol is specified.
 - Fix time left not always being translated in desktop app settings.
+- Fix API address cache to use the supplied ports instead of always using port 443.
 
 #### Windows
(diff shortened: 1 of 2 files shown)


faern (approved):
:lgtm: This looks correct! However, I think it might be cleaner with less interleaved functionality. A rest client specific for the API where we know the URI is a socketaddr and a hostname supplied on the side vs the other case where we always have a "normal" URI. But right now we just need a quick fix, so ignore that this time :)

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2773#-:-MaiKVyfCnJ_mRfP5tVz:b-m48l96)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2778, opened 2021-05-31
Give Windows more time to init IP interfaces

dlon:
Previously, the daemon had a five-second timeout for waiting on the `MibAddInstance` event (i.e., for IP interfaces to attach to the tunnel device) after setting up a new Wintun device. Early after a reboot, this can take considerably more time to occur on some machines (30 seconds or more), causing some machines to get stuck in the error state, especially with auto-connect enabled.

The previous code blocked the tunnel state machine, so it had to time out quickly. To extend/remove the timeout, the tunnel monitors were updated to do this asynchronously/on a separate thread. The blocking code was also removed from `libwg` and now uses the same code as the openvpn tunnel monitor.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,7 +59,8 @@ Line wrap the file at 100 chars.                                              Th
 - Prevent tray icons from being extraced to `%TEMP%` directory.
 - Fix failure to create Wintun adapter due to a residual network interface by upgrading Wintun to
   0.10.4.
-- Wait for IP interfaces to be added to the Wintun adapter before setting metrics on them.
+- Wait indefinitely for IP interfaces to attach to the tunnel device to prevent early timeouts,
+  and errors setting interface metrics.
 - Prevent Microsoft Store from dropping packets in WireGuard tunnels.
 
 #### Linux
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2524,6 +2524,7 @@ dependencies = [
 name = "talpid-core"
 version = "0.1.0"
 dependencies = [
+ "async-trait",
  "atty",
  "byteorder",
  "cfg-if 1.0.0",
--- a/talpid-core/Cargo.toml
+++ b/talpid-core/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2018"
 publish = false
 
 [dependencies]
+async-trait = "0.1"
 atty = "0.2"
 cfg-if = "1.0"
 duct = "0.13"
--- a/talpid-core/src/routing/windows.rs
+++ b/talpid-core/src/routing/windows.rs
@@ -98,6 +98,11 @@ impl RouteManager {
         }
     }
 
+    /// Retrieve handle for the tokio runtime.
+    pub fn runtime_handle(&self) -> tokio::runtime::Handle {
+        self.runtime.clone()
+    }
+
     async fn listen(mut manage_rx: UnboundedReceiver<RouteManagerCommand>) {
         while let Some(command) = manage_rx.next().await {
             match command {
--- a/talpid-core/src/tunnel/mod.rs
+++ b/talpid-core/src/tunnel/mod.rs
@@ -24,6 +24,9 @@ pub mod wireguard;
 /// A module for low level platform specific tunnel device management.
 pub(crate) mod tun_provider;
 
+#[cfg(target_os = "windows")]
+mod windows;
+
 const OPENVPN_LOG_FILENAME: &str = "openvpn.log";
 const WIREGUARD_LOG_FILENAME: &str = "wireguard.log";
 
--- a/talpid-core/src/tunnel/openvpn/mod.rs
+++ b/talpid-core/src/tunnel/openvpn/mod.rs
@@ -22,7 +22,7 @@ use std::{
     process::ExitStatus,
     sync::{
         atomic::{AtomicBool, Ordering},
-        mpsc, Arc,
+        mpsc, Arc, Mutex,
     },
     thread,
     time::Duration,
@@ -33,12 +33,9 @@ use std::{collections::HashSet, net::IpAddr};
 use std::{
     ffi::{OsStr, OsString},
     os::windows::ffi::OsStrExt,
-    sync::Mutex,
     time::Instant,
 };
-use talpid_types::net::openvpn;
-#[cfg(any(windows, target_os = "linux"))]
-use talpid_types::ErrorExt;
+use talpid_types::{net::openvpn, ErrorExt};
 use tokio::task;
 #[cfg(target_os = "linux")]
 use which;
@@ -180,6 +177,10 @@ pub enum Error {
     #[error(display = "OpenVPN process died unexpectedly")]
     ChildProcessDied,
 
+    /// Failed before OpenVPN started
+    #[error(display = "Failed to start OpenVPN")]
+    StartProcessError,
+
     /// The IP routing program was not found.
     #[cfg(target_os = "linux")]
     #[error(display = "The IP routing program `ip` was not found")]
@@ -260,9 +261,15 @@ const OPENVPN_BIN_FILENAME: &str = "openvpn.exe";
 /// Struct for monitoring an OpenVPN process.
 #[derive(Debug)]
 pub struct OpenVpnMonitor<C: OpenVpnBuilder = OpenVpnCommand> {
-    child: Arc<C::ProcessHandle>,
+    spawn_task: Option<
+        tokio::task::JoinHandle<
+            std::result::Result<io::Result<C::ProcessHandle>, futures::future::Aborted>,
+        >,
+    >,
+    abort_spawn: futures::future::AbortHandle,
+
+    child: Arc<Mutex<Option<Arc<C::ProcessHandle>>>>,
     proxy_monitor: Option<Box<dyn ProxyMonitor>>,
-    log_path: Option<PathBuf>,
     closed: Arc<AtomicBool>,
     /// Keep the `TempFile` for the user-pass file in the struct, so it's removed on drop.
     _user_pass_file: mktemp::TempFile,
@@ -274,9 +281,52 @@ pub struct OpenVpnMonitor<C: OpenVpnBuilder = OpenVpnCommand> {
     server_join_handle: Option<task::JoinHandle<std::result::Result<(), event_server::Error>>>,
 
     #[cfg(windows)]
-    wintun_adapter: Option<windows::TemporaryWintunAdapter>,
-    #[cfg(windows)]
-    _wintun_logger: Option<windows::WintunLoggerHandle>,
+    wintun: Arc<Box<dyn WintunContext>>,
+}
+
+#[cfg(windows)]
+#[async_trait::async_trait]
+trait WintunContext: Send + Sync {
+    fn luid(&self) -> NET_LUID;
+    fn ipv6(&self) -> bool;
+    async fn wait_for_interfaces(&self) -> io::Result<()>;
+}
+
+#[cfg(windows)]
+impl std::fmt::Debug for dyn WintunContext {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "WintunContext {{ luid: {}, ipv6: {} }}",
+            self.luid().Value,
+            self.ipv6()
+        )
+    }
+}
+
+#[cfg(windows)]
+#[derive(Debug)]
+struct WintunContextImpl {
+    adapter: windows::TemporaryWintunAdapter,
+    wait_v6_interface: bool,
+    _logger: windows::WintunLoggerHandle,
+}
+
+#[cfg(windows)]
+#[async_trait::async_trait]
+impl WintunContext for WintunContextImpl {
+    fn luid(&self) -> NET_LUID {
+        self.adapter.adapter().luid()
+    }
+
+    fn ipv6(&self) -> bool {
+        self.wait_v6_interface
+    }
+
+    async fn wait_for_interfaces(&self) -> io::Result<()> {
+        let luid = self.adapter.adapter().luid();
+        super::windows::wait_for_interfaces(luid, true, self.wait_v6_interface).await
+    }
 }
 
 
@@ -399,14 +449,6 @@ impl OpenVpnMonitor<OpenVpnCommand> {
                 log::warn!("You may need to restart Windows to complete the install of Wintun");
             }
 
-            log::debug!("Wait for IP interfaces");
-            windows::wait_for_interfaces(
-                &adapter.adapter().luid(),
-                true,
-                params.generic_options.enable_ipv6,
-            )
-            .map_err(Error::IpInterfacesError)?;
-
             let assigned_guid = adapter.adapter().guid();
             let assigned_guid = assigned_guid.as_ref().unwrap_or_else(|error| {
                 log::error!(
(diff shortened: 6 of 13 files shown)


pinkisemils (changes requested):
> Reviewed 5 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778)* status: 5 of 13 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/tunnel/windows.rs, line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5pPXu1IJoKReAmbq_:-Mb5pPXv7RjQa4ezta0o:b-7uwbc6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/windows.rs#L15)):*
> ```XML
> /// Context for [`notify_ip_interface_change`]. When it is dropped,
> /// the callback is unregistered.
> pub struct IpNotifierHandle<'a> {
> ```

Could this not be refactored to be a `Handle` and a `InnerHandle`, like `Handle { inner: Mutex<InnerHandle> }`? Or is it the case that the raw handle should be cancelled before acquiring the mutex as to not get into a deadlock?
___
*[talpid-core/src/tunnel/windows.rs, line 98 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5qYTEDyEeBMDkEA4f:-Mb5qYTEDyEeBMDkEA4g:b-d56sww) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/windows.rs#L98)):*
> ```XML
>     match get_ip_interface_entry(family, luid) {
>         Ok(_) => Ok(true),
>         Err(error) => match error.raw_os_error() {
> ```

You could lose the nested match statement by having two error cases
`Err(error) if error.raw_os_error() == Some(ERROR_NOT_FOUND) => Ok(false)` and
`Err(error) => Err(error)`.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778)* status: 5 of 13 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/tunnel/windows.rs, line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5pPXu1IJoKReAmbq_:-Mb5xALm7BU3WnHRQtZ-:b-vqp6sq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/windows.rs#L15)):*

Could this not be refactored to be a `Handle` and a `InnerHandle`, like `Handle { inner: Mutex<InnerHandle> }`? Or is it the case that the raw handle should be cancelled before acquiring the mutex as to not get into a deadlock?
</blockquote></details>

It prevents use after free. Consider this outcome in `IpNotifierHandle::drop()`:
```
let inner = mutex.lock();
// an event occurs, and inner_callback waits to acquire the mutex
cancel(inner.raw_handle);
// mutex is released, IpNotifierHandle is destroyed
// inner_callback is still using the inner context/mutex which has been destroyed
```

pinkisemils (changes requested):
> Reviewed 7 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778)* status: 12 of 13 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/tunnel/windows.rs, line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5pPXu1IJoKReAmbq_:-Mb62ujQCc_hFesj7t7S:b-iovvxc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/windows.rs#L15)):*

It prevents use after free. Consider this outcome in `IpNotifierHandle::drop()`:
```
let inner = mutex.lock();
// an event occurs, and inner_callback waits to acquire the mutex
cancel(inner.raw_handle);
// mutex is released
// inner_callback is still using the inner context/mutex which has been destroyed
```
</blockquote></details>

Ah, so it'd have to be a`Mutex<Option<InnerHandle>>` then, no ?
___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 762 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5z41pCf0Q39XpgE_I:-Mb5z41q9lnBZ1Wh-fof:bkhttwy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/openvpn/mod.rs#L762)):*
> ```XML
>             Some(task) => task,
>             None => {
>                 log::error!("BUG: spawn_task == None");
> ```

I wonder if it'd make sense to panic here. Most probably not because `wait_tunnel()` can be called multiple times. Then again, `wait_tunnel()` could take `self` instead of `&mut self` it seems. In which case, I'd advocate for panicking here. This would definitely be caught in a test.
___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 948 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb60zJ2BaL5jkSBZ78x:-Mb60zJ2BaL5jkSBZ78y:bhtuzzn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/openvpn/mod.rs#L948)):*
> ```XML
>                 child.kill()
>             } else {
>                 self.abort_spawn.abort();
> ```

Isn't there a small chance of still spawning a child here? Regardless, I think there's no harm and it's simpler to always abort the spawn first, and then kill the child if it exists.
___
*[talpid-core/src/tunnel/wireguard/mod.rs, line 229 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb62D363WRYnOaKoTA9:-Mb62D363WRYnOaKoTAA:b3ov5cm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/wireguard/mod.rs#L229)):*
> ```XML
>                     futures::select! {
>                         result = setup_future.fuse() => {
>                             match result {
> ```

This match statement could be a dirty `.map_err(|err| iface_close_sender.send(...` hack. Less lines and a bit more readable, I think.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778)* status: 10 of 13 files reviewed, 3 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-core/src/tunnel/windows.rs, line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5pPXu1IJoKReAmbq_:-Mb66zL0ES6c-4T-1S80:b-odjt40) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/windows.rs#L15)):*

Ah, so it'd have to be a`Mutex<Option<InnerHandle>>` then, no ?
</blockquote></details>

Still doesn't work, I think. The pointer to `IpNotifierHandle` is invalid after it is dropped, and the mutex has been dropped at that point.
___
*[talpid-core/src/tunnel/windows.rs, line 98 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5qYTEDyEeBMDkEA4f:-Mb6Lhj40W5CHago8ZFv:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/windows.rs#L98)):*

You could lose the nested match statement by having two error cases
`Err(error) if error.raw_os_error() == Some(ERROR_NOT_FOUND) => Ok(false)` and
`Err(error) => Err(error)`.
</blockquote></details>

Done.
___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 762 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5z41pCf0Q39XpgE_I:-Mb6Ls9T3zC7TmEmJBGP:b-qri8p7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/openvpn/mod.rs#L762)):*

I wonder if it'd make sense to panic here. Most probably not because `wait_tunnel()` can be called multiple times. Then again, `wait_tunnel()` could take `self` instead of `&mut self` it seems. In which case, I'd advocate for panicking here. This would definitely be caught in a test.
</blockquote></details>

Done. Usually when I hesitate, I go for not panicking. But it really should be impossible here.
___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 948 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb60zJ2BaL5jkSBZ78x:-Mb6M4fDEI8NOF0CmaeD:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/openvpn/mod.rs#L948)):*

Isn't there a small chance of still spawning a child here? Regardless, I think there's no harm and it's simpler to always abort the spawn first, and then kill the child if it exists.
</blockquote></details>

Done.
___
*[talpid-core/src/tunnel/wireguard/mod.rs, line 229 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb62D363WRYnOaKoTA9:-Mb6M93P9oH1pNBWuRm5:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/wireguard/mod.rs#L229)):*

This match statement could be a dirty `.map_err(|err| iface_close_sender.send(...` hack. Less lines and a bit more readable, I think.
</blockquote></details>

Done.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @dlon and @pinkisemils)

___
*[talpid-core/src/tunnel/windows.rs, line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5pPXu1IJoKReAmbq_:-Mb72YYU-CJ157B5HSbf:b-v270un) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/windows.rs#L15)):*

Still doesn't work, I think. The pointer to `IpNotifierHandle` is invalid after it is dropped, and the mutex has been dropped at that point.
</blockquote></details>

It'd have to be an `Arc<Mutex<Option<<<InnerHandle>>>`, and  the callback should try and convert the pointer to a weak reference, which then you'd ultimately try and upgrade to a strong one.
___
*[talpid-core/src/tunnel/openvpn/mod.rs, line 762 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5z41pCf0Q39XpgE_I:-Mb6kSxvBMWzDS-hjgHF:b-pjgexn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/openvpn/mod.rs#L762)):*

Done. Usually when I hesitate, I go for not panicking. But it really should be impossible here.
</blockquote></details>

I agree, but we have tests that should exercise this function anyway, right? We might benefit from some coverage tools.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/tunnel/windows.rs, line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5pPXu1IJoKReAmbq_:-Mb7Cj0G5cM_rlfcnzyV:bo5dq3r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/windows.rs#L15)):*

It'd have to be an `Arc<Mutex<Option<<<InnerHandle>>>`, and  the callback should try and convert the pointer to a weak reference, which then you'd ultimately try and upgrade to a strong one.
</blockquote></details>

Wouldn't the `Arc` itself be destroyed along with `IpNotifierHandle`? I might need a more complete explanation.

I've made some changes. The mutex wraps the callback function, which is the only thing that is being accessed in the callback anyway.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @dlon and @pinkisemils)

___
*[talpid-core/src/tunnel/windows.rs, line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5pPXu1IJoKReAmbq_:-Mb7XV4W9UdgBk1jja8M:bl39e0r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/windows.rs#L15)):*

Wouldn't the `Arc` itself be destroyed along with `IpNotifierHandle`? I might need a more complete explanation.

I've made some changes. The mutex wraps the callback function, which is the only thing that is being accessed in the callback anyway.
</blockquote></details>

It would be, but it'd be safe to check if it's still alive from the callback. Isn't the issue here that `inner_callback` be called when it's context has already been dropped? If not, then the current solution is good enough. The issue with using a `Mutex` without an `Arc` is that you have to be really certain about the lifetimes of of the stuff that's being protected by the mutex. In case of a mutex being accessed by multiple threads, you'd still want an Arc, right? If you don't need an arc, do you need a mutex?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/tunnel/windows.rs, line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2778#-Mb5pPXu1IJoKReAmbq_:-MbB4SrJ-rBfYppMrpNm:b-2ogq0q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1a5907fca68046a6f883141b62254a1449766a76/talpid-core/src/tunnel/windows.rs#L15)):*

It would be, but it'd be safe to check if it's still alive from the callback. Isn't the issue here that `inner_callback` be called when it's context has already been dropped? If not, then the current solution is good enough. The issue with using a `Mutex` without an `Arc` is that you have to be really certain about the lifetimes of of the stuff that's being protected by the mutex. In case of a mutex being accessed by multiple threads, you'd still want an Arc, right? If you don't need an arc, do you need a mutex?
</blockquote></details>

I've found that `CancelMibChangeNotify2` doesn't return until all calls to `inner_callback` are complete. So the code as it is works correctly.

PR #2784, opened 2021-06-04
Handle requests to split tunnel driver on a separate thread

dlon:
Occasionally, e.g. because some program is refusing to release the WFP transaction lock, `DeviceIoControl` gets stuck and blocks the tunnel monitor thread. The expected behavior was that `DeviceIoControl` would always pend operations and blocking would occur due to `WaitForSingleObject`, eventually timing out. But if there are no ongoing operations, at least, all processing seems to occur on the application thread. As a result, the entire daemon might lock up.

To fix this, this PR makes all driver requests on a dedicated thread, and refuses to queue further requests in case the driver becomes unresponsive. This also protects the daemon against driver deadlocks, should they exist.

### DIFF
--- a/talpid-core/src/split_tunnel/windows/driver.rs
+++ b/talpid-core/src/split_tunnel/windows/driver.rs
@@ -124,20 +124,13 @@ pub enum SplittingChangeReason {
 
 pub struct DeviceHandle {
     handle: fs::File,
-    overlapped: OVERLAPPED,
 }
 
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2784)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2785, opened 2021-06-04
Fix small into-tunnel leak window for excluded apps

dlon:
Previously, the driver was not given the IP address of the tunnel interface until it had reached the `Connected` state. This is usually fine: All TCP connections where the destination addresses is a public IP address are redirected to the specified non-tunnel interface. Also, non-TCP (eg UDP) sockets that are bound to 0.0.0.0 (or implicitly bound) are rebound correctly. UDP sockets that were explicitly bound to the tunnel interface could still communicate inside the tunnel, though.

This PR addresses this by registering tunnel address with the driver in the `InterfaceUp`  tunnel event, before routes are set up.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2785)

### DIFF
--- a/talpid-core/src/tunnel/mod.rs
+++ b/talpid-core/src/tunnel/mod.rs
@@ -73,8 +73,8 @@ pub enum Error {
 pub enum TunnelEvent {
     /// Sent when the tunnel fails to connect due to an authentication error.
     AuthFailed(Option<String>),
-    /// Sent when the tunnel interface has been created.
-    InterfaceUp(String),
+    /// Sent when the tunnel interface has been created, before routes are set up.
+    InterfaceUp(TunnelMetadata),
     /// Sent when the tunnel comes up and is ready for traffic.
     Up(TunnelMetadata),
     /// Sent when the tunnel goes down.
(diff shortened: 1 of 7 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2785)* status: 0 of 7 files reviewed, 1 unresolved discussion

___
*[talpid-core/src/tunnel/wireguard/mod.rs, line 224 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2785#-MbMbHpBATQbEl0Fs7xk:-MbMbHpBATQbEl0Fs7xl:b9wih0q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/141838af2ebeadd0040b18dcf90b6c05af9e4abd/talpid-core/src/tunnel/wireguard/mod.rs#L224)):*
> ```XML
> 
>         std::thread::spawn(move || {
>             (on_event)(TunnelEvent::InterfaceUp(metadata.clone()));
> ```

There is still a tiny risk for leaks here, iff `set_split_tunnel_addresses` fails and then disconnects, since routes may still be added (there is a race involving `stop_setup_tx`).

Maybe `InterfaceUp` specifically needs to be given a channel with a response. Or a different design should be used in the first place (eg, a handle could be passed in and used to call `set_split_tunnel_addresses`).

pinkisemils (commented):
> Reviewed 4 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2785)* status: 4 of 7 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/tunnel/wireguard/mod.rs, line 224 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2785#-MbMbHpBATQbEl0Fs7xk:-Mb_udPf1RXg7INxCO6V:bf3x2hz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/141838af2ebeadd0040b18dcf90b6c05af9e4abd/talpid-core/src/tunnel/wireguard/mod.rs#L224)):*

There is still a tiny risk for leaks here, iff `set_split_tunnel_addresses` fails and then disconnects, since routes may still be added (there is a race involving `stop_setup_tx`).

Maybe `InterfaceUp` specifically needs to be given a channel with a response. Or a different design should be used in the first place (eg, a handle could be passed in and used to call `set_split_tunnel_addresses`).
</blockquote></details>

If we could futurify this, I imagine it'd help with stopping unwanted stuff quicker, maybe? I don't think that's feasible though. A proper way to deal with this might be reverting changes we've made to the firewall to use it for controlling leaks here. Then it doesn't matter if there are routes setup or not, as the last step for entering the connected state would be to reset the firewall to ensure no leaks.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2785)* status: 4 of 7 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/tunnel/wireguard/mod.rs, line 224 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2785#-MbMbHpBATQbEl0Fs7xk:-Mb_vl1v9KRjUz_wCpKP:b-isnw94) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/141838af2ebeadd0040b18dcf90b6c05af9e4abd/talpid-core/src/tunnel/wireguard/mod.rs#L224)):*

If we could futurify this, I imagine it'd help with stopping unwanted stuff quicker, maybe? I don't think that's feasible though. A proper way to deal with this might be reverting changes we've made to the firewall to use it for controlling leaks here. Then it doesn't matter if there are routes setup or not, as the last step for entering the connected state would be to reset the firewall to ensure no leaks.
</blockquote></details>

Those firewall rules are overridden, so that wouldn't work, unfortunately. I think we do need to register the addresses before setting up routes.

pinkisemils (approved):
> Reviewed 3 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2785)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2792, opened 2021-06-09
Fix broken prepare_release.sh. Don't pass args to env.sh

faern:
Just doing `source env.sh` automatically passes all command line arguments from the parent script down to `env.sh`. This does not work as `env.sh` expects the target as the first argument, which is not what we have as arguments to `prepare_release.sh`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2792)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -242,7 +242,7 @@ function buildTargets() {
             build "$t"
         done
     else
-        source env.sh
+        source env.sh ""
         build
     fi
 }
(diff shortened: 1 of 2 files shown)


raksooo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2792#-:-MbkWKvGCQqT_t2Rj9fa:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2793, opened 2021-06-09
Fix issue of daemon not restarting when "fast startup" is enabled

dlon:
Sometimes, e.g. when `openvpn.exe` cannot be killed gracefully, it can take around 30 seconds to stop the service. `net stop mullvadvpn`, which was used to restart the service after resuming from "hibernation", has a timeout that is approximately 30 seconds as well, so it sometimes fails before the daemon has stopped. Rather than relying on `cmd /c "net stop mullvadvpn & net start mullvadvpn"`, a `--restart-service` option was added, and `mullvad-daemon --restart-service` is run instead. This has a timeout of two minutes.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ Line wrap the file at 100 chars.                                              Th
 - Make offline monitor aware of routing table changes.
 - Assign local DNS servers to more appropriate interfaces when using systemd-resolved.
 
+#### Windows
+- Fix failure to restart the daemon when resuming from "fast startup" hibernation.
+
 
 ## [2021.4-beta1] - 2021-06-09
 This release is for desktop only.
--- a/mullvad-daemon/src/cli.rs
+++ b/mullvad-daemon/src/cli.rs
@@ -9,6 +9,7 @@ pub struct Config {
     pub log_stdout_timestamps: bool,
     pub run_as_service: bool,
     pub register_service: bool,
+    pub restart_service: bool,
 }
 
 pub fn get_config() -> &'static Config {
@@ -32,6 +33,7 @@ pub fn create_config() -> Config {
 
     let run_as_service = cfg!(windows) && matches.is_present("run_as_service");
     let register_service = cfg!(windows) && matches.is_present("register_service");
+    let restart_service = cfg!(windows) && matches.is_present("restart_service");
 
     Config {
         log_level,
@@ -39,6 +41,7 @@ pub fn create_config() -> Config {
         log_stdout_timestamps,
         run_as_service,
         register_service,
+        restart_service,
     }
 }
 
@@ -91,10 +94,16 @@ fn create_app() -> App<'static, 'static> {
             Arg::with_name("run_as_service")
                 .long("run-as-service")
                 .help("Run as a system service. On Windows this option must be used when running a system service"),
-        ).arg(
+        )
+        .arg(
             Arg::with_name("register_service")
                 .long("register-service")
                 .help("Register itself as a system service"),
+        )
+        .arg(
+            Arg::with_name("restart_service")
+                .long("restart-service")
+                .help("Restarts the existing system service"),
         )
     }
     app
--- a/mullvad-daemon/src/main.rs
+++ b/mullvad-daemon/src/main.rs
@@ -75,6 +75,12 @@ fn get_log_dir(config: &cli::Config) -> Result<Option<PathBuf>, String> {
 async fn run_platform(config: &cli::Config, log_dir: Option<PathBuf>) -> Result<(), String> {
     if config.run_as_service {
         system_service::run()
+    } else if config.restart_service {
+        let restart_result = system_service::restart_service().map_err(|e| e.display_chain());
+        if restart_result.is_ok() {
+            log::info!("Restarted the service.");
+        }
+        restart_result
     } else {
         if config.register_service {
             let install_result = system_service::install_service().map_err(|e| e.display_chain());
--- a/mullvad-daemon/src/system_service.rs
+++ b/mullvad-daemon/src/system_service.rs
@@ -2,7 +2,7 @@ use crate::cli;
 use mullvad_daemon::{runtime::new_runtime_builder, DaemonShutdownHandle};
 use std::{
     env,
-    ffi::OsString,
+    ffi::{OsStr, OsString},
     ptr, slice,
     sync::{
         atomic::{AtomicBool, AtomicUsize, Ordering},
@@ -45,6 +45,8 @@ static SERVICE_TYPE: ServiceType = ServiceType::OWN_PROCESS;
 
 const SERVICE_RECOVERY_LAST_RESTART_DELAY: Duration = Duration::from_secs(60 * 10);
 const SERVICE_FAILURE_RESET_PERIOD: Duration = Duration::from_secs(60 * 15);
+
+const SERVICE_RESTART_TIMEOUT: Duration = Duration::from_secs(60 * 2);
 
 lazy_static::lazy_static! {
     static ref SERVICE_ACCESS: ServiceAccess = ServiceAccess::QUERY_CONFIG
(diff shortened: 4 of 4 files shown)


mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[CHANGELOG.md, line 77 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793#-MbkcafI0od73Iipg_i8:-MbkcafI0od73Iipg_i9:b-ti0n8v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5c7ac16dc49b85fcd11efe3b5898e419489882f/CHANGELOG.md#L77)):*
> ```
> start
> ```

`restart`?
___
*[mullvad-daemon/src/system_service.rs, line 435 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793#-MbkigAECnPE6s7vZ41_:-MbkigAF74dowiH8RcYW:bci4vjl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5c7ac16dc49b85fcd11efe3b5898e419489882f/mullvad-daemon/src/system_service.rs#L435)):*
> ```XML
>         if status.current_state == ServiceState::Stopped {
>             break service
>                 .start(&*LAUNCH_ARGUMENTS)
> ```

Are you sure this has to be specified again? I'm surprised it's even possible to override the command line which is registered for the service.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[mullvad-daemon/src/system_service.rs, line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793#-MblERpK-FZJs1wsrvZB:-MbowH7k9EcymI2qfE2a:b-xtjb28) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5c7ac16dc49b85fcd11efe3b5898e419489882f/mullvad-daemon/src/system_service.rs#L439)):*

If you find that execution gets stuck here, in the loop I mean, you could locate all instances of `openvpn.exe` which have been spawned by the daemon, and terminate them forcefully.
</blockquote></details>

We have spawned them and should have handles to them. There should not be any need to iterate all and find which ones belong to us. We should already know that. Otherwise it feels like a lot of duct tape and just working around the fact that something is not killing openvpn properly. Sounds way cleaner to find out why the tunnel state machine does not terminate openVPN in a timely manner. It has a built in timeout of just a few seconds before it should force-kill openvpn already.

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/system_service.rs, line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793#-MblERpK-FZJs1wsrvZB:-Mbp32DCAC6viwLdPW6J:bbd1wrq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5c7ac16dc49b85fcd11efe3b5898e419489882f/mullvad-daemon/src/system_service.rs#L439)):*

We have spawned them and should have handles to them. There should not be any need to iterate all and find which ones belong to us. We should already know that. Otherwise it feels like a lot of duct tape and just working around the fact that something is not killing openvpn properly. Sounds way cleaner to find out why the tunnel state machine does not terminate openVPN in a timely manner. It has a built in timeout of just a few seconds before it should force-kill openvpn already.
</blockquote></details>

Well, the handle(s) are stored in another instance of the daemon, so we can't just access them.

Isn't the issue that the timeout is too long, 30 seconds?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon, @faern, and @mvd-ows)

___
*[mullvad-daemon/src/system_service.rs, line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793#-MblERpK-FZJs1wsrvZB:-Mbp1LME6bRlKMB3b2dz:b-urj2a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5c7ac16dc49b85fcd11efe3b5898e419489882f/mullvad-daemon/src/system_service.rs#L439)):*

We have spawned them and should have handles to them. There should not be any need to iterate all and find which ones belong to us. We should already know that. Otherwise it feels like a lot of duct tape and just working around the fact that something is not killing openvpn properly. Sounds way cleaner to find out why the tunnel state machine does not terminate openVPN in a timely manner. It has a built in timeout of just a few seconds before it should force-kill openvpn already.
</blockquote></details>

It is force-killed after 30 seconds on Windows. I'm not sure why the interval is so much greater on Windows, but it does work as it should.

https://github.com/mullvad/mullvadvpn-app/blob/0bdcecee4f0848afc759e6ca6269a5e5170d477f/talpid-core/src/tunnel/openvpn/mod.rs#L246

As for why OpenVPN sometimes does not gracefully, I suspect there's a race in our patch. The thread that waits for EOF may set `signal_received` to ["SIGTERM"](https://github.com/mullvad/openvpn/blob/64d91f8e931e4e656a90057cadf643a707ed6e50/src/openvpn/openvpn.c#L125) before it is [cleared](https://github.com/mullvad/openvpn/blob/64d91f8e931e4e656a90057cadf643a707ed6e50/src/openvpn/openvpn.c#L200). If the handle is closed too soon, the "signal" is ignored.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon, @faern, and @mvd-ows)

___
*[mullvad-daemon/src/system_service.rs, line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793#-MblERpK-FZJs1wsrvZB:-Mbp3jLt3Rie2s7jyVNA:bpthalg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5c7ac16dc49b85fcd11efe3b5898e419489882f/mullvad-daemon/src/system_service.rs#L439)):*

It is force-killed after 30 seconds on Windows. I'm not sure why the interval is so much greater on Windows, but it does work as it should.

https://github.com/mullvad/mullvadvpn-app/blob/0bdcecee4f0848afc759e6ca6269a5e5170d477f/talpid-core/src/tunnel/openvpn/mod.rs#L246

As for why OpenVPN sometimes does not gracefully, I suspect there's a race in our patch. The thread that waits for EOF may set `signal_received` to ["SIGTERM"](https://github.com/mullvad/openvpn/blob/64d91f8e931e4e656a90057cadf643a707ed6e50/src/openvpn/openvpn.c#L125) before it is [cleared](https://github.com/mullvad/openvpn/blob/64d91f8e931e4e656a90057cadf643a707ed6e50/src/openvpn/openvpn.c#L200). If the handle is closed too soon, the "signal" is ignored.
</blockquote></details>

I think we should work around this by making sure that the process has been alive for some short interval before dropping the handle. But also the patch should be fixed so that the workaround can be removed the next time we upgrade/rebuild OpenVPN.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @faern and @mvd-ows)

___
*[mullvad-daemon/src/system_service.rs, line 435 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2793#-MbkigAECnPE6s7vZ41_:-MbpJwcz7-2c-GoAAd8c:byo8zq5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5c7ac16dc49b85fcd11efe3b5898e419489882f/mullvad-daemon/src/system_service.rs#L435)):*

Are you sure this has to be specified again? I'm surprised it's even possible to override the command line which is registered for the service.
</blockquote></details>

Good question. These arguments are not necessary. I'd prefer to pass a `nullptr` (to `StartServiceW`), but this crate does not let me do that. An empty list works too, though.

PR #2797, opened 2021-06-14
Improve OpenVPN event handling

dlon:
This PR updates the OpenVPN tunnel monitor to stop relying on panics for error handling and execute async/sync code correctly.

Implementing all event handling in the service trait (with async methods) was a simple improvement over using a sync callback function.

Instead of panicking, errors are returned to `talpid-openvpn-plugin`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2797)

PR #2799, opened 2021-06-14
Refactor OpenVPN event handling

dlon:
The last PR (#2797) was reverted. It turns out that returning an error in the route up event didn't *reliably* shut down the monitor. Often it would just keep going. To fix this, the event server is now told to shut down if an error occurs. This doesn't work unless the client is also dropped, so the plugin was updated to drop the client as well.

Only the last commit differs from the previous PR.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2799)

### DIFF
--- a/talpid-core/src/tunnel/mod.rs
+++ b/talpid-core/src/tunnel/mod.rs
@@ -1,7 +1,5 @@
 use self::tun_provider::TunProvider;
 use crate::{logging, routing::RouteManager};
-#[cfg(not(target_os = "android"))]
-use std::collections::HashMap;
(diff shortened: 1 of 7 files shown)


pinkisemils (approved):
> Reviewed 2 of 5 files at r1, 3 of 4 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2799)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2801, opened 2021-06-16
Only store a single account number in the history

dlon:
Overview of changes:
* The account history only contains the last used account. It is only used to remember the most recently used account when logged out. `AccountHistory` doesn't do any key management.
* The WireGuard key and addresses are stored in the settings.
* The file permissions of `settings.json` are set to `600` due to the change mentioned in the previous point.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ Line wrap the file at 100 chars.                                              Th
 
 ## [Unreleased]
 
+### Changed
+- Only use the account history file to store the last used account.
+
 ### Fixed
 - Fix link to download page not always using the beta URL when it should.
 
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ipc/Event.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ipc/Event.kt
@@ -17,7 +17,7 @@ sealed class Event : Message.EventMessage() {
     protected override val messageKey = MESSAGE_KEY
 
     @Parcelize
-    data class AccountHistory(val history: List<String>?) : Event()
+    data class AccountHistory(val history: String?) : Event()
 
     @Parcelize
     data class AppVersionInfo(val versionInfo: AppVersionInfoData?) : Event()
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ipc/Request.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ipc/Request.kt
@@ -54,7 +54,7 @@ sealed class Request : Message.RequestMessage() {
     data class RegisterListener(val listener: Messenger) : Request()
 
     @Parcelize
-    data class RemoveAccountFromHistory(val account: String?) : Request()
+    object ClearAccountHistory : Request()
 
     @Parcelize
     data class RemoveCustomDnsServer(val address: InetAddress) : Request()
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadDaemon.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/MullvadDaemon.kt
@@ -52,7 +52,7 @@ class MullvadDaemon(val vpnService: MullvadVpnService) {
         return getAccountData(daemonInterfaceAddress, accountToken)
     }
 
-    fun getAccountHistory(): ArrayList<String>? {
+    fun getAccountHistory(): String? {
         return getAccountHistory(daemonInterfaceAddress)
     }
 
@@ -92,8 +92,8 @@ class MullvadDaemon(val vpnService: MullvadVpnService) {
         reconnect(daemonInterfaceAddress)
     }
 
-    fun removeAccountFromHistory(accountToken: String) {
-        removeAccountFromHistory(daemonInterfaceAddress, accountToken)
+    fun clearAccountHistory() {
+        clearAccountHistory(daemonInterfaceAddress)
     }
 
     fun setAccount(accountToken: String?) {
@@ -159,7 +159,7 @@ class MullvadDaemon(val vpnService: MullvadVpnService) {
         daemonInterfaceAddress: Long,
         accountToken: String
     ): GetAccountDataResult
-    private external fun getAccountHistory(daemonInterfaceAddress: Long): ArrayList<String>?
+    private external fun getAccountHistory(daemonInterfaceAddress: Long): String?
     private external fun getWwwAuthToken(daemonInterfaceAddress: Long): String?
     private external fun getCurrentLocation(daemonInterfaceAddress: Long): GeoIpLocation?
     private external fun getCurrentVersion(daemonInterfaceAddress: Long): String?
@@ -169,10 +169,7 @@ class MullvadDaemon(val vpnService: MullvadVpnService) {
     private external fun getVersionInfo(daemonInterfaceAddress: Long): AppVersionInfo?
     private external fun getWireguardKey(daemonInterfaceAddress: Long): PublicKey?
     private external fun reconnect(daemonInterfaceAddress: Long)
-    private external fun removeAccountFromHistory(
-        daemonInterfaceAddress: Long,
-        accountToken: String
-    )
+    private external fun clearAccountHistory(daemonInterfaceAddress: Long)
     private external fun setAccount(daemonInterfaceAddress: Long, accountToken: String?)
     private external fun setAllowLan(daemonInterfaceAddress: Long, allowLan: Boolean)
     private external fun setAutoConnect(daemonInterfaceAddress: Long, alwaysOn: Boolean)
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/service/endpoint/AccountCache.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/service/endpoint/AccountCache.kt
@@ -42,7 +42,7 @@ class AccountCache(private val endpoint: ServiceEndpoint) {
 
     val onAccountNumberChange = EventNotifier<String?>(null)
     val onAccountExpiryChange = EventNotifier<DateTime?>(null)
-    val onAccountHistoryChange = EventNotifier<List<String>>(listOf<String>())
+    val onAccountHistoryChange = EventNotifier<String?>(null)
     val onLoginStatusChange = EventNotifier<LoginStatus?>(null)
 
     var newlyCreatedAccount = false
@@ -104,10 +104,8 @@ class AccountCache(private val endpoint: ServiceEndpoint) {
                 invalidateAccountExpiry(request.expiry)
             }
 
-            registerHandler(Request.RemoveAccountFromHistory::class) { request ->
-                request.account?.let { account ->
-                    removeAccountFromHistory(account)
-                }
+            registerHandler(Request.ClearAccountHistory::class) { _ ->
+                clearAccountHistory()
             }
         }
     }
@@ -162,9 +160,9 @@ class AccountCache(private val endpoint: ServiceEndpoint) {
         }
     }
 
-    private fun removeAccountFromHistory(accountToken: String) {
-        jobTracker.newBackgroundJob("removeAccountFromHistory $accountToken") {
-            daemon.await().removeAccountFromHistory(accountToken)
+    private fun clearAccountHistory() {
+        jobTracker.newBackgroundJob("clearAccountHistory") {
+            daemon.await().clearAccountHistory()
             fetchAccountHistory()
         }
     }
@@ -227,7 +225,7 @@ class AccountCache(private val endpoint: ServiceEndpoint) {
 
     private fun fetchAccountHistory() {
         jobTracker.newBackgroundJob("fetchHistory") {
-            daemon.await().getAccountHistory()?.let { history ->
+            daemon.await().getAccountHistory().let { history ->
                 accountHistory = history
             }
         }
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/LoginFragment.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/LoginFragment.kt
@@ -51,7 +51,7 @@ class LoginFragment : ServiceDependentFragment(OnNoService.GoToLaunchScreen), Na
 
         accountLogin = view.findViewById<AccountLogin>(R.id.account_login).apply {
             onLogin = { accountToken -> login(accountToken) }
-            onRemoveFromHistory = { account -> accountCache.removeAccountFromHistory(account) }
+            onClearHistory = { -> accountCache.clearAccountHistory() }
         }
 
         view.findViewById<Button>(R.id.create_account)
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/serviceconnection/AccountCache.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/serviceconnection/AccountCache.kt
@@ -11,7 +11,7 @@ import org.joda.time.DateTime
 class AccountCache(private val connection: Messenger, eventDispatcher: EventDispatcher) {
     val onAccountNumberChange = EventNotifier<String?>(null)
     val onAccountExpiryChange = EventNotifier<DateTime?>(null)
-    val onAccountHistoryChange = EventNotifier<List<String>>(listOf<String>())
+    val onAccountHistoryChange = EventNotifier<String?>(null)
     val onLoginStatusChange = EventNotifier<LoginStatus?>(null)
 
     private var accountHistory by onAccountHistoryChange.notifiable()
@@ -20,7 +20,7 @@ class AccountCache(private val connection: Messenger, eventDispatcher: EventDisp
     init {
         eventDispatcher.apply {
             registerHandler(Event.AccountHistory::class) { event ->
-                accountHistory = event.history ?: listOf<String>()
+                accountHistory = event.history
             }
 
             registerHandler(Event.LoginStatus::class) { event ->
@@ -54,8 +54,8 @@ class AccountCache(private val connection: Messenger, eventDispatcher: EventDisp
         connection.send(request.message)
     }
 
-    fun removeAccountFromHistory(account: String) {
-        connection.send(Request.RemoveAccountFromHistory(account).message)
+    fun clearAccountHistory() {
+        connection.send(Request.ClearAccountHistory.message)
     }
 
     fun onDestroy() {
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/widget/AccountHistoryAdapter.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/widget/AccountHistoryAdapter.kt
@@ -14,12 +14,12 @@ class AccountHistoryAdapter : Adapter<AccountHistoryHolder>() {
         }
     }
 
-    var accountHistory by observable(listOf<String>()) { _, _, _ ->
+    var accountHistory by observable<String?>(null) { _, _, _ ->
         notifyDataSetChanged()
     }
 
     var onSelectEntry: ((String) -> Unit)? = null
-    var onRemoveEntry: ((String) -> Unit)? = null
+    var onRemoveEntry: (() -> Unit)? = null
     var onChildFocusChanged: ((String, Boolean) -> Unit)? = null
 
     override fun onCreateViewHolder(parentView: ViewGroup, type: Int): AccountHistoryHolder {
@@ -28,14 +28,14 @@ class AccountHistoryAdapter : Adapter<AccountHistoryHolder>() {
 
         return AccountHistoryHolder(view, formatter).apply {
             onSelect = { account -> onSelectEntry?.invoke(account) }
-            onRemove = { account -> onRemoveEntry?.invoke(account) }
+            onRemove = { _ -> onRemoveEntry?.invoke() }
             onFocusChanged = { account, hasFocus -> onChildFocusChanged?.invoke(account, hasFocus) }
         }
     }
 
     override fun onBindViewHolder(holder: AccountHistoryHolder, position: Int) {
-        holder.accountToken = accountHistory[position]
+        holder.accountToken = accountHistory ?: ""
     }
 
-    override fun getItemCount() = accountHistory.size
+    override fun getItemCount() = if (accountHistory !== null) 1 else 0
 }
--- a/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/widget/AccountLogin.kt
+++ b/android/src/main/kotlin/net/mullvad/mullvadvpn/ui/widget/AccountLogin.kt
@@ -99,13 +99,12 @@ class AccountLogin : RelativeLayout {
     val hasFocus
         get() = focused
 
-    var accountHistory by observable<List<String>?>(null) { _, _, history ->
-        val entryCount = history?.size ?: 0
-
-        historyHeight = entryCount * (historyEntryHeight + dividerHeight)
-
+    var accountHistory by observable<String?>(null) { _, _, history ->
         if (history != null) {
+            historyHeight = historyEntryHeight + dividerHeight
             historyAdapter.accountHistory = history
+        } else {
+            historyHeight = 0
         }
     }
 
@@ -123,7 +122,7 @@ class AccountLogin : RelativeLayout {
         get() = input.onLogin
         set(value) { input.onLogin = value }
 
-    var onRemoveFromHistory: ((String) -> Unit)?
+    var onClearHistory: (() -> Unit)?
         get() = historyAdapter.onRemoveEntry
         set(value) { historyAdapter.onRemoveEntry = value }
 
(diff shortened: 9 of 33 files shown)


pinkisemils (changes requested):
> Reviewed 5 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 5 of 14 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[mullvad-daemon/src/lib.rs, line 1235 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McKFwUtDQm-3WdDdFLx:-McKFwUtDQm-3WdDdFLy:b2zn50l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7111937254c71bc0015419fbacae10bdb43775b2/mullvad-daemon/src/lib.rs#L1235)):*
> ```XML
>                             .notify_key_event(KeygenEvent::NewKey(public_key));
>                         if is_first_key {
>                             self.ensure_key_rotation().await;
> ```

What's the risk for always ensuring that the key is rotated?
___
*[mullvad-daemon/src/lib.rs, line 1573 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McKGyvcCYcY065nYYzg:-McKGyvd1KwZGkkqh4J1:b-cqircq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7111937254c71bc0015419fbacae10bdb43775b2/mullvad-daemon/src/lib.rs#L1573)):*
> ```XML
>             self.account_history
>                 .get()
>                 .map(|token| vec![token])
> ```

Should this RPC not be changed to only return the last account that was used?

pinkisemils (changes requested):
> Reviewed 7 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 12 of 14 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[mullvad-daemon/src/lib.rs, line 1573 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McKGyvcCYcY065nYYzg:-McKNkbI1jyd8RZV3Bw-:b-70ppir) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7111937254c71bc0015419fbacae10bdb43775b2/mullvad-daemon/src/lib.rs#L1573)):*

Should this RPC not be changed to only return the last account that was used?
</blockquote></details>

More precisely, it should be changed to return a single account rather than a list of accounts.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 13 of 14 files reviewed, 2 unresolved discussions (waiting on @dlon)

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 13 of 14 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[mullvad-daemon/src/lib.rs, line 583 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McKOCcFDMTUCYNdglQb:-McKOCcFDMTUCYNdglQc:b3p7nkb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7111937254c71bc0015419fbacae10bdb43775b2/mullvad-daemon/src/lib.rs#L583)):*
> ```XML
>         );
>         tokio::spawn(version_updater.run());
>         let mut account_history = account_history::AccountHistory::new(&cache_dir, &settings_dir)
> ```

The creation of account history and migration could be put into an async function that takes the directories and the settings struct as parameters.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 13 of 14 files reviewed, 3 unresolved discussions (waiting on @dlon)

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[mullvad-daemon/src/lib.rs, line 583 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McKOCcFDMTUCYNdglQb:-McNvsGZ9XiE8Sh-FVAQ:b-x6bf19) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7111937254c71bc0015419fbacae10bdb43775b2/mullvad-daemon/src/lib.rs#L583)):*

The creation of account history and migration could be put into an async function that takes the directories and the settings struct as parameters. 
</blockquote></details>

Do you think this belongs in `AccountHistory` or this file?
___
*[mullvad-daemon/src/lib.rs, line 1235 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McKFwUtDQm-3WdDdFLx:-McNucwB9N9Wk32Wci-I:be25nj2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7111937254c71bc0015419fbacae10bdb43775b2/mullvad-daemon/src/lib.rs#L1235)):*

What's the risk for always ensuring that the key is rotated?
</blockquote></details>

No real risk, I think. It just resets the rotation task unnecessarily.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[mullvad-daemon/src/lib.rs, line 583 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McKOCcFDMTUCYNdglQb:-McOt7Hz9rLVD1jtLxeJ:b-n2yuom) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7111937254c71bc0015419fbacae10bdb43775b2/mullvad-daemon/src/lib.rs#L583)):*

Do you think this belongs in `AccountHistory` or this file?
</blockquote></details>

I think `AccountHistory` is as good a place as any.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 10 of 31 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[mullvad-daemon/src/lib.rs, line 583 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McKOCcFDMTUCYNdglQb:-McP2D6E6FsqDrO2R34J:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7111937254c71bc0015419fbacae10bdb43775b2/mullvad-daemon/src/lib.rs#L583)):*

I think `AccountHistory` is as good a place as any.
</blockquote></details>

Done.
___
*[mullvad-daemon/src/lib.rs, line 1573 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McKGyvcCYcY065nYYzg:-McP2Dk30YpTGCxP0x89:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7111937254c71bc0015419fbacae10bdb43775b2/mullvad-daemon/src/lib.rs#L1573)):*

More precisely, it should be changed to return a single account rather than a list of accounts.
</blockquote></details>

Done.

pinkisemils (commented):
> Reviewed 14 of 21 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 24 of 31 files reviewed, 1 unresolved discussion (waiting on @dlon and @pinkisemils)

___
*[gui/src/renderer/components/Login.tsx, line 328 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McP8QXQ2iq50dpmqMa_:-McP8QXR0qwl5RJ60r79:b-1mfu01) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/11cccfc4b480148c16a8f6cd8e4ab8758e6e395b/gui/src/renderer/components/Login.tsx#L328)):*
> ```XML
>   const label = formatAccountToken(token);
>   return (
>     <>
> ```

@raksooo Is the surrounding  `<>` element really required here?

pinkisemils (commented):
> Reviewed 2 of 21 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 26 of 31 files reviewed, 1 unresolved discussion (waiting on @dlon)

raksooo (changes requested):
> Reviewed 6 of 21 files at r2, 7 of 18 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 23 of 34 files reviewed, 7 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[gui/src/main/index.ts, line 1166 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPVyXG2nJ9zf_QnJrg:-McPVyXG2nJ9zf_QnJrh:bsb5otb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/11cccfc4b480148c16a8f6cd8e4ab8758e6e395b/gui/src/main/index.ts#L1166)):*
> ```XML
>       const args = ['collect', '--output', reportPath];
>       if (toRedact) {
>         args.push('--redact', toRedact!);
> ```

`!` isn't needed since the if-statement makes sure that `toRedact` has a value.
___
*[gui/src/renderer/components/Login.tsx, line 328 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McP8QXQ2iq50dpmqMa_:-McPZh2WDzMXy7iG-XEt:bozusd5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/11cccfc4b480148c16a8f6cd8e4ab8758e6e395b/gui/src/renderer/components/Login.tsx#L328)):*

@raksooo Is the surrounding  `<>` element really required here?
</blockquote></details>

@pinkisemils That's correct. IIRC it was there before to make the typechecker happy, although it wasn't really needed. Since the contents isn't an array anymore this should work fine without `<>`. 
___
*[gui/src/renderer/components/Login.tsx, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPXt4J3PQIhs81TCuc:-McPXt4J3PQIhs81TCud:b-859qha) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2de8584a4362ecb5b7981706ae0d7cca764d0a79/gui/src/renderer/components/Login.tsx#L211)):*
> ```XML
> 
>   private shouldShowAccountHistory() {
>     return this.allowInteraction() && this.state.isActive && !!this.props.accountHistory;
> ```

I'm not a big fan of converting truthy/falsy values to a bool when there are multiple possible falsy values. In this situation it isn't very intuitive that this corresponds to `this.props.accountHistory !== undefined && this.props.accountHistory !== ""`. I can imagine that if someone refactors/rewrites this in some other way in the future they could miss the empty string case. Although I've probably done this many times myself and it results in less code so I'll let you decide.
___
*[gui/src/renderer/components/Login.tsx, line 291 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPaODs6pUJt-RtXhl-:-McPaODs6pUJt-RtXhl0:b3knkm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2de8584a4362ecb5b7981706ae0d7cca764d0a79/gui/src/renderer/components/Login.tsx#L291)):*
> ```XML
>             <StyledAccountDropdownContainer>
>               <AccountDropdown
>                 item={this.props.accountHistory!}
> ```

I try to avoid the non-null assertion operator. Instead of making this assumption here I think it would make more sense to set the default value of `accountHistory` to the empty string in the redux reducer instead.

If you would replace the empty string with an optional/undefined, then the simplest way to handle this would probably be to just return `null` in `AccountDropdown` if `accountHistory === undefined`.
___
*[gui/src/renderer/components/Login.tsx, line 330 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McP_n4T0qLWgqj_TPqx:-McP_n4T0qLWgqj_TPqy:bwqia3d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2de8584a4362ecb5b7981706ae0d7cca764d0a79/gui/src/renderer/components/Login.tsx#L330)):*
> ```XML
>     <>
>       <AccountDropdownItem
>         key={token}
> ```

`key` isn't needed anymore either since it was used to help react keep track of which array element is which before.
___
*[gui/src/renderer/containers/SupportPage.tsx, line 13 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPcc9j5QYNC2ES4AIg:-McPcc9j5QYNC2ES4AIh:b-j6181f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2de8584a4362ecb5b7981706ae0d7cca764d0a79/gui/src/renderer/containers/SupportPage.tsx#L13)):*
> ```XML
>   defaultEmail: state.support.email,
>   defaultMessage: state.support.message,
>   account: state.account.accountToken,
> ```

Why not use `accountHistory` here? This would only work while logged in.
___
*[mullvad-management-interface/proto/management_interface.proto, line 49 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPUaUH9BMYj48aBxtC:-McPUaUI3P_HdizWZCPU:b-z9ml2g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/11cccfc4b480148c16a8f6cd8e4ab8758e6e395b/mullvad-management-interface/proto/management_interface.proto#L49)):*
> ```protobuf
> 	rpc SetAccount(google.protobuf.StringValue) returns (google.protobuf.Empty) {}
> 	rpc GetAccountData(google.protobuf.StringValue) returns (AccountData) {}
> 	rpc GetAccountHistory(google.protobuf.Empty) returns (google.protobuf.StringValue) {}
> ```

I tried it out and noticed that if there is no history then this returns an empty string. Wouldn't it be better to make this optional?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 14 of 34 files reviewed, 4 unresolved discussions (waiting on @pinkisemils and @raksooo)

___
*[gui/src/main/index.ts, line 1166 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPVyXG2nJ9zf_QnJrg:-McTFXuA7qkTm_kj84pZ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/11cccfc4b480148c16a8f6cd8e4ab8758e6e395b/gui/src/main/index.ts#L1166)):*

`!` isn't needed since the if-statement makes sure that `toRedact` has a value.
</blockquote></details>

Done.
___
*[gui/src/renderer/components/Login.tsx, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPXt4J3PQIhs81TCuc:-McTEtPt9WMHES3l-QeX:bctpdcc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2de8584a4362ecb5b7981706ae0d7cca764d0a79/gui/src/renderer/components/Login.tsx#L211)):*

I'm not a big fan of converting truthy/falsy values to a bool when there are multiple possible falsy values. In this situation it isn't very intuitive that this corresponds to `this.props.accountHistory !== undefined && this.props.accountHistory !== ""`. I can imagine that if someone refactors/rewrites this in some other way in the future they could miss the empty string case. Although I've probably done this many times myself and it results in less code so I'll let you decide.
</blockquote></details>

Makes sense. Done.
___
*[gui/src/renderer/components/Login.tsx, line 291 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPaODs6pUJt-RtXhl-:-McTF2twC8LWa-W2kR8H:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2de8584a4362ecb5b7981706ae0d7cca764d0a79/gui/src/renderer/components/Login.tsx#L291)):*

I try to avoid the non-null assertion operator. Instead of making this assumption here I think it would make more sense to set the default value of `accountHistory` to the empty string in the redux reducer instead.

If you would replace the empty string with an optional/undefined, then the simplest way to handle this would probably be to just return `null` in `AccountDropdown` if `accountHistory === undefined`.
</blockquote></details>

Done.
___
*[mullvad-management-interface/proto/management_interface.proto, line 49 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPUaUH9BMYj48aBxtC:-McTFCDJ8DVDUj_9uYnD:bn570qa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/11cccfc4b480148c16a8f6cd8e4ab8758e6e395b/mullvad-management-interface/proto/management_interface.proto#L49)):*

I tried it out and noticed that if there is no history then this returns an empty string. Wouldn't it be better to make this optional?
</blockquote></details>

Yes. Fixed, though I made it return 404 instead.
___
*[gui/src/renderer/containers/SupportPage.tsx, line 13 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPcc9j5QYNC2ES4AIg:-McTF_1006XkHp6wbvBe:bye9n9u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2de8584a4362ecb5b7981706ae0d7cca764d0a79/gui/src/renderer/containers/SupportPage.tsx#L13)):*

Why not use `accountHistory` here? This would only work while logged in.
</blockquote></details>

Fixed! I didn't think about the fact that you can send problem reports when logged out. The reason was that the history was potentially empty. It was only updated when switching between accounts or logging out, but I've changed that.

raksooo (commented):
> Reviewed 6 of 10 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 20 of 34 files reviewed, 2 unresolved discussions (waiting on @pinkisemils and @raksooo)

___
*[gui/src/renderer/components/Login.tsx, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPXt4J3PQIhs81TCuc:-McTHubAAGpji-8CAxLu:b-1ah3mt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2de8584a4362ecb5b7981706ae0d7cca764d0a79/gui/src/renderer/components/Login.tsx#L211)):*

Makes sense. Done.
</blockquote></details>

Since you changed the daemon response to 404 and made `daemon-rpc.ts` return `undefined` instead of `""` you shouldn't need the the empty string check anymore :slightly_smiling_face:

raksooo (commented):
> Reviewed 1 of 10 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 21 of 34 files reviewed, 2 unresolved discussions (waiting on @dlon, @pinkisemils, and @raksooo)

___
*[gui/src/renderer/containers/SupportPage.tsx, line 13 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPcc9j5QYNC2ES4AIg:-McTJFQu8zFWoqVyp1aL:bwasv2h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2de8584a4362ecb5b7981706ae0d7cca764d0a79/gui/src/renderer/containers/SupportPage.tsx#L13)):*

Fixed! I didn't think about the fact that you can send problem reports when logged out. The reason was that the history was potentially empty. It was only updated when switching between accounts or logging out, but I've changed that.
</blockquote></details>

Now when thinking about it. Shouldn't `mullvad-problem-report` redact everything that looks like an account number? If the user switches accounts the old account won't be redacted. This could of course have happened before this change as well, but might be more common now.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 21 of 34 files reviewed, 1 unresolved discussion (waiting on @pinkisemils and @raksooo)

___
*[gui/src/renderer/containers/SupportPage.tsx, line 13 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPcc9j5QYNC2ES4AIg:-McTLMz0-M5-p6whX_JL:b-o34f9q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2de8584a4362ecb5b7981706ae0d7cca764d0a79/gui/src/renderer/containers/SupportPage.tsx#L13)):*

Now when thinking about it. Shouldn't `mullvad-problem-report` redact everything that looks like an account number? If the user switches accounts the old account won't be redacted. This could of course have happened before this change as well, but might be more common now.
</blockquote></details>

Asked on Slack on was informed that all 16 digit numbers are redacted.

pinkisemils (changes requested):
> Reviewed 8 of 18 files at r3, 1 of 10 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 30 of 34 files reviewed, 2 unresolved discussions (waiting on @dlon, @pinkisemils, and @raksooo)

___
*[mullvad-daemon/src/management_interface.rs, line 442 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McTOyF90-MeCmh7Gltv:-McTOyF90-MeCmh7Gltw:b-dtdvhl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0799c0cb3c1d0e37a233a0bf7bb3fcf9881c7b52/mullvad-daemon/src/management_interface.rs#L442)):*
> ```XML
>         match self.wait_for_result(rx).await? {
>             Some(history) => Ok(Response::new(history)),
>             None => Err(Status::not_found("history is empty")),
> ```

Does it have to be an error? Is an empty history invalid?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 29 of 34 files reviewed, 2 unresolved discussions (waiting on @pinkisemils and @raksooo)

___
*[gui/src/renderer/components/Login.tsx, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McPXt4J3PQIhs81TCuc:-McTQfLi4O_yA_Y5pESr:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2de8584a4362ecb5b7981706ae0d7cca764d0a79/gui/src/renderer/components/Login.tsx#L211)):*

Since you changed the daemon response to 404 and made `daemon-rpc.ts` return `undefined` instead of `""` you shouldn't need the the empty string check anymore :slightly_smiling_face: 
</blockquote></details>

Done.
___
*[mullvad-daemon/src/management_interface.rs, line 442 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McTOyF90-MeCmh7Gltv:-McTQiAb22Tu9r_6it_f:b-9jtqz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0799c0cb3c1d0e37a233a0bf7bb3fcf9881c7b52/mullvad-daemon/src/management_interface.rs#L442)):*

Does it have to be an error? Is an empty history invalid?
</blockquote></details>

It doesn't have to be, but I found it simpler than creating another message just to wrap a `StringValue`.

raksooo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-:-McTSS8NCGrXIoIFXoYc:bnfp4nl)* status: 30 of 34 files reviewed, 2 unresolved discussions (waiting on @pinkisemils and @raksooo)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 28 of 34 files reviewed, 1 unresolved discussion (waiting on @pinkisemils and @raksooo)

___
*[mullvad-daemon/src/management_interface.rs, line 442 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-McTOyF90-MeCmh7Gltv:-McTXdY7Cw4EYSNBbkqb:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0799c0cb3c1d0e37a233a0bf7bb3fcf9881c7b52/mullvad-daemon/src/management_interface.rs#L442)):*

It doesn't have to be, but I found it simpler than creating another message just to wrap a `StringValue`.
</blockquote></details>

Done.

raksooo (commented):
> Reviewed 2 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801)* status: 30 of 34 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

pinkisemils (approved):
:lgtm:

> Reviewed 2 of 18 files at r3, 3 of 10 files at r4, 1 of 1 files at r5, 3 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2801#-:-McTqESgCMn0SkEtLmZC:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2804, opened 2021-06-17
Improve out of time and new account screens

raksooo:
This PR improves the setup process by adding screens that walk users through the process of adding time and getting started. Some other changes were made to make this possible:
* Separate `Connect.tsx` into two views, one with the map and one for the out of time views
* Adding the result of `submitVoucher` to `account-data-cache.ts` instead of just storing it in the Redux state
* Save the previous expiry date when there's a new one (this is used to make it possible to show the amount of days that was added)
* Make it possible to remove all history before the current location to make it the root location
* Add callback parameter to functions in `history` that changes the location, these callbacks are then passed to listeners and in this case called when the animation is finished
* Made sure that `push` transitions won't be used if the navigation action was `POP` and vice-versa.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Line wrap the file at 100 chars.                                              Th
 
 ### Changed
 - Only use the account history file to store the last used account.
+- Updated the out of time-view and new account-view to make it more user friendly.
 
 ### Fixed
 - Fix link to download page not always using the beta URL when it should.
--- a/gui/src/config.json
+++ b/gui/src/config.json
@@ -4,6 +4,7 @@
     "purchase": "https://mullvad.net/account/",
     "manageKeys": "https://mullvad.net/account/ports/",
     "faq": "https://mullvad.net/help/tag/mullvad-app/",
+    "privacyGuide": "https://mullvad.net/help/first-steps-towards-online-privacy/",
     "download": "https://mullvad.net/download/",
     "betaDownload": "https://mullvad.net/download/beta"
   },
--- a/gui/src/main/account-data-cache.ts
+++ b/gui/src/main/account-data-cache.ts
@@ -1,5 +1,5 @@
 import { closeToExpiry, hasExpired } from '../shared/account-expiry';
-import { AccountToken, IAccountData } from '../shared/daemon-rpc-types';
+import { AccountToken, IAccountData, VoucherResponse } from '../shared/daemon-rpc-types';
 import { DateComponent, dateByAddingComponent } from '../shared/date-helper';
 import log from '../shared/logging';
 import consumePromise from '../shared/promise';
@@ -64,6 +64,12 @@ export default class AccountDataCache {
     this.notifyWatchers((watcher) => {
       watcher.onError(new Error('Cancelled'));
     });
+  }
+
+  public handleVoucherResponse(accountToken: AccountToken, voucherResponse: VoucherResponse) {
+    if (accountToken === this.currentAccount && voucherResponse.type === 'success') {
+      this.setValue({ expiry: voucherResponse.newExpiry });
+    }
   }
 
   private setValue(value: IAccountData) {
--- a/gui/src/main/index.ts
+++ b/gui/src/main/index.ts
@@ -1124,9 +1124,16 @@ class ApplicationMain {
     IpcMainEventChannel.account.handleLogin((token: AccountToken) => this.login(token));
     IpcMainEventChannel.account.handleLogout(() => this.logout());
     IpcMainEventChannel.account.handleGetWwwAuthToken(() => this.daemonRpc.getWwwAuthToken());
-    IpcMainEventChannel.account.handleSubmitVoucher((voucherCode: string) =>
-      this.daemonRpc.submitVoucher(voucherCode),
-    );
+    IpcMainEventChannel.account.handleSubmitVoucher(async (voucherCode: string) => {
+      const currentAccountToken = this.settings.accountToken;
+      const response = await this.daemonRpc.submitVoucher(voucherCode);
+
+      if (currentAccountToken) {
+        this.accountDataCache.handleVoucherResponse(currentAccountToken, response);
+      }
+
+      return response;
+    });
 
     IpcMainEventChannel.accountHistory.handleClear(async () => {
       await this.daemonRpc.clearAccountHistory();
--- a/gui/src/main/index.ts
+++ b/gui/src/main/index.ts
@@ -196,10 +196,16 @@ class ApplicationMain {
       return this.daemonRpc.getAccountData(accountToken);
     },
     (accountData) => {
-      this.accountData = accountData;
+      this.accountData = accountData && {
+        ...accountData,
+        previousExpiry:
+          accountData.expiry !== this.accountData?.expiry
+            ? this.accountData?.expiry
+            : this.accountData?.previousExpiry,
+      };
 
       if (this.windowController) {
-        IpcMainEventChannel.account.notify(this.windowController.webContents, accountData);
+        IpcMainEventChannel.account.notify(this.windowController.webContents, this.accountData);
       }
 
       this.handleAccountExpiry();
--- a/gui/src/renderer/app.tsx
+++ b/gui/src/renderer/app.tsx
@@ -145,7 +145,7 @@ export default class AppRenderer {
     });
 
     IpcRendererEventChannel.account.listen((newAccountData?: IAccountData) => {
-      this.setAccountExpiry(newAccountData && newAccountData.expiry);
+      this.setAccountExpiry(newAccountData?.expiry, newAccountData?.previousExpiry);
     });
 
     IpcRendererEventChannel.accountHistory.listen((newAccountHistory?: AccountToken) => {
@@ -219,7 +219,10 @@ export default class AppRenderer {
       initialState.translations.relayLocations,
     );
 
-    this.setAccountExpiry(initialState.accountData && initialState.accountData.expiry);
+    this.setAccountExpiry(
+      initialState.accountData?.expiry,
+      initialState.accountData?.previousExpiry,
+    );
     this.handleAccountChange(undefined, initialState.settings.accountToken);
     this.setAccountHistory(initialState.accountHistory);
     this.setSettings(initialState.settings);
@@ -829,8 +832,8 @@ export default class AppRenderer {
     this.reduxActions.settings.updateGuiSettings(guiSettings);
   }
 
-  private setAccountExpiry(expiry?: string) {
-    this.reduxActions.account.updateAccountExpiry(expiry);
+  private setAccountExpiry(expiry?: string, previousExpiry?: string) {
+    this.reduxActions.account.updateAccountExpiry(expiry, previousExpiry);
   }
 
   private storeAutoStart(autoStart: boolean) {
--- a/gui/src/renderer/app.tsx
+++ b/gui/src/renderer/app.tsx
@@ -620,7 +620,7 @@ export default class AppRenderer {
   private resetNavigation() {
     if (this.connectedToDaemon) {
       if (this.settings.accountToken) {
-        this.history.resetWithIfDifferent('/connect');
+        this.history.resetWithIfDifferent('/main');
       } else {
         this.history.resetWithIfDifferent('/login');
       }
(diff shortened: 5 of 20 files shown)


pronebird (commented):
> Reviewed 2 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804)* status: 2 of 22 files reviewed, all discussions resolved

pronebird (commented):
> Reviewed 3 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804)* status: 5 of 22 files reviewed, all discussions resolved

___
*[gui/src/main/index.ts, line 1137 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-Mci3lle6kRHy_Fg7aad:-Mci3lle6kRHy_Fg7aae:b-s2ru2x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26264bcf80083ef9649c46d7b5147dd6b5da96c8/gui/src/main/index.ts#L1137)):*
> ```XML
> 
>       if (response.type === 'success') {
>         this.accountDataCache.setValue({ expiry: response.newExpiry });
> ```

FYI this violates encapsulation of `AccountData`. 

I'd probably add a dedicated method to handle voucher responses, which doesn't fully protect us from ourselves, but still:

```
class AccountDataCache {

  func handleVoucherResponse(voucherAccount: String, response: VoucherResponse): Bool {
    if (response.type === 'success' && this.currentAccount === voucherAccount) {
      this.setValue({
        expiry: response.newExpiry
      });
      return true;
    } else {
      return false;
    }
  }

}
```

or even better, make `AccountDataCache` handle that for us:

```tax
class AccountDataCache {

  async func redeemVoucher(code: String): Promise<boolean> {
    let voucherAccount = this.currentAccount;
    let response = await this.myRedeemVoucherHandler(code);

    if (response.type === 'success' && this.currentAccount === voucherAccount) {
      this.setValue({
        expiry: response.newExpiry
      });
      return true;
    } else {
      return false;
    }
  }

}
```

Also I think at some point we should revise the name of that class. I think we can call it `AccountDataProxy` or something like that. But that's not important right now.

pronebird (commented):
> Reviewed 3 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804)* status: 8 of 22 files reviewed, all discussions resolved

pronebird (changes requested):
> Reviewed 14 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[gui/src/renderer/components/RedeemVoucher.tsx, line 84 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-McmiTBY7zA2Dfe2VC0k:-McmiTBZ8PdtZoqG7nyi:b7y0qsw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26264bcf80083ef9649c46d7b5147dd6b5da96c8/gui/src/renderer/components/RedeemVoucher.tsx#L84)):*
> ```XML
>     const submitDuration = Date.now() - submitTimestamp;
>     if (response.type !== 'success' && submitDuration < 500) {
>       await new Promise((resolve) => setTimeout(resolve, 500 - submitDuration));
> ```

FYI alternatively you can avoid showing the spinner unless the request takes longer than 500ms if you want to avoid flickering. This can easily be expressed with a React component. This way you avoid the artificial delays and never show the spinner to users if things happen very quickly.
___
*[gui/src/renderer/lib/history.ts, line 7 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-McmjYyD87WPmow3I0jv:-McmjYyECApBvJZbYR8Q:b7mq5t7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26264bcf80083ef9649c46d7b5147dd6b5da96c8/gui/src/renderer/lib/history.ts#L7)):*
> ```XML
>   action: Action,
>   entries: Location<S>[],
>   callback?: () => void,
> ```

Not quite getting what the callback is used for?

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804)* status: 12 of 22 files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[gui/src/main/index.ts, line 1137 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-Mci3lle6kRHy_Fg7aad:-Mcmt21f0zdA15M2mDom:b73nwxf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26264bcf80083ef9649c46d7b5147dd6b5da96c8/gui/src/main/index.ts#L1137)):*

FYI this violates encapsulation of `AccountData`. 

I'd probably add a dedicated method to handle voucher responses, which doesn't fully protect us from ourselves, but still:

```
class AccountDataCache {

  func handleVoucherResponse(voucherAccount: String, response: VoucherResponse): Bool {
    if (response.type === 'success' && this.currentAccount === voucherAccount) {
      this.setValue({
        expiry: response.newExpiry
      });
      return true;
    } else {
      return false;
    }
  }

}
```

or even better, make `AccountDataCache` handle that for us:

```tax
class AccountDataCache {

  async func redeemVoucher(code: String): Promise<boolean> {
    let voucherAccount = this.currentAccount;
    let response = await this.myRedeemVoucherHandler(code);

    if (response.type === 'success' && this.currentAccount === voucherAccount) {
      this.setValue({
        expiry: response.newExpiry
      });
      return true;
    } else {
      return false;
    }
  }

}
```

Also I think at some point we should revise the name of that class. I think we can call it `AccountDataProxy` or something like that. But that's not important right now.
</blockquote></details>

I prefer the first suggestion and implemented it.
___
*[gui/src/renderer/components/RedeemVoucher.tsx, line 84 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-McmiTBY7zA2Dfe2VC0k:-McmtATd7WJH9OdBRDov:but5l3u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26264bcf80083ef9649c46d7b5147dd6b5da96c8/gui/src/renderer/components/RedeemVoucher.tsx#L84)):*

FYI alternatively you can avoid showing the spinner unless the request takes longer than 500ms if you want to avoid flickering. This can easily be expressed with a React component. This way you avoid the artificial delays and never show the spinner to users if things happen very quickly.
</blockquote></details>

Nice idea, I'll bring it up with Matilda tomorrow when she's working since this was her idea.
___
*[gui/src/renderer/lib/history.ts, line 7 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-McmjYyD87WPmow3I0jv:-McmtGnH34pXVude_GdT:b-r0tpa5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26264bcf80083ef9649c46d7b5147dd6b5da96c8/gui/src/renderer/lib/history.ts#L7)):*

Not quite getting what the callback is used for?
</blockquote></details>

It was used to make it possible to reset the history **after** the navigation transition had finished. I've replaced this solution with a new `resetTo` method in `History`. The difference between `resetTo` and `resetWith` is that `resetTo` reports the current location (the one before navigation) as the previous one while `resetWith` reports `this.entries[1]` as the previous location.

I also replaced `entries` in the location listener event with `previousLocation` since that's what it was used for.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804)* status: 12 of 22 files reviewed, 1 unresolved discussion (waiting on @pronebird and @raksooo)

___
*[gui/src/renderer/lib/history.ts, line 7 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-McmjYyD87WPmow3I0jv:-McryM5QDrHohpZ76TNM:b-jagn1h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26264bcf80083ef9649c46d7b5147dd6b5da96c8/gui/src/renderer/lib/history.ts#L7)):*

It was used to make it possible to reset the history **after** the navigation transition had finished. I've replaced this solution with a new `resetTo` method in `History`. The difference between `resetTo` and `resetWith` is that `resetTo` reports the current location (the one before navigation) as the previous one while `resetWith` reports `this.entries[1]` as the previous location.

I also replaced `entries` in the location listener event with `previousLocation` since that's what it was used for.
</blockquote></details>

Transitions aren't part of history. History is just a stack, so maybe coordinating animations should be done elsewhere? Just my two cents but I realise it's really difficult with React. I used to experiment by adding events to `TransitionContainer` to catch when animations complete.

pronebird (approved):
> Reviewed 10 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[gui/src/renderer/lib/history.ts, line 74 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-Mcx2P5k3te7QCZEpqI5:-Mcx2P5k3te7QCZEpqI6:blcw1ye) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c89c6dff961672ad5b5f1864858abebbe333fd1/gui/src/renderer/lib/history.ts#L74)):*
> ```XML
>   // Resets the history and reports the second entry as the previous location.
>   public resetWith = (nextLocation: LocationDescriptor<S>, nextState?: S) => {
>     const previousLocation = this.entries[1];
> ```

Assumption that the history has more than > 1 element. Probably will return `undefined`, although previous location is likely `entries[0]`

raksooo (commented):
* ~~Make it possible to remove all history before the current location to make it the root location~~
* ~~Add callback parameter to functions in history that changes the location, these callbacks are then passed to listeners and in this case called when the animation is finished~~

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-:-MdGQYD5BturHhDq6PVl:b-sjgeb)* status: 18 of 22 files reviewed, 1 unresolved discussion (waiting on @pronebird and @raksooo)

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804)* status: 18 of 22 files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[gui/src/renderer/lib/history.ts, line 7 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-McmjYyD87WPmow3I0jv:-MdGrmbp4h3IWbeVrtsp:b-rs3g0o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26264bcf80083ef9649c46d7b5147dd6b5da96c8/gui/src/renderer/lib/history.ts#L7)):*

Transitions aren't part of history. History is just a stack, so maybe coordinating animations should be done elsewhere? Just my two cents but I realise it's really difficult with React. I used to experiment by adding events to `TransitionContainer` to catch when animations complete.
</blockquote></details>

Reverted as agreed on Slack. The current plan is to merge this without any changes to the transition, route or history systems. And then fix this in a follow-up PR. The only thing that isn't working is the transition between `/main/setup-finished` and `/main`, it just switches without any kind of transition.
___
*[gui/src/renderer/lib/history.ts, line 74 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-Mcx2P5k3te7QCZEpqI5:-MdGsE5n252izvTl1FFG:bp3aow6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c89c6dff961672ad5b5f1864858abebbe333fd1/gui/src/renderer/lib/history.ts#L74)):*

Assumption that the history has more than > 1 element. Probably will return `undefined`, although previous location is likely `entries[0]`
</blockquote></details>

Reverted.

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2804#-:-MdH5bbQ9pC6qdpLVB8i:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2809, opened 2021-06-21
Fix deadlock when entering the connecting state

dlon:
When an API request fails while the tunnel state machine is entering the connecting state, the API address cache tries to switch to a different address. It waits for the tunnel state machine to handle `TunnelCommand::AllowEndpoint`. At the same time, the tunnel state machine waits for the daemon to provide it with tunnel parameters (via `GenerateTunnelParameters`). This results in a deadlock.

To fix this, the REST request handler spawns a separate task for updating the address cache.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2809)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Line wrap the file at 100 chars.                                              Th
 
 ### Fixed
 - Fix link to download page not always using the beta URL when it should.
+- Fix deadlock that may occur when the API cannot be reached while entering the connecting state.
 
 #### Linux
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
This is a very good catch.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2809#-:-MciPF6F6efgcTBpq6UE:b-gl11wc)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2855, opened 2021-07-14
Resolve links and junctions for excluded apps on Windows

dlon:
Previously, reparse points were completely ignored (i.e., not resolved), since correctly handling them is somewhat complicated. If a symlink is only resolved initially, and is later replaced with a link pointing to some other target, then the old target path will be split until the configuration is manually updated again. So changes to the link must be monitored. This PR implements this feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ Line wrap the file at 100 chars.                                              Th
 
 ## [Unreleased]
 
+### Added
+#### Windows
+- Resolve symbolic links and junctions for excluded apps.
+
 ### Changed
 - Only use the account history file to store the last used account.
 - Update the out of time-view and new account-view to make it more user friendly.
--- a/talpid-core/src/split_tunnel/windows/mod.rs
+++ b/talpid-core/src/split_tunnel/windows/mod.rs
@@ -17,8 +17,9 @@ use std::{
     io, mem,
     net::{IpAddr, Ipv4Addr, Ipv6Addr},
     os::windows::io::{AsRawHandle, RawHandle},
+    path::Path,
     ptr,
-    sync::{mpsc as sync_mpsc, Arc, Weak},
+    sync::{mpsc as sync_mpsc, Arc, Mutex, Weak},
     time::Duration,
 };
 use talpid_types::{tunnel::ErrorStateCause, ErrorExt};
@@ -86,6 +87,10 @@ pub enum Error {
     /// The request handling thread is down
     #[error(display = "The ST request thread is down")]
     RequestThreadDown,
+
+    /// Failed to start the NTFS reparse point monitor
+    #[error(display = "Failed to start path monitor")]
+    StartPathMonitor(#[error(source)] io::Error),
 }
 
 /// Manages applications whose traffic to exclude from the tunnel.
@@ -289,6 +294,13 @@ impl SplitTunnel {
         let (tx, rx): (RequestTx, _) = sync_mpsc::sync_channel(3);
         let (init_tx, init_rx) = sync_mpsc::channel();
 
+        let no_paths: [&Path; 0] = [];
+        let (path_monitor, path_change_rx) =
+            path_monitor::PathMonitor::spawn(&no_paths).map_err(Error::StartPathMonitor)?;
+
+        let last_set_paths = Arc::new(Mutex::new(vec![]));
+        let last_set_paths_copy = last_set_paths.clone();
+
         std::thread::spawn(move || {
             let result = driver::DeviceHandle::new()
                 .map(Arc::new)
@@ -307,11 +319,25 @@ impl SplitTunnel {
             while let Ok((request, response_tx)) = rx.recv() {
                 let response = match request {
                     Request::SetPaths(paths) => {
-                        if paths.len() > 0 {
+                        let result = if paths.len() > 0 {
                             handle.set_config(&paths).map_err(Error::SetConfiguration)
                         } else {
                             handle.clear_config().map_err(Error::SetConfiguration)
+                        };
+
+                        if result.is_ok() {
+                            {
+                                *last_set_paths.lock().unwrap() = paths.to_vec();
+                            }
+                            if let Err(error) = path_monitor.set_paths(&paths) {
+                                log::error!(
+                                    "{}",
+                                    error.display_chain_with_msg("Failed to update path monitor")
+                                );
+                            }
                         }
+
+                        result
                     }
                     Request::RegisterIps(
                         mut tunnel_ipv4,
@@ -332,13 +358,40 @@ impl SplitTunnel {
                     log::error!("A response could not be sent for a completed request");
                 }
             }
+
+            if let Err(error) = path_monitor.shutdown() {
+                log::error!(
+                    "{}",
+                    error.display_chain_with_msg("Failed to shut down path monitor")
+                );
+            }
         });
 
         let handle = init_rx
             .recv_timeout(REQUEST_TIMEOUT)
             .map_err(|_| Error::RequestThreadStuck)??;
 
-        Ok((tx, handle))
+        let handle_copy = handle.clone();
+
+        std::thread::spawn(move || {
+            while let Ok(()) = path_change_rx.recv() {
+                log::debug!("Re-resolving excluded paths");
+                let paths = last_set_paths_copy.lock().unwrap();
+                let result = if paths.len() > 0 {
+                    handle_copy.set_config(&*paths)
+                } else {
+                    handle_copy.clear_config()
+                };
+                if let Err(error) = result {
+                    log::error!(
+                        "{}",
+                        error.display_chain_with_msg("Failed to reset excluded paths")
+                    );
+                }
+            }
+        });
+
+        Ok((tx, handle.clone()))
     }
 
     fn send_request(&self, request: Request) -> Result<(), Error> {
--- a/talpid-core/src/split_tunnel/windows/mod.rs
+++ b/talpid-core/src/split_tunnel/windows/mod.rs
@@ -1,4 +1,5 @@
 mod driver;
+mod path_monitor;
 mod windows;
 
 use crate::{
--- a/talpid-core/src/split_tunnel/windows/path_monitor.rs
+++ b/talpid-core/src/split_tunnel/windows/path_monitor.rs
@@ -251,7 +251,7 @@ impl DirContext {
         Ok(DirContext {
             path: path.as_ref().to_path_buf(),
             dir_handle,
-            buffer: vec![0u32; 512],
+            buffer: vec![0u32; 1024],
             overlapped: unsafe { std::mem::zeroed() },
         })
     }
@@ -501,11 +501,7 @@ impl PathMonitor {
                 }
 
                 if result.bytes_returned == 0 {
-                    let new_size = 2 * monitor.dir_contexts[result.completion_key].buffer.len();
-                    monitor.dir_contexts[result.completion_key]
-                        .buffer
-                        .resize(new_size, 0);
-                    log::debug!("Resized change event buffer as it was too small");
+                    log::debug!("Change event buffer is empty");
                 }
 
                 if let Err(error) =
--- a/talpid-core/src/split_tunnel/windows/path_monitor.rs
+++ b/talpid-core/src/split_tunnel/windows/path_monitor.rs
@@ -1,4 +1,5 @@
 use std::{
+    cmp,
     ffi::OsString,
     fs, io,
     os::windows::{
@@ -20,6 +21,7 @@ use winapi::{
             PostQueuedCompletionStatus,
         },
         minwinbase::OVERLAPPED,
+        stringapiset::CompareStringOrdinal,
         winbase::{
             ReadDirectoryChangesW, FILE_FLAG_BACKUP_SEMANTICS, FILE_FLAG_OPEN_REPARSE_POINT,
             FILE_FLAG_OVERLAPPED, INFINITE,
@@ -34,6 +36,8 @@ use winapi::{
 };
 
 const PATH_MONITOR_COMPLETION_KEY_IGNORE: usize = usize::MAX;
+
+const CSTR_EQUAL: i32 = 2;
 
 const ANYSIZE_ARRAY: usize = 1;
 const SYMLINK_FLAG_RELATIVE: u32 = 0x00000001;
@@ -532,9 +536,26 @@ impl PathMonitor {
                         if path.prefix != monitor.dir_contexts[result.completion_key].path() {
                             continue;
                         }
-                        if path.tail.starts_with(file_name) {
-                            changed = true;
-                            break;
+                        if path.tail.len() < file_name.len() {
+                            continue;
+                        }
+                        let cmp_status = unsafe {
+                            CompareStringOrdinal(
+                                path.tail.as_ptr(),
+                                cmp::min(file_name.len(), path.tail.len()) as i32,
+                                file_name.as_ptr(),
+                                file_name.len() as i32,
+                                1,
+                            )
+                        };
+                        match cmp_status {
+                            CSTR_EQUAL => {
+                                log::debug!("A path changed!");
+                                changed = true;
+                                break;
+                            }
+                            0 => log::error!("Bug: CompareStringOrdinal failed"),
+                            _ => (),
                         }
                     }
 
--- a/talpid-core/src/split_tunnel/windows/path_monitor.rs
+++ b/talpid-core/src/split_tunnel/windows/path_monitor.rs
@@ -495,6 +495,11 @@ impl PathMonitor {
                     }
                 };
 
+                if result.completion_key >= monitor.dir_contexts.len() {
+                    log::debug!("Ignoring out-of-bounds completion key");
+                    continue;
+                }
+
                 if result.bytes_returned == 0 {
                     let new_size = 2 * monitor.dir_contexts[result.completion_key].buffer.len();
                     monitor.dir_contexts[result.completion_key]
--- a/talpid-core/src/split_tunnel/windows/path_monitor.rs
+++ b/talpid-core/src/split_tunnel/windows/path_monitor.rs
@@ -1,0 +1,634 @@
+use std::{
+    ffi::OsString,
+    fs, io,
+    os::windows::{
+        ffi::{OsStrExt, OsStringExt},
+        fs::OpenOptionsExt,
+        io::{AsRawHandle, RawHandle},
+    },
+    path::{Path, PathBuf},
+    ptr,
+    sync::{mpsc as sync_mpsc, Arc},
+};
+use winapi::{
+    self,
+    um::{
+        fileapi::{GetFileAttributesW, GetFullPathNameW},
+        handleapi::{CloseHandle, INVALID_HANDLE_VALUE},
+        ioapiset::{
+            CreateIoCompletionPort, DeviceIoControl, GetQueuedCompletionStatus,
+            PostQueuedCompletionStatus,
+        },
+        minwinbase::OVERLAPPED,
+        winbase::{
+            ReadDirectoryChangesW, FILE_FLAG_BACKUP_SEMANTICS, FILE_FLAG_OPEN_REPARSE_POINT,
+            FILE_FLAG_OVERLAPPED, INFINITE,
+        },
+        winioctl::FSCTL_GET_REPARSE_POINT,
+        winnt::{
+            FILE_ATTRIBUTE_REPARSE_POINT, FILE_NOTIFY_CHANGE_DIR_NAME,
+            FILE_NOTIFY_CHANGE_FILE_NAME, FILE_NOTIFY_INFORMATION, HANDLE,
+            IO_REPARSE_TAG_MOUNT_POINT, IO_REPARSE_TAG_SYMLINK, MAXIMUM_REPARSE_DATA_BUFFER_SIZE,
+        },
+    },
+};
+
+const PATH_MONITOR_COMPLETION_KEY_IGNORE: usize = usize::MAX;
+
+const ANYSIZE_ARRAY: usize = 1;
+const SYMLINK_FLAG_RELATIVE: u32 = 0x00000001;
+
+
+// See https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/ca069dad-ed16-42aa-b057-b6b207f447cc.
+#[repr(C)]
+struct ReparseData {
+    tag: u32,
+    data_length: u16,
+    reserved: i16,
+    // Offset to a pathname pointing to the target path.
+    sub_name_offset: u16,
+    sub_name_length: u16,
+    // Offset to a user-displayable pathname.
+    print_name_offset: u16,
+    print_name_length: u16,
+    path_buffer: [u16; ANYSIZE_ARRAY],
+}
+
+// See https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/b41f1cbf-10df-4a47-98d4-1c52a833d913.
+#[repr(C)]
+struct ReparseDataSymlink {
+    tag: u32,
+    data_length: u16,
+    reserved: i16,
+    // Offset to a pathname pointing to the target path.
+    sub_name_offset: u16,
+    sub_name_length: u16,
+    // Offset to a user-displayable pathname.
+    print_name_offset: u16,
+    print_name_length: u16,
+    flags: u32,
+    path_buffer: [u16; ANYSIZE_ARRAY],
+}
+
+fn strip_namespace<P: AsRef<Path>>(path: P) -> PathBuf {
+    // \??: symlink to "DosDevices"
+    path.as_ref()
+        .strip_prefix(r"\\??")
+        .map(PathBuf::from)
+        .unwrap_or(path.as_ref().to_path_buf())
+}
+
+/// Returns the target of a reparse point as an absolute path.
+/// If `path` is not a link, `None` is returned.
+fn resolve_link<T: AsRef<Path> + Copy>(path: T) -> io::Result<Option<PathBuf>> {
+    let mut data_buffer = vec![0u8; MAXIMUM_REPARSE_DATA_BUFFER_SIZE as usize];
+
+    let mut stripped_path = strip_namespace(path);
+    if !stripped_path.starts_with(r"\\?\") {
+        stripped_path = Path::new(r"\\?\").join(stripped_path);
+    }
+
+    // Note: `file_attributes()` doesn't include all attributes, so we must use GetfileAttributesW.
+    let mut u16_path: Vec<u16> = stripped_path
+        .as_os_str()
+        .encode_wide()
+        .chain(std::iter::once(0u16))
+        .collect();
+    let attributes = unsafe { GetFileAttributesW(u16_path.as_mut_ptr()) };
+
+    if (attributes & FILE_ATTRIBUTE_REPARSE_POINT) == 0 {
+        return Ok(None);
+    }
+
+    let file = fs::OpenOptions::new()
+        .read(true)
+        .custom_flags(FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS)
+        .open(path)?;
+
+    let mut _bytes_returned = 0u32;
+
+    if unsafe {
+        DeviceIoControl(
+            file.as_raw_handle() as *mut _,
+            FSCTL_GET_REPARSE_POINT,
+            ptr::null_mut(),
+            0u32,
+            data_buffer.as_mut_ptr() as *mut _,
+            data_buffer.len() as u32,
+            &mut _bytes_returned,
+            ptr::null_mut(),
+        )
+    } == 0
+    {
+        return Err(io::Error::last_os_error());
+    }
+
+    let reparse_tag = unsafe { &*(data_buffer.as_mut_ptr() as *mut ReparseData) }.tag;
+    match reparse_tag {
+        IO_REPARSE_TAG_SYMLINK => {
+            let reparse_data = unsafe { &*(data_buffer.as_mut_ptr() as *mut ReparseDataSymlink) };
+            let parsed_path = unsafe {
+                std::slice::from_raw_parts(
+                    ((&reparse_data.path_buffer) as *const u16).offset(
+                        reparse_data.sub_name_offset as isize / std::mem::size_of::<u16>() as isize,
+                    ),
+                    reparse_data.sub_name_length as usize / std::mem::size_of::<u16>(),
+                )
+            };
+            let mut path_buf = PathBuf::from(OsString::from_wide(parsed_path));
+
+            if reparse_data.flags & SYMLINK_FLAG_RELATIVE != 0 {
+                if let Some(parent) = stripped_path.parent() {
+                    let path_buf_os: Vec<u16> = parent
+                        .join(path_buf)
+                        .into_os_string()
+                        .encode_wide()
+                        .chain(std::iter::once(0u16))
+                        .collect();
+
+                    let mut full_path_buffer = vec![0u16; 2048 / std::mem::size_of::<u16>()];
+
+                    let full_length = loop {
+                        let required_length = unsafe {
+                            GetFullPathNameW(
+                                path_buf_os.as_ptr(),
+                                full_path_buffer.len() as u32,
+                                full_path_buffer.as_mut_ptr(),
+                                ptr::null_mut(),
+                            )
+                        } as usize;
+
+                        if required_length == 0 {
+                            return Err(io::Error::last_os_error());
+                        }
+
+                        if required_length > full_path_buffer.len() {
+                            full_path_buffer.resize(required_length, 0);
+                        } else {
+                            break required_length;
+                        }
+                    };
+
+                    full_path_buffer.resize(full_length, 0);
+                    path_buf = PathBuf::from(OsString::from_wide(&full_path_buffer));
+                }
+            } else {
+                path_buf = strip_namespace(path_buf);
+            }
+
+            Ok(Some(path_buf))
+        }
+        IO_REPARSE_TAG_MOUNT_POINT => {
+            let reparse_data = unsafe { &*(data_buffer.as_mut_ptr() as *mut ReparseData) };
+            let parsed_path = unsafe {
+                std::slice::from_raw_parts(
+                    ((&reparse_data.path_buffer) as *const u16).offset(
+                        reparse_data.sub_name_offset as isize / std::mem::size_of::<u16>() as isize,
+                    ),
+                    reparse_data.sub_name_length as usize / std::mem::size_of::<u16>(),
+                )
+            };
+            Ok(Some(strip_namespace(PathBuf::from(OsString::from_wide(
+                parsed_path,
+            )))))
+        }
+        // unknown reparse tag
+        _ => Ok(None),
+    }
+}
+
+/// The same as [`resolve_all_links`] but for a set of paths.
+fn resolve_all_links_multiple<P: AsRef<Path>>(paths: &[P]) -> std::collections::HashSet<PathBuf> {
+    let mut monitored_paths = std::collections::HashSet::new();
+    for path in paths {
+        match resolve_all_links(path) {
+            Ok(paths) => monitored_paths.extend(paths),
+            Err(error) => {
+                log::error!("Failed to identify paths to monitor: {:?}", error);
+            }
+        }
+    }
+    monitored_paths
+}
+
+/// Returns all links and targets for a given path (including any of its parent directories).
+fn resolve_all_links<P: AsRef<Path>>(path: P) -> io::Result<Vec<PathBuf>> {
+    let mut monitor_paths = vec![path.as_ref().to_path_buf()];
+    let mut iter = path.as_ref().components();
+
+    let mut partial_path = PathBuf::new();
+    for _ in 0..2 {
+        partial_path.push(iter.next().ok_or(io::Error::new(
+            io::ErrorKind::Other,
+            "path must be absolute",
+        ))?);
+    }
+
+    for component in &mut iter {
+        partial_path.push(component);
+        if let Ok(Some(target)) = resolve_link(&partial_path) {
+            monitor_paths.extend(resolve_all_links(target.join(iter))?);
+            break;
+        }
+    }
+
+    Ok(monitor_paths)
+}
+
+struct DirContext {
+    path: PathBuf,
+    dir_handle: fs::File,
+    buffer: Vec<u32>,
+    overlapped: OVERLAPPED,
+}
+
+impl DirContext {
+    fn new<P: AsRef<Path>>(path: P) -> io::Result<DirContext> {
+        let dir_handle = fs::OpenOptions::new()
+            .read(true)
+            .custom_flags(FILE_FLAG_OVERLAPPED | FILE_FLAG_BACKUP_SEMANTICS)
+            .open(&path)?;
+        Ok(DirContext {
+            path: path.as_ref().to_path_buf(),
+            dir_handle,
+            buffer: vec![0u32; 512],
+            overlapped: unsafe { std::mem::zeroed() },
+        })
+    }
+
+    fn attach_to_io_port(
+        &self,
+        io_completion_port: &CompletionPort,
+        completion_key: usize,
+    ) -> io::Result<()> {
+        let handle = unsafe {
+            CreateIoCompletionPort(
+                self.dir_handle.as_raw_handle() as *mut _,
+                io_completion_port.as_raw_handle() as *mut _,
+                completion_key,
+                // num of threads is ignored here
+                0,
+            )
+        };
+
+        if handle == ptr::null_mut() {
+            Err(io::Error::last_os_error())
+        } else {
+            Ok(())
+        }
+    }
+
+    fn read_directory_changes(&mut self) -> io::Result<()> {
+        let mut _bytes_returned = 0;
+        if unsafe {
+            ReadDirectoryChangesW(
+                self.dir_handle.as_raw_handle() as *mut _,
+                self.buffer.as_mut_ptr() as *mut _,
+                (self.buffer.len() * std::mem::size_of::<u32>()) as u32,
+                1,
+                FILE_NOTIFY_CHANGE_FILE_NAME | FILE_NOTIFY_CHANGE_DIR_NAME,
+                &mut _bytes_returned,
+                &mut self.overlapped,
+                None,
+            )
+        } == 0
+        {
+            return Err(io::Error::last_os_error());
+        }
+        Ok(())
+    }
+
+    fn path(&self) -> &Path {
+        &self.path
+    }
+}
+
+unsafe impl Send for DirContext {}
+unsafe impl Sync for DirContext {}
+
+struct CompletionPort {
+    handle: HANDLE,
+}
+
+impl CompletionPort {
+    // `concurrent_threads`: 0 ==> number of processors
+    fn create(concurrent_threads: u32) -> io::Result<Self> {
+        let handle = unsafe {
+            CreateIoCompletionPort(INVALID_HANDLE_VALUE, ptr::null_mut(), 0, concurrent_threads)
+        };
+        if handle == ptr::null_mut() {
+            return Err(io::Error::last_os_error());
+        }
+        Ok(CompletionPort { handle })
+    }
+
+    fn get_queued_completion_status(&self) -> io::Result<CompletionStatus> {
+        let mut result = CompletionStatus {
+            bytes_returned: 0,
+            completion_key: 0,
+            used_overlapped: ptr::null_mut(),
+        };
+
+        if unsafe {
+            GetQueuedCompletionStatus(
+                self.handle,
+                &mut result.bytes_returned,
+                &mut result.completion_key,
+                &mut result.used_overlapped,
+                INFINITE,
+            )
+        } == 0
+        {
+            return Err(io::Error::last_os_error());
+        }
+
+        Ok(result)
+    }
+
+    fn post_queued_completion_status(
+        &self,
+        bytes_transferred: u32,
+        completion_key: usize,
+        overlapped: *mut OVERLAPPED,
+    ) -> io::Result<()> {
+        if unsafe {
+            PostQueuedCompletionStatus(self.handle, bytes_transferred, completion_key, overlapped)
+        } == 0
+        {
+            return Err(io::Error::last_os_error());
+        }
+        Ok(())
+    }
+}
+
+impl AsRawHandle for CompletionPort {
+    fn as_raw_handle(&self) -> RawHandle {
+        self.handle as *mut _
+    }
+}
+
+impl Drop for CompletionPort {
+    fn drop(&mut self) {
+        unsafe { CloseHandle(self.handle) };
+    }
+}
+
+unsafe impl Send for CompletionPort {}
+unsafe impl Sync for CompletionPort {}
+
+struct CompletionStatus {
+    bytes_returned: u32,
+    completion_key: usize,
+    used_overlapped: *mut OVERLAPPED,
+}
+
+#[derive(Clone, Hash, PartialEq, Eq)]
+struct StrippedPath {
+    prefix: PathBuf,
+    tail: Vec<u16>,
+}
+
+pub struct PathMonitorHandle {
+    port_handle: Arc<CompletionPort>,
+    tx: sync_mpsc::Sender<PathMonitorCommand>,
+}
+
+impl PathMonitorHandle {
+    pub fn set_paths<P: AsRef<Path>>(&self, paths: &[P]) -> io::Result<()> {
+        let _ = self.tx.send(PathMonitorCommand::SetPaths(
+            paths.iter().map(|p| p.as_ref().to_path_buf()).collect(),
+        ));
+        self.notify_monitor()
+    }
+
+    pub fn shutdown(&self) -> io::Result<()> {
+        let _ = self.tx.send(PathMonitorCommand::Shutdown);
+        self.notify_monitor()
+    }
+
+    fn notify_monitor(&self) -> io::Result<()> {
+        self.port_handle.post_queued_completion_status(
+            0,
+            PATH_MONITOR_COMPLETION_KEY_IGNORE,
+            ptr::null_mut(),
+        )
+    }
+}
+
+pub type PathChangeNotifyRx = sync_mpsc::Receiver<()>;
+
+enum PathMonitorCommand {
+    Shutdown,
+    SetPaths(Vec<PathBuf>),
+}
+
+pub struct PathMonitor {
+    port_handle: Arc<CompletionPort>,
+    dir_contexts: Vec<DirContext>,
+    stripped_paths: std::collections::HashSet<StrippedPath>,
+}
+
+impl PathMonitor {
+    pub fn spawn<P: AsRef<Path>>(
+        paths: &[P],
+    ) -> io::Result<(PathMonitorHandle, PathChangeNotifyRx)> {
+        let port_handle = Arc::new(CompletionPort::create(0)?);
+        let mut original_paths: Vec<PathBuf> =
+            paths.iter().map(|p| p.as_ref().to_path_buf()).collect();
+
+        let mut resolved_paths = resolve_all_links_multiple(&original_paths);
+        let stripped_paths = resolved_paths
+            .iter()
+            .filter_map(|p| Self::strip_path(p).ok())
+            .collect();
+
+        let mut monitor = Self {
+            port_handle: port_handle.clone(),
+            dir_contexts: vec![],
+            stripped_paths,
+        };
+
+        monitor.update_directory_contexts()?;
+
+        let (cmd_tx, cmd_rx) = sync_mpsc::channel();
+        let (notify_tx, notify_rx) = sync_mpsc::channel();
+
+        std::thread::spawn(move || {
+            loop {
+                let mut stop_monitor = false;
+                while let Some(cmd) = cmd_rx.try_iter().next() {
+                    match cmd {
+                        PathMonitorCommand::Shutdown => {
+                            stop_monitor = true;
+                            break;
+                        }
+                        PathMonitorCommand::SetPaths(new_paths) => {
+                            original_paths = new_paths;
+                            resolved_paths = resolve_all_links_multiple(&original_paths);
+                            monitor.stripped_paths = resolved_paths
+                                .iter()
+                                .filter_map(|p| Self::strip_path(p).ok())
+                                .collect();
+                            if let Err(error) = monitor.update_directory_contexts() {
+                                log::error!("Failed to set open new directory handles: {}", error);
+                                stop_monitor = true;
+                                break;
+                            }
+                        }
+                    }
+                }
+                if stop_monitor {
+                    break;
+                }
+
+                let result = match monitor.port_handle.get_queued_completion_status() {
+                    Ok(result) if result.completion_key == PATH_MONITOR_COMPLETION_KEY_IGNORE => {
+                        continue
+                    }
+                    Ok(result) => result,
+                    Err(error) => {
+                        log::error!(
+                            "GetQueuedCompletionStatus failed: {:?}",
+                            error.raw_os_error()
+                        );
+                        break;
+                    }
+                };
+
+                if result.bytes_returned == 0 {
+                    let new_size = 2 * monitor.dir_contexts[result.completion_key].buffer.len();
+                    monitor.dir_contexts[result.completion_key]
+                        .buffer
+                        .resize(new_size, 0);
+                    log::debug!("Resized change event buffer as it was too small");
+                }
+
+                if let Err(error) =
+                    monitor.dir_contexts[result.completion_key].read_directory_changes()
+                {
+                    log::error!("Failed to queue new directory change event: {}", error);
+                    break;
+                }
+
+                if result.bytes_returned == 0 || result.used_overlapped == ptr::null_mut() {
+                    continue;
+                }
+
+                let mut info = monitor.dir_contexts[result.completion_key].buffer.as_ptr()
+                    as *const FILE_NOTIFY_INFORMATION;
+                let mut changed = false;
+                loop {
+                    let current_field = unsafe { &*info };
+
+                    let file_name = unsafe {
+                        std::slice::from_raw_parts(
+                            current_field.FileName.as_ptr(),
+                            current_field.FileNameLength as usize / std::mem::size_of::<u16>(),
+                        )
+                    };
+
+                    for path in &monitor.stripped_paths {
+                        if path.prefix != monitor.dir_contexts[result.completion_key].path() {
+                            continue;
+                        }
+                        if path.tail.starts_with(file_name) {
+                            changed = true;
+                            break;
+                        }
+                    }
+
+                    if changed || current_field.NextEntryOffset == 0 {
+                        break;
+                    }
+                    info =
+                        unsafe { (info as *mut u8).offset(current_field.NextEntryOffset as isize) }
+                            as *const FILE_NOTIFY_INFORMATION;
+                }
+                if changed {
+                    let new_resolved_paths = resolve_all_links_multiple(&original_paths);
+                    if new_resolved_paths != resolved_paths {
+                        resolved_paths = new_resolved_paths;
+                        monitor.stripped_paths = resolved_paths
+                            .iter()
+                            .filter_map(|p| Self::strip_path(p).ok())
+                            .collect();
+                        if let Err(error) = monitor.update_directory_contexts() {
+                            log::error!("Failed to set open new directory handles: {}", error);
+                            break;
+                        }
+                        let _ = notify_tx.send(());
+                    }
+                }
+            }
+            log::debug!("Shutting down reparse point monitor");
+        });
+
+        Ok((
+            PathMonitorHandle {
+                port_handle,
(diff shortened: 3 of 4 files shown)


mvd-ows (changes requested):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 2 of 4 files reviewed, 12 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/split_tunnel/windows/mod.rs, line 297 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Meyi1Uo5x8KM5kuyP5J:-Meyi1UpE7gszvY-5amP:bmx57j0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/mod.rs#L297)):*
> ```XML
>         let (init_tx, init_rx) = sync_mpsc::channel();
> 
>         let no_paths: [&Path; 0] = [];
> ```

This variable isn't used for anything. Can you make `spawn` stop requiring it?
___
*[talpid-core/src/split_tunnel/windows/mod.rs, line 329 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MeyZj0FEGvxS9BEXuI6:-MeyZj0FEGvxS9BEXuI7:b-vo3ib2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/mod.rs#L329)):*
> ```XML
> 
>                         if result.is_ok() {
>                             {
> ```

Is there no `with(lockable as l) { ... }` construct?
___
*[talpid-core/src/split_tunnel/windows/mod.rs, line 333 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Mey_unt9yjlXkgfW1Nc:-Mey_unt9yjlXkgfW1Nd:b-sfel72) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/mod.rs#L333)):*
> ```XML
>                             }
>                             if let Err(error) = path_monitor.set_paths(&paths) {
>                                 log::error!(
> ```

It's unfortunate this isn't more visible to the user, e.g. presented using a notification.

Alas, as we discussed previously, the request could be coming from a short-lived CLI client.
___
*[talpid-core/src/split_tunnel/windows/mod.rs, line 380 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Meyn7-yArHGlJos7J8i:-Meyn7-yArHGlJos7J8j:bs9iyp4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/mod.rs#L380)):*
> ```XML
>                 log::debug!("Re-resolving excluded paths");
>                 let paths = last_set_paths_copy.lock().unwrap();
>                 let result = if paths.len() > 0 {
> ```

The request thread logic allows the path monitor to fail to register a new set of paths.

Therefore, the active configuration could be empty, and the path monitor could e.g. be using a previous configuration.

We have the following cases, which might be better handled if explicitly addressed:

* driver no config, monitor no config
* driver no config, monitor stale config
* driver has config, monitor no config
* driver has config, monitor stale config
* driver has config, monitor has config

Applying `clear_config` seems wrong. If anything, it should be attempting to clear the configuration in the monitor again.
___
*[talpid-core/src/split_tunnel/windows/mod.rs, line 388 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Meymo1DCC66qHSI3xIE:-Meymo1DCC66qHSI3xIF:bdlpjst) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/mod.rs#L388)):*
> ```
> reset
> ```

Is this the correct verb to use?
___
*[talpid-core/src/split_tunnel/windows/mod.rs, line 394 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Meyk6Ld1KXf9agLOWmc:-Meyk6Ld1KXf9agLOWmd:bzbyog2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/mod.rs#L394)):*
> ```XML
>         });
> 
>         Ok((tx, handle.clone()))
> ```

New thread is using `handle_copy` so cloning again is unnecessary?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 87 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Mf1znoh0uc4nqVBUyqJ:-Mf1znoi5UCfh9IBIYaG:bs2d1i0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L87)):*
> ```
> MAXIMUM_REPARSE_DATA_BUFFER_SIZE
> ```

This is defined as 16 Kb. You could probe the required size by calling `FSCTL_GET_REPARSE_POINT` with a null pointer for the output buffer.

But pre-allocating is probably better since the code is simpler.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Mf2HoGB2JZFy0mtmYoJ:-Mf2HoGB2JZFy0mtmYoK:b-w5qd52) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L136)):*
> ```XML
>                 std::slice::from_raw_parts(
>                     ((&reparse_data.path_buffer) as *const u16).offset(
>                         reparse_data.sub_name_offset as isize / std::mem::size_of::<u16>() as isize,
> ```

You should verify that offset + length stays inside the buffer. The data structure is literally read from the link file and could have invalid fields?

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 2 of 4 files reviewed, 13 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 144 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Mf8sUFW-ksnzqMYVwkk:-Mf8sUFY5izkjVQriGNS:b-bloy2v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L144)):*
> ```XML
> 
>             if reparse_data.flags & SYMLINK_FLAG_RELATIVE != 0 {
>                 if let Some(parent) = stripped_path.parent() {
> ```

This entire if-statement could be a function. Less indentation and shorter functions.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 2 of 4 files reviewed, 17 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 157 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_TnG-7FgIkVJjZ7qS:-MfDhKdp1OeMVdtO2Iz1:b4j4qf0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L157)):*

Are short path names anything to worry about here?
</blockquote></details>

What is this code trying to do? I think the loop and associated logic can be replaced by a single call to `GetLongPathNameW`, specifying the path you've just assembled.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 258 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_U3sdEvLS57NwDCd-:-MfE-qAe6jRnOztD_NZ2:b-hvnjq4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L258)):*

This buffer is pretty arbitrarily sized. Resizing it as necessary causes strange behavior, but this should probably be looked at. Currently, if the buffer is too small, then the result is simply discarded.
</blockquote></details>

The only thing I can think of, is you have a pending operation using the buffer already. If not, you should be able to resize it as much and as often as you want.

How come the item type isn't `u8`? I suppose because of the alignment requirements put forth by the API. But aren't all buffers guaranteed to be native-word aligned?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_UkMdCm40rjgW5mR3:-MfE6v-_Dz3AB6KZUZIQ:bengtv8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L439)):*

`NumberOfConcurrentThreads=0` allows many threads as there are processors to process the completion packets. Should this be lowered?
</blockquote></details>

That's fine. All code since the beginning of time has used `0`. You could even make it a default argument value, oh wait...
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 96 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfCzI954lZJ2HzQrCmM:-MfCzI954lZJ2HzQrCmN:b-kbqfwf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L96)):*
> ```XML
>     // Note: `file_attributes()` doesn't include all attributes, so we must use GetfileAttributesW.
>     let mut u16_path: Vec<u16> = stripped_path
>         .as_os_str()
> ```

If you make a function `pathbuf_into_widestr()` or somesuch, you won't have to repeat this sequence as many times.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 144 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Mf8sUFW-ksnzqMYVwkk:-MfDeHjp1Bg6TTsfGqYq:b3okzda) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L144)):*

This entire if-statement could be a function. Less indentation and shorter functions.
</blockquote></details>

How is the else-case (which is not handled) not an error?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 187 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfDjJ1T1TI8oEpVAEWM:-MfDjJ1UEVF1tE2ezUAo:b8en535) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L187)):*
> ```XML
>             let reparse_data = unsafe { &*(data_buffer.as_mut_ptr() as *mut ReparseData) };
>             let parsed_path = unsafe {
>                 std::slice::from_raw_parts(
> ```

Per above comment, verify length etc.

Also, you can generalize this and turn it into a function.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 223 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfDp6YnDc7LfCkafZi5:-MfDoI3N3L8tlf86J7hN:bn4vff9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L223)):*
> ```XML
> 
>     let mut partial_path = PathBuf::new();
>     for _ in 0..2 {
> ```

This is counting up from 0 to 2 minus 1 and discarding the results of all computations. Is it possible to define and apply a closure instead?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 291 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfE2ZvZ6JXuBITX5cm1:-MfE2ZvZ6JXuBITX5cm2:bo8av4n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L291)):*
> ```XML
>                 self.buffer.as_mut_ptr() as *mut _,
>                 (self.buffer.len() * std::mem::size_of::<u32>()) as u32,
>                 1,
> ```

Is there no `winapi::TRUE` or similar?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 382 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfE5VW13-q9cD-XbUds:-MfE5VW13-q9cD-XbUdt:b-k5t86g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L382)):*
> ```XML
> unsafe impl Sync for CompletionPort {}
> 
> struct CompletionStatus {
> ```

This is being used above. Define sooner?

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 2 of 4 files reviewed, 22 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 292 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfHxUTKE4INTUF2HYDk:-MfHxUTLARZqhEMYHl97:be8qvil) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L292)):*
> ```XML
>                 (self.buffer.len() * std::mem::size_of::<u32>()) as u32,
>                 1,
>                 FILE_NOTIFY_CHANGE_FILE_NAME | FILE_NOTIFY_CHANGE_DIR_NAME,
> ```

For completeness, I think `FILE_NOTIFY_CHANGE_ATTRIBUTES` needs be included as well. In the unlikely event that an existing file (leaf) is changed into being a link.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 374 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfHzEeYDJpkJzim3PlC:-MfHzEeYDJpkJzim3PlD:b-wx4y41) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L374)):*
> ```XML
> 
> impl Drop for CompletionPort {
>     fn drop(&mut self) {
> ```

Can we be sure that some other thread is not blocked inside `get_queued_completion_status` when this method is activated?

Is there a need to cancel any outstanding I/O requests before closing the handle?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 389 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfIdAkZ6ml8UqPADSY1:-MfIdAkZ6ml8UqPADSY2:buqxj7p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L389)):*
> ```XML
> 
> #[derive(Clone, Hash, PartialEq, Eq)]
> struct StrippedPath {
> ```

This type is not sufficiently documented. The `prefix` member seems to be volume + first node, and `tail` seems to be the remaining characters that make up a complete path.

You could add a ctor that operates on a `PathBuf`?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 407 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfIXXis4IcHydeZ4GgV:-MfIXXis4IcHydeZ4GgW:bxlr45j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L407)):*
> ```XML
>     }
> 
>     pub fn shutdown(&self) -> io::Result<()> {
> ```

This requests a shutdown but does not wait for processing to complete. Could this cause issues?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 423 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfIZ_tLANUwr5saa3gK:-MfIZ_tLANUwr5saa3gL:bpv68c6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L423)):*
> ```XML
> pub type PathChangeNotifyRx = sync_mpsc::Receiver<()>;
> 
> enum PathMonitorCommand {
> ```

Used above. Define sooner?

Why aren't types etc sorted by private, public?

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 2 of 4 files reviewed, 23 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 620 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_Vb0Q9QIBIp41VT3-:-MfPtLHI0pANXkhZQGH2:bvifx6q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L620)):*

Essentially, every entire drive is monitored. If `d` in `C:\a\b\c\d` is a symlink that is being excluded, we also want to detect if `C:\a` is removed (because we don't want to exclude the old link target anymore).

If only `C:\a\b\c` is monitored, `GetQueuedCompletionStatus` apparently fails with `ERROR_ACCESS_DENIED`. But I don't know that this is something that can be relied on.
</blockquote></details>

I don't understand the first comment. `prefix` is `C:\a`, right? When we monitor that path, does that include changes to the folder `a` itself or only changes made to its children? I don't see how the entire drive is monitored.

What you said first about including the topmost folder makes sense, so I don't think we should ever monitor only `C:\a\b\c`.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 604 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfPp8Qi9rodFnVBWCyl:-MfPp8Qi9rodFnVBWCym:bma14wi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L604)):*
> ```XML
>                 .any(|p| p.prefix == self.dir_contexts[i].path)
>             {
>                 self.dir_contexts.remove(i);
> ```

This looks suspicious. The deleted instance may have/is likely to have outstanding I/O registered using the item's index as a key. Beneath here, as you add a new instance of `DirContext`, it will be placed at the same index and the index value is again used as the key when queuing an I/O event.

Eventually, you'll have a dozen outstanding requests all with a key of e.g. `5`. When an event comes in it may be for the current `DirContext` item at index 5, or it may be for any old `DirContext` instance which you do not care about any longer, and which has already been removed from the collection.

I think you would be better off using unique keys, or preferably, cancelling outstanding requests as `DirContext` instances are being removed.

mvd-ows (commented):
> Reviewed 1 of 4 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 30 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 462 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfUrrcy-qR3zVhXhUrt:-MfUrrcy-qR3zVhXhUru:b-kd7y1w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L462)):*
> ```XML
>             loop {
>                 let mut stop_monitor = false;
>                 while let Some(cmd) = cmd_rx.try_iter().next() {
> ```

You could put this inside a function, something like `service_commands()`, and there won't be a super long thread body.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 476 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfUrmqp9bk5R_eW4_OL:-MfUrmqqCl2SNyGzeRqL:b-lihd00) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L476)):*
> ```
> set open
> ```

?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 507 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfUtXXSAbW210KEPrQ6:-MfUtXXSAbW210KEPrQ7:b-iqlm39) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L507)):*
> ```XML
> 
>                 if result.bytes_returned == 0 {
>                     log::debug!("Change event buffer is empty");
> ```

I was expecting to see `continue` here, but then realized we would be missing out on registering for the next event.

You need to register for another event only after completing all processing, see comment below. Try putting the event processing inside a function.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 511 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfUtftn-H23v2rnfnGn:-MfUtftn-H23v2rnfnGo:bnoq7pe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L511)):*
> ```XML
> 
>                 if let Err(error) =
>                     monitor.dir_contexts[result.completion_key].read_directory_changes()
> ```

Isn't this registering the same buffer/`OVERLAPPED` again to receive another event? If so, there's a race between the logic beneath here which is processing the current event, and the system that wants to write the next event into that same buffer.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 524 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfV-ydj75c_5vrJCrfm:-MfV-ydj75c_5vrJCrfn:b-i8fhpt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L524)):*
> ```XML
>                     as *const FILE_NOTIFY_INFORMATION;
>                 let mut changed = false;
>                 loop {
> ```

Why isn't this loop a separate function?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 542 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfV-W5G3nX-WtJR_n8_:-MfV-W5G3nX-WtJR_n8a:bq5wgn7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L542)):*
> ```XML
>                         }
>                         let cmp_status = unsafe {
>                             CompareStringOrdinal(
> ```

If `path.tail` is `b\c\d\e\f\g\h` and `file_name` is `b`... that's a match? I suppose that's correct, since `b` may have just been rewritten as a different link.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 567 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfV2aPk5oKOamGjuVNW:-MfV2aPk5oKOamGjuVNX:b-mp0eut) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L567)):*
> ```XML
>                             as *const FILE_NOTIFY_INFORMATION;
>                 }
>                 if changed {
> ```

Can't we know that one of the paths we're interested in have changed, based on the fact that we received an event? I suppose this has something to do with what you said about monitoring entire drives?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 604 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfPp8Qi9rodFnVBWCyl:-MfUnKfx7JghIy9YnOHF:bp3qyj7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L604)):*

This looks suspicious. The deleted instance may have/is likely to have outstanding I/O registered using the item's index as a key. Beneath here, as you add a new instance of `DirContext`, it will be placed at the same index and the index value is again used as the key when queuing an I/O event.

Eventually, you'll have a dozen outstanding requests all with a key of e.g. `5`. When an event comes in it may be for the current `DirContext` item at index 5, or it may be for any old `DirContext` instance which you do not care about any longer, and which has already been removed from the collection.

I think you would be better off using unique keys, or preferably, cancelling outstanding requests as `DirContext` instances are being removed.
</blockquote></details>

Right, any outstanding I/O needs to be cancelled before you delete this instance. Recall that the system owns the `OVERLAPPED` structure and the event buffer that you presented in a previous call to `ReadDirectoryChangesW`. If you don't properly cancel there will be writes into heap memory at some later time, that will overwrite arbitrary data.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 2 of 4 files reviewed, 24 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/mod.rs, line 380 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Meyn7-yArHGlJos7J8i:-MgjeGI368OXn1pFBmU1:b7xszpu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/mod.rs#L380)):*
Good points. Out of the cases mentioned, two are handled correctly:
* driver no config, monitor no config
* driver has config, monitor has config

Two cases can result in the active configuration being overwritten by a stale configuration if files change:

* driver no config, monitor stale config
* driver has config, monitor stale config

The remaining case will effectively disable monitoring (so it's relatively benign):

* driver has config, monitor no config

I'm leaning towards always disabling the monitor if this fails.

> Applying `clear_config` seems wrong. If anything, it should be attempting to clear the configuration in the monitor again.

We could probably just return in this case. This might happen if `last_set_paths` is cleared at the same time as this callback is invoked.
___
*[talpid-core/src/split_tunnel/windows/mod.rs, line 388 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Meymo1DCC66qHSI3xIE:-Mgf6N4TBtz0OJ7l_Jt2:b-cqw21p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/mod.rs#L388)):*

> ```
> reset
> ```

Is this the correct verb to use?
</blockquote></details>

Would `update` make more sense?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 157 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_TnG-7FgIkVJjZ7qS:-MgejiUg8vNOjf11xIvK:bwkpsxk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L157)):*

What is this code trying to do? I think the loop and associated logic can be replaced by a single call to `GetLongPathNameW`, specifying the path you've just assembled.
</blockquote></details>

It is trying to "expand" instances of "..", etc., but without resolving symbolic links. I'm pretty sure `GetLongPathNameW` on its own does not do the trick, though.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 620 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_Vb0Q9QIBIp41VT3-:-MgfP_HH8zwPvFbLLx5y:b-wh18i0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L620)):*

I don't understand the first comment. `prefix` is `C:\a`, right? When we monitor that path, does that include changes to the folder `a` itself or only changes made to its children? I don't see how the entire drive is monitored.

What you said first about including the topmost folder makes sense, so I don't think we should ever monitor only `C:\a\b\c`.
</blockquote></details>

No, `prefix` is the volume. Changes to `C:\a` are not detected if only `C:\a` is monitored.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Mf2HoGB2JZFy0mtmYoJ:-MgekCEfCNQUsDBMe6bv:b-z10uk8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L136)):*

You should verify that offset + length stays inside the buffer. The data structure is literally read from the link file and could have invalid fields?
</blockquote></details>

Very good point. Done.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 144 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Mf8sUFW-ksnzqMYVwkk:-MgeicJT7tDTVxvlk6Jb:b-ck84g3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L144)):*

How is the else-case (which is not handled) not an error?
</blockquote></details>

It is an error, but it should never happen since the parent exists. It probably wouldn't hurt to log it, though.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 291 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfE2ZvZ6JXuBITX5cm1:-Mgjjmh_6xwLlZSnJun6:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L291)):*

Is there no `winapi::TRUE` or similar?
</blockquote></details>

Done.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 407 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfIXXis4IcHydeZ4GgV:-MgfHoI46DsbG0CC1TIP:b-p1z2ht) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L407)):*

This requests a shutdown but does not wait for processing to complete. Could this cause issues?
</blockquote></details>

It should not cause any issues.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 542 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfV-W5G3nX-WtJR_n8_:-MgfO_4u5o4-pcSw_1c7:b-ma344i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L542)):*

If `path.tail` is `b\c\d\e\f\g\h` and `file_name` is `b`... that's a match? I suppose that's correct, since `b` may have just been rewritten as a different link.
</blockquote></details>

Yes. Or `b` might be deleted, for instance.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 567 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfV2aPk5oKOamGjuVNW:-Mgjk5Nx9uaalmmgXLG0:b338dwx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L567)):*

Can't we know that one of the paths we're interested in have changed, based on the fact that we received an event? I suppose this has something to do with what you said about monitoring entire drives?
</blockquote></details>

Correct. A lot of the notifications are irrelevant.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 2 of 4 files reviewed, 19 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/mod.rs, line 380 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Meyn7-yArHGlJos7J8i:-Mgl2jOh5gDaxTqEw1RF:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/mod.rs#L380)):*

Good points. Out of the cases mentioned, two are handled correctly:
* driver no config, monitor no config
* driver has config, monitor has config

Two cases can result in the active configuration being overwritten by a stale configuration if files change:

* driver no config, monitor stale config
* driver has config, monitor stale config

The remaining case will effectively disable monitoring (so it's relatively benign):

* driver has config, monitor no config

I'm leaning towards always disabling the monitor if this fails.

> Applying `clear_config` seems wrong. If anything, it should be attempting to clear the configuration in the monitor again.

We could probably just return in this case. This might happen if `last_set_paths` is cleared at the same time as this code is callback is invoked.
</blockquote></details>

Done.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 374 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfHzEeYDJpkJzim3PlC:-MgkQ_aWEKVXrQatKqpm:b-k9jrpl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L374)):*

Can we be sure that some other thread is not blocked inside `get_queued_completion_status` when this method is activated?

Is there a need to cancel any outstanding I/O requests before closing the handle?
</blockquote></details>

If some thread is still blocked inside `get_queued_completion_status`, then the instance cannot have been dropped. So I think that we can be sure.

I've made sure that a `CompletionPort` object can never be dropped before a `DirContext` that uses it, though.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 507 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfUtXXSAbW210KEPrQ6:-Mgl4SuADIvi3TgmFVVE:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L507)):*

I was expecting to see `continue` here, but then realized we would be missing out on registering for the next event.

You need to register for another event only after completing all processing, see comment below. Try putting the event processing inside a function.
</blockquote></details>

Done.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 604 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfPp8Qi9rodFnVBWCyl:-Mgkk3xSB3RBZFS2e3pe:b11ljid) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L604)):*

Right, any outstanding I/O needs to be cancelled before you delete this instance. Recall that the system owns the `OVERLAPPED` structure and the event buffer that you presented in a previous call to `ReadDirectoryChangesW`. If you don't properly cancel there will be writes into heap memory at some later time, that will overwrite arbitrary data.
</blockquote></details>

True. I've updated the code to cancel outstanding requests (see `DirContext::drop`).

mvd-ows (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 3 of 4 files reviewed, 11 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/mod.rs, line 388 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Meymo1DCC66qHSI3xIE:-MhJVQqoDYladWc8dm3H:bnmf9xf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/mod.rs#L388)):*

Would `update` make more sense?
</blockquote></details>

Probably, yes.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 157 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_TnG-7FgIkVJjZ7qS:-MhJYrQi34ff8qjrt57p:bywxd3v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L157)):*

It is trying to "expand" instances of "..", etc., but without resolving symbolic links. I'm pretty sure `GetLongPathNameW` on its own does not do the trick, though.
</blockquote></details>

Aha, OK. Did you verify to see if symlinks can contain embedded `..`? If they cannot, you would be able to reject the initial path even before trying to resolve anything?
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 620 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_Vb0Q9QIBIp41VT3-:-MhJb_FA9sEyTTxl4S32:b29gca1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L620)):*

No, `prefix` is the volume. Changes to `C:\a` are not detected if only `C:\a` is monitored.
</blockquote></details>

Aha, OK.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Mf2HoGB2JZFy0mtmYoJ:-MhJX3MvEwKym2vn1jTt:b-8zt0r9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L136)):*

Very good point. Done.
</blockquote></details>

I'd also want to verify that `DeviceIoControl` has actually updated enough bytes in the buffer before trying to interpret them as a structure.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 374 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfHzEeYDJpkJzim3PlC:-MhJ_jc4Bq5l3ws8m5VI:btojbfd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L374)):*

If some thread is still blocked inside `get_queued_completion_status`, then the instance cannot have been dropped. So I think that we can be sure.

I've made sure that a `CompletionPort` object can never be dropped before a `DirContext` that uses it, though.
</blockquote></details>

Great!
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 621 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhJcLvT2k7MvsxZafO1:-MhJcLvT2k7MvsxZafO2:b-efd3tt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/201f1f3167e988c2c6452b1960f0aadd21689a39/talpid-core/src/split_tunnel/windows/path_monitor.rs#L621)):*
> ```XML
>         };
> 
>         if result.completion_key >= self.dir_contexts.len() {
> ```

Isn't there a race condition here? You assert the container length and proceed, but there is no locking?

Some other actor could truncate/clear the container?

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 10 unresolved discussions (waiting on @dlon)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 10 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 157 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_TnG-7FgIkVJjZ7qS:-MhSfSxH1S8wkgx5_p8B:bit5tu8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L157)):*

Aha, OK. Did you verify to see if symlinks can contain embedded `..`? If they cannot, you would be able to reject the initial path even before trying to resolve anything?
</blockquote></details>

Yep, I checked. Relative links kind of have to. Otherwise you couldn't have a link to something in the parent directory, etc.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 621 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhJcLvT2k7MvsxZafO1:-MhS_WvP5M-fVWvu8-un:bjdjvkw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/201f1f3167e988c2c6452b1960f0aadd21689a39/talpid-core/src/split_tunnel/windows/path_monitor.rs#L621)):*

Isn't there a race condition here? You assert the container length and proceed, but there is no locking?

Some other actor could truncate/clear the container?
</blockquote></details>

The issue seems to be that an I/O packet could still be scheduled after the list has been updated (if the request is not aborted in time)? I don't think locking would help here. Maybe the suggestion you made elsewhere to use unique keys would make sense.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 11 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 327 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhSj-GJ2TR8HqjwNXqA:-MhSj-GJ2TR8HqjwNXqB:bvwdq4n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/201f1f3167e988c2c6452b1960f0aadd21689a39/talpid-core/src/split_tunnel/windows/path_monitor.rs#L327)):*
This looks a bit suspect. The `OVERLAPPED` object is dropped right away, but the [docs](https://docs.microsoft.com/en-us/windows/win32/fileio/cancelioex-func) state:

> The cancel operation for all pending I/O operations issued by the calling process for the specified file handle was successfully requested. The application must not free or reuse the OVERLAPPED structure associated with the canceled I/O operations until they have completed.

Could it simply wait using `GetOverlappedResult`? If the operation could not be cancelled in time, it also seems wrong to have dropped the `OVERLAPPED` struct.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 11 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 157 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_TnG-7FgIkVJjZ7qS:-MhSmx1559YuhIf29jqX:bnxwhlh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L157)):*

Yep, I checked. Relative links kind of have to. Otherwise you couldn't have a link to something in the parent directory, etc.
</blockquote></details>

Hmm OK. I suppose you'd want to handle short names as well then.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 11 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 327 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhSj-GJ2TR8HqjwNXqA:-MhSnMjOCv0uPFxvj3WF:brhmymg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/201f1f3167e988c2c6452b1960f0aadd21689a39/talpid-core/src/split_tunnel/windows/path_monitor.rs#L327)):*

This looks a bit suspect. The `OVERLAPPED` object is dropped right away, but the [docs](https://docs.microsoft.com/en-us/windows/win32/fileio/cancelioex-func) state:

> The cancel operation for all pending I/O operations issued by the calling process for the specified file handle was successfully requested. The application must not free or reuse the OVERLAPPED structure associated with the canceled I/O operations until they have completed.

Could it simply wait using `GetOverlappedResult`? If the operation cannot be cancelled in time, it also seems wrong to have dropped the `OVERLAPPED` struct.
</blockquote></details>

Using `GetOverlappedResult` sounds like a good idea. Ultimately you'll need to allow the code to get stuck waiting or to leak the object(s), which of these approaches is more reasonable?

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 11 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 621 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhJcLvT2k7MvsxZafO1:-MhSoivnCghHqrFuC4Z5:b-9uuhpm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/201f1f3167e988c2c6452b1960f0aadd21689a39/talpid-core/src/split_tunnel/windows/path_monitor.rs#L621)):*

The issue seems to be that an I/O packet could still be scheduled after the list has been updated (if the request is not aborted in time)? I don't think locking would help here. Maybe the suggestion you made elsewhere to use unique keys would make sense.
</blockquote></details>

Right, the request to update the set of paths being monitored is indirect, i.e. is posted to the same worker thread that receives and processes I/O completions. As such, the race I had in mind doesn't exist.

Regarding the issue you raise, I have no input on that at this time.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 3 of 4 files reviewed, 8 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 157 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_TnG-7FgIkVJjZ7qS:-MhU2rGg6vtUFewxidcd:b-4xi48) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L157)):*

Hmm OK. I suppose you'd want to handle short names as well then.
</blockquote></details>

Incidentally, I've moved this to its own function. I think it makes this function more readable.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 258 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Me_U3sdEvLS57NwDCd-:-MhU72hvCouesdr_O-za:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55ea95d04bd7aeaf4b7cf4a9b7426dce6871e44c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L258)):*

The only thing I can think of, is you have a pending operation using the buffer already. If not, you should be able to resize it as much and as often as you want.

How come the item type isn't `u8`? I suppose because of the alignment requirements put forth by the API. But aren't all buffers guaranteed to be native-word aligned?
</blockquote></details>

Done.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Mf2HoGB2JZFy0mtmYoJ:-MhU2_Fe3jBZxSucHdy0:ba54kxw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L136)):*

I'd also want to verify that `DeviceIoControl` has actually updated enough bytes in the buffer before trying to interpret them as a structure.
</blockquote></details>

How much validation is necessary? I've made sure that it writes at least `sizeof(ReparseDataMountPoint)` bytes now, but it would be more correct/strict to compare it to the specific type, I guess. Also, the buffer could be shrunk so that the pathname offsets may not refer to any data that it didn't update.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 187 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MfDjJ1T1TI8oEpVAEWM:-MhSgKr8-YKEoBlFRiQX:b8guxy8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L187)):*
> you can generalize this and turn it into a function.

Done. I (think I) had to define a macro for this in order to keep things simple, but it's probably an improvement regardless.

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 6 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Mf2HoGB2JZFy0mtmYoJ:-Mh_IbfF32UMXL2oYf5P:b9l2j00) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7a306f6a58860b4cf45cefc9ec26c8309d92dcca/talpid-core/src/split_tunnel/windows/path_monitor.rs#L136)):*

How much validation is necessary? I've made sure that it writes at least `sizeof(ReparseDataMountPoint)` bytes now, but it would be more correct/strict to compare it to the specific type, I guess. Also, the buffer could be shrunk so that the pathname offsets may not refer to any data that it didn't update.
</blockquote></details>

This is probably fine.

mvd-ows (approved):
> Reviewed 3 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 327 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhSj-GJ2TR8HqjwNXqA:-MhT0Gx15WNAwc5JXAYu:b-7uyr6t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/201f1f3167e988c2c6452b1960f0aadd21689a39/talpid-core/src/split_tunnel/windows/path_monitor.rs#L327)):*

Using `GetOverlappedResult` sounds like a good idea. Ultimately you'll need to allow the code to get stuck waiting or to leak the object(s), which of these approaches is more reasonable?
</blockquote></details>

I've updated it to keep the directory handle until `ERROR_OPERATION_ABORTED` is received. If it's never received, it leaks.

mvd-ows (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 327 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhSj-GJ2TR8HqjwNXqA:-MhwCKmi2ljTynOjjaHu:baav58o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/201f1f3167e988c2c6452b1960f0aadd21689a39/talpid-core/src/split_tunnel/windows/path_monitor.rs#L327)):*

I've updated it to keep the directory handle until `ERROR_OPERATION_ABORTED` is received. If it's never received, it leaks.
</blockquote></details>

It looks like any elements in member container `discarded_contexts` would be dropped when the owning `PathMonitor` instance is dropped? And that the current code merely delays the dropping of contexts.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 612 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhwDTN74rivs28AJVvt:-MhwDTN74rivs28AJVvu:b-djszt9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/29e644a1e4a04ce247033409681d8634436ef418/talpid-core/src/split_tunnel/windows/path_monitor.rs#L612)):*
> ```XML
>             }
>             Err(error) if error.raw_os_error() == Some(ERROR_OPERATION_ABORTED as i32) => {
>                 self.discarded_contexts.pop_front();
> ```

An I/O request associated with the port has been cancelled.

You assume the I/O request that was first (requested to be) cancelled will be the first to be signaled here.

I don't think it's safe to make that assumption.

It could just as well be the case that requests are checked in the order they were queued on the port.

mvd-ows (commented):
> Reviewed 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 612 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhwDTN74rivs28AJVvt:-MhwcUeo8Eju4dcvauLx:bthc6wj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/29e644a1e4a04ce247033409681d8634436ef418/talpid-core/src/split_tunnel/windows/path_monitor.rs#L612)):*

An I/O request associated with the port has been cancelled.

You assume the I/O request that was first (requested to be) cancelled will be the first to be signaled here.

I don't think it's safe to make that assumption.

It could just as well be the case that requests are checked in the order they were queued on the port.
</blockquote></details>

This can actually be done by checking `used_overlapped`. It should be enough to remove the matching element in `discarded_contexts`. When shutting down, all requests could be aborted, and then waited for using `GetQueuedCompletionStatus` until the list is empty or there's a timeout. Does that sound good?

mvd-ows (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 612 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhwDTN74rivs28AJVvt:-MhweyXlDT4ICQ5_ZXCD:b4ceia7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/29e644a1e4a04ce247033409681d8634436ef418/talpid-core/src/split_tunnel/windows/path_monitor.rs#L612)):*

This can actually be done by checking `used_overlapped`. It should be enough to remove the matching element in `discarded_contexts`. When shutting down, all requests could be aborted, and then waited for using `GetQueuedCompletionStatus` until the list is empty or there's a timeout. Does that sound good?
</blockquote></details>

Hmm, are you sure? `used_overlapped` seems to only be returned for I/O requests that succeed. Or maybe the code can be updated to always return it.

Sounds good if you can positively identify the request that was aborted.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 327 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhSj-GJ2TR8HqjwNXqA:-Mhy6w3rDtWExwk3Ofy_:b-hfqepg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/201f1f3167e988c2c6452b1960f0aadd21689a39/talpid-core/src/split_tunnel/windows/path_monitor.rs#L327)):*

It looks like any elements in member container `discarded_contexts` would be dropped when the owning `PathMonitor` instance is dropped? And that the current code merely delays the dropping of contexts.
</blockquote></details>

Done. It now waits for cancellation to complete before shutting down.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 612 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhwDTN74rivs28AJVvt:-Mhy649wA20gDhYH0hrI:b-98biwl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/29e644a1e4a04ce247033409681d8634436ef418/talpid-core/src/split_tunnel/windows/path_monitor.rs#L612)):*

Hmm, are you sure? `used_overlapped` seems to only be returned for I/O requests that succeed. Or maybe the code can be updated to always return it.

Sounds good if you can positively identify the request that was aborted.
</blockquote></details>

Done. `GetQueuedCompletionStatus` does seem to return `used_overlapped`, at least when the error is `ERROR_OPERATION_ABORTED`.
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 644 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-MhyAqfP4caazqa6H3PP:-MhyAqfP4caazqa6H3PQ:bcxgijb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/33008fad0666fae8dd01b5794271e8b8f1ad1e4c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L644)):*
> ```XML
>         };
> 
>         if self.free_discarded_context(result.used_overlapped) {
> ```

This handles the case where the cancel request wasn't submitted in time to cancel the operation. (I haven't seen this actually occur, but it is apparently [possible](https://docs.microsoft.com/en-us/windows/win32/fileio/cancelioex-func).)
___
*[talpid-core/src/split_tunnel/windows/path_monitor.rs, line 648 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2855#-Mhy7sYxDYpI7PmOn6hr:-Mhy7sYxDYpI7PmOn6hs:b-bdebqt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/33008fad0666fae8dd01b5794271e8b8f1ad1e4c/talpid-core/src/split_tunnel/windows/path_monitor.rs#L648)):*
> ```XML
>         }
> 
>         let ctx_index = self
> ```

This looks at `used_overlapped` to identify the right context object instead of the completion key. Completion keys as indices could probably become stale if some element in `dir_contexts` were removed.

PR #2856, opened 2021-07-15
Add WireGuard transport protocol constraint

dlon:
This previously existed only for custom relays. The PR adds the option for normal relays. Because the relay list does not currently include ports/TCP endpoints, all WireGuard relays are assumed to run `tcp2udp` on a set of hardcoded ports.

Usage:
> mullvad relay set tunnel wireguard any --protocol tcp

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Line wrap the file at 100 chars.                                              Th
 
 ### Added
 - Added possibility to filter locations by provider in the desktop app.
+- Add WireGuard over TCP CLI option for all relays.
 
 ### Changed
 - Only use the account history file to store the last used account.
--- a/gui/src/main/daemon-rpc.ts
+++ b/gui/src/main/daemon-rpc.ts
@@ -1154,8 +1154,12 @@ function convertFromOpenVpnConstraints(
 function convertFromWireguardConstraints(
   constraints: grpcTypes.WireguardConstraints,
 ): IWireguardConstraints {
-  const port = convertFromConstraint(constraints.getPort());
-  return { port };
+  const transportPort = convertFromConstraint(constraints.getPort());
+  if (transportPort !== 'any' && 'only' in transportPort) {
+    const port = convertFromConstraint(transportPort.only.getPort());
+    return { port };
+  }
+  return { port: 'any' };
 }
 
 function convertFromTunnelTypeConstraint(
@@ -1260,7 +1264,10 @@ function convertToWireguardConstraints(
     const wireguardConstraints = new grpcTypes.WireguardConstraints();
     const port = liftConstraint(constraint.port);
     if (port) {
-      wireguardConstraints.setPort(port);
+      const portConstraints = new grpcTypes.TransportPort();
+      portConstraints.setPort(port);
+      portConstraints.setProtocol(grpcTypes.TransportProtocol.UDP);
+      wireguardConstraints.setPort(portConstraints);
     }
     return wireguardConstraints;
   }
--- a/mullvad-cli/src/cmds/relay.rs
+++ b/mullvad-cli/src/cmds/relay.rs
@@ -167,6 +167,14 @@ impl Command for Relay {
                                             .required(true)
                                     )
                                     .arg(
+                                        clap::Arg::with_name("transport protocol")
+                                            .help("Transport protocol. If TCP is selected, traffic is \
+                                                   sent over TCP using a udp-over-tcp proxy")
+                                            .long("protocol")
+                                            .default_value("any")
+                                            .possible_values(&["any", "udp", "tcp"]),
+                                    )
+                                    .arg(
                                         clap::Arg::with_name("ip version")
                                             .long("ipv")
                                             .default_value("any")
@@ -528,6 +536,7 @@ impl Relay {
 
     async fn set_wireguard_constraints(&self, matches: &clap::ArgMatches<'_>) -> Result<()> {
         let port = parse_port_constraint(matches.value_of("port").unwrap())?;
+        let protocol = parse_protocol_constraint(matches.value_of("transport protocol").unwrap());
         let ip_version = parse_ip_version_constraint(matches.value_of("ip version").unwrap());
         let entry_location =
             parse_entry_location_constraint(matches.values_of("entry location").unwrap());
@@ -537,6 +546,11 @@ impl Relay {
                 NormalRelaySettingsUpdate {
                     wireguard_constraints: Some(WireguardConstraints {
                         port: port.unwrap_or(0) as u32,
+                        protocol: protocol
+                            .option()
+                            .map(|protocol| TransportProtocolConstraint {
+                                protocol: protocol as i32,
+                            }),
                         ip_version: ip_version.option().map(|protocol| IpVersionConstraint {
                             protocol: protocol as i32,
                         }),
--- a/mullvad-cli/src/cmds/relay.rs
+++ b/mullvad-cli/src/cmds/relay.rs
@@ -12,7 +12,7 @@ use mullvad_management_interface::types::{
     connection_config::{self, OpenvpnConfig, WireguardConfig},
     relay_settings, relay_settings_update, ConnectionConfig, CustomRelaySettings, IpVersion,
     IpVersionConstraint, NormalRelaySettingsUpdate, OpenvpnConstraints, ProviderUpdate,
-    RelayListCountry, RelayLocation, RelaySettingsUpdate, TransportProtocol,
+    RelayListCountry, RelayLocation, RelaySettingsUpdate, TransportPort, TransportProtocol,
     TransportProtocolConstraint, TunnelType, TunnelTypeConstraint, TunnelTypeUpdate,
     WireguardConstraints,
 };
@@ -164,15 +164,18 @@ impl Command for Relay {
                                     .arg(
                                         clap::Arg::with_name("port")
                                             .help("Port to use. Either 'any' or a specific port")
-                                            .required(true)
+                                            .long("port")
+                                            .takes_value(true)
+                                            .requires("transport protocol"),
                                     )
                                     .arg(
                                         clap::Arg::with_name("transport protocol")
                                             .help("Transport protocol. If TCP is selected, traffic is \
                                                    sent over TCP using a udp-over-tcp proxy")
                                             .long("protocol")
-                                            .default_value("any")
-                                            .possible_values(&["any", "udp", "tcp"]),
+                                            .takes_value(true)
+                                            .possible_values(&["udp", "tcp"])
+                                            .requires("port"),
                                     )
                                     .arg(
                                         clap::Arg::with_name("ip version")
@@ -535,8 +538,19 @@ impl Relay {
     }
 
     async fn set_wireguard_constraints(&self, matches: &clap::ArgMatches<'_>) -> Result<()> {
-        let port = parse_port_constraint(matches.value_of("port").unwrap())?;
-        let protocol = parse_protocol_constraint(matches.value_of("transport protocol").unwrap());
+        let protocol = matches.value_of("transport protocol").map(parse_protocol);
+        let port = match matches.value_of("port").map(parse_port_constraint) {
+            None => None,
+            Some(Err(error)) => return Err(error),
+            Some(Ok(Constraint::Any)) => Some(TransportPort {
+                protocol: protocol.unwrap() as i32,
+                port: 0,
+            }),
+            Some(Ok(Constraint::Only(port))) => Some(TransportPort {
+                protocol: protocol.unwrap() as i32,
+                port: u32::from(port),
+            }),
+        };
         let ip_version = parse_ip_version_constraint(matches.value_of("ip version").unwrap());
         let entry_location =
             parse_entry_location_constraint(matches.values_of("entry location").unwrap());
@@ -545,12 +559,7 @@ impl Relay {
             r#type: Some(relay_settings_update::Type::Normal(
                 NormalRelaySettingsUpdate {
                     wireguard_constraints: Some(WireguardConstraints {
-                        port: port.unwrap_or(0) as u32,
-                        protocol: protocol
-                            .option()
-                            .map(|protocol| TransportProtocolConstraint {
-                                protocol: protocol as i32,
-                            }),
+                        port,
                         ip_version: ip_version.option().map(|protocol| IpVersionConstraint {
                             protocol: protocol as i32,
                         }),
@@ -750,12 +759,12 @@ impl Relay {
         if let Some(constraints) = constraints {
             let mut out = format!(
                 "{} over {} over {}",
-                Self::format_port(constraints.port),
+                Self::format_port(constraints.port.as_ref().map(|port| port.port).unwrap_or(0)),
                 Self::format_transport_protocol(
                     constraints
-                        .protocol
+                        .port
                         .clone()
-                        .map(|protocol| TransportProtocol::from_i32(protocol.protocol).unwrap())
+                        .map(|port| TransportProtocol::from_i32(port.protocol).unwrap())
                 ),
                 Self::format_ip_version(
                     constraints
@@ -832,8 +841,14 @@ fn parse_port_constraint(raw_port: &str) -> Result<Constraint<u16>> {
 fn parse_protocol_constraint(raw_protocol: &str) -> Constraint<TransportProtocol> {
     match raw_protocol {
         "any" => Constraint::Any,
-        "udp" => Constraint::Only(TransportProtocol::Udp),
-        "tcp" => Constraint::Only(TransportProtocol::Tcp),
+        protocol => Constraint::Only(parse_protocol(protocol)),
+    }
+}
+
+fn parse_protocol(raw_protocol: &str) -> TransportProtocol {
+    match raw_protocol {
+        "udp" => TransportProtocol::Udp,
+        "tcp" => TransportProtocol::Tcp,
         _ => unreachable!(),
     }
 }
--- a/mullvad-cli/src/cmds/relay.rs
+++ b/mullvad-cli/src/cmds/relay.rs
@@ -2,7 +2,7 @@ use crate::{location, new_rpc_client, Command, Error, Result};
 use clap::{value_t, values_t};
 use itertools::Itertools;
 use std::{
-    fmt::Write,
+    convert::TryFrom,
     io::{self, BufRead},
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     str::FromStr,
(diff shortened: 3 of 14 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 0 of 5 files reviewed, 1 unresolved discussion

___
*[mullvad-daemon/src/relays.rs, line 897 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MeiLgkNDiKhfop2Woms:-MeiLgkO1ggIxwNUP9lX:b-slkxap) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9e53f8358d7e9cabf9ffde447ab0248de5b659ef/mullvad-daemon/src/relays.rs#L897)):*
> ```XML
>     ) -> Option<u16> {
>         // TODO: The relay list does not currently provide TCP ports.
>         let port_ranges = match constraints.protocol {
> ```

Maybe `WireguardEndpointData` should be modified right away instead.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[mullvad-types/src/relay_constraints.rs, line 522 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MezCq_D4NbYoAevNrlf:-MezCq_D4NbYoAevNrlg:bmxefz3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9e53f8358d7e9cabf9ffde447ab0248de5b659ef/mullvad-types/src/relay_constraints.rs#L522)):*
> ```XML
>                     .port_ranges
>                     .iter()
>                     .any(|range| (port >= range.0 && port <= range.1)),
> ```

So for a constraint with a specified port but transport protocol set to `Any` it will match if the port is in the allowed UDP ranges? :thinking: That's only correct under the assumption that transport protocol `Any` will deterministically select UDP for the actual connection at a later stage. Do we want to make those assumptions in `mullvad-types` really?

On top of that, it's unfortunate that the allowed TCP ports has to be defined in `mullvad-types`. It feels like our design forces more code than desired to land in the `-types` crates.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[mullvad-types/src/relay_constraints.rs, line 522 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MezCq_D4NbYoAevNrlf:-Mge_oHp6b_Mn4p98E4S:b-t6uvh4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9e53f8358d7e9cabf9ffde447ab0248de5b659ef/mullvad-types/src/relay_constraints.rs#L522)):*

So for a constraint with a specified port but transport protocol set to `Any` it will match if the port is in the allowed UDP ranges? :thinking: That's only correct under the assumption that transport protocol `Any` will deterministically select UDP for the actual connection at a later stage. Do we want to make those assumptions in `mullvad-types` really?

On top of that, it's unfortunate that the allowed TCP ports has to be defined in `mullvad-types`. It feels like our design forces more code than desired to land in the `-types` crates.
</blockquote></details>

Yes. The protocol and ports probably belong in `WireguardEndpointData`? I think the assumption was that this would eventually be included in the relay list instead of hardcoded, so it didn't matter much.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[mullvad-types/src/relay_constraints.rs, line 482 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-Mgf1pFl74cYEHbFsusA:-Mgf1pFl74cYEHbFsusB:bdacwzu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9e53f8358d7e9cabf9ffde447ab0248de5b659ef/mullvad-types/src/relay_constraints.rs#L482)):*
> ```XML
> pub struct WireguardConstraints {
>     pub port: Constraint<u16>,
>     pub protocol: Constraint<TransportProtocol>,
> ```

Can we consider going with a port+protocol representation that does not allow invalid combinations such as `protocol: Any, port: Only(443)`? As discussed on Slack.

Maybe we can consider

```
pub struct WireGuardConstraints {
    pub port: Constraint<TransportPort>,
}

pub struct TransportPort {
    pub protocol: TransportProtocol,
    pub port: Constraint<u16>,
}
```

And maybe later do the same for OpenVPN. But changing will require settings migration code. So if we can avoid this "mistake" on WireGuard from the start it will need less migration code in the long run.
___
*[mullvad-types/src/relay_constraints.rs, line 522 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MezCq_D4NbYoAevNrlf:-Mgf1fvcDohv0dbOxdg8:b-nj47ni) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9e53f8358d7e9cabf9ffde447ab0248de5b659ef/mullvad-types/src/relay_constraints.rs#L522)):*

Yes. The protocol and ports probably belong in `WireguardEndpointData`? I think the assumption was that this would eventually be included in the relay list instead of hardcoded, so it didn't matter much.
</blockquote></details>

We should not rely on the ports being added to the relay list in the near future. So it's better to make a design that's nice and clean with hardcoded ports, because we might be keeping it for a while.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[mullvad-types/src/relay_constraints.rs, line 482 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-Mgf1pFl74cYEHbFsusA:-MgyYd2v4iW7H20-3VRo:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9e53f8358d7e9cabf9ffde447ab0248de5b659ef/mullvad-types/src/relay_constraints.rs#L482)):*

Can we consider going with a port+protocol representation that does not allow invalid combinations such as `protocol: Any, port: Only(443)`? As discussed on Slack.

Maybe we can consider

```
pub struct WireGuardConstraints {
    pub port: Constraint<TransportPort>,
}

pub struct TransportPort {
    pub protocol: TransportProtocol,
    pub port: Constraint<u16>,
}
```

And maybe later do the same for OpenVPN. But changing will require settings migration code. So if we can avoid this "mistake" on WireGuard from the start it will need less migration code in the long run.
</blockquote></details>

Done.
___
*[mullvad-types/src/relay_constraints.rs, line 522 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MezCq_D4NbYoAevNrlf:-MgyYle48nvOOUvPRzdA:b-3g5qlm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9e53f8358d7e9cabf9ffde447ab0248de5b659ef/mullvad-types/src/relay_constraints.rs#L522)):*

We should not rely on the ports being added to the relay list in the near future. So it's better to make a design that's nice and clean with hardcoded ports, because we might be keeping it for a while.
</blockquote></details>

I'm not sure it's "clean", but the TCP ports are only defined in the relay selector now.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[mullvad-types/src/relay_constraints.rs, line 522 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MezCq_D4NbYoAevNrlf:-MhJI5zJAtvd9PkOcP3Z:bkgrqik) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9e53f8358d7e9cabf9ffde447ab0248de5b659ef/mullvad-types/src/relay_constraints.rs#L522)):*

I'm not sure it's "clean", but the TCP ports are only defined in the relay selector now.
</blockquote></details>

It doesn't look like this matching code takes `port.protocol` into consideration at all?

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 0 of 12 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[mullvad-daemon/src/relays.rs, line 917 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MhJMwlL9Y3cxSKDuROv:-MhJMwlL9Y3cxSKDuROw:bisgjzl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/36968b4111798ce1ce824eb5f718389c19d2d003/mullvad-daemon/src/relays.rs#L917)):*
> ```XML
>         constraints: &WireguardConstraints,
>     ) -> Option<u16> {
>         // TODO: The relay list does not currently provide TCP ports.
> ```

It's not clear that the relay list ever will provide TCP ports actually. Or it's not on the roadmap currently. So I'm not sure this is a `TODO` item because it's not something we know we will change.
___
*[mullvad-daemon/src/relays.rs, line 922 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MhJN2qI4xbwHnswlkYh:-MhJN2qI4xbwHnswlkYi:b-ze01kc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/36968b4111798ce1ce824eb5f718389c19d2d003/mullvad-daemon/src/relays.rs#L922)):*
> ```XML
>                 &WIREGUARD_TCP_PORTS[..]
>             }
>             _ => &data.port_ranges,
> ```

So for `Constraints::Any` this will return the UDP ports? That seems wrong since `Any` means automatic, meaning some fallback algorithm can toggle between UDP and TCP... But yeah, for now `Any` does indeed mean `UDP` so maybe we can cut some corners. But this will be changed rather soon, so it's important we don't mess stuff up too much and instead prepare for a version where TCP will be used by the fallback algorithm.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[mullvad-daemon/src/relays.rs, line 922 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MhJN2qI4xbwHnswlkYh:-MhNIpz622b50b9CJQIN:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/36968b4111798ce1ce824eb5f718389c19d2d003/mullvad-daemon/src/relays.rs#L922)):*

So for `Constraints::Any` this will return the UDP ports? That seems wrong since `Any` means automatic, meaning some fallback algorithm can toggle between UDP and TCP... But yeah, for now `Any` does indeed mean `UDP` so maybe we can cut some corners. But this will be changed rather soon, so it's important we don't mess stuff up too much and instead prepare for a version where TCP will be used by the fallback algorithm.
</blockquote></details>

Done.
___
*[mullvad-types/src/relay_constraints.rs, line 522 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MezCq_D4NbYoAevNrlf:-MhJUTzC75_iRMDL_mqn:b-5vpz91) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9e53f8358d7e9cabf9ffde447ab0248de5b659ef/mullvad-types/src/relay_constraints.rs#L522)):*

It doesn't look like this matching code takes `port.protocol` into consideration at all?
</blockquote></details>

It didn't, but I've improved the design now.

pinkisemils (changes requested):
> Reviewed 5 of 12 files at r2, 3 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 8 of 14 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-cli/src/cmds/relay.rs, line 762 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MhNlG5R9i_RD1ZuQZTr:-MhNlG5R9i_RD1ZuQZTs:b-wywmnr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/463a744d04770669b29daa380937cbde63ce7b03/mullvad-cli/src/cmds/relay.rs#L762)):*
> ```XML
>             let mut out = format!(
>                 "{} over {} over {}",
>                 Self::format_port(constraints.port.as_ref().map(|port| port.port).unwrap_or(0)),
> ```

Shouldn't the `unwrap_or(0)` zero be converted into a `any`?

In fact, it seems like you often need to access the port number from the constraint, which could be made into a method that returns an optional u16.

pinkisemils (commented):
> Reviewed 3 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 11 of 14 files reviewed, 4 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-types/src/relay_list.rs, line 146 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MhNn6J22VqJqTNfgBZy:-MhNn6J22VqJqTNfgBZz:bflbns6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/463a744d04770669b29daa380937cbde63ce7b03/mullvad-types/src/relay_list.rs#L146)):*
> ```XML
>     pub public_key: wireguard::PublicKey,
>     #[serde(default = "default_wg_transport")]
>     #[serde(skip)]
> ```

I'm not sure if there's a better way, but having a type that the fetched relay list deserializes to have these trap-door properties that have a default value seems a bit hacky. On the other hand, having two different types with the only difference between them being a single property is also insane.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 11 of 14 files reviewed, 4 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-cli/src/cmds/relay.rs, line 762 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MhNlG5R9i_RD1ZuQZTr:-MhNwMVeBFJknnXrF3Ur:b-anv2a6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/463a744d04770669b29daa380937cbde63ce7b03/mullvad-cli/src/cmds/relay.rs#L762)):*

Shouldn't the `unwrap_or(0)` zero be converted into a `any`?

In fact, it seems like you often need to access the port number from the constraint, which could be made into a method that returns an optional u16. 
</blockquote></details>

The `constraints.port` field is a `TransportPort` (the gRPC type). If it is not set, then the port is `any`. But if `constraints.port.port` is set to 0, then the port is also `any` (but the transport protocol is set). So `format_port` needs to check if it's 0 anyway.

Maybe the cleanest way to improve this would be to convert the gRPC type to back to the `mullvad-types`/`talpid-types` equivalent and then use its `Display` impl to format this? If it can be done without reworking all of the CLI.
___
*[mullvad-types/src/relay_list.rs, line 146 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MhNn6J22VqJqTNfgBZy:-MhNzDeGCGeMFPL2Qt7r:b-v4dl8e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/463a744d04770669b29daa380937cbde63ce7b03/mullvad-types/src/relay_list.rs#L146)):*

I'm not sure if there's a better way, but having a type that the fetched relay list deserializes to have these trap-door properties that have a default value seems a bit hacky. On the other hand, having two different types with the only difference between them being a single property is also insane.
</blockquote></details>

We could just also leave this field there in the file, even though the file itself will not contain the "TCP endpoints". I don't know if it should -- probably not? If it should, then the extra endpoints must be appended to the relay list at some point. I considered this but it seemed wrong to do it in `mullvad-rpc`. But it cannot simply be done in `mullvad_daemon::relays` before writing to the file, since the `mullvad_rpc::relay_list` bin must do it as well.

pinkisemils (changes requested):
> Reviewed 1 of 12 files at r2, 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 13 of 14 files reviewed, 4 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-cli/src/cmds/relay.rs, line 762 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MhNlG5R9i_RD1ZuQZTr:-MhOWchz4DXzUqV4Z2RB:b52ku1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/463a744d04770669b29daa380937cbde63ce7b03/mullvad-cli/src/cmds/relay.rs#L762)):*

The `constraints.port` field is a `TransportPort` (the gRPC type). If it is not set, then the port is `any`. But if `constraints.port.port` is set to 0, then the port is also `any` (but the transport protocol is set). So `format_port` needs to check if it's 0 anyway.

Maybe the cleanest way to improve this would be to convert the gRPC type to back to the `mullvad-types`/`talpid-types` equivalent and then use its `Display` impl to format this? If it can be done without reworking all of the CLI.
</blockquote></details>

I think that'd be the best thing to do, yes.
___
*[mullvad-types/src/relay_list.rs, line 146 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MhNn6J22VqJqTNfgBZy:-MhOWsIVDbjtH4G38tx_:b2ksftb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/463a744d04770669b29daa380937cbde63ce7b03/mullvad-types/src/relay_list.rs#L146)):*

We could just also leave this field there in the file, even though the file itself will not contain the "TCP endpoints". I don't know if it should -- probably not? If it should, then the extra endpoints must be appended to the relay list at some point. I considered this but it seemed wrong to do it in `mullvad-rpc`. But it cannot simply be done in `mullvad_daemon::relays` before writing to the file, since the `mullvad_rpc::relay_list` bin must do it as well.
</blockquote></details>

I guess this truly is the best solution to a weird problem.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: 11 of 14 files reviewed, 4 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[mullvad-cli/src/cmds/relay.rs, line 762 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856#-MhNlG5R9i_RD1ZuQZTr:-MhSUfsvAISEdlAz8WcN:bbrp1qr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/463a744d04770669b29daa380937cbde63ce7b03/mullvad-cli/src/cmds/relay.rs#L762)):*

I think that'd be the best thing to do, yes.
</blockquote></details>

Done. This should really be done for all types, but that work can be continued in a separate PR.

pinkisemils (approved):
> Reviewed 3 of 3 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2856)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

PR #2864, opened 2021-07-16
Ignore API IP rotations when the machine is offline

dlon:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2864)

### DIFF
--- a/mullvad-daemon/src/management_interface.rs
+++ b/mullvad-daemon/src/management_interface.rs
@@ -425,13 +425,7 @@ impl ManagementService for ManagementServiceImpl {
                     }),
                 })
             })
-            .map_err(|error: RestError| {
-                log::error!(
-                    "Unable to get account data from API: {}",
-                    error.display_chain()
-                );
-                map_rest_error(error)
-            })
+            .map_err(map_rest_error)
     }
 
     async fn get_account_history(&self, _: Request<()>) -> ServiceResult<types::AccountHistory> {
@@ -458,13 +452,7 @@ impl ManagementService for ManagementServiceImpl {
         let (tx, rx) = oneshot::channel();
         self.send_command_to_daemon(DaemonCommand::GetWwwAuthToken(tx))?;
         let result = self.wait_for_result(rx).await?;
-        result.map(Response::new).map_err(|error| {
-            log::error!(
-                "Unable to get account data from API: {}",
-                error.display_chain()
-            );
-            map_daemon_error(error)
-        })
+        result.map(Response::new).map_err(map_daemon_error)
     }
 
     async fn submit_voucher(
--- a/mullvad-rpc/src/address_cache.rs
+++ b/mullvad-rpc/src/address_cache.rs
@@ -122,7 +122,7 @@ impl AddressCache {
 
             tokio::task::block_in_place(move || {
                 if (*self.change_listener)(Self::get_address_inner(&transaction)).is_err() {
-                    log::error!("Failed to select a new API endpoint");
+                    log::debug!("Failed to select a new API endpoint");
                     return;
                 }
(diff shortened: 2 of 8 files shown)


pinkisemils (changes requested):
> Reviewed 3 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2864)* status: 3 of 8 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/tunnel_state_machine/connected_state.rs, line 200 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2864#-MfXEi-8AoFpb3I0ykGA:-MfXEi-A3UBCogUiRYh1:b-yrptc6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9aa39115946f4d00b1b03c0d5e7801b1272e58f/talpid-core/src/tunnel_state_machine/connected_state.rs#L200)):*
> ```XML
>             }
>             Some(TunnelCommand::AllowEndpoint(endpoint, tx)) => {
>                 if shared_values.is_offline {
> ```

I think that when the machine is offline, the TSM should never even receive the _allow endpoint_ command - the address cache shouldn't be rotating the address anyway. Otherwise, if the endpoint is rotated but a hole in the firewall isn't punched, requests will fail regardless.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2864)* status: 2 of 8 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/tunnel_state_machine/connected_state.rs, line 200 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2864#-MfXEi-8AoFpb3I0ykGA:-MgegNjL3Eef8PqSHEfF:bq1fwhx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9aa39115946f4d00b1b03c0d5e7801b1272e58f/talpid-core/src/tunnel_state_machine/connected_state.rs#L200)):*

I think that when the machine is offline, the TSM should never even receive the _allow endpoint_ command - the address cache shouldn't be rotating the address anyway. Otherwise, if the endpoint is rotated but a hole in the firewall isn't punched, requests will fail regardless.
</blockquote></details>

If the TSM rejects it, then the rotation doesn't actually occur. But I agree that this is more of a band-aid fix than a good one.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2864)* status: 2 of 8 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/tunnel_state_machine/connected_state.rs, line 200 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2864#-MfXEi-8AoFpb3I0ykGA:-MgexnmZCND5l09sk5vV:b-wsx8ov) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9aa39115946f4d00b1b03c0d5e7801b1272e58f/talpid-core/src/tunnel_state_machine/connected_state.rs#L200)):*

If the TSM rejects it, then the rotation doesn't actually occur. But I agree that this is more of a band-aid fix than a good one.
</blockquote></details>

I'd rework this to have the address cache check if it's online or offline. Maybe we need to multicast the offline state to both the TSM and the address cache? Maybe there should be another channel that allows the TSM to send messages to the daemon - this could be useful for the custom resolver to be able to report errors when it fails to bind to port 53 without stopping the whole TSM during startup.

dlon:
Superseded: #2937.

PR #2895, opened 2021-08-04
Always disable DNS over TLS in systemd-resolved

### LINKED ISSUE #2893: DNS doesn't work if DoT is enabled in systemd-resolved
## Issue report

Operating system: Fedora 34 (Workstation Edition)

App version: 2021.4

### Issue description

If DNS over TLS is enabled in systemd-resolved:

```
$ grep DNSOverTLS /etc/systemd/resolved.conf 
DNSOverTLS=yes
```

then after connecting, domains will not resolve. Looking at the output of resolvectl:

```
$ resolvectl
Global
           Protocols: LLMNR=resolve -mDNS +DNSOverTLS DNSSEC=yes/supported
    resolv.conf mode: stub
Fallback DNS Servers: 8.8.8.8 8.8.4.4

Link 2 (enp4s0)
Current Scopes: LLMNR/IPv4 LLMNR/IPv6
     Protocols: -DefaultRoute +LLMNR -mDNS +DNSOverTLS DNSSEC=yes/supported

Link 14 (wg-mullvad)
    Current Scopes: DNS
         Protocols: +DefaultRoute +LLMNR -mDNS +DNSOverTLS DNSSEC=yes/supported
Current DNS Server: fc00:bbbb:bbbb:bb01::1
       DNS Servers: 10.64.0.1 fc00:bbbb:bbbb:bb01::1
        DNS Domain: ~.
```

it looks like DNS over TLS is enabled for the wg-mullvad interface, and I guess Mullvad's DNS servers do not support that protocol. Manually disabling it with `resolvectl dnsovertls wg-mullvad no` helps until reconnect.

pinkisemils:
If a user has enabled DNS over TLS globally in `systemd-resolved` via editing `/etc/systemd/resolved.conf` and set `DNSOverTLS` to `yes`, then our daemon will fail to configure DNS for our tunnel interface, as systemd-resolved will always try to use DoT with our interface. To remedy this, I've added a call to `SetDNSOverTLS` to disable DoT for our interface.  This should fix #2893.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ Line wrap the file at 100 chars.                                              Th
 #### Linux
 - Make offline monitor aware of routing table changes.
 - Assign local DNS servers to more appropriate interfaces when using systemd-resolved.
+- Disable DNS over TLS for tunnel's DNS config when using systemd-resolved.
 
 #### Windows
 - Fix failure to restart the daemon when resuming from "fast startup" hibernation.
--- a/talpid-core/src/dns/linux/systemd_resolved.rs
+++ b/talpid-core/src/dns/linux/systemd_resolved.rs
@@ -75,6 +75,11 @@ impl SystemdResolved {
         self.tunnel_index = tunnel_index;
         let mut last_result = Ok(());
 
+        if let Err(error) = self.dbus_interface.disable_dot(self.tunnel_index).await {
+            log::error!("Failed to disable DoT: {}", error.display_chain());
+        }
+
+
         {
             let mut initial_states = self.initial_states.lock().unwrap();
             for (iface_index, iface_config) in &initial_config {
@@ -112,6 +117,7 @@ impl SystemdResolved {
                 }
             }
         }
+
 
         if let Err(error) = last_result {
             let _ = self.reset();
--- a/talpid-dbus/src/systemd_resolved.rs
+++ b/talpid-dbus/src/systemd_resolved.rs
@@ -83,6 +83,7 @@ const DNS_DOMAINS: &str = "Domains";
 const DNS_SERVERS: &str = "DNS";
 const GET_LINK_METHOD: &str = "GetLink";
 const SET_DNS_METHOD: &str = "SetDNS";
+const SET_DNS_OVER_TLS_METHOD: &str = "SetDNSOverTLS";
 const SET_DOMAINS_METHOD: &str = "SetDomains";
 const REVERT_METHOD: &str = "Revert";
 
@@ -350,6 +351,26 @@ impl SystemdResolved {
             .map_err(Error::DBusRpcError)
     }
 
+    fn link_disable_dns_over_tls<'a, 'b: 'a>(&'a self, interface_index: u32) -> Result<()> {
+        let link_object_path = self
+            .fetch_link(interface_index)
+            .map_err(|e| Error::GetLinkError(Box::new(e)))?;
+
+        let link_object = self.as_link_object(link_object_path.clone());
+
+        link_object.method_call(LINK_INTERFACE, SET_DNS_OVER_TLS_METHOD, ("no",))
+            .or_else(|error| {
+            if error.name() == Some("org.freedesktop.DBus.Error.UnknownMethod") {
+                log::debug!(
+                    "Didn't disable DNSOverTLS because systemd-resolved doesn't have 'SetDnsOverTLS' method. {}",
+                    error);
+                Ok(())
+            } else {
+                Err(error)
+            }
+        }).map_err(Error::DBusRpcError)
+    }
+
     fn get_link_dns_domains<'a, 'b: 'a>(
         &'a self,
         link_object_path: &'b dbus::Path<'static>,
@@ -530,6 +551,13 @@ impl AsyncHandle {
     pub async fn set_dns(&self, interface_index: u32, servers: Vec<IpAddr>) -> Result<DnsState> {
         let interface = self.dbus_interface.clone();
         tokio::task::spawn_blocking(move || interface.set_dns(interface_index, servers))
+            .await
+            .map_err(Error::AsyncTaskError)?
+    }
+
+    pub async fn disable_dot(&self, interface_index: u32) -> Result<()> {
+        let interface = self.dbus_interface.clone();
+        tokio::task::spawn_blocking(move || interface.link_disable_dns_over_tls(interface_index))
             .await
             .map_err(Error::AsyncTaskError)?
     }


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895)* status: all files reviewed, 3 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge1-Z_A4vdjrIXRQEk:-Mge1-Za8ZEkj_qqtjGh:bu0codq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*
> ```XML
>         let mut last_result = Ok(());
> 
>         if let Err(error) = self.dbus_interface.disable_dot(self.tunnel_index).await {
> ```

Does this actually override the global setting?
___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge2Iuj64dQ2iynqvwn:-Mge2Iuj64dQ2iynqvwo:b1bon8v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*
> ```XML
>         let mut last_result = Ok(());
> 
>         if let Err(error) = self.dbus_interface.disable_dot(self.tunnel_index).await {
> ```

Should this be configured on other interfaces as well, if custom DNS servers are set? This is somewhat uncommon, though.
___
*[talpid-dbus/src/systemd_resolved.rs, line 364 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge03fq4oG-ewaAwg9u:-Mge03fr7_i_mvqy6qpv:b-wy16pg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-dbus/src/systemd_resolved.rs#L364)):*
> ```XML
>             .or_else(|error| {
>             if error.name() == Some("org.freedesktop.DBus.Error.UnknownMethod") {
>                 log::debug!(
> ```

This looks like an error to me, i.e. something that the caller should deal with. But it's not that important.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge1-Z_A4vdjrIXRQEk:-MgeL5KLDnvrlup_pgCw:b-hepf4t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*

Does this actually override the global setting?
</blockquote></details>

No, this only configures the specified interface to not use DoT.
___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge2Iuj64dQ2iynqvwn:-MgeL9dZ32aXpmVStS6M:b-q18xws) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*

Should this be configured on other interfaces as well, if custom DNS servers are set? This is somewhat uncommon, though.
</blockquote></details>

Maybe, but I'm not sure. Do you mean in the case where we set the custom config on all interfaces?
___
*[talpid-dbus/src/systemd_resolved.rs, line 364 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge03fq4oG-ewaAwg9u:-MgeLleV0LY5zKaGMOLz:b-x6w50h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-dbus/src/systemd_resolved.rs#L364)):*

This looks like an error to me, i.e. something that the caller should deal with. But it's not that important.
</blockquote></details>

This is so that on older distros, where this RPC is not available, the RPC is a no-op. In those cases, there's nothing we can do anyway.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge2Iuj64dQ2iynqvwn:-MgeYb285vQwJonApHVp:b90xseg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*

Maybe, but I'm not sure. Do you mean in the case where we set the custom config on all interfaces?
</blockquote></details>

I meant if addresses in "private" ranges are used, yes.
___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge1-Z_A4vdjrIXRQEk:-MgeZ-SD8YhpAPp1LNPq:b-7tbmkz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*
I was wondering whether the setting in `/etc/systemd/resolved.conf ` would cause any per-interface setting to be ignored. This comment from the linked issue seems to answer that question:
> Manually disabling it with `resolvectl dnsovertls wg-mullvad` no helps until reconnect.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge2Iuj64dQ2iynqvwn:-MgejEQY01CN1YSVbaQV:bwgky4h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*

I meant if addresses in "private" ranges are used, yes.
</blockquote></details>

In that case, the resolver should have a valid DoT config anyway, right? And in those cases, we don't force resolution to happen through our config, right?
___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge1-Z_A4vdjrIXRQEk:-MgejaY23d-O3H3dogJ7:b-15rk4o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*

I was wondering whether the setting in `/etc/systemd/resolved.conf ` would cause any per-interface setting to be ignored. This comment from the linked issue seems to answer that question:
> Manually disabling it with `resolvectl dnsovertls wg-mullvad` no helps until reconnect.
</blockquote></details>

I've tested this on Arch, Fedora 34 and Ubuntu 20.10, and it does make resolution work.
Specifically, I set `DNSOverTLS=yes` in `/etc/systemd/resolved.conf`. Without these changes, any resolution fails.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge1-Z_A4vdjrIXRQEk:-MgesHjv1hE-Sb8JE0Bj:b-8fx2u7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*

I've tested this on Arch, Fedora 34 and Ubuntu 20.10, and it does make resolution work.
Specifically, I set `DNSOverTLS=yes` in `/etc/systemd/resolved.conf`. Without these changes, any resolution fails.
</blockquote></details>

:+1: 
___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge2Iuj64dQ2iynqvwn:-MgesoY47FOyi4O1WQvs:beipou1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*

In that case, the resolver should have a valid DoT config anyway, right? And in those cases, we don't force resolution to happen through our config, right?
</blockquote></details>

Don't we sort of force it? Only those servers that are specified are supposed to be used for resolution, and the firewall drops other traffic on port 53. DoT might work anyway, but it feels a bit unexpected.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge2Iuj64dQ2iynqvwn:-MgewqCm7RSIkYCIArtc:b8xx518) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*

Don't we sort of force it? Only those servers that are specified are supposed to be used for resolution, and the firewall drops other traffic on port 53. DoT might work anyway, but it feels a bit unexpected.
</blockquote></details>

We don't drop DoT traffic explicitly, and if the hostname of the DoT server doesn't have an address specified in systemd-resolved's config, then it should at least use our config. 

I think us breaking DoT for users who have enabled it isn't the right thing to do here. Thus, I think we should only be enabling this when we're setting the `~.` search domain for the tunnel interface.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2895#-Mge2Iuj64dQ2iynqvwn:-MgeyBh0DIijBZCxdnEg:bcyp4fc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1439680adf18fcac6d3a3b72f8403b65688ee6b5/talpid-core/src/dns/linux/systemd_resolved.rs#L78)):*

We don't drop DoT traffic explicitly, and if the hostname of the DoT server doesn't have an address specified in systemd-resolved's config, then it should at least use our config. 

I think us breaking DoT for users who have enabled it isn't the right thing to do here. Thus, I think we should only be enabling this when we're setting the `~.` search domain for the tunnel interface.
</blockquote></details>

That's reasonable.

YaBoiBurner:
I know that it's probably a _little_ late to bring this up, but it might be better to set `DNSOverTLS=opportunistic` instead of `DNSOverTLS=no`.

When set to `opportunistic`, DNS over TLS gets turned off when a server doesn't support it, but stays on when a server does support it.

That should allow the default DNS servers to work fine by default, while not unnecessarily disabling DNS over TLS for custom DNS servers that support it.

PR #2906, opened 2021-08-17
Fix icon opacity

raksooo:
This PR removes the opacity values in SVG-files and corrects the CSS values instead. We previously set the icon opacity to `0.6` in a lot of SVGs and then added another `0.6` with CSS which resulted in a final opacity of `0.36`. This probably wasn't intentional and has now been changed to only be specified with CSS and is now specified to `0.4`.

Git checklist:


faern:
Can you provide screenshots of the difference? If it's noticeable.

raksooo:
The difference isn't very noticeable. The difference when hovering is a bit more noticeable though, but if we want it exactly the same we could fix that of course.

Before:
before

After:
after

pronebird (approved):
:lgtm: although means that we would have to account for that change on iOS and possibly android too when we re-import those assets.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2906#-:-MhNO_7u3oyq63f3UML2:b2pfaqp)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2919, opened 2021-08-23
Add `TransportPort` OpenVPN constraint

dlon:
This refactors the OpenVPN relay constraints to use the `TransportPort` type introduced in #2856.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)

### DIFF
--- a/gui/src/main/daemon-rpc.ts
+++ b/gui/src/main/daemon-rpc.ts
@@ -1137,18 +1137,21 @@ function convertFromKeygenEvent(data: grpcTypes.KeygenEvent): KeygenEvent {
 function convertFromOpenVpnConstraints(
   constraints: grpcTypes.OpenvpnConstraints,
 ): IOpenVpnConstraints {
-  const port = convertFromConstraint(constraints.getPort());
-  let protocol: Constraint<RelayProtocol> = 'any';
-  switch (constraints.getProtocol()?.getProtocol()) {
-    case grpcTypes.TransportProtocol.TCP:
-      protocol = { only: 'tcp' };
-      break;
-    case grpcTypes.TransportProtocol.UDP:
-      protocol = { only: 'udp' };
-      break;
+  const transportPort = convertFromConstraint(constraints.getPort());
+  if (transportPort !== 'any' && 'only' in transportPort) {
+    const port = convertFromConstraint(transportPort.only.getPort());
+    let protocol: Constraint<RelayProtocol> = 'any';
+    switch (transportPort.only.getProtocol()) {
+      case grpcTypes.TransportProtocol.TCP:
+        protocol = { only: 'tcp' };
+        break;
+      case grpcTypes.TransportProtocol.UDP:
+        protocol = { only: 'udp' };
+        break;
+    }
+    return { port, protocol };
   }
-
-  return { port, protocol };
+  return { port: 'any', protocol: 'any' };
 }
 
 function convertFromWireguardConstraints(
@@ -1241,15 +1244,15 @@ function convertToOpenVpnConstraints(
 ): grpcTypes.OpenvpnConstraints | undefined {
   const openvpnConstraints = new grpcTypes.OpenvpnConstraints();
   if (constraints) {
-    const port = liftConstraint(constraints.port);
-    if (port) {
-      openvpnConstraints.setPort(port);
-    }
     const protocol = liftConstraint(constraints.protocol);
     if (protocol) {
-      const transportConstraint = new grpcTypes.TransportProtocolConstraint();
-      transportConstraint.setProtocol(convertToTransportProtocol(protocol));
-      openvpnConstraints.setProtocol(transportConstraint);
+      const portConstraints = new grpcTypes.TransportPort();
+      const port = liftConstraint(constraints.port);
+      if (port) {
+        portConstraints.setPort(port);
+      }
+      portConstraints.setProtocol(convertToTransportProtocol(protocol));
+      openvpnConstraints.setPort(portConstraints);
     }
     return openvpnConstraints;
   }
--- a/mullvad-cli/src/cmds/relay.rs
+++ b/mullvad-cli/src/cmds/relay.rs
@@ -13,8 +13,7 @@ use mullvad_management_interface::types::{
     relay_settings, relay_settings_update, ConnectionConfig, CustomRelaySettings, IpVersion,
     IpVersionConstraint, NormalRelaySettingsUpdate, OpenvpnConstraints, ProviderUpdate,
     RelayListCountry, RelayLocation, RelaySettingsUpdate, TransportPort, TransportProtocol,
-    TransportProtocolConstraint, TunnelType, TunnelTypeConstraint, TunnelTypeUpdate,
-    WireguardConstraints,
+    TunnelType, TunnelTypeConstraint, TunnelTypeUpdate, WireguardConstraints,
 };
 use mullvad_types::relay_constraints::Constraint;
 use talpid_types::net::all_of_the_internet;
@@ -149,13 +148,15 @@ impl Command for Relay {
                                     .arg(
                                         clap::Arg::with_name("port")
                                             .help("Port to use. Either 'any' or a specific port")
-                                            .required(true)
+                                            .long("port")
+                                            .default_value("any"),
                                     )
                                     .arg(
                                         clap::Arg::with_name("transport protocol")
+                                            .help("Transport protocol")
                                             .long("protocol")
-                                            .default_value("any")
-                                            .possible_values(&["any", "udp", "tcp"]),
+                                            .possible_values(&["any", "udp", "tcp"])
+                                            .default_value("any"),
                                     )
                             )
                             .subcommand(
@@ -165,17 +166,15 @@ impl Command for Relay {
                                         clap::Arg::with_name("port")
                                             .help("Port to use. Either 'any' or a specific port")
                                             .long("port")
-                                            .takes_value(true)
-                                            .requires("transport protocol"),
+                                            .default_value("any"),
                                     )
                                     .arg(
                                         clap::Arg::with_name("transport protocol")
                                             .help("Transport protocol. If TCP is selected, traffic is \
                                                    sent over TCP using a udp-over-tcp proxy")
                                             .long("protocol")
-                                            .takes_value(true)
-                                            .possible_values(&["udp", "tcp"])
-                                            .requires("port"),
+                                            .possible_values(&["any", "udp", "tcp"])
+                                            .default_value("any"),
                                     )
                                     .arg(
                                         clap::Arg::with_name("ip version")
@@ -516,20 +515,11 @@ impl Relay {
     }
 
     async fn set_openvpn_constraints(&self, matches: &clap::ArgMatches<'_>) -> Result<()> {
-        let port = parse_port_constraint(matches.value_of("port").unwrap())?;
-        let protocol = parse_protocol_constraint(matches.value_of("transport protocol").unwrap());
-
+        let port = parse_transport_port(matches)?;
         self.update_constraints(RelaySettingsUpdate {
             r#type: Some(relay_settings_update::Type::Normal(
                 NormalRelaySettingsUpdate {
-                    openvpn_constraints: Some(OpenvpnConstraints {
-                        port: port.unwrap_or(0) as u32,
-                        protocol: protocol
-                            .option()
-                            .map(|protocol| TransportProtocolConstraint {
-                                protocol: protocol as i32,
-                            }),
-                    }),
+                    openvpn_constraints: Some(OpenvpnConstraints { port }),
                     ..Default::default()
                 },
             )),
@@ -538,19 +528,7 @@ impl Relay {
     }
 
     async fn set_wireguard_constraints(&self, matches: &clap::ArgMatches<'_>) -> Result<()> {
-        let protocol = matches.value_of("transport protocol").map(parse_protocol);
-        let port = match matches.value_of("port").map(parse_port_constraint) {
-            None => None,
-            Some(Err(error)) => return Err(error),
-            Some(Ok(Constraint::Any)) => Some(TransportPort {
-                protocol: protocol.unwrap() as i32,
-                port: 0,
-            }),
-            Some(Ok(Constraint::Only(port))) => Some(TransportPort {
-                protocol: protocol.unwrap() as i32,
-                port: u32::from(port),
-            }),
-        };
+        let port = parse_transport_port(matches)?;
         let ip_version = parse_ip_version_constraint(matches.value_of("ip version").unwrap());
         let entry_location =
             parse_entry_location_constraint(matches.values_of("entry location").unwrap());
@@ -722,26 +700,12 @@ impl Relay {
         }
     }
 
-    fn format_port(port: u32) -> String {
-        if port != 0 {
-            format!("port {}", port)
-        } else {
-            "any port".to_string()
-        }
-    }
-
     fn format_openvpn_constraints(constraints: Option<&OpenvpnConstraints>) -> String {
         if let Some(constraints) = constraints {
-            format!(
-                "{} over {}",
-                Self::format_port(constraints.port),
-                Self::format_transport_protocol(
-                    constraints
-                        .protocol
-                        .clone()
-                        .map(|protocol| TransportProtocol::from_i32(protocol.protocol).unwrap())
-                )
-            )
+            let ovpn_constraints =
+                mullvad_types::relay_constraints::OpenVpnConstraints::try_from(constraints)
+                    .unwrap();
+            format!("{}", ovpn_constraints)
         } else {
             "any port over any transport protocol".to_string()
         }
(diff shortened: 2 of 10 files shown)


faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[mullvad-cli/src/cmds/relay.rs, line 161 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrB8568KnLWLKB_Zpo:-MhrB857B1olCxo7J8xD:bx3n1sv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5da62facb40cf572551c7e19ace460149ce42972/mullvad-cli/src/cmds/relay.rs#L161)):*
> ```XML
>                                             .takes_value(true)
>                                             .possible_values(&["udp", "tcp"])
>                                             .requires("port"),
> ```

You should be able to set the transport protocol without setting the port. This prevents that. I see we have the same issue with WireGuard now.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[mullvad-types/src/settings/migrations/v2.rs, line 132 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrCXIrBUxgtTy1aL2Q:-MhrCXIrBUxgtTy1aL2R:b-1ckp03) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5da62facb40cf572551c7e19ace460149ce42972/mullvad-types/src/settings/migrations/v2.rs#L132)):*
> ```XML
>         "port": {
>           "only": {
>             "protocol": "udp",
> ```

Always having to change *all* old migration files on each new update feels strange. Would be better if they were fixed in time once they landed. There is also a lot of duplication of the same `NEW_SETTINGS` string in all files. I think. I don't see any difference at least.

But it makes sense to test old migrations with new format maybe. So could we dump the desired content of `NEW_SETTINGS` into a `new_settings_format.json` instead and have this code be:
```rust
pub const NEW_SETTINGS: &str = include_str!("new_settings_format.json");
```

?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[mullvad-cli/src/cmds/relay.rs, line 161 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrB8568KnLWLKB_Zpo:-MhrJNtb-FC4xlZA_BNH:b-1fkoe2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5da62facb40cf572551c7e19ace460149ce42972/mullvad-cli/src/cmds/relay.rs#L161)):*

You should be able to set the transport protocol without setting the port. This prevents that. I see we have the same issue with WireGuard now.
</blockquote></details>

The obvious thing if the port isn't set is to default to `any`, which seems unintuitive. But the current port constraint could also be fetched from the daemon. Is that what you had in mind?
___
*[mullvad-types/src/settings/migrations/v2.rs, line 132 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrCXIrBUxgtTy1aL2Q:-MhrLf7X9dY_JC8e8rrb:b-w1rho8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5da62facb40cf572551c7e19ace460149ce42972/mullvad-types/src/settings/migrations/v2.rs#L132)):*

Always having to change *all* old migration files on each new update feels strange. Would be better if they were fixed in time once they landed. There is also a lot of duplication of the same `NEW_SETTINGS` string in all files. I think. I don't see any difference at least.

But it makes sense to test old migrations with new format maybe. So could we dump the desired content of `NEW_SETTINGS` into a `new_settings_format.json` instead and have this code be:
```rust
pub const NEW_SETTINGS: &str = include_str!("new_settings_format.json");
```

?
</blockquote></details>

The `NEW_SETTINGS` strings aren't exactly identical, but I bet they could be.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[mullvad-types/src/settings/migrations/v2.rs, line 132 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrCXIrBUxgtTy1aL2Q:-MhrVN6oACI66c7cidAC:be0g472) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5da62facb40cf572551c7e19ace460149ce42972/mullvad-types/src/settings/migrations/v2.rs#L132)):*

They aren't exactly identical, but I bet they could be.
</blockquote></details>

On second thought, this wouldn't work that well. `v3`, for example, tests migration of custom DNS settings. There's no such option in version 1, so the new settings must differ.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[mullvad-cli/src/cmds/relay.rs, line 161 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrB8568KnLWLKB_Zpo:-MhrOJew5B-9ZdQFgoAA:bwvllt3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5da62facb40cf572551c7e19ace460149ce42972/mullvad-cli/src/cmds/relay.rs#L161)):*

The obvious thing if the port isn't set is to default to `any`, which seems unintuitive. But the current port constraint could be fetched from the daemon. Is that what you had in mind?
</blockquote></details>

This gets somewhat weird if you have set a custom relay. Then there is no way to get the "current" port or transport protocol. I suppose it could fail in that case.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-cli/src/cmds/relay.rs, line 161 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrB8568KnLWLKB_Zpo:-MhrqtU50TqIwxgsnTw4:b65nwry) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5da62facb40cf572551c7e19ace460149ce42972/mullvad-cli/src/cmds/relay.rs#L161)):*

This gets somewhat weird if you have set a custom relay. Then there is no way to get the "current" port or transport protocol. I suppose it could fail in that case.
</blockquote></details>

We can't use "current port" since that might be a UDP port and the transport protocol is changed to TCP. Then the port is not valid any longer anyway. Or at least does not make sense. Until we implement so not specifying a setting does not touch it the behavior here has to be that a non-specified port defaults to `any`. If I say `... wireguard --protocol tcp` I interpret that as "I want to use TCP, I don't care about the other specifics"

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[mullvad-cli/src/cmds/relay.rs, line 161 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrB8568KnLWLKB_Zpo:-Mhrsm_m6r8Lsm0GTFMO:bh7t6wr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5da62facb40cf572551c7e19ace460149ce42972/mullvad-cli/src/cmds/relay.rs#L161)):*

We can't use "current port" since that might be a UDP port and the transport protocol is changed to TCP. Then the port is not valid any longer anyway. Or at least does not make sense. Until we implement so not specifying a setting does not touch it the behavior here has to be that a non-specified port defaults to `any`. If I say `... wireguard --protocol tcp` I interpret that as "I want to use TCP, I don't care about the other specifics"
</blockquote></details>

That makes sense. Done.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-cli/src/cmds/relay.rs, line 161 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrtZkf1_Cm1DpHLfmn:-MhrtZkf1_Cm1DpHLfmo:b-bobs9r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b0bc49c0f3e9a6175650ae1d7f772e9b02ebf64/mullvad-cli/src/cmds/relay.rs#L161)):*
> ```XML
>                                             .long("protocol")
>                                             .takes_value(true)
>                                             .possible_values(&["udp", "tcp"]),
> ```

It's a bit strange that you can't explicitly set `any` but the only way to obtain it is to omit this argument completely :thinking: Can't we allow `any` and then error out later when parsing all the settings instead? There is already a bit too much hidden magic in the CLI. I would not call this intuitive.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[mullvad-cli/src/cmds/relay.rs, line 161 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrtZkf1_Cm1DpHLfmn:-MhrvjmRD9Cz76W4VRf7:b-mcxqnw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b0bc49c0f3e9a6175650ae1d7f772e9b02ebf64/mullvad-cli/src/cmds/relay.rs#L161)):*

It's a bit strange that you can't explicitly set `any` but the only way to obtain it is to omit this argument completely :thinking: Can't we allow `any` and then error out later when parsing all the settings instead? There is already a bit too much hidden magic in the CLI. I would not call this intuitive.
</blockquote></details>

Then, following the existing pattern, `transport protocol` should probably default to `any` as well? Or should it be refactored to keep unspecified constraints right away?

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-cli/src/cmds/relay.rs, line 161 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrtZkf1_Cm1DpHLfmn:-MhrwOlSBl8ePVZZad7F:b-xh47dk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b0bc49c0f3e9a6175650ae1d7f772e9b02ebf64/mullvad-cli/src/cmds/relay.rs#L161)):*

Then, following the existing pattern, `transport protocol` should probably default to `any` as well? Or should it be refactored to keep unspecified constraints right away?
</blockquote></details>

The part about keeping unspecified constraints is a separate issue IMO. Since it's bigger than just this settings refactor. And that change does not change the settings format afaik. So in this PR the transport protocol should probably default to `any`, yes. But if you either implicitly or explicitly try to run the equivalent of `--protocol any --port 1000` the CLI should error out saying that's an invalid combination of relay constraints/settings.

faern (commented):
> Reviewed 1 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 1 of 10 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[mullvad-types/src/settings/migrations/v2.rs, line 132 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrCXIrBUxgtTy1aL2Q:-Mhs-0cO7RRkcub9mbSg:b-6a6n1i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5da62facb40cf572551c7e19ace460149ce42972/mullvad-types/src/settings/migrations/v2.rs#L132)):*

On second thought, this wouldn't work that well. `v3`, for example, tests migration of custom DNS settings. There's no such option in version 1, so the new settings must differ.
</blockquote></details>

Ah yeah! Any migration converts from itself to latest. I just thought `v2` converted to `v3` etc. But no. We convert all the way from `v2 -> latest`. So then the migration of course needs to know the format of the latest settings.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 1 of 10 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[mullvad-cli/src/cmds/relay.rs, line 161 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrtZkf1_Cm1DpHLfmn:-Mhs025A0d9fxd2azkhg:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b0bc49c0f3e9a6175650ae1d7f772e9b02ebf64/mullvad-cli/src/cmds/relay.rs#L161)):*

The part about keeping unspecified constraints is a separate issue IMO. Since it's bigger than just this settings refactor. And that change does not change the settings format afaik. So in this PR the transport protocol should probably default to `any`, yes. But if you either implicitly or explicitly try to run the equivalent of `--protocol any --port 1000` the CLI should error out saying that's an invalid combination of relay constraints/settings.
</blockquote></details>

Done.

faern (changes requested):
> Reviewed 1 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 2 of 10 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-types/src/settings/migrations/v2.rs, line 132 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhrCXIrBUxgtTy1aL2Q:-Mhs4FsV4Ny4HTUXbPCf:bumv4x3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5da62facb40cf572551c7e19ace460149ce42972/mullvad-types/src/settings/migrations/v2.rs#L132)):*

Ah yeah! Any migration converts from itself to latest. I just thought `v2` converted to `v3` etc. But no. We convert all the way from `v2 -> latest`. So then the migration of course needs to know the format of the latest settings.
</blockquote></details>

After looking further I have to open this again issue again. I was right in that `v2 -> v3` and not all the way to latest. It's the way this test is designed that makes it look like that. The tests import `super::super::try_migrate_settings`. So that tries to migrate all the way. But this does not really follow the open-closed principle. And it creates a lot of duplication.

Can't we redesign the tests here a little bit so that each migration module only tests the upgrade to the *next step*. And then maybe, if we see a need for it, a top level test that tests migrating through all layers or something like that.

faern (commented):
> Reviewed 2 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 4 of 10 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-types/src/settings/migrations/v4.rs, line 93 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-Mhs93Y-8pGITS7_c1Zb:-Mhs93Y-8pGITS7_c1Zc:bo0kfk5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5da62facb40cf572551c7e19ace460149ce42972/mullvad-types/src/settings/migrations/v4.rs#L93)):*
> ```XML
>                     protocol,
>                     port: Constraint::Any,
>                 }),
> ```

A bit of a nitpick. But the two branches that has `transport_protocol` set to `Only(protocol)` can be simplified to a single branch:

```
(port, Constraint::Only(protocol) => Constraint::Only(TransportPort {
    protocol,
    port,
}
```

faern (commented):
> Reviewed 3 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 7 of 10 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

faern (commented):
> Reviewed 1 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 4 of 10 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 5 of 10 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-cli/src/cmds/relay.rs, line 153 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919#-MhsKElYAQRzNkF2uYLj:-MhsKElYAQRzNkF2uYLk:b-5ch7vr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ce4f0a86bb21160f9fe1b5de63e103764a4adce3/mullvad-cli/src/cmds/relay.rs#L153)):*
> ```XML
>                                             .long("port")
>                                             .takes_value(true)
>                                             .default_value("any"),
> ```

`default_value` implies` `takes_value(true)`. So the line above can be removed. Same below where both are specified.

https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.default_value

faern (commented):
> Reviewed 2 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 6 of 10 files reviewed, all discussions resolved (waiting on @faern)

faern (commented):
> Reviewed 1 of 4 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 8 of 10 files reviewed, all discussions resolved (waiting on @faern)

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: 9 of 10 files reviewed, all discussions resolved

pinkisemils (approved):
> Reviewed 1 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2919)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2931, opened 2021-08-27
Disable GUI notification for newer version

tarkah:
I'm packaging Mullvad for [Solus OS](https://getsol.us/) and it'd be nice to disable the version check and GUI notification. Package updates are shipped to users every Friday, so the notice is really unnecessary, especially because they can't just "update" via a `.deb` or `.rpm`.

I've added a feature flag on the daemon, `disable-version-check`, which just replaces the app version request with an "empty" response that won't show as an update.

This seemed like the best place to disable this, as the `VersionUpdater` is pretty well tied into the daemon.

I've tested by patching this against 2021.3 and successfully did not see any version update notification. The output of `mullvad version` is:

```
Current version: 2021.3
	Is supported: false
	No newer version is available
```

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2931)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@ Line wrap the file at 100 chars.                                              Th
 ### Added
 - Added possibility to filter locations by provider in the desktop app.
 - Add WireGuard over TCP CLI option for all relays.
+- New GUI environment variable `DISABLE_VERSION_CHECK` has been added. If set to `1`, latest
+  version check will be disabled in the GUI, preventing a notification from occuring when a
+  newer version is available.
 
 #### Android
 - Added toggle for Split tunneling view to be able to show system apps
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,9 +27,8 @@ Line wrap the file at 100 chars.                                              Th
 ### Added
 - Added possibility to filter locations by provider in the desktop app.
 - Add WireGuard over TCP CLI option for all relays.
-- New GUI environment variable `DISABLE_VERSION_CHECK` has been added. If set to `1`, latest
-  version check will be disabled in the GUI, preventing a notification from occuring when a
-  newer version is available.
+- Add GUI environment variable `MULLVAD_DISABLE_UPDATE_NOTIFICATION`. If set to `1`, GUI
+  notification will be disabled when an update is available.
 
 #### Android
 - Added toggle for Split tunneling view to be able to show system apps
--- a/gui/src/main/index.ts
+++ b/gui/src/main/index.ts
@@ -88,6 +88,8 @@ const GUI_VERSION = app.getVersion().replace('.0', '');
 /// Mirrors the beta check regex in the daemon. Matches only well formed beta versions
 const IS_BETA = /^(\d{4})\.(\d+)-beta(\d+)$/;
 
+const VERSION_CHECK_DISABLED = process.env.DISABLE_VERSION_CHECK === '1';
+
 const SANDBOX_DISABLED = app.commandLine.hasSwitch('no-sandbox');
 
 enum AppQuitStage {
@@ -585,7 +587,9 @@ class ApplicationMain {
     }
 
     // fetch the latest version info in background
-    void this.fetchLatestVersion();
+    if (!VERSION_CHECK_DISABLED) {
+      void this.fetchLatestVersion();
+    }
 
     // reset the reconnect backoff when connection established.
     this.reconnectBackoff.reset();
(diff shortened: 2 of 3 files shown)


tarkah:
I just noticed the `IS_DEV_BUILD` flag that "disables" the updater, so I changed this to use that same code path

faern:
Am I understanding you correctly that the annoying part with the current app is that users get update notifications in the GUI?

If so. I think this better be solved in the GUI only. Change the GUI so it ignores version update broadcasts in one way or another. Also, make it configurable at runtime rather than build time. Then you can easily set an environment variable in your GUI launch `.desktop` file or similar instead of having to touch the daemon code.

tarkah:
It appears this would be the correct place to return early, to avoid setting the upgrade version and notifying.

https://github.com/mullvad/mullvadvpn-app/blob/ef7c3ca7e712354376cb7994264bd2dc5371bc90/gui/src/main/index.ts#L931-L966

This is called when the daemon is first connected and on daemon subscription:

https://github.com/mullvad/mullvadvpn-app/blob/ef7c3ca7e712354376cb7994264bd2dc5371bc90/gui/src/main/index.ts#L590-L591

https://github.com/mullvad/mullvadvpn-app/blob/ef7c3ca7e712354376cb7994264bd2dc5371bc90/gui/src/main/index.ts#L659-L676

tarkah:
I've updated this to the proposed GUI change using `DISABLE_VERSION_CHECK=1` env var.

I've backported this to 2021.3 and tested, it works as expected.

- Running without the variable shows the notification:

```
/usr/share/mullvad/mullvad-vpn
```

- Running with the variable doesn't show the notification:

```
DISABLE_VERSION_CHECK=1 /usr/share/mullvad/mullvad-vpn
```

- .desktop entry can be updated with the env var:

```
/usr/bin/env DISABLE_VERSION_CHECK=1 /usr/share/mullvad/mullvad-vpn %U
```

- CLI still reports the new version:

```
❯ mullvad version
Current version: 2021.3
        Is supported: true
        Suggested update: 2021.4
        Latest stable version: 2021.4
```

gui/src/main/index.ts
```diff
@@ -585,7 +587,9 @@ class ApplicationMain {
     }
 
     // fetch the latest version info in background
-    void this.fetchLatestVersion();
+    if (!VERSION_CHECK_DISABLED) {
+      void this.fetchLatestVersion();
+    }
```
tarkah, line 592: I've added the check here as well to disable the request from firing.

raksooo (commented):
Code looks good :+1:

Could you add this environment variable to the readme? It would fit well under this heading: https://github.com/mullvad/mullvadvpn-app#supported-environment-variables

faern:
Can we please prefix the env variable name with `MULLVAD_`? All our other env vars start with `MULLVAD_` or `TALPID_` depending on which component they belong to. I don't want to risk overlapping with other env vars spuriously defined on the system the user is running on.

CHANGELOG.md
```diff
@@ -27,6 +27,9 @@ Line wrap the file at 100 chars.                                              Th
 ### Added
 - Added possibility to filter locations by provider in the desktop app.
 - Add WireGuard over TCP CLI option for all relays.
+- New GUI environment variable `DISABLE_VERSION_CHECK` has been added. If set to `1`, latest
```
raksooo, line 30: Changelog entries should begin with a verb, e.g. add, fix, increase or force. As specified further up in the changelog.
> Entries should have the imperative form, just like commit messages. Start each entry with words like
add, fix, increase, force etc.. Not added, fixed, increased, forced etc.

Just changing the first sentence to this would solve it:
> Add GUI environment variable `DISABLE_VERSION_CHECK`

gui/src/main/index.ts
```diff
@@ -88,6 +88,8 @@ const GUI_VERSION = app.getVersion().replace('.0', '');
 /// Mirrors the beta check regex in the daemon. Matches only well formed beta versions
 const IS_BETA = /^(\d{4})\.(\d+)-beta(\d+)$/;
 
+const VERSION_CHECK_DISABLED = process.env.DISABLE_VERSION_CHECK === '1';
```
raksooo, line 91: In addition to the suggestion from faern, I would like to change the name a bit since it doesn't actually disable version checks. It just disables version checks initiated by the Electron app and ignores the result. How about something like `MULLVAD_DISABLE_UPDATE_NOTIFICATION`?

tarkah:
> Can we please prefix the env variable name with `MULLVAD_`? All our other env vars start with `MULLVAD_` or `TALPID_` depending on which component they belong to. I don't want to risk overlapping with other env vars spuriously defined on the system the user is running on.

Good recommendation, we definitely don't want to risk any overlaps.

raksooo (approved):
Great! Thanks for the contribution! 😄

PR #2943, opened 2021-09-06
Fix ping monitor when multiple peers are used

dlon:
Previously, the ping monitor would take any traffic received from any peer to mean that the tunnel was up. This worked poorly with multihop, since the entry hop could respond even when the exit hop could not be reached, incorrectly causing the TSM to enter the connected state.

This was fixed by requiring data to be received from every peer before considering the tunnel to be up.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2943)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ Line wrap the file at 100 chars.                                              Th
   desktop app.
 - Fix in-app notification button not working for some notifications.
 - Fix incorrectly positioned navigation bar title when navigating back to a scrolled down view.
+- Fix connectivity check for WireGuard multihop when the exit hop is down.
 
 #### Linux
 - Make offline monitor aware of routing table changes.
--- a/talpid-core/src/tunnel/wireguard/connectivity_check.rs
+++ b/talpid-core/src/tunnel/wireguard/connectivity_check.rs
@@ -1,6 +1,6 @@
 use crate::{
     ping_monitor::{new_pinger, Pinger},
-    tunnel::wireguard::stats::Stats,
+    tunnel::wireguard::stats::StatsMap,
 };
 use std::{
     net::Ipv4Addr,
@@ -110,12 +110,6 @@ impl ConnectivityMonitor {
         let start = Instant::now();
         while start.elapsed() < PING_TIMEOUT {
             if self.check_connectivity(Instant::now())? {
-                #[cfg(target_os = "linux")]
-                self.tunnel_handle.upgrade().and_then::<(), _>(|tunnel| {
-                    let tunnel = tunnel.lock().ok()?;
-                    tunnel.as_ref()?.slow_stats_refresh_rate();
-                    None
-                });
                 return Ok(true);
             }
             if self.should_shut_down(DELAY_ON_INITIAL_SETUP) {
@@ -182,7 +176,7 @@ impl ConnectivityMonitor {
 
     /// If None is returned, then the underlying tunnel has already been closed and all subsequent
     /// calls will also return None.
-    fn get_stats(&self) -> Option<Result<Stats, Error>> {
+    fn get_stats(&self) -> Option<Result<StatsMap, Error>> {
         self.tunnel_handle
             .upgrade()?
             .lock()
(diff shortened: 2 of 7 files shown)


pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2943)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/tunnel/wireguard/connectivity_check.rs, line 283 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2943#-MiuUzwl6fGCt3tRAaSW:-MiuUzwl6fGCt3tRAaSX:b-rferk6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/947798af06ab198ce7eb40c8517483930335ab95/talpid-core/src/tunnel/wireguard/connectivity_check.rs#L283)):*
> ```XML
>             } => {
>                 let mut rx_incremented = !stats.is_empty();
>                 for (key, peer_stats) in stats.iter() {
> ```

`stats.iter().all(|(key, peer_stats)| { new_stats.get(key).map(|new_stats| new_stats.rx_bytes > peer_stats.rx_bytes).unwrap_or(false))` - not sure if this is more readable, but it's less lines.

If for whatever reason a peer gets removed during the lifecycle of the tunnel, this would also kill the tunnel. This might well be good behavior, is it deliberate? 
___
*[talpid-core/src/tunnel/wireguard/stats.rs, line 100 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2943#-MiuRZTdBzhTlznd9N5q:-MiuRZTdBzhTlznd9N5r:b-wattgi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/947798af06ab198ce7eb40c8517483930335ab95/talpid-core/src/tunnel/wireguard/stats.rs#L100)):*
> ```XML
>             if let DeviceNla::Peers(peers) = nla {
>                 // iterate over all peer attributes
>                 let peer_iter = peers.iter().map(|peer| peer.0.as_slice()).flatten();
> ```

I think it makes sense to have a nested loop here instead of iterating through all peer NLAs. That way, there's no chance of confusing which byte NLAs belong to which public key, IMO.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2943)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-core/src/tunnel/wireguard/connectivity_check.rs, line 283 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2943#-MiuUzwl6fGCt3tRAaSW:-MiujytN-Z8Q_OpVsabe:b-z5ue4b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/947798af06ab198ce7eb40c8517483930335ab95/talpid-core/src/tunnel/wireguard/connectivity_check.rs#L283)):*
I like that better.

> If for whatever reason a peer gets removed during the lifecycle of the tunnel, this would also kill the tunnel. This might well be good behavior, is it deliberate?

That is deliberate. It could fail more explicitly, I suppose. But I think that this is good.
___
*[talpid-core/src/tunnel/wireguard/stats.rs, line 100 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2943#-MiuRZTdBzhTlznd9N5q:-MiukH3078HKerGBKa4W:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/947798af06ab198ce7eb40c8517483930335ab95/talpid-core/src/tunnel/wireguard/stats.rs#L100)):*

I think it makes sense to have a nested loop here instead of iterating through all peer NLAs. That way, there's no chance of confusing which byte NLAs belong to which public key, IMO.
</blockquote></details>

Done.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2943)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2945, opened 2021-09-06
Always send DNS requests inside the tunnel for public IPs on Linux

dlon:
Previously, whenever DNS was not managed by some system service (e.g., a simple static `resolv.conf`), excluded processes always tried to reach resolvers on public IPs _outside_ the tunnel. Ad blocking DNS servers are in the "public" range (according to the definition used by the app), so ad blocking made DNS unusable for excluded processes.

This PR ensures that public DNS requests are sent _inside_ the tunnel. This solves the problem above and overall reduces the complexity of how DNS is handled for excluded apps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,9 @@ Line wrap the file at 100 chars.                                              Th
 - Move OpenVPN and WireGuard settings in the advanced settings view into separate settings views.
 - Return to main view in desktop app after being hidden/closed for two minutes.
 
+#### Linux
+- Always send DNS requests inside the tunnel for excluded processes when using public custom DNS.
+
 #### Windows
 - Upgrade Wintun from 0.10.4 to 0.13.
 
@@ -64,6 +67,7 @@ Line wrap the file at 100 chars.                                              Th
 - Make offline monitor aware of routing table changes.
 - Assign local DNS servers to more appropriate interfaces when using systemd-resolved.
 - Disable DNS over TLS for tunnel's DNS config when using systemd-resolved.
+- Fix DNS when combining static resolv.conf with ad blocking DNS.
 
 #### Windows
 - Fix failure to restart the daemon when resuming from "fast startup" hibernation.
--- a/docs/split-tunneling.md
+++ b/docs/split-tunneling.md
@@ -43,7 +43,7 @@ Some definitions of terms used later to describe behavior:
 *: On platforms where we have custom firewall integration. This is currently on desktop operating
   systems, and not mobile.
 
-### Windows
+### Windows and Linux
 
 | In-app DNS setting | Normal & Excluded app |
 |-|-|
@@ -51,24 +51,13 @@ Some definitions of terms used later to describe behavior:
 | **Private custom DNS** (e.g. 10.0.1.1) | LAN (to 10.0.1.1) |
 | **Public custom DNS** (e.g. 8.8.8.8) | In tunnel (to 8.8.8.8) |
 
-In other words: Normal and excluded processes always behave the same. This is due to the
-Windows DNS cache service is the single origin for all DNS requests.
+In other words: Normal and excluded processes always behave the same. This is because DNS is
+typically handled by a service, e.g. DNS cache on Windows or systemd-resolved's resolver on Linux,
+which is not an excluded process.
 
-### Linux
-
-| In-app DNS setting | Normal app | Excluded app |
-|-|-|-|
-| **Default DNS** | In tunnel (to relay) | In tunnel (to relay) |
-| **Private custom DNS** (e.g. 10.0.1.1) | LAN (to 10.0.1.1) | LAN (to 10.0.1.1) |
-| **Public custom DNS** (e.g. 8.8.8.8) | In tunnel (to 8.8.8.8) | Outside tunnel* (to 8.8.8.8) |
-
-*: Only if a local DNS resolver, such as systemd-resolved is **not in use**. Because if a
-local DNS resolver is in use the requests will go there and that resolver in turn will then
-send requests in the tunnel.
-
-In other words: Normal and excluded processes behave the same in all cases except when Custom DNS
-is enabled, pointed to a publicly available IP and the system is not set up to use a localhost DNS
-resolver.
+For the sake of simplicity and consistency, public custom DNS is also sent *in tunnel* when using a
+plain old static `resolv.conf`, even though it is technically possible to exclude public custom DNS
+in that case.
 
 ### Android
 
--- a/talpid-core/src/dns/linux/mod.rs
+++ b/talpid-core/src/dns/linux/mod.rs
@@ -8,8 +8,12 @@ use self::{
     network_manager::NetworkManager, resolvconf::Resolvconf, static_resolv_conf::StaticResolvConf,
     systemd_resolved::SystemdResolved,
 };
-use crate::routing::RouteManagerHandle;
-use std::{env, fmt, net::IpAddr, path::Path};
+use crate::{
+    firewall::is_local_address,
+    routing::{Node, RequiredRoute, RouteManagerHandle},
+};
+use ipnetwork::IpNetwork;
+use std::{collections::HashSet, env, fmt, net::IpAddr, path::Path};
 
 
 const RESOLV_CONF_PATH: &str = "/etc/resolv.conf";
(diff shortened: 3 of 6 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945)* status: 0 of 6 files reviewed, all discussions resolved

___
*[docs/split-tunneling.md, line 31 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945#-Mizvv7V3Glwh0dHx_UD:-Mizvv7V3Glwh0dHx_UE:b-97xkcy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a1ca95a736f29f8c629e8d8882904ea8ee860e7/docs/split-tunneling.md#L31)):*
> ```Markdown
> 
> * **In tunnel** - DNS requests are sent in the VPN tunnel. Firewall rules ensure they
>     are not allowed outside the tunnel*.
> ```

I guess this is not quite true for excluded apps. We don't block DNS traffic outside the tunnel in general for those, at least on Linux.
___
*[talpid-core/src/dns/linux/mod.rs, line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945#-MizwJ72-0gocVmbOYeW:-MizwJ72-0gocVmbOYeX:b-s30vrh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a1ca95a736f29f8c629e8d8882904ea8ee860e7/talpid-core/src/dns/linux/mod.rs#L83)):*
> ```XML
>         let routes = get_dns_routes(interface, servers)?;
>         self.handle
>             .block_on(self.route_manager.add_routes(routes.clone()))
> ```

A strange side effect is that you can't reach these IPs outside the tunnel on any port.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945)* status: 0 of 6 files reviewed, 1 unresolved discussion

___
*[talpid-core/src/dns/linux/mod.rs, line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945#-MizzFUL32VGiFlTYUCk:-MizzFUM3IeXDKGh3aU8:b-s746n4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a1ca95a736f29f8c629e8d8882904ea8ee860e7/talpid-core/src/dns/linux/mod.rs#L83)):*
> ```XML
>         let routes = get_dns_routes(interface, servers)?;
>         self.handle
>             .block_on(self.route_manager.add_routes(routes.clone()))
> ```

The issue that was mentioned could probably be solved by using some `ip rule` trickery involving port 53 or perhaps using only nftables. Is it worth doing?

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[docs/split-tunneling.md, line 31 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945#-Mizvv7V3Glwh0dHx_UD:-Mj-1ZOO640Gz9nEGz1p:b2fco2g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a1ca95a736f29f8c629e8d8882904ea8ee860e7/docs/split-tunneling.md#L31)):*

I guess this is not quite true for excluded apps. We don't block DNS traffic outside the tunnel in general for those, at least on Linux.
</blockquote></details>

Then this should be updated to reflect that.
___
*[talpid-core/src/dns/linux/mod.rs, line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945#-MizzFUL32VGiFlTYUCk:-Mj-4PpL2LaP8t5HC4cu:b-evltrz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a1ca95a736f29f8c629e8d8882904ea8ee860e7/talpid-core/src/dns/linux/mod.rs#L83)):*

The issue that was mentioned could probably be solved by using some `ip rule` trickery involving port 53 or perhaps using only nftables. Is it worth doing?
</blockquote></details>

One way to aid this would be to just use firewall rules to clear the mark after it's been applied. Not sure if that's less of a hack though.
___
*[talpid-core/src/firewall/linux.rs, line 356 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945#-Mj-326t47jlexGigtOp:-Mj-326t47jlexGigtOq:b7ot5c6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a1ca95a736f29f8c629e8d8882904ea8ee860e7/talpid-core/src/firewall/linux.rs#L356)):*
> ```XML
>         {
>             for server in dns_servers.iter().filter(|server| {
>                 !super::is_local_address(server)
> ```

This closure could be a named function, if only for legibility.

Also, I'm know I too am guilty of this, but `ipv6_gateway == Some(server)` is also a valid way to check for equality with options, I think.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/firewall/linux.rs, line 356 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945#-Mj-326t47jlexGigtOp:-Mj08sJyBZtbsR2fXejW:b-dt3hac) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a1ca95a736f29f8c629e8d8882904ea8ee860e7/talpid-core/src/firewall/linux.rs#L356)):*

This closure could be a named function, if only for legibility.

Also, I'm know I too am guilty of this, but `ipv6_gateway == Some(server)` is also a valid way to check for equality with options, I think.
</blockquote></details>

Done. `ipv6_gateway == Some(server)` didn't work (`ipv6_gateway` is an `Option<Ipv6Addr>` rather than `Option<IpAddr>`), but I simplified it slightly.
___
*[talpid-core/src/dns/linux/mod.rs, line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945#-MizzFUL32VGiFlTYUCk:-Mj0JhTZ4_Nvsp5fXetN:b6xl5k0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a1ca95a736f29f8c629e8d8882904ea8ee860e7/talpid-core/src/dns/linux/mod.rs#L83)):*

One way to aid this would be to just use firewall rules to clear the mark after it's been applied. Not sure if that's less of a hack though.
</blockquote></details>

Done. I think that this is better, actually.

pinkisemils (approved):
> Reviewed 4 of 6 files at r2, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[talpid-core/src/dns/linux/mod.rs, line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945#-MizzFUL32VGiFlTYUCk:-Mj439A4Em5IpSYrnd1I:brbd4rv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a1ca95a736f29f8c629e8d8882904ea8ee860e7/talpid-core/src/dns/linux/mod.rs#L83)):*

Done. I think that this is better, actually.
</blockquote></details>

This might have implications on the split tunneling guide that's going to be published later :/ What's the priority used for that rule? I wonder if there's a way to circumvent this specific rule so that users can configure the firewall so that DNS traffic *is* being split. Still, this solution is the superior one given that it's just that much simpler.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945)* status: 5 of 6 files reviewed, all discussions resolved (waiting on @pinkisemils)

___
*[talpid-core/src/dns/linux/mod.rs, line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2945#-MizzFUL32VGiFlTYUCk:-Mj4f1iI2k66HAFT3AAG:b-ta3z7s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a1ca95a736f29f8c629e8d8882904ea8ee860e7/talpid-core/src/dns/linux/mod.rs#L83)):*

This might have implications on the split tunneling guide that's going to be published later :/ What's the priority used for that rule? I wonder if there's a way to circumvent this specific rule so that users can configure the firewall so that DNS traffic *is* being split. Still, this solution is the superior one given that it's just that much simpler.
</blockquote></details>

Hopefully it won't have any implications. All this does is to *not* set the `fwmark` for processes in the cgroup, if the destination is one of the public resolvers. It shouldn't mess with other rules setting marks.

PR #2960, opened 2021-09-15
Update TypeScript

raksooo:
This PR updates TypeScript from 4.3.2 to 4.4.3 and adjusts our code to the breaking change where the catch-clause argument is now of type `unknown` instead of `any`.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960)


pronebird (approved):
> Reviewed 21 of 21 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[gui/src/main/gui-settings.ts, line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960#-MjcjPqg7oVtEH9ck-0y:-MjcjPqg7oVtEH9ck-0z:b-kp5y2j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/414fc9082877c4c018aec21ab52ab648a56eab81/gui/src/main/gui-settings.ts#L106)):*
> ```XML
>       };
>     } catch (e) {
>       const error = e as NodeJS.ErrnoException;
> ```

FYI

`JSON.parse` throws `SyntaxError`, so if you want to handle it properly, then I believe you should really use  `instanceof` here. Big question if it really works though.


```
if (e instanceof NodeJS.ErrnoException && e.code == 'ENOENT') {
    // branch A
} else {
  let e = e as Error
  // branch B
}
```

However `SyntaxError.code` does not exist, so this should still work correctly.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960)* status: 20 of 22 files reviewed, all discussions resolved (waiting on @pronebird)

___
*[gui/src/main/gui-settings.ts, line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960#-MjcjPqg7oVtEH9ck-0y:-Mje6nZ_90cg2UO60N-Z:b-iqoyng) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/414fc9082877c4c018aec21ab52ab648a56eab81/gui/src/main/gui-settings.ts#L106)):*

FYI

`JSON.parse` throws `SyntaxError`, so if you want to handle it properly, then I believe you should really use  `instanceof` here. Big question if it really works though.


```
if (e instanceof NodeJS.ErrnoException && e.code == 'ENOENT') {
    // branch A
} else {
  let e = e as Error
  // branch B
}
```

However `SyntaxError.code` does not exist, so this should still work correctly.
</blockquote></details>

Yes it would have worked correctly but I added a check to make sure it's an error with a code. Makes no difference when running but makes the types correct.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960)* status: 20 of 22 files reviewed, all discussions resolved (waiting on @pronebird)

___
*[gui/src/main/gui-settings.ts, line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960#-MjcjPqg7oVtEH9ck-0y:-MjhmqyABhOWebnJecnk:b-jw7g0g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/414fc9082877c4c018aec21ab52ab648a56eab81/gui/src/main/gui-settings.ts#L106)):*

Yes it would have worked correctly but I added a check to make sure it's an error with a code. Makes no difference when running but makes the types correct.
</blockquote></details>

Why not just cast to `Error` and then go with`instanceof`? This would avoid adding another helper that's only used once.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960)* status: 20 of 22 files reviewed, all discussions resolved (waiting on @pronebird)

___
*[gui/src/main/gui-settings.ts, line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960#-MjcjPqg7oVtEH9ck-0y:-Mjhngvi0tntK2RWdHPC:b-o4whr8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/414fc9082877c4c018aec21ab52ab648a56eab81/gui/src/main/gui-settings.ts#L106)):*

Why not just cast to `Error` and then go with`instanceof`? This would avoid adding another helper that's only used once.
</blockquote></details>

It's not possible to do this
```
e instanceof NodeJS.ErrnoException
```
since `NodeJS.ErrnoException` is just a TypesScript interface and there's no exported error to which you can compare it to. The `instanceof` operator checks if a value is an instance of a JavaScript class.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960)* status: 20 of 22 files reviewed, all discussions resolved (waiting on @pronebird)

___
*[gui/src/main/gui-settings.ts, line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960#-MjcjPqg7oVtEH9ck-0y:-MjhqR2J8Aa_oWFYsbcy:buiqnjd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/414fc9082877c4c018aec21ab52ab648a56eab81/gui/src/main/gui-settings.ts#L106)):*

It's not possible to do this
```
e instanceof NodeJS.ErrnoException
```
since `NodeJS.ErrnoException` is just a TypesScript interface and there's no exported error to which you can compare it to. The `instanceof` operator checks if a value is an instance of a JavaScript class.
</blockquote></details>

Hmmm hold on, but `Error` has `code` according to Node.js:

https://nodejs.org/api/errors.html#errors_error_code

So casting to `Error` should be enough.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960)* status: 20 of 22 files reviewed, all discussions resolved (waiting on @pronebird)

___
*[gui/src/main/gui-settings.ts, line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960#-MjcjPqg7oVtEH9ck-0y:-MjhyZdR2sQhHThkPoLL:bhiduf1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/414fc9082877c4c018aec21ab52ab648a56eab81/gui/src/main/gui-settings.ts#L106)):*

Hmmm hold on, but `Error` has `code` according to Node.js:

https://nodejs.org/api/errors.html#errors_error_code

So casting to `Error` should be enough.
</blockquote></details>

Since the TypeScript `Error` type doesn't contain `code` while the JavaScript `Error` does we decided to type it as `Error & { code?: string }`.

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2960#-:-Mjhz4ABACPMwMzc1Apy:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2981, opened 2021-09-23
Fix initial version cache fetch

dlon:
`GetVersionInfo` updates the version cache if there's no cache yet, but it doesn't wait on the fetch to complete, causing the first call to `mullvad version` to fail if there's no cache yet. This fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2981)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2981)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2982, opened 2021-09-23
Retry failed API requests made indirectly by frontends

dlon:
This primarily affects these account-related RPCs:
* `CreateNewAccount`
* `GetAccountData`
* `GetWwwAuthToken`
* `SubmitVoucher`

In general, a request is sent up to 3 times. On failure, the daemon will use a different endpoint to reach the API.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982)

### DIFF
--- a/mullvad-daemon/src/account.rs
+++ b/mullvad-daemon/src/account.rs
@@ -2,16 +2,21 @@ use chrono::{DateTime, Utc};
 use futures::future::{abortable, AbortHandle};
 use mullvad_rpc::{
     availability::ApiAvailabilityHandle,
-    rest::{self, MullvadRestHandle},
+    rest::{self, Error as RestError, MullvadRestHandle},
     AccountsProxy,
 };
 use mullvad_types::account::{AccountToken, VoucherSubmission};
-use std::time::Duration;
-use talpid_core::future_retry::{retry_future_with_backoff, ExponentialBackoff, Jittered};
+use std::{future::Future, time::Duration};
+use talpid_core::future_retry::{
+    constant_interval, retry_future_n, retry_future_with_backoff, ExponentialBackoff, Jittered,
+};
 
-const RETRY_INTERVAL_INITIAL: Duration = Duration::from_secs(4);
-const RETRY_INTERVAL_FACTOR: u32 = 5;
-const RETRY_INTERVAL_MAX: Duration = Duration::from_secs(24 * 60 * 60);
+const RETRY_ACTION_INTERVAL: Duration = Duration::from_millis(500);
+const RETRY_ACTION_MAX_RETRIES: usize = 2;
+
+const RETRY_EXPIRY_CHECK_INTERVAL_INITIAL: Duration = Duration::from_secs(4);
+const RETRY_EXPIRY_CHECK_INTERVAL_FACTOR: u32 = 5;
+const RETRY_EXPIRY_CHECK_INTERVAL_MAX: Duration = Duration::from_secs(24 * 60 * 60);
 
 
 pub struct Account(());
@@ -20,12 +25,42 @@ pub struct Account(());
 pub struct AccountHandle {
     api_availability: ApiAvailabilityHandle,
     initial_check_abort_handle: AbortHandle,
-    pub proxy: AccountsProxy,
+    proxy: AccountsProxy,
 }
 
 impl AccountHandle {
+    pub fn create_account(&self) -> impl Future<Output = Result<AccountToken, rest::Error>> {
+        let mut proxy = self.proxy.clone();
+        retry_future_n(
+            move || proxy.create_account(),
+            Self::should_retry,
+            constant_interval(RETRY_ACTION_INTERVAL),
+            RETRY_ACTION_MAX_RETRIES,
+        )
+    }
+
+    pub fn get_www_auth_token(
+        &self,
+        account: AccountToken,
+    ) -> impl Future<Output = Result<String, rest::Error>> {
+        let proxy = self.proxy.clone();
+        retry_future_n(
+            move || proxy.get_www_auth_token(account.clone()),
+            Self::should_retry,
+            constant_interval(RETRY_ACTION_INTERVAL),
+            RETRY_ACTION_MAX_RETRIES,
+        )
+    }
+
     pub async fn check_expiry(&self, token: AccountToken) -> Result<DateTime<Utc>, rest::Error> {
-        let result = self.proxy.get_expiry(token).await;
+        let proxy = self.proxy.clone();
+        let result = retry_future_n(
+            move || proxy.get_expiry(token.clone()),
+            Self::should_retry,
+            constant_interval(RETRY_ACTION_INTERVAL),
+            RETRY_ACTION_MAX_RETRIES,
+        )
+        .await;
         if handle_expiry_result_inner(&result, &self.api_availability) {
             self.initial_check_abort_handle.abort();
         }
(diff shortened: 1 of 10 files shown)


pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[mullvad-daemon/src/account.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982#-MkarhsY0N-JWCJrIzR7:-MkarhsY0N-JWCJrIzR8:b-53qo07) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8745d1507497561929069e708627b47a2df0898d/mullvad-daemon/src/account.rs#L14)):*
> ```XML
> };
> 
> const RETRY_ACTION_INTERVAL: Duration = Duration::from_millis(500);
> ```

If the retry for a request fails due to network reasons and it's expected that another API address will be tried, should the retry interval even be used? I'd assume that the request will already take some time to time out. 
___
*[mullvad-daemon/src/account.rs, line 90 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982#-MkamrbD9ZgIaQ1f1hUS:-MkamrbD9ZgIaQ1f1hUT:bqqjfjz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8745d1507497561929069e708627b47a2df0898d/mullvad-daemon/src/account.rs#L90)):*
> ```XML
>     }
> 
>     fn should_retry<T>(result: &Result<T, RestError>) -> bool {
> ```

Whole function could be expressed in a match statement.
___
*[mullvad-daemon/src/wireguard.rs, line 36 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982#-Mkaw60mEKTVlCPnTgLV:-Mkaw60mEKTVlCPnTgLW:bbhkaex) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8745d1507497561929069e708627b47a2df0898d/mullvad-daemon/src/wireguard.rs#L36)):*
> ```XML
> 
> #[cfg(not(target_os = "android"))]
> const SHORT_RETRY_INTERVAL: Duration = Duration::from_millis(500);
> ```

What's special about Android here?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[mullvad-daemon/src/account.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982#-MkarhsY0N-JWCJrIzR7:-MkazztzD9K02VnbGviE:b-1cqboa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8745d1507497561929069e708627b47a2df0898d/mullvad-daemon/src/account.rs#L14)):*

If the retry for a request fails due to network reasons and it's expected that another API address will be tried, should the retry interval even be used? I'd assume that the request will already take some time to time out. 
</blockquote></details>

It could fail for other reasons. Is it worth extra complexity just to handle the timeout case differently? Or would it be fine to set the interval to `0` even if the server returns a 500 status, for example?
___
*[mullvad-daemon/src/account.rs, line 90 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982#-MkamrbD9ZgIaQ1f1hUS:-Mkbm_A4DKZS_4mgTHiu:b-896fja) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8745d1507497561929069e708627b47a2df0898d/mullvad-daemon/src/account.rs#L90)):*

Whole function could be expressed in a match statement.
</blockquote></details>

Done!
___
*[mullvad-daemon/src/wireguard.rs, line 36 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982#-Mkaw60mEKTVlCPnTgLV:-MkayrMG0QWFD_n5CaSf:bfz7bfw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8745d1507497561929069e708627b47a2df0898d/mullvad-daemon/src/wireguard.rs#L36)):*

What's special about Android here?
</blockquote></details>

It doesn't have a factory reset function, which is the only place where `remove_key` is used. Setting a new account token calls `remove_key_with_backoff` instead since there's no need to complete it in a timely manner.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982)* status: all files reviewed, 2 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-daemon/src/account.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982#-MkarhsY0N-JWCJrIzR7:-Mkfv36R6dNkXZt3Vk6z:b-8ue1sm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8745d1507497561929069e708627b47a2df0898d/mullvad-daemon/src/account.rs#L14)):*

It could fail for other reasons. Is it worth extra complexity just to handle the timeout case differently? Or would it be fine to set the interval to `0` even if the server returns a 500 status, for example?
</blockquote></details>

I'm not sure that waiting an extra 500 miliseconds will help resolve any issues preventing the app from reaching the API. But I think maybe @faern might be a better arbiter here - this is sensitive enough to require more than 2 opinions IMO.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982)* status: 4 of 9 files reviewed, 2 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-daemon/src/account.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982#-MkarhsY0N-JWCJrIzR7:-MkgXxx7DBc4u9do2yDL:b-85rjjf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8745d1507497561929069e708627b47a2df0898d/mullvad-daemon/src/account.rs#L14)):*

I'm not sure that waiting an extra 500 miliseconds will help resolve any issues preventing the app from reaching the API. But I think maybe @faern might be a better arbiter here - this is sensitive enough to require more than 2 opinions IMO.
</blockquote></details>

Done. Interval is zero and retries are only made when the error is network-related.

pinkisemils (approved):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> Reviewed 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[mullvad-daemon/src/wireguard.rs, line 179 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982#-MkvTTHXDslBueNOX1pu:-MkvTTHYERqIGzN3r1kw:by0gh69) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4a22e409ce7a5816af1ef2c638e1ed9be3090044/mullvad-daemon/src/wireguard.rs#L179)):*
> ```XML
>         let mut rpc = mullvad_rpc::WireguardKeyProxy::new(self.http_handle.clone());
>         let api_handle = self.availability_handle.clone();
>         let api_handle_2 = api_handle.clone();
> ```

`api_handle_2` could just be `retry_api_handle`. 
___
*[mullvad-daemon/src/wireguard.rs, line 201 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2982#-MkvTfO2CgxqxVLoj0k8:-MkvTfO2CgxqxVLoj0k9:bf4trb3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4a22e409ce7a5816af1ef2c638e1ed9be3090044/mullvad-daemon/src/wireguard.rs#L201)):*
> ```XML
>     }
> 
>     fn should_retry_removal(error: &RestError, api_handle: &ApiAvailabilityHandle) -> bool {
> ```

If the removal fails due to the host being offline, I think we should probably try and remove the key once the host gets online - but I see this is as a higher level issue, not to be fixed in this module, and not to be fixed in this PR.

PR #2991, opened 2021-09-28
Use larger buffer for `ReadDirectoryChangesW`

dlon:
It's not uncommon for the log to be flooded with these messages:

```
[talpid_core::split_tunnel::imp::path_monitor][DEBUG] Change event buffer is empty
```

This seems to be because the buffer used by the path monitor is too small.

It's not super critical, so I've just enlarged the buffer and changed the log level to `trace`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2991)


mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2991)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2994, opened 2021-09-29
Use `socket2` for conversion to or from OS/C address types

dlon:


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2994)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[talpid-core/src/windows.rs, line 414 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2994#-MklOY8L8W6xWQwj0hIr:-MklOY8M1f564CJnpNhY:b-3hs0tg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c07a556ba1d4041acd0aaee56895216648eb5132/talpid-core/src/windows.rs#L414)):*
> ```XML
>             )
>             .as_socket()
>             .ok_or(Error::UnknownAddressFamily(0)),
> ```

This could probably be unwrapped instead.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2994)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/windows.rs, line 389 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2994#-MklrUArDTrsAmXP3BIU:-MklrUAsAnI_Mm1qqHfx:b-6v371l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/21f275af3bb2117dffb63cb9f3caec76c984395e/talpid-core/src/windows.rs#L389)):*
> ```XML
> /// Converts a `SocketAddr` to `SOCKADDR_INET`
> pub fn inet_sockaddr_from_socketaddr(addr: SocketAddr) -> SOCKADDR_INET {
>     unsafe { *(SockAddr::from(addr).as_ptr() as *const _) }
> ```

Is it safe to treat a `libc::sockaddr` as `SOCKADDR_INET`? Are they guaranteed to be the same type/size?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2994)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[talpid-core/src/windows.rs, line 389 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2994#-MklrUArDTrsAmXP3BIU:-Mklu_oK7XxnhWneBcYU:bs2krno) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/21f275af3bb2117dffb63cb9f3caec76c984395e/talpid-core/src/windows.rs#L389)):*

Is it safe to treat a `libc::sockaddr` as `SOCKADDR_INET`? Are they guaranteed to be the same type/size?
</blockquote></details>

It is safe because it is a pointer to a `sockaddr_storage`, which is large enough to contain `SOCKADDR_INET` (which itself can be treated as either `sockaddr_in6` or `sockaddr_in`).

Granted, this makes an assumption about hidden/internal data in `SockAddr`. The previous revision did not, but it's not as simple.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2994)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/windows.rs, line 389 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2994#-MklrUArDTrsAmXP3BIU:-MkpT_nd6YdHMrCP5RFs:bkzj1qq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/21f275af3bb2117dffb63cb9f3caec76c984395e/talpid-core/src/windows.rs#L389)):*

It is safe because it is a pointer to a `sockaddr_storage`, which is large enough to contain `SOCKADDR_INET` (which itself can be treated as either `sockaddr_in6` or `sockaddr_in`).

Granted, this makes an assumption about hidden/internal data in `SockAddr`. The previous revision did not, but it's not as simple.
</blockquote></details>

`SockAddr::as_ptr` returns a pointer to a `libc::sockaddr`. So if it's guaranteed safe to treat such a pointer as a pointer to a `SOCKADDR_INET` and dereference it, then we are safe. But then I think that should be documented with a `// SAFETY: ` comment here.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2994)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[talpid-core/src/windows.rs, line 389 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/2994#-MklrUArDTrsAmXP3BIU:-MkpjTrT82kbM0B1t9RK:b-dtbec1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/21f275af3bb2117dffb63cb9f3caec76c984395e/talpid-core/src/windows.rs#L389)):*

`SockAddr::as_ptr` returns a pointer to a `libc::sockaddr`. So if it's guaranteed safe to treat such a pointer as a pointer to a `SOCKADDR_INET` and dereference it, then we are safe. But then I think that should be documented with a `// SAFETY: ` comment here.
</blockquote></details>

Done. I didn't like the assumption that was made about `*sockaddr` pointing to a `sockaddr_storage`, so I had to make it slightly more complicated.

PR #3077, opened 2021-11-02
Ignore local resolvers for systemd-resolved and simplify

pinkisemils:
To simplify and not clash with other system configurators, I've simplified the systemd-resolved code significantly, by doing 2 things:
- ignore local IP addresses
- remove all route watching code
- only apply config to tunnel interface
- stop watching for changes to the DNS config
The last point might be somewhat contentious, so I'll conduct some further testing on older distributions, where this used to be a problem.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077)


dlon (changes requested):
> Reviewed 3 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077)* status: all files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 44 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077#-MnWdzZSCb-BkM9mCLi1:-MnWdzZSCb-BkM9mCLi2:b-wgtyuz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d0d91b163ac36241efbeb9eb4340dbe6bf33a594/talpid-core/src/dns/linux/systemd_resolved.rs#L44)):*
> ```XML
>         _route_manager: RouteManagerHandle,
>         interface_name: &str,
>         servers: &[IpAddr],
> ```

Maybe we could filter `servers` by just passing in `tunnel_gateways: &[IpAddr]` as an argument and using `!is_local_address(addr) || tunnel_gateways.contains(addr)`. As was done in #2425.
___
*[talpid-core/src/tunnel_state_machine/connected_state.rs, line 98 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077#-MnWcAJd-5VFu2_1oMB_:-MnWcAJe94koben6fy7U:bf2sh7s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d0d91b163ac36241efbeb9eb4340dbe6bf33a594/talpid-core/src/tunnel_state_machine/connected_state.rs#L98)):*
> ```XML
>                     .iter()
>                     .cloned()
>                     .filter(|ip| !crate::firewall::is_local_address(ip))
> ```

The firewall will likely drop DNS requests/responses for local addresses if you do this. It also likely breaks the non-systemd-resolved managers. I don't think this should be changed.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 44 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077#-MnWdzZSCb-BkM9mCLi1:-MnWh-kJ5zyw_Zcuydwl:btb2iy6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d0d91b163ac36241efbeb9eb4340dbe6bf33a594/talpid-core/src/dns/linux/systemd_resolved.rs#L44)):*

Maybe we could filter `servers` by just passing in `tunnel_gateways: &[IpAddr]` as an argument and using `!is_local_address(addr) || tunnel_gateways.contains(addr)`. As was done in #2425.
</blockquote></details>

I think the filtering should be done at a higher level.
___
*[talpid-core/src/tunnel_state_machine/connected_state.rs, line 98 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077#-MnWcAJd-5VFu2_1oMB_:-MnWfdb609yR5LZ7sScl:b-ef3mf4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d0d91b163ac36241efbeb9eb4340dbe6bf33a594/talpid-core/src/tunnel_state_machine/connected_state.rs#L98)):*

The firewall will likely drop DNS requests/responses for local addresses if you do this. It also likely breaks the non-systemd-resolved managers. I don't think this should be changed.
</blockquote></details>

I'll fix the issue of breaking DNS with other implementations by not applying an empty list.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 44 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077#-MnWdzZSCb-BkM9mCLi1:-MnWox2b5pDghayDl9SP:b-bwd4s2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d0d91b163ac36241efbeb9eb4340dbe6bf33a594/talpid-core/src/dns/linux/systemd_resolved.rs#L44)):*

I think the filtering should be done at a higher level.
</blockquote></details>

If a static resolv.conf is used, doesn't it still need to include the local addresses? This is sort of fine if you only use local resolvers, as it just leaves resolv.conf alone, but it doesn't work if for whatever reason you mix public and non-public addresses. That will be obsoleted eventually, but it is still possible for now.

faern (commented):
> Reviewed 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 44 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077#-MnWdzZSCb-BkM9mCLi1:-Mn_9q3Z7ikrTMS07uro:b8s8wsw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d0d91b163ac36241efbeb9eb4340dbe6bf33a594/talpid-core/src/dns/linux/systemd_resolved.rs#L44)):*

If a static resolv.conf is used, doesn't it still need to include the local addresses? This is sort of fine if you only use local resolvers, as it just leaves resolv.conf alone, but it doesn't work if for whatever reason you mix public and non-public addresses. That will be obsoleted eventually, but it is still possible for now.
</blockquote></details>

This is a fair point, but it won't leave DNS in a broken state. I don't think this is that big of an issue.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[talpid-core/src/dns/linux/systemd_resolved.rs, line 44 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3077#-MnWdzZSCb-BkM9mCLi1:-Mn_BDK26d4GWpgtUgdW:bdlrsdd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d0d91b163ac36241efbeb9eb4340dbe6bf33a594/talpid-core/src/dns/linux/systemd_resolved.rs#L44)):*

This is a fair point, but it won't leave DNS in a broken state. I don't think this is that big of an issue.
</blockquote></details>

Good point. It will *work*. We should probably just do the same thing on Windows/macOS, but that can wait.

PR #3081, opened 2021-11-03
CI: Switch to macOS 11 and iOS 13

pronebird:

Git checklist:

This PR migrates iOS tests to iOS 13 simulator runtime and macOS 11. The change is needed because older Xcode does not recognise API introduced in iOS 15 and fails to run tests even though all calls to newer API are guarded with availability checks. 

Xcode 13 on the other side, does not have that problem, however it does not come with iOS 12 simulator runtime, so I had to bump the test suite to use iOS 13 instead. 

[The alternative solution that preserves the iOS 12 runtime](https://github.com/mullvad/mullvadvpn-app/commit/1d647738d40fc3be58abf3049ce592f2c62423aa
), involves installing it on macOS 11 image and then caching it with `actions/cache@v2`, however that alone adds about 3 minute of extra overhead to set up the testing environment and takes 2 GB extra cache space.

Hence I favour running tests on iOS 13, partly because we'll eventually have to abandon iOS 12. The tests themselves aren't iOS 12 specific so hence even more reasons to not complicate our test setup.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3081)

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -9,10 +9,10 @@ on:
 jobs:
     test:
         name: Unit tests
-        runs-on: macos-10.15
+        runs-on: macos-11
         strategy:
             matrix:
-                destination: ['platform=iOS Simulator,OS=12.4,name=iPhone 8']
+                destination: ['platform=iOS Simulator,OS=13.7,name=iPhone 8']
         env:
             source_packages_dir: .spm
         steps:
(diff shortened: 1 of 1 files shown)


raksooo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3081#-:-Mn_Wohu7rCAPppp4GpF:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3094, opened 2021-11-05
Fix/improve window proportion after display scale factor change on windows.

raksooo:
This PR improves (almost fixes) the issue where the app window isn't always resized properly after changing display scale factor. I still managed to reproduce the issue once with this fix active but compared to before it happens *way* less. I was able to reproduce it easily and now I've tried a lot and only managed to reproduce it once.

The fix is to resize the window when the display scale changes. I also had to switch to `setContentSize` instead of `setSize`, since `setSize` sometimes used the wrong scale factor. Always settings content size seems like a better fit for us since our UI is designed in relation to these measurements.

I also improved the debug logging for when the window is the wrong size. And we no longer need to set `minHeight`, `minWidth`, `maxHeight` and `maxWidth` since those doesn't seem to make a difference.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3094)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ Line wrap the file at 100 chars.                                              Th
 - Keep unspecified constraints unchanged in the CLI when providing specific tunnel constraints
   instead of setting them to default values.
 
+### Fixed
+#### Windows
+- Fix app size after changing display scale.
+
 
 ## [2021.6-beta1] - 2021-11-03
 ### Added
--- a/gui/src/main/index.ts
+++ b/gui/src/main/index.ts
@@ -1693,45 +1693,13 @@ class ApplicationMain {
     }
   }
 
-  // On both Linux and Windows the app height is applied incorrectly:
-  // https://github.com/electron/electron/issues/28777
-  private getContentHeight(): number {
-    // The height we want achieve.
-    const contentHeight = 568;
-
-    switch (process.platform) {
-      case 'darwin': {
-        // The size of transparent area around arrow on macOS.
-        const headerBarArrowHeight = 12;
-
-        return this.guiSettings.unpinnedWindow
-          ? contentHeight
-          : contentHeight + headerBarArrowHeight;
-      }
-      case 'win32':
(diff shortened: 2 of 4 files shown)


pronebird (commented):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3094)* status: 2 of 4 files reviewed, all discussions resolved

___
*[gui/src/main/window-controller.ts, line 277 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3094#-MoYVeOA583umqq3wJwC:-MoYVeOB4CcM-Vz44vqM:btwqyu3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c2890e52fd5702786cf37d4eafaf0ddaa8a40788/gui/src/main/window-controller.ts#L277)):*
> ```XML
>     }
> 
>     // On linux, the window won't be properly rescaled back to it's original
> ```

FYI On Linux and Windows I suppose?

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3094)* status: 1 of 4 files reviewed, all discussions resolved (waiting on @pronebird)

___
*[gui/src/main/window-controller.ts, line 277 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3094#-MoYVeOA583umqq3wJwC:-MoZ6jd826FVMGmTj7bp:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c2890e52fd5702786cf37d4eafaf0ddaa8a40788/gui/src/main/window-controller.ts#L277)):*

FYI On Linux and Windows I suppose?
</blockquote></details>

Done.

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3094#-:-MocEMGqAAsglmcMr8k7:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3107, opened 2021-11-11
Fix Windows double tray icon on update

raksooo:
This PR fixes the issue where the tray icon of the app running before upgrade remains until hovered. It's fixed by:
* Adding support for quitting the app without disconnecting.
* Quit without disconnecting from `installer.nsh` if the old version is 2021.2 or older.

The kill/quit in `customRemoveFiles` had to be moved to an earlier stage of the uninstallation to prevent the app from changing the tray icon to the disconnected one when it loses connection to the daemon.

The solution for checkign the version in `installer.nsh` has the drawback that it assumes that we won't release more than 9 versions during 2021.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Line wrap the file at 100 chars.                                              Th
 #### Windows
 - Fix app size after changing display scale.
 - Fix daemon not starting if all excluded app paths reside on non-existent/unmounted volumes.
+- Remove tray icon of current running app version when upgrading.
 
 
 ## [2021.6] - 2021-11-17
--- a/dist-assets/windows/installer.nsh
+++ b/dist-assets/windows/installer.nsh
@@ -911,29 +911,32 @@
 # The default behavior may cause the daemon to disconnect.
 #
 !macro customCheckAppRunning
+	push $R0
+	push $R1
 
 	# This must be done here for compatibility with <= 2021.2,
 	# since those versions do not kill the GUI in the uninstaller.
-	# This is fine as long as /f is used.
+	Var /GLOBAL OldVersion
+	ReadRegStr $OldVersion HKLM "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\${APP_GUID}" "DisplayVersion"
+	StrCpy $R0 $OldVersion 4 # Major version
+	StrCpy $R1 $OldVersion 1 5 # Minor version
+	${If} $R0 > 2021
+		Goto customCheckAppRunning_skip_kill
+	${OrIf} $R0 == 2021
+		${If} $R1 > 2
+			Goto customCheckAppRunning_skip_kill
+		${EndIf}
+	${EndIf}
 
-	${KillGui}
-
-!macroend
-
-#
-# KillGui
-#
-# Kill "Mullvad VPN.exe" if it is running. Killing without /f may cause the daemon to disconnect.
-#
-!macro KillGui
-
+	# Killing without /f will likely cause the daemon to disconnect.
 	nsExec::Exec `taskkill /f /t /im "${APP_EXECUTABLE_FILENAME}"` $R0
-
 	Sleep 500
 
-!macroend
+	customCheckAppRunning_skip_kill:
+	pop $R1
+	pop $R0
 
-!define KillGui '!insertmacro "KillGui"'
+!macroend
 
 #
 # customInstall
@@ -1231,6 +1234,10 @@
 
 	Pop $FullUninstall
 
+	nsExec::Exec '"$INSTDIR\Mullvad VPN.exe" --quit-without-disconnect' $0
+	Sleep 500
+	nsExec::Exec `taskkill /f /t /im "${APP_EXECUTABLE_FILENAME}"` $0
+
 	${If} $FullUninstall == 0
 		# Save the target tunnel state if we're upgrading
 		nsExec::ExecToStack '"$TEMP\mullvad-setup.exe" prepare-restart'
@@ -1250,8 +1257,6 @@
 	${If} $R0 != 0
 		Goto customRemoveFiles_abort
 	${EndIf}
-
-	${KillGui}
 
 	# Remove application files
 	log::Log "Deleting $INSTDIR"
--- a/gui/src/main/index.ts
+++ b/gui/src/main/index.ts
@@ -98,6 +98,8 @@ const SANDBOX_DISABLED = app.commandLine.hasSwitch('no-sandbox');
 
 const ALLOWED_PERMISSIONS = ['clipboard-sanitized-write'];
 
+const QUIT_WITHOUT_DISCONNECT_FLAG = '--quit-without-disconnect';
+
 enum AppQuitStage {
   unready,
   initiated,
@@ -245,6 +247,8 @@ class ApplicationMain {
 
   private macOsScrollbarVisibility?: MacOsScrollbarVisibility;
 
+  private quitWithoutDisconnect = false;
+
   public run() {
     // Remove window animations to combat window flickering when opening window. Can be removed when
     // this issue has been resolved: https://github.com/electron/electron/issues/12130
@@ -254,9 +258,15 @@ class ApplicationMain {
 
     this.overrideAppPaths();
 
-    if (this.ensureSingleInstance()) {
+    // This ensures that only a single instance is running at the same time, but also exits if
+    // there's no already running instance when the quit without disconnect flag is supplied.
+    if (!app.requestSingleInstanceLock() || process.argv.includes(QUIT_WITHOUT_DISCONNECT_FLAG)) {
+      this.quitWithoutDisconnect = true;
+      app.quit();
       return;
     }
+
+    this.addSecondInstanceEventHandler();
 
     this.initLogging();
 
@@ -287,18 +297,17 @@ class ApplicationMain {
     app.on('before-quit', this.onBeforeQuit);
   }
 
-  private ensureSingleInstance() {
-    if (app.requestSingleInstanceLock()) {
-      app.on('second-instance', (_event, _commandLine, _workingDirectory) => {
-        if (this.windowController) {
-          this.windowController.show();
-        }
-      });
-      return false;
-    } else {
-      app.quit();
-      return true;
-    }
+  private addSecondInstanceEventHandler() {
+    app.on('second-instance', (_event, argv, _workingDirectory) => {
+      if (argv.includes(QUIT_WITHOUT_DISCONNECT_FLAG)) {
+        // Quit if another instance is started with the quit without disconnect flag.
+        this.quitWithoutDisconnect = true;
+        app.quit();
+      } else if (this.windowController) {
+        // If no action was provided to the new instance the window is opened.
+        this.windowController.show();
+      }
+    });
   }
 
   private overrideAppPaths() {
@@ -377,16 +386,20 @@ class ApplicationMain {
   };
 
   private async prepareToQuit() {
-    if (this.connectedToDaemon) {
-      try {
-        await this.daemonRpc.disconnectTunnel();
-        log.info('Disconnected the tunnel');
-      } catch (e) {
-        const error = e as Error;
-        log.error(`Failed to disconnect the tunnel: ${error.message}`);
-      }
+    if (this.quitWithoutDisconnect) {
+      log.info('Not disconnecting tunnel on quit');
     } else {
-      log.info('Cannot close the tunnel because there is no active connection to daemon.');
+      if (this.connectedToDaemon) {
+        try {
+          await this.daemonRpc.disconnectTunnel();
+          log.info('Disconnected the tunnel');
+        } catch (e) {
+          const error = e as Error;
+          log.error(`Failed to disconnect the tunnel: ${error.message}`);
+        }
+      } else {
+        log.info('Cannot close the tunnel because there is no active connection to daemon.');
+      }
     }
 
     // Unsubscribe the event handler


dlon (changes requested):
> it assumes that we won't release more than 9 versions during 2021

I think we'll be fine. :stuck_out_tongue: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-:-MoEHPd28c239CmRQCd2:b12qvtv)* status: 2 of 3 files reviewed, 6 unresolved discussions (waiting on @raksooo)

___
*[dist-assets/windows/installer.nsh, line 919 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoEEQqaBjvtdkHzuoOC:-MoEEQqaBjvtdkHzuoOD:b7k9em4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L919)):*
> ```NSIS
> 	Var /GLOBAL OldVersion
> 	ReadRegStr $OldVersion HKLM "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\${APP_GUID}" "DisplayVersion"
> 	StrCpy $R0 $OldVersion 4 # Major version
> ```

Because of insane NSIS/electron-builder things, we always push any registers to use (`$R0` and `$R1` in this case) to the stack at the start of a macro and pop them before leaving it. Just in case they already contained some value that is used later on.
___
*[dist-assets/windows/installer.nsh, line 920 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoEKUB30W3hrovIwdcX:-MoEKUB30W3hrovIwdcY:b-7ifurd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L920)):*
> ```NSIS
> 	ReadRegStr $OldVersion HKLM "SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\${APP_GUID}" "DisplayVersion"
> 	StrCpy $R0 $OldVersion 4 # Major version
> 	StrCpy $R1 $OldVersion 1 5 # Minor version
> ```

I'm not sure what happens if the offset is out of bounds. I suspect `$R1` just ends up as an empty string. That's the case if the string to be copied is shorter than the length, at least.
___
*[dist-assets/windows/installer.nsh, line 921 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoEMpe5AfXNCGKeCRP_:-MoEMpe5AfXNCGKeCRPa:b9761cv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L921)):*
> ```NSIS
> 	StrCpy $R0 $OldVersion 4 # Major version
> 	StrCpy $R1 $OldVersion 1 5 # Minor version
> 	${If} $R0 > 2021
> ```

What happens to the comparisons if the registers don't contain integers (e.g. if `DisplayVersion` does not exist and `$OldVersion` is just an empty string)? I *think* they will just evaluate to false and cause `taskkill` to run, which does seem correct.
___
*[dist-assets/windows/installer.nsh, line 933 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoED6aXBQ5rCYK0B-l4:-MoED6aXBQ5rCYK0B-l5:b-i7njdw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L933)):*
> ```NSIS
> 	Sleep 500
> 
> 	skip_kill:
> ```

Let's use the same label naming convention here as we do elsewhere, i.e. prefix this with `customCheckAppRunning_`. For some stupid reason--I think because these "functions" are actually macros--the labels are global.
___
*[dist-assets/windows/installer.nsh, line 1231 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoEGOtT0cfXdx5WB33N:-MoEGOtT0cfXdx5WB33O:b3iryhh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L1231)):*
> ```NSIS
> 	${EndIf}
> 
> 	nsExec::Exec '"$INSTDIR\Mullvad VPN.exe" --quit-without-disconnect' $R1
> ```

Same thing applies here as above about `$R1`. But you could just use `$0` to store the discarded value instead of `$R1`.
___
*[dist-assets/windows/installer.nsh, line 1234 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoEG7ua200qQN-nsYcO:-MoEG7ua200qQN-nsYcP:bv6fksf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L1234)):*
> ```NSIS
> 	Sleep 500
> 
> 	Pop $FullUninstall
> ```

It feels safer and simpler to pop this before the `Exec` call. But it should be fine as `nsExec::Exec` isn't supposed to mess with the stack.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[dist-assets/windows/installer.nsh, line 919 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoEEQqaBjvtdkHzuoOC:-MoF1T2N7OS480R25u7_:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L919)):*

Because of insane NSIS/electron-builder things, we always push any registers to use (`$R0` and `$R1` in this case) to the stack at the start of a macro and pop them before leaving it. Just in case they already contained some value that is used later on.
</blockquote></details>

Done.
___
*[dist-assets/windows/installer.nsh, line 920 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoEKUB30W3hrovIwdcX:-MoF1_6xEMi87dfprKD7:b-3frkpn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L920)):*

I'm not sure what happens if the offset is out of bounds. I suspect `$R1` just ends up as an empty string. That's the case if the string to be copied is shorter than the length, at least.
</blockquote></details>

I haven't found any documentation describing this scenario but when testing it I got an empty string.
___
*[dist-assets/windows/installer.nsh, line 921 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoEMpe5AfXNCGKeCRP_:-MoF280h3sU6ylz-O0GF:b-za06ve) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L921)):*

What happens to the comparisons if the registers don't contain integers (e.g. if `DisplayVersion` does not exist and `$OldVersion` is just an empty string)? I *think* they will just evaluate to false and cause `taskkill` to run, which does seem correct.
</blockquote></details>

That's what I found while testing.
___
*[dist-assets/windows/installer.nsh, line 933 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoED6aXBQ5rCYK0B-l4:-MoF2Q-t-LeSaZgd4XFT:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L933)):*

Let's use the same label naming convention here as we do elsewhere, i.e. prefix this with `customCheckAppRunning_`. For some stupid reason--I think because these "functions" are actually macros--the labels are global.
</blockquote></details>

Done.
___
*[dist-assets/windows/installer.nsh, line 1231 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoEGOtT0cfXdx5WB33N:-MoF2UG056_zggGRnV4s:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L1231)):*

Same thing applies here as above about `$R1`. But you could just use `$0` to store the discarded value instead of `$R1`.
</blockquote></details>

Done.
___
*[dist-assets/windows/installer.nsh, line 1234 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MoEG7ua200qQN-nsYcO:-MoF2WVo6wuqzim6nN7O:b-1wglij) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6a74d5ea4aa4094abc28ca6cb2bfdd90c3846624/dist-assets/windows/installer.nsh#L1234)):*

It feels safer and simpler to pop this before the `Exec` call. But it should be fine as `nsExec::Exec` isn't supposed to mess with the stack.
</blockquote></details>

Yeah, I realized this too and when testing it more it turned out that it's messing with the stack. I've now moved it.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: 2 of 3 files reviewed, all discussions resolved

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[gui/src/main/index.ts, line 306 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-Moc_xNa3cvoKJuRYLr7:-Moc_xNa3cvoKJuRYLr8:bv7sbn1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L306)):*
> ```XML
>       // Quit if this instance is called with arguments that should be passed to a running one when
>       // there is none.
>       if (Object.values(SecondInstanceActions).some((action) => process.argv.includes(action))) {
> ```

Not sure why we need to extract object values when we only respect one flag. This could be simplified as:

```
process.argv.includes("--quit-without-disconnect")
```

Overall I think the entire branch of code can be simplified even further:

```
const QUIT_WITHOUT_DISCONNECT_FLAG = "--quit-without-disconnect"

if ( !app.requestSingleInstanceLock() || process.argv.includes(QUIT_WITHOUT_DISCONNECT_FLAG) ) {
   app.quit()
   return false
}

app.on('second-instance', (event, argv, workdir) => {
  if ( argv.includes(QUIT_WITHOUT_DISCONNECT_FLAG) ) {
    // raise the flag to avoid disconnecting the tunnel during the 'quit' event, somewhere in `onBeforeQuit`.
    this.preventTunnelDisconneectOnQuit = true

    app.quit()
  } else {
    this.windowController?.show()
  }
})

return true
```

pronebird (changes requested):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: all files reviewed, 2 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/index.ts, line 111 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MocevBr6RuEK-6n59LM:-MocevBr6RuEK-6n59LN:b8di236) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L111)):*
> ```XML
> }
> 
> enum InstanceType {
> ```

We don't really need an extra entity in my opinion. `boolean` should work just fine to tell the app to stop initialization.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: all files reviewed, 2 unresolved discussions (waiting on @pronebird and @raksooo)

___
*[gui/src/main/index.ts, line 111 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MocevBr6RuEK-6n59LM:-ModU6MEAKVc-2GMhyMl:b3j9p0d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L111)):*

We don't really need an extra entity in my opinion. `boolean` should work just fine to tell the app to stop initialization.
</blockquote></details>

My intention is to expand the usage of this by adding more flags in the future. An idea I have is to add a flag for reloading the renderer instead of using `SIGUSR2` to make it work on Windows as well. Although, I don't know how well this will work yet so I can't say for certain that there will be other usecases. I could simplify it now and change it to something like this when it's actually needed.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: all files reviewed, 2 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/index.ts, line 111 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MocevBr6RuEK-6n59LM:-MohlB4rC1vJ1zAziUjA:blcpbdg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L111)):*

My intention is to expand the usage of this by adding more flags in the future. An idea I have is to add a flag for reloading the renderer instead of using `SIGUSR2` to make it work on Windows as well. Although, I don't know how well this will work yet so I can't say for certain that there will be other usecases. I could simplify it now and change it to something like this when it's actually needed.
</blockquote></details>

I think it's a little premature at this point.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: all files reviewed, 3 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/index.ts, line 313 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-Moi4tBb6sUy-Bkbwn-X:-Moi4tBb6sUy-Bkbwn-Y:b-cvb8ss) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L313)):*
> ```XML
>         app.on('second-instance', (_event, argv, _workingDirectory) => {
>           if (argv.includes(SecondInstanceActions.quitWithoutDisconnect)) {
>             void this.onBeforeQuit(undefined, false);
> ```

FYI This is ugly. 
___
*[gui/src/main/index.ts, line 323 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-Moi5FNhDNbO6VeQL1xC:-Moi5FNhDNbO6VeQL1xD:blijv0v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L323)):*
> ```XML
>     } else {
>       // Quit if there's already an instance running.
>       app.quit();
> ```

It occurred to me that the secondary instance should make sure that it does not attempt to disconnect the tunnel before it quits. OR does it simply relay on the fact that it didn't connect the daemon yet at this point?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: all files reviewed, 3 unresolved discussions (waiting on @raksooo)

___
*[gui/src/main/index.ts, line 381 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-Moi6AKN5Xj3WtFcN-td:-Moi6AKN5Xj3WtFcN-te:b-z19izp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L381)):*
> ```XML
>     switch (this.quitStage) {
>       case AppQuitStage.unready:
>         // postpone the app shutdown
> ```

FYI We could rewrite to make the termination a little faster (without a need for an extra call to `onBeforeQuit`:

```
if disconnectTunnel {
   event?.preventDefault()
   this.quitStage = AppQuitStage.initiated

   await this.prepareToQuit(disconnectTunnel);

   this.quitStage = AppQuitStage.ready
   app.quit()
} else {
    // log that the tunnel won't be disconnected and do nothing else...
    // no need for app.quit() again.
    this.quitStage = AppQuitStage.ready
}
break
```

This is just a suggestion but I realize that it's quite convenient to have all logging related to disconnecting the tunnel within `prepareToQuit()`

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @dlon and @pronebird)

___
*[gui/src/main/index.ts, line 111 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-MocevBr6RuEK-6n59LM:-MoiRHHO3YGAcXzR2icU:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L111)):*

I think it's a little premature at this point.
</blockquote></details>

Done.
___
*[gui/src/main/index.ts, line 306 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-Moc_xNa3cvoKJuRYLr7:-MoiRh5H4nOChi2o3uKI:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L306)):*

Not sure why we need to extract object values when we only respect one flag. This could be simplified as:

```
process.argv.includes("--quit-without-disconnect")
```

Overall I think the entire branch of code can be simplified even further:

```
const QUIT_WITHOUT_DISCONNECT_FLAG = "--quit-without-disconnect"

if ( !app.requestSingleInstanceLock() || process.argv.includes(QUIT_WITHOUT_DISCONNECT_FLAG) ) {
   app.quit()
   return false
}

app.on('second-instance', (event, argv, workdir) => {
  if ( argv.includes(QUIT_WITHOUT_DISCONNECT_FLAG) ) {
    // raise the flag to avoid disconnecting the tunnel during the 'quit' event, somewhere in `onBeforeQuit`.
    this.preventTunnelDisconneectOnQuit = true

    app.quit()
  } else {
    this.windowController?.show()
  }
})

return true
```
</blockquote></details>

Done.
___
*[gui/src/main/index.ts, line 313 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-Moi4tBb6sUy-Bkbwn-X:-MoiRkGE1j-oqKfM8wL_:bb83zkh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L313)):*

FYI This is ugly. 
</blockquote></details>

I agree that it's ugly passing `undefined` but also feels a bit unnecessary to add an instance variable. But I'm fine with an instance variable as well so I've changed it to that.
___
*[gui/src/main/index.ts, line 323 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-Moi5FNhDNbO6VeQL1xC:-MoiRxBWCwehijqiTXTi:b-i84ip) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L323)):*
> OR does it simply relay on the fact that it didn't connect the daemon yet at this point?

Yes, but I agree it would be nice to make it more clear in code. Done.
___
*[gui/src/main/index.ts, line 381 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-Moi6AKN5Xj3WtFcN-td:-MoiSsDJ38Y7HsyBvKSt:b-5lck2w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L381)):*

FYI We could rewrite to make the termination a little faster (without a need for an extra call to `onBeforeQuit`:

```
if disconnectTunnel {
   event?.preventDefault()
   this.quitStage = AppQuitStage.initiated

   await this.prepareToQuit(disconnectTunnel);

   this.quitStage = AppQuitStage.ready
   app.quit()
} else {
    // log that the tunnel won't be disconnected and do nothing else...
    // no need for app.quit() again.
    this.quitStage = AppQuitStage.ready
}
break
```

This is just a suggestion but I realize that it's quite convenient to have all logging related to disconnecting the tunnel within `prepareToQuit()`
</blockquote></details>

We do a lot more than disconnect in `prepareToQuit` though, and everything in `prepareToQuit` should be quarded by if-statements.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @dlon and @pronebird)

___
*[gui/src/main/index.ts, line 323 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-Moi5FNhDNbO6VeQL1xC:-MoiUzNp9VFbvizBLX8p:bzfq0kt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L323)):*

> OR does it simply relay on the fact that it didn't connect the daemon yet at this point?
Yes, but I agree it would be nice to make it more clear in code. Done.
</blockquote></details>

My reply wasn't supposed to be included in the quote :laughing:

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @dlon and @pronebird)

___
*[gui/src/main/index.ts, line 381 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-Moi6AKN5Xj3WtFcN-td:-MomNU4oCpxNKn7k5e24:b-cqp8yf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6384c4edd0c4fb07454e296e288d7a3f4b968d6e/gui/src/main/index.ts#L381)):*

We do a lot more than disconnect in `prepareToQuit` though, and everything in `prepareToQuit` should be quarded by if-statements.
</blockquote></details>

Right sorry I guess reviewable messed up with me. I had to expand the code and I see that we have plenty more stuff going on in there. My bad.

pronebird (approved):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107#-:-MomQApi0VDJMzqAQ6YH:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (approved):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3107)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3108, opened 2021-11-11
Keep unspecified constraints unchanged in the CLI when providing protocol-specific constraints

dlon:
This is a tweak to make the CLI more intuitive. Previously, setting protocol-specific constraints would clear/set to default any options that weren't provided. E.g., the second command would reset the port and entry locations (to `any` and `none`, respectively) in this example:

```
mullvad relay set tunnel wireguard --protocol udp --port 1234 --entry-location se
mullvad relay set tunnel wireguard --protocol tcp
```

The updated behavior in this PR is to leave those settings alone and only change the transport protocol constraint (or whatever constraints are provided).

To avoid modifying the proto file/gRPC service, the old constraints are simply obtained from the gRPC server.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3108)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ Line wrap the file at 100 chars.                                              Th
 
 
 ## [Unreleased]
+### Changed
+- Keep unspecified constraints unchanged in the CLI when providing specific tunnel constraints
+  instead of setting them to default values.
 
 
 ## [2021.6-beta1] - 2021-11-03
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3108#-:-MoExJ3y4QMIqFZyax9d:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3118, opened 2021-11-17
Upgrade tokio due to RUSTSEC-2021-0124

faern:
RUSTSEC-2021-0124 was just submitted. Looks like there is a race condition in their `oneshot` implementation. As far as I can see we don't use it. We use `oneshot` channels provided by the `futures` crate. So I think we should be fine. But upgrading to get rid of the audit warning and to not accidentally use this broken channel.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3118)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2635,9 +2635,9 @@ dependencies = [
 
 [[package]]
 name = "tokio"
-version = "1.12.0"
+version = "1.14.0"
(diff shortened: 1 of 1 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3118)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3124, opened 2021-11-18
Kill sslocal if setting up the tunnel monitor fails

dlon:
Previously, `sslocal` would sometimes continue running even after the daemon had been killed. This resulted in the uninstaller failing on Windows. As the process is only explicitly killed in `OpenVpnMonitor::wait`, if anything in `OpenVpnMonitor::start` failed *after* the proxy monitor had started, the process would keep running. To fix this and to make it less likely to occur in the future, the process is now killed when a `ShadowsocksProxyMonitor` is dropped, even if `ShadowsocksProxyMonitorCloseHandle::close()` is never called.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3124)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ Line wrap the file at 100 chars.                                              Th
   instead of setting them to default values.
 
 ### Fixed
+- Always kill `sslocal` if the tunnel monitor fails to start when using bridges.
+
 #### Windows
 - Fix app size after changing display scale.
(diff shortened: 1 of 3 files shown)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3124#-:-MosEzn32G0npuZ6qrN0:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3127, opened 2021-11-19
Remove workaround for TAP adapter alias bug

dlon:
We previously couldn't rely on `FriendlyName` in `IP_ADAPTER_ADDRESSES_LH` to provide the correct alias on Windows 8. This PR removes that workaround since the affected driver is no longer used. Also, the interfaces are now compared using LUIDs instead of aliases, which should work regardless.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3127)


mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3127)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #3062, opened 2021-10-27
Support Wayland natively


faern:
Wayland is more and more replacing X. More and more people want to run a Wayland native desktop. Most Linux GUI programs run well under Wayland these days. This issue is for tracking the Wayland support in the Mullvad VPN app GUI.

## Current status

The latest stable release, `2021.5`, runs on Electron 11. There is no Wayland support in Electron 11, it was introduced in Electron 12.

The `master` branch in this git repository is using Electron 15. So the next release we make will support Wayland. However, you will need to supply some rather long arguments to the app to make it start in native Wayland mode:
```
/opt/Mullvad\ VPN/mullvad-vpn --enable-features=UseOzonePlatform --ozone-platform=wayland
```

## Goal

In a future release we hope to modify `/opt/Mullvad VPN/mullvad-vpn` in such a way that it will automatically detect if Wayland is being used, and then automatically add the arguments needed to run the app in native Wayland mode.

johns2s:
On F35/GNOME 41, using fractional scaling: Works fine, but titlebars missing and cursor way too small, both of which are already being worked on upstream IIRC
https://imgur.com/7WFlRk6

faern:
Thanks for reporting this! Do you have links to the upstream issues?

lkcv:
I believe the missing titlebars is related to this issue: https://gitlab.gnome.org/GNOME/mutter/-/issues/217

johns2s:
> I believe the missing titlebars is related to this issue: https://gitlab.gnome.org/GNOME/mutter/-/issues/217
> GNOME Wayland doesn't support server-side decorations, so if client-side decorations aren't implemented in an application running in that environment, the titlebar will be missing.

Yeah, looks like support is being worked on in election itself though. https://github.com/electron/electron/pull/29618

pinballmachine:
I am having some small and strange difficulties using the Mullvad GUI app with Wayland on KDE (KDE Plasma version 5.23.5, KDE Frameworks version 5.90.0, Qt version 5.15.2) on OpenSUSE Tumbleweed (rolling release up to date, kernel 5.16)
I know Wayland is not officially supported yet, and on top of that I am running the .rpm on OpenSUSE instead of Fedora, but here goes:

With the titlebar on the bottom of the screen, as is standard for most systems and people, clicking the systray icon to open the app works fine.
[...]

PR #3145, opened 2021-11-24
Add electron flags to run wayland native if in a wayland session.

pmkap:
Check if in a wayland session and add electron flags to run wayland native as explained here: #3062

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3145)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@ Line wrap the file at 100 chars.                                              Th
   default 1500. The 1500 is still the fallback if this for some reason fails. This may stop
   fragmentation.
 
+#### Linux
+- GUI: Add electron flags to run Wayland native if in a compositor/desktop known to work well
+
 ### Changed
 - Reject invalid WireGuard ports in the CLI.
 - Reorganize settings into more logical categories.
--- a/dist-assets/linux/mullvad-gui-launcher.sh
+++ b/dist-assets/linux/mullvad-gui-launcher.sh
@@ -8,5 +8,15 @@ else
     SANDBOX_FLAG=""
 fi
 
+SUPPORTED_COMPOSITORS="sway river"
+if [ "${XDG_SESSION_TYPE:-""}"  = "wayland" ] && \
+    echo " $SUPPORTED_COMPOSITORS " | \
+    grep -qi -e " ${XDG_CURRENT_DESKTOP:-""} " -e " ${XDG_SESSION_DESKTOP:-""} "
+then
+    WAYLAND_FLAGS="--ozone-platform=wayland --enable-features=WaylandWindowDecorations"
+else
+    WAYLAND_FLAGS=""
+fi
+
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-exec "$SCRIPT_DIR/mullvad-gui" $SANDBOX_FLAG "$@"
+exec "$SCRIPT_DIR/mullvad-gui" $SANDBOX_FLAG $WAYLAND_FLAGS "$@"


faern:
This is awesome. Thank you! The main reason why we have not done this ourselves yet is that we want input from users if running the current app works well under Wayland or not. We want to be fairly confident that using the app in native Wayland mode renders it correctly and does not have significant performance impact or any other downside.

Have you tried running our app in native Wayland mode for a decent amount of time? Does it seem to work well? It works fine for me here when I try it! But I know a bunch of other electron apps that behave pretty badly in native Wayland mode. Especially if there are multiple monitors and some of them have scaling set to something other than 1.0. I want these special cases tested before we merge this.

pmkap:
I've only run it for a few hours. So far everything works perfectly but I can come back in a couple weeks or so to confirm.

Also with a second monitor and fractional scaling everything's perfect on my machine (wlroots-based compositor).

faern:
This is the type of stuff we need to find and have fixed before we merge this PR: https://github.com/mullvad/mullvadvpn-app/issues/3062#issuecomment-978260018

faern:
It is now enough to pass only `--ozone-platform=wayland`, since Ozone is enabled by default in newer Chromium. However, this does still not work perfectly under some window managers. So we might still not want to merge this wayland-by-default-PR. But it's getting closer!

faern:
I think the changes looks good! We just need to collect some more user data first to see if this is stable enough to enable by default. For example, I want to know what the solution to #3570 ends up being.

faern:
I tested this under Gnome in Ubuntu 20.04 in wayland mode now. I manually specified the flags that this PR adds. And it did not work well. The window came up in the wrong size and halfway off the screen. It still works way better in X11 mode in gnome sadly. I'm not sure if that's a gnome issue and it just works better in Ubuntu 22.04 or something. Or if there are still issues in Electron that we need to wait for.

faern:
Let's go ahead with this! But in addition to checking `XDG_SESSION_TYPE`, let's also check `XDG_SESSION_DESKTOP` against a predefined list of allowlist session types. If the current session type is **not** in our list we fall back to X. That's the most conservative and safe approach. Then people can report if it works well for them in other WMs.

We know it works well in sway. So that's a good starting point for the allowlist.

pmkap:
This is a great idea.

As far as I know, sway itself does not set `XDG_SESSION_DESKTOP` and I think most users do not set this.

There's also the `XDG_CURRENT_DESKTOP` environment variable which is probably set by more users.

So not sure which variable (or both) to check here.

dist-assets/linux/mullvad-gui-launcher.sh
```diff
@@ -8,5 +8,15 @@ else
     SANDBOX_FLAG=""
 fi
 
+SUPPORTED_COMPOSITORS="sway river"
+if [ "$XDG_SESSION_TYPE" = "wayland" ] && \
```
faern, line 12: If it does not become too cumbersome, our bash coding guidelines prefers `${VARIABLE:-""}` over removing `set -u`. Otherwise it allows unwanted behavior to silently sneak into other parts of the script. https://github.com/mullvad/coding-guidelines/blob/master/bash.md

faern (approved):
More or less good to go IMO. But please keep `set -u`

dist-assets/linux/mullvad-gui-launcher.sh
```diff
@@ -8,5 +8,15 @@ else
     SANDBOX_FLAG=""
 fi
 
+SUPPORTED_COMPOSITORS="sway river"
```
faern, line 11: I think you need spaces at the start and end of this string. Otherwise `" sway "` will never exist in this string.

dist-assets/linux/mullvad-gui-launcher.sh
pmkap: Ah yeah, sure.

dist-assets/linux/mullvad-gui-launcher.sh
pmkap: This is done in the `echo " $SUPPORTED_COMPOSITORS "` line below.

faern: Ah, of course. My own little test did not do that dance. Bash is such a hack :facepalm: This really should be an array. But bash is bash, so OK.

CHANGELOG.md
```diff
@@ -35,6 +35,9 @@ Line wrap the file at 100 chars.                                              Th
   default 1500. The 1500 is still the fallback if this for some reason fails. This may stop
   fragmentation.
 
+#### Linux
+- GUI: Add electron flags to run Wayland native if in a Wayland session.
```
faern, line 39: Please add something along the lines of "in Window managers known to work well".

faern (approved):

faern:
Please rebase this on our latest master branch, then we can merge it.

pmkap:
> Please rebase this on our latest master branch, then we can merge it.

Done :)

faern:
Thank you for the contribution! We'll add more WMs to the list later.

PR #3151, opened 2021-11-26
Remove explicitly passing handles to tokio runtimes in mullvad-rpc

faern:
What do you think about this? I agree it's a bit implicit that some stuff has to run in an async context. But on the other hand, there is a lot of arguments to pass around for basically no reason here...

Some parts of the problem report tool was the only things that called into this code from a non async context. So I slightly refactored that to be an `async fn`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3151)

### DIFF
--- a/mullvad-problem-report/src/lib.rs
+++ b/mullvad-problem-report/src/lib.rs
@@ -268,49 +268,61 @@ pub fn send_problem_report(
             }
         })?,
     );
-    let metadata =
-        ProblemReport::parse_metadata(&report_content).unwrap_or_else(|| metadata::collect());
 
     let runtime = tokio::runtime::Builder::new_multi_thread()
         .worker_threads(2)
         .enable_all()
         .build()
         .map_err(Error::CreateRuntime)?;
+    runtime.block_on(send_problem_report_inner(
+        user_email,
+        user_message,
+        &report_content,
+        cache_dir,
+    ))
+}
 
-    let mut rpc_manager = runtime
-        .block_on(mullvad_rpc::MullvadRpcRuntime::with_cache(
-            None,
-            cache_dir,
-            false,
-            #[cfg(target_os = "android")]
-            None,
-        ))
-        .map_err(Error::CreateRpcClientError)?;
+async fn send_problem_report_inner(
+    user_email: &str,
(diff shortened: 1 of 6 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3151)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[mullvad-rpc/src/lib.rs, line 203 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3151#-MpRNX6R1TXkUfI_kSD6:-MpRNX6R1TXkUfI_kSD7:bcp41f2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/982df7252fdcffac30e6510df6b198cbab2dd4b3/mullvad-rpc/src/lib.rs#L203)):*
> ```XML
>         );
>         let handle = service.handle();
>         tokio::spawn(service.into_future());
> ```

I'd prefer the explicit handle, because it's hard to know/keep track of whether `mullvad_rest_handle` and `rest_handle` are only ever called in async contexts. It's probably the case, though.
___
*[mullvad-rpc/src/rest.rs, line 285 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3151#-MpRM2bC7zo9lGhRQesz:-MpRM2bD6I-MK91dt0VL:bffp30p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/982df7252fdcffac30e6510df6b198cbab2dd4b3/mullvad-rpc/src/rest.rs#L285)):*
> ```XML
> 
>     /// Spawns a future on the RPC runtime.
>     pub fn spawn<T: Send + 'static>(&self, future: impl Future<Output = T> + Send + 'static) {
> ```

This method seems pointless. Can it be removed?

pinkisemils (changes requested):
> Reviewed 5 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3151)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

PR #3164, opened 2021-12-01
Make the initial allowed endpoint argument non-optional

dlon:
Previously, `FirewallArguments` included an optional `allowed_endpoint` field as well as a boolean `initialize_blocked`. In reality, `allowed_endpoint` will never be `None` if `initialize_blocked` is `true`, and it will only ever be `None` if `initialize_blocked` is `false`. Moreover, setting the firewall policy to `connecting` or `blocked` after init involves a non-optional `allowed_endpoint` argument. So this PR gets rid of the "invalid" combinations.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3164)

### DIFF
--- a/mullvad-setup/src/main.rs
+++ b/mullvad-setup/src/main.rs
@@ -4,7 +4,7 @@ use mullvad_rpc::MullvadRpcRuntime;
 use mullvad_types::version::ParsedAppVersion;
 use std::{path::PathBuf, process, time::Duration};
 use talpid_core::{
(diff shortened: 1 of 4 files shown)


mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3164)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3000, opened 2021-10-01
Update to electron 15

raksooo:
This PR updates Electron from 11.4.10 to 15.0.0. The update to 12 was held back by a change in Electron/Chromium that caused the map to become blurry during transitions. I've finally found a working workaround for this. A few other changes was required as well:
* Set map CSS zoom to `100,01%` to prevent blurriness during transitions
* Set the size of `TransitionContainer` to prevent it from growing and becoming unscrollable
* Wait until switch input has finished it's transition before calling `onChange`. This is due to the transition being glitchy when communicating with the daemon at the same time.
* Writing to clipboard is now a permission which needs to be allowed.

Improvements thanks to this version:
* Removed workaround for "uncaught error" in error messages passed between the processes
* We now deny `window.open()`
* Since Node 16.5 contains a function for generating UUIDs we no longer need the `uuid` module
* The sandbox now works on Ubuntu 21.10 again.

Major version changelogs:
https://www.electronjs.org/releases/stable?page=9#12.0.0
https://www.electronjs.org/releases/stable?page=5#13.0.0
https://www.electronjs.org/releases/stable?page=1#14.0.0
https://www.electronjs.org/releases/stable?page=1#15.0.0

Git checklist:

* [x] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header.
* [x] Check that commits follow the [Mullvad coding guidelines](https://github.com/mullvad/coding-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3000)

PR #3170, opened 2021-12-03
Improve switch confirmation

raksooo:
This PR changes the behavior of the block when disconnected-switch which previously showed a confirmation dialog after the setting had changed and now it shows it before changing it instead. This moves part of the logic to the `Switch` component and makes it reusable.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3170)

### DIFF
--- a/gui/src/renderer/components/Switch.tsx
+++ b/gui/src/renderer/components/Switch.tsx
@@ -11,7 +11,7 @@ interface IProps {
   onChange?: (isOn: boolean) => void;
   className?: string;
   disabled?: boolean;
-  forwardedRef?: React.Ref<HTMLDivElement>;
+  innerRef?: React.Ref<HTMLDivElement>;
 }
 
 interface IState {
@@ -103,9 +103,13 @@ export default class Switch extends React.PureComponent<IProps, IState> {
     );
   }
 
+  public setOn(isOn: boolean) {
+    this.setState({ isOn });
+  }
+
   private refCallback = (element: HTMLDivElement | null) => {
     assignToRef(element, this.containerRef);
-    assignToRef(element, this.props.forwardedRef);
+    assignToRef(element, this.props.innerRef);
   };
 
   private onTransitionEnd = (event: React.TransitionEvent) => {
--- a/gui/src/renderer/components/AdvancedSettings.tsx
+++ b/gui/src/renderer/components/AdvancedSettings.tsx
@@ -25,6 +25,7 @@ import {
 } from './NavigationBar';
 import { ISelectorItem } from './cell/Selector';
 import SettingsHeader, { HeaderTitle } from './SettingsHeader';
+import Switch from './Switch';
 
 type OptionalTunnelProtocol = TunnelProtocol | undefined;
 
@@ -50,6 +51,8 @@ export default class AdvancedSettings extends React.Component<IProps, IState> {
   public state = {
     showConfirmBlockWhenDisconnectedAlert: false,
   };
+
+  private blockWhenDisconnectedRef = React.createRef<Switch>();
 
   public render() {
     const hasWireguardKey = this.props.wireguardKeyState.type === 'key-set';
@@ -118,6 +121,7 @@ export default class AdvancedSettings extends React.Component<IProps, IState> {
                     </AriaLabel>
                     <AriaInput>
                       <Cell.Switch
+                        ref={this.blockWhenDisconnectedRef}
                         isOn={this.props.blockWhenDisconnected}
                         onChange={this.setBlockWhenDisconnected}
                       />
@@ -257,7 +261,6 @@ export default class AdvancedSettings extends React.Component<IProps, IState> {
 
   private setBlockWhenDisconnected = (newValue: boolean) => {
     if (newValue) {
-      this.props.setBlockWhenDisconnected(true);
       this.setState({ showConfirmBlockWhenDisconnectedAlert: true });
     } else {
       this.props.setBlockWhenDisconnected(false);
@@ -265,13 +268,13 @@ export default class AdvancedSettings extends React.Component<IProps, IState> {
   };
 
   private hideConfirmBlockWhenDisconnectedAlert = () => {
-    this.props.setBlockWhenDisconnected(false);
     this.setState({ showConfirmBlockWhenDisconnectedAlert: false });
+    this.blockWhenDisconnectedRef.current?.setOn(this.props.blockWhenDisconnected);
   };
 
   private confirmEnableBlockWhenDisconnected = () => {
-    this.props.setBlockWhenDisconnected(true);
     this.setState({ showConfirmBlockWhenDisconnectedAlert: false });
+    this.props.setBlockWhenDisconnected(true);
   };
 
   private onSelectTunnelProtocol = (protocol?: TunnelProtocol) => {
--- a/gui/src/renderer/components/CustomDnsSettings.tsx
+++ b/gui/src/renderer/components/CustomDnsSettings.tsx
@@ -194,7 +194,7 @@ export default function CustomDnsSettings() {
           </AriaLabel>
           <AriaInput>
             <Cell.Switch
-              ref={switchRef}
+              innerRef={switchRef}
               isOn={dns.state === 'custom' || inputVisible}
               onChange={setCustomDnsEnabled}
             />
--- a/gui/src/renderer/components/cell/Input.tsx
+++ b/gui/src/renderer/components/cell/Input.tsx
@@ -8,10 +8,10 @@ import ImageView from '../ImageView';
 
 export const Switch = React.forwardRef(function SwitchT(
   props: StandaloneSwitch['props'],
-  ref: React.Ref<HTMLDivElement>,
+  ref: React.Ref<StandaloneSwitch>,
 ) {
   const disabled = useContext(CellDisabledContext);
-  return <StandaloneSwitch forwardedRef={ref} disabled={disabled} {...props} />;
+  return <StandaloneSwitch ref={ref} disabled={disabled} {...props} />;
 });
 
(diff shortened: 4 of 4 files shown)


gui/src/renderer/components/Switch.tsx
```diff
@@ -156,7 +163,14 @@ export default class Switch extends React.PureComponent<IProps, IState> {
...
 
     if (this.props.isOn !== this.state.isOn) {
-      this.notify();
+      if (
+        this.props.confirmation === undefined ||
+        (await this.props.confirmation(this.state.isOn))
+      ) {
+        this.notify();
+      } else {
+        this.setState({ isOn: !this.state.isOn });
```
MrChocolatine, line 172: If I may, updating state based on the current state should be done with a function.

```suggestion
        this.setState((prevState) => ({ isOn: !prevState.isOn }));
```

gui/src/renderer/components/Switch.tsx
```diff
@@ -114,13 +115,19 @@ export default class Switch extends React.PureComponent<IProps, IState> {
...
 
     if (!this.changedDuringPan) {
-      this.setState((state) => ({ isOn: !state.isOn, notifyOnTransitionEnd: true }));
+      const newIsOn = !this.state.isOn;
+
+      if (this.props.confirmation !== undefined && !(await this.props.confirmation(newIsOn))) {
+        return;
+      }
+
+      this.setState({ isOn: newIsOn, notifyOnTransitionEnd: true });
```
MrChocolatine, line 130: If I may, updating state based on the current state should be done with a function.

```suggestion
      this.setState((prevState) => ({
        isOn: !prevState.isOn,
        notifyOnTransitionEnd: true,
      }));
```

gui/src/renderer/components/Switch.tsx
raksooo: Thanks for your feedback. This has been changed to `this.props.isOn` so this is no longer the case.

gui/src/renderer/components/Switch.tsx
raksooo: In this situation the variable `newIsOn` is also passed to an asynchronous callback for confirmation so the new state must be the value that has been confirmed. The only current use of this within our app is to show a confirmation dialog to the user.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3170)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @MrChocolatine)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3170#-MqE9-BJ4YPRopfbpapr:-MqE9-BJ4YPRopfbpaps:b958nf6) (no related file):*
Had another look at that. Although I somewhat understand the intent here to prevent the switch form changing state before confirmation, I don't see a problem with the switch going into "on" state before returning to "off" state upon user confirmation.

That kind of behaviour can be managed at the higher level parent component that propagates the `props` and `state` based on what's happening with the user input, for example:


```
class MyComponent extends React.Component {

  init(props) {
    this.state = { isOn: props.isOn, userConfirmation: false }
  }

  componentDidUpdate(oldProps) {
    this.setState((state) => { isOn: state.userConfirmation ? state.isOn : this.props.isOn })
  }

  onChange = (newValue) => {
    this.setState({ userConfirmation: true, isOn: newValue })
  }

  onConfirmChange = (isConfirmed) => {
    if (isConfirmed) {
      this.setState({ userConfirmation: false })

      // persist new value
    } else {
      this.setState({ userConfirmation: false, isOn: this.props.isOn })
    }
  }

  render() {
    return (
      <div>
        {this.state.userConfirmation && <MyUserConfirmation onConfirm={this.onConfirmChange}>}
        <Switch isOn={this.state.isOn} onChange={this.onChange}>
      </div>
    )
  }

}
```

Note that the state is persisted only after user confirmation, otherwise the change is discarded.

There are many ways to solve it but overall, I think the `Preferences` or any other components should handle the intermediate state of its child components, i.e part of toggles that require confirmation can use `state` and the rest can feed directly from `props`. This feels cleaner to me.

Regardless all of the above, I think the following improvements can be done:

1.  It's more common in HTML world to `return false` or use `event.preventDefault()` from event handler to discard the event. So maybe that would be the way to go here? So for example we could send `props.onChange(newValue)` before updating the state  and then if the handler returns `props.onChange(newValue) === false` then we simply bail in the input handler.
2. Since we're not going with React-native any longer, the drag handling in the `Switch` is arguably not needed. This in a way should simplify the code and remove edge cases associated with the drag input and allow for easier adoption of the solution proposed above.

---

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3170)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @MrChocolatine)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3170#-MqE9-BJ4YPRopfbpapr:-MqELFXR7HlRra8Yjy3a:b-mezmvi) (no related file):*

Had another look at that. Although I somewhat understand the intent here to prevent the switch form changing state before confirmation, I don't see a problem with the switch going into "on" state before returning to "off" state upon user confirmation.

That kind of behaviour can be managed at the higher level parent component that propagates the `props` and `state` based on what's happening with the user input, for example:


```
class MyComponent extends React.Component {

  init(props) {
    this.state = { isOn: props.isOn, userConfirmation: false }
  }

  componentDidUpdate(oldProps) {
    this.setState((state) => { isOn: state.userConfirmation ? state.isOn : this.props.isOn })
  }

  onChange = (newValue) => {
    this.setState({ userConfirmation: true, isOn: newValue })
  }

  onConfirmChange = (isConfirmed) => {
    if (isConfirmed) {
      this.setState({ userConfirmation: false })

      // persist new value
    } else {
      this.setState({ userConfirmation: false, isOn: this.props.isOn })
    }
  }

  render() {
    return (
      <div>
        {this.state.userConfirmation && <MyUserConfirmation onConfirm={this.onConfirmChange}>}
        <Switch isOn={this.state.isOn} onChange={this.onChange}>
      </div>
    )
  }

}
```

Note that the state is persisted only after user confirmation, otherwise the change is discarded.

There are many ways to solve it but overall, I think the `Preferences` or any other components should handle the intermediate state of its child components, i.e part of toggles that require confirmation can use `state` and the rest can feed directly from `props`. This feels cleaner to me.

Regardless all of the above, I think the following improvements can be done:

1.  It's more common in HTML world to `return false` or use `event.preventDefault()` from event handler to discard the event. So maybe that would be the way to go here? So for example we could send `props.onChange(newValue)` before updating the state  and then if the handler returns `props.onChange(newValue) === false` then we simply bail in the input handler.
2. Since we're not going with React-native any longer, the drag handling in the `Switch` is arguably not needed. This in a way should simplify the code and remove edge cases associated with the drag input and allow for easier adoption of the solution proposed above.


</blockquote></details>

In https://github.com/mullvad/mullvadvpn-app/pull/3000 I changed the behavior of the `Switch` component to call `onChange` after the transition had finished due to transition lag introduced when upgrading to Electron 15. This prevents such a solution unfortunately. I could spend the time to understand why that lag is happening but I'm not sure if it's worth the time.

---

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3170)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @MrChocolatine)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3170#-MqE9-BJ4YPRopfbpapr:-MqESoOu4vevtWzpf0MF:bu8zp9e) (no related file):*

In https://github.com/mullvad/mullvadvpn-app/pull/3000 I changed the behavior of the `Switch` component to call `onChange` after the transition had finished due to transition lag introduced when upgrading to Electron 15. This prevents such a solution unfortunately. I could spend the time to understand why that lag is happening but I'm not sure if it's worth the time.
</blockquote></details>

Alright. I think propagating new state on `transitionEnd` makes sense. 

There are ways to tackle the lag though and you don't have to make `Switch` aware of this specific issue.  It's not a responsibility of a `Switch` control to account for platform quirks going as far as RPC calls in the main process.

For example, you can call an instance method directly to update the state of `Switch` component, i.e:

```
const onTransitionEnd = () => {
  // persist state
}

switchRef.setOn(true, onTransitionEnd)
```

 ... or if you use an intermediate state, you can add `onTransitionEnd` to be notified when animation has finished and act there:

```
onChange = () => {
  // request confirmation
  this.requestedConfirmation = true
}

transitionEnd = ()  => {
  if (this.requestedConfirmation) {
      // persist state

     this.requestedConfirmation = false
  }
}

render() {
  <Switch ref={this.ref} isOn={this.state.isOn} onChange={this.onChange} onTransitionEnd={this.transitionEnd}>
}
```

Moreover, in your proposal (PR), the switch can still be dragged into the right-most position, with confirmation triggering only on `mouseUp`. I see inconsistency in this solution and I feel like this can be solved in a better way.

---

pronebird (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3170)* status: all files reviewed, 2 unresolved discussions (waiting on @MrChocolatine)

pronebird (approved):
I like it! :lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3170#-:-MqFOl7X6_7UlwUMW6gl:b-mg7t6l)* status: all files reviewed, 2 unresolved discussions (waiting on @MrChocolatine)

PR #3172, opened 2021-12-03
Refactor api addr override

faern:
Rename `MULLVAD_API_ADDRESS` to `MULLVAD_API_ADDR` to be more like the corresponding `MULLVAD_API_HOST`.

Also doing some refactoring to allow better error handling in case their values are set in the wrong way. This feels better than silently failing. The silent failing makes it harder to figure out if a tester correctly set it or not.

Here I'm going more with `cfg!` rather than `#[cfg` because it allows the compiler/CI to catch errors in the not used path in all cases. If completely compiling out code it's easier to have errors that go undetected. The optimizer will remove all not used cases anyway when it sees `if false {}`

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3172)

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,7 +459,7 @@ echo "org.gradle.jvmargs=-Xmx4608M" >> ~/.gradle/gradle.properties
 
 * `MULLVAD_API_HOST` - Set the hostname to use in API requests. E.g. `api.mullvad.net`.
 
-* `MULLVAD_API_ADDRESS` - Set the IP address and port to use in API requests. E.g. `10.10.1.2:443`.
+* `MULLVAD_API_ADDR` - Set the IP address and port to use in API requests. E.g. `10.10.1.2:443`.
 
 #### Setting environment variable
 - On Windows, one can use `setx` from an elevated shell, like so
--- a/mullvad-rpc/Cargo.toml
+++ b/mullvad-rpc/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2021"
 publish = false
 
 [features]
-# Allow the API server to use to be configured via MULLVAD_API_HOST and MULLVAD_API_ADDRESS.
+# Allow the API server to use to be configured via MULLVAD_API_HOST and MULLVAD_API_ADDR.
 api-override = []
 
 [dependencies]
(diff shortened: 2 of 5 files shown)


dlon (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3172#-:-Mq0At1uAOckYp5Suf-_:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[mullvad-rpc/src/lib.rs, line 170 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3172#-Mq0BNO_2p4tZMb22_x8:-Mq0BNO_2p4tZMb22_x9:bdewcls) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b345cd98c77b2e47b97165df58068609887336c4/mullvad-rpc/src/lib.rs#L170)):*
> ```XML
>         let handle = tokio::runtime::Handle::current();
>         if API.disable_address_rotation {
> ```

Maybe this was misnamed before. It doesn't just disable rotation, but also prevents the address from being read from the files in the cache and resource dirs.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3172)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[mullvad-rpc/src/lib.rs, line 170 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3172#-Mq0BNO_2p4tZMb22_x8:-MqDqtMc8DkucqeCz2X9:b-lyxiuy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b345cd98c77b2e47b97165df58068609887336c4/mullvad-rpc/src/lib.rs#L170)):*

Maybe this was misnamed before. It doesn't just disable rotation, but also prevents the address from being read from the files in the cache and resource dirs.
</blockquote></details>

Renaming to `disable_address_cache`!

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3172)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3183, opened 2021-12-09
Remove unused dependencies found by cargo-udeps

faern:
I ran `cargo udeps` and found we had three unused dependencies. They are still in our dependency tree, so we don't get rid of them. But at least our crates should not specify dependencies we don't use.

I'm considering adding a CI job for `cargo udeps`, but not sure it's worth it given that one must compile it from source, which will take time in CI.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3183)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1464,7 +1464,6 @@ version = "0.1.0"
 dependencies = [
  "chrono",
  "err-derive",
- "filetime",
  "futures",
  "http",
  "hyper",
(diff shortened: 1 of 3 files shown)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3183#-:-Mq_6Du94Mecs1r_StXb:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3186, opened 2021-12-13
Improve unsafe code usage in system service module

dlon:
This attempts to make the `system_service` module conform a bit better to the [coding guidelines](https://github.com/mullvad/coding-guidelines).

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186)

### DIFF
--- a/mullvad-daemon/src/system_service.rs
+++ b/mullvad-daemon/src/system_service.rs
@@ -3,7 +3,7 @@ use mullvad_daemon::{runtime::new_runtime_builder, DaemonShutdownHandle};
 use std::{
     env,
     ffi::{OsStr, OsString},
-    ptr, slice,
+    mem, ptr, slice,
     sync::{
         atomic::{AtomicBool, AtomicUsize, Ordering},
         mpsc, Arc,
@@ -14,11 +14,7 @@ use std::{
 use talpid_types::ErrorExt;
 use winapi::{
     ctypes::c_void,
-    shared::{
-        minwindef::ULONG,
-        ntdef::{LUID, PVOID},
-        ntstatus::STATUS_SUCCESS,
-    },
+    shared::{minwindef::ULONG, ntdef::LUID, ntstatus::STATUS_SUCCESS},
     um::ntlsa::{
         LsaEnumerateLogonSessions, LsaFreeReturnBuffer, LsaGetLogonSessionData,
         SECURITY_LOGON_SESSION_DATA,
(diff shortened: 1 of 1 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[mullvad-daemon/src/system_service.rs, line 472 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186#-MqnPLct2xc58XYiG_Vu:-MqnPLct2xc58XYiG_Vv:bfsadid) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/07b2387f0fffdfd76a363f3ac09129f23e7a0919/mullvad-daemon/src/system_service.rs#L472)):*
> ```XML
>             return false;
>         }
>         // SAFETY: `logon_session_list` is not mutated before being freed.
> ```

I'm not sure exactly [what](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety) needs to be spelled out.

Take this: `The total size len * mem::size_of::<T>() of the slice must be no larger than isize::MAX.` I don't think we have to mention that it the total size can't exceed 8 exabytes. :stuck_out_tongue: 

Regarding alignment and contiguity, I think they can be taken for granted as well?
___
*[mullvad-daemon/src/system_service.rs, line 492 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186#-MqnPaGT0i8G2lzEE5TJ:-MqnPaGT0i8G2lzEE5TK:bpw154k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/07b2387f0fffdfd76a363f3ac09129f23e7a0919/mullvad-daemon/src/system_service.rs#L492)):*
> ```XML
>             }
>         }
>         mem::drop(logons);
> ```

Is it necessary to drop the slice here? I assume so, because of the requirement that the data must not be mutated.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @dlon)

___
*[mullvad-daemon/src/system_service.rs, line 463 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186#-MqnUtslAhUmVlLh9Y_w:-MqnUtslAhUmVlLh9Y_x:b-f1863v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/07b2387f0fffdfd76a363f3ac09129f23e7a0919/mullvad-daemon/src/system_service.rs#L463)):*
> ```XML
>     }
> 
>     fn interactive_session(session_id: u32) -> bool {
> ```

Nit: Maybe rename this `is_interactive_session`?

And it's correct to remove the `unsafe` from the function itself if it's not possible to call it in a way that causes UB/invalid memory access.
___
*[mullvad-daemon/src/system_service.rs, line 472 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186#-MqnPLct2xc58XYiG_Vu:-MqnUL1624RNV2BRgRwM:b-lpz3sm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/07b2387f0fffdfd76a363f3ac09129f23e7a0919/mullvad-daemon/src/system_service.rs#L472)):*

I'm not sure exactly [what](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety) needs to be spelled out.

Take this: `The total size len * mem::size_of::<T>() of the slice must be no larger than isize::MAX.` I don't think we have to mention that it the total size can't exceed 8 exabytes. :stuck_out_tongue: 

Regarding alignment and contiguity, I think they can be taken for granted as well?
</blockquote></details>

I agree that most of the safety stuff outlined by `slice::from_raw_parts` does not need to be reiterated here. It's hard to write good `// SAFETY` docs. One thing we could say is that the pointer and length are valid and that `LsaEnumerateLogonSessions` is responsible for that. But maybe that's just more noise than value. The important part is not document the non obvious unsafe constraints that are being upheld. Here most of it is pretty obvious.

I think your current docs make sense.
___
*[mullvad-daemon/src/system_service.rs, line 485 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186#-MqnYKdJBH-hW15GoZ8T:-MqnYKdJBH-hW15GoZ8U:b9mi1es) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/07b2387f0fffdfd76a363f3ac09129f23e7a0919/mullvad-daemon/src/system_service.rs#L485)):*
> ```XML
>                 continue;
>             }
>             let candidate_correct_session = unsafe { *session_data }.Session == session_id;
> ```

This variable is now unused.. No? I can't see where it's used any longer.
___
*[mullvad-daemon/src/system_service.rs, line 492 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186#-MqnPaGT0i8G2lzEE5TJ:-MqnVSdl9TpIi55qOcvs:b-ownb93) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/07b2387f0fffdfd76a363f3ac09129f23e7a0919/mullvad-daemon/src/system_service.rs#L492)):*

Is it necessary to drop the slice here? I assume so, because of the requirement that the data must not be mutated.
</blockquote></details>

Yes. We should make sure the safe references goes out of scope before the data is freed. :+1:

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/system_service.rs, line 485 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186#-MqnYKdJBH-hW15GoZ8T:-MqncH4-65opLQ96D5eu:bqp8ylg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/07b2387f0fffdfd76a363f3ac09129f23e7a0919/mullvad-daemon/src/system_service.rs#L485)):*

This variable is now unused.. No? I can't see where it's used any longer.
</blockquote></details>

I've added it back. The simpler code was incorrect in the case where there was no matching session.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3186)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

PR #3204, opened 2021-12-16
Fix relay selector with bridge constraints

dlon:
Changes:
* Stop always preferring OpenVPN when the tunnel protocol is set to `any` and bridges are enabled. It doesn't make much sense if WireGuard multihop is enabled especially.
* Enabling bridges no longer resets the tunnel protocol to `any` if it is currently set to `wireguard`. This can currently only occur in the CLI.


pinkisemils (approved):
> Reviewed 4 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3204)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3205, opened 2021-12-16
Increase WFP transaction lock timeout

dlon:
Acquiring the lock appears to fail every now and then, perhaps more so on slow systems. It doesn't seem to be obviously caused by third-party software.

The timeout is increased somewhat here to hopefully improve things. A very long timeout would be a bad idea, because, among other reasons, firewall state changes block the tunnel state machine.

(Changing the tunnel state machine architecture at some point to allow for async transitions that could be interrupted by `mullvad disconnect`, etc., seems like a good idea, irrespective of what the timeout is here.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ Line wrap the file at 100 chars.                                              Th
 - Update wireguard-nt to 0.10.1.
 - Make wireguard-nt the default driver for WireGuard. This is used instead of wireguard-go and
   Wintun.
+- Increase firewall transaction timeout from 2 to 6 seconds.
 
 ### Fixed
 - Always kill `sslocal` if the tunnel monitor fails to start when using bridges.
--- a/talpid-core/src/firewall/windows.rs
+++ b/talpid-core/src/firewall/windows.rs
@@ -1,6 +1,6 @@
 use crate::{logging::windows::log_sink, tunnel::TunnelMetadata};
 
-use std::{net::IpAddr, path::Path, ptr};
+use std::{net::IpAddr, path::Path, ptr, time::Duration};
 
 use self::winfw::*;
 use super::{FirewallArguments, FirewallPolicy, FirewallT, InitialFirewallState};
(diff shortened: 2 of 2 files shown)


mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3205)* status: :shipit: complete! all files reviewed, all discussions resolved

mvd-ows (approved):
I would have probably picked 4 seconds, since the current limit of 2 seconds seems to almost always provide more than enough time.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3205#-:-Mr4JYZY0D4Z6qHuiDVL:b-z1j3hf)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3205)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[talpid-core/src/firewall/windows.rs, line 60 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3205#-Mr6pbU50Pjiwnm72c7f:-Mr6pbU50Pjiwnm72c7g:be85sar) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6e986d8197a36eca99b3fd0ff056d3b9734d2ea8/talpid-core/src/firewall/windows.rs#L60)):*
> ```XML
>         let logging_context = b"WinFw\0".as_ptr();
> 
>         let timeout_secs = u32::try_from(WINFW_TXN_TIMEOUT.as_secs()).unwrap();
> ```

I'm usually all for representing values as correctly as possible in the type system. But IMHO, this does not improve the timeout representation or code readability / reliability. It's more complex and introduces a hypothetical panic code path.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3205)* status: :shipit: complete! all files reviewed, all discussions resolved

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3205)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3221, opened 2021-12-27
Make build.sh produce single architecture pkg by default (add --universal)

faern:
In order to make it easier to use `build.sh` while developing we make it the default to build for the current architecture only. Since currently developers type `build.sh --target ..-darwin-...` every time. Add --universal flag for building universal app for both Intel and Apple Silicon. Make the build server script use the new flag, so it keeps producing universal binaries.

Maybe it's a little strange to carry around the `TARGETS` variable now when it's only ever used to list the two macOS target triples if it should be built for both.. But maybe it's worth keeping? I guess we are not sure how we will build arm64 Linux later? Or do you have another design that you prefer?

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3221)

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,8 +310,8 @@ Building this requires at least 1GB of memory.
 
 #### macOS
 
-By default a universal app is built on macOS. To build specifically for x86_64 or ARM64 add
-`--target x86_64-apple-darwin` or `--target aarch64-apple-darwin`.
+By default, `build.sh` produces a pkg for your current architecture only. To build a universal
+app that works on both Intel and Apple Silicon macs, build with `--universal`.
 
 ##### Apple ARM64 (aka Apple Silicon)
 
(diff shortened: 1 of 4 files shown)


faern:
The only Mac I currently have access to can't build for Apple Silicon. So I have only tried the non-universal code path to the very end. For the `--universal` case I have verified that it's triggered and tries to build for both architectures, but I have not been able to verify the output pkg.

raksooo (approved):
When I was experimenting with building for arm64 Linux I tried cross compiling but wasn't successful. We could remove the `TARGETS` variable and add it back if needed. I don't mind it being there but I'm also fine with removing it.

I tried building a universal on my computer and it works as expected. Although, I haven't tried running it on an arm64 mac.

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3221#-:-Ms-NL8_8Pw6p9_OG83-:byxwytt)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3228, opened 2021-12-30
build.sh improvements to support debug builds and be more flexible

faern:
`build.sh` has since the start been focused on being used mostly for building signed release artifacts. Using it as a developer has not been a high priority. But I want to change that. This script is often being overlooked, making it not work, because developers don't use it. And we want to make it easier and faster for developers to build test artifacts for themselves and their test machines.

So here I invert the arguments and make them more flexible. Instead of passing `--dev-build` to not build a signed release artifact, one does no longer need to pass anything. And the build server instead of passing nothing will now pass `--optimize --sign` to enable compiler optimizations and binary signing.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228)

### DIFF
--- a/.github/workflows/daemon.yml
+++ b/.github/workflows/daemon.yml
@@ -29,8 +29,6 @@ on:
             - integration-tests.sh
             - prepare-release.sh
             - rustfmt.toml
-            - update-api-address.sh
-            - update-relays.sh
             - version-metadata.sh
     # Build if requested manually from the Actions tab
     workflow_dispatch:
--- a/README.md
+++ b/README.md
@@ -300,9 +300,10 @@ storePassword = keystore-password
 ### Desktop
 
 The simplest way to build the entire app and generate an installer is to just run the build script.
-`--dev-build` is added to skip some release checks and signing of the binaries:
+`--optimize` can be added to enable compiler optimizations. This will take longer to build but will
+produce a smaller installer and installed binaries:
 ```bash
-./build.sh --dev-build
+./build.sh [--optimize]
 ```
 This should produce an installer exe, pkg or rpm+deb file in the `dist/` directory.
 
--- a/README.md
+++ b/README.md
@@ -382,7 +382,7 @@ echo "org.gradle.jvmargs=-Xmx4608M" >> ~/.gradle/gradle.properties
 
 1. On Windows, make sure to start bash first (e.g., Git BASH). Then build the C++ libraries:
    ```bash
-    ./build-windows-modules.sh --dev-build
+    ./build-windows-modules.sh
    ```
 
 1. Build the system daemon plus the other Rust tools and programs:
@@ -628,8 +628,7 @@ If you're using GNOME, try installing one of these GNOME Shell extensions:
 
 ### Building, testing and misc
 - **build-windows-modules.sh** - Compiles the C++ libraries needed on Windows
-- **build.sh** - Sanity checks the working directory state and then builds release artifacts for
-  the app
+- **build.sh** - Sanity checks the working directory state and then builds installers for the app
 
 ### Mullvad Daemon
 
--- a/build-apk.sh
+++ b/build-apk.sh
@@ -126,8 +126,10 @@ for ARCHITECTURE in ${ARCHITECTURES:-aarch64 armv7 x86_64 i686}; do
     $STRIP_TOOL --strip-debug --strip-unneeded -o "$STRIPPED_LIB_PATH" "$UNSTRIPPED_LIB_PATH"
 done
 
-./update-relays.sh
-./update-api-address.sh
+echo "Updating relays.json..."
+cargo run --bin relay_list $CARGO_ARGS > dist-assets/relays.json
+echo "Updating api-ip-address..."
+cargo run --bin address_cache $CARGO_ARGS > dist-assets/api-ip-address.txt
 
 cd "$SCRIPT_DIR/android"
 $GRADLE_CMD --console plain "$GRADLE_TASK"
--- a/build-apk.sh
+++ b/build-apk.sh
@@ -56,7 +56,7 @@ if [[ "$BUILD_TYPE" == "debug" || $product_version_commit_hash != $current_head_
     echo "Modifying product version to $PRODUCT_VERSION"
 else
     echo "Removing old Rust build artifacts"
-    cargo +stable clean
+    cargo clean
     CARGO_ARGS+=" --locked"
 fi
 
@@ -117,7 +117,7 @@ for ARCHITECTURE in ${ARCHITECTURES:-aarch64 armv7 x86_64 i686}; do
     esac
 
     echo "Building mullvad-daemon for $TARGET"
-    cargo +stable build $CARGO_ARGS --target "$TARGET" --package mullvad-jni
+    cargo build $CARGO_ARGS --target "$TARGET" --package mullvad-jni
 
     STRIP_TOOL="${NDK_TOOLCHAIN_DIR}/${LLVM_TRIPLE}-strip"
     STRIPPED_LIB_PATH="$SCRIPT_DIR/android/app/build/extraJni/$ABI/libmullvad_jni.so"
--- a/build-windows-modules.sh
+++ b/build-windows-modules.sh
@@ -9,18 +9,16 @@ CPP_BUILD_MODES=${CPP_BUILD_MODES:-"Debug"}
 # Common platforms include "x86" and "x64".
 CPP_BUILD_TARGETS=${CPP_BUILD_TARGETS:-"x64"}
 
-if [[ "${1:-""}" == "--dev-build" ]]; then
-    DEV_BUILD=true
-fi
+IS_RELEASE=${IS_RELEASE:-"false"}
 
 function clean_solution {
     local path="$1"
 
-    if [[ -z ${DEV_BUILD+x} ]]; then
+    if [[ "$IS_RELEASE" == "true" ]]; then
         # Clean all intermediate and output files
         rm -r "${path:?}/bin/"* || true
     else
-        echo "Will NOT clean intermediate files in $path/bin/"
+        echo "Will NOT clean intermediate files in $path/bin/ in dev builds"
     fi
 }
 
--- a/build.sh
+++ b/build.sh
@@ -1,11 +1,12 @@
 #!/usr/bin/env bash
 
-# This script is used to build, and sign a release artifact. See `README.md` for further
-# instructions.
-#
-# Invoke the script with --dev-build in order to skip checks, cleaning and signing.
+# This script is used to build, and optionally sign the app.
+# See `README.md` for further instructions.
 
 set -eu
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$SCRIPT_DIR"
 
 function log {
     local NO_COLOR="0m"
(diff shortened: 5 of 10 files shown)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228)* status: 0 of 10 files reviewed, 2 unresolved discussions

___
*[build.sh, line 104 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-Ms9nqDj5dljrUfb4_Kn:-Ms9nqDj5dljrUfb4_Ko:b-miir6v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L104)):*
> ```Shell
> # to use the DLLs from those paths. If we find value in being able to build apps with C++
> # in debug mode, we can fix this.
> CPP_BUILD_MODE="Release"
> ```

@dlon Is there any reason to want to be able to build an installer with Debug mode? We have never supported it so far in this script, but `build-windows-modules.sh` supports it. I guess there is a reason for that?
___
*[build.sh, line 164 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-Ms9oAOP3E84wVOunudr:-Ms9oAOP3E84wVOunuds:b-7pnj4q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L164)):*
> ```Shell
> 
>     if [[ "$(uname -s)" == "Darwin" ]]; then
>         log_info "Disabling Apple notarization of installer in dev build"
> ```

Putting it behind this `if` makes it print only on macOS, cleeean.
___
*[build.sh, line 186 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-Ms9oMfV7y9QhQI08dbk:-Ms9oMfV7y9QhQI08dbl:bpqpsh8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L186)):*
> ```Shell
>     ./version-metadata.sh restore-backup --desktop
>     mv Cargo.lock.bak Cargo.lock || true
>     popd > /dev/null
> ```

By default `pushd` and `popd` prints the path stack it's operating on. This is usually fine, but it creates some noise. Especially here since this function is always run on script exit. I found it ugly to have a bunch of paths be printed even after a successful script exit.
___
*[build.sh, line 241 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-Ms9ok6n1PgE3Xp0KDNE:-Ms9ok6n1PgE3Xp0KDNF:b-xqmspa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L241)):*
> ```Shell
>     if [[ "$SIGN" == "true" && "$(uname -s)" == "MINGW"* ]]; then
>         # Windows can only be built for this one target anyway, so it can be hardcoded.
>         sign_win "build/lib/x86_64-pc-windows-msvc/libwg.dll"
> ```

`libwg.dll` was magically signed much further down before. I think it's way easier to follow the flow if it's signed right after building it.
___
*[build.sh, line 358 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-Ms9pG1x-vmOB01bM3iq:-Ms9pG1x-vmOB01bM3ir:b7zylc3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L358)):*
> ```Shell
> cargo run --bin relay_list "${CARGO_ARGS[@]}" > dist-assets/relays.json
> log_info "Updating api-ip-address..."
> cargo run --bin address_cache "${CARGO_ARGS[@]}" > dist-assets/api-ip-address.txt
> ```

I moved these into here and removed the scripts in order to be able to make it build using the same arguments. Previously it was hardcoded as `--release` and that would double compile everything if using `build.sh` in debug mode.. Which would defeat one purpose of debug mode (build faster). I could make it so that `update-relays.sh` etc take some cargo arguments as arguments, that is another solution if you prefer it.

Is anyone ever using those scripts standalone? Otherwise I think they pollute the repository root more than they help us.

raksooo (commented):
A lot of nice improvements to the structure of the file! I just tried it and it's much easier to read the output :) 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-:-MsAbsJJ1I3-VUK6HV4K:b-c2lvee)* status: all files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[build.sh, line 291 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-MsAVRbK17f-WSQ86ASy:-MsAVRbK17f-WSQ86ASz:b9p906u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L291)):*
> ```Shell
>     else
>         local src_dir="$CARGO_TARGET_DIR/$RUST_BUILD_MODE"
>         local dst_dir="dist-assets"
> ```

We don't really use a lot of acronyms for variables in this script. I'm fine with `dir` since that's very common but I would prefer `destination` instead of `dst`. Also, it would make more sense to call `src_dir` something like `build_target_dir`. What do you think?
___
*[build.sh, line 300 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-MsAaPuT5b45FIBeFXrW:-MsAaPuU2d5A8x689lJy:b4eztrj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L300)):*
> ```Shell
>         if [[ "$(uname -s)" == "MINGW"* || "$binary" == *.dylib ]]; then
>             log_info "Copying $src => $dst"
>             cp "$src" "$dst"
> ```

I just noticed that the environment variable `OUT_DIR` will make cargo move the artifacts to the specified directory. Maybe it's not very useful here since we also run `strip` on them for Linux and macOS.
___
*[build.sh, line 358 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-Ms9pG1x-vmOB01bM3iq:-MsAWfUL5amuPQSx1aif:boddrvw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L358)):*

I moved these into here and removed the scripts in order to be able to make it build using the same arguments. Previously it was hardcoded as `--release` and that would double compile everything if using `build.sh` in debug mode.. Which would defeat one purpose of debug mode (build faster). I could make it so that `update-relays.sh` etc take some cargo arguments as arguments, that is another solution if you prefer it.

Is anyone ever using those scripts standalone? Otherwise I think they pollute the repository root more than they help us.
</blockquote></details>

Looks good :+1:

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228)* status: all files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[build.sh, line 300 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-MsAaPuT5b45FIBeFXrW:-MsAiQHQCN-1zBWnvl8L:b-wdzm6b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L300)):*

I just noticed that the environment variable `OUT_DIR` will make cargo move the artifacts to the specified directory. Maybe it's not very useful here since we also run `strip` on them for Linux and macOS.
</blockquote></details>

Mixed up the `OUT_DIR` environment variable with the `--out-dir` flag. But that flag is unstable so we can't use it yet.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228)* status: all files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[build.sh, line 307 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-MsAnFyJ3vPSPJjigzw5:-MsAnFyJ3vPSPJjigzw6:b-kl8t1u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L307)):*
> ```Shell
> 
>         if [[ "$SIGN" == "true" && "$(uname -s)" == "MINGW"* ]]; then
>             sign_win "$dst"
> ```

Now we sign the files at `$dst` rather than `$src`. I think this makes more sense. Does this have any unexpected caveat that I missed? Any reason why we needed the binaries in `$CARGO_TARGET_DIR/...` to be signed?

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228)* status: 9 of 10 files reviewed, 4 unresolved discussions (waiting on @raksooo)

___
*[build.sh, line 291 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-MsAVRbK17f-WSQ86ASy:-MsAuG5D7qIKq-GSEqql:b-hetvpx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L291)):*

We don't really use a lot of acronyms for variables in this script. I'm fine with `dir` since that's very common but I would prefer `destination` instead of `dst`. Also, it would make more sense to call `src_dir` something like `build_target_dir`. What do you think?
</blockquote></details>

What about these new names? `cargo_output_dir` instead of `src_dir` to clearly indicate not that it's the source to copy from, but rather where cargo puts binaries. And then just expand the abbreviation `dst`.

For the loop later I choose to stay with `source` + `destination`. This loop just moves+signs files, so I think it makes sense to think of the two paths as source and destination, not much else. What else would you call `source`? `cargo_output_binary` makes no sense :D
___
*[build.sh, line 203 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-MsAu4wh6A7PeBkr43XH:-MsAu4wi6W4ZDzOJd9oh:b-mb34hv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/69a521173f0ab355538c0abe2ea42a7a4e00229b/build.sh#L203)):*
> ```Shell
>         # be contacted.
>         for i in $(seq 0 ${NUM_RETRIES}); do
>             log_info "Signing $binary..."
> ```

I have not *yet* even tested this script with signing. I will do that once we have bikeshedded/agreed on most of the changes. But I figured a completely silent signing process is not optimal.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228)* status: 9 of 10 files reviewed, 3 unresolved discussions (waiting on @raksooo)

___
*[build.sh, line 291 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-MsAVRbK17f-WSQ86ASy:-MsU-FQN2jlPyfICtQ3v:brsuyt3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L291)):*

What about these new names? `cargo_output_dir` instead of `src_dir` to clearly indicate not that it's the source to copy from, but rather where cargo puts binaries. And then just expand the abbreviation `dst`.

For the loop later I choose to stay with `source` + `destination`. This loop just moves+signs files, so I think it makes sense to think of the two paths as source and destination, not much else. What else would you call `source`? `cargo_output_binary` makes no sense :D
</blockquote></details>

This looks good to me :+1:

raksooo (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228)* status: all files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[build.sh, line 203 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-MsAu4wh6A7PeBkr43XH:-MsUULEc3m2MiQYcFf4Z:bb6mt9b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/69a521173f0ab355538c0abe2ea42a7a4e00229b/build.sh#L203)):*

I have not *yet* even tested this script with signing. I will do that once we have bikeshedded/agreed on most of the changes. But I figured a completely silent signing process is not optimal.
</blockquote></details>

:+1:

pinkisemils (commented):
> Reviewed 9 of 10 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228)* status: all files reviewed, 3 unresolved discussions (waiting on @faern)

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228)* status: all files reviewed, 3 unresolved discussions (waiting on @faern)

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228)* status: all files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[build.sh, line 104 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3228#-Ms9nqDj5dljrUfb4_Kn:-MsUWuMx-H3tHHDuOkFY:bb5cpny) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e105896eee15a2979b007bfd6a973f0a0bb168a7/build.sh#L104)):*

@dlon Is there any reason to want to be able to build an installer with Debug mode? We have never supported it so far in this script, but `build-windows-modules.sh` supports it. I guess there is a reason for that?
</blockquote></details>

The only reason this always uses release mode, as far as I can tell, is that the NSIS script, and `distribution.js`, always points to release binaries. And because everything else was built with `--release`, I suppose. But no, I don't think we *need* these debug builds.

PR #3229, opened 2022-01-03
Position button content with css

raksooo:
This PR positions button content with CSS instead of JavaScript. The solution is similar to the navigation bar but a different requirement is that the text should always be centered. To accomplish this when there are icons of different width on the sides of the label, the icons are added to both sides but hidden on the one where they shouldn't be visible. This makes them take up the width and makes the widest side use up the same amount on both sides.

This made it possible to simplify the code a lot since `ref`s to the button and text aren't needed anymore.

One issue I ran into was that the `buttonRef.current?.getBoundingClientRect()` triggered a reflow which made transitions in `TransitionContainer` work. When removing the calls to `getBoundingClientRect` in `AppButton` some views lost their navigation transition. This was solved by triggering a reflow in `TransitionContainer` to make sure it's always done before setting the CSS that initiates the transition.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3229)

### DIFF
--- a/gui/src/renderer/components/AppButton.tsx
+++ b/gui/src/renderer/components/AppButton.tsx
@@ -1,36 +1,26 @@
-import React, { useCallback, useContext, useEffect, useMemo, useRef, useState } from 'react';
+import React, { useCallback, useContext, useMemo, useState } from 'react';
 import styled from 'styled-components';
 import { colors } from '../../config.json';
 import log from '../../shared/logging';
 import { useMounted } from '../lib/utilityHooks';
 import {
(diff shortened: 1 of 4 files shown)


pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3229#-:-MuBDOkM3UElwYn_1Wxv:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3242, opened 2022-01-07
Always use logging macros prefixed with log::

faern:
I'm thinking about adding a section to the coding guidelines on trying to avoid importing stand alone functions and macros as much as possible. At least where prefixing with the module adds valuable information to the reader. Logging was a prime example of this. So I thought I'd take this change through review before proposing any coding guidelines changes. `log::` is very short. It makes very few lines of code expand into more lines than they already were. And it adds a lot of context to the reader IMO. Only `error!(...)` could very well be some kind of sibling to `try!(...)`. `log::error!` is very clear.

Also, we are already using `log::$macro` in a lot of places in the code. Even in the same file we mix using it qualified and unqualified, so this is also in order to make it more consistent.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3242)

### DIFF
--- a/mullvad-daemon/src/cli.rs
+++ b/mullvad-daemon/src/cli.rs
@@ -1,5 +1,4 @@
 use clap::{crate_authors, crate_description, crate_name, App, Arg};
-use log;
 
 use crate::version;
 
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -29,7 +29,6 @@ use futures::{
     future::{abortable, AbortHandle, Future},
     StreamExt,
 };
-use log::{debug, error, info, warn};
 use mullvad_rpc::availability::ApiAvailabilityHandle;
 use mullvad_types::{
     account::{AccountData, AccountToken, VoucherSubmission},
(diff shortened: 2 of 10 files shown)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3242)* status: 0 of 10 files reviewed, all discussions resolved

___
*[mullvad-daemon/src/cli.rs, line 2 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3242#-Msoa4rO0YVXuGjt0Crf:-Msoa4rO0YVXuGjt0Crg:bukpi8t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dc034b7030d615acea40e3272eaaa114b605c7d7/mullvad-daemon/src/cli.rs#L2)):*
> ```XML
> use clap::{crate_authors, crate_description, crate_name, App, Arg};
> use log;
> ```

These `use` was useless :P Since it's a crate we depend on this namespace is always in scope.

pinkisemils (approved):
> Reviewed 8 of 10 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3242)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3246, opened 2022-01-11
Stable settings migrations (vendor types that a specific version depends on)

faern:
The current settings migration is very fragile to changes. Take for example `v3` that depends on the `DefaultDnsOptions` type. If we were today to remove the `block_trackers` setting in `master`, then this migration would break. It would still migrate settings, but the `v4` settings it would output would not have this flag properly set.

The solution I opted for was to vendor all the relevant dependencies to each settings version. It did not end up being too much.

Did I miss something here, or is this a good way to make the migration more stable and less susceptible to future changes?

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3246)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1333,6 +1333,7 @@ dependencies = [
  "fern",
  "futures",
  "ipnetwork",
+ "jnix",
  "lazy_static",
  "libc",
  "log",
--- a/mullvad-daemon/Cargo.toml
+++ b/mullvad-daemon/Cargo.toml
@@ -41,6 +41,7 @@ mullvad-management-interface = { path = "../mullvad-management-interface" }
 
 [target.'cfg(target_os="android")'.dependencies]
 android_logger = "0.8"
+jnix = { version = "0.4", features = ["derive"] }
 
 [target.'cfg(unix)'.dependencies]
 nix = "0.22.2"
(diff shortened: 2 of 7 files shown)


dlon (approved):
Previously, I found that vendoring all types led to an insane number of duplicates. But in this case, it seems fine.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3246#-:-Mt7i2509iFHlXNPI45U:b-ynojki)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (commented):
I think that's because previously we had an actual settings *struct* in the migration code. So we had to vendor *everything*. Now we treat it mostly as raw JSON with the exception of a few types that we need to modify during migration. So we only need to vendor those affected types.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3246#-:-Mt7pAqMEm4VQblCE3wz:b-ahtqz9)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3246#-:-Mt7x94z8o8GXf-KikYm:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3246)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3250, opened 2022-01-11
Build Windows libraries in debug mode with `build.sh`

dlon:
Changes:
* Build the Windows modules in debug mode when `build.sh` is called without `--optimize`.
* Link Rust crates against the correct Windows libraries. Previously, the DLLs were always built in release mode, and if the Rust code was built in debug mode, Cargo incorrectly linked against the debug builds of the Windows libraries.
* Remove unused `msiutil` NSIS plugin.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: 0 of 11 files reviewed, 1 unresolved discussion

___
*[dist-assets/windows/installer.nsh, line 4 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-Mt8lYLg5Ztos7_rIsav:-Mt8lYLg5Ztos7_rIsaw:b-i58e98) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8bebe9e8a0faa749fd961a0b3b472eb2f0236d1/dist-assets/windows/installer.nsh#L4)):*
> ```NSIS
> !include winver.nsh
> 
> !addplugindir "${BUILD_RESOURCES_DIR}\..\dist-assets"
> ```

I'm not sure how the NSIS plugins are identified here, but it does work to keep them in the same directory.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[build.sh, line 103 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-Mt8ntxO8nFdp9ixQNrA:-Mt8ntxO8nFdp9ixQNrB:bpy5ppc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8bebe9e8a0faa749fd961a0b3b472eb2f0236d1/build.sh#L103)):*
> ```Shell
> # C++ is currently hardcoded to build in release mode since distribution.js is hardcoded
> # to use the DLLs from those paths. If we find value in being able to build apps with C++
> # in debug mode, we can fix this.
> ```

This comment is no longer correct. Please remove.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: 0 of 11 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[build.sh, line 103 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-Mt8ntxO8nFdp9ixQNrA:-Mt8piTT2P02I3GwO7Ur:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8bebe9e8a0faa749fd961a0b3b472eb2f0236d1/build.sh#L103)):*

This comment is no longer correct. Please remove.
</blockquote></details>

Done.
___
*[build.sh, line 333 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-Mt8pjzfEskmAtCfIf5v:-Mt8pjzfEskmAtCfIf5w:b-nvdpei) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3f4fda96527868a5481223949c99e1f2c8d34aaf/build.sh#L333)):*
> ```Shell
>         local destination="$destination_dir/$(basename "$binary")"
>         log_info "Copying $binary => $destination"
>         cp "$binary" "$destination"
> ```

Arguably, the files should be copied in `build-windows-modules.sh`.

mvd-ows (approved):
> Reviewed 9 of 11 files at r1, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon and @faern)

___
*[dist-assets/windows/installer.nsh, line 4 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-Mt8lYLg5Ztos7_rIsav:-MtDCXlq5lkP0Ulvu7PS:b31t00p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8bebe9e8a0faa749fd961a0b3b472eb2f0236d1/dist-assets/windows/installer.nsh#L4)):*

I'm not sure how the NSIS plugins are identified here, but it does work to keep them in the same directory.
</blockquote></details>

What's the concern?
___
*[dist-assets/windows/installer.nsh, line 66 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDBurG3SxncSECbgBb:-MtDBurG3SxncSECbgBc:b-ejx2iw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/dist-assets/windows/installer.nsh#L66)):*
> ```NSIS
> 	File "${BUILD_RESOURCES_DIR}\binaries\x86_64-pc-windows-msvc\wintun\wintun.dll"
> 	File "${BUILD_RESOURCES_DIR}\binaries\x86_64-pc-windows-msvc\wireguard-nt\mullvad-wireguard.dll"
> 	File "${BUILD_RESOURCES_DIR}\..\windows\driverlogic\bin\$%TARGET_CONFIG%\driverlogic.exe"
> ```

I forget what the difference is between using `${XXX}` and `$%XXX%`, and the NSIS documentation is less than stellar.
___
*[gui/tasks/distribution.js, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDD1sm0Sd30encfuCR:-MtDD1sm0Sd30encfuCS:b-rjin8n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L211)):*
> ```JavaScript
> 
> function packWin() {
>   if (release) {
> ```

This could be more clearly expressed using the ternary operator.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 4 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[gui/tasks/distribution.js, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDD1sm0Sd30encfuCR:-MtDGH1g0wrDQUfO2WLs:b8qszwe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L211)):*

This could be more clearly expressed using the ternary operator.
</blockquote></details>

Would it not be even better to just pass the C++ build mode as an argument instead of duplicating the strings here? We compute the same thing in so many places. Let's add a mandatory argument to the script that tells it what CPP build mode it should pull binaries from?

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[gui/tasks/distribution.js, line 322 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDGblg6A5Sh0vTDet4:-MtDGblhBj5m9TaZFe0X:b-fqi2qu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L322)):*
> ```JavaScript
> 
> function windowsModule(name) {
>   if (release) {
> ```

This function partially duplicates what `packWin` does. I think this can be simplified/more code reuse.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 6 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[build.sh, line 327 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDH3UV-pEvT8qotyYZ:-MtDH3UV-pEvT8qotyY_:b-9vra8a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/build.sh#L327)):*
> ```Shell
> 
>     if [[ "$SIGN" == "true" ]]; then
>         sign_win "${SIGN_DEPENDENCIES[@]}"
> ```

I think this entire change can be reverted. Now it's effectively just renaming the variable and reducing the scope of the if. Why put the constant outside of the if statement? We don't use the variable outside the if, so it does not need to be declared if there is no signing going on.

I think the renaming to uppercase letters is good, because it's a constant. But if we aim to give it a better name I'd opt for `CPP_BINARIES` or something that explains **what they are** not **what action we are going to perform on them**. Because the latter becomes obvious from the call to `sign_win` anyway.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 7 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[gui/tasks/distribution.js, line 127 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDTcNm1fZuzkFl6IH5:-MtDTtuI3CO5mhH_YDnm:bl4pmkm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L127)):*

I'd prefer if the `.dll` part stayed here instead of moved into the function. Reading just the string `'winfw'` is pretty unclear what it is. At least `'winfw.dll'` would help me understand we are talking about the DLL artifact.
</blockquote></details>

Would be consistent with all the other entries in this list right here.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 6 unresolved discussions (waiting on @dlon, @faern, and @mvd-ows)

___
*[dist-assets/windows/installer.nsh, line 66 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDBurG3SxncSECbgBb:-MtDgBBq1-5eCLJzR4Te:b-ygo9gm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/dist-assets/windows/installer.nsh#L66)):*

I forget what the difference is between using `${XXX}` and `$%XXX%`, and the NSIS documentation is less than stellar.
</blockquote></details>

The latter refers to an environment variable. The others are defined in the nsis script I believe.
___
*[gui/tasks/distribution.js, line 127 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDTcNm1fZuzkFl6IH5:-MtDgZVQ5iykSsozpECa:b-fgorc7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L127)):*

Would be consistent with all the other entries in this list right here.
</blockquote></details>

I didn't do that because that would mean duplicating the project name for no real reason: `windowsModule('winfw', 'winfw.dll')`. But perhaps that's still more readable?

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 6 unresolved discussions (waiting on @dlon, @faern, and @mvd-ows)

___
*[dist-assets/windows/installer.nsh, line 66 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDBurG3SxncSECbgBb:-MtEvq5e7yh4glqYC9Ni:b-d39kto) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/dist-assets/windows/installer.nsh#L66)):*

The latter refers to an environment variable. The others are defined in the nsis script I believe.
</blockquote></details>

Thanks. Apparently `BUILD_RESOURCES_DIR` is defined by electron-builder before including this file.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 6 unresolved discussions (waiting on @dlon, @faern, and @mvd-ows)

___
*[dist-assets/windows/installer.nsh, line 4 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-Mt8lYLg5Ztos7_rIsav:-MtI0voMEpfBusFgUNN0:bus2a1w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8bebe9e8a0faa749fd961a0b3b472eb2f0236d1/dist-assets/windows/installer.nsh#L4)):*

What's the concern?
</blockquote></details>

That it might bundle stuff it shouldn't bundle, or try to load DLLs it shouldn't. Anyway, this is moot now.
___
*[gui/tasks/distribution.js, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDD1sm0Sd30encfuCR:-MtI0J7w5jgQzwMis2DL:bq8o3kt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L211)):*

Would it not be even better to just pass the C++ build mode as an argument instead of duplicating the strings here? We compute the same thing in so many places. Let's add a mandatory argument to the script that tells it what CPP build mode it should pull binaries from?
</blockquote></details>

Do you mean like this?

```
npm run pack:win -- --cpp-build-mode Release
```

I think I prefer it the other way as it is simpler and requires (almost) no argument parsing.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon, @mvd-ows, and @raksooo)

___
*[gui/tasks/distribution.js, line 127 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDTcNm1fZuzkFl6IH5:-MtICHntCLgaoyP_pC9H:bz9begi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L127)):*

I didn't do that because that would mean duplicating the project name for no real reason: `windowsModule('winfw', 'winfw.dll')`. But perhaps that's still more readable?
</blockquote></details>

To reduce the amount of magic involved I'd probably opt for the slightly more verbose but more readable:
```
{ from: root('windows/winfw/bin/' + cpp_build_mode() + '/winfw.dll'), to: '.' },
```
I don't see the benefit in removing a few character only to make it much harder to see which files are actually included in the package.

I'd really want @raksooo 's opinion here though..
___
*[gui/tasks/distribution.js, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDD1sm0Sd30encfuCR:-MtICA7v-q2xsJH7Wn8a:b-pbu2s9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L211)):*

Do you mean like this? `npm run pack:win -- --cpp-build-mode Release`? I think I prefer it the other way as it is simpler and requires (almost) no argument parsing.
</blockquote></details>

We should almost never call `npm run pack` manually, or does anyone do that in their development process? And it does not matter if it's complicated when called automatically by `build.sh`. But if someone really want to call that command manually we might want to go for the simpler approach you had here. I'm just not a huge fan of duplicating the "computation" of the directory name in so many places.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon, @faern, @mvd-ows, and @raksooo)

___
*[gui/tasks/distribution.js, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDD1sm0Sd30encfuCR:-MtIJNauB8g-Tp2LS7cy:b7fyq0c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L211)):*

We should almost never call `npm run pack` manually, or does anyone do that in their development process? And it does not matter if it's complicated when called automatically by `build.sh`. But if someone really want to call that command manually we might want to go for the simpler approach you had here. I'm just not a huge fan of duplicating the "computation" of the directory name in so many places.
</blockquote></details>

I do sometimes. It's the fastest way to test changes to the NSIS script.

mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon, @faern, and @raksooo)

___
*[gui/tasks/distribution.js, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDD1sm0Sd30encfuCR:-MtINoeUCFcZqOwzpQYt:bbur7k7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L211)):*

I do sometimes. It's the fastest way to test changes to the NSIS script.
</blockquote></details>

Perhaps add code to support `./build.sh --pack-only`?

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon, @mvd-ows, and @raksooo)

___
*[gui/tasks/distribution.js, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDD1sm0Sd30encfuCR:-MtIabUQ99zj7BxLqt9j:bpjfvjq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L211)):*

Perhaps add code to support `./build.sh --pack-only`?
</blockquote></details>

I don't think we should introduce `build.sh --pack-only`. Because it will not be guaranteed to work as expected. It won't build the artifacts to include, so it can't guarantee that it will build the correct thing. I think the flags should only affect how it builds stuff, but the outcome should always be a working artifact, which would not be true only for packing. That's just a convoluted way of making a shortcut for `npm run pack`. Then I much prefer the `--release` flag we have currently.

But to approve this we need to at least not duplicate the strings `x64-Release` and `x64-Debug` in multiple places in this file. We should be able to isolate that to a nice and tidy `cpp_build_mode()`. Or if that full string is not called a "mode" in VS-lingo then name it something else. But at least have a function that computes this string in a single place.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: 9 of 12 files reviewed, 3 unresolved discussions (waiting on @faern, @mvd-ows, and @raksooo)

___
*[build.sh, line 327 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDH3UV-pEvT8qotyYZ:-MtIdqzj4sKWeehZhUz4:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/build.sh#L327)):*

I think this entire change can be reverted. Now it's effectively just renaming the variable and reducing the scope of the if. Why put the constant outside of the if statement? We don't use the variable outside the if, so it does not need to be declared if there is no signing going on.

I think the renaming to uppercase letters is good, because it's a constant. But if we aim to give it a better name I'd opt for `CPP_BINARIES` or something that explains **what they are** not **what action we are going to perform on them**. Because the latter becomes obvious from the call to `sign_win` anyway.
</blockquote></details>

Done.
___
*[gui/tasks/distribution.js, line 127 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDTcNm1fZuzkFl6IH5:-MtIdzH14W3-znaZImf2:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L127)):*

To reduce the amount of magic involved I'd probably opt for the slightly more verbose but more readable:
```
{ from: root('windows/winfw/bin/' + cpp_build_mode() + '/winfw.dll'), to: '.' },
```
I don't see the benefit in removing a few character only to make it much harder to see which files are actually included in the package.

I'd really want @raksooo 's opinion here though..
</blockquote></details>

Done.
___
*[gui/tasks/distribution.js, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDD1sm0Sd30encfuCR:-MtIe7uW7shQH1choXaj:bcxa0oh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L211)):*

I don't think we should introduce `build.sh --pack-only`. Because it will not be guaranteed to work as expected. It won't build the artifacts to include, so it can't guarantee that it will build the correct thing. I think the flags should only affect how it builds stuff, but the outcome should always be a working artifact, which would not be true only for packing. That's just a convoluted way of making a shortcut for `npm run pack`. Then I much prefer the `--release` flag we have currently.

But to approve this we need to at least not duplicate the strings `x64-Release` and `x64-Debug` in multiple places in this file. We should be able to isolate that to a nice and tidy `cpp_build_mode()`. Or if that full string is not called a "mode" in VS-lingo then name it something else. But at least have a function that computes this string in a single place.
</blockquote></details>

I decided to use the same approach as is used by the macOS script instead. Thoughts?

faern (changes requested):
> Reviewed 2 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @dlon and @mvd-ows)

___
*[gui/tasks/distribution.js, line 211 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtDD1sm0Sd30encfuCR:-MtIf24eBJC6ExfEzLRg:bsqsc19) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02dca36c816f5dd513a23c9898134a4708ee38c2/gui/tasks/distribution.js#L211)):*

I decided to use the same approach as is used by the macOS script instead. Thoughts?
</blockquote></details>

I'm not a huge fan of using process env vars to send state between components within the same script. But since we have precedence already doing so, this is fine :)
___
*[gui/tasks/distribution.js, line 128 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtIfNEaC-eRlDJtkCdo:-MtIfNEaC-eRlDJtkCdp:b92014w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aac5ea5fcddfab1bde3eda1f4cafcb7d6625ff12/gui/tasks/distribution.js#L128)):*
> ```JavaScript
>       { from: root(path.join('windows', 'winfw', 'bin', 'x64-${env.CPP_BUILD_MODE}', 'winfw.dll')), to: '.' },
>       { from: root(path.join('windows', 'windns', 'bin', 'x64-${env.CPP_BUILD_MODE}', 'windns.dll')), to: '.' },
>       { from: root(path.join('windows', 'winnet', 'bin', 'x64-${env.CPP_BUILD_MODE}', 'winnet.dll')), to: '.' },
> ```

Looks like `path.join` can support having more than one path component per argument. So can't we simplify the first three arguments into `'windows/winnet/bin'` etc? We already do that in `function distAssets` further down. Not a blocker, but would look a bit shorter.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @dlon and @mvd-ows)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @faern and @mvd-ows)

___
*[gui/tasks/distribution.js, line 128 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250#-MtIfNEaC-eRlDJtkCdo:-MtIgpsQ2J0X5eTDW-7c:b1m5aei) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aac5ea5fcddfab1bde3eda1f4cafcb7d6625ff12/gui/tasks/distribution.js#L128)):*

Looks like `path.join` can support having more than one path component per argument. So can't we simplify the first three arguments into `'windows/winnet/bin'` etc? We already do that in `function distAssets` further down. Not a blocker, but would look a bit shorter.
</blockquote></details>

In that case, we might as well not use `path.join` at all:

```
root('windows/winnet/bin/x64-${env.CPP_BUILD_MODE}/winnet.dll')
```

mvd-ows (approved):
> Reviewed 1 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3250)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

PR #3254, opened 2022-01-11
Simplify custom resolver to not leak any traffic

pinkisemils:
As it turns out, it's not necessary to leak traffic to appease the network connectivity check on macOS, as such the custom resolver can be simplified a lot.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)


faern (changes requested):
> Reviewed 18 of 29 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: 18 of 29 files reviewed, 8 unresolved discussions (waiting on @pinkisemils)

___
*[docs/allow-macos-network-check.md, line 13 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCcWuu5yAIqvwSFkQ9:-MtCcWuu5yAIqvwSFkQA:b2ni54z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/docs/allow-macos-network-check.md#L13)):*
> ```Markdown
> always incurred, which forces the app into the offline error state for a prolonged time.
> 
> To not have to wait for macOS to time out it's captive portal check, the app should allow the
> ```

its
___
*[docs/allow-macos-network-check.md, line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCce2RDd_boio6CCDP:-MtCce2RDd_boio6CCDQ:b2ni54z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/docs/allow-macos-network-check.md#L17)):*
> ```Markdown
> traffic. However, only a DNS response is required to appease the connectivity check - and it doesn't
> even need to be valid. As such, during blocked states the app can run a custom resolver that only
> responds to queries for `captive.apple.com` to allow macOS to do it's connectivity check. Since no
> ```

its
___
*[docs/allow-macos-network-check.md, line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCch8pCeMUH1GgoTEU:-MtCch8pCeMUH1GgoTEV:b-8i5jbq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/docs/allow-macos-network-check.md#L20)):*
> ```Markdown
> lookups have to be made, no traffic needs to be leaked.
> 
> # Overcoming hese issues in the daemon.
> ```

s/hese/these/ ?
___
*[docs/security.md, line 248 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCd0ae6bFmzhzqOzHi:-MtCd0ae6bFmzhzqOzHj:bvne6mx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6ce47447cbfe5bbffebf79568f6cb9600271b380/docs/security.md#L248)):*
> ```Markdown
> connectivity check will fail in the blocked state imposing a hefty timeout before a tunnel can be
> connected. The connectivity check requires a working DNS resolver and access to `captive.apple.com`.
> The feature is discussed in detail [here](allow-macos-network-check.md).
> ```

Is `allow-macos-network-check.md` linked to from somewhere else? I don't think we should have unlinked documentation lying around. The discoverability of that is very bad. Can we find a good place to add a link to it? Since there are no leaks, `security.md` might no longer be the best place? Maybe `architecture.md`?
___
*[talpid-core/src/firewall/macos.rs, line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCf602EjlYmw-w4ebC:-MtCf602EjlYmw-w4ebD:b2k8uvj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/firewall/macos.rs#L24)):*
> ```XML
>     rule_logging: RuleLogging,
>     dns_redirect_port: Option<u16>,
>     _exclusion_gid: u32,
> ```

Why is the field kept if it's unused?
___
*[talpid-core/src/firewall/macos.rs, line 103 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCgIjR7ptWifLw3y93:-MtCgIjR7ptWifLw3y94:bvy0gwd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/firewall/macos.rs#L103)):*
> ```XML
>                 .dns_redirect_port
>                 .iter()
>                 .map(|redirect_port| {
> ```

I don't think this is optimal use of functional style programming. It looks like this will support an arbitrary number of redirect ports. But that would not make sense in practice. Would you redirect port 53 to multiple other ports? Personally I think a `if let Some(redirect_port) = self.dns_redirect_port {...}` here would be clearer and more readable.

Seee my other comment about this field being an `Option` before spending time on changing this :)
___
*[talpid-core/src/firewall/macos.rs, line 107 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCgwPO0MotxtHz0ZRC:-MtCgwPO0MotxtHz0ZRD:bpb8gg1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/firewall/macos.rs#L107)):*
> ```XML
>                         .action(pfctl::RedirectRuleAction::Redirect)
>                         .interface("lo0")
>                         .proto(pfctl::Proto::Udp)
> ```

Don't we need to also redirect TCP? Well, I guess the `captive.apple.com` lookup never uses TCP. But on this layer of abstraction the firewall is not/should not be aware of what the DNS redirect is used for. It would be strange if UDP lookups went to one resolver and TCP lookups to another, which is possible with this code if there is a resolver running on port 53.

But if fixing this is more complicated than just adding another rule, then it's probably more expensive than it's worth to fix. Just noting this.
___
*[talpid-core/src/firewall/mod.rs, line 230 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCi-e76QwxjLPUPCSg:-MtCi-e76QwxjLPUPCSh:b-hbt2iy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/firewall/mod.rs#L230)):*
> ```XML
>     #[cfg(target_os = "macos")]
>     /// DNS redirect port. DNS traffic in blocking states will be redirected to this port
>     pub dns_redirect_port: Option<u16>,
> ```

Would this make more sense as a field in `FirewallPolicy::Blocked`? I guess it's put here since it does not change over time. But it's a value only relevant for that specific policy. I'm thinking it's easier to read and understand what the policy does if as much of its data as possible is stored in the policy, and not some background state. What do you think? Downsides of having it in the policy? It needs to be passed in every time the blocking state is engaged, is it complicated to get hold of the desired port in all those places?

A downside with having it placed right here is that `mullvad-setup` needs to care about it and set it to `None`. If we move it to the policy it looks like we can model it directly as a `u16` and not an `Option<u16>`. That would be nicer.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: 16 of 30 files reviewed, 8 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[docs/allow-macos-network-check.md, line 13 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCcWuu5yAIqvwSFkQ9:-MtD6XAw3vw4eJHNMSwY:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/docs/allow-macos-network-check.md#L13)):*

its
</blockquote></details>

Done.
___
*[docs/allow-macos-network-check.md, line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCce2RDd_boio6CCDP:-MtD6WhG6Lekx_-Vfh3W:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/docs/allow-macos-network-check.md#L17)):*

its
</blockquote></details>

Done.
___
*[docs/allow-macos-network-check.md, line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCch8pCeMUH1GgoTEU:-MtD6WYp7kD6lK7f1DGB:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/docs/allow-macos-network-check.md#L20)):*

s/hese/these/ ?
</blockquote></details>

Done.
___
*[docs/security.md, line 248 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCd0ae6bFmzhzqOzHi:-MtD6YZc2o4idcV9CcJ4:ba9im4p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6ce47447cbfe5bbffebf79568f6cb9600271b380/docs/security.md#L248)):*

Is `allow-macos-network-check.md` linked to from somewhere else? I don't think we should have unlinked documentation lying around. The discoverability of that is very bad. Can we find a good place to add a link to it? Since there are no leaks, `security.md` might no longer be the best place? Maybe `architecture.md`?
</blockquote></details>

I've linked to it from `architecture.md` now. 
___
*[talpid-core/src/firewall/macos.rs, line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCf602EjlYmw-w4ebC:-MtD6YPt9pAIZxoPjTjM:b-3wfpfi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/firewall/macos.rs#L24)):*

Why is the field kept if it's unused?
</blockquote></details>

It's being kept in hopes of it being useful to enable split tunneling on macOS.
___
*[talpid-core/src/firewall/macos.rs, line 107 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCgwPO0MotxtHz0ZRC:-MtD3NVCDKn5fPlH5otf:b-z49uj0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/firewall/macos.rs#L107)):*

Don't we need to also redirect TCP? Well, I guess the `captive.apple.com` lookup never uses TCP. But on this layer of abstraction the firewall is not/should not be aware of what the DNS redirect is used for. It would be strange if UDP lookups went to one resolver and TCP lookups to another, which is possible with this code if there is a resolver running on port 53.

But if fixing this is more complicated than just adding another rule, then it's probably more expensive than it's worth to fix. Just noting this.
</blockquote></details>

The resolver currently isn't listening on TCP, but I can make it do so again. However, the DNS query for now will only ever need a single UDP packet, as it's just a query for `captive.apple.com`.  Since we will never have to spoof a response big enough to warrant TCP, I'll just add a redirect for TCP too, which will then make sure that a query won't ever reach a would-be resolver listening on port 53 for TCP, even if that resolver wouldn't be able to leak any traffic.
___
*[talpid-core/src/firewall/mod.rs, line 230 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtCi-e76QwxjLPUPCSg:-MtD5I_MAqztvvOJORjG:b-zb0fjr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/firewall/mod.rs#L230)):*

Would this make more sense as a field in `FirewallPolicy::Blocked`? I guess it's put here since it does not change over time. But it's a value only relevant for that specific policy. I'm thinking it's easier to read and understand what the policy does if as much of its data as possible is stored in the policy, and not some background state. What do you think? Downsides of having it in the policy? It needs to be passed in every time the blocking state is engaged, is it complicated to get hold of the desired port in all those places?

A downside with having it placed right here is that `mullvad-setup` needs to care about it and set it to `None`. If we move it to the policy it looks like we can model it directly as a `u16` and not an `Option<u16>`. That would be nicer.
</blockquote></details>

I don't believe it belongs in the firewall policy, since this won't change depending on policy. I don't think the fact that `mullvad-setup` has to deal with an optional argument is that big of a deal.

faern (changes requested):
> Reviewed 2 of 29 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: 21 of 30 files reviewed, 3 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[talpid-core/src/dns/macos.rs, line 66 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDJWxXDw9mXadNR_cw:-MtDJWxXDw9mXadNR_cx:by8zh7s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/dns/macos.rs#L66)):*
> ```XML
> 
> impl State {
>     fn send_new_config(&self) {}
> ```

Can this be removed now instead of keeping an empty function? If it has a reason to be left as dead code, please document why that is.

dlon (commented):
> Reviewed 7 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: 26 of 30 files reviewed, 4 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[talpid-core/src/resolver/mod.rs, line 140 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDHgXSENn8vlK5cRFi:-MtDHgXSENn8vlK5cRFj:b4na0sa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/resolver/mod.rs#L140)):*
> ```XML
>         }
> 
>         std::mem::drop(self);
> ```

This is redundant.

faern (changes requested):
> Reviewed 1 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: 27 of 30 files reviewed, 6 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[talpid-core/src/tunnel_state_machine/disconnected_state.rs, line 100 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDSlvyCW_0FXvwE0vA:-MtDSlvyCW_0FXvwE0vB:b-kx7cp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/tunnel_state_machine/disconnected_state.rs#L100)):*
> ```XML
>         #[cfg(target_os = "macos")]
>         if shared_values.block_when_disconnected {
>             if let Err(err) = DisconnectedState::setup_local_dns_config(shared_values) {
> ```

Probably more idiomatic to use `Self::` instead of `DisconnectedState::`
___
*[talpid-core/src/tunnel_state_machine/disconnected_state.rs, line 117 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDSw613HMU8rGJm2lv:-MtDSw613HMU8rGJm2lw:b-ya81ng) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/tunnel_state_machine/disconnected_state.rs#L117)):*
> ```XML
>         #[cfg(windows)]
>         Self::register_split_tunnel_addresses(shared_values, should_reset_firewall);
>         DisconnectedState::set_firewall_policy(shared_values, should_reset_firewall);
> ```

Same as above

dlon (approved):
> Reviewed 3 of 29 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: all files reviewed, 6 unresolved discussions (waiting on @faern and @pinkisemils)

faern (commented):
> Reviewed 2 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: all files reviewed, 6 unresolved discussions (waiting on @faern and @pinkisemils)

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: all files reviewed, 7 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[talpid-core/src/resolver/mod.rs, line 35 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDlqLb-aOOZOL9SO8W:-MtDlqLb-aOOZOL9SO8X:b-ebxnqn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/resolver/mod.rs#L35)):*
> ```XML
> const CAPTIVE_PORTAL_DOMAIN: &str = "captive.apple.com";
> const TTL_SECONDS: u32 = 3;
> const RESOLVED_ADDR: Ipv4Addr = Ipv4Addr::new(198, 51, 100, 1);
> ```

Can we get some documentation on this constant please. Why this IP and what is it used for.

faern (commented):
> Reviewed 1 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: all files reviewed, 9 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[talpid-core/src/resolver/mod.rs, line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDt0X37nj_CGfUsJZ5:-MtDt0X37nj_CGfUsJZ6:b-j9k8j9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6ce47447cbfe5bbffebf79568f6cb9600271b380/talpid-core/src/resolver/mod.rs#L1)):*
> ```XML
> use socket2::{Domain, Socket, Type};
> ```

Nit: Can't this file be `talpid-core/src/resolver.rs` since there is just a single file in this module? Easier to where in the code I am on the IDE tabs as few as possible are named `mod.rs` among other things.
___
*[talpid-core/src/resolver/mod.rs, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDtW-2Bms59oJmDbU6:-MtDtW-2Bms59oJmDbU7:b85254z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/resolver/mod.rs#L42)):*
> ```XML
>     let (resolver, resolver_handle) = FilteringResolver::new().await?;
>     tokio::spawn(resolver.run());
>     Ok(resolver_handle)
> ```

Looks like we use the handle for two things:
1) Stop the server when we drop it.
2) Get the listening port

I think we can have fewer types and a bit simpler design if we skip `start_resolver` and `ResolverHandle` completely. Unless we have an idea how they can be more utilized in the future. To me we are kind of reinventing `futures::future::Abortable` here :thinking:. Personally I would probably consider adding `listening_port` directly on the resolver type and use it something like:

```rust
let resolver = resolver::FilteringResolver::new().await?;
let resolver_port = resolver.listening_port()?;
let (abortable_resolver, resolver_handle) = futures::future::abortable(resolver.run());
tokio::spawn(abortable_resolver);
...
```

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: all files reviewed, 9 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[talpid-core/src/resolver/mod.rs, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDtW-2Bms59oJmDbU6:-MtDw-pP0IixZHA5MNYq:bh903k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/resolver/mod.rs#L42)):*

Looks like we use the handle for two things:
1) Stop the server when we drop it.
2) Get the listening port

I think we can have fewer types and a bit simpler design if we skip `start_resolver` and `ResolverHandle` completely. Unless we have an idea how they can be more utilized in the future. To me we are kind of reinventing `futures::future::Abortable` here :thinking:. Personally I would probably consider adding `listening_port` directly on the resolver type and use it something like:

```rust
let resolver = resolver::FilteringResolver::new().await?;
let resolver_port = resolver.listening_port()?;
let (abortable_resolver, resolver_handle) = futures::future::abortable(resolver.run());
tokio::spawn(abortable_resolver);
...
```
</blockquote></details>

Or simply save the join handle and use `JoinHandle::abort`.

faern (commented):
> Reviewed 1 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: all files reviewed, 9 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[talpid-core/src/resolver/mod.rs, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDtW-2Bms59oJmDbU6:-MtDx-Yd8X1bvMTVoBe1:bf625px) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/resolver/mod.rs#L42)):*

Or simply save the join handle and use `JoinHandle::abort`.
</blockquote></details>

Ah, even better! I forgot about the joinhandle :sweat_smile:. One potential downside: It does not abort automagically on drop. I have not checked exactly how it's used, so it could or could not be an issue.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: all files reviewed, 9 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[talpid-core/src/resolver/mod.rs, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDtW-2Bms59oJmDbU6:-MtDxU8O7bVDD5l6XpnS:bmt71c2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/resolver/mod.rs#L42)):*

Ah, even better! I forgot about the joinhandle :sweat_smile:. One potential downside: It does not abort automagically on drop. I have not checked exactly how it's used, so it could or could not be an issue.
</blockquote></details>

Maybe this change does not really help :shrug: But I could not avoid feeling it was a remain of something not properly adapted to shedding 99% of it's functionality. Maybe it's still reasonable to have a helper type that takes care of cleanup and access to the interesting data.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: all files reviewed, 8 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[talpid-core/src/resolver/mod.rs, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDtW-2Bms59oJmDbU6:-MtEYUOfCp35lIzV8Z_Q:br40wrv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/resolver/mod.rs#L42)):*

Maybe this change does not really help :shrug: But I could not avoid feeling it was a remain of something not properly adapted to shedding 99% of it's functionality. Maybe it's still reasonable to have a helper type that takes care of cleanup and access to the interesting data.
</blockquote></details>

If anything, the return type of `start_resolver()` would be turned into tuple of both the port and the abort handle. The issue with a plain `AbortHandle` is that it doesn't abort on a drop. So it needs to be wrapped in a different type that does implement `Drop`.  Since a type is already needed, I don't believe it hurts to store the port in there too, especially since this field will eventually be used in `SharedTunnelValues` so that it can be passed to `FirewallPolicy::Block` later. If this type is to be cloneable (it doesn't have to be), more wrapping has to be done to ensure the task is only aborted once and only when all the handles are gone.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: 22 of 30 files reviewed, 6 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[talpid-core/src/tunnel_state_machine/disconnected_state.rs, line 100 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDSlvyCW_0FXvwE0vA:-MtJ0iU9CCK2efZLcPyd:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/tunnel_state_machine/disconnected_state.rs#L100)):*

Probably more idiomatic to use `Self::` instead of `DisconnectedState::`
</blockquote></details>

Done.
___
*[talpid-core/src/tunnel_state_machine/disconnected_state.rs, line 117 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDSw613HMU8rGJm2lv:-MtJ0iGgB0Sloua5V29K:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/tunnel_state_machine/disconnected_state.rs#L117)):*

Same as above
</blockquote></details>

Done.

faern (commented):
> Reviewed 4 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: 26 of 30 files reviewed, 1 unresolved discussion (waiting on @dlon, @faern, and @pinkisemils)

___
*[talpid-core/src/resolver/mod.rs, line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDtW-2Bms59oJmDbU6:-MtJ8yfk7FniWVHtfinB:bb6mt9b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/resolver/mod.rs#L42)):*

If anything, the return type of `start_resolver()` would be turned into tuple of both the port and the abort handle. The issue with a plain `AbortHandle` is that it doesn't abort on a drop. So it needs to be wrapped in a different type that does implement `Drop`.  Since a type is already needed, I don't believe it hurts to store the port in there too, especially since this field will eventually be used in `SharedTunnelValues` so that it can be passed to `FirewallPolicy::Block` later. If this type is to be cloneable (it doesn't have to be), more wrapping has to be done to ensure the task is only aborted once and only when all the handles are gone. 
</blockquote></details>

:+1:

dlon (approved):
> Reviewed 4 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[talpid-core/src/dns/macos.rs, line 66 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtDJWxXDw9mXadNR_cw:-MtJWx1HAU8OHzNULUAM:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/940c0da292541c1fdd88158b0154c2e114860f1f/talpid-core/src/dns/macos.rs#L66)):*

Can this be removed now instead of keeping an empty function? If it has a reason to be left as dead code, please document why that is.
</blockquote></details>

Done.

faern (approved):
> Reviewed 1 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (approved):
> Reviewed 1 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (approved):
> Reviewed 1 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[talpid-core/src/firewall/mod.rs, line 141 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254#-MtNuczU1CHreY8owk0U:-MtNuczU1CHreY8owk0V:b371pli) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4ac857daf98f7f39d7879a45814acd29ac6e22b6/talpid-core/src/firewall/mod.rs#L141)):*
This documentation does not say anything that is not already conveyed by the name of the field. I still have no idea what this does. Can definitely be improved with something like

> Local DNS traffic (`127.0.0.1:53/UDP`) will be redirected to this port (`127.0.0.1:$dns_redirect_port/UDP`)

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3254)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3261, opened 2022-01-14
Simplify WireGuard tunnel monitor

dlon:
Changes:
* Previously, `Runtime::block_on` was used all over the place to execute async code in the connectivity monitor thread. It turned out much nicer to spawn a tokio task and use `spawn_blocking` for the blocking code instead.
* Remove channel used to stop waiting on the the IP interface to be set up. It's not needed (anymore?) because we can simply rely on a sender being dropped instead.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3261)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3261)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3262, opened 2022-01-14
If matching relays have 0 weight, pick one randomly

pinkisemils:
A small change to the relay selector to pick a relay with 0 weight if all matching relays have 0 weight.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262)

### DIFF
--- a/mullvad-daemon/src/relays.rs
+++ b/mullvad-daemon/src/relays.rs
@@ -773,15 +773,18 @@ impl RelaySelector {
             .collect()
     }
 
-    /// Pick a random relay from the given slice. Will return `None` if the given slice is empty
-    /// or all relays in it has zero weight.
+    /// Pick a random relay from the given slice. Will return `None` if the given slice is empty.
+    /// If all of the relays have a weight of 0, one will be picked at random without bias,
+    /// otherwise roulette wheel selection will be used to pick only relays with non-zero
+    /// weights.
     fn pick_random_relay<'a>(&mut self, relays: &'a [Relay]) -> Option<&'a Relay> {
         let total_weight: u64 = relays.iter().map(|relay| relay.weight).sum();
         if total_weight == 0 {
-            None
+            relays.choose(&mut self.rng)
         } else {
-            // Pick a random number in the range 0 - total_weight. This choses the relay.
-            let mut i: u64 = self.rng.gen_range(0, total_weight + 1);
+            // Pick a random number in the range 1 - total_weight. This choses the relay with a
+            // non-zero weight.
+            let mut i: u64 = self.rng.gen_range(1, total_weight + 1);
             Some(
                 relays
                     .iter()
@@ -789,7 +792,7 @@ impl RelaySelector {
                         i = i.saturating_sub(relay.weight);
                         i == 0
                     })
-                    .unwrap(),
+                    .expect("At least one relay must've had a weight above 0"),
             )
         }
     }


faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[mullvad-daemon/src/relays.rs, line 783 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNTpOzAE9YIMDrgDBu:-MtNTpOzAE9YIMDrgDBv:b-t47iyk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d916c889c580c2b10869dd0f62d61a111c9f658/mullvad-daemon/src/relays.rs#L783)):*
> ```XML
>         let total_weight: u64 = relays.iter().map(|relay| relay.weight).sum();
>         if total_weight == 0 {
>             relays.get(self.rng.gen_range(0, relays.len()))
> ```

The `rand` crates has a `choose` method on slices. It returns `None` if the slice is empty. So with that I think we can simplify this. It can get rid of the initial `relays.is_empty()` check and also not have to "compute" the index ranges when choosing uniformly.
___
*[mullvad-daemon/src/relays.rs, line 787 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNUNVd1qUkZ_odBGpG:-MtNUNVd1qUkZ_odBGpH:b711ok4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d916c889c580c2b10869dd0f62d61a111c9f658/mullvad-daemon/src/relays.rs#L787)):*
> ```XML
>             // Pick a random number in the range 1 - total_weight. This choses the relay with a
>             // non-zero weight.
>             let mut i: u64 = self.rng.gen_range(1, total_weight + 1);
> ```

Ooh.. This must have been a slight bug up until now? With the range `[0, total_weight +1]` did we give the first relay a slight statistical advantage? Nice fix.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[mullvad-daemon/src/relays.rs, line 783 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNTpOzAE9YIMDrgDBu:-MtN_1Yf-4WKX3zmvGH3:b-ahyjhh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d916c889c580c2b10869dd0f62d61a111c9f658/mullvad-daemon/src/relays.rs#L783)):*

The `rand` crates has a `choose` method on slices. It returns `None` if the slice is empty. So with that I think we can simplify this. It can get rid of the initial `relays.is_empty()` check and also not have to "compute" the index ranges when choosing uniformly.
</blockquote></details>

I was sure it existed, but `select` and `pick` didn't return any results from the docs :sweat_smile: 
___
*[mullvad-daemon/src/relays.rs, line 787 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNUNVd1qUkZ_odBGpG:-MtN_AzWB9Tbt17fECA-:br84zef) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d916c889c580c2b10869dd0f62d61a111c9f658/mullvad-daemon/src/relays.rs#L787)):*

Ooh.. This must have been a slight bug up until now? With the range `[0, total_weight +1]` did we give the first relay a slight statistical advantage? Nice fix.
</blockquote></details>

 I changed the range of values that could be assigned to i to not include 0, so instead it can be [1; total_weight +1). This is so that 0 weighted relays can't be picked in the case that i would be set to 0 and the first relay in the slice would have a weight of 0. 

But I guess it also fixes the slight preference for the first relay, but I'm not so sure anymore. A safer bet of fixing the issue of picking a  relay with a 0 weight would be to just filter them out explicitly and have the behavior remain as it always has been by leaving `i` to be in the range of `[0; total_weight]`.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-daemon/src/relays.rs, line 777 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNcc103EF0RTe4Ph1X:-MtNcc116W7w_2j3tRBX:bthd9ed) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fa62c258f0777b85e94b7df59f384f49686b04f9/mullvad-daemon/src/relays.rs#L777)):*
> ```XML
> 
>     /// Pick a random relay from the given slice. Will return `None` if the given slice is empty
>     /// or all relays in it has zero weight.
> ```

Good to remove the invalid documentation. But I think we should also add the new functionality. Say that if all relays has a weight of zero, then one random relay is picked.
___
*[mullvad-daemon/src/relays.rs, line 779 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNcUGcBQf0_mgOxre8:-MtNcUGcBQf0_mgOxre9:b98ia97) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6e52197b87108d51510ae6fe57ac3bc1b2a328b1/mullvad-daemon/src/relays.rs#L779)):*
> ```XML
>     fn pick_random_relay<'a>(&mut self, relays: &'a [Relay]) -> Option<&'a Relay> {
>         if relays.is_empty() {
>             return None;
> ```

This is not needed. If the slice is empty then `total_weight` will be `0` and `relays.choose` will return `None` for you.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-daemon/src/relays.rs, line 787 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNUNVd1qUkZ_odBGpG:-MtNdC3_83YlCC_WUWZq:b-9lffod) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d916c889c580c2b10869dd0f62d61a111c9f658/mullvad-daemon/src/relays.rs#L787)):*

 I changed the range of values that could be assigned to i to not include 0, so instead it can be [1; total_weight +1). This is so that 0 weighted relays can't be picked in the case that i would be set to 0 and the first relay in the slice would have a weight of 0. 

But I guess it also fixes the slight preference for the first relay, but I'm not so sure anymore. A safer bet of fixing the issue of picking a  relay with a 0 weight would be to just filter them out explicitly and have the behavior remain as it always has been by leaving `i` to be in the range of `[0; total_weight]`. 
</blockquote></details>

Depending on the weights the advantage of being the first relay could have been huge :P

I don't think we should fix this in two ways with a filter. If the the rest of the algorithm is correct it should not be needed. Better to fail loudly. We could even replace `saturating_sub` with `checked_sub` and panic there as well to fail even louder if there is an issue with the algorithm.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-daemon/src/relays.rs, line 777 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNcc103EF0RTe4Ph1X:-MtNfv8qDwTDbtQdEyOs:b-sde1ef) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fa62c258f0777b85e94b7df59f384f49686b04f9/mullvad-daemon/src/relays.rs#L777)):*

Good to remove the invalid documentation. But I think we should also add the new functionality. Say that if all relays has a weight of zero, then one random relay is picked.
</blockquote></details>

Or be clear that it's uniform random. And with non-zero weight a roulette wheel selection is used.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-daemon/src/relays.rs, line 787 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNUNVd1qUkZ_odBGpG:-MtNkbCZ9jPeByhoJ_-4:b1vb6s5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d916c889c580c2b10869dd0f62d61a111c9f658/mullvad-daemon/src/relays.rs#L787)):*

Depending on the weights the advantage of being the first relay could have been huge :P

I don't think we should fix this in two ways with a filter. If the the rest of the algorithm is correct it should not be needed. Better to fail loudly. We could even replace `saturating_sub` with `checked_sub` and panic there as well to fail even louder if there is an issue with the algorithm.
</blockquote></details>

Given 3 relays each with a weight of 10 and `i` being set to 5, we'll panic for no reason. `saturating_sub` is the correct method to use here.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[mullvad-daemon/src/relays.rs, line 777 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNcc103EF0RTe4Ph1X:-MtNmUxqDIKz8vYVVXru:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fa62c258f0777b85e94b7df59f384f49686b04f9/mullvad-daemon/src/relays.rs#L777)):*

Or be clear that it's uniform random. And with non-zero weight a roulette wheel selection is used.
</blockquote></details>

Done.
___
*[mullvad-daemon/src/relays.rs, line 779 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNcUGcBQf0_mgOxre8:-MtNmVDW5HUPMFl1KD0e:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6e52197b87108d51510ae6fe57ac3bc1b2a328b1/mullvad-daemon/src/relays.rs#L779)):*

This is not needed. If the slice is empty then `total_weight` will be `0` and `relays.choose` will return `None` for you.
</blockquote></details>

Done.

faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-:-MtNn7fQArALEhAdnVEB:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[mullvad-daemon/src/relays.rs, line 787 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3262#-MtNUNVd1qUkZ_odBGpG:-MtNn-kCAx2QTNWEw7A9:b-55oa2c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d916c889c580c2b10869dd0f62d61a111c9f658/mullvad-daemon/src/relays.rs#L787)):*

Given 3 relays each with a weight of 10 and `i` being set to 5, we'll panic for no reason. `saturating_sub` is the correct method to use here.
</blockquote></details>

You are correct! I had a slight brain fart.

PR #3267, opened 2022-01-17
Make `mullvad relay set hostname` not case sensitive to the input

faern:
This has been done previously for `mullvad relay set location`, but the hostname seems to have been ignored/forgotten.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3267)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,8 @@ Line wrap the file at 100 chars.                                              Th
 - Gradually increase the WireGuard connectivity check timeout, lowering the timeout for the first
   few attempts.
 - Stop preferring OpenVPN when bridge mode is enabled.
+- CLI command for setting a specific server by hostname is no longer case sensitive.
+  Example: `mullvad relay set hostname SE9-WIREGUARD` should now work.
 
 #### Android
 - Avoid running in foreground when not connected.
--- a/mullvad-cli/src/cmds/relay.rs
+++ b/mullvad-cli/src/cmds/relay.rs
@@ -393,11 +393,15 @@ impl Relay {
         let countries = Self::get_filtered_relays().await?;
 
         let find_relay = || {
-            for country in &countries {
-                for city in &country.cities {
-                    for relay in &city.relays {
-                        if relay.hostname == hostname {
-                            return Some((country, city, relay));
+            for country in countries {
+                for city in country.cities {
+                    for relay in city.relays {
+                        if relay.hostname.to_lowercase() == hostname.to_lowercase() {
+                            return Some(types::RelayLocation {
+                                country: country.code,
+                                city: city.code,
+                                hostname: relay.hostname,
+                            });
                         }
                     }
                 }
@@ -408,19 +412,13 @@ impl Relay {
         if let Some(location) = find_relay() {
             println!(
                 "Setting location constraint to {} in {}, {}",
-                location.2.hostname, location.1.name, location.0.name
+                location.hostname, location.city, location.country
             );
-
-            let location_constraint = types::RelayLocation {
-                country: location.0.code.clone(),
-                city: location.1.code.clone(),
-                hostname: location.2.hostname.clone(),
-            };
 
             self.update_constraints(types::RelaySettingsUpdate {
                 r#type: Some(types::relay_settings_update::Type::Normal(
                     types::NormalRelaySettingsUpdate {
-                        location: Some(location_constraint),
+                        location: Some(location),
                         ..Default::default()
                     },
                 )),


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3267)* status: 0 of 2 files reviewed, all discussions resolved

___
*[mullvad-cli/src/cmds/relay.rs, line 403 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3267#-MtbdWjaASzm66S507RI:-MtbdWjaASzm66S507RJ:b-683pjp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/80d2476c066813745947ee0e5a4b272f3cdb476b/mullvad-cli/src/cmds/relay.rs#L403)):*
> ```XML
>                                 country: country.code,
>                                 city: city.code,
>                                 hostname: relay.hostname,
> ```

Please note that this is not lowercased. Only the comparison is. Otherwise the command would be able to find and set a hostname that does not exist. Let's say there is a relay list containing a host with name `SE8-WIREGUARD`. Then this command should be able to find and configure its usage by `mullvad relay set hostname se8-wireguard`, but the config should be updated to the hostname that the relay actually has (`SE8-WIREGUARD`).

That's why only the comparison has lowercasing, not the values later sent back to the daemon.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3267)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[mullvad-cli/src/cmds/relay.rs, line 399 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3267#-Mtbs4GA5MPnHeUraHzO:-Mtbs4GA5MPnHeUraHzP:b-dt7erp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/80d2476c066813745947ee0e5a4b272f3cdb476b/mullvad-cli/src/cmds/relay.rs#L399)):*
> ```XML
>                 for city in country.cities {
>                     for relay in city.relays {
>                         if relay.hostname.to_lowercase() == hostname.to_lowercase() {
> ```

This doesn't actually matter, but you could create a lowercase copy of `hostname` only once when defining it instead of once per relay.
___
*[mullvad-cli/src/cmds/relay.rs, line 403 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3267#-MtbdWjaASzm66S507RI:-MtbtdggDz9fRS_1idjj:b-4s3lx6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/80d2476c066813745947ee0e5a4b272f3cdb476b/mullvad-cli/src/cmds/relay.rs#L403)):*

Please note that this is not lowercased. Only the comparison is. Otherwise the command would be able to find and set a hostname that does not exist. Let's say there is a relay list containing a host with name `SE8-WIREGUARD`. Then this command should be able to find and configure its usage by `mullvad relay set hostname se8-wireguard`, but the config should be updated to the hostname that the relay actually has (`SE8-WIREGUARD`).

That's why only the comparison has lowercasing, not the values later sent back to the daemon.
</blockquote></details>

That's a good point. It looks like we do always send back the lowercase values for `mullvad relay set location ...`, though.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3267)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[mullvad-cli/src/cmds/relay.rs, line 399 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3267#-Mtbs4GA5MPnHeUraHzO:-Mtbx_MdCYCjpbJ6JBjR:bkx92m1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/80d2476c066813745947ee0e5a4b272f3cdb476b/mullvad-cli/src/cmds/relay.rs#L399)):*

This doesn't actually matter, but you could create a lowercase copy of `hostname` only once when defining it instead of once per relay.
</blockquote></details>

That's a good point. But I think I prefer the clarity of this version.
___
*[mullvad-cli/src/cmds/relay.rs, line 403 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3267#-MtbdWjaASzm66S507RI:-Mtbxzvm4izrs4Vflz9S:b-wp2f63) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/80d2476c066813745947ee0e5a4b272f3cdb476b/mullvad-cli/src/cmds/relay.rs#L403)):*

That's a good point. It looks like we don't do this for `mullvad relay set location ...`, though.
</blockquote></details>

Maybe we want to fix that at some point. But this is more of a theoretical issue than a practical issue. Since all our country codes and hostnames will be lowercase for the foreseeable future. I just did not want the CLI tool to be relying on this fact.

PR #3277, opened 2022-01-19
Fix auto-connect on app removed from recent apps

albin-mullvad:
By default in Android, a service will be automatically killed if it
belongs to a task that was removed (i.e. from recent apps) without a
chance to run any cleanup logic.

In this case, the lack of cleanup would lead to a crash and auto-restart
(triggered by the system) resulting in a auto-connect. To fix this, a
graceful stop will be ran if the task is removed while the tunnel is
disconnected (service not running as foreground). However if the tunnel
is in other states than disconnected (service running as foreground),
the service should continue to run as it's bound by the system.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3277)

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -51,7 +51,8 @@
         </activity>
         <service android:name="net.mullvad.mullvadvpn.service.MullvadVpnService"
                  android:permission="android.permission.BIND_VPN_SERVICE"
-                 android:process=":mullvadvpn_daemon">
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3277#-:-MtwLhbB-f18bwQRjK3z:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3279, opened 2022-01-19
Use routes for offline state monitoring on macOS

pinkisemils:
It seems like there are times when the system-configuration watcher either doesn't receive a notification at all or receives one at a significant delay even when a default route is setup in the routing table. E.g. it's sometimes possible to end up in the offline state during boot-up due to this. To circumvent this, I've added a more eager way to deduce the offline state - monitor the output of `route monitor` and query the routing table for a default route every time it a change is registered. If there exists a default route that isn't routed through a tunnel interface (which is inferred by checking if the device name contains `tun`), the host is considered to be online.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,10 @@ Line wrap the file at 100 chars.                                              Th
 - Change Quick Settings tile label to reflect the action of clicking the tile. Also add a subtitle
   on supported Android versions (Q and above) to reflect the state.
 
+#### macOS
+- Use `route monitor` instead of relying on `SCNetworkReachability` API to infer whether the host is
+  offline.
+
 #### Windows
 - Update wireguard-nt to 0.10.1.
 - Make wireguard-nt the default driver for WireGuard. This is used instead of wireguard-go and
--- a/talpid-core/src/offline/macos.rs
+++ b/talpid-core/src/offline/macos.rs
@@ -1,40 +1,11 @@
-use futures::channel::mpsc::UnboundedSender;
-use std::{
-    net::{Ipv4Addr, SocketAddr},
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        mpsc, Arc, Weak,
-    },
-    thread,
-};
-use system_configuration::{
-    core_foundation::{
-        array::CFArray,
-        base::{CFType, TCFType, ToVoid},
-        boolean::CFBoolean,
-        dictionary::CFDictionary,
-        runloop::{kCFRunLoopCommonModes, CFRunLoop},
-        string::CFString,
-    },
-    dynamic_store::{SCDynamicStore, SCDynamicStoreBuilder, SCDynamicStoreCallBackContext},
-    network_configuration::{self, SCNetworkInterface, SCNetworkInterfaceType},
-    network_reachability::{
-        ReachabilityFlags, SCNetworkReachability, SchedulingError, SetCallbackError,
-    },
-};
-
-const PRIMARY_INTERFACE_KEY: &str = "State:/Network/Global/IPv4";
+use futures::{channel::mpsc::UnboundedSender, StreamExt};
+use std::sync::{Arc, Weak};
+use talpid_types::ErrorExt;
 
 #[derive(err_derive::Error, Debug)]
 pub enum Error {
-    #[error(display = "Failed to initialize dynamic store")]
-    DynamicStoreInitError,
-    #[error(display = "Failed to schedule reachability callback")]
-    ScheduleReachabilityCallbackError(#[error(source)] SchedulingError),
-    #[error(display = "Failed to set reachability callback")]
-    SetCallbackError(#[error(source)] SetCallbackError),
-    #[error(display = "Panic during initialization")]
-    InitializationError,
+    #[error(display = "Failed to initialize route monitor")]
+    StartMonitorError(#[error(source)] crate::routing::PlatformError),
 }
 
 pub struct MonitorHandle {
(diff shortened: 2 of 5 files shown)


faern (changes requested):
> Reviewed 4 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279)* status: 4 of 5 files reviewed, 5 unresolved discussions (waiting on @pinkisemils)

___
*[CHANGELOG.md, line 59 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-Mtql7klBX58gm9teDvF:-Mtql7klBX58gm9teDvG:b-4qzy1b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d63d675e53487aa18f8d41cd196fdb17377ee54/CHANGELOG.md#L59)):*
> ```Markdown
> #### macOS
> - Use `route monitor` instead of relying on `SCNetworkReachability` API to infer whether the host is
>   offline.
> ```

Iff this fixes a known issue/bug then this should be under `### Fixed`. And then it should also explain what it fixes, not only what was changed.
___
*[talpid-core/src/routing/macos.rs, line 171 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtqivJN3eWm2qmtFALX:-MtqivJN3eWm2qmtFALY:b-6bf7dw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d63d675e53487aa18f8d41cd196fdb17377ee54/talpid-core/src/routing/macos.rs#L171)):*
> ```XML
>     // Retrieves the node that's currently used to reach 0.0.0.0/0
>     // Arguments can be either -inet or -inet6
>     pub(crate) async fn get_default_node_cmd(if_family: &'static str) -> Result<Option<Node>> {
> ```

Optimal would be if this function did not have `_cmd` in its name, nor was possible to call with invalid arguments. Or at least that it did not happily executed the subcommand if it was invalid. It's not a very clean abstraction to force the caller to know arguments to the `route` command IMO.

I'm not sure if there is any existing enum for `v4` vs `v6` in a reasonable place in our dependency tree that we could rely on or not.
___
*[talpid-core/src/routing/macos.rs, line 299 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtqkFjb7uH0wSqoX0j9:-MtqkFjb7uH0wSqoX0jA:b-9no9dm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d63d675e53487aa18f8d41cd196fdb17377ee54/talpid-core/src/routing/macos.rs#L299)):*
> ```XML
> /// Returns a stream that produces an item whenever a default route is either added or deleted from
> /// the routing table.
> pub(crate) async fn listen_for_default_route_changes(
> ```

This function does not do a single `.await`, does it need to be `async`? :thinking: 
___
*[talpid-core/src/routing/unix.rs, line 270 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtqksfJCrUHL5mLR71S:-MtqksfJCrUHL5mLR71T:bfd0xl6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d63d675e53487aa18f8d41cd196fdb17377ee54/talpid-core/src/routing/unix.rs#L270)):*
> ```XML
> 
> #[cfg(target_os = "macos")]
> pub(crate) async fn get_default_routes() -> Result<(Option<super::Node>, Option<super::Node>), Error>
> ```

It's not obvious what the returned tuple is. Can the types be tighter to signal that it's v4 + v6? Can we at least document it, would be great :+1:
___
*[talpid-core/src/routing/unix.rs, line 273 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtqkoQmB9pdMGe83Va1:-MtqkoQmB9pdMGe83Va2:b-9ox0qn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d63d675e53487aa18f8d41cd196fdb17377ee54/talpid-core/src/routing/unix.rs#L273)):*
> ```XML
> {
>     let v4 = imp::RouteManagerImpl::get_default_node_cmd("-inet").await?;
>     let v6 = imp::RouteManagerImpl::get_default_node_cmd("-inet").await?;
> ```

It's called with `-inet` both for `v4` and `v6`.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279)* status: 1 of 5 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[CHANGELOG.md, line 59 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-Mtql7klBX58gm9teDvF:-Mtr9QWD9GkfNc-6uUNq:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d63d675e53487aa18f8d41cd196fdb17377ee54/CHANGELOG.md#L59)):*

Iff this fixes a known issue/bug then this should be under `### Fixed`. And then it should also explain what it fixes, not only what was changed.
</blockquote></details>

Done.
___
*[talpid-core/src/routing/unix.rs, line 270 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtqksfJCrUHL5mLR71S:-Mtr8je4A66jSVKqLQF9:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d63d675e53487aa18f8d41cd196fdb17377ee54/talpid-core/src/routing/unix.rs#L270)):*

It's not obvious what the returned tuple is. Can the types be tighter to signal that it's v4 + v6? Can we at least document it, would be great :+1:
</blockquote></details>

Done.
___
*[talpid-core/src/routing/unix.rs, line 273 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtqkoQmB9pdMGe83Va1:-Mtr8jVd57ioKZHY9rTP:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d63d675e53487aa18f8d41cd196fdb17377ee54/talpid-core/src/routing/unix.rs#L273)):*

It's called with `-inet` both for `v4` and `v6`.
</blockquote></details>

Done.

dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279)* status: all files reviewed, 4 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[talpid-core/src/routing/unix.rs, line 270 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtqksfJCrUHL5mLR71S:-MtrKFGx0Qed96tR0T0q:bk3w41f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d63d675e53487aa18f8d41cd196fdb17377ee54/talpid-core/src/routing/unix.rs#L270)):*

Done.
</blockquote></details>

On Linux, we have `RouteManagerHandle::change_listener()` rather than a separate function. Makes more sense there because it's dependent on the netlink socket.

Regardless, would it be nicer to use the same interface here? That way the listeners could also be dropped along with the route manager.
___
*[talpid-core/src/routing/unix.rs, line 274 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtrKbuV22gBEWp3mTTj:-MtrKbuV22gBEWp3mTTk:b-xwdn2f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8bdd97124aac4f37420089fd213971b9290579bc/talpid-core/src/routing/unix.rs#L274)):*
> ```XML
> pub(crate) async fn get_default_routes() -> Result<(Option<super::Node>, Option<super::Node>), Error>
> {
>     let v4 = imp::RouteManagerImpl::get_default_node(IpVersion::V4).await?;
> ```

I suppose these futures could be `future::join`ed?

faern (changes requested):
> Reviewed 2 of 4 files at r2, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[CHANGELOG.md, line 57 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtrUKss4easWocT2zJU:-MtrUKss4easWocT2zJV:bi3dhqt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/141d5332819539fe24930d24ae0e889a204da266/CHANGELOG.md#L57)):*
> ```Markdown
>   on supported Android versions (Q and above) to reflect the state.
> 
> #### macOS
> ```

The `#### macOS` section is now empty... The header should not remain here
___
*[talpid-core/src/routing/macos.rs, line 171 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtqivJN3eWm2qmtFALX:-MtrV2iF3dXKR2Zzohx3:bn4kh5d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d63d675e53487aa18f8d41cd196fdb17377ee54/talpid-core/src/routing/macos.rs#L171)):*

Optimal would be if this function did not have `_cmd` in its name, nor was possible to call with invalid arguments. Or at least that it did not happily executed the subcommand if it was invalid. It's not a very clean abstraction to force the caller to know arguments to the `route` command IMO.

I'm not sure if there is any existing enum for `v4` vs `v6` in a reasonable place in our dependency tree that we could rely on or not.
</blockquote></details>

Great :clap: This is way nicer!

dlon (approved):
> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279)* status: all files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-core/src/offline/macos.rs, line 10 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtrXHd43moY0aRpkBTG:-MtrXHd43moY0aRpkBTH:b-ofryk4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9b5da7a14f2be306817fe414543124b7bdc9398d/talpid-core/src/offline/macos.rs#L10)):*
> ```XML
> //! address that isn't using a tunnel adapter. On macOS, there were various ways of deducing this:
> //! - watching the `State:/Network/Global/IPv4`  key in SystemConfiguration via
> //!  `system-configuration-rs`, relying on a CoreFoudnation runloop to drive callbacks.
> ```

Nit: CoreFoundation

faern (approved):
:lgtm:
I did not read the details of the implementation. But I looked at the general idea of the solution. And i LOVE the new documentation on the offline module.

> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-:-MtsMjeyBGOF0or94BVc:b-un51jh)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

___
*[talpid-core/src/routing/unix.rs, line 274 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtrKbuV22gBEWp3mTTj:-MtxOBkzB_2FPD2WkgVJ:b-lqpkl3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8bdd97124aac4f37420089fd213971b9290579bc/talpid-core/src/routing/unix.rs#L274)):*

I suppose these futures could be `future::join`ed?
</blockquote></details>

You're right, but I think more changes were necessary anyway, as even if the daemon fails to get one of the default routes, it's ok not to error out.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279)* status: all files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-core/src/routing/unix.rs, line 274 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtrKbuV22gBEWp3mTTj:-MtxYwTo3Cr3iiyqp-lc:b-vemoxs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8bdd97124aac4f37420089fd213971b9290579bc/talpid-core/src/routing/unix.rs#L274)):*

You're right, but I think more changes were necessary anyway, as even if the daemon fails to get one of the default routes, it's ok not to error out.
</blockquote></details>

It might be ok. Say the daemon fails to obtain the IPv4 route even though it actually exists, but it correctly finds no IPv6 route. Now the offline monitor will think that there's no connectivity. So I think it is better to fail if either of them fails.
___
*[talpid-core/src/routing/unix.rs, line 283 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3279#-MtxYrV-1S0EzsjrmTva:-MtxYrV-1S0EzsjrmTvb:bdic300) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cd72cbd7fce879caeeb53ecfa56a7e5e133de5a8/talpid-core/src/routing/unix.rs#L283)):*
> ```XML
>         (Ok(v4), Ok(v6)) => Ok((v4, v6)),
>         (Ok(node), Err(err)) => {
>             log::error!("Failed to get IPv4 node: {}", err);
> ```

This should say `IPv6`.

Issue #2268, opened 2020-11-15
[Feature Request] Add flag/subcommand to `mullvad relay list` to only list hostnames


rgo3:
## Feature Request
Hi Mullvad!

Would it be possible to add a flag/sucommand to the `mullvad relay list` command in order to only list hostnames?
That would allow for easier scripting when using the cli or e.g. combine the cli with a tool like fzf.

Again happy to help if this is accepted :)

faern:
Sure. Sounds like a good thing to add. It can be a flag that just makes the output simpler to parse but contain more than just hostname. I'd suggest `-p` for parse. I'm thinking to print each relay on one line on the format:
```
{country code} {city code} {hostnam}
```
with a space between. Then it's easy to extract the wanted components in a script.

Feel free to submit a PR :+1:

BrendanThompson:
This would be incredible, especially if you could also switch the output to be `json`. That would allow for far easier filtering for automation.

cole-h:
It looks like there's already a tool that generates this in JSON: https://github.com/mullvad/mullvadvpn-app/blob/master/mullvad-rpc/src/bin/relay_list.rs

You can then use `jq` to process this easily -- to get hostnames only, something like `jq -cr '.countries[].cities[].relays[].hostname' < relays.json` would give you all hostnames on a separate line (e.g. easy stdin piping). If you want to filter for wireguard only, you could do something like: `jq -cr '.countries[].cities[].relays[] | select(.tunnels | .openvpn == []) | .hostname' < relays.json` (replace `.openvpn == []` with `.wireguard == []` for only openvpn).

---

That said, it would be nice to integrate this into `mullvad relay list` with a `--json` flag. I did a small experiment and found that the `types::RelayListCountry` in `relay.rs` is actually generated from the proto format in `mullvad-management-interface` (so it's not trivial to attach a `#[derive(Serialize)]` and use `serde-json`, unlike the `RelayListCountry` from `mullvad-types`).

Attaching `#[derive(Serialize)]` *is* possible, but it's a very manual process. I've pushed up a PR at https://github.com/mullvad/mullvadvpn-app/pull/3304 that does just that, but it may not be "up to snuff".

faern:
[...]

PR #3304, opened 2022-01-29
Add `--json` flag to `mullvad relay list`

cole-h:
This prints the list of relays in JSON so that it can be easily parsed
and manipulated without having to resort to ugly `sed` or `awk` (or what
have you) invocations.

#### What

This PR:

* adds a `--json` flag to `mullvad relay list`
* changes `mullvad-management-interface`'s proto compilation to allow for additional configuration

#### Why

* To potentially close https://github.com/mullvad/mullvadvpn-app/issues/2268 (as the functionality implemented in this PR has more general usecases than just displaying the hostname and some additional information in a specific format).
* Because I wanted to be able to do some testing based on information that is most easily accessible from `mullvad relay list` (e.g. testing latency on various endpoints)

#### How

* By adding the `serde` crate with the `derive` feature to `mullvad-management-interface` and the `serde_json` crate to `mullvad-cli`
* By adding a `json` flag to the `relay list` subcommand in order to communicate intent
* By branching on whether or not JSON output is desired -- if not, the current functionality of specially-formatted output is used
* By using `tonic_build::configure()` to allow additional type attributes (e.g. `#[derive(serde::Serialize)]`) for automatic conversion to JSON via the `serde_json` crate
* By manually specifying which types should have `Serialize` derived for them
  * This is necessary because the `Serialize` derivation fails on the `Timestamp` type (for reasons I don't know currently)

Git checklist:

  * Despite ticking the above box, I will admit I mostly skimmed the document... but I believe myself to be mostly conforming to its guidelines.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3304)


faern:
Please consider what I wrote about in https://github.com/mullvad/mullvadvpn-app/issues/2268#issuecomment-1025459443. An automatically updated relay list in JSON format that is readable by all users is already available, it's written by the daemon. Would it fit your use case to parse that instead of doing what this PR does?

What I'm not a fan of with this PR is that it adds a *third* format of the relay list. Currently the daemon fetches the relay list from our API. That's in json but it has a different format than the internal structure desired by the daemon. So the daemon processes the server list into our desired format and then it outputs it to a cache so it's persisted, that's the second format.

cole-h:
Ah, I see -- fair enough. I totally forgot about that file ^^;

Thanks for reminding me!

PR #3313, opened 2022-02-01
Call the API over bridges

dlon:
Previously, the app would use addresses supplied by the `/v1/api-addrs` endpoint in order to reach the API. This PR removes most of that behavior in favor of using bridges provided by the relay list. Currently, the implementation only supports Shadowsocks over TCP, but it can be extended to support other protocols.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)


faern (changes requested):
> Reviewed 6 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 6 of 22 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[mullvad-rpc/src/proxy.rs, line 54 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupSgB0ARGwoBoTqRxl:-MupSgB0ARGwoBoTqRxm:b-mhim6j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L54)):*
> ```XML
> 
> impl ProxyConfig {
>     /// Reads the proxy config from `CURRENT_API_CONFIG_FILENAME`.
> ```

I only see references to `CURRENT_API_CONFIG_FILENAME` in the comments, not in actual code. Is this some old leftover?
___
*[mullvad-rpc/src/proxy.rs, line 64 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupS78A8JxGGJ9_ftpV:-MupS78A8JxGGJ9_ftpW:bx20f8b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L64)):*
> ```XML
>                     error.display_chain_with_msg(stringify!(
>                         "Failed to deserialize ",
>                         CURRENT_CONFIG_FILENAME
> ```

Are you sure this works? I believe this will construct the string "Failed to deserialize CURRENT_CONFIG_FILENAME"

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 6 of 22 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[mullvad-rpc/Cargo.toml, line 38 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupU41W9ZLXd58kL0M6:-MupU41W9ZLXd58kL0M7:buw2m7w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ce59966ea373e2f129e8ec3063b4a1f7fbebb9c9/mullvad-rpc/Cargo.toml#L38)):*
> ```TOML
> 
> shadowsocks = { version = "1.12" }
> shadowsocks-crypto = { version = "0.2" }
> ```

It looks like `shadowsocks` provides `pub use shadowsocks_crypto as crypto;`. So I think you can pull in the needed types from there instead of explicitly depending on `shadowsocks-crypto`

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 6 of 22 files reviewed, 7 unresolved discussions (waiting on @dlon)

___
*[mullvad-daemon/src/api.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupZz7X8m-UqzCbt2f5:-MupZz7X8m-UqzCbt2f6:bgabk77) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-daemon/src/api.rs#L14)):*
> ```XML
> 
> pub(crate) struct MullvadProxyConfigProvider {
>     retry_attempt: Mutex<u32>,
> ```

Would maybe an `AtomicU32` be less bulky to use? The increment-and-fetch of `retry_attempt` further down looks a bit cumbersome. Atomic integers by definition wraps around on overflow, so you'll get that for free.
___
*[mullvad-daemon/src/api.rs, line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupZhp0BKXVW5EY30oR:-MupZhp0BKXVW5EY30oS:b-3cqxrx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-daemon/src/api.rs#L30)):*
> ```XML
>     fn first(&self) -> ProxyConfig {
>         let mut attempt = self.retry_attempt.lock().unwrap();
>         *attempt = 0;
> ```

It's a bit implicit that `first` is also supposed to reset the "iterator"
___
*[mullvad-daemon/src/api.rs, line 45 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mup_fvcCeHxHZRxMub6:-Mup_fvcCeHxHZRxMub7:b-9dwnxv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-daemon/src/api.rs#L45)):*
> ```XML
> 
>         let _ = self.tx.send(ApiProxyRequest {
>             response_tx: tx,
> ```

Can we do better naming here maybe? It took multiple reads of `self.tx.send` before I realized you don't send on the channel you created in this method itself. Having multiple channels and calling them all `tx`/`rx` is hard to read. How about `self.daemon_sender`?
___
*[mullvad-setup/src/main.rs, line 192 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupauO-03BSh10rw0Fr:-MupauO-03BSh10rw0Fs:bkh960) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-setup/src/main.rs#L192)):*
> ```XML
>                                 error.display_chain_with_msg("Failed to read API endpoint cache")
>                             );
>                             ProxyConfig::Tls
> ```

I think this chunk of new code could be better explained in a standalone method/function than embedded inside `remove_wireguard_key`. In general I think we need to work more on having smaller more self explanatory functions instead of huge behemoths that do many unrelated operations.

A separate

```rust
/// Reads the cache and creates a proxy provider. Falls back to plain TLS
/// directly towards the API if it fails to initialize from cache.
async fn create_proxy_provider(cache_path) -> impl ProxyConfigProvider {
    ....
}
```
Would be easier to read

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 6 of 22 files reviewed, 9 unresolved discussions (waiting on @dlon)

___
*[mullvad-rpc/src/proxy.rs, line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupcEBHC9skiZ3Ov-yL:-MupcEBHC9skiZ3Ov-yM:bw7bdhg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L25)):*
> ```XML
> pub enum ProxyConfig {
>     /// Connect directly to the target.
>     Tls,
> ```

"Proxy = no proxy" feels like a strange variant to me :thinking:. How is this enum better than having `Option<ProxyConfigSettings>` to describe direct connection vs proxy?
___
*[mullvad-rpc/src/proxy.rs, line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupcRw61GSW5Q48LNyj:-MupcRw61GSW5Q48LNyk:b7z64rd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L34)):*
> ```XML
>         match self {
>             ProxyConfig::Tls => write!(f, "unproxied"),
>             ProxyConfig::Proxied(settings) => write!(f, "{}", settings),
> ```

Nitpick: `settings.fmt(f)` would be a shorter and more idiomatic way of writing this.
___
*[mullvad-rpc/src/proxy.rs, line 91 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mupd9ow4_bB-04-YKGG:-Mupd9ow4_bB-04-YKGH:b-o6x8xh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L91)):*
> ```XML
>         file.sync_data().await?;
> 
>         fs::rename(&temp_path, path).await
> ```

Should we not close `file` before we attempt to move the file? Even if not strictly needed that feels cleaner. If you just put the writing into a separate method `fn write(&self, path)` then you get that automatically thanks to the scope, and the methods/operations become even more focused/cleaner on one thing only.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 6 of 22 files reviewed, 10 unresolved discussions (waiting on @dlon)

___
*[mullvad-rpc/src/proxy.rs, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupfKNm6cW4vNxeJyUi:-MupfKNm6cW4vNxeJyUj:b4gkoui) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L80)):*
> ```XML
> 
>     /// Stores this config to `CURRENT_API_CONFIG_FILENAME`.
>     pub async fn save(&self, cache_dir: &Path) -> io::Result<()> {
> ```

Given the implementation I suppose the intent here is to write to the file as atomically as possible. So readers of the file should not end up with partial content, right? I think the documentation for the method should mention that.

I know this has been mentioned regarding other files also. That are "published" for a wider audience. If so, it would be cool to have some wrapper type `MoveOnCloseFile` that's some sort of atomic write file. But that's outside the scope of this PR of course.
___
*[mullvad-setup/src/main.rs, line 192 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupauO-03BSh10rw0Fr:-Mupm9YJEXDsgzfPzF9i:bglfamj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-setup/src/main.rs#L192)):*

I think this chunk of new code could be better explained in a standalone method/function than embedded inside `remove_wireguard_key`. In general I think we need to work more on having smaller more self explanatory functions instead of huge behemoths that do many unrelated operations.

A separate

```rust
/// Reads the cache and creates a proxy provider. Falls back to plain TLS
/// directly towards the API if it fails to initialize from cache.
async fn create_proxy_provider(cache_path) -> impl ProxyConfigProvider {
    ....
}
```
Would be easier to read
</blockquote></details>

Feels like this exact chunk of code exists in the problem report binary as well. Could there not be a helper function in `mullvad-rpc` to `load_from_cache_or_fall_back_to_tls` (but with a better name)?



_Code quote:_
```XML
ProxyConfigProviderNoop
```

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 6 of 22 files reviewed, 11 unresolved discussions (waiting on @dlon)

___
*[mullvad-rpc/src/proxy.rs, line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupcEBHC9skiZ3Ov-yL:-Mupn2GQCUg9YhdyF2Fo:bh02ca9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L25)):*

"Proxy = no proxy" feels like a strange variant to me :thinking:. How is this enum better than having `Option<ProxyConfigSettings>` to describe direct connection vs proxy?
</blockquote></details>

I guess because it implements loading and storing to disk, and we can't attach methods to `Option`. But then the name feels a bit off. What do you think about something like
```rust
pub enum ApiConnectionMode {
    Direct,
    Proxied(ProxySettings),
}
```

I'm a bit divided on the proxied/obfuscated part. It's a proxy of sorts. But the point of it is to get obfuscation. So maybe that's a more direct description of what's going on?
___
*[mullvad-rpc/src/proxy.rs, line 40 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupnjAW1X_2H-7RZ0xw:-MupnjAW1X_2H-7RZ0xx:b9bbj0l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L40)):*
> ```XML
> 
> #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
> pub enum ProxyConfigSettings {
> ```

`ConfigSettings` is pretty repetitive. What's the difference between these words? To me this is a code smell that some layer here is not optimally named. See comment above on maybe calling `ProxyConfig` => `ApiConnectionMode` or similar. Since it does not describe the proxy, it describes **if a proxy should be used or not**. So it's not a proxy config, it's an API connection config/mode

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 6 of 22 files reviewed, 11 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/api.rs, line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupZhp0BKXVW5EY30oR:-Muqs5bW3Geh2dSJ_he2:b-a745x0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-daemon/src/api.rs#L30)):*

It's a bit implicit that `first` is also supposed to reset the "iterator"
</blockquote></details>

Maybe `init` would make more sense?

I'm tempted to dispense with the trait altogether in favor of `Stream<Item = ProxyConfig>`. Probably using `stream::unfold` to avoid having to implement the poll methods.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 6 of 22 files reviewed, 11 unresolved discussions (waiting on @dlon)

___
*[mullvad-daemon/src/api.rs, line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupZhp0BKXVW5EY30oR:-MutnTGkB89WFfflGvf0:b6en947) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-daemon/src/api.rs#L30)):*

Maybe `init` would make more sense?

I'm tempted to dispense with the trait altogether in favor of `Stream<Item = ProxyConfig>`. Probably using `stream::unfold` to avoid having to implement the poll methods.
</blockquote></details>

If we can make due with existing types instead of inventing our own semi-iterators I think that would indeed be cleaner!

faern (commented):
> Reviewed 1 of 22 files at r1, 2 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 9 of 22 files reviewed, 10 unresolved discussions (waiting on @dlon)

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 9 of 22 files reviewed, 11 unresolved discussions (waiting on @dlon)

___
*[mullvad-rpc/src/proxy.rs, line 125 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MuttRpE1O0ZHGXlJt4K:-MuttRpE1O0ZHGXlJt4L:b-j9unrm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L125)):*
> ```XML
> }
> 
> pub struct ProxyConfigProviderNoop(pub ProxyConfig);
> ```

This implementation just repeats the same `ProxyConfig` over and over. I think `Repeat` is a better name than `Noop` (see https://docs.rs/futures/latest/futures/stream/fn.repeat.html for reference). If you manage to get rid of the trait, as you have mentioned, this type can also go away in favor of the `Repeat` type I linked to.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 8 of 22 files reviewed, 10 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/api.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupZz7X8m-UqzCbt2f5:-MuzyuF1-lO0wVEXQKGt:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-daemon/src/api.rs#L14)):*

Would maybe an `AtomicU32` be less bulky to use? The increment-and-fetch of `retry_attempt` further down looks a bit cumbersome. Atomic integers by definition wraps around on overflow, so you'll get that for free.
</blockquote></details>

Done.
___
*[mullvad-daemon/src/api.rs, line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupZhp0BKXVW5EY30oR:-Muzyy8A2gkCFPu_dgoL:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-daemon/src/api.rs#L30)):*

If we can make due with existing types instead of inventing our own semi-iterators I think that would indeed be cleaner!
</blockquote></details>

Done.
___
*[mullvad-daemon/src/api.rs, line 45 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mup_fvcCeHxHZRxMub6:-Muzz2Kj6JP_5hNKkJMh:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-daemon/src/api.rs#L45)):*

Can we do better naming here maybe? It took multiple reads of `self.tx.send` before I realized you don't send on the channel you created in this method itself. Having multiple channels and calling them all `tx`/`rx` is hard to read. How about `self.daemon_sender`?
</blockquote></details>

Done.
___
*[mullvad-rpc/Cargo.toml, line 38 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupU41W9ZLXd58kL0M6:-MuzymHE-Sa9BXk5IeVN:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ce59966ea373e2f129e8ec3063b4a1f7fbebb9c9/mullvad-rpc/Cargo.toml#L38)):*

It looks like `shadowsocks` provides `pub use shadowsocks_crypto as crypto;`. So I think you can pull in the needed types from there instead of explicitly depending on `shadowsocks-crypto`
</blockquote></details>

Done.
___
*[mullvad-rpc/src/proxy.rs, line 54 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupSgB0ARGwoBoTqRxl:-MuzzyZl7Cm2YLnbEkbZ:b-5bl2jp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L54)):*

I only see references to `CURRENT_API_CONFIG_FILENAME` in the comments, not in actual code. Is this some old leftover?
</blockquote></details>

Done. It was renamed.
___
*[mullvad-rpc/src/proxy.rs, line 125 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MuttRpE1O0ZHGXlJt4K:-Mv-1FIZ6SpzMcIoOxo4:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L125)):*

This implementation just repeats the same `ProxyConfig` over and over. I think `Repeat` is a better name than `Noop` (see https://docs.rs/futures/latest/futures/stream/fn.repeat.html for reference). If you manage to get rid of the trait, as you have mentioned, this type can also go away in favor of the `Repeat` type I linked to.
</blockquote></details>

Done.
___
*[mullvad-setup/src/main.rs, line 192 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupauO-03BSh10rw0Fr:-Mv-1ZFL2LfM7qWF2bGH:bmalu7v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-setup/src/main.rs#L192)):*

Feels like this exact chunk of code exists in the problem report binary as well. Could there not be a helper function in `mullvad-rpc` to `load_from_cache_or_fall_back_to_tls` (but with a better name)?
</blockquote></details>

This is what `ProxyConfig::try_from_cache` does now.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 8 of 22 files reviewed, 10 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-rpc/src/proxy.rs, line 64 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupS78A8JxGGJ9_ftpV:-Mv--y4z6ZlbVZTyrfJu:bpexld7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L64)):*

Are you sure this works? I believe this will construct the string "Failed to deserialize CURRENT_CONFIG_FILENAME"
</blockquote></details>

It definitely does not work! Fixed.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 8 of 22 files reviewed, 8 unresolved discussions (waiting on @faern)

___
*[mullvad-rpc/src/proxy.rs, line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupcEBHC9skiZ3Ov-yL:-Mv-DB831k6BJZeHj2E_:b-7v6gwd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L25)):*

I guess because it implements loading and storing to disk, and we can't attach methods to `Option`. But then the name feels a bit off. What do you think about something like
```rust
pub enum ApiConnectionMode {
    Direct,
    Proxied(ProxySettings),
}
```

I'm a bit divided on the proxied/obfuscated part. It's a proxy of sorts. But the point of it is to get obfuscation. So maybe that's a more direct description of what's going on?
</blockquote></details>

That's definitely a better name. I'm also not sure whether "obfuscated" or "proxied" is better, so I've left it unchanged for now.
___
*[mullvad-rpc/src/proxy.rs, line 40 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupnjAW1X_2H-7RZ0xw:-Mv-BU4y0gNDpfszV9Zg:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L40)):*

`ConfigSettings` is pretty repetitive. What's the difference between these words? To me this is a code smell that some layer here is not optimally named. See comment above on maybe calling `ProxyConfig` => `ApiConnectionMode` or similar. Since it does not describe the proxy, it describes **if a proxy should be used or not**. So it's not a proxy config, it's an API connection config/mode
</blockquote></details>

Done.
___
*[mullvad-rpc/src/proxy.rs, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MupfKNm6cW4vNxeJyUi:-Mv-C1_a0OFgsXB2n57j:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L80)):*

Given the implementation I suppose the intent here is to write to the file as atomically as possible. So readers of the file should not end up with partial content, right? I think the documentation for the method should mention that.

I know this has been mentioned regarding other files also. That are "published" for a wider audience. If so, it would be cool to have some wrapper type `MoveOnCloseFile` that's some sort of atomic write file. But that's outside the scope of this PR of course.
</blockquote></details>

Done.
___
*[mullvad-rpc/src/proxy.rs, line 91 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mupd9ow4_bB-04-YKGG:-Mv-CbmmDfCPDPz20RXa:b-sdaypc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L91)):*

Should we not close `file` before we attempt to move the file? Even if not strictly needed that feels cleaner. If you just put the writing into a separate method `fn write(&self, path)` then you get that automatically thanks to the scope, and the methods/operations become even more focused/cleaner on one thing only.
</blockquote></details>

True. I've moved that code to its own block. Do you think a separate method would be preferable anyway?

pinkisemils (changes requested):
> Reviewed 3 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 9 of 22 files reviewed, 10 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-rpc/src/https_client_with_sni.rs, line 55 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3bsH42RlLUId9dvqE:-Mv3bsH42RlLUId9dvqF:bji4zp1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/https_client_with_sni.rs#L55)):*
> ```XML
> 
>     /// Stop all streams produced by this connector
>     pub fn set_proxy(&self, proxy: ApiConnectionMode) {
> ```

Would be better if it was named `set_connection_config` or `set_connection_mode`
___
*[mullvad-rpc/src/https_client_with_sni.rs, line 133 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3c_iTAkGQeVJpY8DO:-Mv3c_iTAkGQeVJpY8DP:bpqfgta) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/https_client_with_sni.rs#L133)):*
> ```XML
>         let inner_copy = inner.clone();
>         let notify = abort_notify.clone();
>         handle.spawn(async move {
> ```

You probably want to use `tokio::spawn` here.
___
*[mullvad-rpc/src/proxy.rs, line 152 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3X5GUAXUMAbk8Sslh:-Mv3X5GUAXUMAbk8Ssli:b-5xtoo9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/proxy.rs#L152)):*
> ```XML
> 
> /// Stream that is either a regular TLS stream or TLS via shadowsocks
> pub enum MaybeProxyStream {
> ```

`MaybeProxyStream` isn't the best name for what is used and works exactly like a TCP connection. I think a `WebConnection` or just `Connection` might be a better name.
___
*[mullvad-rpc/src/rest.rs, line 202 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3bmd-0nWykuWK2lBU:-Mv3bmd-0nWykuWK2lBV:b-aj6dwb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/rest.rs#L202)):*
> ```XML
>             RequestCommand::NextApiConfig => {
>                 if let Some(new_config) = self.proxy_config_provider.next().await {
>                     self.connector_handle.set_proxy(new_config);
> ```

What happens to in-flight requests when a new config is set?

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 9 of 22 files reviewed, 10 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-rpc/src/https_client_with_sni.rs, line 72 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3dOjd3Wbp9WC3eG8J:-Mv3dOjd3Wbp9WC3eG8K:b1iwpe8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/https_client_with_sni.rs#L72)):*
> ```XML
>     Direct,
>     /// Connect to the destination via a proxy.
>     Proxied(ServerConfig),
> ```

Instead of storing a shadowsocks type, you could store a struct that only contains the data we can have so that later there is no need to fail if the config contains a hostname address instead of a socket address and the like.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 9 of 22 files reviewed, 10 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[mullvad-rpc/src/rest.rs, line 202 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3bmd-0nWykuWK2lBU:-Mv3spts78x0QiNuNcE9:b-f0s6lk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/rest.rs#L202)):*

What happens to in-flight requests when a new config is set?
</blockquote></details>

Good point. They'll probably time out. I'll check if that's the case and call `reset` as well.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 9 of 22 files reviewed, 5 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[mullvad-daemon/src/relays/mod.rs, line 259 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv425mD6I8xAyrPyQKT:-Mv425mD6I8xAyrPyQKU:bjpbu9i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-daemon/src/relays/mod.rs#L259)):*
> ```XML
>     /// Returns the average location of relays that match the given constraints.
>     /// This returns none if the location is `any` or if no relays match the constraints.
>     pub fn get_relay_midpoint(&self, relay_constraints: &RelayConstraints) -> Option<Coordinates> {
> ```

I don't follow when we need this :thinking: When do we want the midpoint of a number of relays?
___
*[mullvad-rpc/src/https_client_with_sni.rs, line 55 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3bsH42RlLUId9dvqE:-Mv3yRll1A_ovGjfbphw:b-glpbdy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/https_client_with_sni.rs#L55)):*

Would be better if it was named `set_connection_config` or `set_connection_mode`
</blockquote></details>

I agree. Same goes for various enums and variants below as well.

* `SetProxy` -> `SetConnectionMode`
* `InnerProxyConfig` -> `InnerConnectionMode`
___
*[mullvad-rpc/src/https_client_with_sni.rs, line 133 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3c_iTAkGQeVJpY8DO:-Mv3zTt9EbHl32O6AWbu:b-4g9nzq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/https_client_with_sni.rs#L133)):*

You probably want to use `tokio::spawn` here.
</blockquote></details>

If so, we can remove the entire `handle` argument. This is the only place I see it used. And that's completely in line with how I think we should design more. Assume a function is called on the relevant runtime instead of passing a handle around.
___
*[mullvad-rpc/src/proxy.rs, line 91 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mupd9ow4_bB-04-YKGG:-Mv4-ufrCZBlUZhROd9M:bs6uyl6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a32f1a3c2bfddd2a1fc916a44f10ee3161535f4c/mullvad-rpc/src/proxy.rs#L91)):*

True. I've moved that code to its own block. Do you think a separate method would be preferable anyway?
</blockquote></details>

This works! Very nice with the random filename as well to even further avoid collisions :)
___
*[mullvad-rpc/src/proxy.rs, line 152 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3X5GUAXUMAbk8Sslh:-Mv40UnM62Y_VAU5u_oW:bpp4oex) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/proxy.rs#L152)):*

`MaybeProxyStream` isn't the best name for what is used and works exactly like a TCP connection. I think a `WebConnection` or just `Connection` might be a better name.
</blockquote></details>

I agree with Emils. And I think that the first variant should maybe be called `Direct` to be in line with what the corresponding variant is called in all the other similar enums?
___
*[mullvad-setup/src/main.rs, line 177 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv41N7d8E3x2bWaFNTc:-Mv41N7d8E3x2bWaFNTd:bru7ktz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-setup/src/main.rs#L177)):*
> ```XML
>     if let Some(token) = settings.get_account_token() {
>         if let Some(wg_data) = settings.get_wireguard() {
>             let mut rpc_runtime = MullvadRpcRuntime::with_cache(&cache_path, false)
> ```

Here we initialize the RPC runtime with a cache. And then further down where we get the REST handle we also load from the cache. Should this logic really be split like this? It feels like either we should not use one of the caches at all, or they should be loaded at the same time/by the same object?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 8 of 23 files reviewed, 5 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-daemon/src/relays/mod.rs, line 259 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv425mD6I8xAyrPyQKT:-Mv4PUED5O0xptVEbIEB:b-mq84c7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-daemon/src/relays/mod.rs#L259)):*

I don't follow when we need this :thinking: When do we want the midpoint of a number of relays?
</blockquote></details>

When connecting using bridge mode, a specific relay is first selected and then the closest bridge to that relay is selected. But here we want the closest bridge given a location *constraint*, not the bridge closest to some specific relay.
___
*[mullvad-rpc/src/https_client_with_sni.rs, line 133 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3c_iTAkGQeVJpY8DO:-MvIOmJTDjuwDxrckX9F:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/https_client_with_sni.rs#L133)):*

If so, we can remove the entire `handle` argument. This is the only place I see it used. And that's completely in line with how I think we should design more. Assume a function is called on the relevant runtime instead of passing a handle around.
</blockquote></details>

Done.
___
*[mullvad-rpc/src/proxy.rs, line 152 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3X5GUAXUMAbk8Sslh:-MvIOp2qBHX00Z2e7QMv:b-f1xuup) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/proxy.rs#L152)):*

I agree with Emils. And I think that the first variant should maybe be called `Direct` to be in line with what the corresponding variant is called in all the other similar enums?
</blockquote></details>

Done. Is this better or too specific?
___
*[mullvad-setup/src/main.rs, line 177 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv41N7d8E3x2bWaFNTc:-Mv4XQ_10awKAK1ISx6F:bvo0nrg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-setup/src/main.rs#L177)):*

Here we initialize the RPC runtime with a cache. And then further down where we get the REST handle we also load from the cache. Should this logic really be split like this? It feels like either we should not use one of the caches at all, or they should be loaded at the same time/by the same object?
</blockquote></details>

We use the API IP cache in the daemon, but never load the "API connection mode" file. For the problem report tool, etc., both are loaded. That's why there's a split.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 8 of 25 files reviewed, 5 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-rpc/src/rest.rs, line 202 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3bmd-0nWykuWK2lBU:-MvJ7HxK8Ck3lwkwCs4A:btf3oem) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/rest.rs#L202)):*

Good point. They'll probably time out. I'll check if that's the case and call `reset` as well.
</blockquote></details>

In the above case, in-flight requests were aborted because the socket was closed. But the error was not mapped to `Error::Aborted`, so it could cause the client to switch to yet another bridge for no reason.

I've fixed the above, and I was also able to get rid of the `BTreeMap` containing all in-flight requests and simply rely on the socket being closed (i.e., calling `connector_handle.reset()`) to cancel all the requests. (I'm not sure why this didn't work before.)

pinkisemils (commented):
> Reviewed 1 of 7 files at r8.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 9 of 26 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[mullvad-rpc/src/proxy.rs, line 152 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3X5GUAXUMAbk8Sslh:-MvJAtg1-sJUdLj74f5G:bf3phwp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/proxy.rs#L152)):*

Done. Is this better or too specific?
</blockquote></details>

I think this is good. I'm a bit unsure about the _Api_ bit, but I calling it a web or a tcp connection also feels wrong, so I'm OK with leaving it as is :thumbsup: 
___
*[mullvad-rpc/src/rest.rs, line 202 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3bmd-0nWykuWK2lBU:-MvJB8u-1BJHjlcYAdvx:bla6gbx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/rest.rs#L202)):*

In the above case, in-flight requests were aborted because the socket was closed. But the error was not mapped to `Error::Aborted`, so it could cause the client to switch to yet another bridge for no reason.

I've fixed the above, and I was also able to get rid of the `BTreeMap` containing all in-flight requests and simply rely on the socket being closed (i.e., calling `connector_handle.reset()`) to cancel all the requests. (I'm not sure why this didn't work before.)
</blockquote></details>

Great work :tada:

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 9 of 26 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/relays/mod.rs, line 259 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv425mD6I8xAyrPyQKT:-MvOky7oAK4AMsTrvCh3:bjyu97a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-daemon/src/relays/mod.rs#L259)):*

When connecting using bridge mode, a specific relay is first selected and then the closest bridge to that relay is selected. But here we want the closest bridge given a location *constraint*, not the bridge closest to some specific relay.
</blockquote></details>

It looks like the current implementation would place the midpoint very close to datacenter `A` if both datacenter `A` and `B` have relays matching the constraint but `A` has way more servers? Is it pulling in the midpoint by having more relays in the same location? That's a bit strange.

Also, is this not going to give a very strange midpoint if the location is set (via the CLI) to `Any`? Then it's all servers in the entire world. The midpoint here makes very little sense :thinking:

Once again I feel like our post-processing of the relay list is working against us. It's only the app that has locations per relay. The API relay list has the locations separate from the relays and it would then be easier to get coordinates *per location* instead of per relay.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 9 of 26 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[mullvad-daemon/src/relays/mod.rs, line 259 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv425mD6I8xAyrPyQKT:-MvOo5NZEiCDGkDT4s-e:be6ukgz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-daemon/src/relays/mod.rs#L259)):*
> Is it pulling in the midpoint by having more relays in the same location? That's a bit strange.

Yes. Per-city coordinates are stored in the processed list as well. So duplicate cities could easily be filtered out.

> Also, is this not going to give a very strange midpoint if the location is set (via the CLI) to `Any`?

No. It simply returns `None` (i.e., does not use a bridge) in that special case. It might be more correct to just pick a random bridge, though.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 9 of 26 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/relays/mod.rs, line 259 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv425mD6I8xAyrPyQKT:-MvOt8r98PyI61qmSzTo:bfixmey) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-daemon/src/relays/mod.rs#L259)):*

> Is it pulling in the midpoint by having more relays in the same location? That's a bit strange.

Yes. Per-city coordinates are stored in the processed list as well. So duplicate cities could easily be filtered out.

> Also, is this not going to give a very strange midpoint if the location is set (via the CLI) to `Any`?

No. It simply returns `None` (i.e., does not use a bridge) in that special case. It might be more correct to just pick a random bridge, though.
</blockquote></details>

Yeah, having `Any` location constraint fall back to a random bridge is better than not being able use the bridged API mode at all.
___
*[mullvad-setup/src/main.rs, line 177 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv41N7d8E3x2bWaFNTc:-MvOuS6R4KIRbPds0kSr:b-wbunvp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-setup/src/main.rs#L177)):*

We use the API IP cache in the daemon, but never load the "API connection mode" file. For the problem report tool, etc., both are loaded. That's why there's a split.
</blockquote></details>

Ah. I understand.
___
*[mullvad-types/src/location.rs, line 57 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MvOu8zkD8OLX9Rr0T2q:-MvOu8zkD8OLX9Rr0T2r:bswso7j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4cff6d3513e26c67da75d1bb7bcbcb2c8ab883a4/mullvad-types/src/location.rs#L57)):*
> ```XML
> impl Coordinates {
>     /// Computes the approximate midpoint of a set of locations
>     pub fn midpoint(locations: &[Location]) -> Self {
> ```

Does this midpoint algorithm have a name? If so, can we include that in the documentation and maybe even link to the corresponding wikipedia page or something? So it's possible to compare the implementation with some reference implementations and research exactly what "approximate" here means etc.

faern (changes requested):
> Reviewed 1 of 7 files at r8, 1 of 3 files at r9, 1 of 1 files at r11.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 12 of 26 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/lib.rs, line 929 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MvOwLPQBtiLvGHcyzCm:-MvOwLPREz4ih5OafvKe:birp6m5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4cff6d3513e26c67da75d1bb7bcbcb2c8ab883a4/mullvad-daemon/src/lib.rs#L929)):*
> ```XML
>             }
>             GenerateApiProxyConfig(request) => {
>                 self.handle_generate_api_connection_mode(request).await
> ```

The enum variant and the corresponding `handle_` method has different naming. I think we should try to keep them in sync to make it easier to follow along the code.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 12 of 26 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-types/src/location.rs, line 57 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MvOu8zkD8OLX9Rr0T2q:-MvOwvS71EtSAS6I46-p:b2dip42) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4cff6d3513e26c67da75d1bb7bcbcb2c8ab883a4/mullvad-types/src/location.rs#L57)):*

Does this midpoint algorithm have a name? If so, can we include that in the documentation and maybe even link to the corresponding wikipedia page or something? So it's possible to compare the implementation with some reference implementations and research exactly what "approximate" here means etc.
</blockquote></details>

I'm not sure if there's a name for it. This is really just converting spherical coordinates to cartesian coordinates, averaging them, and then converting them back to spherical coordinates. I suppose the docs could just explain what it's doing?

The reason it's approximate is that it assumes that the earth is a perfect sphere. But I doubt the error is large enough to matter for our purpose.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 11 of 26 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/lib.rs, line 929 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MvOwLPQBtiLvGHcyzCm:-MvPAj-f1_Xa2qU3tOVn:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4cff6d3513e26c67da75d1bb7bcbcb2c8ab883a4/mullvad-daemon/src/lib.rs#L929)):*

The enum variant and the corresponding `handle_` method has different naming. I think we should try to keep them in sync to make it easier to follow along the code.
</blockquote></details>

Done.
___
*[mullvad-rpc/src/https_client_with_sni.rs, line 72 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3dOjd3Wbp9WC3eG8J:-MvPAg_w1eWL9DbV5q8x:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/https_client_with_sni.rs#L72)):*

Instead of storing a shadowsocks type, you could store a struct that only contains the data we can have so that later there is no need to fail if the config contains a hostname address instead of a socket address and the like. 
</blockquote></details>

Done.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 11 of 26 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[mullvad-daemon/src/relays/mod.rs, line 259 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv425mD6I8xAyrPyQKT:-MvSpD1o-wyik_Fa0dR0:b-a2j5ve) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-daemon/src/relays/mod.rs#L259)):*

Yeah, having `Any` location constraint fall back to a random bridge is better than not being able use the bridged API mode at all.
</blockquote></details>

Done. Cities are only included once now, and `any` causes a random bridge to be selected.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 12 of 26 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/lib.rs, line 336 at r13](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MvSwIVMC_y-LWzCJvaS:-MvSwIVMC_y-LWzCJvaT:b4iwavl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9eefa42f325509a04fee84778cb234754734baad/mullvad-daemon/src/lib.rs#L336)):*
> ```XML
>     GenerateApiConnectionMode(api::ApiConnectionModeRequest),
>     /// Endpoint used to reach the API.
>     NewApiEndpoint(SocketAddr),
> ```

It would be great if we did not need two messages for this feature :thinking: After talking to David I understand that the current protocol is

1. mullvad-rpc asks the daemon for an ApiConnectionMode when the current one fails.
2. When mullvad-rpc gets the new connection mode it sends the `SocketAddr` to the daemon so it can unlock it in the firewall

It's a bit redundant since the daemon already should know the corresponding `SocketAddr` in all cases except for `ApiConnectionMode::Direct`... Because that IP+port is known and managed by mullvad-rpc.
___
*[mullvad-rpc/src/proxy.rs, line 152 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mv3X5GUAXUMAbk8Sslh:-MvSuXxW2IdWxE6UVhEY:brxp9sd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/23943238b39e0b2b85336bd8fb7e333eb61c5877/mullvad-rpc/src/proxy.rs#L152)):*

I think this is good. I'm a bit unsure about the _Api_ bit, but I calling it a web or a tcp connection also feels wrong, so I'm OK with leaving it as is :thumbsup: 
</blockquote></details>

Sounds good!

pinkisemils (commented):
> Reviewed 1 of 7 files at r8, 1 of 3 files at r9, 1 of 3 files at r12.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 15 of 26 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 15 of 26 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[mullvad-daemon/src/lib.rs, line 336 at r13](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MvSwIVMC_y-LWzCJvaS:-MvsKyZb7jZDomWOL5Ee:b-wzv1e6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9eefa42f325509a04fee84778cb234754734baad/mullvad-daemon/src/lib.rs#L336)):*

It would be great if we did not need two messages for this feature :thinking: After talking to David I understand that the current protocol is

1. mullvad-rpc asks the daemon for an ApiConnectionMode when the current one fails.
2. When mullvad-rpc gets the new connection mode it sends the `SocketAddr` to the daemon so it can unlock it in the firewall

It's a bit redundant since the daemon already should know the corresponding `SocketAddr` in all cases except for `ApiConnectionMode::Direct`... Because that IP+port is known and managed by mullvad-rpc.
</blockquote></details>

Some changes were made. There are still two callbacks, but:
* The `SocketAddr` callback used to be invoked only when the direct endpoint (returned by `/v1/api-addrs`) was updated. Now it is sent for bridge endpoints as well, and firewall rules are set only in this callback, not in both. The `GenerateApiConnectionMode` cmd has no real side effects, and only returns a connection mode.
* The closure returns a future. This makes the types messier, but it is more correct since the code is async on both ends.

faern (changes requested):
> Reviewed 2 of 22 files at r1, 1 of 3 files at r9, 1 of 3 files at r12.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 15 of 26 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[mullvad-daemon/src/lib.rs, line 1386 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mvx29EW7A1swtSkH0lO:-Mvx29EXAZgwRCw2_Lsl:bcyonpu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e0bdd06607da8688926718d83114c71323334e8/mullvad-daemon/src/lib.rs#L1386)):*
> ```XML
>     }
> 
>     async fn handle_generate_api_connection_mode(
> ```

Can this method be documented with a description of how it decides on a good API endpoint to use next? So it can be understood without reading all the code. And so the code has a reference description to rely on when checking for potential bugs.

I'm also not a huge fan of this being in `lib.rs`. This is already a god-module. Why is something so specific as this computed at the root of the crate? We can skip this change for now if we really take care of refactoring `lib.rs` a lot soon after.

pinkisemils (changes requested):
> Reviewed 3 of 22 files at r1, 1 of 4 files at r13.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 16 of 26 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[mullvad-daemon/src/lib.rs, line 1386 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mvx29EW7A1swtSkH0lO:-MwBmEaZCUoyTyKjErHw:b-plzpeo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e0bdd06607da8688926718d83114c71323334e8/mullvad-daemon/src/lib.rs#L1386)):*

Can this method be documented with a description of how it decides on a good API endpoint to use next? So it can be understood without reading all the code. And so the code has a reference description to rely on when checking for potential bugs.

I'm also not a huge fan of this being in `lib.rs`. This is already a god-module. Why is something so specific as this computed at the root of the crate? We can skip this change for now if we really take care of refactoring `lib.rs` a lot soon after.
</blockquote></details>

I agree that this might benefit from better documentation, but I don't have strong feelings about pushing this logic anywhere else.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 13 of 26 files reviewed, 3 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-daemon/src/lib.rs, line 1386 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-Mvx29EW7A1swtSkH0lO:-MwC2V3l0JJVrWxHukiz:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e0bdd06607da8688926718d83114c71323334e8/mullvad-daemon/src/lib.rs#L1386)):*

I agree that this might benefit from better documentation, but I don't have strong feelings about pushing this logic anywhere else. 
</blockquote></details>

Done.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 12 of 26 files reviewed, 3 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-types/src/location.rs, line 57 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MvOu8zkD8OLX9Rr0T2q:-MwC9STE5ZNvDQegdOmn:bnu8ujv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4cff6d3513e26c67da75d1bb7bcbcb2c8ab883a4/mullvad-types/src/location.rs#L57)):*

I'm not sure if there's a name for it. This is really just converting spherical coordinates to cartesian coordinates, averaging them, and then converting them back to spherical coordinates. I suppose the docs could just explain what it's doing?

The reason it's approximate is that it assumes that the earth is a perfect sphere. But I doubt the error is large enough to matter for our purpose.
</blockquote></details>

Is this clear enough?

pinkisemils (commented):
> Reviewed 2 of 8 files at r14, 4 of 8 files at r15.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 18 of 26 files reviewed, 3 unresolved discussions (waiting on @faern and @pinkisemils)

pinkisemils (commented):
> Reviewed 2 of 22 files at r1, 2 of 8 files at r14, 2 of 8 files at r15.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 22 of 26 files reviewed, 3 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-rpc/src/rest.rs, line 111 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MwR8iFQAVuMtLt2RLZl:-MwR8iFQAVuMtLt2RLZm:bl6rak7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6233dd1a8427110a7c252d25d2919ff1f8d1cee1/mullvad-rpc/src/rest.rs#L111)):*
> ```XML
>     T: Stream<Item = ApiConnectionMode>,
>     F: Fn(SocketAddr) -> Fut,
>     Fut: Future<Output = std::result::Result<(), ()>>,
> ```

Does this future need to return a result if in both cases the result carries no information?   If the callback gets a more descriptive name, it might as well just return a boolean indicating whether the socket address is useable or not.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 22 of 26 files reviewed, 2 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[mullvad-daemon/src/lib.rs, line 1455 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MwRXSWv7Xzijz_-V23-:-MwRXSWv7Xzijz_-V230:bbhiayi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6233dd1a8427110a7c252d25d2919ff1f8d1cee1/mullvad-daemon/src/lib.rs#L1455)):*
> ```XML
>                 }
>                 _ => {
>                     log::error!("Received unexpected proxy settings type");
> ```

Is it not better to leave out this catch all branch? Then we'll get a compilation error whenever we add another proxy mode instead of silently compiling and getting a runtime error.
___
*[mullvad-rpc/src/rest.rs, line 111 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MwR8iFQAVuMtLt2RLZl:-MwROvMKBUKk_ouFV6W1:bsllj62) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6233dd1a8427110a7c252d25d2919ff1f8d1cee1/mullvad-rpc/src/rest.rs#L111)):*

Does this future need to return a result if in both cases the result carries no information?   If the callback gets a more descriptive name, it might as well just return a boolean indicating whether the socket address is useable or not.
</blockquote></details>

I have not read the full context yet. But using `Result<(), ()>` instead of a bool has the advantage of it being obvious which value is the error. A boolean could either represent `it_failed` or `it_worked`.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 22 of 26 files reviewed, 2 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[mullvad-rpc/src/rest.rs, line 110 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MwVpdyoDdrqAOoF4l3W:-MwVpdyoDdrqAOoF4l3X:b-5iyp8m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6233dd1a8427110a7c252d25d2919ff1f8d1cee1/mullvad-rpc/src/rest.rs#L110)):*
> ```XML
> pub(crate) struct RequestService<
>     T: Stream<Item = ApiConnectionMode>,
>     F: Fn(SocketAddr) -> Fut,
> ```

I will probably switch to using `impl Fn(SocketAddr) -> Fut` instead of a type parameter here.
___
*[mullvad-rpc/src/rest.rs, line 111 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MwR8iFQAVuMtLt2RLZl:-MwVoi4W98LxTIIdk8nv:b-q0nrrh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6233dd1a8427110a7c252d25d2919ff1f8d1cee1/mullvad-rpc/src/rest.rs#L111)):*

I have not read the full context yet. But using `Result<(), ()>` instead of a bool has the advantage of it being obvious which value is the error. A boolean could either represent `it_failed` or `it_worked`.
</blockquote></details>

If it were named `AcceptedAddress`, or something like that, instead of `Fut`, then maybe a boolean would be sufficient.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 17 of 26 files reviewed, 2 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[mullvad-daemon/src/lib.rs, line 1455 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313#-MwRXSWv7Xzijz_-V23-:-MwX8I9T2vLt4ZDBnB9L:b-iyjcz5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6233dd1a8427110a7c252d25d2919ff1f8d1cee1/mullvad-daemon/src/lib.rs#L1455)):*

Is it not better to leave out this catch all branch? Then we'll get a compilation error whenever we add another proxy mode instead of silently compiling and getting a runtime error.
</blockquote></details>

`ProxySettings` also contains the `Local` and `Remote` variants (i.e., for custom proxies). But these are never returned by the relay selector. (You could argue that it should return a different type.)

pinkisemils (approved):
> Reviewed 1 of 8 files at r14, 2 of 8 files at r15, 6 of 6 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: all files reviewed, 2 unresolved discussions (waiting on @faern)

dlon (commented):
> Dismissed @faern from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: 24 of 27 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

pinkisemils (approved):
> Reviewed 2 of 3 files at r17, 1 of 1 files at r18, 1 of 1 files at r20, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3313)* status: all files reviewed, 1 unresolved discussion

PR #3314, opened 2022-02-01
Fix port confusion when picking WG endpoints

pinkisemils:
The relay selector was buggy in three ways - when selecting an entry endpoint for multi-hop, it would consider both TCP and UDP endpoints and when trying to construct a valid configuration, the endpoint protocol constraint wouldn't be taken into consideration and finally, the config construction would assume that only UDP endpoints have been selected if the transport protocol was not specified. Thus, when using multi-hop the relay selector could end up picking a UDP configuration with TCP ports.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3314)

### DIFF
--- a/mullvad-daemon/src/relays/matcher.rs
+++ b/mullvad-daemon/src/relays/matcher.rs
@@ -8,7 +8,7 @@ use mullvad_types::{
 };
 use rand::{seq::SliceRandom, Rng};
 use std::net::{IpAddr, SocketAddr};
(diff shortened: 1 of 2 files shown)


mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3314)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3321, opened 2022-02-02
Remove upstart config file

pinkisemils:
We've had an upstart configuration file since forever, yet in the current master, I don't think this file is used by anything or referenced anywhere. As such, I think we can safely remove it.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3321)

### DIFF
--- a/dist-assets/linux/mullvad-daemon.conf
+++ b/dist-assets/linux/mullvad-daemon.conf
@@ -1,6 +1,0 @@
-# Upstart job configuration file for the Mullvad VPN daemon
-
-start on local-filesystems and net-device-up IFACE!=lo
-respawn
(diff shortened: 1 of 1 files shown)


faern:
It was added in #396. And as far as I can see it's still referenced in https://github.com/mullvad/mullvadvpn-app/blob/master/dist-assets/linux/after-install.sh#L10 and https://github.com/mullvad/mullvadvpn-app/blob/master/gui/tasks/distribution.js#L167 at least

dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3321)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3322, opened 2022-02-02
Forward device arrival and removal window messages that are only received by users (not services)

dlon:
This makes it possible to detect, for example, VeraCrypt volumes.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ Line wrap the file at 100 chars.                                              Th
 - Add malware blocking to the desktop app. Implemented via DNS on the relays.
 - Add changes dialog which will include the most notable changes in each new version.
 
+#### Windows
+- Detect mounting and dismounting of volumes, such as VeraCrypt volumes or USB drives,
+  and exclude paths from the tunnel correctly when these occur. This sometimes only works
+  when the GUI frontend is running.
+
 ### Changed
 - Keep unspecified constraints unchanged in the CLI when providing specific tunnel constraints
   instead of setting them to default values.
--- a/gui/src/main/daemon-rpc.ts
+++ b/gui/src/main/daemon-rpc.ts
@@ -501,6 +501,10 @@ export class DaemonRpc {
     await this.callBool(this.client.setSplitTunnelState, enabled);
   }
 
+  public async checkVolumes(): Promise<void> {
+    await this.callEmpty(this.client.checkVolumes);
+  }
+
   private subscriptionId(): number {
     const current = this.nextSubscriptionId;
     this.nextSubscriptionId += 1;
--- a/gui/src/main/index.ts
+++ b/gui/src/main/index.ts
@@ -1857,6 +1857,20 @@ class ApplicationMain {
           // https://github.com/electron/electron/blob/main/docs/faq.md#the-font-looks-blurry-what-is-this-and-what-can-i-do
           backgroundColor: '#fff',
         });
+        const WM_DEVICECHANGE = 0x0219;
+        const DBT_DEVICEARRIVAL = 0x8000;
+        const DBT_DEVICEREMOVECOMPLETE = 0x8004;
+        appWindow.hookWindowMessage(WM_DEVICECHANGE, (wParam) => {
+          const wParamL = wParam.readBigInt64LE(0);
+          if (wParamL != DBT_DEVICEARRIVAL && wParamL != DBT_DEVICEREMOVECOMPLETE) {
+            return;
+          }
+          this.daemonRpc
+            .checkVolumes()
+            .catch((error) =>
+              log.error(`Unable to notify daemon of device event: ${error.message}`),
+            );
+        });
 
         appWindow.removeMenu();
 
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -293,6 +293,9 @@ pub enum DaemonCommand {
     /// Toggle wireguard-nt on or off
     #[cfg(target_os = "windows")]
     UseWireGuardNt(ResponseTx<(), Error>, bool),
+    /// Notify the split tunnel monitor that a volume was mounted or dismounted
+    #[cfg(target_os = "windows")]
+    CheckVolumes(ResponseTx<(), Error>),
     /// Makes the daemon exit the main loop and quit.
     Shutdown,
     /// Saves the target tunnel state and enters a blocking state. The state is restored
@@ -546,6 +549,8 @@ pub struct Daemon<L: EventListener> {
     shutdown_tasks: Vec<Pin<Box<dyn Future<Output = ()>>>>,
     /// oneshot channel that completes once the tunnel state machine has been shut down
     tunnel_state_machine_shutdown_signal: oneshot::Receiver<()>,
+    #[cfg(target_os = "windows")]
+    volume_update_tx: mpsc::UnboundedSender<()>,
 }
 
 impl<L> Daemon<L>
@@ -627,6 +632,8 @@ where
             Self::get_allowed_endpoint(rpc_runtime.address_cache.peek_address());
 
         let (offline_state_tx, offline_state_rx) = mpsc::unbounded();
+        #[cfg(target_os = "windows")]
+        let (volume_update_tx, volume_update_rx) = mpsc::unbounded();
         let tunnel_command_tx = tunnel_state_machine::spawn(
             tunnel_state_machine::InitialTunnelState {
                 allow_lan: settings.allow_lan,
@@ -643,6 +650,8 @@ where
             internal_event_tx.to_specialized_sender(),
             offline_state_tx,
             tunnel_state_machine_shutdown_tx,
+            #[cfg(target_os = "windows")]
+            volume_update_rx,
             #[cfg(target_os = "macos")]
             exclusion_gid,
             #[cfg(target_os = "android")]
@@ -742,6 +751,8 @@ where
             app_version_info,
             shutdown_tasks: vec![],
             tunnel_state_machine_shutdown_signal,
+            #[cfg(target_os = "windows")]
+            volume_update_tx,
         };
 
         daemon.ensure_wireguard_keys_for_current_account().await;
@@ -1241,6 +1252,8 @@ where
             SetSplitTunnelState(tx, enabled) => self.on_set_split_tunnel_state(tx, enabled).await,
             #[cfg(target_os = "windows")]
             UseWireGuardNt(tx, state) => self.on_use_wireguard_nt(tx, state).await,
+            #[cfg(target_os = "windows")]
+            CheckVolumes(tx) => self.on_check_volumes(tx).await,
             Shutdown => self.trigger_shutdown_event(),
             PrepareRestart => self.on_prepare_restart(),
             #[cfg(target_os = "android")]
@@ -1977,6 +1990,13 @@ where
                 );
                 Self::oneshot_send(tx, Err(error), "use_wireguard_nt response");
             }
+        }
+    }
+
+    #[cfg(windows)]
+    async fn on_check_volumes(&mut self, tx: ResponseTx<(), Error>) {
+        if self.volume_update_tx.unbounded_send(()).is_ok() {
+            let _ = tx.send(Ok(()));
         }
     }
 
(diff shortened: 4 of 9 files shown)


mvd-ows (commented):
> Reviewed 7 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322)* status: 6 of 9 files reviewed, 7 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-MuzhUxW6uEslwvKityv:-MuzhUxW6uEslwvKityw:bsmb5pq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L32)):*
> ```XML
> }
> 
> impl VolumeMonitorHandle {
> ```

`impl Drop for ...` as well?
___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-Muzl_Kn43pMkOlC2vtc:-Muzl_Kn43pMkOlC2vtd:b-5z38zl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L80)):*
> ```XML
> /// Monitors update requests from frontends. This checks if the known state of mounted volumes
> /// has change, and, if so, reapplies the ST config.
> async fn start_frontend_monitor(
> ```

`start` in the function name implies that some async thing is spawned before the function returns.

Maybe using just `frontend_monitor` would be more accurate/succinct?

Or maybe keep the name and move `tokio::spawn` into here?
___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 144 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-Muzy1Oy93Mj2l-iwtyl:-Muzy1Oy93Mj2l-iwtym:blon4m8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L144)):*
> ```XML
>                 // Reapply config
>                 let _ = update_tx.send(());
>                 let _ = path_monitor.refresh();
> ```

Paths and the update event sender could be kept in the path monitor only?

If they are, only the path monitor will need to be shared everywhere. There will be less sharing and fewer arguments throughout.

Why is the channel being signaled before the refresh?
___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 154 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-MuzsQrr5txCjPzHGidc:-MuzsQrr5txCjPzHGidd:bqlebrs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L154)):*
> ```XML
> 
> /// Return a bitmask representing all currently available disk drives.
> /// Each bit refers to a volume letter. The first bit refers to 'A', the
> ```

`bit 0` is probably more universally understood compared to `first bit`?

Checked official docs now, they use bit 0 etc.
___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 172 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-MuzuRtXErVXiVAk0NuT:-MuzuRtXErVXiVAk0NuU:b-gbrb6l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L172)):*
> ```XML
>                 path::Prefix::VerbatimDisk(disk) | path::Prefix::Disk(disk) => {
>                     let disk = disk.saturating_sub('A' as u8);
>                     if (volumes.wrapping_shr(u32::from(disk)) & 1) != 0 {
> ```

`disk` is the bit index?

`volumes & (1 << disk) != 0`
___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 183 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-MuzqCc1B9kGn6QgUDwQ:-MuzqCc1B9kGn6QgUDwR:bh243i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L183)):*
> ```XML
> }
> 
> fn is_broadcast_message(message: u32, w_param: usize) -> bool {
> ```

Many different messages of vastly different types are being broadcast in Windows.

This function name needs to be more specific.
___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 190 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-MuzriSl6b7SLLRZiM3y:-MuzriSl6b7SLLRZiM3z:b-5j968w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L190)):*
> ```XML
> /// Return volumes affected by the device arrival or removal message as a mask.
> /// This has the same format as `get_logical_drives()`.
> unsafe fn parse_broadcast(broadcast: &DEV_BROADCAST_HDR) -> u32 {
> ```

Function name.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322)* status: 6 of 9 files reviewed, 7 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 172 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-MuzuRtXErVXiVAk0NuT:-Mv38zmeD1HRgkHNYLXj:bspie2r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L172)):*

`disk` is the bit index?

`volumes & (1 << disk) != 0`
</blockquote></details>

Those are equivalent, but I can change it if it gets rid of the "safe" method. I'd still be a bit paranoid that the shift might cause an overflow panic if `disk` is not in the A..Z range.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322)* status: 6 of 9 files reviewed, 7 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 172 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-MuzuRtXErVXiVAk0NuT:-MvIFLli0JsxC6h4tdpU:b12q100) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L172)):*

Those are equivalent, but I can change it if it gets rid of the "safe" method. I'd still be a bit paranoid that the shift might cause an overflow panic if `disk` is not in the A..Z range.
</blockquote></details>

I think we have to assume that the system provides valid data.

Or you would also have to implement some try/except equivalent for any struct parsing where the data comes from the system.

If you're concerned that the value is not in range you could e.g. validate the value, log any issues, and resume the loop on the next iteration.

This would allow you to apply the correct check and avoid obfuscating the comparison.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322)* status: 7 of 9 files reviewed, all discussions resolved

___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 144 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-Muzy1Oy93Mj2l-iwtyl:-MvImbvG0aSsmQPrK7Kq:b-6zbv68) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L144)):*

> Why is the channel being signaled before the refresh?

The first signal causes all real paths to be sent to the driver, and exclude files on the new volume. I.e., if D: was not previously mounted, it figures out what D: maps to, and sends the path to the driver.

The refresh just tells the path monitor to monitor symlinks on the new volume (or to stop monitoring them).

The order shouldn't matter.

> Paths and the update event sender could be kept in the path monitor only?

I think I prefer to keep them separate, as they're independent. E.g., if the path monitor dies, mounts will still be detected this way.
</blockquote></details>

OK thanks for clarifying.

Is the issue then that the volume monitor knows about the path monitor?

Shouldn't whoever is receiving the signal poke the path monitor as part of its processing?
___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 183 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-MuzqCc1B9kGn6QgUDwQ:-MvImvktAOGN6cON-ip1:b-x5hh8u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L183)):*

Done. Is this better?
</blockquote></details>

Much clearer :+1: 
___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 190 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-MuzriSl6b7SLLRZiM3y:-MvInuRcCGW7RQN3pi9O:b8dotcb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L190)):*

Done. Is this better?
</blockquote></details>

Yep, better.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322)* status: 7 of 9 files reviewed, all discussions resolved

___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 144 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-Muzy1Oy93Mj2l-iwtyl:-MvJKAZ88iemO_ZlRI1T:b12sdki) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L144)):*

OK thanks for clarifying.

Is the issue then that the volume monitor knows about the path monitor?

Shouldn't whoever is receiving the signal poke the path monitor as part of its processing?
</blockquote></details>

This channel is shared with the path monitor. One could create yet another channel that pokes the path monitor in addition to sending new paths to the driver. I'm not sure if it would make things simpler.

mvd-ows (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322)* status: 7 of 9 files reviewed, all discussions resolved

___
*[talpid-core/src/split_tunnel/windows/volume_monitor.rs, line 144 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322#-Muzy1Oy93Mj2l-iwtyl:-MvJQuzBDCc7O5_liGUH:bial5hx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16db69bf7fd657fe74886f489471fb12cfd73678/talpid-core/src/split_tunnel/windows/volume_monitor.rs#L144)):*

This channel is shared with the path monitor. One could create yet another channel that pokes the path monitor in addition to sending new paths to the driver. I'm not sure if it would make things simpler.
</blockquote></details>

This is probably fine, I trust your judgement.

raksooo (commented):
> Reviewed 2 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3322)* status: 8 of 9 files reviewed, all discussions resolved (waiting on @mvd-ows)

PR #892, opened 2019-05-29
Activate LTO and size optimization

faern:
We should check what effects this has on build time and size of the result on macOS and Windows as well before we go ahead with anything like this. But some (very statistically unsecure) numbers point to strict improvements for Linux.

My numbers on current master branch doing `./build.sh --dev-build` after a `cargo clean`:
```
68M	dist/MullvadVPN-2019.4-dev-3e9d2fcf_amd64.deb
69M	dist/MullvadVPN-2019.4-dev-3e9d2fcf_x86_64.rpm
4.5M	dist-assets/mullvad
17M	dist-assets/mullvad-daemon
6.5M	dist-assets/problem-report

real	16m49.443s
user	27m20.172s
sys	0m31.170s
```

Same test as above, but with this LTO + opt-level change:
```
66M	dist/MullvadVPN-2019.4-dev-3e9d2fcf_amd64.deb
68M	dist/MullvadVPN-2019.4-dev-3e9d2fcf_x86_64.rpm
2.7M	dist-assets/mullvad
13M	dist-assets/mullvad-daemon
5.2M	dist-assets/problem-report

real	15m54.787s
user	24m9.212s
sys	0m31.842s
```

You can see the three Rust binaries became quite a lot smaller. However the resulting installers lost much less weight. I guess the `xz` compression was able to get rid of most of the extra binary weight in the first place as well? However, my build is also *faster* with the changes applied. Not sure if the opt-level set to `z` made the compile faster, or if the more incompressible binaries made the final `xz` compression faster. I did not measure Rust building and electron packaging individually.

Git checklist:

* [ ] Describe the change in **`CHANGELOG.md`** under the `[Unreleased]` header.
* [ ] Check that commits follow the [Mullvad coding guidelines](https://github.com/mullvad/coding-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/892)

PR #3326, opened 2022-02-03
Enable LTO and opt-level=3 in the Rust workspace

faern:
A long time ago we enabled LTO and `opt-level="z"` (#892). A while later we reverted that (#902 + #1459) due to segfaults somewhere around loading TLS certificates in OpenSSL. We never tracked down the real cause. We suspect it might have been something around how OpenSSL was statically compiled and/or a rustc compiler bug even. But we took the easy route of just disabling optimizations to get rid of the issue.

A long time has passed since. We don't even use/link to OpenSSL in the daemon/Rust code any longer. And the Rust compiler has seen many new releases. Even if we don't know exactly what caused the segfault we can't live in fear forever. Let's enable full optimizations and see what happens!

The result of this change is longer compile times, but smaller binaries. Hopefully it leads to more efficient code also. Not that we do much that is performance critical, but anyway.

The binary size differences that I can observe on Linux when building with `./build.sh --optimize` is:

| Binary | Size before this PR | Size after this PR |
| --- | :-: | :-: |
| mullvad | 5.3 MiB | 5.0 MiB |
| mullvad-daemon | 17 MiB | 16 MiB |
| mullvad-setup | 10 MiB | 6.5 MiB |

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,5 @@ members = [
 exclude = ["dist-assets/binaries/shadowsocks-rust"]
 
 [profile.release]
-# FIXME: This is here as a temporary hack to stop the mullvad-daemon from segfaulting
-# on Windows. Can hopefully be removed after we have replaced OpenSSL with Rustls.
-# https://github.com/mullvad/mullvadvpn-app/pull/1438
-# Other opt-levels that stops the daemon from crashing include 0 and 1. Crashes has
-# only been detected on 2.
-opt-level = "s"
+opt-level = 3
+lto = true


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3326)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3326)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3330, opened 2022-02-04
Fix missing fdroid metadata

albin-mullvad:
F-Droid requires the metadata to be present in a specific location which
isn't compatible with our mono-repo structure, therefore this change
moves the Gradle Play Publisher (GPP) directories to a location
compatible with F-Droid and adds a symlink at the location GPP expects,
which leads to the F-Droid location.

The reason for doing it this way rather than symlinking the other way
around is that we don't use GPP itself atm. Even if we start using it,
it's still easier to make sure everything is working (in regards to the
symlink) in the GPP build environment compared to the F-Droid build
environment.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3330)


pinkisemils (approved):
> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3330)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3332, opened 2022-02-04
Explicitly set feature resolver version for workspace

dlon:
The feature resolution currently used enables platform-specific features for all targets. This is [no longer true](https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2) for version 2 of the resolver. The old behavior can be problematic when dependencies or features depend on the build target.

Specifying 2021 as the Rust edition in the root package causes the resolver to be 2 by default. Unfortunately, the edition cannot be set in the `[workspace]` section. So this PR explicitly sets `resolver = "2"` in the `[workspace]` section instead (as is suggested [here](https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions)).

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3332)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 members = [
     "android/translations-converter",
     "mullvad-daemon",


faern (approved):
Oh darn. Cargo workspace limitations strikes again! Great catch. We should of course use resolver 2! :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3332#-:-Mv3uq4fAx2TqRqa0tXg:bcq3a40)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3373, opened 2022-02-18
Enable "Always require VPN" by default if settings cannot be parsed

dlon:
This prevents the app from entering the (non-blocking) disconnected state unexpectedly. For example, if "always require VPN" or auto-connect was previously enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3373)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,10 @@ Line wrap the file at 100 chars.                                              Th
 - Fix settings file being truncated before being read.
 - Improve performance for automatically scrolling text in desktop app.
 
+### Security
+- Enable "Always require VPN" by default if the settings cannot be parsed. This reduces the number
+  of errors that lead to the daemon unexpectedly starting into non-blocking mode.
+
 
 ## [2022.1-beta1] - 2022-02-14
 ### Added
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -584,11 +584,7 @@ where
                 error.display_chain_with_msg("Failed to migrate settings or cache")
             );
         }
-        let mut settings = SettingsPersister::load(&settings_dir).await;
-
-        if version::is_beta_version() {
-            let _ = settings.set_show_beta_releases(true).await;
-        }
+        let settings = SettingsPersister::load(&settings_dir).await;
 
         let target_state = if settings.get_account_token().is_none() {
             PersistentTargetState::force(&cache_dir, TargetState::Unsecured).await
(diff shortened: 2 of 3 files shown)


faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3373)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[CHANGELOG.md, line 40 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3373#-MwBQsT03X2c_29x2h86:-MwBQsT03X2c_29x2h87:b-i9f763) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dd8cb45ca69e3fcee0c2c3c5314f91415e5c11d4/CHANGELOG.md#L40)):*
> ```Markdown
>   the machine will not transition directly from the `disconnected` to the `disconnecting` state
>   if an error occurs.
> - Enable "Always require VPN" by default if the settings cannot be parsed.
> ```

I kind of want this under `### Security`. Can we also add a second sentence something like "This improves the kill switch, reducing the number of errors that lead to the daemon unexpectedly starting into non-blocking mode."
___
*[mullvad-daemon/src/settings.rs, line 62 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3373#-MwBT3Rr9qC4AjIPMz3S:-MwBT3Rs6qed83VJH3Sy:b4e88hx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dd8cb45ca69e3fcee0c2c3c5314f91415e5c11d4/mullvad-daemon/src/settings.rs#L62)):*
> ```XML
>                     error.display_chain_with_msg("Failed to load settings. Using defaults.")
>                 );
>                 (Self::default_settings(true), true)
> ```

Boolean arguments are always a bit hard to know what they do from the callsite. `default_settings(true)` does not tell me anything about what the bool is doing.

Given that we only call it with `true` in a single place, would it maybe be cleaner and easier to read to remove the `fallback` functionality of `default_settings` and instead explicitly set `block_when_disconnected` in this error path? Then it can be properly documented here with a comment. Saying something like "When failing to load settings, protect the user by blocking the internet by default".

faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3373#-:-MwR-18IAohV9rP3Yr2U:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3384, opened 2022-02-24
Add will-change CSS property for transition properties

raksooo:
This PR aims to optimize transitions by adding the `will-change` CSS property and better define `transition-property` instead of using `all` for some transitions.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3384)

### DIFF
--- a/gui/src/renderer/components/Accordion.tsx
+++ b/gui/src/renderer/components/Accordion.tsx
@@ -18,6 +18,7 @@ const Container = styled.div((props: { height: string; animationDuration: number
   display: 'flex',
   height: props.height,
   overflow: 'hidden',
+  willChange: 'height',
   transition: `height ${props.animationDuration}ms ease-in-out`,
 }));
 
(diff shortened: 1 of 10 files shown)


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3384)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @raksooo)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3384#-MwkDuTk6ZZn0joooIA1:-MwkDuTk6ZZn0joooIA2:b-fak5cn) (no related file):*
Based on documentation examples on MDN, it looks like willChange should only be set before animating the element and removed after completing animation. Documentation also states that in other cases it may worsen performance.

https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

Quote from documentation:

> Use sparingly. The normal behavior for optimizations that the browser make is to remove the optimizations as soon as it can and revert back to normal. But adding will-change directly in a stylesheet implies that the targeted elements are always a few moments away from changing and the browser will keep the optimizations for much longer time than it would have otherwise. So it is a good practice to switch will-change on and off using script code before and after the change occurs.

I haven’t dug deep enough but that’s the first thing that came up when I looked up the willChange property that I never used before. Did you measure the performance impact somehow?

---

raksooo:
Closing since we came to the conclusion that this wasn't the way forward, and since Hank has made other improvements.

PR #3407, opened 2022-03-09
Shut down REST request service when all handles have been dropped

dlon:
Since the dawn of time, the REST request service has owned a copy of its own command sender. Because of this, the actor never stops waiting for commands to process. This is made apparent by the fact that sockets remain open long after location lookups are made. Notably, it also means that the lookup RPC previously leaked some memory every time it was called.

The PR fixes this by only keeping a weak reference to the sender.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3407)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ Line wrap the file at 100 chars.                                              Th
 - Fix the sometimes incorrect time added text after adding time to the account.
 - Fix scrollbar no longer responsive and usable when covered by other elements.
 - Improve tunnel bypass for the API sometimes not working in the connecting state.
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3407)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3407)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3414, opened 2022-03-11
Store weak value using closure to trim down on Any* types

pronebird:

Git checklist:

- I found [a neat trick](https://www.swiftbysundell.com/articles/combining-value-and-reference-types-in-swift/#collections-of-weak-references), which explains how to store a weak reference to object without additional constraints on the stored type, which previously prevented us from storing a collection of instances conforming to a certain protocol and forced us to implement `Any*` types to wrap the actual underlying type.

- Switch `ObserverList` from `NSRecursiveLock` to `NSLock` which is cheaper.


raksooo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3414)* status: all files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[ios/MullvadVPN/ObserverList.swift, line 58 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3414#-My2VPCl0V57yGQCJ1iK:-My2VPCl0V57yGQCJ1iL:b-bc7ecd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b9a63ec239ad2bcad56d13df39f63f308420fe0/ios/MullvadVPN/ObserverList.swift#L58)):*
> ```Swift
>         if let index = index {
>             observers.remove(at: index)
>         }
> ```

Why isn't `removeAll` a good solution here?
___
*[ios/MullvadVPN/ObserverList.swift, line 63 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3414#-My2XISt1w58DOce4QeJ:-My2XISt1w58DOce4QeK:b-y5uixk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b9a63ec239ad2bcad56d13df39f63f308420fe0/ios/MullvadVPN/ObserverList.swift#L63)):*
> ```Swift
>     }
> 
>     func forEach(_ body: (T) -> Void) {
> ```

What's the reason behind removing observers that are `nil`? Is it just to have an up to date list or are there additional reasons?

raksooo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3414#-:-My6pSf442GfsjivrWmc:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[ios/MullvadVPN/ObserverList.swift, line 58 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3414#-My2VPCl0V57yGQCJ1iK:-My6n_gk35b1j_v6gjbu:bnddhc9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b9a63ec239ad2bcad56d13df39f63f308420fe0/ios/MullvadVPN/ObserverList.swift#L58)):*

Because it would iterate over the entire collection. Each unique object can only be added once as observer.
</blockquote></details>

Ah, of course. Thanks for clarifying!
___
*[ios/MullvadVPN/ObserverList.swift, line 63 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3414#-My2XISt1w58DOce4QeJ:-My6nk3S40UUTcyP69I3:b-oxdg4z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b9a63ec239ad2bcad56d13df39f63f308420fe0/ios/MullvadVPN/ObserverList.swift#L63)):*

Since observers are stored as collection of `WeakBox<ObserverType>` which hold the underlying value using weak reference, it means that once the object is released elsewhere, this reference turns into `nil`. In other words to unregister the observer, it is enough to destroy the observing object. 

Once box does not hold any value (turns `nil`), it becomes a dead weight which adds to extra iterations when looping over observers. Hence I opt-in for cleaning such boxes when iterating over observers.
</blockquote></details>

Okay, I was thinking that it would make sense to move the cleaning part into it's own function but that would probably result in additional iterations.

PR #3421, opened 2022-03-14
Warn if IPs cannot be obtained for default routes

dlon:
It might be that we cannot obtain any addresses for the interface used for excluded apps, even if a route exists. This PR doesn't fix the issue but adds some logging that might be helpful.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3421)


mvd-ows (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3421)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3429, opened 2022-03-17
Delete leftover settings in Keychain during login

pronebird:

Git checklist:

We had issues with users reinstalling the app because of broken tunnel, only to find themselves using the same WireGuard key. That's because Keychain entry isn't automatically removed with the app unless the user explicitly logged out before uninstalling the app from device.

This small PR makes sure that if the leftover entry is found in Keychain during login, the app will try to remove WG stored in the leftover settings, then wipe the Keychain entry before adding new blank settings with new key.


pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3429)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[ios/CHANGELOG.md, line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3429#-MyN_1dv3uW-4j1tMon3:-MyNdKQG8FPN6nuUggIz:bybrxss) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/676e7cecb39e300dce7ded3d52f4f92096f585f5/ios/CHANGELOG.md#L27)):*

It looks like this and the next line ends with two spaces.
</blockquote></details>

I'll fix it before merging. Thanks

PR #3433, opened 2022-03-22
Move imported types out of `v5` module

dlon:
This doesn't eliminate non-vendored types, but makes sure that outside modules are only imported in `migrations::device`, not in any `v{n}` module.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3433)

### DIFF
--- a/mullvad-daemon/src/device.rs
+++ b/mullvad-daemon/src/device.rs
@@ -356,7 +356,7 @@ impl AccountManager {
         Ok(())
     }
 
-    async fn logout_inner(&mut self) -> tokio::task::JoinHandle<()> {
+    fn logout_inner(&mut self) -> tokio::task::JoinHandle<()> {
         let prev_data = self.data.take();
         let service = self.device_service.clone();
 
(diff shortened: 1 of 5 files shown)


faern (commented):
> but makes sure that outside modules are only imported in migrations::device, not in any v{n} module.

The important part is not where the code resides. Rather that the conversion of the settings format does not rely on types outside the migration module, so that it can change. This is a special case where the "migration" also needs a REST client etc. That can rely on things that change. But the actual settings *format* conversion from `V5->V6` cannot rely on any type outside of the migration module.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3433#-:-MylmNxJAjqvIwfGAaDW:bqzku5t)* status: 0 of 5 files reviewed, all discussions resolved

dlon (commented):
> The important part is not where the code resides. Rather that the conversion of the settings format does not rely on types outside the migration module, so that it can change.

I agree with that, but I still think it's nicer to not mix that stuff with the pure format conversion.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3433#-:-MylnI_L7mNNwhq2iFWR:byggwgg)* status: 0 of 5 files reviewed, all discussions resolved

PR #3439, opened 2022-03-23
Fix bug with `PidManager::add_list`

faern:
This first caught my attention when seeing the code `T: Into<i32> + ToString`. `T` was never used as integers. It was only converted to string. So this felt like a strange way to try to encourage the user to send `i32`s instead of just taking an `&[i32]` directly. I first converted to that and tried it. But it did not work. I then realized that the code did not put any separators between the PIDs it wrote. So essentially trying to add PIDs `&[1, 2, 3]` resulted in `"123"` being written to the file.

I guess this bug was never found because we never called this method with more than one PID.

Some more possible improvements I see here that is out of scope for this PR is:
* Convert the OpenVPN Shadowsocks proxying into using our bundled Shadowsocks library, not `sslocal` as a subprocess. I have created a backlog item for it.
* Move CGroup management to a separate crate. I really don't think we should define constants such as `SPLIT_TUNNEL_CGROUP_NAME` and functions such as `find_net_cls_mount` in `talpid-types`. This crate has become too much of a kitchen sink.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3439)

### DIFF
--- a/talpid-core/src/proxy/shadowsocks.rs
+++ b/talpid-core/src/proxy/shadowsocks.rs
@@ -186,17 +186,14 @@ impl ShadowsocksProxyMonitor {
                     error.display_chain_with_msg("Failed to initialize PidManager"),
                 )
             })?;
-            let i32_pids = subproc
-                .pids()
-                .iter()
-                .map(|pid| *pid as i32)
-                .collect::<Vec<_>>();
-            excluded_pids.add_list(&i32_pids).map_err(|error| {
-                Error::new(
-                    ErrorKind::Other,
-                    error.display_chain_with_msg("Failed to exclude Shadowsocks process"),
-                )
-            })?;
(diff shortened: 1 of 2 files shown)


dlon (approved):
> Reviewed 2 of 2 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3439)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3439)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3465, opened 2022-03-28
Don't abort the uninstallation if removing the device fails on macOS

dlon:
This might happen because the API is not reachable, or simply because the device no longer exists. Regardless of the reason, it should not prevent the app from being uninstalled.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3465)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3465#-:-MzGE6Yl28V25k4j-xjC:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3473, opened 2022-03-31
Use a mutex to share device data

pinkisemils:
Since obtaining device data from the daemon can block due to the device manager waiting on an RPC to finish, this can block the daemon. To get around this deadlock, the easiest and seemingly the cleanest solution has been taken - instead of sharing the device data via messages, it can be shared via a mutex. This way, the device data is always accessible without waiting on any one actor in the system.


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3473)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @pinkisemils)

PR #3483, opened 2022-04-04
Improve CLI version output slightly

dlon:
The indentation in the output of `mullvad version` isn't helpful:

```
Current version: 2022.1
	Is supported: true
	No newer version is available
	Latest stable version: 2022.1
	Latest beta version: 2022.1-beta2
```

"No newer version is available" is also not exactly correct.

I've updated it to look more like `mullvad account get`:

```
Current version      : 2022.1
Is supported         : true
Suggested upgrade    : none
Latest stable version: 2022.1
Latest beta version  : 2022.1-beta2
```

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3483)

### DIFF
--- a/mullvad-cli/src/cmds/version.rs
+++ b/mullvad-cli/src/cmds/version.rs
@@ -20,27 +20,33 @@ impl Command for Version {
             .await
             .map_err(|error| Error::RpcFailedExt("Failed to obtain current version", error))?
             .into_inner();
-        println!("Current version: {}", current_version);
+        println!("{:21}: {}", "Current version", current_version);
         let version_info = rpc
             .get_version_info(())
             .await
             .map_err(|error| Error::RpcFailedExt("Failed to obtain version info", error))?
             .into_inner();
-        println!("\tIs supported: {}", version_info.supported);
+        println!("{:21}: {}", "Is supported", version_info.supported);
 
         if !version_info.suggested_upgrade.is_empty() {
-            println!("\tSuggested update: {}", version_info.suggested_upgrade);
+            println!(
+                "{:21}: {}",
+                "Suggested upgrade", version_info.suggested_upgrade
+            );
         } else {
-            println!("\tNo newer version is available");
+            println!("{:21}: none", "Suggested upgrade");
         }
 
         if !version_info.latest_stable.is_empty() {
-            println!("\tLatest stable version: {}", version_info.latest_stable);
+            println!(
+                "{:21}: {}",
+                "Latest stable version", version_info.latest_stable
+            );
         }
 
         let settings = rpc.get_settings(()).await?.into_inner();
         if settings.show_beta_releases {
-            println!("\tLatest beta version: {}", version_info.latest_beta);
+            println!("{:21}: {}", "Latest beta version", version_info.latest_beta);
         };
 
         Ok(())


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3483)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[mullvad-cli/src/cmds/version.rs, line 29 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3483#-Mzp7avuCaToON29CJNm:-Mzp7avuCaToON29CJNn:bppzoj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/edcacf160727ebd33269fae708a66b0c41d2fd0e/mullvad-cli/src/cmds/version.rs#L29)):*
> ```XML
>             .map_err(|error| Error::RpcFailedExt("Failed to obtain version info", error))?
>             .into_inner();
>         println!("Is supported         : {}", version_info.supported);
> ```

It is very hard to visually see that the colons are aligned. And when adding/changing fields it's going to be pretty hard to fix the alignment again. I won't mark this as a blocker, this is OK, but I think it would be an improvement if one of the following could be done:
1. Rearrange the code so all `println!` are next to each other and with the same indentation, so that the strings they print are aligned also.
2. Use some formatting helper stuff that does the alignment for us rather than manually inserting it. (https://doc.rust-lang.org/stable/std/fmt/index.html#fillalignment)
3. skip having it aligned. Reading the following is not that much harder:
    ```
    Current version: 2022.1
    Is supported: true
    Suggested upgrade: none
    Latest stable version: 2022.1
    Latest beta version: 2022.1-beta2
    ```

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3483)* status: 0 of 1 files reviewed, all discussions resolved

___
*[mullvad-cli/src/cmds/version.rs, line 29 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3483#-Mzp7avuCaToON29CJNm:-MzseTdD0LU3tOMcX2Yf:b-q6jrw8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/edcacf160727ebd33269fae708a66b0c41d2fd0e/mullvad-cli/src/cmds/version.rs#L29)):*

It is very hard to visually see that the colons are aligned. And when adding/changing fields it's going to be pretty hard to fix the alignment again. I won't mark this as a blocker, this is OK, but I think it would be an improvement if one of the following could be done:
1. Rearrange the code so all `println!` are next to each other and with the same indentation, so that the strings they print are aligned also.
2. Use some formatting helper stuff that does the alignment for us rather than manually inserting it. (https://doc.rust-lang.org/stable/std/fmt/index.html#fillalignment)
3. skip having it aligned. Reading the following is not that much harder:
    ```
    Current version: 2022.1
    Is supported: true
    Suggested upgrade: none
    Latest stable version: 2022.1
    Latest beta version: 2022.1-beta2
    ```
</blockquote></details>

Done (2).

faern (approved):
:lgtm: Awesome!

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3483#-:-MzsfxlU6x8nJXBQS4rw:bswp6el)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3485, opened 2022-04-05
Add information about post upgrade tasks in UI

raksooo:
This PR retrieves the post upgrade status from the daemon and tells the user if post upgrade tasks are running.

Git checklist:

### DIFF
--- a/gui/src/main/daemon-rpc.ts
+++ b/gui/src/main/daemon-rpc.ts
@@ -499,6 +499,11 @@ export class DaemonRpc {
     await this.callEmpty(this.client.checkVolumes);
   }
 
+  public async isPerformingPostUpgrade(): Promise<boolean> {
+    const response = await this.callEmpty<BoolValue>(this.client.isPerformingPostUpgrade);
+    return response.getValue();
+  }
+
   public async getDevice(): Promise<IDeviceConfig | undefined> {
     try {
       const response = await this.callEmpty<grpcTypes.DeviceConfig>(this.client.getDevice);
--- a/gui/src/main/index.ts
+++ b/gui/src/main/index.ts
@@ -127,6 +127,7 @@ class ApplicationMain {
   private reconnectBackoff = new ReconnectionBackoff();
   private beforeFirstDaemonConnection = true;
   private connectedToDaemon = false;
+  private isPerformingPostUpgrade?: boolean;
   private quitStage = AppQuitStage.unready;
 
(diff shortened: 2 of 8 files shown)


pronebird (changes requested):
> Reviewed 3 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3485)* status: 4 of 9 files reviewed, 1 unresolved discussion (waiting on @pronebird and @raksooo)

___
*[gui/src/main/index.ts, line 630 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3485#-Mzy7-jIEWsUdnzca4Gr:-Mzy7-jIEWsUdnzca4Gs:baeby23) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a64c38052e9cf316d2819adcee89802f747dd630/gui/src/main/index.ts#L630)):*
> ```XML
>     }
> 
>     if (firstDaemonConnection) {
> ```

Shouldn't we check if daemon is performing a post upgrade check each time we reconnect to it instead of just on first connection attempt?

pronebird (commented):
> Reviewed 5 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3485)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3485)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[gui/src/main/index.ts, line 630 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3485#-Mzy7-jIEWsUdnzca4Gr:-Mzyg24u2jvSLuEC14Wt:bmz00nr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a64c38052e9cf316d2819adcee89802f747dd630/gui/src/main/index.ts#L630)):*

Shouldn't we check if daemon is performing a post upgrade check each time we reconnect to it instead of just on first connection attempt? 
</blockquote></details>

I'm not sure. This can only happen after upgrading the app, which should kill both the daemon and the frontend. In what situations can the daemon be in this state when connected with the frontend for a second time? The ones I can think of is if the frontend isn't killed during upgrade, and if the connection fails before this call on the first connection but then would work the second time. Are there any more?

PR #3487, opened 2022-04-06
Use clocks that take into account system sleep for long-running timers

dlon:
Fun fact: `Instant::now()` includes time spent in suspend mode on Windows, but not on Linux or macOS. I've added a `sleep` function that is guaranteed to use clocks that account for suspension.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3487)


pinkisemils (commented):
We might want to push some changes to the standard library for Rust or tokio, or have a crate of our own. There is a timer that does increment during suspension on Linux (`CLOCK_BOOTTIME`), and should probably be used here.  I think the optimal solution is to use both a monotonic clock and the system clock.

As for the changes, I was envisioning an abstraction, a future based on a minute interval that would check both clocks and fire whenever either has expired.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3487#-:-N--vckfCQbUu6CmHU5L:b-idnnss)* status: 0 of 4 files reviewed, all discussions resolved

dlon (commented):
There's an issue for `Instant` already: https://github.com/rust-lang/rust/issues/79462.

> I was envisioning an abstraction, a future based on a minute interval that would check both clocks and fire whenever either has expired.

~~Isn't this sort of what `future_retry::sleep` is? It could be more general.~~

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3487#-:-N-1t8xK6sx2_fd3s683:b-2rh0gg)* status: 0 of 4 files reviewed, all discussions resolved

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3487)* status: 0 of 13 files reviewed, 1 unresolved discussion

___
*[talpid-time/src/unix.rs, line 32 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3487#-N-8Kz-62_qAl7cOEiqd:-N-8Kz-62_qAl7cOEiqe:bklpzwk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f4a9ccc70fb6d941f3cd5cf97f4c54f519d2260/talpid-time/src/unix.rs#L32)):*
> ```XML
>         let (tv_sec, tv_nsec) = if self.t.tv_nsec < earlier.t.tv_nsec {
>             (
>                 self.t.tv_sec - earlier.t.tv_sec - 1,
> ```

I think overflows are possible here.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3487)* status: 0 of 13 files reviewed, all discussions resolved

___
*[talpid-time/src/unix.rs, line 32 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3487#-N-8Kz-62_qAl7cOEiqd:-N-9-UCU1q9294FTm48-:bro2c66) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f4a9ccc70fb6d941f3cd5cf97f4c54f519d2260/talpid-time/src/unix.rs#L32)):*

I think underflows are possible here.
</blockquote></details>

This is fine. `i64::MIN == -i64::MAX - 1`

pinkisemils (approved):
> Reviewed 2 of 4 files at r1, 7 of 12 files at r2, 1 of 2 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3487)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #2897, opened 2021-08-08
Adding support for Siri Shortcuts in iOS


rpecka:
## Issue report

Operating system: iOS 14.6

App version: 2021.2

### Issue description

I have a use case where I'd like to be able to enable and disable the VPN on iOS using the native automation solution, Shortcuts. I did some research and testing to see if this could be done in Mullvad and I ran into enough issues that I think it's worth some discussion.

### Problems with `IntentsExtension`s

It looks like the recommended way to support shortcuts is to add an `IntentsExtension` to your app. This allows for the minimal functionality of your app to be packaged into its own executable and run by the system to execute your app's actions. Keeping the code in a minimal executable helps keep the footprint of the action small and helps the app stay within the 10 second time limit.

I tried implementing an `IntentsExtension` for Mullvad and ran into [this issue](https://developer.apple.com/forums/thread/96020?answerId=291486022#291486022). TLDR: VPN configurations in iOS seem like they can only be modified by the executable with the bundle ID that created them. Since the `IntentsExtension` must have a different bundle ID from the host app, this means that an extension is probably out of the question until this is resolved.

### In-app Intent Handling

In iOS 14, support for in-app intent handling was added. This means that intents which are listed in the host app's Info.plist will be forwarded directly to the app rather than to an `IntentsExtension` (Not sure what the behavior is if the intent is listed in the extension and the host).

I tried this out by implementing the required AppDelegate plumbing but found that none of the delegate functions were being called. This led me to [this WWDC session](https://developer.apple.com/videos/play/wwdc2020/10073/?time=329) where the speaker outlines that to support in-app intent handling:

1. Add support for multiple windows using the `UIScene` lifecycle
2. Add the list of supported intents to Info.plist
3. Add the AppDelegate plumbing for intent handling

2 and 3 are easy, 1 is the main reason I'm making this post.

### Supporting the `UIScene` Lifecycle

Supporting the `UIScene` Lifecycle in the canonical way would require us to remove all UI code from `application(_:didFinishLaunchingWithOptions:)`, refactor the app/account initialization process as it relates to the UI, and make heavy modifications to how we are storing ViewControllers i.e. not storing them in AppDelegate and allowing for more than one of each controller to exist simultaneously.

Alternatively, I believe that there's a way that we can force the app to only allow one window even with the `UIScene` lifecycle, and use that ability to hack the current architecture into the `UIScene` lifecycle.

[...]

PR #3489, opened 2022-04-07
Add toggle tunnel Siri intent to iOS app

nchlscs:
## Motivation
iOS does not support split tunneling. But there is a workaround using Siri Shortcuts app. Automation allows you to perform specific actions when opening and closing selected applications. This Pull Request adds Siri Shortcuts integration to the Mullvad iOS app.

## Example
Let's say we want to turn off the tunnel while using maps and calendar.
1. Open Shortcuts
2. Tap on "Automation"
3. Tap on "Create Personal Automation"
4. Choose "App"
5. Choose "Maps" and "Calendar"
6. Select "Is Opened" option
7. Tap on "Next"
8. Tap on "Search for apps and actions"
9. Type and select "Toggle VPN Connection" from Mullvad
10. Set "Off" state
11. Tap on "Next"
12. Turn off "Ask Before Running" and "Notify When Run"
13. Tap on "Done"

Follow the same steps by selecting "Is Closed" in step 6 and "On" in step 10. Now, when opening maps or calendar, the tunnel will turn off and turn back on when changing the application.

## Availability
The minimum version is iOS 13. iOS 15.4 allows you to disable Automation notifications, which makes it less annoying to use.

## Git checklist

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3489)


pronebird:
Hi,

I think the reason why this workaround works is because existing connections are not terminated when VPN comes back up. We do plan to implement [includeAllNetworks](https://developer.apple.com/documentation/networkextension/nevpnprotocol/3131931-includeallnetworks) which should tackle this, however we had a few obstacles in enabling that specific setting on iOS. I don't think this PR offers a reliable solution.

I wish I could accept that PR, but sadly not this time. 

Note that we also had another [request](https://github.com/mullvad/mullvadvpn-app/issues/2897) to add intents to start or stop VPN with Siri, but that would only work if we were to implement the intents handling inside the main bundle, since we'd have to turn off the on-demand setting in VPN configuration to prevent the tunnel from coming back up. Intents extension simply does not have sufficient permissions to persist VPN configuration changes. So if we were to implement intents, then we'd probably handle them inside the main bundle.

nchlscs:
@pronebird 
I know it’s closed, just want to share some observations about Intents extensions. Most likely, extensions have the same permissions as the main app to modify VPN configurations. There is, though, an undocumented behaviour ([or a bag](https://developer.apple.com/forums/thread/96020)), that does not allow to modify configurations created by another bundle (doesn’t matter, was it the main app or an extension). Meanwhile, if you make an exact copy of that configuration, it’ll be saved successfully, and allows us to toggle on-demand setting from the extension. I’ve updated this [pull request on my fork](https://github.com/nchlscs/mullvadvpn-app/pull/1/files#diff-99a899c76a44c7aebc6322f5da8fc5ffcedcdf6bf9b07bf86e69461fd19e9786) to demonstrate how it works, if you are interested.

By the way, do you have any plans to implement intents handling, or don't you see this feature in your roadmap?

pronebird:
@nchlscs Sorry for a late reply. I must have missed your message. Interesting approach, however I think it's probably more reliable to use intents eligible for in-app handling (#3807). There intents to control the VPN tunnel should become available with the next release. (Also see: https://github.com/mullvad/mullvadvpn-app/issues/2897)

nchlscs:
@pronebird 
Thanks for a reply. I missed the fact there is the background handling starting iOS 14, thought it was a privilege of an extension only. Agree that keeping this in the main app is more reliable and maintainable. I tried the latest master, works like a charm. The only thing I’m missing now is to be able to add shortcuts via `INUIEditVoiceShortcutViewController` from the app so no need to configure them manually on each device. Arguable, but I think this is a more obvious way to add shortcuts for the most users. I was going to suggest this small feature with the new PR, but if you say this is redundant functionality, I’ll do nothing.

PR #3493, opened 2022-04-08
Improve random port distribution

pronebird:

Git checklist:

- Improve random port distribution. Should be less biased towards port 53. See Rust code for reference: 
  https://github.com/mullvad/mullvadvpn-app/blob/66ff8978f94d4ad1c8c45872de9f540cbcbcf612/mullvad-daemon/src/relays/matcher.rs#L202-L238
- Replace `CloseRange<UInt16>` in relay model with `[[UInt16]]`. Parse and validate  port ranges manually in `RelaySelector`, skip invalid ranges but don't error.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3493)

### DIFF
--- a/ios/CHANGELOG.md
+++ b/ios/CHANGELOG.md
@@ -37,6 +37,9 @@ Line wrap the file at 100 chars.                                              Th
   credits on accounts that no longer exist on our backend. Usually the case with newly created 
   accounts that went stale.
 
+### Fixed
+- Improve random port distribution. Should be less biased towards port 53.
+
 
 ## [2022.1] - 2022-02-15
 ### Added
--- a/ios/MullvadVPN/REST/ServerRelaysResponse.swift
+++ b/ios/MullvadVPN/REST/ServerRelaysResponse.swift
@@ -34,7 +34,7 @@ extension REST {
     struct ServerWireguardTunnels: Codable {
         let ipv4Gateway: IPv4Address
         let ipv6Gateway: IPv6Address
-        let portRanges: [ClosedRange<UInt16>]
+        let portRanges: [[UInt16]]
         let relays: [ServerRelay]
     }
 
--- a/ios/MullvadVPN/RelaySelector.swift
+++ b/ios/MullvadVPN/RelaySelector.swift
@@ -8,6 +8,7 @@
 
 import Foundation
 import Network
+import Logging
 
 struct RelaySelectorResult: Codable {
     var endpoint: MullvadEndpoint
@@ -38,16 +39,16 @@ extension RelaySelector {
     static func evaluate(relays: REST.ServerRelaysResponse, constraints: RelayConstraints) -> RelaySelectorResult? {
         let filteredRelays = applyConstraints(constraints, relays: Self.parseRelaysResponse(relays))
 
-        guard let relayWithLocation = pickRandomRelay(relays: filteredRelays) else {
-            return nil
-        }
-
-        guard let port = relays.wireguard.portRanges.randomElement()?.randomElement() else {
-            return nil
-        }
+        guard let relayWithLocation = pickRandomRelay(relays: filteredRelays),
+              let port = pickRandomPort(rawPortRanges: relays.wireguard.portRanges) else {
+                  return nil
+              }
 
         let endpoint = MullvadEndpoint(
-            ipv4Relay: IPv4Endpoint(ip: relayWithLocation.relay.ipv4AddrIn, port: port),
+            ipv4Relay: IPv4Endpoint(
+                ip: relayWithLocation.relay.ipv4AddrIn,
+                port: port
+            ),
             ipv6Relay: nil,
             ipv4Gateway: relays.wireguard.ipv4Gateway,
             ipv6Gateway: relays.wireguard.ipv6Gateway,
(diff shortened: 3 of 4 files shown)


raksooo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3493)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[ios/MullvadVPN/RelaySelector.swift, line 119 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3493#-N-8XJwJ6Aj8SyT0iiR6:-N-8XJwJ6Aj8SyT0iiR7:bmu29mr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a244060c5c9d0454379877877f0ed020bacd80a/ios/MullvadVPN/RelaySelector.swift#L119)):*
> ```Swift
>     }
> 
>     private static func pickRandomPort(rawPortRanges: [[UInt16]]) -> UInt16? {
> ```

Wouldn't it be easier to do something like this or am I missing somehting?
```
return rawPortRanges.joined().randomElement()
```
or if that doesn't work:
```
return parseRawPortRanges(rawPortRanges).joined().randomElement()
```

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3493)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[ios/MullvadVPN/RelaySelector.swift, line 119 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3493#-N-8XJwJ6Aj8SyT0iiR6:-N-8hRzg8ClrGK34DpCZ:b-oqkaq5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a244060c5c9d0454379877877f0ed020bacd80a/ios/MullvadVPN/RelaySelector.swift#L119)):*

Wouldn't it be easier to do something like this or am I missing somehting?
```
return rawPortRanges.joined().randomElement()
```
or if that doesn't work:
```
return parseRawPortRanges(rawPortRanges).joined().randomElement()
```
</blockquote></details>

Port ranges are expressed as array of arrays of `UInt16` as returned from API, i.e:

```
[
  [53, 53], // single port 53
  [1024, 2048], // port range of 1024 - 2048
  // etc..
]
```

So these are not ports themselves but sequence of arrays with two elements, where the first element tells the start port and the second element is the end port of the range.

Since we have over 40k of ports overall, expressing them as array of individual port numbers would take around 100KB of memory. It's more efficient to use a range and quickly loop through it to find the port we need.

raksooo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3493#-:-N-MpIlkBJu5cVoYAyT5:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[ios/MullvadVPN/RelaySelector.swift, line 119 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3493#-N-8XJwJ6Aj8SyT0iiR6:-N-MnnpuEJMXpN-O6eBq:b-n21xts) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a244060c5c9d0454379877877f0ed020bacd80a/ios/MullvadVPN/RelaySelector.swift#L119)):*

Port ranges are expressed as array of arrays of `UInt16` as returned from API, i.e:

```
[
  [53, 53], // single port 53
  [1024, 2048], // port range of 1024 - 2048
  // etc..
]
```

So these are not ports themselves but sequence of arrays with two elements, where the first element tells the start port and the second element is the end port of the range.

Since we have over 40k of ports overall, expressing them as array of individual port numbers would take around 100KB of memory. It's more efficient to use a range and quickly loop through it to find the port we need.
</blockquote></details>

Makes sense! Thanks for clarifying!
___
*[ios/MullvadVPN/RelaySelector.swift, line 125 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3493#-N-MntFpBQdFNM-KJLld:-N-MntFq7PEmTMSjl3Zj:bbotuod) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a244060c5c9d0454379877877f0ed020bacd80a/ios/MullvadVPN/RelaySelector.swift#L125)):*
> ```Swift
>         }
> 
>         guard var portIndex = (0..<portAmount).randomElement() else {
> ```

How does Swift handle the conversion from range to collection and `randomElement` here? Is it lazy evaluating, or will it initialize the whole collection?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3493)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[ios/MullvadVPN/RelaySelector.swift, line 125 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3493#-N-MntFpBQdFNM-KJLld:-N-MrKI01UWPa2HQd76p:bu8jr27) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a244060c5c9d0454379877877f0ed020bacd80a/ios/MullvadVPN/RelaySelector.swift#L125)):*

How does Swift handle the conversion from range to collection and `randomElement` here? Is it lazy evaluating, or will it initialize the whole collection?
</blockquote></details>

I don't think the call to `randomElement()` converts the `Range<Int>` to Collection. I think it simply produces a random number between the lower and upper bound of the range, i.e:

```

let range: Range<UInt16> = (5..<10)
var randomGenerator = SystemRandomNumberGenerator()
let randomNumber = range.lowerBound + randomGenerator.next(
  upperBound: range.upperBound - range.lowerBound
)
```

Also note that `0..<0` is an empty range, so `randomElement()` will always return `nil` in that case, hence I never check if `portAmount > 0`.

PR #3503, opened 2022-04-13
Fix overzealous relay selector test

pinkisemils:
The relay selector test was testing if a WireGuard endpoint wasn't using
the same ports as TCP2UDP relays will be listening on. This doesn't make
much sense as a relay could very well be using those same ports, and in
some cases it was, thus making this test fail. The faulty assertion has
been removed.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3503)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3503)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3510, opened 2022-04-19
Fix macos dns thrash

### LINKED ISSUE #3400: High CPU use by configd process with macOS client connected
## Issue report

Operating system: MacOS Monterey 12.2

App version: 2022.1

### Issue description

When the Mullvad client is connected, I see the configd process running at 100-110% CPU in Activity Monitor.
When I disconnect, configd returns to 1-2%
Leaving the client connected but changing settings for blocking ads, trackers, and malware has no effect.

pinkisemils:
These changes are an attempt to resolve  #3400. Whilst I haven't reproduced the issue itself, I do believe that only comparing the resovler IPs when comparing current and desired DNS configs should improve the thrashing, and in case it doesn't, debouncing logic has been added so that when there's too many DNS changes, the DNS monitor will just capitulate and force the daemon into the error state.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,9 @@ Line wrap the file at 100 chars.                                              Th
 - Fix navigation resetting to main view when toggling the unpinned window setting.
 - Log splitting event reason correctly.
 
+#### macOS
+- Fix thrashing due to DNS config monitoring.
+
 ### Security
 #### Android
 - Prevent location request responses from being received outside the tunnel when in the connected
--- a/talpid-core/src/dns/macos.rs
+++ b/talpid-core/src/dns/macos.rs
@@ -1,6 +1,6 @@
 use parking_lot::Mutex;
 use std::{
-    collections::HashMap,
+    collections::{BTreeSet, HashMap},
     fmt,
     net::{AddrParseError, IpAddr},
     sync::{mpsc as sync_mpsc, Arc},
@@ -131,6 +131,10 @@ impl DnsSettings {
             .unwrap_or(Vec::new())
     }
 
+    pub fn address_set(&self) -> BTreeSet<String> {
+        BTreeSet::from_iter(self.server_addresses().into_iter())
+    }
+
     pub fn interface_config(&self, interface_path: &str) -> Result<Vec<IpAddr>> {
         let addresses = self
             .server_addresses()
@@ -241,7 +245,7 @@ impl super::DnsMonitorT for DnsMonitor {
                 }
             }
             Some(state) => {
-                if servers != state.dns_settings.server_addresses() {
+                if BTreeSet::from_iter(servers.into_iter()) != state.dns_settings.address_set() {
                     for service_path in state.backup.keys() {
                         settings.save(&self.store, service_path.as_str())?;
                     }
@@ -387,7 +391,7 @@ fn dns_change_callback_internal(
                 true
             }
             Some(new_settings) => {
-                if new_settings.dict != state.dns_settings.dict {
+                if new_settings.address_set() != state.dns_settings.address_set() {
                     log::debug!("Detected DNS change for {}", *path);
                     state.backup.insert(path.to_string(), Some(new_settings));
                     true
--- a/talpid-core/src/dns/macos.rs
+++ b/talpid-core/src/dns/macos.rs
@@ -1,10 +1,12 @@
+use futures::channel::mpsc;
 use parking_lot::Mutex;
 use std::{
     collections::{BTreeSet, HashMap},
     fmt,
     net::{AddrParseError, IpAddr},
-    sync::{mpsc as sync_mpsc, Arc},
+    sync::{mpsc as sync_mpsc, Arc, Weak},
     thread,
+    time::Duration,
 };
 use system_configuration::{
     core_foundation::{
@@ -18,6 +20,10 @@ use system_configuration::{
     dynamic_store::{SCDynamicStore, SCDynamicStoreBuilder, SCDynamicStoreCallBackContext},
     sys::schema_definitions::{kSCPropNetDNSServerAddresses, kSCPropNetInterfaceDeviceName},
 };
+use talpid_time::Instant;
+use talpid_types::tunnel::ErrorStateCause;
+
+use crate::tunnel_state_machine::TunnelCommand;
 
 pub type Result<T> = std::result::Result<T, Error>;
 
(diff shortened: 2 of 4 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510)* status: all files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[talpid-core/src/dns/macos.rs, line 146 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510#-N01BydO1v5_PM-2mhDd:-N01BydO1v5_PM-2mhDe:b-8hwcfg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e535d0ca41e7d7f5df5cdf23921718cc4bbfd259/talpid-core/src/dns/macos.rs#L146)):*
> ```XML
>                         }
> 
>                         if let Err(err) = self.reset(&store) {
> ```

Will the tunnel state transition not trigger this as well?
___
*[talpid-core/src/dns/macos.rs, line 174 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510#-N01Cgkx7aATZDdjB1D9:-N01Cgky8rprlM9uNtuE:b8q3bcn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e535d0ca41e7d7f5df5cdf23921718cc4bbfd259/talpid-core/src/dns/macos.rs#L174)):*
> ```XML
>     fn reset(&mut self, store: &SCDynamicStore) -> Result<()> {
>         log::trace!("Restoring DNS settings to: {:#?}", self.backup);
>         let old_backup = std::mem::replace(&mut self.backup, HashMap::new());
> ```

Nit: You could use `std::mem::take()` here.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[talpid-core/src/dns/macos.rs, line 146 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510#-N01BydO1v5_PM-2mhDd:-N01NMKzCzrEwdhD2Mvp:b-vt6mdl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e535d0ca41e7d7f5df5cdf23921718cc4bbfd259/talpid-core/src/dns/macos.rs#L146)):*

Will the tunnel state transition not trigger this as well?
</blockquote></details>

It will, the state is reset here so that the thrashing is stopped sooner rather than later.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[talpid-core/src/dns/macos.rs, line 146 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510#-N01BydO1v5_PM-2mhDd:-N053Yzu42I72zbDEyYR:boyor8j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e535d0ca41e7d7f5df5cdf23921718cc4bbfd259/talpid-core/src/dns/macos.rs#L146)):*

It will, the state is reset here so that the thrashing is stopped sooner rather than later.
</blockquote></details>

Alright. I do wonder if moving to the error state is necessary. Could you see a temporary burst happening that isn't due to fighting over the DNS config? Just limiting the rate of changes would prevent the transition from being triggered in that case. But it also wouldn't stop the fighting, so I guess that this is better.
___
*[talpid-core/src/dns/macos.rs, line 497 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510#-N057T9gBMeG7_Ey43ec:-N057T9h32jNGxZqqQvt:bt58l93) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e535d0ca41e7d7f5df5cdf23921718cc4bbfd259/talpid-core/src/dns/macos.rs#L497)):*
> ```XML
> }
> 
> const MAX_CHANGES_PER_5_SECONDS: usize = 25;
> ```

Feel free to ignore this, but it is somewhat strange to include the interval (5 seconds) in the constant name itself, IMO. Why not `MAX_CHANGES_PER_INTERVAL`, etc.?

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[talpid-core/src/dns/macos.rs, line 146 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510#-N01BydO1v5_PM-2mhDd:-N05NDVY3JDh5DI_kdeQ:b-mjj9n5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e535d0ca41e7d7f5df5cdf23921718cc4bbfd259/talpid-core/src/dns/macos.rs#L146)):*

Alright. I do wonder if moving to the error state is necessary. Could you see a temporary burst happening that isn't due to fighting over the DNS config? Just limiting the rate of changes would prevent the transition from being triggered in that case. But it also wouldn't stop the fighting, so I guess that this is better.
</blockquote></details>

Deciding on what's the biggest burst that should still be allowed will be a balancing act and it's incredibly hard to determine what the optimal value will be just because we cannot reproduce this issue.
But I do think that entering the error state should be the best thing to do here, adding a timeout would just throttle the thrash,

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[talpid-core/src/dns/macos.rs, line 146 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510#-N01BydO1v5_PM-2mhDd:-N05OOCYDoq3hhTQto6n:buviv3e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e535d0ca41e7d7f5df5cdf23921718cc4bbfd259/talpid-core/src/dns/macos.rs#L146)):*

Deciding on what's the biggest burst that should still be allowed will be a balancing act and it's incredibly hard to determine what the optimal value will be just because we cannot reproduce this issue.
But I do think that entering the error state should be the best thing to do here, adding a timeout would just throttle the thrash, 
</blockquote></details>

And throttling the thrashing wouldn't fix it. If there's another agent trying to reconfigure DNS, chances are, DNS will be broken all the time.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3510)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3520, opened 2022-04-22
Pick API bridges directly using the relay selector

dlon:
Main changes:
* The relay selector handle can be cloned, so the API connection mode no longer needs to be returned by an internal daemon command. This makes deadlocks less likely to be introduced.
* The relay selector has been moved to its own crate.
* The transport protocol will no longer be set to UDP when the bridge mode is set to `auto` and a bridge is returned. Previously, this occurred some of the time.
* Update the bridge when the constraints change.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520)* status: 0 of 11 files reviewed, 4 unresolved discussions

___
*[mullvad-daemon/src/api.rs, line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520#-N0GCxNL3Q8oTyaut_ci:-N0GCxNL3Q8oTyaut_cj:b-httjn4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9598f5761f71d0ca4b6c6d6066d3015f7e9d40ea/mullvad-daemon/src/api.rs#L34)):*
> ```XML
> /// * When no matching bridge is found, e.g. if the selected hosting providers don't
> ///   match any bridge, [`ApiConnectionMode::Direct`] is returned.
> pub struct ApiConnectionModeProvider {
> ```

Instead of providing a stream to `mullvad-api` at all, it should perhaps receive a relay selector handle? I've stuck with the stream in part to limit the scope of the PR, but also because:
1. The API client needs to be created before the relay selector, as migration code depends on it. But the relay selector must be created after that code runs.
2. If the API client selects bridges on its own, then they must be correctly forwarded to the TSM in the `mullvad-problem-report`, etc. The correct solution there might be to add a daemon command that sets the allowed endpoint.
___
*[mullvad-daemon/src/lib.rs, line 744 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520#-N0GAtQ33w-C3WK7E0Tc:-N0GAtQ33w-C3WK7E0Td:bl0xzip) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9598f5761f71d0ca4b6c6d6066d3015f7e9d40ea/mullvad-daemon/src/lib.rs#L744)):*
> ```XML
>         proxy_provider_handle.set_relay_selector(relay_selector.clone());
> 
>         let mut relay_list_updater = RelayListUpdater::new(
> ```

There's probably no benefit to separating this from the selector.
___
*[mullvad-relay-selector/src/lib.rs, line 220 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520#-N0GCAof1lxG_7udXxGr:-N0GCAof1lxG_7udXxGs:b-3cz6n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9598f5761f71d0ca4b6c6d6066d3015f7e9d40ea/mullvad-relay-selector/src/lib.rs#L220)):*
> ```XML
> 
> #[derive(Clone)]
> pub struct RelaySelector {
> ```

This should probably be renamed to `RelaySelectorHandle`.
___
*[mullvad-relay-selector/src/lib.rs, line 262 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520#-N0GBx3x8DhW4M8Fuo8S:-N0GBx3x8DhW4M8Fuo8T:bj8omk8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9598f5761f71d0ca4b6c6d6066d3015f7e9d40ea/mullvad-relay-selector/src/lib.rs#L262)):*
> ```XML
> 
>     /// Returns a random relay and relay endpoint matching the current constraints.
>     pub fn get_relay(
> ```

This should probably return the obfuscator as well.

pinkisemils (commented):
> Reviewed 5 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520)* status: 5 of 11 files reviewed, 4 unresolved discussions (waiting on @dlon)

___
*[`mullvad-daemon/src/api.rs` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520#-N0GCxNL3Q8oTyaut_ci:-N0VDjUIAZlo4nr-87jv:b2u1rnb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9598f5761f71d0ca4b6c6d6066d3015f7e9d40ea/mullvad-daemon/src/api.rs#L34)):*

Instead of providing a stream to `mullvad-api` at all, it should perhaps receive a relay selector handle? I've stuck with the stream in part to limit the scope of the PR, but also because:
1. The API client needs to be created before the relay selector, as migration code depends on it. But the relay selector must be created after that code runs.
2. If the API client selects bridges on its own, then they must be correctly forwarded to the TSM in the `mullvad-problem-report`, etc. The correct solution there might be to add a daemon command that sets the allowed endpoint.
</blockquote></details>

Could the relay selector not be created before the API client? AFAIU, it only needs relay settings, which might well be deserialized before the account migration takes place.
___
*[`mullvad-daemon/src/lib.rs` line 744 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520#-N0GAtQ33w-C3WK7E0Tc:-N0VGEwM-kPv5puEmHNw:b-zf5jvv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9598f5761f71d0ca4b6c6d6066d3015f7e9d40ea/mullvad-daemon/src/lib.rs#L744)):*

There's probably no benefit to separating this from the selector.
</blockquote></details>

It depends on the API handle, so I guess the original intention was to be able to spawn the relay selector before the API runtime.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520)* status: 5 of 11 files reviewed, 4 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[`mullvad-daemon/src/api.rs` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520#-N0GCxNL3Q8oTyaut_ci:-N0VYMsi9RkbZdaiUYiA:b8wvoy1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9598f5761f71d0ca4b6c6d6066d3015f7e9d40ea/mullvad-daemon/src/api.rs#L34)):*

Could the relay selector not be created before the API client? AFAIU, it only needs relay settings, which might well be deserialized before the account migration takes place.
</blockquote></details>

I think doing the settings migration first (including changes to the relay constraints) and the account migration later would make sense.

pinkisemils (commented):
> Reviewed 2 of 11 files at r1, 2 of 2 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520)* status: all files reviewed, 4 unresolved discussions (waiting on @dlon and @pinkisemils)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520)* status: 9 of 12 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`mullvad-daemon/src/api.rs` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520#-N0GCxNL3Q8oTyaut_ci:-N0ViM9x8yiiSRMWuUgP:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9598f5761f71d0ca4b6c6d6066d3015f7e9d40ea/mullvad-daemon/src/api.rs#L34)):*

I think doing the settings migration first (including changes to the relay constraints) and the account migration later would make sense.
</blockquote></details>

Done.
___
*[`mullvad-daemon/src/lib.rs` line 744 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520#-N0GAtQ33w-C3WK7E0Tc:-N0VhzxL96f4W6OuHLPF:bhcbbjy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9598f5761f71d0ca4b6c6d6066d3015f7e9d40ea/mullvad-daemon/src/lib.rs#L744)):*

It depends on the API handle, so I guess the original intention was to be able to spawn the relay selector before the API runtime.
</blockquote></details>

That is true.

pinkisemils (approved):
> Reviewed 2 of 3 files at r4, 1 of 2 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3520)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3530, opened 2022-04-27
Fix manual key rotation

dlon:
There were a couple of problems here:
1. The key rotation response was not sent back to the caller.
2. When updating the device, it was removed. The old device is only supposed to be removed when logging in to a new account or generating a new device.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3530)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3530)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3531, opened 2022-04-27
Pause API interactions when no daemon commands are received

dlon:
This PR makes it so that API requests are paused after some interval (currently 3 days) has elapsed, provided that no daemon commands were received during that time. This makes the daemon a bit less talkative even if the current account is still valid. The timer is only enabled when the tunnel state is not `disconnected`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3531)


dlon (commented):
> Should this not be the other way around?

You are right! The PR description is completely incorrect, but the code does what you'd expect.

> the code only resets the inactivity timer if the daemon changes state

The timer is restarted when any daemon command is received as well.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3531#-:-N0kkmXx0swKQ8lBeNNT:b-nwpfn7)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3540, opened 2022-04-28
Rework output of status subcommand in CLI

pinkisemils:
The output of the status command is reworked to show hostnames instead of IP addresses and trim the fat. The extra information (tunnel protocols, IP addrresses) are now available with the verbose flag.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3540)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Line wrap the file at 100 chars.                                              Th
 - Make login field keep previous value when submitting an incorrect account number in desktop app.
 - Decrease the time it takes to connect to WireGuard relays by sending an ICMP packet immediately.
 - Pause API interactions when the daemon has not been used for 3 days.
+- Simplified output of `mullvad status` command.
 
 ### Fixed
 - Fix the sometimes incorrect time added text after adding time to the account.
--- a/mullvad-cli/src/cmds/connect.rs
+++ b/mullvad-cli/src/cmds/connect.rs
@@ -34,7 +34,7 @@ impl Command for Connect {
             if let Some(mut receiver) = receiver_option {
                 while let Some(state) = receiver.next().await {
                     let state = state?;
-                    format::print_state(&state);
+                    format::print_state(&state, false);
                     match state.state.unwrap() {
                         State::Connected(_) => return Ok(()),
                         State::Error(_) => return Err(Error::CommandFailed("connect")),
--- a/mullvad-cli/src/cmds/disconnect.rs
+++ b/mullvad-cli/src/cmds/disconnect.rs
@@ -34,7 +34,7 @@ impl Command for Disconnect {
             if let Some(mut receiver) = receiver_option {
                 while let Some(state) = receiver.next().await {
                     let state = state?;
-                    format::print_state(&state);
+                    format::print_state(&state, false);
                     match state.state.unwrap() {
                         Disconnected(_) => return Ok(()),
                         _ => {}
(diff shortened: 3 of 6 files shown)


faern (changes requested):
> Reviewed 3 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3540)* status: all files reviewed, 3 unresolved discussions (waiting on @pinkisemils)

___
*[`mullvad-cli/src/format.rs` line 150 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3540#-N0okdzI3Wrsfc3xV-AP:-N0okdzI3Wrsfc3xV-AQ:b-smw60d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1c9b1636be078848003b5e24880b43265084d17e/mullvad-cli/src/format.rs#L150)):*
> ```XML
>         .map(|addr| format!(" ({addr})"))
>         .unwrap_or_else(String::new);
>     format!("{hostname}/{protocol}{sockaddr_suffix}")
> ```

I don't think the protocol should be shown alongside the hostname. On technical level / abstraction level the transport protocol belongs with the IP addr etc, and not with the hostname. I now for example get:

```
Connected to se17-wireguard/UDP in Malmö, Sweden
```

I don't think `/UDP` belongs there as it's on another technical level.

The way I prefer to think about `mullvad status` vs `mullvad status -v` is that the former answers only the question "What's the current status and where am I connected". While the verbose mode also answers the question "**how** am I connected?".

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3540)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[`mullvad-cli/src/format.rs` line 150 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3540#-N0okdzI3Wrsfc3xV-AP:-N0pTGsH7sf2p_PeIc1j:b-1tf5dr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1c9b1636be078848003b5e24880b43265084d17e/mullvad-cli/src/format.rs#L150)):*

I don't think the protocol should be shown alongside the hostname. On technical level / abstraction level the transport protocol belongs with the IP addr etc, and not with the hostname. I now for example get:

```
Connected to se17-wireguard/UDP in Malmö, Sweden
```

I don't think `/UDP` belongs there as it's on another technical level.

The way I prefer to think about `mullvad status` vs `mullvad status -v` is that the former answers only the question "What's the current status and where am I connected". While the verbose mode also answers the question "**how** am I connected?".
</blockquote></details>

That's a fair point, I've changed it so that the transport protocol is only shown in _verbose_ mode.
___
*[`mullvad-cli/src/cmds/status.rs` line 64 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3540#-N0oKIjZBX5gCWJ6fQAU:-N0pTXZV3ic1iU8vv9iC:b-i8wig4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1c9b1636be078848003b5e24880b43265084d17e/mullvad-cli/src/cmds/status.rs#L64)):*

Should the second argument not be `verbose` here? I also kind of wonder if we need both `location` and `verbose`.
</blockquote></details>

The output here should always be a single line, and the location will be printed further down in full anyway.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3540)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3540)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

PR #3544, opened 2022-04-29
Remove pubkey from `device.json`

dlon:
This PR removes the WireGuard pubkey from `device.json`. The key must now always be derived from the private key, meaning that they cannot be out of sync.

Some other changes and fixes:
* Log the result of `validate_device` again.
* Fixed a regression: Calling `validate_device` lots of times would always send a request to the API. It is only supposed to do so if at least 10 seconds have passed since the last call.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3544)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3544)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`mullvad-types/src/device.rs` line 66 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3544#-N0pAqXsEBxfN_Wn_Rbl:-N13nRAt3nzeGumgMung:bli5lz2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fcfee038a47a67bff2878fb0f55aaf9ae4acc2de/mullvad-types/src/device.rs#L66)):*

It'd be cool to name this type `Device` and move the current `Device` struct to the API crate. Then the naming can become slightly less awkward.
</blockquote></details>

Perhaps it would be better to just have one "private" `Device` type (containing a private key) and one "public" type?

`mullvad_types`:

```
pub struct Device {
    pub id: DeviceId,
    pub name: DeviceName,
    pub pubkey: PublicKey,
    pub ports: Vec<DevicePort>,
}
```

`mullvad_daemon::device`:

```
pub struct Device {
    pub id: DeviceId,
    pub name: DeviceName,
    pub wg_data: wireguard::WireguardData,
    pub ports: Vec<DevicePort>,
}
```

pinkisemils (approved):
> Reviewed 10 of 10 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3544)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3553, opened 2022-05-05
Add tags to filename artifacts

Jontified:
Causes build.sh to append "+<commit-tag>" to the end of artifacts (.deb, .rpm, etc) if the build is a development build and this commit has a tag. This is useful because we're able to more easily find the tag name on releases.mullvad.net.
Opted to not change the PRODUCT_VERSION inside of build.sh since having an appended "+<tag>" caused issues with electron builder and other build tools. Instead just rename the artifacts after building.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -316,10 +316,10 @@ if [[ "$(uname -s)" == "MINGW"* ]]; then
 
     if [[ "$SIGN" == "true" ]]; then
         CPP_BINARIES=(
-            windows/winfw/bin/x64-$CPP_BUILD_MODE/winfw.dll
-            windows/windns/bin/x64-$CPP_BUILD_MODE/windns.dll
-            windows/winnet/bin/x64-$CPP_BUILD_MODE/winnet.dll
-            windows/driverlogic/bin/x64-$CPP_BUILD_MODE/driverlogic.exe
+            "windows/winfw/bin/x64-$CPP_BUILD_MODE/winfw.dll"
+            "windows/windns/bin/x64-$CPP_BUILD_MODE/windns.dll"
+            "windows/winnet/bin/x64-$CPP_BUILD_MODE/winnet.dll"
+            "windows/driverlogic/bin/x64-$CPP_BUILD_MODE/driverlogic.exe"
             # The nsis plugin is always built in 32 bit release mode
             windows/nsis-plugins/bin/Win32-Release/*.dll
         )
--- a/build.sh
+++ b/build.sh
@@ -379,8 +379,17 @@ esac
 popd
 
 SEMVER_VERSION=$(echo "$PRODUCT_VERSION" | sed -Ee 's/($|-.*)/.0\1/g')
+# `git tag --points-at` defaults to point at HEAD
+current_head_commit_tag=$(git tag --points-at)
 for semver_path in dist/*"$SEMVER_VERSION"*; do
-    product_path=$(echo "$semver_path" | sed -Ee "s/$SEMVER_VERSION/$PRODUCT_VERSION/g")
+    # If there is a tag for this commit then we append that to the produced artifacts
+    # We don't want to change the actual PRODUCT_VERSION as metadata in the form of +<metadata> is ignored by electron builder etc
+    version_suffix=""
+    if [[ -n "$current_head_commit_tag" ]]; then
+        # Remove disallowed version characters from the tag
+        version_suffix="+${current_head_commit_tag//[^0-9a-z_-]/}"
+    fi
+    product_path=$(echo "$semver_path" | sed -Ee "s/$SEMVER_VERSION/$PRODUCT_VERSION$version_suffix/g")
     log_info "Moving $semver_path -> $product_path"
     mv "$semver_path" "$product_path"
 


faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553)* status: all files reviewed, 2 unresolved discussions (waiting on @Jontified)

___
*[`build.sh` line 383 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553#-N1NgjHs7RYg9daLVzl7:-N1NgjHs7RYg9daLVzl8:b-bqxlq4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a514888e31a8cd17003cf7cc97f05d457cc9a8f9/build.sh#L383)):*
> ```Shell
> SEMVER_VERSION=$(echo "$PRODUCT_VERSION" | sed -Ee 's/($|-.*)/.0\1/g')
> for semver_path in dist/*"$SEMVER_VERSION"*; do
>     # If there is a tag for this commit then we append that to the produced binaries
> ```

Nitpick, but I think "binaries" should be changed to "artifacts" here. In this script we mostly use binaries to refer to the files `mullvad-daemon`, `mullvad-problem-report` and the other executables that make up the app. We mostly use artifacts to mean the installers produced by the script.
___
*[`build.sh` line 386 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553#-N1NgNWu7gxAPpCKfH1T:-N1NgNWu7gxAPpCKfH1U:byb8jyb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a514888e31a8cd17003cf7cc97f05d457cc9a8f9/build.sh#L386)):*
> ```Shell
>     # We don't want to change the actual PRODUCT_VERSION as metadata in the form of +<metadata> is ignored by electron builder etc
>     # `git tag --points-at` defaults to point at HEAD
>     TAG=$(git tag --points-at)
> ```

This could be computed outside the loop to avoid doing it once per artifact. Like we do with `SEMVER_VERSION`.

I'm thinking that maybe it should be named `current_head_commit_tag` to match the variable we already have named `current_head_commit_hash`? This script is old and does not 100% conform to our own best practices. It's very unclear what is to be treated as a constant vs a non constant variable here. So the use of uppercase vs lowercase is a bit unclear. But apparently we went with lowercase for the HEAD commit hash. And I'd say the tag name attached to HEAD is very related. So for the sake of consistency let's treat them equally non-const?
___
*[`build.sh` line 389 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553#-N1NiBGD-xZOhhWKiDLX:-N1NiBGD-xZOhhWKiDLY:b-eq8p72) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a514888e31a8cd17003cf7cc97f05d457cc9a8f9/build.sh#L389)):*
> ```Shell
>     if [[ -n $TAG ]]; then
>         # Remove disallowed version characters from the tag
>         TAG=${TAG//[^0-9a-z_-]/}
> ```

Nitpick, but I think this variable should be in lowercase, since it's not a constant but rather something we compute "at runtime" (bash :shrug:).

if `TAG` is moved outside the loop, this should have a different name. Even if the operation is idempotent we should avoid applying the operation multiple times to the same data, as to not overwrite itself. 

How about
```bash
cleaned_tag=${current_head_commit_tag//[^0-9a-z_-]/}
```

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[`build.sh` line 386 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553#-N1NgNWu7gxAPpCKfH1T:-N1btCEJ21kf_l6aj9rg:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a514888e31a8cd17003cf7cc97f05d457cc9a8f9/build.sh#L386)):*

This could be computed outside the loop to avoid doing it once per artifact. Like we do with `SEMVER_VERSION`.

I'm thinking that maybe it should be named `current_head_commit_tag` to match the variable we already have named `current_head_commit_hash`? This script is old and does not 100% conform to our own best practices. It's very unclear what is to be treated as a constant vs a non constant variable here. So the use of uppercase vs lowercase is a bit unclear. But apparently we went with lowercase for the HEAD commit hash. And I'd say the tag name attached to HEAD is very related. So for the sake of consistency let's treat them equally non-const?
</blockquote></details>

Done.
___
*[`build.sh` line 389 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553#-N1NiBGD-xZOhhWKiDLX:-N1btDxi1FxJZNlq1xKa:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a514888e31a8cd17003cf7cc97f05d457cc9a8f9/build.sh#L389)):*

Nitpick, but I think this variable should be in lowercase, since it's not a constant but rather something we compute "at runtime" (bash :shrug:).

if `TAG` is moved outside the loop, this should have a different name. Even if the operation is idempotent we should avoid applying the operation multiple times to the same data, as to not overwrite itself. 

How about
```bash
cleaned_tag=${current_head_commit_tag//[^0-9a-z_-]/}
```
</blockquote></details>

Done.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`build.sh` line 383 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553#-N1dDrqZ0FIQz2-sFoej:-N1dDrqZ0FIQz2-sFoek:bz0weem) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/61cf96d626dcfe10e9bb525d8e22c3e2ce75da84/build.sh#L383)):*
> ```Shell
> SEMVER_VERSION=$(echo "$PRODUCT_VERSION" | sed -Ee 's/($|-.*)/.0\1/g')
> # `git tag --points-at` defaults to point at HEAD
> current_head_commit_tag=$(git tag --points-at)
> ```

I love that you add documentation! :heart_eyes:  Just noting that this line is just on the edge of being superfluous/redundant IMHO. Given the name of the variable, it's pretty clear what it's expected to contain. So a reader can probably assume the thing in the comment pretty easily. Another way is to explicitly add `HEAD` to the command.

This is fine. I'm just saying that it's probably not 100% needed.
___
*[`build.sh` line 390 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553#-N1dFFLs9V5mJ30VIcZo:-N1dFFLt2vS47yP9vvbM:b-9o887d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/61cf96d626dcfe10e9bb525d8e22c3e2ce75da84/build.sh#L390)):*
> ```Shell
>     if [[ -n "$current_head_commit_tag" ]]; then
>         # Remove disallowed version characters from the tag
>         cleaned_tag="+${current_head_commit_tag//[^0-9a-z_-]/}"
> ```

Now I'm getting nitpicky. But if the variable contains `+$tag` or is optionally empty, then the contents is not perfectly described with `cleaned_tag` any longer. The full explanation would be `empty_string_or_tag_suffix`, which is a bit long. But maybe `version_suffix` is slightly more explanatory of the contents? :thinking:

faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553#-:-N1dFwHp1-kkYXfukjA2:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`build.sh` line 390 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3553#-N1dFFLs9V5mJ30VIcZo:-N1grV80BErHGPwoNpOz:bgnv5vh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/61cf96d626dcfe10e9bb525d8e22c3e2ce75da84/build.sh#L390)):*

Now I'm getting nitpicky. But if the variable contains `+$tag` or is optionally empty, then the contents is not perfectly described with `cleaned_tag` any longer. The full explanation would be `empty_string_or_tag_suffix`, which is a bit long. But maybe `version_suffix` is slightly more explanatory of the contents? :thinking: 
</blockquote></details>

Or `optional_tag`

faern:
This is already merged, but due to not pushing the feature branch before merging, github did not properly detect that.

PR #3574, opened 2022-05-11
Check API response code to determine if a device is valid

dlon:
Previously, the response didn't contain a custom status code in the response body, so the device was inferred to be invalid if the HTTP status code was 404. This may cause a device to be "revoked" if an endpoint is removed/deprecated or deployed incorrectly, so this PR switches to using the custom json response code instead.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3574)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3574#-:-N1nhZAG3-piXPpZ2XU2:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3577, opened 2022-05-12
Tag artifact names only when developing

Jontified:
Moves the code that tags artifacts to the buildserver and away from `build.sh`.
This means that the tagging code does not touch anything that a user interacts with and allows us to dodge having to implement the same logic in the `build-apk.sh` script.
The buildserver logic renames the artifact files that it can find inside of `/dist` and appends a cleaned tag if there is one. It only touches the artifacts that include `*-dev-*` in order to not modify release versions.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3577)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -379,16 +379,8 @@ esac
 popd
 
 SEMVER_VERSION=$(echo "$PRODUCT_VERSION" | sed -Ee 's/($|-.*)/.0\1/g')
-current_head_commit_tag=$(git tag --points-at HEAD)
 for semver_path in dist/*"$SEMVER_VERSION"*; do
-    # If there is a tag for this commit then we append that to the produced artifacts
-    # We don't want to change the actual PRODUCT_VERSION as metadata in the form of +<metadata> is ignored by electron builder etc
-    version_suffix=""
-    if [[ -n "$current_head_commit_tag" ]]; then
-        # Remove disallowed version characters from the tag
-        version_suffix="+${current_head_commit_tag//[^0-9a-z_-]/}"
-    fi
-    product_path=$(echo "$semver_path" | sed -Ee "s/$SEMVER_VERSION/$PRODUCT_VERSION$version_suffix/g")
+    product_path=$(echo "$semver_path" | sed -Ee "s/$SEMVER_VERSION/$PRODUCT_VERSION/g")
     log_info "Moving $semver_path -> $product_path"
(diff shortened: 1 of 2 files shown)


faern (approved):
:lgtm:

> Reviewed 2 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3577#-:-N1rcOxzDs6oSoRecG7c:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`ci/buildserver-build.sh` line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3577#-N1sjd3xE14-t6SW-MfI:-N1sjd3xE14-t6SW-MfJ:b-3vctsr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5273a5cfc6ccd0581cd56080944c5001f1543ec4/ci/buildserver-build.sh#L136)):*
> ```Shell
>       # Will only match paths that include *-dev-* which means release builds will not be included
>       for artifact_path in dist/MullvadVPN-*-dev-*{_amd64.deb,_x86_64.rpm,.exe,.pkg,.apk,.aab}; do
>           tagged_name=$(echo $artifact_path | sed -E "s/(.*)(_amd64\.deb|_x86_64\.rpm|\.exe|\.pkg|\.apk|\.aab)/\1$version_suffix\2/")
> ```

I'm not a huge fan of having to list all the file endings more than once. And here we add it twice even.. I don't see a better way out myself, but this feels fragile.

One problem with this is that the code path is not executed that often. We don't build tagged builds all that often. So it would be easy to introduce an issue not spotted in a while.

As I said. I don't know a better solution right now, so I'll approve this.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3577)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (approved):
This works on the build servers! :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3577#-:-N2B78Pz2XLhzb9uBbTx:b3754oc)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3580, opened 2022-05-12
Remove temporary files used during setup

dlon:
The NSIS installer/uninstaller leaves some files in `%TEMP%` that remain even after uninstalling:

```
mullvad-setup.exe
winfw.dll
driverlogic.exe
mullvad-split-tunnel/
nsXXX.tmp/log.dll
```

Most are straightforward to delete.

`log.dll` is not cleaned up because the module isn't freed until the process dies. NSIS normally unloads plugins after each function call. `log.dll` has global state, so it increases its own reference count by a ton to avoid being unloaded.

Unfortunately, NSIS is inconsistent and calls `FreeLibrary()` many more times than it calls `LoadLibrary()`, so we have no idea how many times to decrement the refcount. As a workaround, `FreeLibrary()` is called until the module is freed.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3580)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ Line wrap the file at 100 chars.                                              Th
 #### Android
 - Remove WireGuard view as it's no longer needed with the new way of managing devices.
 
+### Fixed
+#### Windows
+- Be more scrupulous about removing temporary files used by the installer and uninstaller.
+
 ### Security
 #### Android
 - Prevent location request responses from being received outside the tunnel when in the connected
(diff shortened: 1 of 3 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3580)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3580)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3591, opened 2022-05-16
Replace complex expression with for loop

Jontified:
Fix buildserver bug that causes release builds to not build since the replaced expression will run if there are no pattern matches.
Replace a complicated expression for renaming files in the buildserver with a for loop with a slightly less complex expression.
This hopefully will reduce their fragility.


faern (approved):
Looks good! Let me try it on the build servers before merging.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3591#-:-N2CByDe1_hLkHYyU1J3:b-qvxh32)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3609, opened 2022-05-23
Set mtu on wireguard

Jontified:
Draft for setting the MTU of the wireguard tunnel.
When starting the Wireguard tunnel checks the MTU for the route to the connection endpoint and then sets the MTU in the Wireguard parameters to whatever is lesser between the provided MTU and 1400.  We also log when the provided MTU is different from 1500 in order to detect weirdness.

Next steps include detecting a better MTU over a real network.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3609)

### DIFF
--- a/talpid-core/src/routing/linux.rs
+++ b/talpid-core/src/routing/linux.rs
@@ -113,6 +113,9 @@ pub enum Error {
     #[error(display = "No netlink response for route query")]
     NoRouteError,
 
+    #[error(display = "No link found")]
+    LinkNotFoundError,
+
     /// Unable to create routing table for tagged connections and packets.
     #[error(display = "Cannot find a free routing table ID")]
     NoFreeRoutingTableId,
@@ -362,6 +365,9 @@ impl RouteManagerImpl {
             }
             RouteManagerCommand::GetDestinationRoute(destination, set_mark, result_tx) => {
                 let _ = result_tx.send(self.get_destination_route(&destination, set_mark).await);
+            }
+            RouteManagerCommand::GetMtuForRoute(ip, result_tx) => {
+                let _ = result_tx.send(self.get_mtu_for_route(ip).await);
             }
             RouteManagerCommand::ClearRoutes => {
                 log::debug!("Clearing routes");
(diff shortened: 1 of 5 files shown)


pinkisemils (changes requested):
> Reviewed 1 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3609)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on @Jontified)

___
*[`talpid-core/src/tunnel/mod.rs` line 232 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3609#-N2pq4_b31NcarFFuJfx:-N2pq4_b31NcarFFuJfy:b-fgjh1n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26d35fcc906c0f300e303d2750ba0235e4282c04/talpid-core/src/tunnel/mod.rs#L232)):*
> ```XML
>                         let node = route.get_node();
>                         let device = node.get_device();
>                         match device {
> ```

This `match` statement could match both the `node` and `device`, decreasing some levels of nesting. 
___
*[`talpid-core/src/tunnel/mod.rs` line 246 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3609#-N2pqYCfEMhLDG-kVI-b:-N2poPNy111JujuIIScW:b-no0tu9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26d35fcc906c0f300e303d2750ba0235e4282c04/talpid-core/src/tunnel/mod.rs#L246)):*
> ```XML
>                                 }
>                                 let upstream_mtu = std::cmp::min(1400, mtu);
>                                 params.options.mtu = Some(upstream_mtu);
> ```

The MTU set for the WireGuard interface should be 80 bytes lower than what the physical interface has, to account for the WireGuard header size. And it should never be less than 576 for IPv4. Technically, we should error out if it ends up less than 1280 if IPv6 is enabled too.
___
*[`talpid-core/src/tunnel/mod.rs` line 252 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3609#-N2pon9fEoQtlV9Wpr41:-N2pon9fEoQtlV9Wpr42:bn6ovd0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26d35fcc906c0f300e303d2750ba0235e4282c04/talpid-core/src/tunnel/mod.rs#L252)):*
> ```XML
>                                 Some(ip) => attempted_ip = ip,
>                                 None => {
>                                     log::error!(
> ```

It's ok to panic here since it would be a bug if a route had neither an address nor a network device.
___
*[`talpid-core/src/tunnel/mod.rs` line 274 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3609#-N2po_WmClzfKy6f7gm_:-N2po_WmClzfKy6f7gma:b-o9wd82) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26d35fcc906c0f300e303d2750ba0235e4282c04/talpid-core/src/tunnel/mod.rs#L274)):*
> ```XML
>             return Err(Error::AssignMtuError);
>         });
>         result
> ```

Typically, there's no need to assign a value to a variable if it's going to be returned in the next statement.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3609)* status: 1 of 5 files reviewed, 5 unresolved discussions (waiting on @Jontified)

___
*[`talpid-core/src/tunnel/mod.rs` line 230 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3609#-N2v0yF37fx6kyWNXMYU:-N2v0yF37fx6kyWNXMYV:b-83cqmz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7ef5b956693e63798ad08a33e6293c2c1af51e96/talpid-core/src/tunnel/mod.rs#L230)):*
> ```XML
>                 mtu.checked_sub(WIREGUARD_HEADER_SIZE)
>                     .unwrap_or(MIN_IPV4_MTU),
>                 MIN_IPV4_MTU,
> ```

Instead of duplicating all four max+checked_sub+unwrap_or function calls, we should be able to do it once if we do something like:

```
let min_mtu = match params.generic_options.enable_ipv6 {
    true => MIN_IPV6_MTU,
    false => MIN_IPV4_MTU,
};
```

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3609)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @Jontified)

pinkisemils (commented):
> Reviewed 1 of 5 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3609)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @Jontified)

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3609)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #3619, opened 2022-05-28
gradlew violates POSIX


carbongreat13:
### Issue description
Same as #3617 .
Lines [97](https://github.com/mullvad/mullvadvpn-app/blob/584a4f31cb64a4b02f81a2bade1fa99eeb4d98ca/android/gradlew#L97), and [98](https://github.com/mullvad/mullvadvpn-app/blob/584a4f31cb64a4b02f81a2bade1fa99eeb4d98ca/android/gradlew#L98) are POSIX violations.

Issue #3618, opened 2022-05-28
build-logo-icons.sh violates POSIX


carbongreat13:
### Issue description
Same as #3617 .
Lines [14](https://github.com/mullvad/mullvadvpn-app/blob/144987b5a2b5aff3d003a40ebfb9a80ea0d60738/ios/build-wireguard-go.sh#L14), [16](https://github.com/mullvad/mullvadvpn-app/blob/144987b5a2b5aff3d003a40ebfb9a80ea0d60738/ios/build-wireguard-go.sh#L16), [25](https://github.com/mullvad/mullvadvpn-app/blob/144987b5a2b5aff3d003a40ebfb9a80ea0d60738/ios/build-wireguard-go.sh#L25), and [27](https://github.com/mullvad/mullvadvpn-app/blob/144987b5a2b5aff3d003a40ebfb9a80ea0d60738/ios/build-wireguard-go.sh#L27) are POSIX violations.

faern:
Fixed in #3625

Issue #3617, opened 2022-05-28
prebuild.sh violates POSIX


carbongreat13:
### Issue description
[The sheban of prebuild.sh](https://github.com/mullvad/mullvadvpn-app/blob/13f1ec7a6135c33bb60e39641de61657b6c7ba6a/ios/prebuild.sh#L1) is for POSIX sh.

However, [line 13](https://github.com/mullvad/mullvadvpn-app/blob/13f1ec7a6135c33bb60e39641de61657b6c7ba6a/ios/prebuild.sh#L13) violates POSIX. This can be checked with [checkbashisms](https://packages.qa.debian.org/d/devscripts.html).
The script must be modified to be POSIX compliant.

faern:
> The script must be modified to be POSIX compliant.

Can you please explain why you feel like the script must be posix compliant?

What shebang would you suggest to use if we were to not change the posix compliance?

carbongreat13:
> Can you please explain why you feel like the script must be posix compliant?

I'm not sure if the shebang is `#!/bin/sh` so. That's all. `#!/bin/sh` and the script must be POSIX compliant.

> What shebang would you suggest to use if we were to not change the posix compliance?

`#!/bin/bash`

PR #3625, opened 2022-05-31
Correct bash shebang

faern:
Fixing #3617, #3618 and #3619.

@albin-mullvad Does it even make sense to edit `gradlew` or should we let it be?

@pronebird is the change from `exit -1` into `exit 1` going to break something? I don't think negative exit codes are very standard.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3625)

### DIFF
--- a/ios/build-wireguard-go.sh
+++ b/ios/build-wireguard-go.sh
@@ -25,7 +25,7 @@ fi
 RESOLVED_SOURCE_PACKAGES_PATH="$( cd "$SOURCE_PACKAGES_PATH" && pwd -P )"
 if [ "$RESOLVED_SOURCE_PACKAGES_PATH" == "" ]; then
   echo "Failed to resolve the SourcePackages path: $SOURCE_PACKAGES_PATH"
-  exit -1
+  exit 1
 fi
 
 # Compile the path to the Makefile directory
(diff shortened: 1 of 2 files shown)


albin-mullvad (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3625)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`android/gradlew` line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3625#-N3P6GKd--hlTAABWdNA:-N3P6GKd--hlTAABWdNB:bavj8h0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/db2740d0300df6b02b7884647b14c63c572f6f48/android/gradlew#L1)):*
> ```Shell
> #!/usr/bin/env bash
> ```

To answer your question: I think we should avoid editing the gradle wrapper as it's automatically generated.

albin-mullvad (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3625)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

pronebird (approved):
None of these changes impact the build pipeline. Both negative and positive exit codes are treated as an error by xcodebuild, so either works for me. 👍

PR #3646, opened 2022-06-10
Fix missing IPC event unregistration

albin-mullvad:
The app is not unregistering from events before unbinding from the service, which result in messages (duplicates) delivered to old message handlers after the service has been bound again. This issue has been hidden in the current implementation as the messages were not propagated further within the app.

The issue is fixed by making sure to unregister before unbinding.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3646)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Line wrap the file at 100 chars.                                              Th
 
 #### Android
 - Fix unused dependencies loaded in the service/tile DI graph.
+- Fix missing IPC message unregistration causing multiple copies of some messages to be received.
 
 ### Security
 #### Android
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/serviceconnection/ServiceConnectionContainer.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/serviceconnection/ServiceConnectionContainer.kt
@@ -20,7 +20,7 @@ import org.koin.core.scope.get
 // the service and to get values received from events.
 @OptIn(KoinApiExtension::class)
 class ServiceConnectionContainer(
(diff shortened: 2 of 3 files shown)


pinkisemils (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3646)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
These changes look good but I have to wonder if we still need the indirection of using an integer ID to know what/where to deregister. AFAICT, there will only ever be a single listener, can't the listener be registered by it'd identity hash instead?

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3646#-:-N4WxvD31T-0dP0Cc7UH:b-jq8vli)* status: :shipit: complete! all files reviewed, all discussions resolved

albin-mullvad (commented):
Both the VpnService and TileService register as listeners using different implementations. The TileService uses the `net.mullvad.mullvadvpn.ipc.ServiceConnection`, as seen here:
https://github.com/mullvad/mullvadvpn-app/blob/f249bc175d21c3f732541266be9004700a913487/android/app/src/main/kotlin/net/mullvad/mullvadvpn/ipc/ServiceConnection.kt#L89

I agree that this should be improved (and unified), however I would prefer to not start that type of work right now.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3646#-:-N4X1Did8ygSZksVMauR:b-qkhsnc)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3667, opened 2022-06-15
Clippy semantic manual fixes

Jontified:
This PR fixes all of the remaining Clippy warnings and deals with the less obvious cases on what to do. It attempts as much as possible to not change behavior of code.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)


dlon (changes requested):
> Reviewed 3 of 69 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 10 of 69 files reviewed, 2 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[`mullvad-api/src/address_cache.rs` line 22 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4bvFwZ1Ul6W9cVSU68:-N4bxUkX0u3mmqYohbY2:b5ss3mr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-api/src/address_cache.rs#L22)):*

This and the `Empty` variant are not used anywhere.
</blockquote></details>

Maybe `set_address` should return this instead of an `io::Error`.

dlon (changes requested):
> Reviewed 21 of 69 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 31 of 69 files reviewed, 8 unresolved discussions (waiting on @dlon, @Jontified, and @pinkisemils)

___
*[`mullvad-daemon/src/lib.rs` line 86 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4fVifSBMDzcXwoDu8Q:-N4fVifSBMDzcXwoDu8R:bfy6pma) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-daemon/src/lib.rs#L86)):*
> ```XML
> 
> pub type ResponseTx<T, E> = oneshot::Sender<Result<T, E>>;
> type ShutdownTasks = Vec<Pin<Box<dyn Future<Output = ()>>>>;
> ```

Could this maybe be a vec of [`LocalBoxFuture`](https://docs.rs/futures/latest/futures/future/type.LocalBoxFuture.html)?
___
*[`mullvad-daemon/src/tunnel.rs` line 3 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4fgJIpDJoSGQkBZuad:-N4fgJIpDJoSGQkBZuae:bwh2l05) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-daemon/src/tunnel.rs#L3)):*
> ```XML
> use std::{future::Future, pin::Pin, sync::Arc};
> 
> use tokio::sync::Mutex;
> ```

I don't think you should switch to an async mutex here. We're not holding it for a significant amount of time.
___
*[`talpid-core/src/tunnel/wireguard/wireguard_kernel/wg_message.rs` line 532 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4fdJf5Dhf50bWkH9uf:-N4fdJf5Dhf50bWkH9ug:b65lkr8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel/wireguard/wireguard_kernel/wg_message.rs#L532)):*
> ```XML
>             IpAddr(ip_addr) => {
>                 buffer
>                     .write_all(&ip_addr_to_bytes(ip_addr))
> ```

Nice. This seems to have been incorrect before. `&mut [u8]::write` is infallible.
___
*[`talpid-core/src/tunnel_state_machine/connected_state.rs` line 122 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4fe5a442hKNofeoPul:-N4fe5a442hKNofeoPum:batwkhd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/connected_state.rs#L122)):*
> ```XML
>     }
> 
>     #[allow(clippy::needless_borrow)]
> ```

Is this attribute needed? Could you not just remove the `&` from the second `dns_ips`?
___
*[`talpid-core/src/tunnel_state_machine/connecting_state.rs` line 145 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4feb3tAaxiFDjAyVZ2:-N4feb3tAaxiFDjAyVZ3:b-9k6pvb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/connecting_state.rs#L145)):*
> ```XML
>             };
> 
>             let init_args = TunnelCreationArguments {
> ```

Is there any reasoning behind grouping (only) these parameters in particular? It looks a bit arbitrary to me.
___
*[`talpid-core/src/tunnel_state_machine/mod.rs` line 136 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ff5ArDPoWzOF0ju4e:-N4ff5ArDPoWzOF0ju4f:bncatkb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/mod.rs#L136)):*
> ```XML
>     let weak_command_tx = Arc::downgrade(&command_tx);
> 
>     let init_args = TunnelStateMachineInitArgs {
> ```

Not sure how I feel about moving every single parameter into a single struct. Does that really make things more readable?
___
*[`talpid-core/src/tunnel_state_machine/mod.rs` line 238 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ffKco5ONMd1yhCiLJ:-N4ffKco5ONMd1yhCiLK:b-t14hb8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/mod.rs#L238)):*
> ```XML
>         init_args: TunnelStateMachineInitArgs<impl TunnelParametersGenerator>,
>         let settings = init_args.settings;
> ```

Using `init_args.settings`, etc., directly, would be better, IMO. But see the above comment.

dlon (changes requested):
> Reviewed 26 of 69 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 57 of 69 files reviewed, 12 unresolved discussions (waiting on @dlon, @Jontified, and @pinkisemils)

___
*[`mullvad-relay-selector/src/lib.rs` line 278 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4gCSmrA9SGO8Jkrwy9:-N4gCSmrA9SGO8JkrwyA:bz3he0m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-relay-selector/src/lib.rs#L278)):*
> ```XML
>         match &config.relay_settings {
>             RelaySettings::CustomTunnelEndpoint(custom_relay) => Ok((
>                 SelectedRelay::Custom(Box::new(custom_relay.clone())),
> ```

Why is the relay boxed now?
___
*[`mullvad-relay-selector/src/updater.rs` line 60 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4gCiJrAidbtt1H9Drj:-N4gCiJrAidbtt1H9Drk:b-lzjuv5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-relay-selector/src/updater.rs#L60)):*
> ```XML
> 
> impl RelayListUpdater {
>     pub fn new_handle(
> ```

If this has to be renamed, I think `spawn` would be more consistent with how we start actors elsewhere.
___
*[`mullvad-types/src/relay_constraints.rs` line 632 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4gDML7ENIOCPL3eUnf:-N4gDML7ENIOCPL3eUng:b9ydsua) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-types/src/relay_constraints.rs#L632)):*
> ```XML
>                     !matches!(
>                         &constraints.port,
>                         Constraint::Only(TransportPort {
> ```

Would it be too unreadable use `matches!` on the whole thing (i.e., on `update`)? Like this, I mean:

```
Some(OpenvpnConstraints {
    port: Constraint::Only(TransportPort {
        protocol: TransportProtocol::Udp,
        ..
    }),
    ..
})
```
___
*[`talpid-core/src/dns/linux/resolvconf.rs` line 52 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4gFyqyEkT-xfMqOqil:-N4gFyqyEkT-xfMqOqim:b-rkjgnm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/dns/linux/resolvconf.rs#L52)):*
> ```XML
> 
>         // Check if resolvconf is managing DNS by /etc/resolv.conf
>         if !(is_dnsmasq_running
> ```

I think `!is_dnsmasq_running && !Self::check_if_resolvconf_is_symlinked_correctly() && !Self::check_if_resolvconf_was_generated()` would be more readable than either of these.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 57 of 69 files reviewed, 12 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[`mullvad-api/src/address_cache.rs` line 22 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4bvFwZ1Ul6W9cVSU68:-N4hlSkG3kzHsALSflv1:b650tdj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-api/src/address_cache.rs#L22)):*

Maybe `set_address` should return this instead of an `io::Error`.
</blockquote></details>

I agree with using it in `set_address` and also suggest we use it in `save_to_disk` in order to just return our own `Result`s.  I will remove `Empty`.
___
*[`mullvad-api/src/relay_list.rs` line 348 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4bvzHUAc8VaViCjyMQ:-N4hoInO7GJjToyxddmx:b-1sa65r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-api/src/relay_list.rs#L348)):*

I think instead of appending `_mut`, this change should just remove the `to_` part of the function name. It's not really converting the type or exposing a mutable reference to anything, instead it's mutating the relay value in place. I think the naming of the function may very well remain an open question.
</blockquote></details>

This sounds reasonable. I will change it to `convert_to_lowercase`. This seems like it would be at least as understandable as the previous name if slightly more verbose.
___
*[`mullvad-daemon/src/lib.rs` line 86 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4fVifSBMDzcXwoDu8Q:-N4hrAB0BszMBuxGhwGu:b-54hkpf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-daemon/src/lib.rs#L86)):*

Could this maybe be a vec of [`LocalBoxFuture`](https://docs.rs/futures/latest/futures/future/type.LocalBoxFuture.html)?
</blockquote></details>

That makes sense, fixed.
___
*[`mullvad-daemon/src/tunnel.rs` line 3 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4fgJIpDJoSGQkBZuad:-N4hyRt84MHJaCdhQyf_:b-jquwv1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-daemon/src/tunnel.rs#L3)):*

I don't think you should switch to an async mutex here. We're not holding it for a significant amount of time.
</blockquote></details>

What is the significant downside for switching to an async mutex here? Even if the time we hold this lock is quite short isn't it unnecessary to increase fragility and risk a potential deadlock?
___
*[`mullvad-relay-selector/src/lib.rs` line 278 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4gCSmrA9SGO8Jkrwy9:-N4hyr0pCjhaCPEarl_3:b-rllf8g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-relay-selector/src/lib.rs#L278)):*

Why is the relay boxed now?
</blockquote></details>

The reason is that Clippy complains about there being a large difference in the size of enum variants. I boxed all of the variants that were large. Do you think that could cause issues down the line?
___
*[`mullvad-relay-selector/src/updater.rs` line 60 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4gCiJrAidbtt1H9Drj:-N4hz5HkAwxZD2Ui1SAE:b-vp6lf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-relay-selector/src/updater.rs#L60)):*

If this has to be renamed, I think `spawn` would be more consistent with how we start actors elsewhere.
</blockquote></details>

Good, will fix. I'll fix this for `RequestService` as well.
___
*[`mullvad-types/src/relay_constraints.rs` line 632 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4gDML7ENIOCPL3eUnf:-N4i2sXA1gHlB1WnkMav:by9poav) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-types/src/relay_constraints.rs#L632)):*

Would it be too unreadable use `matches!` on the whole thing (i.e., on `update`)? Like this, I mean:

```
Some(OpenvpnConstraints {
    port: Constraint::Only(TransportPort {
        protocol: TransportProtocol::Udp,
        ..
    }),
    ..
})
```
</blockquote></details>

No, I think it's more readable. Will change to that.
___
*[`talpid-core/src/dns/linux/resolvconf.rs` line 52 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4gFyqyEkT-xfMqOqil:-N4i3s821x9WAfS04t8i:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/dns/linux/resolvconf.rs#L52)):*

I think `!is_dnsmasq_running && !Self::check_if_resolvconf_is_symlinked_correctly() && !Self::check_if_resolvconf_was_generated()` would be more readable than either of these.
</blockquote></details>

Done.
___
*[`talpid-core/src/tunnel_state_machine/connected_state.rs` line 122 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4fe5a442hKNofeoPul:-N4i54o-8YUwgzHZTXSl:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/connected_state.rs#L122)):*

Is this attribute needed? Could you not just remove the `&` from the second `dns_ips`?
</blockquote></details>

Done.
___
*[`talpid-core/src/tunnel_state_machine/connecting_state.rs` line 145 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4feb3tAaxiFDjAyVZ2:-N4i5Qn6AbdPaQKiuGCx:b-hodwbn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/connecting_state.rs#L145)):*

Is there any reasoning behind grouping (only) these parameters in particular? It looks a bit arbitrary to me.
</blockquote></details>

The reasoning is that this was the intersection of parameters of all of the functions with too many arguments. It allows us to only create one type that we can reuse and pass down to the other functions. I was not sure how to group the parameters and this allowed us to only create 1 new type instead of 3.
___
*[`talpid-core/src/tunnel_state_machine/mod.rs` line 136 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ff5ArDPoWzOF0ju4e:-N4i5uRq429P1SzQ4ri5:bvf2bfk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/mod.rs#L136)):*

Not sure how I feel about moving every single parameter into a single struct. Does that really make things more readable?
</blockquote></details>

Not sure what the best grouping is here. It does improve readability by forcing you to name each parameter before calling it. However it might be possible to make it even more readable if you have some more clever grouping. I opted for 1 type instead of several because I thought it made it less complicated, the type is really only meant to be used for passing arguments to this one function anyway so I don't know if it really makes sense to divide it into a bunch more logical types. If we wanted to use them elsewhere in the future I think it makes more sense to create those types then.
___
*[`talpid-core/src/tunnel_state_machine/mod.rs` line 238 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ffKco5ONMd1yhCiLJ:-N4i6slo7yHVLNVg2RC8:b-c6ww5g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/mod.rs#L238)):*

Using `init_args.settings`, etc., directly, would be better, IMO. But see the above comment.
</blockquote></details>

I'm not sure I agree. I think it becomes unnecessarily verbose since we repeat it several times. I'm curious what you think after reading above response.

dlon (commented):
> Reviewed 4 of 69 files at r1, 9 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 60 of 69 files reviewed, 6 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[`mullvad-daemon/src/tunnel.rs` line 3 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4fgJIpDJoSGQkBZuad:-N4kcbmW86rkgPSd5NGm:b-tum6el) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-daemon/src/tunnel.rs#L3)):*

What is the significant downside for switching to an async mutex here? Even if the time we hold this lock is quite short isn't it unnecessary to increase fragility and risk a potential deadlock?
</blockquote></details>

I agree with you after our Slack discussion.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 60 of 69 files reviewed, 6 unresolved discussions (waiting on @dlon, @Jontified, and @pinkisemils)

___
*[`mullvad-relay-selector/src/lib.rs` line 278 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4gCSmrA9SGO8Jkrwy9:-N4kjYjcBgzsPG2mft2C:b-8h9x8m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-relay-selector/src/lib.rs#L278)):*

The reason is that Clippy complains about there being a large difference in the size of enum variants. I boxed all of the variants that were large. Do you think that could cause issues down the line?
</blockquote></details>

No. Is the custom endpoint the larger variant? The problem is just that I doubt that the waste of memory is a problem that justifies boxing everything.

dlon (commented):
> Reviewed 3 of 69 files at r1, 1 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 64 of 69 files reviewed, 6 unresolved discussions (waiting on @dlon, @Jontified, and @pinkisemils)

pinkisemils (changes requested):
> Reviewed 15 of 69 files at r1, 6 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 64 of 69 files reviewed, 5 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`mullvad-setup/src/main.rs` line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ksH28EduWvAByWO0K:-N4ksH28EduWvAByWO0L:b-1s72gb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/mullvad-setup/src/main.rs#L136)):*
> ```XML
> async fn is_older_version(old_version: &str) -> Result<ExitStatus, Error> {
>     let parsed_version =
>         ParsedAppVersion::parse_from_str(old_version).ok_or(Error::ParseVersionStringError)?;
> ```

I think this could be changed to an implementation of `FromStr` or just a function named `parse` instead of `parse_from_str` - it's pretty obvious it is parsing a borrowed string.

pinkisemils (changes requested):
> Reviewed 11 of 69 files at r1, 3 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 65 of 69 files reviewed, 6 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`talpid-core/src/offline/linux.rs` line 28 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4lbDOwC5wn5CN4yzhk:-N4lbDOwC5wn5CN4yzhl:btafeh0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/offline/linux.rs#L28)):*
> ```XML
> 
> impl MonitorHandle {
>     #[allow(clippy::wrong_self_convention)]
> ```

I don't believe this is necessary as I can't get my own clippy to complain about this with and without this.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 65 of 69 files reviewed, 7 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`talpid-core/src/offline/mod.rs` line 35 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4lbKHOEGjVE6tLXFqb:-N4lbKHOEGjVE6tLXFqc:b-f18iw1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/offline/mod.rs#L35)):*
> ```XML
> 
> impl MonitorHandle {
>     #[allow(clippy::wrong_self_convention)]
> ```

Same as the previous comment, I can't get clippy to complain about this function even without this exception.

pinkisemils (changes requested):
> Reviewed 5 of 69 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 65 of 69 files reviewed, 8 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`talpid-core/src/routing/unix.rs` line 28 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4lbkQO3UaDOmD27s_G:-N4lbkQO3UaDOmD27s_H:b-sotkwk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/routing/unix.rs#L28)):*
> ```XML
> pub(crate) use imp::listen_for_default_route_changes;
> 
> #[allow(clippy::module_inception)]
> ```

If there's an exception for Linux, shouldn't there be exceptions for all other modules?

pinkisemils (commented):
> Reviewed 1 of 69 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 65 of 69 files reviewed, 8 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`talpid-core/src/tunnel/mod.rs` line 104 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4lcAKg7HHYkhwNlt6a:-N4lcAKg7HHYkhwNlt6b:b-2xz217) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/tunnel/mod.rs#L104)):*
> ```XML
> pub struct TunnelCreationArguments<'a, L>
> where
>     L: (Fn(TunnelEvent) -> std::pin::Pin<Box<dyn std::future::Future<Output = ()> + Send>>)
> ```

It might make sense to use a type for the `futures` crate here instead of typing out the pinned, boxed, sendable future here.

pinkisemils (changes requested):
> Reviewed 5 of 69 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 67 of 69 files reviewed, 10 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`talpid-core/src/tunnel/mod.rs` line 102 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ledtY60W2VVDl5jFI:-N4ledtY60W2VVDl5jFJ:b-jugru6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/tunnel/mod.rs#L102)):*
> ```XML
> 
> /// Arguments for creating a tunnel.
> pub struct TunnelCreationArguments<'a, L>
> ```

I think it could be named `TunnelArgs` or `TunnelArguments`. 
___
*[`talpid-core/src/tunnel/mod.rs` line 130 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4lcs5-1Lg_yxluqRIB:-N4lcs5-1Lg_yxluqRIC:b-89sj23) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/tunnel/mod.rs#L130)):*
> ```XML
>         retry_attempt: u32,
>         route_manager: RouteManagerHandle,
>         init_args: TunnelCreationArguments<'_, L>,
> ```

These are very good changes!
___
*[`talpid-core/src/tunnel/openvpn/mod.rs` line 377 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ld42n8A32ZXMkuOOQ:-N4ld42n8A32ZXMkuOOR:b-h8wmh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/tunnel/openvpn/mod.rs#L377)):*
> ```XML
> }
> 
> struct OpenVpnTunnelInitArgs {
> ```

This could implement `Default` for test builds just to simplify all of the code in the tests. Or have a test function that initializes just for the test module.
___
*[`talpid-core/src/tunnel/tun_provider/unix.rs` line 25 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ldHKJ3qCHgZmj2mRv:-N4ldHKJ3qCHgZmj2mRw:bxer7zf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/tunnel/tun_provider/unix.rs#L25)):*
> ```XML
> pub struct UnixTunProvider;
> 
> impl Default for UnixTunProvider {
> ```

Can't this implementation be derived?
___
*[`talpid-core/src/tunnel/wireguard/logging.rs` line 115 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ldk234_rdpt67ilCY:-N4ldk234_rdpt67ilCZ:b-etd01n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/tunnel/wireguard/logging.rs#L115)):*
> ```XML
>         let level = match level {
>             WG_GO_LOG_VERBOSE => LogLevel::Verbose,
>             WG_GO_LOG_ERROR => LogLevel::Error,
> ```

This could be simplified to only match against `WG_GO_LOG_VERBOSE` and `_`.
___
*[`talpid-core/src/tunnel/wireguard/mod.rs` line 45 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4leAFm6X1QSE4QZB5h:-N4leAFm6X1QSE4QZB5i:b-f5s10z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/tunnel/wireguard/mod.rs#L45)):*
> ```XML
> type Result<T> = std::result::Result<T, Error>;
> type EventCallback = Box<
>     dyn (Fn(TunnelEvent) -> std::pin::Pin<Box<dyn std::future::Future<Output = ()> + Send>>)
> ```

As in other places, consider using [this](https://docs.rs/futures/latest/futures/future/type.BoxFuture.html) type instead.

pinkisemils (commented):
> Reviewed 14 of 69 files at r1, 2 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: 68 of 69 files reviewed, 10 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`talpid-core/src/tunnel_state_machine/mod.rs` line 238 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ffKco5ONMd1yhCiLJ:-N4lfwwu3M5Og-7cUgMf:bpu6tyl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/mod.rs#L238)):*

I'm not sure I agree. I think it becomes unnecessarily verbose since we repeat it several times. I'm curious what you think after reading above response.
</blockquote></details>

It's easier to not call it `init_args`, you can use something shorter like `args`. Then it won't be as cumbersome to not reassign variables. Otherwise, it seems like a lot of busywork to pack everything in a struct and then one function call deeper, the same fields get reassigned to values.

dlon (approved):
> Reviewed 1 of 69 files at r1, 1 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: all files reviewed, 9 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`talpid-core/src/tunnel_state_machine/connecting_state.rs` line 145 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4feb3tAaxiFDjAyVZ2:-N4lgSJc1oxtJLZs_sAb:bwzpk38) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/connecting_state.rs#L145)):*

The reasoning is that this was the intersection of parameters of all of the functions with too many arguments. It allows us to only create one type that we can reuse and pass down to the other functions. I was not sure how to group the parameters and this allowed us to only create 1 new type instead of 3.
</blockquote></details>

I definitely agree that there are too many parameters. Grouping them is a good idea.

I'm not sure about these three fields in particular. For example, why is `resource_dir` part of these args, but not `log_dir`? Why not the tunnel parameters?

But I also have no great suggestions.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: all files reviewed, 8 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[`mullvad-relay-selector/src/lib.rs` line 278 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4gCSmrA9SGO8Jkrwy9:-N4lh6Un-FGEeIYAzFrf:bvvmp4j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-relay-selector/src/lib.rs#L278)):*

No. Is the custom endpoint the larger variant? The problem is just that I doubt that the waste of memory is a problem that justifies boxing everything.
</blockquote></details>

I've made it so the clippy threshold for variance size is 1000 bytes instead of the default 200.
I've reversed the `SelectedRelay` boxing.
I reverted the boxing on `RequestCommand` since the variance isn't larger than 1000 bytes and most of the time its the large `NewRequest`.
Also reverted the boxing of `LastSelectedRelays` due to not crossing 1000 bytes threshold.
Same for `SelectedRelay`.

In `mullvad-api/src/proxy.rs` I've left `ApiConnection` boxed since the `Direct` variant is 512 bytes and the `Proxied` is 2976 bytes.
I figured this is a large enough difference and a bunch of memory is wasted for anyone not using shadowsocks.
___
*[`mullvad-setup/src/main.rs` line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ksH28EduWvAByWO0K:-N5-4oqi66tU8AJhUcJX:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/mullvad-setup/src/main.rs#L136)):*

I think this could be changed to an implementation of `FromStr` or just a function named `parse` instead of `parse_from_str` - it's pretty obvious it is parsing a borrowed string.
</blockquote></details>

Done.
___
*[`talpid-core/src/offline/linux.rs` line 28 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4lbDOwC5wn5CN4yzhk:-N5-5CXv1aTLH9LxmBpG:bsnzaza) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/offline/linux.rs#L28)):*

I don't believe this is necessary as I can't get my own clippy to complain about this with and without this.
</blockquote></details>

Interesting, I'll remove this. My clippy might be outdated.
___
*[`talpid-core/src/offline/mod.rs` line 35 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4lbKHOEGjVE6tLXFqb:-N5-5HSv0oDfP6HSE6Ze:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/offline/mod.rs#L35)):*

Same as the previous comment, I can't get clippy to complain about this function even without this exception.
</blockquote></details>

Done.
___
*[`talpid-core/src/routing/unix.rs` line 28 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4lbkQO3UaDOmD27s_G:-N5-5rGJ8U0XOHEYr2d-:b-9910kl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/routing/unix.rs#L28)):*

If there's an exception for Linux, shouldn't there be exceptions for all other modules?
</blockquote></details>

Done for Android and Macos. Windows should not need it.
___
*[`talpid-core/src/tunnel/openvpn/mod.rs` line 377 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ld42n8A32ZXMkuOOQ:-N5-AYwW4-aSupWF5BAw:bp1n58p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/tunnel/openvpn/mod.rs#L377)):*

This could implement `Default` for test builds just to simplify all of the code in the tests. Or have a test function that initializes just for the test module.
</blockquote></details>

Good suggestion, updated.
___
*[`talpid-core/src/tunnel/tun_provider/unix.rs` line 25 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ldHKJ3qCHgZmj2mRv:-N5-DKTq6BH4TPD0vdXX:bi5dwqf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1007442f00c9ee89950d7fb66cb9344bbf2534/talpid-core/src/tunnel/tun_provider/unix.rs#L25)):*

Can't this implementation be derived?
</blockquote></details>

It could, although it wouldn't call the `new` function and if we in the future want `new` to do something special then `default` may behave differently from `new`. As such I'm thinking maybe it's cleaner to just let `default` call `new`.
___
*[`talpid-core/src/tunnel_state_machine/connecting_state.rs` line 145 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4feb3tAaxiFDjAyVZ2:-N5-FnQr0uuSLZEZGagb:b487epp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/connecting_state.rs#L145)):*

I definitely agree that there are too many parameters. Grouping them is a good idea.

I'm not sure about these three fields in particular. For example, why is `resource_dir` part of these args, but not `log_dir`? Why not the tunnel parameters?

But I also have no great suggestions.
</blockquote></details>

The reason for that is because we create the `log_file` inside of `TunnelMonitor::start` and then pass that to the other init functions, if we want to reduce that parameter out of them we'll have to either split the type up or create `log_file` before we call `TunnelMonitor::start`.
Creating it before seems a bit weird because it is created through an associated method to the `TunnelMonitor`, and it seems nice to delegate all of that stuff to the `TunnelMonitor` itself. I believe this is the largest intersection of arguments between all the tunnel starting functions. So it's either this, have some `Option`s inside of the init arg struct or to split it up into several types.
___
*[`talpid-core/src/tunnel_state_machine/mod.rs` line 238 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4ffKco5ONMd1yhCiLJ:-N5-Hw5O-A5FzPYlHWz6:b-79a8tv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/mod.rs#L238)):*

It's easier to not call it `init_args`, you can use something shorter like `args`. Then it won't be as cumbersome to not reassign variables. Otherwise, it seems like a lot of busywork to pack everything in a struct and then one function call deeper, the same fields get reassigned to values.
</blockquote></details>

I've refactored it into `args` and used that directly like the original comment suggested.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`mullvad-relay-selector/src/lib.rs` line 278 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4gCSmrA9SGO8Jkrwy9:-N4kqKyU5xAzjIGVaUEB:brxp9sd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/mullvad-relay-selector/src/lib.rs#L278)):*

I've made it so the clippy threshold for variance size is 1000 bytes instead of the default 200.
I've reversed the `SelectedRelay` boxing.
I reverted the boxing on `RequestCommand` since the variance isn't larger than 1000 bytes and most of the time its the large `NewRequest`.
Also reverted the boxing of `LastSelectedRelays` due to not crossing 1000 bytes threshold.
Same for `SelectedRelay`.

In `mullvad-api/src/proxy.rs` I've left `ApiConnection` boxed since the `Direct` variant is 512 bytes and the `Proxied` is 2976 bytes.
I figured this is a large enough difference and a bunch of memory is wasted for anyone not using shadowsocks.


</blockquote></details>

Sounds good!
___
*[`talpid-core/src/tunnel_state_machine/connecting_state.rs` line 145 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667#-N4feb3tAaxiFDjAyVZ2:-N50LYJq8Qj_SVvAquXc:bqszfbv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a5c35ef0d01cae92c60aaab43e0233d7a441b0b7/talpid-core/src/tunnel_state_machine/connecting_state.rs#L145)):*

The reason for that is because we create the `log_file` inside of `TunnelMonitor::start` and then pass that to the other init functions, if we want to reduce that parameter out of them we'll have to either split the type up or create `log_file` before we call `TunnelMonitor::start`.
Creating it before seems a bit weird because it is created through an associated method to the `TunnelMonitor`, and it seems nice to delegate all of that stuff to the `TunnelMonitor` itself. I believe this is the largest intersection of arguments between all the tunnel starting functions. So it's either this, have some `Option`s inside of the init arg struct or to split it up into several types.
</blockquote></details>

I see! I think we should just go with this, but personally, I would split the type. I don't think that the public interface should be influenced too much by some internal detail. But let's not spend more time on this.

dlon (approved):
> Reviewed 40 of 42 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3667)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3670, opened 2022-06-16
Rework text field

pronebird:

Git checklist:

This PR reworks `AccountTextField` and leverages kerning attribute of `NSAttributedString` to add extra space between groups of digits. The benefit of this approach is that we can add visual space between groups of digits without affecting the input text itself.


AndreasShortcut (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670)* status: 0 of 6 files reviewed, 12 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN/AccountInputGroupView.swift` line 39 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4gdkYS4DlENWSY0CyI:-N4gdkYT8cOHtJm97oas:b-zb5ivk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountInputGroupView.swift#L39)):*
> ```Swift
> 
>     var textField: AccountTextField {
>         return privateTextField
> ```

I'm curious, what is the benefit of defining this as private and then exposing it as a computed property?
___
*[`ios/MullvadVPN/AccountTextField.swift` line 13 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4ge5A_7HEQr4_e9nBj:-N4ge5Aa48Mum4D-OoCf:bg2wfv3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L13)):*
> ```Swift
> class AccountTextField: CustomTextField, UITextFieldDelegate, UITextPasteDelegate {
>     /// The size of groups of digits.
>     static let groupSize = 4
> ```

If you call this digitGroupSize I think it's clear without the comment
___
*[`ios/MullvadVPN/AccountTextField.swift` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4geCGK7f4Sb4D9wNvt:-N4geCGK7f4Sb4D9wNvu:b-wv4ldd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L17)):*
> ```Swift
>     /// Spacing between groups in points.
>     /// Automatically updated using current font.
>     private var groupSpacing: CGFloat = 8
> ```

Same here if it's digitGroupSpacing or spacingBetweenDigitGroups
___
*[`ios/MullvadVPN/AccountTextField.swift` line 21 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4gkTnN2QzrEURUuE7s:-N4gkTnO4YAwaZIZf4Sa:b-h2v9v3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L21)):*
> ```Swift
>     /// Adjust caret by one whitespace when it's at the end of document, unless the given character
>     /// limit reached.
>     static let caretTrailingSpaceAtEndCharacterLimit = 16
> ```

I think it's a bit unclear what this constant is, and the comment is more addressing what it's used for.

I would suggest calling it something like `maxAccountNumberLength`, removing the comment and then using it in a function with a name that describes what it does, e.g.
```
func shouldAddGroupSpacing(for position: UITextPosition) -> Bool {
    return offset(from: beginningOfDocument, to: position) < Self.maxAccountNumberLength
}
```
Also see comment about `caretRect(for:)` further down
___
*[`ios/MullvadVPN/AccountTextField.swift` line 72 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4giEe25ubyk6rDXmqQ:-N4giEe25ubyk6rDXmqR:b2e5f6j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L72)):*
> ```Swift
>     }
> 
>     private class func isDigit(_ ch: Character) -> Bool {
> ```

I don't think you need this. There's already an `isNumber` computed property for Character, so you should be able to use e.g. `.filter { $0.isNumber }` or `.allSatisfy { $0.isNumber }` where needed
___
*[`ios/MullvadVPN/AccountTextField.swift` line 100 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4hBtDL96pyDnkgYVMo:-N4hBtDL96pyDnkgYVMp:b5fu1gu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L100)):*
> ```Swift
>     // MARK: - Input styling
> 
>     private func styleInput(_ string: String) -> NSAttributedString {
> ```

Maybe `spacedAttributedString(from:)` or something to better reflect the return type?
___
*[`ios/MullvadVPN/AccountTextField.swift` line 104 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4h9UEzAtduiUpfcpSA:-N4h9UEzAtduiUpfcpSB:b-9ijck5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L104)):*
> ```Swift
> 
>         for i in 0 ..< string.count {
>             if i > 0 && i % Self.groupSize == 0 {
> ```

Maybe we should wrap the `% Self.groupSize == 0` part in a function e.g. `isEndOfDigitGroup(at:)` or something?
___
*[`ios/MullvadVPN/AccountTextField.swift` line 125 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4hDx008vZm1ET6yTIq:-N4hDx008vZm1ET6yTIr:bsndrlb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L125)):*
> ```Swift
> 
>     func textPasteConfigurationSupporting(_ textPasteConfigurationSupporting: UITextPasteConfigurationSupporting, transform item: UITextPasteItem) {
>         _ = item.itemProvider.loadObject(ofClass: String.self) { str, error in
> ```

Could we call it `string`?
___
*[`ios/MullvadVPN/AccountTextField.swift` line 143 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4hER1M7lMBHMDPiGUa:-N4hER1M7lMBHMDPiGUb:b-nubulh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L143)):*
> ```Swift
>         NotificationCenter.default.post(name: UITextField.textDidChangeNotification, object: self)
> 
>         return self.textRange(from: endOfDocument, to: endOfDocument)!
> ```

Perhaps `?? UITextRange()` instead of force unwrapping?
___
*[`ios/MullvadVPN/AccountTextField.swift` line 167 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4h59jzCIqgNrVrsCBZ:-N4h59jzCIqgNrVrsCB_:buacvtq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L167)):*
> ```Swift
>             var spacing: CGFloat = .zero
> 
>             if position == endOfDocument {
> ```

This if /else part seems to not work properly if I move the caret to e.g. 123`|`4 5678 and type something.
This seems to solve the issue:
```
if position == endOfDocument {
    if shouldAddGroupSpacing(for: position) {
        spacing = groupSpacing
    }
} else if shouldAddGroupSpacing(for: position) {
    spacing = groupSpacing / 2
} else {
    spacing = -groupSpacing
}
```
where I've also used the suggested function from earlier comment
___
*[`ios/MullvadVPN/AccountTextField.swift` line 168 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4h4ZNV25ej35cLCNeS:-N4h4ZNV25ej35cLCNeT:b-fax6zp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L168)):*
> ```Swift
> 
>             if position == endOfDocument {
>                 let textLength = text?.count ?? 0
> ```

Since this part is only relevant if we're at the end of the document, .count should be the same as offset, so I think you could use that instead, getting rid of a constant that you have to unwrap.
___
*[`ios/MullvadVPN/AccountTextField.swift` line 174 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4h6i-6EoJX-BMAciVO:-N4h6i-75MQw-gQGRP4S:b-cqafrr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L174)):*
> ```Swift
>                 }
>             } else {
>                 spacing = -groupSpacing
> ```

I guess this is just a weird pet peeve of mine, and therefore completely optional, but I always find that negative signs look a bit weird in front of a variable, so I tend make an extension like
```
extension CGFloat {
    var negative: CGFloat { -self }
}
```
and call e.g. `spacing = groupSpacing.negative`

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670)* status: 0 of 6 files reviewed, 8 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/AccountTextField.swift` line 13 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4ge5A_7HEQr4_e9nBj:-N4knLuV5h0BDM52Sbqc:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L13)):*

If you call this digitGroupSize I think it's clear without the comment
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/AccountTextField.swift` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4geCGK7f4Sb4D9wNvt:-N4knMXVDCO6M_9YlZAN:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L17)):*

Same here if it's digitGroupSpacing or spacingBetweenDigitGroups
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/AccountTextField.swift` line 21 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4gkTnN2QzrEURUuE7s:-N4kmT9lAmVf02LdXU0I:bg9qlap) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L21)):*

I think it's a bit unclear what this constant is, and the comment is more addressing what it's used for.

I would suggest calling it something like `maxAccountNumberLength`, removing the comment and then using it in a function with a name that describes what it does, e.g.
```
func shouldAddGroupSpacing(for position: UITextPosition) -> Bool {
    return offset(from: beginningOfDocument, to: position) < Self.maxAccountNumberLength
}
```
Also see comment about `caretRect(for:)` further down
</blockquote></details>

Most of account numbers are probably 16 digits by now, so like when entering a serial number or any other number that consists of groups of digits or letters, an extra trailing space at the end hints that there is more to add.


So if you imagine that `|` is a caret, then you can visualise the caret adjustment as such:

```
xxxx |
xxxx xxxx |
xxxx xxxx xxxx |
xxxx xxxx xxxx xxxx|
```


You can still enter more than 16 digits and the groups will be properly formed, however the cursor won't be moved forward by one whitespace at the end of next group.

Adding a function does not add much either as this check is only used in `caretRect(for:)` and is a very simple check.
___
*[`ios/MullvadVPN/AccountTextField.swift` line 72 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4giEe25ubyk6rDXmqQ:-N4kowgeCXhlhyo5GJjb:b-tklo5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L72)):*
As per documentation for `Character.isNumber`, it returns `true` for the following variants of unicode letters:

> Discussion
> For example, the following characters all represent numbers:
> “7” (U+0037 DIGIT SEVEN)
> “⅚” (U+215A VULGAR FRACTION FIVE SIXTHS)
> “㊈” (U+3288 CIRCLED IDEOGRAPH NINE)
> “𝟠” (U+1D7E0 MATHEMATICAL DOUBLE-STRUCK DIGIT EIGHT)
> “๒” (U+0E52 THAI DIGIT TWO)

Our account numbers contain ASCII digits only AFAIK and I could perhaps combine it with `ch.isASCII && ch.isDigit` but I think it's much simpler to check against the specific ASCII range. Refactored into range check instead of switch-case.
___
*[`ios/MullvadVPN/AccountTextField.swift` line 104 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4h9UEzAtduiUpfcpSA:-N4kqejM2Jq4fSS70XLa:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L104)):*

Maybe we should wrap the `% Self.groupSize == 0` part in a function e.g. `isEndOfDigitGroup(at:)` or something?
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/AccountTextField.swift` line 125 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4hDx008vZm1ET6yTIq:-N4krO8S6xpXBirSDGka:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L125)):*

Could we call it `string`?
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/AccountTextField.swift` line 143 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4hER1M7lMBHMDPiGUa:-N4kzMBk7TrMJLuwLyoT:b-9012hk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L143)):*

Perhaps `?? UITextRange()` instead of force unwrapping?
</blockquote></details>

This call must never return `nil` because `endOfDocument` is guaranteed to be a valid text position.

I didn't actually know that `UITextRange()` can be instantiated directly and it looks like it can, but check out what `start` and `end` props contain when you do that:



Note that the definition for `UITextRange` states that `start` an `end` are not optional:

```swift
open class UITextRange : NSObject {
    open var isEmpty: Bool { get } //  Whether the range is zero-length.
    open var start: UITextPosition { get }
    open var end: UITextPosition { get }
}
```

So I think it's a necessary evil to force unwrap here :shrug:

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670)* status: 0 of 6 files reviewed, 6 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/AccountInputGroupView.swift` line 39 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4gdkYS4DlENWSY0CyI:-N4l604R4dQPNtYdH9fm:bwswuo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountInputGroupView.swift#L39)):*

I'm curious, what is the benefit of defining this as private and then exposing it as a computed property?
</blockquote></details>

Just a left over code. Renamed `privateTextField` to `textField` and made it public.
___
*[`ios/MullvadVPN/AccountTextField.swift` line 168 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4h4ZNV25ej35cLCNeS:-N4l3_WR5lMfsL6diZ9P:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L168)):*

Since this part is only relevant if we're at the end of the document, .count should be the same as offset, so I think you could use that instead, getting rid of a constant that you have to unwrap.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/AccountTextField.swift` line 174 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3670#-N4h6i-6EoJX-BMAciVO:-N4l3h0a1RhOLbYPs_2v:ba86e9d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9dbd77f2e59a8c3dca1d03f4355d725d79703ada/ios/MullvadVPN/AccountTextField.swift#L174)):*

I guess this is just a weird pet peeve of mine, and therefore completely optional, but I always find that negative signs look a bit weird in front of a variable, so I tend make an extension like
```
extension CGFloat {
    var negative: CGFloat { -self }
}
```
and call e.g. `spacing = groupSpacing.negative`
</blockquote></details>

There is `negate()` in standard library. However `-x` or `x * -1` is a basic arithmetics that pretty much everyone is familiar with. 

Overall I prefer the KISS principle in recent years. Less extensions we have is better for at least two reasons:

1. It's much easier to drag your code into playground without copying another 10 extensions that make it compile. 
2. But also it lowers the entry bar for new developers since they can swim (and not sink) as long as they know the standard library or base frameworks.

pronebird:
I am gonna leave that branch for now, but I am afraid that we have to abandon that effort since `UITextField` replaces the attributed string with `NSKern` attributes for individual letters  with a single paragraph style with kerning set for each letter, which breaks the visual output.

PR #3671, opened 2022-06-16
Add abstraction for atomic file I/O

dlon:
Replaces ad hoc code in a couple of places.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1667,7 +1667,6 @@ dependencies = [
  "lazy_static",
  "log",
  "mullvad-types",
- "rand 0.7.3",
  "regex",
  "rustls-pemfile",
  "serde",
@@ -1679,6 +1678,7 @@ dependencies = [
  "tokio-rustls",
  "tokio-stream",
  "urlencoding",
+ "uuid",
  "webpki 0.21.4",
 ]
 
--- a/mullvad-api/Cargo.toml
+++ b/mullvad-api/Cargo.toml
@@ -19,7 +19,6 @@ http = "0.2"
 hyper = { version = "0.14", features = ["client", "stream"] }
 ipnetwork = "0.16"
 log = "0.4"
-rand = "0.7"
 regex = "1"
 serde = "1"
 serde_json = "1.0"
@@ -36,6 +35,8 @@ talpid-types = { path = "../talpid-types" }
 talpid-time = { path = "../talpid-time" }
 
 shadowsocks = { version = "1.14.2", default-features = false, features = ["stream-cipher"] }
+
+uuid = { version = "0.8", features = ["v4"] }
 
 [target.'cfg(target_os="macos")'.dependencies]
 tokio-stream = { version = "0.1", features = ["io-util"] }
--- a/mullvad-api/src/address_cache.rs
+++ b/mullvad-api/src/address_cache.rs
@@ -83,15 +83,11 @@ impl AddressCache {
             None => return Ok(()),
         };
 
-        let temp_path = write_path.with_file_name("api-cache.temp");
-
-        let mut file = fs::File::create(&temp_path).await?;
+        let mut file = crate::fs::AtomicFile::new().await?;
         let mut contents = address.to_string();
         contents += "\n";
         file.write_all(contents.as_bytes()).await?;
-        file.sync_data().await?;
-
-        fs::rename(&temp_path, write_path).await
+        file.finalize(write_path).await
     }
 }
 
--- a/mullvad-api/src/fs.rs
+++ b/mullvad-api/src/fs.rs
@@ -1,0 +1,61 @@
+use std::{env, path::{PathBuf, Path}, ops::{Deref, DerefMut}};
+use talpid_types::ErrorExt;
+use tokio::{fs, io};
+
+/// Stores content in a temporary file before moving it to the
+/// final destination, ensuring that consumers of the file never
+/// end up with partial content. Must be moved with `finalize`.
+pub struct AtomicFile {
+    file: Option<fs::File>,
+    path: PathBuf,
+}
+
+impl AtomicFile {
+    pub async fn new() -> io::Result<Self> {
+        let path = temp_path();
+        Ok(Self {
+            file: Some(fs::File::create(&path).await?),
+            path,
+        })
+    }
+
+    /// Flushes and moves the file to `target_path`, replacing it if it exists.
+    pub async fn finalize(mut self, target_path: &Path) -> io::Result<()> {
+        let file = self.file.take().unwrap();
+
+        file.sync_all().await?;
+        let std_file = file.into_std().await;
+        let _ = tokio::task::spawn_blocking(move || drop(std_file)).await;
+
+        fs::rename(&self.path, target_path).await
+    }
+}
+
+impl Drop for AtomicFile {
+    fn drop(&mut self) {
+        // The file will be removed when all file handles are closed
+        if let Err(error) = std::fs::remove_file(&self.path) {
+            if error.kind() != io::ErrorKind::NotFound {
+                log::warn!("{}", error.display_chain_with_msg("Failed to delete AtomicFile"));
+            }
+        }
+    }
+}
+
+impl Deref for AtomicFile {
+    type Target = fs::File;
+
+    fn deref(&self) -> &Self::Target {
+        self.file.as_ref().unwrap()
+    }
+}
+
+impl DerefMut for AtomicFile {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.file.as_mut().unwrap()
+    }
+}
+
+fn temp_path() -> PathBuf {
+    env::temp_dir().join(uuid::Uuid::new_v4().to_string())
+}
--- a/mullvad-api/src/lib.rs
+++ b/mullvad-api/src/lib.rs
@@ -32,6 +32,7 @@ pub use crate::https_client_with_sni::SocketBypassRequest;
 mod access;
 mod address_cache;
 pub mod device;
+mod fs;
 mod relay_list;
 pub use address_cache::AddressCache;
 pub use device::DevicesProxy;
--- a/mullvad-api/src/proxy.rs
+++ b/mullvad-api/src/proxy.rs
@@ -1,7 +1,6 @@
 use crate::tls_stream::TlsStream;
 use futures::Stream;
 use hyper::client::connect::{Connected, Connection};
-use rand::{distributions::Alphanumeric, Rng};
 use serde::{Deserialize, Serialize};
 use shadowsocks::relay::tcprelay::ProxyClientStream;
 use std::{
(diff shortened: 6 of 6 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-api/src/fs.rs` line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4lNG450GzWmMAvVchu:-N4lNG450GzWmMAvVchv:bvaskaf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/413b5c2bfbd9454bd5d4cef8cce30401f95d9852/mullvad-api/src/fs.rs#L42)):*
> ```XML
>     fn drop(&mut self) {
>         // The file will be removed when all file handles are closed
>         if let Err(error) = std::fs::remove_file(&self.temp_path) {
> ```

I think this should be guarded for and not happen on a properly finalized file. Otherwise in the happy path we'll be trying to remove non existent paths all the time.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @faern and @pinkisemils)

___
*[`mullvad-api/src/fs.rs` line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4lNG450GzWmMAvVchu:-N4lVERh-qCWRjdKribO:bfspgo6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/413b5c2bfbd9454bd5d4cef8cce30401f95d9852/mullvad-api/src/fs.rs#L42)):*

I think this should be guarded for and not happen on a properly finalized file. Otherwise in the happy path we'll be trying to remove non existent paths all the time.
</blockquote></details>

Done. This was the least terrible solution I came up with.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-api/src/fs.rs` line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4lNG450GzWmMAvVchu:-N4lqZYK4F49eX5vhJsB:bx63hbt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/413b5c2bfbd9454bd5d4cef8cce30401f95d9852/mullvad-api/src/fs.rs#L42)):*

Done. This was the least terrible solution I came up with.
</blockquote></details>

Instead of this dance, why don't you do the equivalent of `finalize_inner` in the descructor? That way a consumer does not need to finalize it. They just need to do that if they *care about the result of finalizing*.

Since the drop is not blocking, I'd make it `finalize_inner(&mut self)` blocking and then have the `async finalize` just call the entire method in a `spawn_blocking`.

I think that would be an even less terrible solution.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-api/src/fs.rs` line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4lNG450GzWmMAvVchu:-N4lrZlG33srjMoocU8V:br2q8qu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/413b5c2bfbd9454bd5d4cef8cce30401f95d9852/mullvad-api/src/fs.rs#L42)):*

Instead of this dance, why don't you do the equivalent of `finalize_inner` in the descructor? That way a consumer does not need to finalize it. They just need to do that if they *care about the result of finalizing*.

Since the drop is not blocking, I'd make it `finalize_inner(&mut self)` blocking and then have the `async finalize` just call the entire method in a `spawn_blocking`.

I think that would be an even less terrible solution.
</blockquote></details>

I'd argue it would be the least surprising API. Since just dropping normal files makes sure they are properly closed and written to disk.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`mullvad-api/src/fs.rs` line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4lNG450GzWmMAvVchu:-N4ltbxMEOBqrKDlrnUz:b-du92ts) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/413b5c2bfbd9454bd5d4cef8cce30401f95d9852/mullvad-api/src/fs.rs#L42)):*

I'd argue it would be the least surprising API. Since just dropping normal files makes sure they are properly closed and written to disk.
</blockquote></details>

Do you mean that `drop` should just call `finalize_inner`? If so, then finalize would still need to be called in an async context, right? Not doing so would be wrong.

Or do you mean that the destructor should call `spawn_blocking` from a tokio handle as well? In that case, you wouldn't be able to rely on the drop to immediately flush everything to disk, which could also be unexpected.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`mullvad-api/src/fs.rs` line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4lNG450GzWmMAvVchu:-N4lyzaI7GvFEvkAsZwl:b-brqvtk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/413b5c2bfbd9454bd5d4cef8cce30401f95d9852/mullvad-api/src/fs.rs#L42)):*

Do you mean that `drop` should just call `finalize_inner`? If so, then finalize would still need to be called in an async context, right? Not doing so would be wrong.

Or do you mean that the destructor should call `spawn_blocking` from a tokio handle as well? In that case, you wouldn't be able to rely on the drop to immediately flush everything to disk, which could also be unexpected.
</blockquote></details>

I was not aware that `into_std` was async/required async IO :O

Yeah OK. Then the only way to have drop finalize the file would be if the drop impl entered the tokio context and performed some async IO. This would not be impossible, but it's probably overkill for achieving this API behavior.

Async drop, when? :D

I guess then, this is the best impl we can have for now :+1:
___
*[`mullvad-api/src/fs.rs` line 52 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4m-FN_Aoje_31NIQdc:-N4m-FN_Aoje_31NIQdd:b-umlrp6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7894f2178a042b12b5d5c562c3f9d50ee622a6b7/mullvad-api/src/fs.rs#L52)):*
> ```XML
>     fn drop(&mut self) {
>         if self.file.is_some() {
>             log::warn!("Object was not finalized");
> ```

"Object" is extremely generic. I think the best thing would be to print the `self.target_path` here. If not, we should at least call it a file, not just an object.
___
*[`mullvad-api/src/fs.rs` line 63 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4m-Z5CCt3-ML9V03Tv:-N4m-Z5CCt3-ML9V03Tw:b-vzs3nh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7894f2178a042b12b5d5c562c3f9d50ee622a6b7/mullvad-api/src/fs.rs#L63)):*
> ```XML
>             log::warn!(
>                 "{}",
>                 error.display_chain_with_msg("Failed to delete temp file")
> ```

This should probably also contain the `self.temp_path` that we failed to delete. Because the inner error won't. IO errors don't describe the path, that's the responsibility of the caller. Our caller, meaning the user of `AtomicFile` can't know the temp path, nor print it, because that's a hidden property inside the struct. So I think we are responsible for printing it if we are leaving litter on the disk :)

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[`mullvad-api/src/fs.rs` line 40 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4m04X8-b3NzuHNVR2J:-N4m04X8-b3NzuHNVR2K:b-hv7cyq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7894f2178a042b12b5d5c562c3f9d50ee622a6b7/mullvad-api/src/fs.rs#L40)):*
> ```XML
>         file: fs::File,
>         temp_path: &Path,
>         target_path: &Path,
> ```

Why doesn't this just take `&mut self` and do the `self.file.take()` in here? It's not a public method, and we can take responsiblity of calling it only once.

faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-:-N5-5Xiy8m7O_eslLYm8:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-api/src/fs.rs` line 52 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4m-FN_Aoje_31NIQdc:-N5-3jnKAf_TkIBdOF-C:bu40sz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7894f2178a042b12b5d5c562c3f9d50ee622a6b7/mullvad-api/src/fs.rs#L52)):*

"Object" is extremely generic. I think the best thing would be to print the `self.target_path` here. If not, we should at least call it a file, not just an object.
</blockquote></details>

What's the motivation behind printing `temp_path` rather than `target_path`? If a user of this module forgets to call `finalize` they will probably get more help out of seeing which file they forgot to finalize rather than a random UUID that they did not generate themselves.

I will mark the thread as non-blocking since it's a bit nitpicky. But I do think the proper path to print here is `target_path`.
___
*[`mullvad-api/src/fs.rs` line 55 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N5-4cE--eNYCqmFQNK5:-N5-4cE--eNYCqmFQNK6:b4du9tn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b4f0b6a889913378a6e1d5e19d34488795e0cc05/mullvad-api/src/fs.rs#L55)):*
> ```XML
> fn try_remove_file(temp_path: &Path) {
>     if let Err(error) = std::fs::remove_file(temp_path) {
>         if error.kind() != io::ErrorKind::NotFound {
> ```

Not a blocker for the PR, but I want to note this: It's probably a bug/issue if this code hits `NotFound` also. We never remove the file twice, and no other code/process should touch our temporary file. So I'd say we could equally well WARN-log on `NotFound` also, as it indicates something is probably wrong in the system.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`mullvad-api/src/fs.rs` line 52 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4m-FN_Aoje_31NIQdc:-N5-9tBdB02Jx263zAXF:b-h0wxhk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7894f2178a042b12b5d5c562c3f9d50ee622a6b7/mullvad-api/src/fs.rs#L52)):*

What's the motivation behind printing `temp_path` rather than `target_path`? If a user of this module forgets to call `finalize` they will probably get more help out of seeing which file they forgot to finalize rather than a random UUID that they did not generate themselves.

I will mark the thread as non-blocking since it's a bit nitpicky. But I do think the proper path to print here is `target_path`.
</blockquote></details>

Done. That's fair. The temp path will be printed anyway, so that would give them more information.

I do wonder if this message should be removed altogether. There is an expected situation in which it can occur:

```
let f = AtomicFile::new(path);
f.write_all(b"test").await?; // if this fails, maybe no error needs to be logged?
f.finalize().await?;
```
___
*[`mullvad-api/src/fs.rs` line 55 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N5-4cE--eNYCqmFQNK5:-N5-91L-5cQIEK13ltDo:b2c9w9f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b4f0b6a889913378a6e1d5e19d34488795e0cc05/mullvad-api/src/fs.rs#L55)):*

Not a blocker for the PR, but I want to note this: It's probably a bug/issue if this code hits `NotFound` also. We never remove the file twice, and no other code/process should touch our temporary file. So I'd say we could equally well WARN-log on `NotFound` also, as it indicates something is probably wrong in the system.
</blockquote></details>

Done. This is true.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671)* status: 5 of 6 files reviewed, all discussions resolved (waiting on @faern)

___
*[`mullvad-api/src/fs.rs` line 52 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3671#-N4m-FN_Aoje_31NIQdc:-N5-CyE52BCJgXsnFv62:b-gpv4i1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7894f2178a042b12b5d5c562c3f9d50ee622a6b7/mullvad-api/src/fs.rs#L52)):*

Done. That's fair. The temp path will be printed anyway, so that would give them more information.

I do wonder if this message should be removed altogether. There is an expected situation in which it can occur:

```
let f = AtomicFile::new(path);
f.write_all(b"test").await?; // if this fails, maybe no error needs to be logged?
f.finalize().await?;
```
</blockquote></details>

But I will leave it in because I think it's valuable for catching bugs.

PR #3678, opened 2022-06-21
Remove legacy upstart init configuration

evan-a-a:
The supported platforms (Ubuntu 18.04+, Debian 10+, and Fedora 33+) all use systemd, so the old upstart configuration can be safely removed. This change cleans up the post-install and packaging code to remove all references to the upstart configurations.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3678)


faern (approved):
This makes sense to me. There are always weird distributions using weird old software. But according to Wikipedia, no one is really using Upstart any longer. So I'm fine with completely removing it.

faern:
Thanks for the contribution.

PR #3680, opened 2022-06-21
Add CI script for running Clippy

Jontified:
Add a CI script that runs Clippy and clean up the last of the Clippy warnings

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -1,0 +1,24 @@
+name: Rust - Run Clippy to check lints
+on:
+    # Check whenever a file that affects Rust formatting is changed in a pull request
+    pull_request:
+        paths:
+            - .github/workflows/clippy.yml
+            - clippy.toml
+            - '**/*.rs'
+    # Check if requested manually from the Actions tab
+    workflow_dispatch:
+jobs:
+  clippy_check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1.0.6
+        with:
+            toolchain: nightly
+            components: clippy
+            override: true
+      - uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --all-features


dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: all files reviewed, 2 unresolved discussions (waiting on @Jontified)

___
*[`.github/workflows/clippy.yml` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5A_DEW3Lbh0u5BWXhY:-N5A_DEW3Lbh0u5BWXhZ:b-2knwoc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9477dc06bb40ecd53d7d7993814466cde28c8a1/.github/workflows/clippy.yml#L20)):*
> ```YAML
> 
>       - name: Checkout binaries submodule
>         run: git submodule update --init --depth=1 dist-assets/binaries
> ```

I don't think it needs to grab the binaries for this.
___
*[`.github/workflows/clippy.yml` line 31 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5A_5cK3IBgPO57hTNc:-N5A_5cK3IBgPO57hTNd:bc0px52) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9477dc06bb40ecd53d7d7993814466cde28c8a1/.github/workflows/clippy.yml#L31)):*
> ```YAML
>         run: |
>           sudo apt-get update
>           sudo apt-get install libdbus-1-dev
> ```

It might work without this as well.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`.github/workflows/clippy.yml` line 31 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5A_5cK3IBgPO57hTNc:-N5Ebkf5-NLSTHXL54QX:b-dxn4dp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9477dc06bb40ecd53d7d7993814466cde28c8a1/.github/workflows/clippy.yml#L31)):*

It might work without this as well.
</blockquote></details>

Scratch that. `libdbus-rs` definitely needs this.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`.github/workflows/clippy.yml` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5A_DEW3Lbh0u5BWXhY:-N5EfFQd48xU1n1WUgw6:bfcwu0n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9477dc06bb40ecd53d7d7993814466cde28c8a1/.github/workflows/clippy.yml#L20)):*

I don't think it needs to grab the binaries for this.
</blockquote></details>

Not grabbing binaries does not allow compilation.
___
*[`.github/workflows/clippy.yml` line 31 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5A_5cK3IBgPO57hTNc:-N5EfQUr3UxU6hcQmD_R:bnxf80r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9477dc06bb40ecd53d7d7993814466cde28c8a1/.github/workflows/clippy.yml#L31)):*

Scratch that. `libdbus-rs` definitely needs this.
</blockquote></details>

Yeah, should be necessary.

dlon (approved):
The only complaint I have is that Clippy still fails on other targets/platforms, e.g. Windows. But I'm not sure we should to run it for all OSes in this workflow.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-:-N5EiwV02uw5A_pDRO8u:b-sfdjr6)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`.github/workflows/clippy.yml` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5A_DEW3Lbh0u5BWXhY:-N5Eh5eF-yVXqN1FauSL:byxe63g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9477dc06bb40ecd53d7d7993814466cde28c8a1/.github/workflows/clippy.yml#L20)):*

Not grabbing binaries does not allow compilation.
</blockquote></details>

You're absolutely right. It needs `libnftnl` and `libmnl`.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`.github/workflows/clippy.yml` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5A_DEW3Lbh0u5BWXhY:-N5EkxMMEZwY4cDmTwp4:b70u06u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9477dc06bb40ecd53d7d7993814466cde28c8a1/.github/workflows/clippy.yml#L20)):*

You're absolutely right. It needs `libnftnl` and `libmnl`.
</blockquote></details>

If that's all that's needed, maybe you can install those libraries from `apt` and save some time? I don't think it will make any difference if we use the shared libraries from `apt` or our static libraries for the sake of just running clippy?

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`.github/workflows/clippy.yml` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5A_DEW3Lbh0u5BWXhY:-N5FFzsK8RTb4BsiHgpp:b-rfsgxv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9477dc06bb40ecd53d7d7993814466cde28c8a1/.github/workflows/clippy.yml#L20)):*

If that's all that's needed, maybe you can install those libraries from `apt` and save some time? I don't think it will make any difference if we use the shared libraries from `apt` or our static libraries for the sake of just running clippy?
</blockquote></details>

It seems to only take a second to clone/checkout that repo, so I think this is fine as well.

Jontified (commented):
I am not sure either. The downside is that we then run (1-3) more jobs every time we push while most of the code they check is the same. The upside is that we have guarantees of no clippy warnings for those other platforms. I imagine that you're likely to fix the clippy warnings since you'll be forced to fix the linux-related ones and will then notice the non-linux ones. However that's not a guarantee.
I don't think it's a very difficult thing to add however.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-:-N5aDDw0EB2TPtI5lXi6:b-om6632)* status: 10 of 13 files reviewed, all discussions resolved (waiting on @dlon)

faern (changes requested):
> Reviewed 2 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: 10 of 13 files reviewed, 2 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`.github/workflows/clippy.yml` line 35 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5dfe_e6krqHmJOipkv:-N5dfe_e6krqHmJOipkw:bs09m9a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f255ce46560f577053abbbbbdeebbc6c2959e77/.github/workflows/clippy.yml#L35)):*
> ```YAML
>       - name: Clippy check
>         run: |
>           ./ci/check-clippy.sh
> ```

Does this really need to be a separate file? The script only has three commands in it. As far as I can see, this is the only invocation. So can't you list those commands directly here in the yml?
___
*[`ci/check-clippy.sh` line 5 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5deiyS1GfTuR5FZhBz:-N5deiyT94Np9MxfH6EG:b-ops8ri) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f255ce46560f577053abbbbbdeebbc6c2959e77/ci/check-clippy.sh#L5)):*
> ```Shell
> set -eux
> 
> export RUSTFLAGS="--deny warnings"
> ```

This is OK for now. Let's not feature creep this PR. But I don't like that we keep repeating this in more files.

An idea: Move `RUSTFLAGS` initialization to `env.sh` but only set it to `--deny warnings` if `--ci` is passed to `env.sh`? That way developers don't need to change anything in their workflow, and all various CI scripts can initialize their environment with just `source env.sh --ci`. But let's save that for another PR :)

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: 8 of 13 files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

___
*[`.github/workflows/clippy.yml` line 35 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5dfe_e6krqHmJOipkv:-N5esYGi2_S-Ku8FI63j:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f255ce46560f577053abbbbbdeebbc6c2959e77/.github/workflows/clippy.yml#L35)):*

Does this really need to be a separate file? The script only has three commands in it. As far as I can see, this is the only invocation. So can't you list those commands directly here in the yml?
</blockquote></details>

Done.

faern (changes requested):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: 10 of 13 files reviewed, 1 unresolved discussion (waiting on @dlon and @Jontified)

___
*[`talpid-core/src/tunnel/wireguard/wireguard_kernel/wg_message.rs` line 729 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5ip90d3O7o2nasc7Uj:-N5ip90d3O7o2nasc7Uk:b-t6qkiv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/afb4f51c26c708e4fdcbc31f1988096b9a7bea49/talpid-core/src/tunnel/wireguard/wireguard_kernel/wg_message.rs#L729)):*
> ```XML
>                 AllowedIps(
>                     [AllowedIpMessage(
>                         [
> ```

Please don't do the reformatting in this PR. You already fixed it on `master`. So just remove this reformatting commit and rebase please.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: 10 of 13 files reviewed, 2 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`talpid-core/src/offline/linux.rs` line 28 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5iqTHt9tQmNdwaU2yN:-N5iqTHt9tQmNdwaU2yO:b4qc4j4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/afb4f51c26c708e4fdcbc31f1988096b9a7bea49/talpid-core/src/offline/linux.rs#L28)):*
> ```XML
> 
> impl MonitorHandle {
>     pub async fn host_is_offline(&mut self) -> bool {
> ```

I think you changed this because clippy said an `is_` method should not take `&mut`, right? Does it need `&mut` then? It does not look like it needs that. Reducing mutability sounds like a good thing if it's not needed. I feel like the compiler or clippy should be able to detect unused mutability and warn :thinking:

Anyway, I'm not saying the new name is bad. I'm just saying there is probably another way to also avoid the clippy warning.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: 10 of 13 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[`talpid-core/src/offline/linux.rs` line 28 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5iqTHt9tQmNdwaU2yN:-N5jORr395jjNZksrqXR:b-bp1kxo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/afb4f51c26c708e4fdcbc31f1988096b9a7bea49/talpid-core/src/offline/linux.rs#L28)):*

I think you changed this because clippy said an `is_` method should not take `&mut`, right? Does it need `&mut` then? It does not look like it needs that. Reducing mutability sounds like a good thing if it's not needed. I feel like the compiler or clippy should be able to detect unused mutability and warn :thinking:

Anyway, I'm not saying the new name is bad. I'm just saying there is probably another way to also avoid the clippy warning.
</blockquote></details>

I removed the `&mut` but kept the name `host_is_offline`. I think the reason there isn't a warning for this is that the compiler and clippy think we want to keep the signature mutable in order to potentially add mutable behavior later.
___
*[`talpid-core/src/tunnel/wireguard/wireguard_kernel/wg_message.rs` line 729 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5ip90d3O7o2nasc7Uj:-N5jQ3P57B4kOs4gUyYS:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/afb4f51c26c708e4fdcbc31f1988096b9a7bea49/talpid-core/src/tunnel/wireguard/wireguard_kernel/wg_message.rs#L729)):*

Please don't do the reformatting in this PR. You already fixed it on `master`. So just remove this reformatting commit and rebase please.
</blockquote></details>

Done.

faern (approved):
> Reviewed 2 of 3 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`talpid-core/src/tunnel_state_machine/connecting_state.rs` line 154 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5jfrKtBAQZ4APuJePL:-N5jfrKtBAQZ4APuJePM:bs5m3vy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4e0b9d489b93ff82cda28527c18fe16320038c20/talpid-core/src/tunnel_state_machine/connecting_state.rs#L154)):*
> ```XML
>                 tun_provider,
>                 retry_attempt,
>                 route_manager: route_manager_handle,
> ```

I feel like this move of arguments is just a hack to please Clippy. The code itself did not really get better. We just hid the arguments somewhere else so that specific lint won't trigger. I'm not sure what the original reason for having `TunnelArgs` were. Maybe it was the start of the same type of hack? Anway, going to mark this as non-blocking. Just want to flag that I don't really think it solves what Clippy wanted us to solve :)

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`talpid-core/src/tunnel_state_machine/connecting_state.rs` line 154 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5jfrKtBAQZ4APuJePL:-N5jjqruBPSiHKquIHAM:b3mfhqj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4e0b9d489b93ff82cda28527c18fe16320038c20/talpid-core/src/tunnel_state_machine/connecting_state.rs#L154)):*

I feel like this move of arguments is just a hack to please Clippy. The code itself did not really get better. We just hid the arguments somewhere else so that specific lint won't trigger. I'm not sure what the original reason for having `TunnelArgs` were. Maybe it was the start of the same type of hack? Anway, going to mark this as non-blocking. Just want to flag that I don't really think it solves what Clippy wanted us to solve :)
</blockquote></details>

`TunnelArgs` was created to please clippy in one of the previous clippy PRs.
I would argue that this solution solves at least one problem that clippy complains about which is that now arguments are named.
You no longer have to remember what the 6th argument meant, instead you have a name for each argument.

Arguably another problem is that there are too many arguments for the function, but this function seemingly needs all of these arguments so to reduce the amount is not necessarily possible or better.

One could bundle the arguments that one thinks fit logically together in a new struct and reduce the amount of arguments that way. However you'd be creating several structs that are *only* used here. I would argue that this would be unnecessarily complex and that this should only be done if you can bundle the arguments in such a way that the new structs finds use in another part of the code.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680)* status: 13 of 14 files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`talpid-core/build.rs` line 74 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5k8XWN4H1O-VlRa1B_:-N5k8XWN4H1O-VlRa1Ba:bfff3hj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b4514b6c5a9d6fe44733caac153f35275a3bd0b5/talpid-core/build.rs#L74)):*
> ```XML
>         // If we cross-compile for windows in order to run clippy we don't want to panic.
>         // In those cases we should set an env variable called clippy.
>         "windows" if env::var("CLIPPY").is_ok() => "",
> ```

Will we set this env var, or does Clippy set it automatically whet `cargo clippy` is ran? If it's something we are going to set, maybe we should consider another name and/or just happily fall through without a variable? I have wanted to do `cargo check --target=x86_64-pc-windows-gnu` many times to just see if Windows specific code I'm writing on Windows typechecks. So I see the use for this even outside of Clippy/CI.

Jontified (commented):
Upon further reflection I believe that this is a useful feature but that it should be done in another PR. There is a significant number of clippy warnings on the windows build and I think it's better to fix those before releasing a CI clippy build that will just fail. I don't think it's a good idea to block this PR on fixing all of the Windows clippy warnings.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-:-N5naoRJBY_akXR2OA4t:b-rlas2w)* status: 13 of 14 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`talpid-core/build.rs` line 74 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-N5k8XWN4H1O-VlRa1B_:-N5naad_8AuFV7NV4bop:b-s7xbhf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b4514b6c5a9d6fe44733caac153f35275a3bd0b5/talpid-core/build.rs#L74)):*

Will we set this env var, or does Clippy set it automatically whet `cargo clippy` is ran? If it's something we are going to set, maybe we should consider another name and/or just happily fall through without a variable? I have wanted to do `cargo check --target=x86_64-pc-windows-gnu` many times to just see if Windows specific code I'm writing on Windows typechecks. So I see the use for this even outside of Clippy/CI.
</blockquote></details>

Alright, I will update this to just fallthrough without an env-variable set and leave a comment that explains how this does not allow building but is necessary for checking.

faern (approved):
I agree. The Windows clippy warnings should be looked at. But this PR is not the place. Let's merge this and have a separate card for Windows and macOS specific linting fixes.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3680#-:-N5oQgwG52leoXH1-S_w:brdpo3r)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3696, opened 2022-06-23
Only check multihop state when enabling PQ

dlon:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3696)

### DIFF
--- a/mullvad-cli/src/cmds/relay.rs
+++ b/mullvad-cli/src/cmds/relay.rs
@@ -575,7 +575,7 @@ impl Relay {
             wireguard_constraints.entry_location = parse_entry_location_constraint(entry);
             let use_multihop = wireguard_constraints.entry_location.is_some();
             if use_multihop {
-                let use_pq = rpc
+                let use_pq_safe_psk = rpc
                     .get_settings(())
                     .await?
                     .into_inner()
@@ -584,9 +584,9 @@ impl Relay {
                     .wireguard
                     .unwrap()
                     .use_pq_safe_psk;
-                if use_pq {
+                if use_pq_safe_psk {
                     return Err(Error::CommandFailed(
-                        "PQ PSK exchange does not work when multihop is enabled",
+                        "Quantum resistant tunnels do not work when multihop is enabled",
                     ));
                 }
             }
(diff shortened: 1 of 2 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3696)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @faern)

___
*[`mullvad-cli/src/cmds/relay.rs` line 587 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3696#-N5G1Nxt9k3p0ej9JZQb:-N5G4QhK9Tqz5d_ngZex:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b0b93961cb3ec98751565a831dafefd51ebbe601/mullvad-cli/src/cmds/relay.rs#L587)):*

Can't this variable be named `use_pq_safe_psk` like the valiable it sources its value from? Just "Use PQ" means very little since just "post quantum" on its own means very little. One has to use it in combination with "safety" or "cryptography" etc for it to really mean much.
</blockquote></details>

Done.
___
*[`mullvad-cli/src/cmds/tunnel.rs` line 239 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3696#-N5G0ryz1LVG4jdNZwZz:-N5G4TQyB2lXz365VdU0:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b0b93961cb3ec98751565a831dafefd51ebbe601/mullvad-cli/src/cmds/tunnel.rs#L239)):*

IMO `new_state` is a strange/too generic name for this. I would expect it to contain a state. When the state is just a `bool` it becomes hard to know what is what. `if new_state` is hard to know what it is. I see it's called `use_multihop` and `use_pq_safe_psk` in similar places. Can we try to use one of those and both be more consistent and more explicit about what it is?
</blockquote></details>

Done.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3696)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3703, opened 2022-06-28
Windows auto set mtu

Jontified:
This PR makes the Windows daemon look up the MTU of the default interface and use that as the default in the Wireguard tunnel.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3703)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ Line wrap the file at 100 chars.                                              Th
 #### Android
 - Add device management to the Android app. This simplifies knowing which device is which and adds
   the option to log other devices out when the account already has five devices.
+#### Windows
+- Windows daemon now looks up the MTU on the default interface and uses this MTU instead of the
+  default 1500. The 1500 is still the fallback if this for some reason fails. This may stop
+  fragmentation.
 
 ### Changed
 - Reject invalid WireGuard ports in the CLI.
--- a/talpid-core/src/routing/windows.rs
+++ b/talpid-core/src/routing/windows.rs
@@ -7,7 +7,8 @@ use futures::{
     },
     StreamExt,
 };
-use std::collections::HashSet;
+use std::{collections::HashSet, net::IpAddr};
+use winnet::WinNetAddrFamily;
 
 /// Windows routing errors.
 #[derive(err_derive::Error, Debug)]
@@ -30,6 +31,9 @@ pub enum Error {
     /// Attempt to use route manager that has been dropped
     #[error(display = "Cannot send message to route manager since it is down")]
     RouteManagerDown,
+    /// Something went wrong when getting the mtu of the interface
+    #[error(display = "Could not get the mtu of the interface")]
+    GetMtu,
 }
 
 pub type Result<T> = std::result::Result<T, Error>;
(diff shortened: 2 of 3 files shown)


dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3703)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`talpid-core/src/routing/windows.rs` line 129 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3703#-N67mg6ZBhJysb6YOHe2:-N67mg6ZBhJysb6YOHe3:b-dfsdgw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ae0af9e4a44c25472193acebbbdc29bd06b88fe/talpid-core/src/routing/windows.rs#L129)):*
> ```XML
>                     );
>                 }
>                 RouteManagerCommand::GetMtuForRoute(ip, tx) => {
> ```

This doesn't need to be in the route manager at all. Is it here because we want to have the same interface on Linux and Windows? That does make sense to me.
___
*[`talpid-core/src/routing/windows.rs` line 201 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3703#-N67qEa5-62YsOcYqFdR:-N67qEa5-62YsOcYqFdS:b-xm244r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ae0af9e4a44c25472193acebbbdc29bd06b88fe/talpid-core/src/routing/windows.rs#L201)):*
> ```XML
>             };
>             let interface_row = crate::windows::get_ip_interface_entry(addr_family, &luid)
>                 .map_err(|_| Error::GetMtu)?;
> ```

I think either the io error should be logged here, or it should be the source. Currently, it is impossible to figure out why `get_mtu_for_route` fails based on the logs.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3703)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`talpid-core/src/routing/windows.rs` line 129 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3703#-N67mg6ZBhJysb6YOHe2:-N6TOnBg4mjWU43clpvv:b-f3ed8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ae0af9e4a44c25472193acebbbdc29bd06b88fe/talpid-core/src/routing/windows.rs#L129)):*

This doesn't need to be in the route manager at all. Is it here because we want to have the same interface on Linux and Windows? That does make sense to me.
</blockquote></details>

It is because we want to have the same interface for both Linux and windows. I'm not sure if that's the right decision but I'll keep it like this unless you disagree.
___
*[`talpid-core/src/routing/windows.rs` line 201 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3703#-N67qEa5-62YsOcYqFdR:-N6TRQyZ3NIgQIOXixrC:b-dkciup) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ae0af9e4a44c25472193acebbbdc29bd06b88fe/talpid-core/src/routing/windows.rs#L201)):*

I think either the io error should be logged here, or it should be the source. Currently, it is impossible to figure out why `get_mtu_for_route` fails based on the logs.
</blockquote></details>

Good point.
I decided to go with the logging option since
1. We want to both log the error from `winnet` and from `crate::windows` and we would either have to log at least one or create new error variants which encapsulate them.
2. We are not planning on treating these errors differently anyway, an error here will simply result in not setting the MTU.
3. It's nice to not have the interface use foreign errors if avoidable.
4. Creating multiple error variants just to log them and not treat them differently seems unnecessary.

I also realized that we should probably log the fact that setting the MTU fails for whatever reason if an error occurs. I've added this as well.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3703)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`talpid-core/src/routing/windows.rs` line 200 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3703#-N6h3IwI6zn5bdx-AA_Q:-N6h3IwI6zn5bdx-AA_R:b-dxfsxt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/89d71e03c4da67b56587ed06307a6a796a9164e4/talpid-core/src/routing/windows.rs#L200)):*
> ```XML
>                 Value: route.interface_luid,
>             };
>             let interface_row = match crate::windows::get_ip_interface_entry(addr_family, &luid) {
> ```

Nit: The logging could be thrown into the closure passed to `map_err` for brevity.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3703)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3705, opened 2022-06-28
Remove notify dependency and add inotify instead

Jontified:
Switch from notify to inotify since we only used notify on Linux.

Use inotify instead of notify since we only use notify on Linux. This allows us to get rid of old dependencies that we do not need.
Transitioning required some small changes in the API usage.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,7 +32,7 @@ version = "0.7.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9e8b47f52ea9bae42228d07ec09eb676433d7c4ed1ebdf0f1d1c29ed446f1ab8"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if",
  "cipher",
  "cpufeatures",
  "ctr",
@@ -138,7 +138,7 @@ checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
 dependencies = [
  "hermit-abi",
  "libc",
- "winapi 0.3.9",
+ "winapi",
 ]
 
 [[package]]
@@ -155,7 +155,7 @@ checksum = "e7a905d892734eea339e896738c14b9afce22b5318f64b951e70bf3844419b01"
 dependencies = [
  "addr2line",
  "cc",
- "cfg-if 1.0.0",
+ "cfg-if",
  "libc",
  "miniz_oxide",
  "object",
@@ -254,12 +254,6 @@ checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
 
 [[package]]
 name = "cfg-if"
-version = "0.1.10"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4785bdd1c96b2a846b2bd7cc02e86b6b3dbf14e7e53446c4f54c92a361040822"
-
-[[package]]
-name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
@@ -270,7 +264,7 @@ version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "01b72a433d0cf2aef113ba70f62634c56fddb0f244e6377185c56a7cadbd8f91"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if",
  "cipher",
  "cpufeatures",
  "zeroize",
@@ -300,7 +294,7 @@ dependencies = [
  "num-traits",
  "serde",
  "time 0.1.43",
- "winapi 0.3.9",
+ "winapi",
 ]
 
 [[package]]
@@ -354,7 +348,7 @@ checksum = "f4ffc801dacf156c5854b9df4f425a626539c3a6ef7893cc0c5084a23f0b6c59"
 dependencies = [
  "atty",
  "lazy_static",
- "winapi 0.3.9",
+ "winapi",
 ]
 
 [[package]]
@@ -445,7 +439,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a19c6cedffdc8c03a3346d723eb20bd85a13362bb96dc2ac000842c6381ec7bf"
 dependencies = [
  "nix 0.23.1",
- "winapi 0.3.9",
+ "winapi",
 ]
 
 [[package]]
@@ -510,7 +504,7 @@ checksum = "de0a745c25b32caa56b82a3950f5fec7893a960f4c10ca3b02060b0c38d8c2ce"
 dependencies = [
  "libc",
  "libdbus-sys",
- "winapi 0.3.9",
+ "winapi",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


faern (changes requested):
I LOVE what this PR is doing to our dependency tree :heart: 

> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705#-:-N5dgScw2F_tzuErGn0d:b-hxlgf5)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @Jontified)

___
*[`talpid-core/src/dns/linux/static_resolv_conf.rs` line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705#-N5djem32wAHTbup2bcA:-N5djem32wAHTbup2bcB:b-5odf5m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e8f739156a3f942a54844da8b0f6b14e0b0f26a4/talpid-core/src/dns/linux/static_resolv_conf.rs#L1)):*
> ```XML
> use super::RESOLV_CONF_PATH;
> ```

Not really related to your PR. But I think this constant can be moved into this file. As far as I can find, it's only ever used here, but for some reason defined in the parent module.
___
*[`talpid-core/src/dns/linux/static_resolv_conf.rs` line 109 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705#-N5dk6UKBdisLL6em1GX:-N5dk6UKBdisLL6em1GY:b-l5jww5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e8f739156a3f942a54844da8b0f6b14e0b0f26a4/talpid-core/src/dns/linux/static_resolv_conf.rs#L109)):*
> ```XML
> 
> struct DnsWatcher {
>     _pd: std::marker::PhantomData<()>,
> ```

Why do we need a phantom data here?
___
*[`talpid-core/src/dns/linux/static_resolv_conf.rs` line 132 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705#-N5djuWu9DJrEpl7yV8s:-N5djuWu9DJrEpl7yV8t:b-fs2mh8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8ca56fdc02e369aff199111e0ce6af8c2393ab3a/talpid-core/src/dns/linux/static_resolv_conf.rs#L132)):*
> ```XML
>                 .path
>                 .as_ref()
>                 .map(|p| p.as_path() == AsRef::<Path>::as_ref(RESOLV_CONF_PATH))
> ```

Looks like the new code does not check if the changed file is actually `RESOLV_CONF_PATH`? Do we trigger `Self::update` on changes to *all* files in `/etc` now? :thinking:

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705)* status: 3 of 4 files reviewed, 5 unresolved discussions (waiting on @Jontified)

___
*[`talpid-core/src/dns/linux/static_resolv_conf.rs` line 109 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705#-N5dk6UKBdisLL6em1GX:-N5e5yTH40mQeN9XYCHe:b-pj9wy3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e8f739156a3f942a54844da8b0f6b14e0b0f26a4/talpid-core/src/dns/linux/static_resolv_conf.rs#L109)):*

Why do we need a phantom data here?
</blockquote></details>

`struct DnsWatcher(())` is simpler if we just want to make it private.
___
*[`talpid-core/src/dns/linux/static_resolv_conf.rs` line 122 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705#-N5e5oi5DxufWbQkiff9:-N5e5oi5DxufWbQkiffA:b-thbmwy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6cc7b51c2d860db97abbdfcfc644db57a2310464/talpid-core/src/dns/linux/static_resolv_conf.rs#L122)):*
> ```XML
> 
>         let resolv_conf_wd = watcher
>             .add_watch(&RESOLV_CONF_DIR, mask)
> ```

Watch `RESOLV_CONF_PATH` instead
___
*[`talpid-core/src/dns/linux/static_resolv_conf.rs` line 139 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705#-N5e6OTV8qtxBuapH7PL:-N5e6OTV8qtxBuapH7PM:b93bh8l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6cc7b51c2d860db97abbdfcfc644db57a2310464/talpid-core/src/dns/linux/static_resolv_conf.rs#L139)):*
> ```XML
>         const EVENT_BUFFER_SIZE: usize = 1024;
>         let mut buffer = [0; EVENT_BUFFER_SIZE];
>         loop {
> ```

How do we cancel/abort the watcher?

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on @Jontified)

___
*[`talpid-core/src/dns/linux/static_resolv_conf.rs` line 105 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705#-N5ebIqp3QwmC76D8oeM:-N5ebIqp3QwmC76D8oeN:bu4xfs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e480e30ec8c843d5eb282cdac27515fce4d9dff1/talpid-core/src/dns/linux/static_resolv_conf.rs#L105)):*
> ```XML
> 
> struct DnsWatcher {
>     cancel_tx: Trigger,
> ```

Nitpicky: IMHO `tx/rx` are not super good names. This has also been discussed in the Rust community in general. Pointing out that they are pretty technical and not very friendly. `sender/receiver` are usually prefer. We break this in a million places. But unless we have a good reason to keep `tx/rx` I think we should move away from them in favor of more explanatory names.

In this case specifically it's not really a transmitter/receiver really, it's a trigger/listener. What do you think about naming this `cancel_trigger` and `cancel_listener` respectively?
___
*[`talpid-core/src/dns/linux/static_resolv_conf.rs` line 153 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705#-N5ecOSgDk5CQPOko4yL:-N5ecOSgDk5CQPOko4yM:bocptzz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e480e30ec8c843d5eb282cdac27515fce4d9dff1/talpid-core/src/dns/linux/static_resolv_conf.rs#L153)):*
> ```XML
>         loop {
>             tokio::select! {
>                 // To use a Receiver in a tokio::select! loop, add &mut in front of the channel.
> ```

Outdated comment. And probably not relevant?

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @Jontified)

faern (approved):
:lgtm:

> Reviewed 1 of 2 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3705#-:-N5iu4wm09sfwYYqC2TF:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3712, opened 2022-06-30
Remove dependencies reported as unused by cargo udeps

faern:
I was wondering why we had `openssl-probe` in our dependency tree. I looked further and saw it was pulled in as a default feature to `hyper-rustls`. I disabled that feature. Shortly after I realized that the entire crate was unused. So I ran `cargo udeps` and realized we had two unused dependencies.

I'll look at adding `cargo udeps` into our CI.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3712)

PR #3716, opened 2022-07-01
Add a cargo-udeps CI job to find unused dependencies

faern:
Make the CI check for unused dependencies. We just now had two unused dependencies (#3712), one of which pulled in an unwanted crate even. And we have had unused crates before. It's really easy to miss removing something from `Cargo.toml` when the last piece of code stops using it. So better automatically check it!


.github/workflows/rust-unused-dependencies.yml
```diff
@@ -0,0 +1,46 @@
...
+                sudo apt-get update
+                sudo apt-get install libdbus-1-dev
+
+            # cargo-udeps can compile on stable, but not run on stable.
+            # So we need a nighly toolchain. But `rust-cache@v1` keys
+            # the cache on the rustc version. So if we just use `nightly`
+            # the cache will be lost every day. To mitigate this we lock
+            # it to a specific nightly. Obviously we have to bump this from
+            # time to time or when `cargo-udeps` release a version incompatible
+            # with our nightly version.
```
faern, line 30: Another solution to this hack is to install both stable and nightly rustc and install `cargo-udeps` with stable (and have it cached under stable). That way the cache is only invalidated every six weeks (new Rust release). And then we run it under nightly. Installing toolchains is fast, so this is no problem. But if we stay on rolling `nightly` we risk having sudden breakage, which is also not desirable.

pinkisemils (approved):

PR #3719, opened 2022-07-04
Don't update the ST paths copy if applying the config fails

dlon:
Previously, if one had tried to update the list of apps to split but applying the config failed, that new config would still be reapplied after resuming from sleep. That's unexpected: The list should always be synchronized with the paths that are actively being split.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3719)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3719)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3727, opened 2022-07-05
Fix app occasionally getting stuck in the offline state

dlon:
`PowerManagementListener` previously only retained the most recent event. Since `ResumeSuspend` is sometimes sent at the same time as `ResumeAutomatic`, receivers would potentially only see one of the events. This PR fixes this by using a broadcast channel instead of `watch`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3727)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3727)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3736, opened 2022-07-10
Add search and pin locations

nchlscs:
iOS only

1. Add search bar to filter locations by country or city
2. Add button to pin favourite locations to the top of the list
3. Update settings manager to store favourite locations

Also, fixed a little bug with locations list scroll indicator insets on devices with safe area.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3736)


pronebird:
Awesome! I will have to adapt this work but overall great proof of concept. I'll take it with UI/UX folks and see how deep in the inbox the so-called "favourite locations" feature is.

nchlscs:
The current search solution, though, has an unobvious UX behaviour. If location has preceding node in hierarchy, it won’t show up. This can be confusing (and will be), say we are looking for Gothenburg, and on G request it looks like Gothenburg is in Germany, which is definitely not true.

before

The first idea was to disable indentation during the search, turned out, it requires some significant `LocationDataSource` refactor, and my goal was as little impact on the current structure as possible.

Today, I came up with an idea to include preceding nodes in search results, which is much better imho. Still not the best, since there are some neighbours in result (because of the `LocationDataSource` structure, again), but better.

after

 I did update this pull request to the latest solution. I can refactor `LocationDataSource` to better search-and-pin support, if necessary.

pronebird:
I think it makes sense that search results output a tree-like structure, so good work on fixing this. Lots of people are on summer vacation, so it will probably take some time to approve that work.

Also I think we'll be removing the header label above the table view so we could possibly use `navigationItem.searchController` instead and then switch to using `UITableViewController`, which should perhaps lower the amount of boilerplate needed to set up the view controller.

pronebird:
Sorry that it takes that long to get anything from us. It almost feels like it's been forever since this PR was first posted. There hasn't been any design decision on our side and key UI/UX folks are still absent. I hope it's not too terrible to rebase this work given that we re-formatted out codebase with `swiftformat`.

nchlscs:
That’s ok. I realize this work is just a subjective concept that requires a deep review from the design team and the final implementation may change significantly. I have no problem to keep it up to date time to time.

nchlscs:
I tried a different approach here. I’m using this branch as the main one for a while, and noticed, the pin button on each row looks overloaded. Since pinning is not everyday feature (usually, you pin once and forget), I decided to move it to the leading swipe action and just leave a small indicator displaying whether location is pinned.

pronebird:
That's a solid effort and it's a shame I have to close this PR. Search has already been implemeneted, while favorite locations are coming a bit later this year. With more folks working on iOS app we should be able to move at a higher pace. Thanks again for trying to improve the app, however we'd have to stick with the in-house implementation of these features as we need to align the UI across multiple platforms.

PR #3741, opened 2022-07-12
Fix MapKit rendering issues

pronebird:

Git checklist:

1. Fix regression related to interior polygons causing strokes rendered thicker than expected by extracting interior polygons and placing them after exterior polygon. Normally interior polygons should punch holes in the exterior polygon, i.e this is the case when big country has another country inside of. However that hasn't been working as expected as of late.

<table>
<tr>
<td>

BEFORE

</td>
<td>

AFTER

</td>
</tr>
</table>

2. Remove Antarctica since our data source seem to have a combination of coordinates that make it look like self-intersecting polygon which is then rendered inside out. Since we don't have servers there, I guess it's safe to simply remove it instead of trying to fix the data source.

3. Avoid force unwrapping when loading GeoJSON. Instead log errors and carry on.
4. Avoid using iOS 13+ MKGeoJSONDecoder since it would attempt to load interior polygons, see the first point.
5. Strip down metadata when parsing GeoJSON since we don't use any of it. Less code = better.
6. Move `GeoJSON/countries.geo.json` -> `Assets/countries.geo.json`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3741)


raksooo (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, 1 of 1 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3741#-:-N6mFh8v0TPrB_6nKt4d:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3745, opened 2022-07-13
Enter blocking state atomically when service is stopped

dlon:
There exists a small potential leak window when stopping the service and auto-connect is enabled:

```
[2022-07-13 11:37:45.058][talpid_core::firewall][INFO] Resetting firewall policy
[2022-07-13 11:37:45.160][mullvad_daemon][DEBUG] New tunnel state: Disconnected
[2022-07-13 11:37:45.171][talpid_core::firewall][INFO] Applying firewall policy: Blocked. Allowing LAN. Allowing endpoint 45.83.222.100:443/TCP for mullvad-problem-report.exe mullvad-daemon.exe
```

The firewall is supposed to block until the daemon has restarted, but in this particular case, the firewall state would be wiped prior to the blocking filters being added.

This is only relevant when always require VPN is disabled and auto-connect is enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3745)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Line wrap the file at 100 chars.                                              Th
 #### Android
 - Add device management to the Android app. This simplifies knowing which device is which and adds
   the option to log other devices out when the account already has five devices.
+
 #### Windows
 - Windows daemon now looks up the MTU on the default interface and uses this MTU instead of the
   default 1500. The 1500 is still the fallback if this for some reason fails. This may stop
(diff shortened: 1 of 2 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3745)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`mullvad-daemon/src/lib.rs` line 2144 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3745#-N7FBH4W4jqsZZLgcxLJ:-N7FVGkt8tPavs8MCx1j:bxohl57) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/935b71e1516dc118c7b0d1aa714ca915bc06d914/mullvad-daemon/src/lib.rs#L2144)):*

Is there a particular reason that this code is windows-only?
</blockquote></details>

This is done on Windows mainly to ensure that persistent firewall filters are added that keep blocking traffic after reboot as well, before the daemon has restarted. That's not a thing on Linux, so it wouldn't help much there. But it probably wouldn't hurt.

PR #3748, opened 2022-07-14
Migrate a given percentage of windows users to WG

Jontified:
Retrieve a percentage from the version check API response.
Use this percentage to decide if a user should use WG or OpenVpn as
their "automatic" default option.

Currently not implemented for windows or functioning and contains dbg
statements.

Git checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ Line wrap the file at 100 chars.                                              Th
 - Windows daemon now looks up the MTU on the default interface and uses this MTU instead of the
   default 1500. The 1500 is still the fallback if this for some reason fails. This may stop
   fragmentation.
+- The default VPN protocol is slowly being changed from OpenVPN to WireGuard.
+  The app fetches the ratio between the protocols from the API.
 
 #### Linux
 - GUI: Add electron flags to run Wayland native if in a compositor/desktop known to work well
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1727,6 +1727,7 @@ dependencies = [
  "jnix",
  "lazy_static",
  "log",
+ "rand 0.8.5",
  "regex",
  "serde",
  "talpid-types",
--- a/docs/relay-selector.md
+++ b/docs/relay-selector.md
@@ -45,10 +45,13 @@ constraints, following default ones will take effect:
 
 - If no tunnel protocol is specified for tunnel endpoints, then the behavior is different on Windows
   and other platforms.
-  - On Windows, OpenVPN is used.
   - On MacOS and Linux, first two connection attempts will use WireGuard, over a random port at
     first and then port 53. From the third attempt onwards, OpenVPN will be used, alternating
     between UDP on any port and TCP on port 443.
+  - On Windows, a migration to WireGuard is ongoing and a percentage value provided by the API tells
+    clients to randomly decide if they will use WireGuard as a default or OpenVPN as a default.
+    The client's decision will persist over time.
+    If the client decides to use WireGuard it will have the same behavior as MacOS and Linux.
 
 - If the tunnel protocol is specified as WireGuard without any other protocol constraints, then the
   transport protocol is not applicable as only UDP endpoints exist and any port will be matched.
--- a/mullvad-api/src/lib.rs
+++ b/mullvad-api/src/lib.rs
@@ -477,6 +477,16 @@ pub struct AppVersionResponse {
     pub latest: AppVersion,
     pub latest_stable: Option<AppVersion>,
     pub latest_beta: AppVersion,
+    #[serde(default = "default_wg_threshold")]
+    pub x_threshold_wg_default: f32,
+}
+
+/// Temporary function that will be removed later. Used to generate default wg_threshold.
+/// In case there is no `x_threshold_wg_default` returned by the API result we interpret that to
+/// mean that the migration is done and WireGuard should be the default. In that case the threshold
+/// value should be 1.0
+fn default_wg_threshold() -> f32 {
+    1.0
 }
 
 impl AppVersionProxy {
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -586,8 +586,9 @@ where
                 None
             });
         let settings = SettingsPersister::load(&settings_dir).await;
+        let app_version_info = version_check::load_cache(&cache_dir).await;
 
-        let initial_selector_config = new_selector_config(&settings);
+        let initial_selector_config = new_selector_config(&settings, &app_version_info);
         let relay_selector = RelaySelector::new(initial_selector_config, &resource_dir, &cache_dir);
 
         let proxy_provider =
@@ -697,7 +698,6 @@ where
             on_relay_list_update,
         );
 
-        let app_version_info = version_check::load_cache(&cache_dir).await;
         let (version_updater, version_updater_handle) = version_check::VersionUpdater::new(
             api_handle.clone(),
             api_availability.clone(),
@@ -822,7 +822,7 @@ where
             Command(command) => self.handle_command(command).await,
             TriggerShutdown => self.trigger_shutdown_event(),
             NewAppVersionInfo(app_version_info) => {
-                self.handle_new_app_version_info(app_version_info)
+                self.handle_new_app_version_info(app_version_info);
             }
             DeviceEvent(event) => self.handle_device_event(event).await,
             DeviceMigrationEvent(event) => self.handle_device_migration_event(event).await,
@@ -1032,6 +1032,8 @@ where
 
     fn handle_new_app_version_info(&mut self, app_version_info: AppVersionInfo) {
         self.app_version_info = Some(app_version_info.clone());
+        self.relay_selector
+            .set_config(new_selector_config(&self.settings, &self.app_version_info));
         self.event_listener.notify_app_version(app_version_info);
     }
 
@@ -1742,7 +1744,7 @@ where
                     self.event_listener
                         .notify_settings(self.settings.to_settings());
                     self.relay_selector
-                        .set_config(new_selector_config(&self.settings));
+                        .set_config(new_selector_config(&self.settings, &self.app_version_info));
                     log::info!("Initiating tunnel restart because the relay settings changed");
                     self.reconnect_tunnel();
                 }
@@ -1884,7 +1886,7 @@ where
                     self.event_listener
                         .notify_settings(self.settings.to_settings());
                     self.relay_selector
-                        .set_config(new_selector_config(&self.settings));
+                        .set_config(new_selector_config(&self.settings, &self.app_version_info));
                     if let Err(error) = self.api_handle.service().next_api_endpoint().await {
                         log::error!("Failed to rotate API endpoint: {}", error);
                     }
@@ -1914,7 +1916,7 @@ where
                     self.event_listener
                         .notify_settings(self.settings.to_settings());
                     self.relay_selector
-                        .set_config(new_selector_config(&self.settings));
+                        .set_config(new_selector_config(&self.settings, &self.app_version_info));
                     self.reconnect_tunnel();
                 }
                 Self::oneshot_send(tx, Ok(()), "set_obfuscation_settings");
@@ -1940,7 +1942,7 @@ where
                     self.event_listener
                         .notify_settings(self.settings.to_settings());
                     self.relay_selector
-                        .set_config(new_selector_config(&self.settings));
+                        .set_config(new_selector_config(&self.settings, &self.app_version_info));
                     log::info!("Initiating tunnel restart because bridge state changed");
                     self.reconnect_tunnel();
                 }
@@ -2253,11 +2255,34 @@ impl DaemonShutdownHandle {
     }
 }
 
-fn new_selector_config(settings: &Settings) -> SelectorConfig {
+fn new_selector_config(
+    settings: &Settings,
+    app_version_info: &Option<AppVersionInfo>,
+) -> SelectorConfig {
+    // In case of the app not having a version we safety default to OpenVPN to guard against the
+    // case where some error causes users to not recieve a version and in that case all going to
+    // wireguard. Magic number 0.0 implies that 0% of users should use Wireguard.
+    // This will be removed in the future when we the migration is done.
+    let wg_migration_threshold = app_version_info
+        .as_ref()
+        .map(|f| f.wg_migration_threshold)
+        .unwrap_or(0.0);
+
+    let default_tunnel_type = if cfg!(target_os = "windows") {
+        if wg_migration_threshold >= settings.wg_migration_rand_num {
+            TunnelType::Wireguard
+        } else {
+            TunnelType::OpenVpn
+        }
+    } else {
+        TunnelType::Wireguard
+    };
+
     SelectorConfig {
         relay_settings: settings.get_relay_settings(),
         bridge_state: settings.get_bridge_state(),
         bridge_settings: settings.bridge_settings.clone(),
         obfuscation_settings: settings.obfuscation_settings.clone(),
+        default_tunnel_type,
     }
 }
--- a/mullvad-daemon/src/settings.rs
+++ b/mullvad-daemon/src/settings.rs
@@ -5,6 +5,7 @@ use mullvad_types::{
     settings::{DnsOptions, Settings},
     wireguard::RotationInterval,
 };
+use rand::Rng;
 #[cfg(target_os = "windows")]
 use std::collections::HashSet;
 use std::{
@@ -68,6 +69,15 @@ impl SettingsPersister {
                 (settings, true)
             }
         };
+
+        // If the settings file did not contain a wg_migration_rand_num then it will be initialized
+        // to -1.0 by serde. This block ensures that this value is correctly intitialzed to a
+        // percentage.
+        if settings.wg_migration_rand_num < 0.0 || settings.wg_migration_rand_num > 1.0 {
+            let mut rng = rand::thread_rng();
+            settings.wg_migration_rand_num = rng.gen_range(0.0..=1.0);
+            should_save |= true
+        }
 
         // Force IPv6 to be enabled on Android
         if cfg!(target_os = "android") {
@@ -185,6 +195,7 @@ impl SettingsPersister {
     /// is being run or not.
     fn default_settings() -> Settings {
         let mut settings = Settings::default();
+
         if crate::version::is_beta_version() {
             settings.show_beta_releases = true;
         }
--- a/mullvad-daemon/src/version_check.rs
+++ b/mullvad-daemon/src/version_check.rs
@@ -47,7 +47,7 @@ const PLATFORM: &str = "windows";
 #[cfg(target_os = "android")]
 const PLATFORM: &str = "android";
 
-#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 struct CachedAppVersionInfo {
     #[serde(flatten)]
     pub version_info: AppVersionInfo,
@@ -281,11 +281,20 @@ impl VersionUpdater {
             self.show_beta_releases || is_beta_version(),
         );
 
+        let wg_migration_threshold = if response.x_threshold_wg_default.is_nan() {
+            // If the value should for some strange reason be NaN then safe default to 0.0
+            0.0
+        } else {
+            // Make sure that the returned value is between 0% and 100%
+            response.x_threshold_wg_default.clamp(0.0, 1.0)
+        };
+
         AppVersionInfo {
             supported: response.supported,
             latest_stable: response.latest_stable.unwrap_or_else(|| "".to_owned()),
             latest_beta: response.latest_beta,
             suggested_upgrade,
+            wg_migration_threshold,
         }
     }
 
@@ -375,6 +384,7 @@ impl VersionUpdater {
                                     latest_stable: last_app_version_info.latest_stable,
                                     latest_beta: last_app_version_info.latest_beta,
                                     suggested_upgrade,
+                                    wg_migration_threshold: last_app_version_info.wg_migration_threshold,
                                 }).await;
                             }
                         }
@@ -468,6 +478,7 @@ fn dev_version_cache() -> AppVersionInfo {
         latest_stable: PRODUCT_VERSION.to_owned(),
         latest_beta: PRODUCT_VERSION.to_owned(),
         suggested_upgrade: None,
+        wg_migration_threshold: 1.0,
     }
 }
 
--- a/mullvad-relay-selector/src/lib.rs
+++ b/mullvad-relay-selector/src/lib.rs
@@ -169,6 +169,7 @@ pub struct SelectorConfig {
     pub bridge_state: BridgeState,
     pub bridge_settings: BridgeSettings,
     pub obfuscation_settings: ObfuscationSettings,
+    pub default_tunnel_type: TunnelType,
 }
 
 #[derive(Clone)]
@@ -231,8 +232,12 @@ impl RelaySelector {
                 Ok((SelectedRelay::Custom(custom_relay.clone()), None, None))
             }
             RelaySettings::Normal(constraints) => {
-                let relay =
-                    self.get_tunnel_endpoint(constraints, config.bridge_state, retry_attempt)?;
+                let relay = self.get_tunnel_endpoint(
+                    constraints,
+                    config.bridge_state,
+                    retry_attempt,
+                    config.default_tunnel_type,
+                )?;
                 let bridge = match relay.endpoint {
                     MullvadEndpoint::OpenVpn(endpoint)
                         if endpoint.protocol == TransportProtocol::Tcp =>
@@ -271,6 +276,7 @@ impl RelaySelector {
         relay_constraints: &RelayConstraints,
         bridge_state: BridgeState,
         retry_attempt: u32,
+        default_tunnel_type: TunnelType,
     ) -> Result<NormalSelectedRelay, Error> {
         match relay_constraints.tunnel_protocol {
             Constraint::Only(TunnelType::OpenVpn) => self.get_openvpn_endpoint(
@@ -289,9 +295,12 @@ impl RelaySelector {
                 &relay_constraints.wireguard_constraints,
                 retry_attempt,
             ),
-            Constraint::Any => {
-                self.get_any_tunnel_endpoint(relay_constraints, bridge_state, retry_attempt)
-            }
+            Constraint::Any => self.get_any_tunnel_endpoint(
+                relay_constraints,
+                bridge_state,
+                retry_attempt,
+                default_tunnel_type,
+            ),
         }
     }
 
@@ -576,9 +585,14 @@ impl RelaySelector {
         relay_constraints: &RelayConstraints,
         bridge_state: BridgeState,
         retry_attempt: u32,
+        default_tunnel_type: TunnelType,
     ) -> Result<NormalSelectedRelay, Error> {
-        let preferred_constraints =
-            self.preferred_constraints(relay_constraints, bridge_state, retry_attempt);
+        let preferred_constraints = self.preferred_constraints(
+            relay_constraints,
+            bridge_state,
+            retry_attempt,
+            default_tunnel_type,
+        );
 
         if let Ok(result) = self.get_multihop_tunnel_endpoint_internal(&preferred_constraints) {
             log::debug!(
@@ -604,10 +618,12 @@ impl RelaySelector {
         original_constraints: &RelayConstraints,
         bridge_state: BridgeState,
         retry_attempt: u32,
+        default_tunnel_type: TunnelType,
     ) -> RelayConstraints {
         let (preferred_port, preferred_protocol, preferred_tunnel) = self
             .preferred_tunnel_constraints(
                 retry_attempt,
+                default_tunnel_type,
                 &original_constraints.location,
                 &original_constraints.providers,
                 &original_constraints.ownership,
@@ -929,40 +945,46 @@ impl RelaySelector {
     fn preferred_tunnel_constraints(
         &self,
         retry_attempt: u32,
+        default_tunnel_type: TunnelType,
         location_constraint: &Constraint<LocationConstraint>,
         providers_constraint: &Constraint<Providers>,
         ownership_constraint: &Constraint<Ownership>,
     ) -> (Constraint<u16>, TransportProtocol, TunnelType) {
-        #[cfg(target_os = "windows")]
-        {
-            let location_supports_openvpn =
-                self.parsed_relays.lock().relays().iter().any(|relay| {
-                    relay.active
-                        && relay.endpoint_data == RelayEndpointData::Openvpn
-                        && location_constraint.matches(relay)
-                        && providers_constraint.matches(relay)
-                        && ownership_constraint.matches(relay)
-                });
-            if location_supports_openvpn {
-                let (preferred_port, preferred_protocol) =
-                    Self::preferred_openvpn_constraints(retry_attempt);
-                return (preferred_port, preferred_protocol, TunnelType::OpenVpn);
-            }
-        }
+        match default_tunnel_type {
+            TunnelType::OpenVpn => {
+                let location_supports_openvpn =
+                    self.parsed_relays.lock().relays().iter().any(|relay| {
+                        relay.active
+                            && relay.endpoint_data == RelayEndpointData::Openvpn
+                            && location_constraint.matches(relay)
+                            && providers_constraint.matches(relay)
+                            && ownership_constraint.matches(relay)
+                    });
 
-        let location_supports_wireguard = self.parsed_relays.lock().relays().iter().any(|relay| {
-            relay.active
-                && matches!(relay.endpoint_data, RelayEndpointData::Wireguard(_))
-                && location_constraint.matches(relay)
-                && providers_constraint.matches(relay)
-                && ownership_constraint.matches(relay)
-        });
-        // If location does not support WireGuard, defer to preferred OpenVPN tunnel
-        // constraints
-        if !location_supports_wireguard {
-            let (preferred_port, preferred_protocol) =
-                Self::preferred_openvpn_constraints(retry_attempt);
-            return (preferred_port, preferred_protocol, TunnelType::OpenVpn);
+                if location_supports_openvpn {
+                    let (preferred_port, preferred_protocol) =
+                        Self::preferred_openvpn_constraints(retry_attempt);
+                    return (preferred_port, preferred_protocol, TunnelType::OpenVpn);
+                }
+            }
+            TunnelType::Wireguard => {
+                let location_supports_wireguard =
+                    self.parsed_relays.lock().relays().iter().any(|relay| {
+                        relay.active
+                            && matches!(relay.endpoint_data, RelayEndpointData::Wireguard(_))
+                            && location_constraint.matches(relay)
+                            && providers_constraint.matches(relay)
+                            && ownership_constraint.matches(relay)
+                    });
+
+                // If location does not support WireGuard, defer to preferred OpenVPN tunnel
+                // constraints
+                if !location_supports_wireguard {
+                    let (preferred_port, preferred_protocol) =
+                        Self::preferred_openvpn_constraints(retry_attempt);
+                    return (preferred_port, preferred_protocol, TunnelType::OpenVpn);
+                }
+            }
         }
 
         // Try out WireGuard in the first two connection attempts, first with any port,
@@ -1318,6 +1340,7 @@ mod test {
                     ..Default::default()
                 },
                 bridge_state: BridgeState::Auto,
+                default_tunnel_type: TunnelType::Wireguard,
             })),
         }
     }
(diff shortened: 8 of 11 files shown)


dlon (changes requested):
> Reviewed 3 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 3 of 9 files reviewed, 5 unresolved discussions (waiting on @Jontified)

___
*[`CHANGELOG.md` line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70C-CP4g-qXRahMOTn:-N70C-CP4g-qXRahMOTo:bpuwxmc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/CHANGELOG.md#L34)):*
> ```Markdown
>   default 1500. The 1500 is still the fallback if this for some reason fails. This may stop
>   fragmentation.
> - Windows users now sometimes default to using Wireguard tunnels rather than OpenVpn tunnels.
> ```

Nit: "Windows users" is redundant since this is under the Windows heading. :P
___
*[`mullvad-daemon/src/settings.rs` line 300 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70IZJFD5-Ocb9t5PIT:-N70IZJFD5-Ocb9t5PIU:b-ootur4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-daemon/src/settings.rs#L300)):*
> ```XML
>     }
> 
>     pub async fn set_x_wg_migration_rand_num(&mut self, x_wg_migration_rand_num: Option<XWgMigrationRandNum>) -> Result<bool, Error> {
> ```

I don't think this function should needed at all. Look at the `SettingsPersister` constructor (`SettingsPersister::load`). You could just generate the random number there, if it's not set.

It would also be nice for the field to be non-optional, but I'm not sure if that can be solved nicely. Using `#[serde(default = "random_number")]` would probably work, but you'd also have to make sure to save the settings if the field was not present in the file.
___
*[`mullvad-daemon/src/tunnel.rs` line 136 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70NQx14sr3JUMqZhdV:-N70NQx14sr3JUMqZhdW:b-1q684a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-daemon/src/tunnel.rs#L136)):*
> ```XML
> impl InnerParametersGenerator {
>     async fn generate(&mut self, retry_attempt: u32) -> Result<TunnelParameters, Error> {
>         let x_threshold_wg_default = match load_cache(&self.cache_dir).await {
> ```

As we discussed elsewhere, it would probably be simpler to add a `default_tunnel_type` or `preferred_tunnel_type` field to the relay selector config.
___
*[`mullvad-types/src/version.rs` line 40 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70DMVmArhIxBNR9P-x:-N70DMVmArhIxBNR9P-y:bnsgpx1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-types/src/version.rs#L40)):*
> ```XML
>     /// Temporary field used to decide if a user should default to Wireguard or OpenVpn.
>     /// NOTE: This field will be removed completely in future versions.
>     pub x_threshold_wg_default: Option<XThresholdWgDefault>,
> ```

I think you can get around having to newtype this by just removing `Eq` from this type as well as any type that contains it.
___
*[`mullvad-types/src/settings/mod.rs` line 97 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70CP799dD3kZ0HIXCd:-N70CP799dD3kZ0HIXCe:bnwmf3n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-types/src/settings/mod.rs#L97)):*
> ```XML
>     /// use wireguard or openvpn when the automatic feature is set. This variable will be removed
>     /// in future versions.
>     pub x_wg_migration_rand_num: Option<XWgMigrationRandNum>,
> ```

Is the newtype really needed here?
___
*[`mullvad-types/src/settings/mod.rs` line 97 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70MeBG0l3vO8MMtQ09:-N70MeBG0l3vO8MMtQ0A:b-jkugb2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-types/src/settings/mod.rs#L97)):*
> ```XML
>     /// use wireguard or openvpn when the automatic feature is set. This variable will be removed
>     /// in future versions.
>     pub x_wg_migration_rand_num: Option<XWgMigrationRandNum>,
> ```

Nit: Could we keep `settings_version` at the bottom?

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 3 of 9 files reviewed, 7 unresolved discussions (waiting on @Jontified)

___
*[`CHANGELOG.md` line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7Fzm3DDS0Mtd3d0LjN:-N7Fzm3DDS0Mtd3d0LjO:blqwls6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/CHANGELOG.md#L34)):*
> ```Markdown
>   default 1500. The 1500 is still the fallback if this for some reason fails. This may stop
>   fragmentation.
> - Windows users now sometimes default to using Wireguard tunnels rather than OpenVpn tunnels.
> ```

Nit: OpenV**PN** and Wire**G**uard is the correct capitalizations, respectively.
___
*[`mullvad-types/src/version.rs` line 40 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7G1Cl33cVg9UmWWobf:-N7G1Cl33cVg9UmWWobg:b-2po803) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-types/src/version.rs#L40)):*
> ```XML
>     /// Temporary field used to decide if a user should default to Wireguard or OpenVpn.
>     /// NOTE: This field will be removed completely in future versions.
>     pub x_threshold_wg_default: Option<XThresholdWgDefault>,
> ```

`#[serrde(default = 1.0)]`

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 0 of 11 files reviewed, 6 unresolved discussions (waiting on @dlon and @faern)

___
*[`mullvad-api/src/lib.rs` line 488 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7GXFgJ4qNhDij9-Wmi:-N7GXFgJ4qNhDij9-Wmj:b-lqh8j5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0bcba5915d4286125f7ec6f7517cea55e4df3dc1/mullvad-api/src/lib.rs#L488)):*
> ```XML
> fn default_wg_threshold() -> f32 {
>     // MAGIC NUMBER 1.0 is the default threshold of 100%
>     1.0
> ```

After some discussion with @faern we decided to make the threshold value into a `f32` rather than an `Option`. After doing this I'm a little bit unsure if this is a good idea. We end up having a bunch of `1.0`s as default everywhere, it might be more intuitive if there is just a bunch of `None`s everywhere instead and we then unwrap that to `1.0` when the value is used. Thoughts?
___
*[`mullvad-daemon/src/settings.rs` line 300 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70IZJFD5-Ocb9t5PIT:-N7GarXn2-MM5rSO5ddY:b9t5t56) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-daemon/src/settings.rs#L300)):*

I don't think this function should needed at all. Look at the `SettingsPersister` constructor (`SettingsPersister::load`). You could just generate the random number there, if it's not set.

It would also be nice for the field to be non-optional, but I'm not sure if that can be solved nicely. Using `#[serde(default = "random_number")]` would probably work, but you'd also have to make sure to save the settings if the field was not present in the file.
</blockquote></details>

You make a good point. The new solution is to avoid having an option and instead defaulting to a number outside of the range (-1) and inside of `SettingsPersister::load` check if the random value is outside of the range and in that case generate a new value and save it.
___
*[`mullvad-types/src/version.rs` line 40 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70DMVmArhIxBNR9P-x:-N7GYERC0VgR_OjdP2u1:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-types/src/version.rs#L40)):*

I think you can get around having to newtype this by just removing `Eq` from this type as well as any type that contains it.
</blockquote></details>

Done.
___
*[`mullvad-types/src/version.rs` line 40 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7G1Cl33cVg9UmWWobf:-N7GYFGlCF08BQp9TPVw:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-types/src/version.rs#L40)):*

`#[serrde(default = 1.0)]`
</blockquote></details>

Done.
___
*[`mullvad-types/src/settings/mod.rs` line 97 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70CP799dD3kZ0HIXCd:-N7GYGHsEBudIZ7uhQp7:bo8pv6h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-types/src/settings/mod.rs#L97)):*

Is the newtype really needed here?
</blockquote></details>

Removed `Eq`
___
*[`mullvad-daemon/src/tunnel.rs` line 136 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70NQx14sr3JUMqZhdV:-N7GZO9jCyMJg0tAihLh:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-daemon/src/tunnel.rs#L136)):*

As we discussed elsewhere, it would probably be simpler to add a `default_tunnel_type` or `preferred_tunnel_type` field to the relay selector config.
</blockquote></details>

Done.

dlon (changes requested):
> Reviewed 2 of 11 files at r2, 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: all files reviewed, 5 unresolved discussions (waiting on @faern and @Jontified)

___
*[`mullvad-api/src/lib.rs` line 488 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7GXFgJ4qNhDij9-Wmi:-N7H3eAp87WWltxpBMdY:b-mq516d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0bcba5915d4286125f7ec6f7517cea55e4df3dc1/mullvad-api/src/lib.rs#L488)):*

After some discussion with @faern we decided to make the threshold value into a `f32` rather than an `Option`. After doing this I'm a little bit unsure if this is a good idea. We end up having a bunch of `1.0`s as default everywhere, it might be more intuitive if there is just a bunch of `None`s everywhere instead and we then unwrap that to `1.0` when the value is used. Thoughts?
</blockquote></details>

I don't have a strong opinion on this.
___
*[`mullvad-daemon/src/settings.rs` line 300 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70IZJFD5-Ocb9t5PIT:-N7GxWNO2lTHiJRzELaE:b5qtem8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-daemon/src/settings.rs#L300)):*

You make a good point. The new solution is to avoid having an option and instead defaulting to a number outside of the range (-1) and inside of `SettingsPersister::load` check if the random value is outside of the range and in that case generate a new value and save it.
</blockquote></details>

This is pretty good.
___
*[`mullvad-daemon/src/settings.rs` line 76 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7H0eOD542WzkDzqLse:-N7H0eOD542WzkDzqLsf:b1hwo16) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5ec9c9645250acb617b6c7770c7a70ef61e4fd5/mullvad-daemon/src/settings.rs#L76)):*
> ```XML
>             use rand::Rng;
>             let mut rng = rand::thread_rng();
>             settings.wg_migration_rand_num = rng.gen_range(0.0, 1.0);
> ```

This is never reached by `reset`. So the random number might have to be generated in `Self::default_settings` as well, unfortunately.
___
*[`mullvad-relay-selector/src/lib.rs` line 178 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7GxMAe4BmqxoHr0Aql:-N7GxMAf4oNSwPu28sMf:b-sbfy3v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5ec9c9645250acb617b6c7770c7a70ef61e4fd5/mullvad-relay-selector/src/lib.rs#L178)):*
> ```XML
>     pub bridge_settings: BridgeSettings,
>     pub obfuscation_settings: ObfuscationSettings,
>     pub default_tunnel_protocol: DefaultTunnelProtocol,
> ```

Nit: Using `TunnelType` here would probably be fine. Having this additional type doesn't add much.
___
*[`mullvad-types/src/version.rs` line 42 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7H5URH3JjGde13yMHe:-N7H5URH3JjGde13yMHf:b-10on1w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5ec9c9645250acb617b6c7770c7a70ef61e4fd5/mullvad-types/src/version.rs#L42)):*
> ```XML
>     /// NOTE: This field will be removed completely in future versions.
>     // TODO: Should be windows only
>     #[serde(default = "default_wg_threshold")]
> ```

Arguably, the default isn't needed. If the file does not contain the field, it should be discarded anyway (as it will be, if deserialization fails).
___
*[`mullvad-daemon/src/tunnel.rs` line 136 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70NQx14sr3JUMqZhdV:-N7Gy4c232jI3TSNiWOD:b-6znjz4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-daemon/src/tunnel.rs#L136)):*

Done.
</blockquote></details>

Nice! This is much better.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[`mullvad-daemon/src/settings.rs` line 300 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70IZJFD5-Ocb9t5PIT:-N7KS4hC4zDm_VM3d1tP:b-p9aie1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-daemon/src/settings.rs#L300)):*

This is pretty good.
</blockquote></details>

After a bit of thinking I'm wondering if this is a good idea. It in a sense allows bugs to silently fall through, if the random value is left uninitialized due to some oversight (such as the reset path) then we end up with having a percentage initialized as -1 rather than None.
-1 in this case is probably fine, it means you'll default to using wireguard. However the practice in general seems flawed. If we for whatever reason do not want to panic in the case of a bug (and instead log it as an error for example) then we could make sure to use `unwrap_or`.
___
*[`mullvad-daemon/src/settings.rs` line 76 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7H0eOD542WzkDzqLse:-N7KS1Op912HBzTIf8Oo:bt2vr1b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5ec9c9645250acb617b6c7770c7a70ef61e4fd5/mullvad-daemon/src/settings.rs#L76)):*

This is never reached by `reset`. So the random number might have to be generated in `Self::default_settings` as well, unfortunately.
</blockquote></details>

Fixed.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 4 of 11 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[`mullvad-daemon/src/settings.rs` line 76 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7H0eOD542WzkDzqLse:-N7LB-k6DwqoLl___Bm8:b-4nwv5j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f5ec9c9645250acb617b6c7770c7a70ef61e4fd5/mullvad-daemon/src/settings.rs#L76)):*

Fixed.
</blockquote></details>

`reset` uses the `Default::default` initialization and I've now fixed so that this correctly initializes. `load` is as far as I can see the only public API which loads `Settings` from disk and as such this should be the only place we need to ensure that the value is re-initialized.
___
*[`mullvad-types/src/settings/mod.rs` line 142 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7LAa8E6P3D62frN_hA:-N7LAa8E6P3D62frN_hB:b-i6q1vc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f0403d0cc8ef9a4d9090f12ea5a075bb4431c458/mullvad-types/src/settings/mod.rs#L142)):*
> ```XML
>             show_beta_releases: false,
>             // TODO: Should be windows only
>             wg_migration_rand_num: rand_percent(),
> ```

This should make sure that any init of `Settings` that does not go through the filesystem (and serde) contains a correctly initialized value.
It does require that we import `rand` in the types crate. I've imported `rand` 0.8 since that's the one we are trying to target.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 3 of 11 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[`mullvad-types/src/version.rs` line 40 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7G1Cl33cVg9UmWWobf:-N7LOE3P6mb73_8VuuXC:b-w8zvqg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-types/src/version.rs#L40)):*

Done.
</blockquote></details>

After some discussion with @dlon. I believe the correct behavior is to have the serialization fail to indicate that the cached version is out of date and a newer version should be downloaded. The daemon can function with a None value as the app version.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 1 of 11 files reviewed, 4 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`CHANGELOG.md` line 35 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7URswX0StkwgUC1x2V:-N7URswX0StkwgUC1x2W:b-209mku) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/51060200514cd37bcb83f759c3b098e6e0dcbb11/CHANGELOG.md#L35)):*
"sometimes defaults" can be interpreted like it's randomized per connection attempt. I also feel like "the daemon" is a bit too specific. Yeah, the daemon creates the connections, but conceptually, the app is the overall entity that creates tunnels from a user perspective. How about:

> The default VPN protocol is slowly being changed to WireGuard. The app fetches the ratio between the protocols from the API.

My formulation answers the question "Why is this being done?" - Because we want to change the default protocol, but in a controlled manner.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 1 of 11 files reviewed, 3 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`mullvad-types/src/settings/mod.rs` line 142 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7LAa8E6P3D62frN_hA:-N7UYCNu98widy9Xd8sT:b9s4815) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f0403d0cc8ef9a4d9090f12ea5a075bb4431c458/mullvad-types/src/settings/mod.rs#L142)):*

This should make sure that any init of `Settings` that does not go through the filesystem (and serde) contains a correctly initialized value.
It does require that we import `rand` in the types crate. I've imported `rand` 0.8 since that's the one we are trying to target.
</blockquote></details>

Nit: Personally I think it might be simpler to just have
```rust
wg_migration_rand_num: rand::thread_rng().gen_range(0.0..1.0)
```
over a separate function.

And as we talked about in the office. It might be a nice refactor to not make `Settings` implement `Default` at all. Instead define the default values for all fields in `mullvad-daemon`. The benefits of doing so:
* No need to depend on `rand` in `mullvad-types` (also, having a `Default` implementation return different values on different invocations is not idiomatic.
* All defaults in one place. Currently they are spread out in the entire code base due to `Default` requiring us to implement `Default` for all types that are part of `Settings`, which is types spread out in many places.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 1 of 11 files reviewed, 4 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`mullvad-relay-selector/src/lib.rs` line 621 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7UZTeY1CY0bErtoWx-:-N7UZTeY1CY0bErtoWx0:bqo0fs3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/51060200514cd37bcb83f759c3b098e6e0dcbb11/mullvad-relay-selector/src/lib.rs#L621)):*
> ```XML
>         bridge_state: BridgeState,
>         retry_attempt: u32,
>         default_tunnel_type: &TunnelType,
> ```

Tiny `Copy` types are usually passed by ownership rather than by reference. I'm surprised clippy does not pick this up? :thinking:

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 1 of 11 files reviewed, 4 unresolved discussions (waiting on @dlon and @faern)

___
*[`CHANGELOG.md` line 35 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7URswX0StkwgUC1x2V:-N7Uq7iD2J9BnYcNrUtg:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/51060200514cd37bcb83f759c3b098e6e0dcbb11/CHANGELOG.md#L35)):*

"sometimes defaults" can be interpreted like it's randomized per connection attempt. I also feel like "the daemon" is a bit too specific. Yeah, the daemon creates the connections, but conceptually, the app is the overall entity that creates tunnels from a user perspective. How about:

> The default VPN protocol is slowly being changed to WireGuard. The app fetches the ratio between the protocols from the API.

My formulation answers the question "Why is this being done?" - Because we want to change the default protocol, but in a controlled manner.
</blockquote></details>

Done.
___
*[`mullvad-relay-selector/src/lib.rs` line 621 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7UZTeY1CY0bErtoWx-:-N7UphsL4d2TpX9BVYGn:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/51060200514cd37bcb83f759c3b098e6e0dcbb11/mullvad-relay-selector/src/lib.rs#L621)):*

Tiny `Copy` types are usually passed by ownership rather than by reference. I'm surprised clippy does not pick this up? :thinking: 
</blockquote></details>

Done.

dlon (changes requested):
> Reviewed 1 of 7 files at r4, 1 of 7 files at r5, 7 of 8 files at r6, 1 of 1 files at r7, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon, @faern, and @Jontified)

___
*[`mullvad-relay-selector/src/lib.rs` line 964 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7Uwhme2jBUgblVyNak:-N7Uwhme2jBUgblVyNal:b-uvfcfj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/51060200514cd37bcb83f759c3b098e6e0dcbb11/mullvad-relay-selector/src/lib.rs#L964)):*
> ```XML
>                     });
> 
>                 if location_supports_openvpn {
> ```

This code no longer defaults to WireGuard when the constraints only match wg relays.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 10 of 11 files reviewed, 4 unresolved discussions (waiting on @dlon and @faern)

___
*[`mullvad-relay-selector/src/lib.rs` line 964 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N7Uwhme2jBUgblVyNak:-N7V498I74R_jyADOcgh:biyomqq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/51060200514cd37bcb83f759c3b098e6e0dcbb11/mullvad-relay-selector/src/lib.rs#L964)):*

This code no longer defaults to WireGuard when the constraints only match wg relays.
</blockquote></details>

if `location_supports_openvpn` is false then it will default to using WG for the first two retries and OpenVPN for the remainder. So even if `default_tunnel_type` is passed as `OpenVpn` if the constraints do not permit OpenVPN it should use WG.

faern (commented):
> Reviewed 2 of 8 files at r6, 1 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 10 of 11 files reviewed, 2 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`mullvad-daemon/src/settings.rs` line 300 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N70IZJFD5-Ocb9t5PIT:-N87vU3F5fajbl26Bueg:b253tir) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19ec51526b9ed5cdf0ef661073e8950fcc41324c/mullvad-daemon/src/settings.rs#L300)):*

After a bit of thinking I'm wondering if this is a good idea. It in a sense allows bugs to silently fall through, if the random value is left uninitialized due to some oversight (such as the reset path) then we end up with having a percentage initialized as -1 rather than None.
-1 in this case is probably fine, it means you'll default to using wireguard. However the practice in general seems flawed. If we for whatever reason do not want to panic in the case of a bug (and instead log it as an error for example) then we could make sure to use `unwrap_or`.
</blockquote></details>

Exposing something that should be initialized as an option is also a design flaw. It tells the user "this should be a number, but I don't trust myself, so you have to handle my errors for me :shrug:"

As discussed AFK. This should be a float, and it should only be initialized in a single place. `Settings` should not implement `Default` and must be loaded from the persister at all times. Even resets should go through this settings management type.
___
*[`mullvad-daemon/src/settings.rs` line 76 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N87v41i-yo0sNUfVAnt:-N87v41i-yo0sNUfVAnu:b-rqk1jj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/46d4b4554c664af12856799b0ee1c04961d21f2a/mullvad-daemon/src/settings.rs#L76)):*
> ```XML
>         // percentage.
>         if settings.wg_migration_rand_num < 0.0 || settings.wg_migration_rand_num > 1.0 {
>             use rand::Rng;
> ```

Nit. But I don't see any reason not to import this at the top of the file.

dlon (commented):
> Reviewed 1 of 1 files at r9, 3 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (commented):
> Reviewed 3 of 8 files at r6, 2 of 3 files at r10, 1 of 1 files at r11.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`mullvad-daemon/src/settings.rs` line 78 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N95pDEY-TdSPBBEoZZH:-N95pDEY-TdSPBBEoZZI:b-omp6bo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b424c52a4cbb2dfe4d7d72622c8fcd557a26b6ea/mullvad-daemon/src/settings.rs#L78)):*
> ```XML
>         if settings.wg_migration_rand_num < 0.0 || settings.wg_migration_rand_num > 1.0 {
>             let mut rng = rand::thread_rng();
>             settings.wg_migration_rand_num = rng.gen_range(0.0..1.0);
> ```

Not that it really matters. But this range is exclusive of `1.0`. But the if-check above has `>1.0` meaning up there `1.0` is a valid value. Really nitpicky and has absolutely no impact on the results, but a small inconsistency nonetheless

faern (changes requested):
> Reviewed 1 of 7 files at r4, 1 of 8 files at r6, 1 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`mullvad-daemon/src/settings.rs` line 196 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N95sfFNCMMUsaoKFKjA:-N95sfFNCMMUsaoKFKjB:b-5gq13j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b424c52a4cbb2dfe4d7d72622c8fcd557a26b6ea/mullvad-daemon/src/settings.rs#L196)):*
> ```XML
>     /// Modifies `Settings::default()` somewhat, e.g. depending on whether a beta version
>     /// is being run or not.
>     fn default_settings() -> Settings {
> ```

Can we move the randomized field initialization from `Settings::default` into here? Seems more logical since we already have this method for modifying the defaults :thinking:

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 11 of 12 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`mullvad-daemon/src/settings.rs` line 196 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N95sfFNCMMUsaoKFKjA:-N96DqlC3BebOqjRr-Oo:b-biopa1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b424c52a4cbb2dfe4d7d72622c8fcd557a26b6ea/mullvad-daemon/src/settings.rs#L196)):*

Can we move the randomized field initialization from `Settings::default` into here? Seems more logical since we already have this method for modifying the defaults :thinking: 
</blockquote></details>

The problem that occurs is that since `Settings::default` is public it needs to return a correct version of `Settings` and as such must initialize.  We can not remove every public constructor of `Settings` as that would make us unable to create it in the `mullvad-daemon` crate unless we also move `Settings` into this crate.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 9 of 12 files reviewed, 2 unresolved discussions (waiting on @dlon, @faern, and @Jontified)

___
*[`docs/relay-selector.md` line 57 at r13](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N979vuAC8gzABFFK3La:-N979vuAC8gzABFFK3Lb:b-isk81q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6d7d0a9216e3a4b371617e14267ba03faabba218/docs/relay-selector.md#L57)):*
> ```Markdown
> 
> - If no tunnel protocol is specified for tunnel endpoints, then the behavior is different on Windows
>   and other platforms.
> ```

What happened here? This is just a verbatim copy of the above.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: 9 of 12 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[`docs/relay-selector.md` line 57 at r13](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N979vuAC8gzABFFK3La:-N9Apv3h1k1FB6FDLmeQ:b-4da7n8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6d7d0a9216e3a4b371617e14267ba03faabba218/docs/relay-selector.md#L57)):*

What happened here? This is just a verbatim copy of the above.
</blockquote></details>

Oops! I made a copy of the old text in order to have a reference when writing and then I forgot that it wasn't just another paragraph. Removed.

faern (commented):
> Reviewed 2 of 2 files at r13, 1 of 1 files at r14.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`docs/relay-selector.md` line 53 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N9COhpMAv4XZXGeeVcY:-N9COhpMAv4XZXGeeVcZ:b-mke1hv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8748e641a6e11856c5864b8e800de1edc537f86b/docs/relay-selector.md#L53)):*
> ```Markdown
>   - On Windows, a migration to WireGuard is ongoing and a percentage value provided by the API tells
>     clients to randomly decide if they will use WireGuard as a default or OpenVPN as a default.
>     The clients decision will persist over time.
> ```

Nit: "client's" or "clients'".

dlon (approved):
> Reviewed 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (changes requested):
> Reviewed 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`mullvad-relay-selector/src/lib.rs` line 279 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N9GgExJ1slsuFrxb7kU:-N9GgExJ1slsuFrxb7kV:b1xwsl0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/11c596ff8490cd1e72052095aaba17e0fb882216/mullvad-relay-selector/src/lib.rs#L279)):*
> ```XML
>         bridge_state: BridgeState,
>         retry_attempt: u32,
>         default_tunnel_type: TunnelType,
> ```

Why does a method that takes `&self` need to have this argument passed in? It's stored in `self`.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`mullvad-relay-selector/src/lib.rs` line 279 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748#-N9GgExJ1slsuFrxb7kU:-N9VO7IxCTkJPL0ifXmg:b-5p8vwu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/11c596ff8490cd1e72052095aaba17e0fb882216/mullvad-relay-selector/src/lib.rs#L279)):*

Why does a method that takes `&self` need to have this argument passed in? It's stored in `self`.
</blockquote></details>

Discussed on slack. Because of how the relay selector is currently designed. We lean towards removing the locks from within this type and instead let everyone who want to use it have a `Arc<Mutex<RelaySelector>>` in order to greatly simplify the relay selector itself.

faern (approved):
> Reviewed 1 of 1 files at r9, 2 of 2 files at r13.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3748)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3751, opened 2022-07-15
Saner logging when version checks are disabled

dlon:
Previously, when you ran `mullvad version` in a dev build, you'd see the following error:

```
[2022-06-23 13:23:46.360][mullvad_daemon][DEBUG] No version cache found. Fetching new info
[2022-06-23 13:23:46.360][mullvad_daemon][ERROR] Error: Error running version check
Caused by: Version cache update was aborted
```

This is because all commands sent to the version check actor were dropped.

This PR makes the following changes:
1. When running a dev build, always return valid (fake) version info.
2. When running the version check explicitly, just log a message pointing out that version checks are disabled.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3751)

### DIFF
--- a/mullvad-daemon/src/version_check.rs
+++ b/mullvad-daemon/src/version_check.rs
@@ -343,7 +343,12 @@ impl VersionUpdater {
         // If this is a dev build, there's no need to pester the API for version checks.
         if *IS_DEV_BUILD {
             log::warn!("Not checking for updates because this is a development build");
-            while rx.next().await.is_some() {}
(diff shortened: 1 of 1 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3751)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3755, opened 2022-07-18
Upgrade x25519-dalek to latest (pre) version to get newer rand

faern:
Upgrades `x25519-dalek` to the latest pre-release in order to get it to depend on `rand_core 0.6` instead of the old `0.5` version. We already had the newer rand crates in our dependency tree, so this does not really change much in our overall dependency tree. But this unlocks upgrading `rand` in all other crates. So this PR also upgrades `rand` to version `0.8.5` everywhere.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3755)


faern:
I bumped the `rand` version everywhere to make it consistent. I picked the latest version available right now according to https://github.com/mullvad/coding-guidelines/blob/master/rust.md#specifying-version-numbers

PR #3756, opened 2022-07-18
Add swift-format

pronebird:

Git checklist:

After careful examination of [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) and [Apple swift-format](https://github.com/apple/swift-format) I think I lean towards the former tool as it seems to be feature rich and well documented.

This PR adds [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) support on Github actions but does not reformat the entire codebase, which I think deserves a separate PR once we're light on unmerged changes.

The latest `SwiftFormat` can be installed via Homebrew:

```brew install swiftformat```

Use the following command to format the source code within the repository:

```
cd ios
swiftformat .
```

To view the complete list of supported rules, run:

```
swiftformat --ruleinfo
```

## Basic configuration

- Swift 5.5
- 4 space indentation
- 100 character limit per line

## Rule overrides

- `--wraparguments before-first`

  ```diff
  - foo(bar: Int,
  -     baz: String)

  + foo(
  +   bar: Int,
  +   baz: String
  + )

  - class Foo<Bar,
  -           Baz>
  
  + class Foo<
  +   Bar,
  +   Baz
  + >
  ```

- `--wrapternary before-operator`

  ```diff
  - let result = condition ? "Truthy statement" : "False statement"
  + let result = condition
  +     ? "Truthy statement"
  +     : "False statement"
  ```

- `--wrapcollections before-first`:

  ```diff
  - let foo = [bar,
               baz,
  -            quuz]

  + let foo = [
  +   bar,
      baz,
  +   quuz
  + ]
  ```

- `redundanttype`

  ```diff
  - let view: UIView = UIView()
  + let view = UIView()
  ```

## Disabled rules

- `initCoderUnavailable` - Add @available(*, unavailable) attribute to required `init(coder:)` when it hasn't been implemented.

  ```diff
  + @available(*, unavailable)
    required init?(coder aDecoder: NSCoder) {
      fatalError("init(coder:) has not been implemented")
    }
  ```

- `redundantReturn` - all returns become implicit (i.e no `return` statement). Subjective, I dislike implicit returns in Swift.
- `unusedArguments` - replaces all unused arguments with `_`. Subjective, I don't think it has to be enforced. I think there are legit cases in closures, but it can be done manually by developers.
- `redundantRawValues` -  Remove redundant raw string values for enum cases. Subjective, I think it can do some harm, for example settings store keys cannot be renamed without performing migration for older versions of the app. On the other side, normally, raw values are not specified by hand when `enum` is inherited from a primitive type.
  
  ```diff
    enum Foo: String {
  -   case bar = "bar"
      case baz = "quux"
    }

    enum Foo: String {
  +   case bar
      case baz = "quux"
    }
  ```
- `preferKeyPath` - Convert trivial map `{ $0.foo }` closures to keyPath-based syntax. Subjective. I am not used to this expression style that much. Although I don't mind if others on the team do that, I'd probably write closure by hand. Hence I feel like it does not have to be enforced.

  ```diff
  - let barArray = fooArray.map { $0.bar }
  + let barArray = fooArray.map(\.bar)

  - let barArray = fooArray.compactMap { $0.optionalBar }
  + let barArray = fooArray.compactMap(\.optionalBar)
  ```

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756)

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -2,11 +2,29 @@ name: iOS app
 on:
     pull_request:
         paths:
-            - .github/workflows/ios.yml
-            - ios/**
+            - ".github/workflows/ios.yml"
+            - "ios/.swiftformat"
+            - "ios/**/*.swift"
     # Build if requested manually from the Actions tab
     workflow_dispatch:
 jobs:
+    check-formatting:
+        name: Check formatting
+        runs-on: macos-11
+        steps:
+            - name: Install SwiftFormat
+              run: |
+                brew update
+                brew upgrade swiftformat
+
+            - name: Checkout repository
+              uses: actions/checkout@v3
+
+            - name: Check formatting
+              run: |
+                swiftformat --version
+                swiftformat --lint .
+
     test:
         name: Unit tests
         runs-on: macos-11
@@ -17,10 +35,10 @@ jobs:
             source_packages_dir: .spm
         steps:
             - name: Checkout repository
-              uses: actions/checkout@v2
+              uses: actions/checkout@v3
 
             - name: Configure cache
-              uses: actions/cache@v2
+              uses: actions/cache@v3
               with:
                 path: ios/${{ env.source_packages_dir }}
                 key: ${{ runner.os }}-spm-${{ hashFiles('ios/**/Package.resolved') }}
@@ -28,7 +46,7 @@ jobs:
                   ${{ runner.os }}-spm-
 
             - name: Setup go-lang
-              uses: actions/setup-go@v2
+              uses: actions/setup-go@v3
               with:
                   go-version: '1.16.5'
 
--- a/ios/.swiftformat
+++ b/ios/.swiftformat
@@ -1,0 +1,14 @@
+# file options
+--exclude Build, .spm, MullvadVPNScreenshots/SnapshotHelper.swift
+
+# general options
+--swiftversion 5.5
+
+# format options
+--indent 4
+--maxwidth 100
+--wraparguments before-first
+--wrapcollections before-first
+--wrapternary before-operators
+--redundanttype inferred
+--disable initCoderUnavailable, redundantReturn, unusedArguments, redundantRawValues, preferKeyPath
--- a/ios/README.md
+++ b/ios/README.md
@@ -12,6 +12,28 @@ instructions document.
 [changelog]: CHANGELOG.md
 [Configure Xcode project]: BuildInstructions.md#configure-xcode-project
 
+## Code formatting
+
+The codebase is formatted using [SwiftFormat](https://github.com/nicklockwood/SwiftFormat). Please 
+format all contributions using the latest version of formatter.
+
+```
+swiftformat ios/
+```
+
+Install the latest version of SwiftFormat via Homebrew:
+
+```
+brew install swiftformat
+```
+
+CI uses the latest version available on Homebrew to check formatting, so please keep your local 
+installation up to date, if you see it complain:
+
+```
+brew upgrade swiftformat 
+```
+
 ## Screenshots for AppStore
 
 The process of taking AppStore screenshots is automated using a UI Testing bundle and Snapshot tool,

### CODE REVIEW

**Review by AndreasShortcut — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @pronebird)

___
*[`ios/.swiftformat` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756#-N7pLfynDQ3PhCCuFZSX:-N7pLfynDQ3PhCCuFZSY:b-fpty9k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2efdb6c2d97a2b92108afa9a805f79d8091bbb98/ios/.swiftformat#L10)):*
> ```
> --indent 4
> --maxwidth 100
> --wraparguments before-first
> ```

I would much prefer
```
foo(bar: Int,
    baz: String)
```
over
```
foo(
    bar: Int,
    baz: String
)
```
since that's the way I see most Swift code written. The latter feels a bit strange to me
___
*[`ios/.swiftformat` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756#-N7pNGKo-0HsIAlq3cqV:-N7pNGKo-0HsIAlq3cqW:bhapsce) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2efdb6c2d97a2b92108afa9a805f79d8091bbb98/ios/.swiftformat#L12)):*
> ```
> --wraparguments before-first
> --wrapcollections before-first
> --wrapternary before-operators
> ```

I really like this. I wish XCode did this by default when re-indenting
___
*[`ios/.swiftformat` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756#-N7pOQjmE3be1feGf55M:-N7pOQjmE3be1feGf55N:b-c3xsfp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2efdb6c2d97a2b92108afa9a805f79d8091bbb98/ios/.swiftformat#L14)):*
> ```
> --wrapternary before-operators
> --redundanttype inferred
> --disable initCoderUnavailable, redundantReturn, unusedArguments, redundantRawValues, preferKeyPath
> ```

About `redundantRawValues`: I think we should keep this. Could you elaborate on the issue that you described with this one? 
About `redundantReturn`: Maybe we should enable this for the same reason that we don't use redundant `self`? Seems more consistent to me

**Review by pronebird — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/.swiftformat` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756#-N7pOQjmE3be1feGf55M:-N7pRadAEx_iwvN4vacJ:b-a6qs64) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2efdb6c2d97a2b92108afa9a805f79d8091bbb98/ios/.swiftformat#L14)):*
> redundantRawValues

I have one particular case which is why I dislike this rule and this is related to enums that are used to list settings keys stored in `Keychain` or `UserDefaults`, i.e:

```
enum SettingKey: String {
  case importantSetting = "importantSetting"

  // other cases...
}
```

At some point you decide that `importantSetting ` should be renamed to `notVeryImportantSetting` so you rename the case with Xcode. Now when you run the app, the old setting is no longer accessible because there was no explicit raw value assigned. The correct solution of course is to implement settiings migration from `importantSetting` to `notVeryImportantSetting`. 

For majority of cases you're unlikely to define raw values unless you really need them, hence I think this rule is not very useful in the first place.

> About redundantReturn: Maybe we should enable this for the same reason that we don't use redundant self?

This is a little subjective on my side. I don't feel like enforcing implicit `return` everywhere. Perhaps I am simply not ready to abandon explicit return yet. 

On the other side, `self` may be gone from escaping closures soon too, see this recent suggestion. I like it: https://github.com/apple/swift-evolution/blob/main/proposals/0365-implicit-self-weak-capture.md

On the other side I don't see `return` being completely removed like this because `guard` statements require it.

**Comment by AndreasShortcut** [comment]
*[`ios/.swiftformat` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756#-N7pOQjmE3be1feGf55M:-N7pbUwN1Tarxe_cOVOk:bg60vwd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2efdb6c2d97a2b92108afa9a805f79d8091bbb98/ios/.swiftformat#L14)):*

> redundantRawValues

I have one particular case which is why I dislike this rule and this is related to enums that are used to list settings keys stored in `Keychain` or `UserDefaults`, i.e:

```
enum SettingKey: String {
  case importantSetting = "importantSetting"

  // other cases...
}
```

At some point you decide that `importantSetting ` should be renamed to `notVeryImportantSetting` so you rename the case with Xcode. Now when you run the app, the old setting is no longer accessible because there was no explicit raw value assigned. The correct solution of course is to implement settiings migration from `importantSetting` to `notVeryImportantSetting`. 

For majority of cases you're unlikely to define raw values unless you really need them, hence I think this rule is not very useful in the first place.

> About redundantReturn: Maybe we should enable this for the same reason that we don't use redundant self?

This is a little subjective on my side. I don't feel like enforcing implicit `return` everywhere. Perhaps I am simply not ready to abandon explicit return yet. 

On the other side, `self` may be gone from escaping closures soon too, see this recent suggestion. I like it: https://github.com/apple/swift-evolution/blob/main/proposals/0365-implicit-self-weak-capture.md

On the other side I don't see `return` being completely removed like this because `guard` statements require it.
</blockquote></details>

That example for redundantRawValues makes a lot of sense. I agree to disable it.

I still think the reasoning for implicit return holds, i.e. we're not using `self` if it's not needed to compile, so let's not use `return` if it's not needed to compile. But it's not super important to me, so I'll leave it up to you.

**Review by pronebird — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/.swiftformat` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756#-N7pLfynDQ3PhCCuFZSX:-N7pZJjJ9dn0lBce1vg5:befzv63) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2efdb6c2d97a2b92108afa9a805f79d8091bbb98/ios/.swiftformat#L10)):*

I would much prefer
```
foo(bar: Int,
    baz: String)
```
over
```
foo(
    bar: Int,
    baz: String
)
```
since that's the way I see most Swift code written. The latter feels a bit strange to me
</blockquote></details>

It looks fine on `foo(bar, baz)` but it doesn't scale well for a mid-to-long function name because a lot of leading space is then wasted only to align the subsequent argument with the first one. 

I find it more appealing when the leading space is used to define arguments and their respective types and not accommodate whitespace and this is especially the case with arguments that accept closures (blocks). Block declarations are often pretty long.

Also note that `func foo(bar: Int, baz: String)` will not be broken on multiple lines since it fits within 100 character limit. The formatting is only applied to the code that doesn't fit within the line limit.

I did speak on increasing the char limit per line on Slack but we'll prob stick to 100 as we use it elsewhere.
___
*[`ios/.swiftformat` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756#-N7pNGKo-0HsIAlq3cqV:-N7pdGOZ7TKR6Hi3BaOG:btd51a6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2efdb6c2d97a2b92108afa9a805f79d8091bbb98/ios/.swiftformat#L12)):*

I really like this. I wish XCode did this by default when re-indenting
</blockquote></details>

I agree.

AndreasShortcut:
*[`ios/.swiftformat` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756#-N7pLfynDQ3PhCCuFZSX:-N7pfyZwBJfrGKRGZuI6:b-1acqh5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2efdb6c2d97a2b92108afa9a805f79d8091bbb98/ios/.swiftformat#L10)):*

It looks fine on `foo(bar, baz)` but it doesn't scale well for a mid-to-long function name because a lot of leading space is then wasted only to align the subsequent argument with the first one. 

I find it more appealing when the leading space is used to define arguments and their respective types and not accommodate whitespace and this is especially the case with arguments that accept closures (blocks). Block declarations are often pretty long.

Also note that `func foo(bar: Int, baz: String)` will not be broken on multiple lines since it fits within 100 character limit. The formatting is only applied to the code that doesn't fit within the line limit.

I did speak on increasing the char limit per line on Slack but we'll prob stick to 100 as we use it elsewhere.

</blockquote></details>

I see, didn't realise it was only for >100 characters. Maybe we can increase it later if we think it makes sense, since long names are quite common in Swift

AndreasShortcut (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756)* status: :shipit: complete! all files reviewed, all discussions resolved

AndreasShortcut (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3756)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3790, opened 2022-07-27
Drop individual error in relay cache

pronebird:

Git checklist:

1. Drop individual `RelayCache.Error`. Instead `RelayCache` classes re-throw either `CocoaError`, `DecodingError`, `EncodingError` or `REST.Error`. Simplifies a lot of things but also, normally we don't really do much with those errors beyond printing them, so perhaps does not matter where they come from.
2. Throw `NoRelaysSatisfyingConstraintsError` from `RelaySelector` when it cannot find relays matching a given constraint. This makes the return type non-optional. Down the road the primary purpose for this change is  integration with do-catch blocks for automatic early return.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3790)


AndreasShortcut (commented):
> Reviewed 10 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3790)* status: 10 of 11 files reviewed, all discussions resolved

PR #3811, opened 2022-08-04
Add out of time view

AndreasShortcut:
This PR introduces a new out of time view, which is shown when logged into an account with no time left. From this view, it's possible to buy more time via in-app purchase, redeem vouchers (not yet implemented!), and restore previous purchases. 

This change replaces the redirection to account settings, providing a better experience, which is especially important since it's the first view shown after creating a new account in the app.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)

### DIFF
--- a/ios/CHANGELOG.md
+++ b/ios/CHANGELOG.md
@@ -30,8 +30,12 @@ Line wrap the file at 100 chars.                                              Th
 - Add revoked device view displayed when the app detects that device is no longer registered on
   backend.
 - Add ability to manage registered devices if too many devices detected during log-in.
-- Add intents: start VPN, stop VPN, reconnect VPN (acts as start VPN when the tunnel is down, 
+- Add intents: start VPN, stop VPN, reconnect VPN (acts as start VPN when the tunnel is down,
   otherwise picks new relay).
+
+### Changed
+- When logged into an account with no time left, a new view is shown instead of account settings,
+with the option to buy more time.
 
 ### Fixed
 - Improve random port distribution. Should be less biased towards port 53.
@@ -49,8 +53,8 @@ Line wrap the file at 100 chars.                                              Th
 - Delete leftover settings in Keychain during login. WireGuard keys will be removed from
   server too if old settings can be read. This is usually the case when uninstalling the app and
   then reinstalling it without logging out first.
-- Validate account token before charging user (in-app purchases). Safeguards from trying to add 
-  credits on accounts that no longer exist on our backend. Usually the case with newly created 
+- Validate account token before charging user (in-app purchases). Safeguards from trying to add
+  credits on accounts that no longer exist on our backend. Usually the case with newly created
   accounts that went stale.
 
 
@@ -62,13 +66,13 @@ Line wrap the file at 100 chars.                                              Th
 ### Fixed
 - Fix crash occurring after completing in-app purchase.
 - Fix error when changing relays while in airplane mode.
-- Prevent key rotation from clogging the server key list by storing the next key and reusing it 
+- Prevent key rotation from clogging the server key list by storing the next key and reusing it
   until receiving the successful response from Mullvad API. Add up to three retry attempts.
 
 ### Changed
 - Increase hit area of settings (cog) button.
 - Update launch screen.
-- Never use DNS to talk to Mullvad API. Instead use the list of IP addresses bundled with the app 
+- Never use DNS to talk to Mullvad API. Instead use the list of IP addresses bundled with the app
   and update it periodically.
 
 
@@ -86,7 +90,7 @@ Line wrap the file at 100 chars.                                              Th
 - Drop leading replacement characters (`\u{FFFD}`) when decoding UTF-8 from a part of log file.
 
 ### Security
-- Move REST API networking from the packet tunnel process to the main process to prevent leaking 
+- Move REST API networking from the packet tunnel process to the main process to prevent leaking
   traffic outside of the tunnel.
 
 
@@ -106,7 +110,7 @@ Line wrap the file at 100 chars.                                              Th
 - Enable option to "Select all" when viewing app logs.
 - Split view interface for iPad.
 - Add interactive map.
-- Reduce network traffic consumption by leveraging HTTP caching via ETag HTTP header to avoid 
+- Reduce network traffic consumption by leveraging HTTP caching via ETag HTTP header to avoid
   re-downloading the relay list if it hasn't changed.
 - Pin root SSL certificates.
 - Add option to use Mullvad's ad-blocking DNS servers.
@@ -134,7 +138,7 @@ Line wrap the file at 100 chars.                                              Th
 
 ## [2020.5] - 2020-11-04
 ### Fixed
-- Fix regression where "Internal error" was displayed instead of server error (i.e too many 
+- Fix regression where "Internal error" was displayed instead of server error (i.e too many
   WireGuard keys)
 
 
@@ -164,11 +168,11 @@ Line wrap the file at 100 chars.                                              Th
 - Add automatic key rotation every 4 days.
 
 ### Fixed
-- Fix relay selection for country wide constraints by respecting the `include_in_country` 
+- Fix relay selection for country wide constraints by respecting the `include_in_country`
   parameter.
 - Fix defect when manually regenerating the private key from Settings would automatically connect
   the tunnel.
-- Properly format date intervals close to 1 day or less than 1 minute. Enforce intervals between 1 
+- Properly format date intervals close to 1 day or less than 1 minute. Enforce intervals between 1
   and 90 days to always be displayed in days quantity.
 - Fix a number of errors in DNS64 resolution and IPv6 support.
 - Update the tunnel state when the app returns from suspended state.
@@ -184,8 +188,8 @@ Line wrap the file at 100 chars.                                              Th
 
 ### Added
 - Format account number in groups of 4 digits separated by whitespace on login screen.
-- Enable on-demand VPN with a single rule to always connect the tunnel when on Wi-Fi or cellular. 
-  Automatically disable on-demand VPN when manually disconnecting the tunnel from GUI to prevent the 
+- Enable on-demand VPN with a single rule to always connect the tunnel when on Wi-Fi or cellular.
+  Automatically disable on-demand VPN when manually disconnecting the tunnel from GUI to prevent the
   tunnel from coming back up.
 
 
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -307,7 +307,11 @@
 		58FEEB46260A028D00A621A8 /* GeoJSON.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58FEEB45260A028D00A621A8 /* GeoJSON.swift */; };
 		58FEEB58260B662E00A621A8 /* AutomaticKeyboardResponder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58FEEB57260B662E00A621A8 /* AutomaticKeyboardResponder.swift */; };
 		58FF2C03281BDE02009EF542 /* SettingsManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58FF2C02281BDE02009EF542 /* SettingsManager.swift */; };
+		E1187ABC289BBB850024E748 /* OutOfTimeViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1187ABA289BBB850024E748 /* OutOfTimeViewController.swift */; };
+		E1187ABD289BBB850024E748 /* OutOfTimeContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1187ABB289BBB850024E748 /* OutOfTimeContentView.swift */; };
+		E1187ABF289BE76F0024E748 /* RESTCreateApplePaymentResponse+Localization.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1187ABE289BE76F0024E748 /* RESTCreateApplePaymentResponse+Localization.swift */; };
 		E158B360285381C60002F069 /* StringFormatter.swift in Sources */ = {isa = PBXBuildFile; fileRef = E158B35F285381C60002F069 /* StringFormatter.swift */; };
+		E1FD0DF528AA7CE400299DB4 /* StatusActivityView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1FD0DF428AA7CE400299DB4 /* StatusActivityView.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -594,7 +598,11 @@
 		58FEEB45260A028D00A621A8 /* GeoJSON.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = GeoJSON.swift; sourceTree = "<group>"; };
 		58FEEB57260B662E00A621A8 /* AutomaticKeyboardResponder.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AutomaticKeyboardResponder.swift; sourceTree = "<group>"; };
 		58FF2C02281BDE02009EF542 /* SettingsManager.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SettingsManager.swift; sourceTree = "<group>"; };
+		E1187ABA289BBB850024E748 /* OutOfTimeViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = OutOfTimeViewController.swift; sourceTree = "<group>"; };
+		E1187ABB289BBB850024E748 /* OutOfTimeContentView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = OutOfTimeContentView.swift; sourceTree = "<group>"; };
+		E1187ABE289BE76F0024E748 /* RESTCreateApplePaymentResponse+Localization.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "RESTCreateApplePaymentResponse+Localization.swift"; sourceTree = "<group>"; };
 		E158B35F285381C60002F069 /* StringFormatter.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StringFormatter.swift; sourceTree = "<group>"; };
+		E1FD0DF428AA7CE400299DB4 /* StatusActivityView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StatusActivityView.swift; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
 /* Begin PBXFrameworksBuildPhase section */
@@ -774,6 +782,7 @@
 				58554F72280AFA5A00013055 /* RESTAuthenticationProxy.swift */,
 				58B5A898280AB0D7009FDE99 /* RESTAuthorization.swift */,
 				585DA88926B027A300B8C587 /* RESTCoding.swift */,
+				E1187ABE289BE76F0024E748 /* RESTCreateApplePaymentResponse+Localization.swift */,
 				588BCF23280FE43D009ADCEC /* RESTDevicesProxy.swift */,
 				585DA88626B0277200B8C587 /* RESTError.swift */,
 				58095C522760EEC700890776 /* RESTNetworkOperation.swift */,
@@ -920,6 +929,8 @@
 				5871FB9F254C26BF0051A0A4 /* NSRegularExpression+IPAddress.swift */,
 				58CC40EE24A601900019D96E /* ObserverList.swift */,
 				580EE1FF24B3218800F9D8A1 /* Operations */,
+				E1187ABB289BBB850024E748 /* OutOfTimeContentView.swift */,
+				E1187ABA289BBB850024E748 /* OutOfTimeViewController.swift */,
 				584D26C3270C855A004EA533 /* PreferencesDataSource.swift */,
 				587EB6732714520600123C75 /* PreferencesDataSourceDelegate.swift */,
 				58ACF6482655365700ACE4B7 /* PreferencesViewController.swift */,
@@ -955,6 +966,7 @@
 				587A01FB23F1F0BE00B68763 /* SimulatorTunnelProviderHost.swift */,
 				58FD5BEF24238EB300112C88 /* SKProduct+Formatting.swift */,
 				58F19E34228C15BA00C7710B /* SpinnerActivityIndicatorView.swift */,
+				E1FD0DF428AA7CE400299DB4 /* StatusActivityView.swift */,
 				58EF581025D69DB400AEBA94 /* StatusImageView.swift */,
 				5807E2BF2432038B00F5FF30 /* String+Split.swift */,
 				E158B35F285381C60002F069 /* StringFormatter.swift */,
@@ -1350,7 +1362,9 @@
 				58B3F30F2742708B00A2DD38 /* HeaderBarButton.swift in Sources */,
 				584789E026529D72000E45FB /* SSLPinningURLSessionDelegate.swift in Sources */,
 				58161C9C28352F850028ECFD /* MigrateSettingsOperation.swift in Sources */,
+				E1187ABF289BE76F0024E748 /* RESTCreateApplePaymentResponse+Localization.swift in Sources */,
 				58ACF6492655365700ACE4B7 /* PreferencesViewController.swift in Sources */,
+				E1FD0DF528AA7CE400299DB4 /* StatusActivityView.swift in Sources */,
 				58968FAE28743E2000B799DC /* TunnelInteractor.swift in Sources */,
 				588D2FE3248AC27F00E313F7 /* AsyncOperation.swift in Sources */,
 				5820675026E6514100655B05 /* HTTP.swift in Sources */,
@@ -1368,6 +1382,7 @@
 				58FEEB58260B662E00A621A8 /* AutomaticKeyboardResponder.swift in Sources */,
 				5846227326E22A160035F7C2 /* AppStorePaymentObserver.swift in Sources */,
 				58F2E146276A2C9900A79513 /* StopTunnelOperation.swift in Sources */,
+				E1187ABC289BBB850024E748 /* OutOfTimeViewController.swift in Sources */,
 				58DF5B762852108E00E92647 /* InputInjectionBuilder.swift in Sources */,
 				5856D13727450A8A00DFD627 /* UIImage+TintColor.swift in Sources */,
 				58CB0EE024B86751001EF0D8 /* RESTAPIProxy.swift in Sources */,
@@ -1393,6 +1408,7 @@
 				588527B4276B4F2F00BAA373 /* SetAccountOperation.swift in Sources */,
 				585CA70F25F8C44600B47C62 /* UIMetrics.swift in Sources */,
 				58F97A1B280EEBC00050C2FC /* RESTProxyFactory.swift in Sources */,
+				E1187ABD289BBB850024E748 /* OutOfTimeContentView.swift in Sources */,
 				58095C592762155700890776 /* RESTRetryStrategy.swift in Sources */,
 				5840250422B11AB700E4CFEC /* MullvadEndpoint.swift in Sources */,
 				58CC40EF24A601900019D96E /* ObserverList.swift in Sources */,
--- a/ios/MullvadVPN/AccountViewController.swift
+++ b/ios/MullvadVPN/AccountViewController.swift
@@ -422,67 +422,6 @@ class AccountViewController: UIViewController, AppStorePaymentObserver, TunnelOb
     }
 }
 
-private extension REST.CreateApplePaymentResponse {
-    enum Context {
-        case purchase
-        case restoration
-    }
-
-    func alertTitle(context: Context) -> String {
-        switch context {
-        case .purchase:
-            return NSLocalizedString(
-                "TIME_ADDED_ALERT_SUCCESS_TITLE",
-                tableName: "Account",
-                value: "Thanks for your purchase",
-                comment: ""
-            )
-        case .restoration:
-            return NSLocalizedString(
-                "RESTORE_PURCHASES_ALERT_TITLE",
-                tableName: "Account",
-                value: "Restore purchases",
-                comment: ""
-            )
-        }
-    }
-
-    func alertMessage(context: Context) -> String {
-        switch context {
-        case .purchase:
-            return String(
-                format: NSLocalizedString(
-                    "TIME_ADDED_ALERT_SUCCESS_MESSAGE",
-                    tableName: "Account",
-                    value: "%@ have been added to your account",
-                    comment: ""
-                ),
-                formattedTimeAdded ?? ""
-            )
-        case .restoration:
-            switch self {
-            case .noTimeAdded:
-                return NSLocalizedString(
-                    "RESTORE_PURCHASES_ALERT_NO_TIME_ADDED_MESSAGE",
-                    tableName: "Account",
-                    value: "Your previous purchases have already been added to this account.",
-                    comment: ""
-                )
-            case .timeAdded:
-                return String(
-                    format: NSLocalizedString(
-                        "RESTORE_PURCHASES_ALERT_TIME_ADDED_MESSAGE",
-                        tableName: "Account",
-                        value: "%@ have been added to your account",
-                        comment: ""
-                    ),
-                    formattedTimeAdded ?? ""
-                )
-            }
-        }
-    }
-}
-
 private extension AccountViewController {
     enum PaymentState: Equatable {
         case none
--- a/ios/MullvadVPN/HeaderBarView.swift
+++ b/ios/MullvadVPN/HeaderBarView.swift
@@ -31,9 +31,15 @@ class HeaderBarView: UIView {
     class func makeSettingsButton() -> HeaderBarButton {
         let settingsImage = UIImage(named: "IconSettings")?
             .backport_withTintColor(UIColor.HeaderBar.buttonColor, renderingMode: .alwaysOriginal)
+        let disabledSettingsImage = UIImage(named: "IconSettings")?
+            .backport_withTintColor(
+                UIColor.HeaderBar.disabledButtonColor,
+                renderingMode: .alwaysOriginal
+            )
 
         let settingsButton = HeaderBarButton(type: .system)
         settingsButton.setImage(settingsImage, for: .normal)
+        settingsButton.setImage(disabledSettingsImage, for: .disabled)
         settingsButton.translatesAutoresizingMaskIntoConstraints = false
         settingsButton.accessibilityIdentifier = "SettingsButton"
         settingsButton.accessibilityLabel = NSLocalizedString(
--- a/ios/MullvadVPN/OutOfTimeContentView.swift
+++ b/ios/MullvadVPN/OutOfTimeContentView.swift
@@ -1,0 +1,147 @@
+//
+//  OutOfTimeContentView.swift
+//  MullvadVPN
+//
+//  Created by Andreas Lif on 2022-07-26.
+//  Copyright © 2022 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+class OutOfTimeContentView: UIView {
+    let statusActivityView: StatusActivityView = {
+        let statusActivityView = StatusActivityView(state: .failure)
+        statusActivityView.translatesAutoresizingMaskIntoConstraints = false
+        return statusActivityView
+    }()
+
+    private lazy var titleLabel: UILabel = {
+        let label = UILabel()
+        label.text = NSLocalizedString(
+            "OUT_OF_TIME_TITLE",
+            tableName: "OutOfTime",
+            value: "Out of time",
+            comment: ""
+        )
+        label.font = UIFont.systemFont(ofSize: 32)
+        label.textColor = .white
+        return label
+    }()
+
+    lazy var bodyLabel: UILabel = {
+        let label = UILabel()
+        label.text = NSLocalizedString(
+            "OUT_OF_TIME_BODY",
+            tableName: "OutOfTime",
+            value: "You have no more VPN time left on this account. Either buy credit on our website or redeem a voucher.",
+            comment: ""
+        )
+        label.font = UIFont.systemFont(ofSize: 17)
+        label.textColor = .white
+        label.numberOfLines = 0
+        return label
+    }()
+
+    lazy var disconnectButton: AppButton = {
+        let button = AppButton(style: .danger)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.alpha = 0
+        let localizedString = NSLocalizedString(
+            "OUT_OF_TIME_DISCONNECT_BUTTON",
+            tableName: "OutOfTime",
+            value: "Disconnect",
+            comment: ""
+        )
+        button.setTitle(localizedString, for: .normal)
+        return button
+    }()
+
+    lazy var purchaseButton: InAppPurchaseButton = {
+        let button = InAppPurchaseButton()
+        button.translatesAutoresizingMaskIntoConstraints = false
+        let localizedString = NSLocalizedString(
+            "OUT_OF_TIME_PURCHASE_BUTTON",
+            tableName: "OutOfTime",
+            value: "Add 30 days time",
+            comment: ""
+        )
+        button.setTitle(localizedString, for: .normal)
+        return button
+    }()
+
+    lazy var restoreButton: AppButton = {
+        let button = AppButton(style: .default)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.setTitle(NSLocalizedString(
+            "RESTORE_PURCHASES_BUTTON_TITLE",
+            tableName: "OutOfTime",
+            value: "Restore purchases",
+            comment: ""
+        ), for: .normal)
+        return button
+    }()
+
+    private lazy var topStackView: UIStackView = {
+        let stackView = UIStackView(arrangedSubviews: [statusActivityView, titleLabel, bodyLabel])
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        stackView.axis = .vertical
+        stackView.spacing = UIMetrics.sectionSpacing
+        return stackView
+    }()
+
+    private lazy var bottomStackView: UIStackView = {
+        let stackView = UIStackView(
+            arrangedSubviews: [disconnectButton, purchaseButton, restoreButton]
+        )
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        stackView.axis = .vertical
+        stackView.spacing = UIMetrics.sectionSpacing
+        return stackView
+    }()
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        translatesAutoresizingMaskIntoConstraints = false
+        backgroundColor = .secondaryColor
+        layoutMargins = UIMetrics.contentLayoutMargins
+        setUpSubviews()
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+// MARK: - Private Functions
+
+private extension OutOfTimeContentView {
+    func setUpSubviews() {
+        addSubview(topStackView)
+        addSubview(bottomStackView)
+        configureConstraints()
+    }
+
+    func configureConstraints() {
+        NSLayoutConstraint.activate([
+            topStackView.centerYAnchor.constraint(equalTo: centerYAnchor, constant: -20),
+
+            topStackView.leadingAnchor.constraint(
+                equalTo: layoutMarginsGuide.leadingAnchor
+            ),
+            topStackView.trailingAnchor.constraint(
+                equalTo: layoutMarginsGuide.trailingAnchor
+            ),
+
+            bottomStackView.leadingAnchor.constraint(
+                equalTo: layoutMarginsGuide.leadingAnchor
+            ),
+            bottomStackView.trailingAnchor.constraint(
+                equalTo: layoutMarginsGuide.trailingAnchor
+            ),
+            bottomStackView.bottomAnchor.constraint(
+                equalTo: layoutMarginsGuide.bottomAnchor
+            ),
+        ])
+    }
+}
--- a/ios/MullvadVPN/OutOfTimeViewController.swift
+++ b/ios/MullvadVPN/OutOfTimeViewController.swift
@@ -1,0 +1,453 @@
+//
+//  OutOfTimeViewController.swift
+//  MullvadVPN
+//
+//  Created by Andreas Lif on 2022-07-25.
+//  Copyright © 2022 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+import StoreKit
+import UIKit
+
+class OutOfTimeViewController: UIViewController {
+    weak var delegate: SettingsButtonInteractionDelegate?
+
+    private var productState: ProductState = .none
+    private var paymentState: PaymentState = .none
+
+    private let alertPresenter = AlertPresenter()
+
+    private lazy var contentView = OutOfTimeContentView()
+
+    override var preferredStatusBarStyle: UIStatusBarStyle {
+        return .lightContent
+    }
+
+    private var tunnelState: TunnelState = .disconnected {
+        didSet {
+            setNeedsHeaderBarStyleAppearanceUpdate()
+            applyViewState(animated: true)
+        }
+    }
+
+    override func viewDidLoad() {
+        setUpContentView()
+        setUpButtonTargets()
+        setUpInAppPurchases()
+        addObservers()
+        tunnelState = TunnelManager.shared.tunnelStatus.state
+    }
+}
+
+// MARK: - Private Functions
+
+private extension OutOfTimeViewController {
+    func setUpContentView() {
+        view.addSubview(contentView)
+
+        NSLayoutConstraint.activate([
+            contentView.topAnchor.constraint(equalTo: view.topAnchor),
+            contentView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            contentView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            contentView.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+        ])
+    }
+
+    func setUpButtonTargets() {
+        contentView.disconnectButton.addTarget(
+            self,
+            action: #selector(handleDisconnect(_:)),
+            for: .touchUpInside
+        )
+
+        contentView.purchaseButton.addTarget(
+            self,
+            action: #selector(doPurchase),
+            for: .touchUpInside
+        )
+        contentView.restoreButton.addTarget(
+            self,
+            action: #selector(restorePurchases),
+            for: .touchUpInside
+        )
+    }
+
+    @objc func handleDisconnect(_ sender: Any) {
+        TunnelManager.shared.stopTunnel()
+    }
+
+    func addObservers() {
+        AppStorePaymentManager.shared.addPaymentObserver(self)
+        TunnelManager.shared.addObserver(self)
+    }
+
+    func setEnableUserInteraction(_ enableUserInteraction: Bool) {
+        [contentView.purchaseButton, contentView.restoreButton]
+            .forEach { button in
+                button?.isEnabled = enableUserInteraction
+            }
+
+        view.isUserInteractionEnabled = enableUserInteraction
+    }
+
+    func bodyText(for tunnelState: TunnelState) -> String {
+        if tunnelState.isSecured {
+            return NSLocalizedString(
+                "OUT_OF_TIME_BODY_CONNECTED",
+                tableName: "OutOfTime",
+                value: "You have no more VPN time left on this account. To add more, you will need to disconnect and access the Internet with an unsecure connection.",
+                comment: ""
+            )
+        } else {
+            return NSLocalizedString(
+                "OUT_OF_TIME_BODY_DISCONNECTED",
+                tableName: "OutOfTime",
+                value: "You have no more VPN time left on this account. Either buy credit on our website or redeem a voucher.",
+                comment: ""
+            )
+        }
+    }
+}
+
+// MARK: - In App Purchases
+
+private extension OutOfTimeViewController {
+    func setUpInAppPurchases() {
+        if AppStorePaymentManager.canMakePayments {
+            requestStoreProducts()
+        } else {
+            setProductState(.cannotMakePurchases, animated: false)
+        }
+    }
+
+    func requestStoreProducts() {
+        let productKind = AppStoreSubscription.thirtyDays
+
+        setProductState(.fetching(productKind), animated: true)
+
+        _ = AppStorePaymentManager.shared
+            .requestProducts(with: [productKind]) { [weak self] completion in
+                let productState: ProductState = completion.value?.products.first
+                    .map { .received($0) } ?? .failed
+
+                self?.setProductState(productState, animated: true)
+            }
+    }
+
+    func setPaymentState(_ newState: PaymentState, animated: Bool) {
+        paymentState = newState
+
+        applyViewState(animated: animated)
+    }
+
+    func setProductState(_ newState: ProductState, animated: Bool) {
+        productState = newState
+
+        applyViewState(animated: animated)
+    }
+
+    func applyViewState(animated: Bool) {
+        let isInteractionEnabled = paymentState.allowsViewInteraction
+        let purchaseButton = contentView.purchaseButton
+
+        let isOutOfTime = TunnelManager.shared.deviceState.accountData
+            .map { $0.expiry < Date() } ?? false
+
+        let actions = { [weak self] in
+            guard let self = self else { return }
+
+            purchaseButton.setTitle(self.productState.purchaseButtonTitle, for: .normal)
+            self.contentView.purchaseButton.isLoading = self.productState.isFetching
+
+            purchaseButton.isEnabled = self.productState.isReceived && isInteractionEnabled && !self
+                .tunnelState.isSecured
+            self.contentView.restoreButton.isEnabled = isInteractionEnabled
+            self.contentView.disconnectButton.isEnabled = self.tunnelState.isSecured
+            self.contentView.disconnectButton.alpha = self.tunnelState.isSecured ? 1 : 0
+            self.contentView.bodyLabel.text = self.bodyText(for: self.tunnelState)
+
+            if !isInteractionEnabled {
+                self.contentView.statusActivityView.state = .activity
+            } else {
+                self.contentView.statusActivityView.state = isOutOfTime ? .failure : .success
+            }
+
+            self.delegate?.viewController(
+                self,
+                didRequestSettingsButtonEnabled: isInteractionEnabled
+            )
+        }
+        if animated {
+            UIView.animate(withDuration: 0.25, animations: actions)
+        } else {
+            actions()
+        }
+
+        view.isUserInteractionEnabled = isInteractionEnabled
+        if #available(iOS 13.0, *) {
+            isModalInPresentation = !isInteractionEnabled
+        }
+        navigationItem.setHidesBackButton(!isInteractionEnabled, animated: animated)
+    }
+
+    @objc private func doPurchase() {
+        guard case let .received(product) = productState,
+              let accountData = TunnelManager.shared.deviceState.accountData
+        else {
+            return
+        }
+
+        let payment = SKPayment(product: product)
+        AppStorePaymentManager.shared.addPayment(payment, for: accountData.number)
+
+        setPaymentState(.makingPayment(payment), animated: true)
+    }
+
+    @objc func restorePurchases() {
+        guard let accountData = TunnelManager.shared.deviceState.accountData else {
+            return
+        }
+
+        setPaymentState(.restoringPurchases, animated: true)
+
+        _ = AppStorePaymentManager.shared.restorePurchases(for: accountData.number) { completion in
+            switch completion {
+            case let .success(response):
+                self.showAlertIfNoTimeAdded(with: response, context: .restoration)
+            case let .failure(error):
+                self.showRestorePurchasesErrorAlert(error: error)
+
+            case .cancelled:
+                break
+            }
+
+            self.setPaymentState(.none, animated: true)
+        }
+    }
+
+    private func showAlertIfNoTimeAdded(
+        with response: REST.CreateApplePaymentResponse,
+        context: REST.CreateApplePaymentResponse.Context
+    ) {
+        guard case .noTimeAdded = response else { return }
+
+        let alertController = UIAlertController(
+            title: response.alertTitle(context: context),
+            message: response.alertMessage(context: context),
+            preferredStyle: .alert
+        )
+        alertController.addAction(
+            UIAlertAction(
+                title: NSLocalizedString(
+                    "TIME_ADDED_ALERT_OK_ACTION",
+                    tableName: "OutOfTime",
+                    value: "OK",
+                    comment: ""
+                ),
+                style: .cancel
+            )
+        )
+
+        alertPresenter.enqueue(alertController, presentingController: self)
+    }
+
+    func showRestorePurchasesErrorAlert(error: AppStorePaymentManager.Error) {
+        let alertController = UIAlertController(
+            title: NSLocalizedString(
+                "RESTORE_PURCHASES_FAILURE_ALERT_TITLE",
+                tableName: "OutOfTime",
+                value: "Cannot restore purchases",
+                comment: ""
+            ),
+            message: error.errorChainDescription,
+            preferredStyle: .alert
+        )
+        alertController.addAction(
+            UIAlertAction(title: NSLocalizedString(
+                "RESTORE_PURCHASES_FAILURE_ALERT_OK_ACTION",
+                tableName: "OutOfTime",
+                value: "OK",
+                comment: ""
+            ), style: .cancel)
+        )
+        alertPresenter.enqueue(alertController, presentingController: self)
+    }
+
+    func showPaymentErrorAlert(error: AppStorePaymentManager.Error) {
+        let alertController = UIAlertController(
+            title: NSLocalizedString(
+                "CANNOT_COMPLETE_PURCHASE_ALERT_TITLE",
+                tableName: "OutOfTime",
+                value: "Cannot complete the purchase",
+                comment: ""
+            ),
+            message: error.errorChainDescription,
+            preferredStyle: .alert
+        )
+
+        alertController.addAction(
+            UIAlertAction(
+                title: NSLocalizedString(
+                    "CANNOT_COMPLETE_PURCHASE_ALERT_OK_ACTION",
+                    tableName: "OutOfTime",
+                    value: "OK",
+                    comment: ""
+                ), style: .cancel
+            )
+        )
+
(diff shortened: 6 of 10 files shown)


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: all files reviewed, 9 unresolved discussions (waiting on @AndreasShortcut)

___
*[`ios/CHANGELOG.md` line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wW-OyDPSYGbISNHml:-N8wW-OyDPSYGbISNHmm:bssuag) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/CHANGELOG.md#L37)):*
> ```Markdown
> 
> ### Changed
> - When logged into an account with no time left, a new view is shown instead of account settings, with the option to buy more time.
> ```

Please break the long lines at 100 chars as described in the `Format` section of this document.
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wmbPcE20BtIkGn1q6:-N8wmbPd5n2VH91pKs5k:buod54l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/OutOfTimeViewController.swift#L17)):*
> ```Swift
> }
> 
> class OutOfTimeViewController: UIViewController {
> ```

Missing `preferredStatusBarStyle` override with light content.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 581 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8x285_8xamQhNnEtoN:-N8x285_8xamQhNnEtoO:b-tecip5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L581)):*
> ```Swift
>         lastLoginAction = nil
> 
>         setUpOutOfTimeTimer()
> ```

We could call it inside of `tunnelManager(:didUpdateDeviceState:)`. This should keep the timer up to date as well, for example if user added time on account, the timer would be rescheduled too or if device was revoked, the timer would be simply invalidated.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 617 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8x2PWr5J2u5XJ8QyJr:-N8x2PWr5J2u5XJ8QyJs:b-6yt2fc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L617)):*
> ```Swift
>               accountData.expiry > Date() else { return }
> 
>         outOfTimeTimer?.invalidate()
> ```

Please invalidate the timer when in `.loggedOut` or `.revoked` states because the app can move from `loggedIn` to `revoked` state for example and we don't want to go out of time in that case since the user is obliged to log out the revoked device. 

Also note that you can avoid the `guard case` hoop and use `deviceState.deviceData` instead, which should return `nil` in all cases except when `.loggedIn`.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 621 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8x-GW8CNHmLSu4Tomi:-N8x-GW8CNHmLSu4Tomj:b-or0xs5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L621)):*
> ```Swift
>             fireAt: accountData.expiry,
>             interval: 0,
>             target: self,
> ```

`NSTimer` retains its `target` strongly, which is a problem because `SceneDelegate` will never be released as the timer needs to be explicitly invalidated. 

One way to avoid that is to use a block-based `Timer` with weak reference to `self` and then invalidate the timer in `deinit`. 

The other way is to use `DispatchSource.makeTimer()` which will automatically invalidate itself once released, but API is less `NSDate` friendly so you'd have to do something like `.now() + date.timeIntervalSinceNow` when scheduling the timer.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 626 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8x-u-W6097xJ7x-912:-N8x-u-W6097xJ7x-913:b-fog0mo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L626)):*
> ```Swift
>             repeats: false
>         )
>         RunLoop.main.add(outOfTimeTimer ?? Timer(), forMode: .common)
> ```

Basically compiler will likely turn this into something like that:

```swift
if let outOfTimeTimer = outOfTimeTimer {
  RunLoop.main.add(outOfTimeTimer, forMode: .common)
} else {
  RunLoop.main.add(Timer(), forMode: .common)
}
```

The `else` branch makes no sense in our case, it's simply creating a leak which is not desired. This unconfigured `Timer()` would never be released. Using nil coalescing operator is good for sensible fallbacks but here it's used to keep compiler happy while creating undesired potential code paths.

What programmer would normally do here is assert that `outOfTimeTimer` cannot be `nil` here by force unwrapping it:

```swift
RunLoop.main.add(outOfTimeTimer!, forMode: .common)
```

But we don't want unsafe code in our case, so the simplest solution is to use a local variable and then assign the timer back to member variable:

```swift
let timer = Timer(...)
RunLoop.main.add(timer, forMode: .common)

self.outOfTimeTimer?.invalidate()
self.outOfTimeTimer = timer
```



_Code quote:_
```Swift
 RunLoop.main.add(outOfTimeTimer ?? Timer(), forMode: .common)
```
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 704 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8woqGS7vqshtg3H_DP:-N8woqGS7vqshtg3H_DQ:bjdm6jj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L704)):*
> ```Swift
> extension SceneDelegate: OutOfTimeViewControllerDelegate {
>     func outOfTimeViewControllerDidAddTime(_ controller: OutOfTimeViewController) {
>         dismissOutOfTimeController(controller)
> ```

Time can be added in three different ways:

1. Account view
2. Out of time view
3. Website. (which is right now we update account data each time user opens settings)

In all those cases, tunnel manager will call `didUpdateDeviceState`, with new expiry stored in account data. So ideally we should be able to remove the out of time view regardless of how the time was added to account.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 876 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wYMhF91c-OYUC4vDQ:-N8wYMhF91c-OYUC4vDR:b-6k1585) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L876)):*
> ```Swift
>     }
> 
>     func tunnelManager(_: TunnelManager, didFailWithError _: Error) {
> ```

Swiftformat supports removing unused arguments via `unusedArguments` rule, which we have disabled because I find it counter-productive and ugly. (See https://github.com/mullvad/mullvadvpn-app/pull/3756)

Xcode auto-completes delegates with variable names, and so then you go and remove them. One day you need them back, and you have to fill the variable names back in. This generates extra git diff and creates a lot of busy work. 

There are pros and cons in either approach and it's already been discussed in detail in: https://github.com/nicklockwood/SwiftFormat/issues/83. 

In short I am in the camp of leaving those variables alone but I don't mind when it comes to closures. So please leave them be and focus on implementation instead.

More info on the swiftformat rule:
```
swiftformat --ruleinfo unusedArguments
```
___
*[`ios/MullvadVPN/REST/RESTCreateApplePaymentResponse.swift` line 2 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wf9S-CROkAKBzTynu:-N8wf9S-CROkAKBzTynv:b29cgdg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/REST/RESTCreateApplePaymentResponse.swift#L2)):*
> ```Swift
> //
> //  RESTCreateApplePaymentResponse.swift
> ```

Maybe we should rename it to `RESTCreateApplePaymentResponse+Localization.swift` (ugh long name!) and set localisation table to `REST` instead of `Account`. Although strictly speaking, this belong in UI/Controllers helpers.

AndreasShortcut (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 2 of 7 files reviewed, 9 unresolved discussions (waiting on @pronebird)

___
*[`ios/CHANGELOG.md` line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wW-OyDPSYGbISNHml:-N8x8GiA8csEG9BUZ31l:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/CHANGELOG.md#L37)):*

Please break the long lines at 100 chars as described in the `Format` section of this document.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wmbPcE20BtIkGn1q6:-N8x9XOO84aiE-hBJcT4:b-im78w4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/OutOfTimeViewController.swift#L17)):*

Missing `preferredStatusBarStyle` override with light content.
</blockquote></details>

Done. But is there any view where we don't want it to be `.lightContent`? We could just set this in info.plist and not have to deal with it in each VC
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 581 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8x285_8xamQhNnEtoN:-N8xGC2M7wOulLFAHKn1:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L581)):*

We could call it inside of `tunnelManager(:didUpdateDeviceState:)`. This should keep the timer up to date as well, for example if user added time on account, the timer would be rescheduled too or if device was revoked, the timer would be simply invalidated.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 617 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8x2PWr5J2u5XJ8QyJr:-N8xGNp_8BHHfJ6V0p6T:bensqzr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L617)):*

Please invalidate the timer when in `.loggedOut` or `.revoked` states because the app can move from `loggedIn` to `revoked` state for example and we don't want to go out of time in that case since the user is obliged to log out the revoked device. 

Also note that you can avoid the `guard case` hoop and use `deviceState.deviceData` instead, which should return `nil` in all cases except when `.loggedIn`.
</blockquote></details>

Moved `.invalidate()`call so it's called for `.loggedOut` and `.revoked` as well.

About the guard, I think it's nicer to keep it for readability. This way we don't have to rely on the reader knowing that `deviceData != nil` means that we're logged in. What do you think?
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 621 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8x-GW8CNHmLSu4Tomi:-N8xcy5X9h7SY_16favz:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L621)):*

`NSTimer` retains its `target` strongly, which is a problem because `SceneDelegate` will never be released as the timer needs to be explicitly invalidated. 

One way to avoid that is to use a block-based `Timer` with weak reference to `self` and then invalidate the timer in `deinit`. 

The other way is to use `DispatchSource.makeTimer()` which will automatically invalidate itself once released, but API is less `NSDate` friendly so you'd have to do something like `.now() + date.timeIntervalSinceNow` when scheduling the timer.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 626 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8x-u-W6097xJ7x-912:-N8xmbA856-UaFSdK_H0:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L626)):*

Basically compiler will likely turn this into something like that:

```swift
if let outOfTimeTimer = outOfTimeTimer {
  RunLoop.main.add(outOfTimeTimer, forMode: .common)
} else {
  RunLoop.main.add(Timer(), forMode: .common)
}
```

The `else` branch makes no sense in our case, it's simply creating a leak which is not desired. This unconfigured `Timer()` would never be released. Using nil coalescing operator is good for sensible fallbacks but here it's used to keep compiler happy while creating undesired potential code paths.

What programmer would normally do here is assert that `outOfTimeTimer` cannot be `nil` here by force unwrapping it:

```swift
RunLoop.main.add(outOfTimeTimer!, forMode: .common)
```

But we don't want unsafe code in our case, so the simplest solution is to use a local variable and then assign the timer back to member variable:

```swift
let timer = Timer(...)
RunLoop.main.add(timer, forMode: .common)

self.outOfTimeTimer?.invalidate()
self.outOfTimeTimer = timer
```
</blockquote></details>

Done
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 704 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8woqGS7vqshtg3H_DP:-N8xmcCm0Hkc1taEmYBE:b-icwiyi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L704)):*

Time can be added in three different ways:

1. Account view
2. Out of time view
3. Website. (which is right now we update account data each time user opens settings)

In all those cases, tunnel manager will call `didUpdateDeviceState`, with new expiry stored in account data. So ideally we should be able to remove the out of time view regardless of how the time was added to account.
</blockquote></details>

You mean that we should check for time added, and if it is, dismiss the out of time view from didUpdateDeviceState instead?
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 876 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wYMhF91c-OYUC4vDQ:-N8yA8sj1FxDizP-27lr:b-wz1p2b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L876)):*

Swiftformat supports removing unused arguments via `unusedArguments` rule, which we have disabled because I find it counter-productive and ugly. (See https://github.com/mullvad/mullvadvpn-app/pull/3756)

Xcode auto-completes delegates with variable names, and so then you go and remove them. One day you need them back, and you have to fill the variable names back in. This generates extra git diff and creates a lot of busy work. 

There are pros and cons in either approach and it's already been discussed in detail in: https://github.com/nicklockwood/SwiftFormat/issues/83. 

In short I am in the camp of leaving those variables alone but I don't mind when it comes to closures. So please leave them be and focus on implementation instead.

More info on the swiftformat rule:
```
swiftformat --ruleinfo unusedArguments
```
</blockquote></details>

Not sure what action you want me to take here, I just ran the formatter and this was the result.
___
*[`ios/MullvadVPN/REST/RESTCreateApplePaymentResponse.swift` line 2 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wf9S-CROkAKBzTynu:-N8y9u075-5khtYSRjtr:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/REST/RESTCreateApplePaymentResponse.swift#L2)):*

Maybe we should rename it to `RESTCreateApplePaymentResponse+Localization.swift` (ugh long name!) and set localisation table to `REST` instead of `Account`. Although strictly speaking, this belong in UI/Controllers helpers.
</blockquote></details>

Done.

pronebird (changes requested):
> Reviewed 3 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 5 of 7 files reviewed, 4 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wmbPcE20BtIkGn1q6:-N90u9sp41SJCVHe1HQL:bsfgxdu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/OutOfTimeViewController.swift#L17)):*

Done. But is there any view where we don't want it to be `.lightContent`? We could just set this in info.plist and not have to deal with it in each VC
</blockquote></details>

Except the log viewer in problem report, I think everything else is dark UI + light statusbar.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 581 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8x285_8xamQhNnEtoN:-N90xWaWAl_J9Po7QDgJ:b-cepn07) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L581)):*

Done.
</blockquote></details>

Is there a reason to leave this call (`setUpOutOfTimeTimer()`) here given that `tunnelManager(:didUpdateDeviceState:)` should be called prior to ` loginViewControllerDidFinishLogin`?
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 617 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8x2PWr5J2u5XJ8QyJr:-N90urbW93c2cTYCsV0v:b-bk5ip0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L617)):*

Moved `.invalidate()`call so it's called for `.loggedOut` and `.revoked` as well.

About the guard, I think it's nicer to keep it for readability. This way we don't have to rely on the reader knowing that `deviceData != nil` means that we're logged in. What do you think?
</blockquote></details>

Works for me.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 704 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8woqGS7vqshtg3H_DP:-N90vGWR4BAkXEd52j1k:bhgype7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L704)):*

You mean that we should check for time added, and if it is, dismiss the out of time view from didUpdateDeviceState instead?
</blockquote></details>

Maybe. Somehow we need to dismiss it when the time is added to account. For example opening settings refreshes the account data, which should trigger update to device state and provide new expiry too. 

_This scenario can be simulated with a few lines of code I suppose without actually bumping the account time on server._

 I think that the desktop app simply polls account data from time to time. So perhaps we could do the same. I am discussing that right now with Oskar.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 876 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wYMhF91c-OYUC4vDQ:-N90vfW7Av_3_D1q9oJD:bws2sny) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L876)):*

Not sure what action you want me to take here, I just ran the formatter and this was the result.
</blockquote></details>

How come? I ran it on `master`, no changes. It's possible you ran the formatter without `.swiftformat` configuration?

```
pronebird@tiger ios % swiftformat --version
0.49.14
pronebird@tiger ios % git status
On branch master
Your branch is behind 'origin/master' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)

nothing to commit, working tree clean
pronebird@tiger ios % pwd          
/Users/pronebird/mullvad/mullvadvpn-app/ios
pronebird@tiger ios % swiftformat .
Running SwiftFormat...
Reading config file at /Users/pronebird/mullvad/mullvadvpn-app/ios/.swiftformat
SwiftFormat completed in 0.17s.
0/221 files formatted, 2 files skipped.
pronebird@tiger ios % git status
On branch master
Your branch is behind 'origin/master' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)

nothing to commit, working tree clean
```

Please commit relevant changes only.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 618 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N910R_T6ygrhMw7M0NJ:-N910R_T6ygrhMw7M0NK:b-n686r6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7621e2a112e0421221d2151693164f292e08b669/ios/MullvadVPN/SceneDelegate.swift#L618)):*
> ```Swift
> 
>     private func setUpOutOfTimeTimer() {
>         outOfTimeTimer?.invalidate()
> ```

Nit: can you keep blank line between this and `guard`. It's really hard to read when disconnected statements like this are squeezed together. This is akin to using paragraphs in sentences.

pronebird (commented):
> Reviewed 2 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: all files reviewed, 5 unresolved discussions (waiting on @AndreasShortcut)

___
*[`ios/MullvadVPN/SceneDelegate.swift` line 461 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9144kN-A1jXGEjpQdF:-N9144kN-A1jXGEjpQdG:b-pokltr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7621e2a112e0421221d2151693164f292e08b669/ios/MullvadVPN/SceneDelegate.swift#L461)):*
> ```Swift
> 
>     private func handleExpiredAccount() {
>         guard case let .loggedIn(accountData, _) = TunnelManager.shared.deviceState else { return }
> ```

Nit: here it's the same, perhaps add newline between guard and if it's really difficult to read this when it's squashed togther. Also you could move `accountData.expiry <= date()` into guard statement to avoid extra indentation level.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: all files reviewed, 6 unresolved discussions (waiting on @AndreasShortcut)

___
*[`ios/MullvadVPN/SceneDelegate.swift` line 465 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N914qx4-CQnQ5P_84tQ:-N914qx4-CQnQ5P_84tR:b5jjg5x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7621e2a112e0421221d2151693164f292e08b669/ios/MullvadVPN/SceneDelegate.swift#L465)):*
> ```Swift
>             switch UIDevice.current.userInterfaceIdiom {
>             case .phone:
>                 rootContainer.pushViewController(makeOutOfTimeViewController(), animated: false)
> ```

Do you think we need to guard against the potential case where we keep pushing out of time controller into the stack regardless if we already have one in there or not?

AndreasShortcut (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 6 of 7 files reviewed, 5 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/SceneDelegate.swift` line 581 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8x285_8xamQhNnEtoN:-N91LrfO4dZqKviHtfgW:b-ga4u2o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L581)):*

Is there a reason to leave this call (`setUpOutOfTimeTimer()`) here given that `tunnelManager(:didUpdateDeviceState:)` should be called prior to ` loginViewControllerDidFinishLogin`?
</blockquote></details>

Nope, my bad. Removed.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 876 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wYMhF91c-OYUC4vDQ:-N91P-dODjABAN4_7BfO:bbxpdwz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L876)):*

How come? I ran it on `master`, no changes. It's possible you ran the formatter without `.swiftformat` configuration?

```
pronebird@tiger ios % swiftformat --version
0.49.14
pronebird@tiger ios % git status
On branch master
Your branch is behind 'origin/master' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)

nothing to commit, working tree clean
pronebird@tiger ios % pwd          
/Users/pronebird/mullvad/mullvadvpn-app/ios
pronebird@tiger ios % swiftformat .
Running SwiftFormat...
Reading config file at /Users/pronebird/mullvad/mullvadvpn-app/ios/.swiftformat
SwiftFormat completed in 0.17s.
0/221 files formatted, 2 files skipped.
pronebird@tiger ios % git status
On branch master
Your branch is behind 'origin/master' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)

nothing to commit, working tree clean
```

Please commit relevant changes only.
</blockquote></details>

I don't know why.
```
andreas@MBPsomtrAndreas ios % swiftformat --version
0.49.14
andreas@MBPsomtrAndreas ios % git status
On branch add-out-of-time-view
Your branch is ahead of 'origin/add-out-of-time-view' by 5 commits.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean
andreas@MBPsomtrAndreas ios % pwd
/Users/andreas/Development/mullvadvpn-app/ios
andreas@MBPsomtrAndreas ios % swiftformat .
Running SwiftFormat...
Reading config file at /Users/andreas/Development/mullvadvpn-app/ios/.swiftformat
SwiftFormat completed in 0.11s.
0/224 files formatted, 1 file skipped.
andreas@MBPsomtrAndreas ios % 
```

Config file:
```
# file options
--exclude Build, .spm, MullvadVPNScreenshots/SnapshotHelper.swift

# general options
--swiftversion 5.5

# format options
--indent 4
--maxwidth 100
--wraparguments before-first
--wrapcollections before-first
--wrapternary before-operators
--redundanttype inferred
--disable initCoderUnavailable, redundantReturn, unusedArguments, redundantRawValues, preferKeyPath
```
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 461 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9144kN-A1jXGEjpQdF:-N91IYop6l1EOZ9Ji07-:bk8btgn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7621e2a112e0421221d2151693164f292e08b669/ios/MullvadVPN/SceneDelegate.swift#L461)):*

Nit: here it's the same, perhaps add newline between guard and if it's really difficult to read this when it's squashed togther. Also you could move `accountData.expiry <= date()` into guard statement to avoid extra indentation level.
</blockquote></details>

Moved if statement into guard
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 465 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N914qx4-CQnQ5P_84tQ:-N91IeOG9eka6vtwxtKa:b-kw6yto) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7621e2a112e0421221d2151693164f292e08b669/ios/MullvadVPN/SceneDelegate.swift#L465)):*

Do you think we need to guard against the potential case where we keep pushing out of time controller into the stack regardless if we already have one in there or not?
</blockquote></details>

Yeah, that makes sense. Done.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 618 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N910R_T6ygrhMw7M0NJ:-N91MNKWCb2L1Pwq2nt1:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7621e2a112e0421221d2151693164f292e08b669/ios/MullvadVPN/SceneDelegate.swift#L618)):*

Nit: can you keep blank line between this and `guard`. It's really hard to read when disconnected statements like this are squeezed together. This is akin to using paragraphs in sentences.
</blockquote></details>

Done.

AndreasShortcut (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 6 of 7 files reviewed, 5 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/SceneDelegate.swift` line 876 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8wYMhF91c-OYUC4vDQ:-N91V0Ey0KwXtF4JewBB:b-g9kg51) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L876)):*

I don't know why.
```
andreas@MBPsomtrAndreas ios % swiftformat --version
0.49.14
andreas@MBPsomtrAndreas ios % git status
On branch add-out-of-time-view
Your branch is ahead of 'origin/add-out-of-time-view' by 5 commits.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean
andreas@MBPsomtrAndreas ios % pwd
/Users/andreas/Development/mullvadvpn-app/ios
andreas@MBPsomtrAndreas ios % swiftformat .
Running SwiftFormat...
Reading config file at /Users/andreas/Development/mullvadvpn-app/ios/.swiftformat
SwiftFormat completed in 0.11s.
0/224 files formatted, 1 file skipped.
andreas@MBPsomtrAndreas ios % 
```

Config file:
```
# file options
--exclude Build, .spm, MullvadVPNScreenshots/SnapshotHelper.swift

# general options
--swiftversion 5.5

# format options
--indent 4
--maxwidth 100
--wraparguments before-first
--wrapcollections before-first
--wrapternary before-operators
--redundanttype inferred
--disable initCoderUnavailable, redundantReturn, unusedArguments, redundantRawValues, preferKeyPath
```
</blockquote></details>

Reverted changes

AndreasShortcut (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/SceneDelegate.swift` line 465 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N914qx4-CQnQ5P_84tQ:-N91diIw5I9IO1p5UHKQ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7621e2a112e0421221d2151693164f292e08b669/ios/MullvadVPN/SceneDelegate.swift#L465)):*

Same for iPad please.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 833 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N91YZ3RD0XFEnT5uUOw:-N91aF0i61i5lSWwqlG8:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e5fa760f9d1f204a6c611a93249c97968bd4ad5e/ios/MullvadVPN/SceneDelegate.swift#L833)):*

Last one.
</blockquote></details>

Done.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: all files reviewed, 12 unresolved discussions (waiting on @AndreasShortcut)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N959jJo5V0Zf2hqgZ7d:-N959jJo5V0Zf2hqgZ7e:botbteg) (no related file):*
Btw, what happened to idea of making a reusable view and extracting business logic into separate reusable component? A lot of code in "Out of time" view is nearly a verbatim copy of Account view.

___
*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95JXWxAWbGTV8rmSJ4:-N95JXWxAWbGTV8rmSJ5:b83plrt) (no related file):*
As discussed on Slack: out of time view should have "disconnect" button displayed when VPN tunnel is connected as there is no way to add time when account ran out of time and has no internet access. 

We had screenshots shared in private messages. It's easy to reproduce on desktop by creating a new account, then restarting the app and then turning on VPN via CLI `mullvad connect`. The UI prototype is all there.

___
___
*[`ios/MullvadVPN/OutOfTimeContentView.swift` line 60 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95Dvs8BHk3DDv66ckj:-N95Dvs8BHk3DDv66ckk:b-vpz79f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeContentView.swift#L60)):*
> ```Swift
> 
>     // TODO: Implement properly
>     lazy var redeemButton: AppButton = {
> ```

Please do not commit dead code.
___
*[`ios/MullvadVPN/OutOfTimeContentView.swift` line 116 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95CVmAADG3oq8O4kjf:-N95CVmAADG3oq8O4kjg:b-o4gt3s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeContentView.swift#L116)):*
> ```Swift
> private extension OutOfTimeContentView {
>     func disableAutoresizingMask() {
>         translatesAutoresizingMaskIntoConstraints = false
> ```

Why can't initializer set `translatesAutoresizingMaskIntoConstraints = false` directly?
___
*[`ios/MullvadVPN/OutOfTimeContentView.swift` line 120 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95Ckyi3oiyRqdq2y2d:-N95CkyjBbuuVVQ0E0i0:b-be3elk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeContentView.swift#L120)):*
> ```Swift
> 
>     func setBackgroundColor() {
>         backgroundColor = .secondaryColor
> ```

Same here. Can be inlined in `init(frame:)` and that saves us total 6 lines of code (including `disableAutoresizingMask` above).
___
*[`ios/MullvadVPN/OutOfTimeContentView.swift` line 134 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95F4xFAT9CWyTGGkBa:-N95F4xFAT9CWyTGGkBb:b-2hfq3e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeContentView.swift#L134)):*
> ```Swift
>             topStackView.leadingAnchor.constraint(
>                 equalTo: layoutMarginsGuide.leadingAnchor,
>                 constant: 16
> ```

Very much all view controller views we have use `layoutMargins = UIMetrics.contentLayoutMargins` and then the content inside is pinned to layout margins guide. Can we please stick to one system used in the codebase? I assume you looked at account and login views.
___
*[`ios/MullvadVPN/OutOfTimeContentView.swift` line 141 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95E88W48lpq0sSTCq6:-N95E88W48lpq0sSTCq7:b-64mm4a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeContentView.swift#L141)):*
> ```Swift
>             ),
> 
>             purchaseButton.heightAnchor.constraint(equalToConstant: 48),
> ```

The buttons should have intrinsic size automatically calculated. There is no need to specify the height.
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 142 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N959ebABn9vjiwK5JC2:-N959ebABn9vjiwK5JC3:bsu4tox) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeViewController.swift#L142)):*
> ```Swift
>     }
> 
>     func setProduct(_ product: SKProduct, animated _: Bool) {
> ```

If `animated` is not used then perhaps it's not needed at all?
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 159 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95B4kf42vYZVoVNQ2n:-N95B4kf42vYZVoVNQ2o:b6jravz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeViewController.swift#L159)):*
> ```Swift
>     }
> 
>     func didFailLoadingProducts(with _: Error) {
> ```

Please review your code more thoroughly. I can't be going line by line.
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 269 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N950doFDoXw93KXP5am:-N950doFDoXw93KXP5an:bjjpni) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeViewController.swift#L269)):*
> ```Swift
> extension OutOfTimeViewController: AppStorePaymentObserver {
>     func appStorePaymentManager(
>         _: AppStorePaymentManager,
> ```

All variables wiped again.
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 305 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N951HUU1pPKOXfmxlWW:-N951HUU1pPKOXfmxlWX:b-273dy3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeViewController.swift#L305)):*
> ```Swift
> 
>     func appStorePaymentManager(
>         _: AppStorePaymentManager,
> ```

Same variables wiped. Please review the code. I can't be going line by line.
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 321 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N950NqxCDDfKzU6JiYj:-N950NqxCDDfKzU6JiYk:b-wrpqbh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeViewController.swift#L321)):*
> ```Swift
> extension OutOfTimeViewController: RootContainment {
>     var preferredHeaderBarPresentation: HeaderBarPresentation {
>         .init(style: .unsecured, showsDivider: false)
> ```

It shouldn't paint the header bar in unsecured style (red) if VPN is connected. The user is safe for as long as VPN connection is up, even when out of time, hence the UI should reflect that. 

I have been thinking to move that logic into `SceneDelegate` as it seems that each controller has to monitor tunnel connection in order to paint the right header. Excessive. Although knowing the tunnel status is important for "out of time" controller.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 4 of 7 files reviewed, 10 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 123 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9B7EdpDI-M2gyZtBJs:-N9B7Edq46-iAY6BCYbX:b-l4s14j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fe930f2c03327bb3b71a2dd15f402d37c310eac4/ios/MullvadVPN/OutOfTimeViewController.swift#L123)):*
> ```Swift
>             contentView.disconnectButton.isEnabled = true
>             contentView.disconnectButton.alpha = 1
>             contentView.purchaseButton.isEnabled = false
> ```

So the `purchaseButtonInteractionRestriction` thing does the following:

```
// Make sure to disable the button if the product is not loaded
self.contentView.purchaseButton.isEnabled = enableUserInteraction &&
                self.product != nil &&
                AppStorePaymentManager.canMakePayments &&
                !self.tunnelState.isSecured
```

But the code here would enable or disable the purchase button solely based on `tunnelState` and disregard other conditions. 

I start thinking that `UserInterfaceInteractionRestriction` is actually over engineering at best and we could perhaps use a few booleans and one function to update the UI. 

What do you think?
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 392 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9BBBUMCkP9kFTHpfkH:-N9BBBUMCkP9kFTHpfkI:b856vm1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fe930f2c03327bb3b71a2dd15f402d37c310eac4/ios/MullvadVPN/OutOfTimeViewController.swift#L392)):*
> ```Swift
>             )
>         case .loggedOut:
>             return HeaderBarPresentation(style: .default, showsDivider: true)
> ```

I think we can disregard the `.loggedOut` state in here and keep it shorter:

```swift
let tunnelState = TunnelManager.shared.tunnelStatus.state

return HeaderBarPresentation(
  style: tunnelState.isSecured ? .secured : .unsecured,
  showsDivider: false
)
```

We do override the `.loggedOut` state in connect controller because on iPad it's always on screen and we don't want to flash red across the entire screen estate when user is logged out. But this view shouldn't be on screen when user is logged out. Hope that makes sense.

pronebird (commented):
> Reviewed 2 of 3 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: all files reviewed, 4 unresolved discussions (waiting on @AndreasShortcut)

AndreasShortcut (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: all files reviewed, 4 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N959jJo5V0Zf2hqgZ7d:-N95uBgk1-AA_0cFSSZP:bxv8f23) (no related file):*

Btw, what happened to idea of making a reusable view and extracting business logic into separate reusable component? A lot of code in "Out of time" view is nearly a verbatim copy of Account view.
</blockquote></details>

Discussed in slack

___
*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95JXWxAWbGTV8rmSJ4:-N96bRv3Bp8F-Ikq2n_L:bxv8f23) (no related file):*

As discussed on Slack: out of time view should have "disconnect" button displayed when VPN tunnel is connected as there is no way to add time when account ran out of time and has no internet access. 

We had screenshots shared in private messages. It's easy to reproduce on desktop by creating a new account, then restarting the app and then turning on VPN via CLI `mullvad connect`. The UI prototype is all there.
</blockquote></details>

Discussed in slack

___
___
*[`ios/MullvadVPN/OutOfTimeContentView.swift` line 60 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95Dvs8BHk3DDv66ckj:-N95vv0kDUHigltzmFYf:bt2vr1b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeContentView.swift#L60)):*

Please do not commit dead code.
</blockquote></details>

Fixed.
___
*[`ios/MullvadVPN/OutOfTimeContentView.swift` line 116 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95CVmAADG3oq8O4kjf:-N95w0vvEkt2TU_blI2h:bea3knw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeContentView.swift#L116)):*

Why can't initializer set `translatesAutoresizingMaskIntoConstraints = false` directly?
</blockquote></details>

It can. I think it's nice to split things out of the main functions, since it makes it clear at a glance which things are connected and not. Also, if you need to change something, it's easier to see if it will have any unintended consequences.

Moved to init.
___
*[`ios/MullvadVPN/OutOfTimeContentView.swift` line 120 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95Ckyi3oiyRqdq2y2d:-N95wdK7DdlUQsYHfGdJ:b-h7ncbn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeContentView.swift#L120)):*

Same here. Can be inlined in `init(frame:)` and that saves us total 6 lines of code (including `disableAutoresizingMask` above).
</blockquote></details>

Same reasoning. I assumed saving lines of code was not a priority, since the rest of the code is so verbose and previous suggestions to save space have been rejected.

Moved to init.
___
*[`ios/MullvadVPN/OutOfTimeContentView.swift` line 134 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95F4xFAT9CWyTGGkBa:-N96bVCQ0TIQczNn2gTz:bi8f774) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeContentView.swift#L134)):*

Very much all view controller views we have use `layoutMargins = UIMetrics.contentLayoutMargins` and then the content inside is pinned to layout margins guide. Can we please stick to one system used in the codebase? I assume you looked at account and login views.
</blockquote></details>

Yep, my bad. Fixed.
___
*[`ios/MullvadVPN/OutOfTimeContentView.swift` line 141 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95E88W48lpq0sSTCq6:-N96h1_nEcs6Pfy7N-MU:b2ct678) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeContentView.swift#L141)):*

The buttons should have intrinsic size automatically calculated. There is no need to specify the height.
</blockquote></details>

My bad, fixed.
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 142 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N959ebABn9vjiwK5JC2:-N96hHiR76g_uz4MF_PF:bv1jw8d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeViewController.swift#L142)):*

If `animated` is not used then perhaps it's not needed at all?
</blockquote></details>

True, fixed
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 159 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95B4kf42vYZVoVNQ2n:-N96hnrc7UEoL9ANDaC8:b-886gsx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeViewController.swift#L159)):*

Please review your code more thoroughly. I can't be going line by line.
</blockquote></details>

Fixed
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 269 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N950doFDoXw93KXP5am:-N96i71r5RXbvHLIeflH:b-886gsx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeViewController.swift#L269)):*

All variables wiped again.
</blockquote></details>

Fixed
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 305 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N951HUU1pPKOXfmxlWW:-N96iN_yBwCps2-zMpCU:b-886gsx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeViewController.swift#L305)):*

Same variables wiped. Please review the code. I can't be going line by line.
</blockquote></details>

Fixed
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 321 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N950NqxCDDfKzU6JiYj:-N96iUCYBWyFyXxQaSh8:b-42sqq3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeViewController.swift#L321)):*

It shouldn't paint the header bar in unsecured style (red) if VPN is connected. The user is safe for as long as VPN connection is up, even when out of time, hence the UI should reflect that. 

I have been thinking to move that logic into `SceneDelegate` as it seems that each controller has to monitor tunnel connection in order to paint the right header. Excessive. Although knowing the tunnel status is important for "out of time" controller.
</blockquote></details>

Ah, I didn't realise this was the case. I assumed you were disconnected when time ran out.

Fixed and added conditional disconnect button.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 704 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8woqGS7vqshtg3H_DP:-N9BGa6i-SnXKYKUdn4U:b-i41ces) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L704)):*

Maybe. Somehow we need to dismiss it when the time is added to account. For example opening settings refreshes the account data, which should trigger update to device state and provide new expiry too. 

_This scenario can be simulated with a few lines of code I suppose without actually bumping the account time on server._

 I think that the desktop app simply polls account data from time to time. So perhaps we could do the same. I am discussing that right now with Oskar.
</blockquote></details>

Added a check in `didUpdateDeviceState` as suggested, after discussion in Slack

pronebird (commented):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: all files reviewed, 3 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/OutOfTimeContentView.swift` line 120 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N95Ckyi3oiyRqdq2y2d:-N9BZCvEEAXZLe20FkLt:b-3aud0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d2b86e80efa3c13fe4be615c7c5962f16f2508e/ios/MullvadVPN/OutOfTimeContentView.swift#L120)):*

Same reasoning. I assumed saving lines of code was not a priority, since the rest of the code is so verbose and previous suggestions to save space have been rejected.

Moved to init.
</blockquote></details>

It depends, as long as it code does not sacrifice readability or comprehension. I think we better touch on that on Slack or have a call.
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 392 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9C0_zr8Hr0MN1U6K5l:-N9C0_zr8Hr0MN1U6K5m:bm1uzp0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5fe1a044e25ea864dbbf38aef9c97cce074e9d0e/ios/MullvadVPN/OutOfTimeViewController.swift#L392)):*
> ```Swift
>     var preferredHeaderBarPresentation: HeaderBarPresentation {
>         return HeaderBarPresentation(
>             style: TunnelManager.shared.tunnelStatus.state.isSecured ? .secured : .unsecured,
> ```

FYI Minor: since `tunnelState` is stored as a member variable, then perhaps you can avoid going through `TunnelManager.shared.tunnelStatus.state` once again.



_Code quote:_
```Swift
 tunnelState
```

AndreasShortcut (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 6 of 7 files reviewed, 3 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 392 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9BBBUMCkP9kFTHpfkH:-N9BIV2uEYMs9iTa9EUZ:b-1eh3wi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fe930f2c03327bb3b71a2dd15f402d37c310eac4/ios/MullvadVPN/OutOfTimeViewController.swift#L392)):*

I think we can disregard the `.loggedOut` state in here and keep it shorter:

```swift
let tunnelState = TunnelManager.shared.tunnelStatus.state

return HeaderBarPresentation(
  style: tunnelState.isSecured ? .secured : .unsecured,
  showsDivider: false
)
```

We do override the `.loggedOut` state in connect controller because on iPad it's always on screen and we don't want to flash red across the entire screen estate when user is logged out. But this view shouldn't be on screen when user is logged out. Hope that makes sense.
</blockquote></details>

Sure, done.
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 392 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9C0_zr8Hr0MN1U6K5l:-N9CM1IWBWpTAhoZIcww:bn043xi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5fe1a044e25ea864dbbf38aef9c97cce074e9d0e/ios/MullvadVPN/OutOfTimeViewController.swift#L392)):*

FYI Minor: since `tunnelState` is stored as a member variable, then perhaps you can avoid going through `TunnelManager.shared.tunnelStatus.state` once again.
</blockquote></details>

Nice catch, thanks. Fixed.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: all files reviewed, 3 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

AndreasShortcut (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 2 of 7 files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 123 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9B7EdpDI-M2gyZtBJs:-N9W0eLu2kGAVSn3ACB0:b-886gsx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fe930f2c03327bb3b71a2dd15f402d37c310eac4/ios/MullvadVPN/OutOfTimeViewController.swift#L123)):*

So the `purchaseButtonInteractionRestriction` thing does the following:

```
// Make sure to disable the button if the product is not loaded
self.contentView.purchaseButton.isEnabled = enableUserInteraction &&
                self.product != nil &&
                AppStorePaymentManager.canMakePayments &&
                !self.tunnelState.isSecured
```

But the code here would enable or disable the purchase button solely based on `tunnelState` and disregard other conditions. 

I start thinking that `UserInterfaceInteractionRestriction` is actually over engineering at best and we could perhaps use a few booleans and one function to update the UI. 

What do you think?
</blockquote></details>

Fixed
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 704 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N8woqGS7vqshtg3H_DP:-N9W2NUyEW3C7ztk3Fq_:b-qt6oj4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/24ff81317cbb4f09a77455b8b27a114a2dd448fa/ios/MullvadVPN/SceneDelegate.swift#L704)):*

Maybe this delegate method is not needed if `didUpdateDeviceState` takes care of dismissing the controller and updating the timer?
</blockquote></details>

Good point. Fixed

pronebird (commented):
> Reviewed 1 of 4 files at r13, 4 of 6 files at r14.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 7 of 9 files reviewed, 1 unresolved discussion (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/SceneDelegate.swift` line 47 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9_vu425d3LJl0NMqgN:-N9_vu433nY6dYEWv-Jz:bqkhcha) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dbb2ef8f2f54f74851f82ec1e980f417b8798c93/ios/MullvadVPN/SceneDelegate.swift#L47)):*
> ```Swift
>         case .pad:
>             return modalRootContainer.viewControllers
>                 .first(where: { $0 is OutOfTimeViewController }) != nil
> ```

nit: we could actually use `.contains(where: (T) -> Bool)` but that doesn't matter much.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 7 of 9 files reviewed, 2 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/RootContainerViewController.swift` line 253 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9a3bcOByf-PNihh4UG:-N9a3bcOByf-PNihh4UH:bsa3nvc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dbb2ef8f2f54f74851f82ec1e980f417b8798c93/ios/MullvadVPN/RootContainerViewController.swift#L253)):*
> ```Swift
>         headerBarView.settingsButton.alpha = isEnabled ? 1 : 0.5
>         headerBarView.settingsButton.isEnabled = isEnabled
>         presentationContainerSettingsButton?.isEnabled = isEnabled
> ```

`presentationContainerSettingsButton` is basically a second button that's added into modal presentation container. It very much duplicates the `headerBarView.settingsButton` and has identical position. So perhaps you may want to apply alpha to both of them.

Another way would be to override `makeSettingsButton()` in `HeaderBarView` and set the image for disabled set, i.e:

```swift
let disabledSettingsImage = UIImage(named: "IconSettings")?
            .backport_withTintColor(UIColor.HeaderBar.disabledButtonColor, renderingMode: .alwaysOriginal)

settingsButton.setImage(disabledSettingsImage, for: .disabled)
```

Then UIKit will take care of opacity once you flip `isEnabled`.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 7 of 9 files reviewed, 2 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/RootContainerViewController.swift` line 253 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9a3bcOByf-PNihh4UG:-N9a6klW0vSl8b_b4yfa:bw18byw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dbb2ef8f2f54f74851f82ec1e980f417b8798c93/ios/MullvadVPN/RootContainerViewController.swift#L253)):*

`presentationContainerSettingsButton` is basically a second button that's added into modal presentation container. It very much duplicates the `headerBarView.settingsButton` and has identical position. So perhaps you may want to apply alpha to both of them.

Another way would be to override `makeSettingsButton()` in `HeaderBarView` and set the image for disabled set, i.e:

```swift
let disabledSettingsImage = UIImage(named: "IconSettings")?
            .backport_withTintColor(UIColor.HeaderBar.disabledButtonColor, renderingMode: .alwaysOriginal)

settingsButton.setImage(disabledSettingsImage, for: .disabled)
```

Then UIKit will take care of opacity once you flip `isEnabled`.
</blockquote></details>

I actually realised that since both `addSettingsButtonToPresentationContainer()` and `removeSettingsButtonFromPresentationContainer()` use the `opacity` to toggle the visibility of original settings button, the best way to not interfere with this is to assign the image with desired opacity for `.disabled` state.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 18 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9a78kGDmvMwgAE1MNE:-N9a78kGDmvMwgAE1MNF:b-ilra75) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dbb2ef8f2f54f74851f82ec1e980f417b8798c93/ios/MullvadVPN/OutOfTimeViewController.swift#L18)):*
> ```Swift
>     private var productState: ProductState = .none
>     private var paymentState: PaymentState = .none
>     private var product: SKProduct?
> ```

As outlined in our conversation earlier today, both `product` and `pendingPayment` seem to be the leftovers from rebase. `productState` and `paymentState` are meant to replace them.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 7 of 9 files reviewed, 4 unresolved discussions (waiting on @AndreasShortcut and @pronebird)

___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 99 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9a8bMwE_9zMmoSbU9J:-N9a8bMwE_9zMmoSbU9K:bw18uo6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dbb2ef8f2f54f74851f82ec1e980f417b8798c93/ios/MullvadVPN/OutOfTimeViewController.swift#L99)):*
> ```Swift
>         if tunnelState.isSecured {
>             return NSLocalizedString(
>                 "OUT_OF_TIME_BOD_CONNECTEDY",
> ```

Minor: typo in `OUT_OF_TIME_BOD_CONNECTEDY` instead of `OUT_OF_TIME_BODY_CONNECTED`

AndreasShortcut (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 5 of 11 files reviewed, 4 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 18 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9a78kGDmvMwgAE1MNE:-N9aIvVEE4a8WgY3igqK:bss4a2b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dbb2ef8f2f54f74851f82ec1e980f417b8798c93/ios/MullvadVPN/OutOfTimeViewController.swift#L18)):*

As outlined in our conversation earlier today, both `product` and `pendingPayment` seem to be the leftovers from rebase. `productState` and `paymentState` are meant to replace them.
</blockquote></details>

Removed.
___
*[`ios/MullvadVPN/OutOfTimeViewController.swift` line 99 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9a8bMwE_9zMmoSbU9J:-N9aJDjn9GGmaKWLtZDz:bt2vr1b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dbb2ef8f2f54f74851f82ec1e980f417b8798c93/ios/MullvadVPN/OutOfTimeViewController.swift#L99)):*

Minor: typo in `OUT_OF_TIME_BOD_CONNECTEDY` instead of `OUT_OF_TIME_BODY_CONNECTED`
</blockquote></details>

Fixed.
___
*[`ios/MullvadVPN/RootContainerViewController.swift` line 253 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9a3bcOByf-PNihh4UG:-N9aQAsrDJGD1IDgc1z1:bt2vr1b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dbb2ef8f2f54f74851f82ec1e980f417b8798c93/ios/MullvadVPN/RootContainerViewController.swift#L253)):*

I actually realised that since both `addSettingsButtonToPresentationContainer()` and `removeSettingsButtonFromPresentationContainer()` use the `opacity` to toggle the visibility of original settings button, the best way to not interfere with this is to assign the image with desired opacity for `.disabled` state.
</blockquote></details>

Fixed.
___
*[`ios/MullvadVPN/SceneDelegate.swift` line 47 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-N9_vu425d3LJl0NMqgN:-N9a16lcC0LeDpJAzbOM:b-nc3pl7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dbb2ef8f2f54f74851f82ec1e980f417b8798c93/ios/MullvadVPN/SceneDelegate.swift#L47)):*

nit: we could actually use `.contains(where: (T) -> Bool)` but that doesn't matter much.
</blockquote></details>

I agree, that's cleaner. Fixed

pronebird (commented):
> Reviewed 4 of 6 files at r15.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 9 of 11 files reviewed, all discussions resolved (waiting on @pronebird)

pronebird (commented):
> Reviewed 1 of 6 files at r15.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811)* status: 10 of 11 files reviewed, all discussions resolved (waiting on @pronebird)

pronebird (approved):
:lgtm:

> Reviewed 2 of 2 files at r16.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3811#-:-N9aTp-t1o7f-nBvDUnL:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3820, opened 2022-08-09
Refactor index.ts

raksooo:
This PR refactors `index.ts` and `app.tsx` by moving code out of the main classes since especially `index.ts` had become way to large and contained a lot of logic which was easily extracted. There's more work to do here but this serves as a good starting point.

The `notify` functions in `IpcMainEventChannel` is now also curried with the current `WebContents` to make it easier to call those functions. The function will be replaced by `undefined` if there's no `WebContents`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820)

### DIFF
--- a/gui/src/main/account.ts
+++ b/gui/src/main/account.ts
@@ -11,21 +11,21 @@ import log from '../shared/logging';
 import {
   AccountExpiredNotificationProvider,
   CloseToAccountExpiryNotificationProvider,
-  SystemNotification,
 } from '../shared/notifications/notification';
 import { Scheduler } from '../shared/scheduler';
 import AccountDataCache from './account-data-cache';
 import { DaemonRpc } from './daemon-rpc';
 import { InvalidAccountError } from './errors';
 import { IpcMainEventChannel } from './ipc-event-channel';
+import { NotificationSender } from './notification-controller';
+import { TunnelStateProvider } from './tunnel-state';
+
+export interface LocaleProvider {
+  getLocale(): string;
+}
 
 export interface AccountDelegate {
-  notify(notification: SystemNotification): void;
-  getTunnelState(): TunnelState;
-  getLocale(): string;
-  isPerformingPostUpgradeCheck(): boolean;
-  performPostUpgradeCheck(): void;
-  setTrayContextMenu(): void;
+  onDeviceEvent(): void;
 }
 
 export default class Account {
@@ -47,7 +47,10 @@ export default class Account {
 
   private deviceStateValue?: DeviceState;
 
-  public constructor(private delegate: AccountDelegate, private daemonRpc: DaemonRpc) {}
+  public constructor(
+    private delegate: AccountDelegate & TunnelStateProvider & LocaleProvider & NotificationSender,
+    private daemonRpc: DaemonRpc,
+  ) {}
 
   public get accountData() {
     return this.accountDataValue;
@@ -123,10 +126,6 @@ export default class Account {
   public handleDeviceEvent(deviceEvent: DeviceEvent) {
     this.deviceStateValue = deviceEvent.deviceState;
 
-    if (this.delegate.isPerformingPostUpgradeCheck()) {
-      void this.delegate.performPostUpgradeCheck();
-    }
-
     switch (deviceEvent.deviceState.type) {
       case 'logged in':
         this.accountDataCache.fetch(deviceEvent.deviceState.accountAndDevice.accountToken);
@@ -138,7 +137,7 @@ export default class Account {
     }
 
     void this.updateAccountHistory();
-    this.delegate.setTrayContextMenu();
+    this.delegate.onDeviceEvent();
 
     IpcMainEventChannel.account.notifyDevice?.(deviceEvent);
   }
--- a/gui/src/main/account.ts
+++ b/gui/src/main/account.ts
@@ -1,0 +1,232 @@
+import {
+  AccountToken,
+  DeviceEvent,
+  DeviceState,
+  IAccountData,
+  IDeviceRemoval,
+  TunnelState,
+} from '../shared/daemon-rpc-types';
+import { messages } from '../shared/gettext';
+import log from '../shared/logging';
+import {
+  AccountExpiredNotificationProvider,
+  CloseToAccountExpiryNotificationProvider,
+  SystemNotification,
+} from '../shared/notifications/notification';
+import { Scheduler } from '../shared/scheduler';
+import AccountDataCache from './account-data-cache';
+import { DaemonRpc } from './daemon-rpc';
+import { InvalidAccountError } from './errors';
+import { IpcMainEventChannel } from './ipc-event-channel';
+
+export interface AccountDelegate {
+  notify(notification: SystemNotification): void;
+  getTunnelState(): TunnelState;
+  getLocale(): string;
+  isPerformingPostUpgradeCheck(): boolean;
+  performPostUpgradeCheck(): void;
+  setTrayContextMenu(): void;
+}
+
+export default class Account {
+  private accountDataValue?: IAccountData = undefined;
+  private accountHistoryValue?: AccountToken = undefined;
+  private accountExpiryNotificationScheduler = new Scheduler();
+  private accountDataCache = new AccountDataCache(
+    (accountToken) => {
+      return this.daemonRpc.getAccountData(accountToken);
+    },
+    (accountData) => {
+      this.accountDataValue = accountData;
+
+      IpcMainEventChannel.account.notify?.(this.accountData);
+
+      this.handleAccountExpiry();
+    },
+  );
+
+  private deviceStateValue?: DeviceState;
+
+  public constructor(private delegate: AccountDelegate, private daemonRpc: DaemonRpc) {}
+
+  public get accountData() {
+    return this.accountDataValue;
+  }
+
+  public get accountHistory() {
+    return this.accountHistoryValue;
+  }
+
+  public get deviceState() {
+    return this.deviceStateValue;
+  }
+
+  public registerIpcListeners() {
+    IpcMainEventChannel.account.handleCreate(() => this.createNewAccount());
+    IpcMainEventChannel.account.handleLogin((token: AccountToken) => this.login(token));
+    IpcMainEventChannel.account.handleLogout(() => this.logout());
+    IpcMainEventChannel.account.handleGetWwwAuthToken(() => this.daemonRpc.getWwwAuthToken());
+    IpcMainEventChannel.account.handleSubmitVoucher(async (voucherCode: string) => {
+      const currentAccountToken = this.getAccountToken();
+      const response = await this.daemonRpc.submitVoucher(voucherCode);
+
+      if (currentAccountToken) {
+        this.accountDataCache.handleVoucherResponse(currentAccountToken, response);
+      }
+
+      return response;
+    });
+    IpcMainEventChannel.account.handleUpdateData(() => this.updateAccountData());
+
+    IpcMainEventChannel.accountHistory.handleClear(async () => {
+      await this.daemonRpc.clearAccountHistory();
+      void this.updateAccountHistory();
+    });
+
+    IpcMainEventChannel.account.handleGetDeviceState(async () => {
+      try {
+        await this.daemonRpc.updateDevice();
+      } catch (e) {
+        const error = e as Error;
+        log.warn(`Failed to update device info: ${error.message}`);
+      }
+      return this.daemonRpc.getDevice();
+    });
+    IpcMainEventChannel.account.handleListDevices((accountToken: AccountToken) => {
+      return this.daemonRpc.listDevices(accountToken);
+    });
+    IpcMainEventChannel.account.handleRemoveDevice((deviceRemoval: IDeviceRemoval) => {
+      return this.daemonRpc.removeDevice(deviceRemoval);
+    });
+  }
+
+  public isLoggedIn(): boolean {
+    return this.deviceState?.type === 'logged in';
+  }
+
+  public updateAccountData = () => {
+    if (this.daemonRpc.isConnected && this.isLoggedIn()) {
+      this.accountDataCache.fetch(this.getAccountToken()!);
+    }
+  };
+
+  public detectStaleAccountExpiry(tunnelState: TunnelState) {
+    const hasExpired = !this.accountData || new Date() >= new Date(this.accountData.expiry);
+
+    // It's likely that the account expiry is stale if the daemon managed to establish the tunnel.
+    if (tunnelState.state === 'connected' && hasExpired) {
+      log.info('Detected the stale account expiry.');
+      this.accountDataCache.invalidate();
+    }
+  }
+
+  public handleDeviceEvent(deviceEvent: DeviceEvent) {
+    this.deviceStateValue = deviceEvent.deviceState;
+
+    if (this.delegate.isPerformingPostUpgradeCheck()) {
+      void this.delegate.performPostUpgradeCheck();
+    }
+
+    switch (deviceEvent.deviceState.type) {
+      case 'logged in':
+        this.accountDataCache.fetch(deviceEvent.deviceState.accountAndDevice.accountToken);
+        break;
+      case 'logged out':
+      case 'revoked':
+        this.accountDataCache.invalidate();
+        break;
+    }
+
+    void this.updateAccountHistory();
+    this.delegate.setTrayContextMenu();
+
+    IpcMainEventChannel.account.notifyDevice?.(deviceEvent);
+  }
+
+  public setAccountHistory(accountHistory?: AccountToken) {
+    this.accountHistoryValue = accountHistory;
+
+    IpcMainEventChannel.accountHistory.notify?.(accountHistory);
+  }
+
+  private async createNewAccount(): Promise<string> {
+    try {
+      return await this.daemonRpc.createNewAccount();
+    } catch (e) {
+      const error = e as Error;
+      log.error(`Failed to create account: ${error.message}`);
+      throw error;
+    }
+  }
+
+  private async login(accountToken: AccountToken): Promise<void> {
+    try {
+      await this.daemonRpc.loginAccount(accountToken);
+    } catch (e) {
+      const error = e as Error;
+      log.error(`Failed to login: ${error.message}`);
+
+      if (error instanceof InvalidAccountError) {
+        throw Error(messages.gettext('Invalid account number'));
+      } else {
+        throw error;
+      }
+    }
+  }
+
+  private async logout(): Promise<void> {
+    try {
+      await this.daemonRpc.logoutAccount();
+
+      this.accountExpiryNotificationScheduler.cancel();
+    } catch (e) {
+      const error = e as Error;
+      log.info(`Failed to logout: ${error.message}`);
+
+      throw error;
+    }
+  }
+
+  private handleAccountExpiry() {
+    if (this.accountData) {
+      const expiredNotification = new AccountExpiredNotificationProvider({
+        accountExpiry: this.accountData.expiry,
+        tunnelState: this.delegate.getTunnelState(),
+      });
+      const closeToExpiryNotification = new CloseToAccountExpiryNotificationProvider({
+        accountExpiry: this.accountData.expiry,
+        locale: this.delegate.getLocale(),
+      });
+
+      if (expiredNotification.mayDisplay()) {
+        this.accountExpiryNotificationScheduler.cancel();
+        this.delegate.notify(expiredNotification.getSystemNotification());
+      } else if (
+        !this.accountExpiryNotificationScheduler.isRunning &&
+        closeToExpiryNotification.mayDisplay()
+      ) {
+        this.delegate.notify(closeToExpiryNotification.getSystemNotification());
+
+        const twelveHours = 12 * 60 * 60 * 1000;
+        const remainingMilliseconds = new Date(this.accountData.expiry).getTime() - Date.now();
+        const delay = Math.min(twelveHours, remainingMilliseconds);
+        this.accountExpiryNotificationScheduler.schedule(() => this.handleAccountExpiry(), delay);
+      }
+    }
+  }
+
+  private async updateAccountHistory(): Promise<void> {
+    try {
+      this.setAccountHistory(await this.daemonRpc.getAccountHistory());
+    } catch (e) {
+      const error = e as Error;
+      log.error(`Failed to fetch the account history: ${error.message}`);
+    }
+  }
+
+  private getAccountToken(): AccountToken | undefined {
+    return this.deviceState?.type === 'logged in'
+      ? this.deviceState.accountAndDevice.accountToken
+      : undefined;
+  }
+}
--- a/gui/src/main/daemon-rpc.ts
+++ b/gui/src/main/daemon-rpc.ts
@@ -80,7 +80,10 @@ const invalidErrorStateCause = new Error(
 );
 
 export class ConnectionObserver {
-  constructor(private openHandler: () => void, private closeHandler: (error?: Error) => void) {}
+  constructor(
+    private openHandler: () => void,
+    private closeHandler: (wasConnected: boolean, error?: Error) => void,
+  ) {}
 
   // Only meant to be called by DaemonRpc
   // @internal
@@ -90,8 +93,8 @@ export class ConnectionObserver {
 
   // Only meant to be called by DaemonRpc
   // @internal
-  public onClose = (error?: Error) => {
-    this.closeHandler(error);
+  public onClose = (wasConnected: boolean, error?: Error) => {
+    this.closeHandler(wasConnected, error);
   };
 }
 
@@ -152,7 +155,7 @@ export class DaemonRpc {
     return new Promise((resolve, reject) => {
       this.client.waitForReady(this.deadlineFromNow(), (error) => {
         if (error) {
-          this.connectionObservers.forEach((observer) => observer.onClose(error));
+          this.onClose(error);
           this.ensureConnectivity();
           reject(error);
         } else {
@@ -644,6 +647,13 @@ export class DaemonRpc {
     }
   }
 
+  private onClose(error?: Error) {
+    const wasConnected = this.isConnectedValue;
+    this.isConnectedValue = false;
+
+    this.connectionObservers.forEach((observer) => observer.onClose(wasConnected, error));
+  }
+
   private removeSubscription(id: number) {
     const subscription = this.subscriptions.get(id);
     if (subscription !== undefined) {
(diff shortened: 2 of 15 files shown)


pronebird (changes requested):
> Reviewed 7 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820)* status: 7 of 15 files reviewed, 8 unresolved discussions (waiting on @raksooo)

___
*[`gui/src/main/user-interface.ts` line 65 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96IyupCcwqlRg4_Zhk:-N96IyupCcwqlRg4_Zhl:bwhzus4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L65)):*
> ```XML
> 
>   public async initializeWindow() {
>     if (!this.windowController?.window) {
> ```

How can `windowController` be `null`?  It's defined as:

```
private windowController: WindowController;
```
___
*[`gui/src/main/user-interface.ts` line 85 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96J__c8tsEgwHELvIZ:-N96J__c8tsEgwHELvI_:bpxlv0a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L85)):*
> ```XML
>       case 'win32':
>         this.installWindowsMenubarAppWindowHandlers(
>           this.windowController,
> ```

Inconsistency. Unless there is a reason, perhaps you should either stick to local variable `windowController` or member `this.windowController`? Some occurrences below access member too.
___
*[`gui/src/main/user-interface.ts` line 138 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96IlHq6xREAq5FZ55A:-N96IlHq6xREAq5FZ55B:b-ktbrs0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L138)):*
> ```XML
> 
>   public async recreateWindow(): Promise<void> {
>     if (this.tray && this.windowController) {
> ```

`this.windowController` is always `true`.



_Code quote:_
```XML
windowController: WindowController;
```
___
*[`gui/src/main/user-interface.ts` line 267 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96L6Qu0jG_uvZ1SwMc:-N96L6Qu0jG_uvZ1SwMd:b2274au) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L267)):*
> ```XML
>       case 'linux':
>         return new BrowserWindow({
>           ...options,
> ```

You could just give it `options` instead of spreading it. Also `default` case does exactly that, so perhaps case can be made to remove the `linux` branch?
___
*[`gui/src/main/user-interface.ts` line 417 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96M7pDC4vpycE1ROlH:-N96M7pDC4vpycE1ROlI:btoz1u2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L417)):*
> ```XML
>     browsingFiles: boolean,
>   ) {
>     if (!this.delegate.getUnpinnedWindow()) {
> ```

The `if` condition engulfs the entire function body, perhaps you could use early return instead?

```js
if (this.delegate.getUnpinnedWindow()) {
  return;
}

// Confiigure event monitor
```
___
*[`gui/src/main/user-interface.ts` line 437 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96LfzoCGgRqor1Gf9c:-N96LfzoCGgRqor1Gf9d:b6i2wsj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L437)):*
> ```XML
>   // see https://github.com/electron/electron/issues/8689
>   private installMacOsMenubarAppWindowHandlers(windowController: WindowController) {
>     if (!this.delegate.getUnpinnedWindow()) {
> ```

Same here, if statements engulfs the function body.
___
*[`gui/src/main/user-interface.ts` line 629 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96MoMMA6ZoAEELJoHO:-N96MoMMA6ZoAEELJoHP:b-e81i80) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L629)):*
> ```XML
>   // WindowControllerDelegate
>   public getTrayBounds = () => this.tray.getBounds();
>   public getUnpinnedWindow = this.delegate.getUnpinnedWindow;
> ```

Method above captures `this` with fat arrow closure but this one doesn't. Is it correct?
___
*[`gui/src/main/window-controller.ts` line 19 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96NExDCEYJXIq5x2Wz:-N96NExDCEYJXIq5x2X-:b-5dpu8d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/window-controller.ts#L19)):*
> ```XML
> 
> export interface WindowControllerDelegate {
>   getTrayBounds: Tray['getBounds'];
> ```

FYI maybe we should expand that definition to `getTrayBounds(): Rectangle`. Kinda looks funny otherwise, but I understand where you come from.
___
*[`gui/src/main/window-controller.ts` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96Nhgx9lKDb9N6sv0l:-N96Nhgx9lKDb9N6sv0m:b2j0jlq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/window-controller.ts#L20)):*
> ```XML
> export interface WindowControllerDelegate {
>   getTrayBounds: Tray['getBounds'];
>   getUnpinnedWindow(): boolean;
> ```

Maybe should be `isUnpinnedWindow(): boolean`?

pronebird (changes requested):
> Reviewed 7 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820)* status: 14 of 15 files reviewed, 10 unresolved discussions (waiting on @raksooo)

___
*[`gui/src/main/index.ts` line 66 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96StgkDt1ifDzYbM2P:-N96StgkDt1ifDzYbM2Q:b-owa598) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/index.ts#L66)):*
> ```XML
> 
> const SANDBOX_DISABLED = app.commandLine.hasSwitch('no-sandbox');
> const QUIT_WITHOUT_DISCONNECT = process.argv.includes(CommandLineOptions.quitWithoutDisconnect);
> ```

Second instance handler checks `argv` separately. I'd inline these constants to avoid confusion.
___
*[`gui/src/main/index.ts` line 917 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96Rdou7mLJPB2dL-ZL:-N96Rdou7mLJPB2dL-ZM:b8z2e1n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/index.ts#L917)):*
> ```XML
>         return this.settings.gui.unpinnedWindow && !this.settings.gui.startMinimized;
>       default:
>         return true;
> ```

Can't really tell where this comes from. Is it `true` on other platforms, i.e `freebsd` for a reason or that `switch-case` was optimised to the point where all major platforms have the same behaviour?
___
*[`gui/src/main/settings.ts` line 166 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96Pkgm7vRqEcmqQzxu:-N96Pkgn7q9hdeZHI0nN:b1dmke4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/settings.ts#L166)):*
> ```XML
>   }
> 
>   public get allowLan() {
> ```

FYI all these methods should use dynamic getter in my book but I guess its not easy to express that in TS? Hence a lot of hardcode.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820)* status: 14 of 15 files reviewed, 10 unresolved discussions (waiting on @raksooo)

___
*[`gui/src/main/account.ts` line 22 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96Vb2Y0nc_TT0vJa1g:-N96Vb2Y0nc_TT0vJa1h:b-wdogtv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/account.ts#L22)):*
> ```XML
> import { IpcMainEventChannel } from './ipc-event-channel';
> 
> export interface AccountDelegate {
> ```

FYI I must say that this is less of a delegate and more of collection of unrelated things. 

The delegation that I see in practice is when:

1. You either react to object doing something (similar to observer). For example imagine you have a Login component. It notifies its delegate when it begins and ends logging in. The delegate may then block or unblock other UI components in response to those calls. More like enhancing the behaviour.

```ts
interface Delegate {
   willLogin(): void;
   didFinishLogin(Boolean): void;
}

class Login {
  init(private delegate: Delegate)
  
  login() {
       this.delegate.wilLogin()

       this.startLogin((error) => {
            // ...

            this.delegate.didFinishLogin(error === null)
       })
  }
}
```

2.Or when object consults the delegate on what to do next, i.e imaging the same `Login` component that asks the `delegate` for account validation. For example:

```ts
interface Delegate {
   validateAccountNumber(String): Bool;
}

class Login {
  init(private delegate: Delegate)

  login() {
       if (this.delegate.validateAccountNumber(this.input)) {
           throw Error('Account is not valid')
       } else {
           // Start login
       }
  }
}
```

3. Or a combination of both.

What you have here is more like a collection of unrelated objects that perform some work. A lot of this fits Command pattern very well too, i.e:

```
interface ICommand {
  execute(): void;
}
```

So in other words you have split that delegate on multiple small objects doing different work:

```ts
interface TunnelStatusProvider {
  isConnectedToDaemon(): boolean;
  getTunnelState(): TunnelState;
}

interface PostUpgradeCheckHandler {
  isPerformingPostUpgradeCheck(): boolean;
  performPostUpgradeCheck(): void;
}

interface NotifiicationSender {
    send(notification: SystemNotification): void;
}

interface TrayContextMenuHandler {
   update(): void;
}

interface LocaleProvider {
  getLocale(): string;
}

interface AccountStatusProvider {
  isLoggedIn(): boolean;
}
```

But the odd thing is that `Account` has access to `this.daemonRpc` but then it somehow needs to know if it's connected to daemon (`isConnectedToDaemon`). Also since it has device state handy, can't it deduce if it's logged in from device state?

pronebird (commented):
> Reviewed 1 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820)* status: all files reviewed, 10 unresolved discussions (waiting on @raksooo)

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820)* status: 8 of 15 files reviewed, 10 unresolved discussions (waiting on @pronebird)

___
*[`gui/src/main/account.ts` line 22 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96Vb2Y0nc_TT0vJa1g:-N9GlaPr8w7kB4ClZeua:b-gr46n2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/account.ts#L22)):*

FYI I must say that this is less of a delegate and more of collection of unrelated things. 

The delegation that I see in practice is when:

1. You either react to object doing something (similar to observer). For example imagine you have a Login component. It notifies its delegate when it begins and ends logging in. The delegate may then block or unblock other UI components in response to those calls. More like enhancing the behaviour.

```ts
interface Delegate {
   willLogin(): void;
   didFinishLogin(Boolean): void;
}

class Login {
  init(private delegate: Delegate)
  
  login() {
       this.delegate.wilLogin()

       this.startLogin((error) => {
            // ...

            this.delegate.didFinishLogin(error === null)
       })
  }
}
```

2.Or when object consults the delegate on what to do next, i.e imaging the same `Login` component that asks the `delegate` for account validation. For example:

```ts
interface Delegate {
   validateAccountNumber(String): Bool;
}

class Login {
  init(private delegate: Delegate)

  login() {
       if (this.delegate.validateAccountNumber(this.input)) {
           throw Error('Account is not valid')
       } else {
           // Start login
       }
  }
}
```

3. Or a combination of both.

What you have here is more like a collection of unrelated objects that perform some work. A lot of this fits Command pattern very well too, i.e:

```
interface ICommand {
  execute(): void;
}
```

So in other words you have split that delegate on multiple small objects doing different work:

```ts
interface TunnelStatusProvider {
  isConnectedToDaemon(): boolean;
  getTunnelState(): TunnelState;
}

interface PostUpgradeCheckHandler {
  isPerformingPostUpgradeCheck(): boolean;
  performPostUpgradeCheck(): void;
}

interface NotifiicationSender {
    send(notification: SystemNotification): void;
}

interface TrayContextMenuHandler {
   update(): void;
}

interface LocaleProvider {
  getLocale(): string;
}

interface AccountStatusProvider {
  isLoggedIn(): boolean;
}
```

But the odd thing is that `Account` has access to `this.daemonRpc` but then it somehow needs to know if it's connected to daemon (`isConnectedToDaemon`). Also since it has device state handy, can't it deduce if it's logged in from device state?
</blockquote></details>

I agree. I've moved the tracking of the daemon connection state to daemon-rpc. I've also provided a lot more of the required data as arguments instead of providing getters. I'll look into splitting it up a bit more as well.
___
*[`gui/src/main/index.ts` line 66 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96StgkDt1ifDzYbM2P:-N9Gm6bM6-i87DTKfYmW:b-z904gp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/index.ts#L66)):*

Second instance handler checks `argv` separately. I'd inline these constants to avoid confusion.
</blockquote></details>

I don't get what you're suggestion, could you clarify? The `argv` in the handler of `seconds-instance` isn't the same as this one, it's the argv of the other process provided as a parameter.
___
*[`gui/src/main/index.ts` line 917 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96Rdou7mLJPB2dL-ZL:-N9GngTf7CPseEJ-PpTd:b46wof4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/index.ts#L917)):*

Can't really tell where this comes from. Is it `true` on other platforms, i.e `freebsd` for a reason or that `switch-case` was optimised to the point where all major platforms have the same behaviour?
</blockquote></details>

Janito wrote this code in 2018 so I don't know the reasoning. But we don't support any other platforms so it shouldn't be needed.
___
*[`gui/src/main/settings.ts` line 166 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96Pkgm7vRqEcmqQzxu:-N9GpKKn9LlR3j0F5WQ9:box4r3m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/settings.ts#L166)):*

FYI all these methods should use dynamic getter in my book but I guess its not easy to express that in TS? Hence a lot of hardcode.
</blockquote></details>

There's no simple way to do that in TS as far as I'm aware.
___
*[`gui/src/main/user-interface.ts` line 65 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96IyupCcwqlRg4_Zhk:-N9GpUMf3cCMsvLdXFdr:brfvw65) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L65)):*

How can `windowController` be `null`?  It's defined as:

```
private windowController: WindowController;
```
</blockquote></details>

Too much copy-paste. Done.
___
*[`gui/src/main/user-interface.ts` line 85 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96J__c8tsEgwHELvIZ:-N9Gp_PtBt4N8aKjLAwp:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L85)):*

Inconsistency. Unless there is a reason, perhaps you should either stick to local variable `windowController` or member `this.windowController`? Some occurrences below access member too.
</blockquote></details>

Done.
___
*[`gui/src/main/user-interface.ts` line 138 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96IlHq6xREAq5FZ55A:-N9Gpbb-2idfbgjvsEd1:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L138)):*

`this.windowController` is always `true`.
</blockquote></details>

Done.
___
*[`gui/src/main/user-interface.ts` line 267 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96L6Qu0jG_uvZ1SwMc:-N9GpcfDEax2dCNIUfaf:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L267)):*

You could just give it `options` instead of spreading it. Also `default` case does exactly that, so perhaps case can be made to remove the `linux` branch?
</blockquote></details>

Done.
___
*[`gui/src/main/user-interface.ts` line 417 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96M7pDC4vpycE1ROlH:-N9Gpfi1-ny-eCfkZ9M2:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L417)):*

The `if` condition engulfs the entire function body, perhaps you could use early return instead?

```js
if (this.delegate.getUnpinnedWindow()) {
  return;
}

// Confiigure event monitor
```
</blockquote></details>

Done.
___
*[`gui/src/main/user-interface.ts` line 437 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96LfzoCGgRqor1Gf9c:-N9GphAXA7uBm9MBB-rJ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L437)):*

Same here, if statements engulfs the function body.
</blockquote></details>

Done.
___
*[`gui/src/main/user-interface.ts` line 629 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96MoMMA6ZoAEELJoHO:-N9GpmoMCff8YYTRuAiz:bdhalh5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L629)):*

Method above captures `this` with fat arrow closure but this one doesn't. Is it correct?
</blockquote></details>

This just forwards a method from the delagate so it's IMO up to the delegate to decide on whether or not it should capture `this`, but I'll change it for consistensy.
___
*[`gui/src/main/window-controller.ts` line 19 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96NExDCEYJXIq5x2Wz:-N9GpxUp2w87tMl0Z_uN:b-wputss) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/window-controller.ts#L19)):*

FYI maybe we should expand that definition to `getTrayBounds(): Rectangle`. Kinda looks funny otherwise, but I understand where you come from.
</blockquote></details>

I decided on going this way to have it match the Electron API exactly. If Electron where to change the types, although very unlikely, it would require changes in less places.
___
*[`gui/src/main/window-controller.ts` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96Nhgx9lKDb9N6sv0l:-N9Gq9Q03EoTaHxvsZZ5:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/window-controller.ts#L20)):*

Maybe should be `isUnpinnedWindow(): boolean`? 
</blockquote></details>

Done.

pronebird (changes requested):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[`gui/src/main/index.ts` line 66 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96StgkDt1ifDzYbM2P:-N9HpyU9DDfHwxpZ5VjS:b-vgq2t4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/index.ts#L66)):*

I don't get what you're suggestion, could you clarify? The `argv` in the handler of `seconds-instance` isn't the same as this one, it's the argv of the other process provided as a parameter.
</blockquote></details>

I know that `second-instance` is called with different args. I'd just use `process.argv.includes(..)` directly as opposed to defining them as constants. Last time I checked I think most of these are only referenced once during app initialization.

So in other words, I would use `process.argv.includes(CommandLineOptions...)` directly instead where needed instead of defining these as constants. 

I'll mark it as non-blocking issue.
___
*[`gui/src/main/index.ts` line 917 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96Rdou7mLJPB2dL-ZL:-N9HqzIj2W09vknHn-_A:b9yedf1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/index.ts#L917)):*

Janito wrote this code in 2018 so I don't know the reasoning. But we don't support any other platforms so it shouldn't be needed.
</blockquote></details>

 Ugh.. Maybe there is no tray on other platforms. Otherwise we could have replaced this with:

```
return this.settings.gui.unpinnedWindow && !this.settings.gui.startMinimized;
```
___
*[`gui/src/main/user-interface.ts` line 267 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96L6Qu0jG_uvZ1SwMc:-N9HrlApCr810C6nBL5q:bx21ngs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L267)):*

Done.
</blockquote></details>

Mhm `default` case does exactly the same thing! Just remove `case 'linux'` altogether! :-) Also `default` case has parenthesis for some reason `{ }`

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820)* status: 10 of 15 files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`gui/src/main/index.ts` line 66 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96StgkDt1ifDzYbM2P:-N9klgnr4lftUr_cPGqm:b-a2x6a7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/index.ts#L66)):*

I know that `second-instance` is called with different args. I'd just use `process.argv.includes(..)` directly as opposed to defining them as constants. Last time I checked I think most of these are only referenced once during app initialization.

So in other words, I would use `process.argv.includes(CommandLineOptions...)` directly instead where needed instead of defining these as constants. 

I'll mark it as non-blocking issue.
</blockquote></details>

Done, I inlined those that are used only once.
___
*[`gui/src/main/index.ts` line 917 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96Rdou7mLJPB2dL-ZL:-N9kloXQ-_QHKXRD8Hec:b1jrml9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/index.ts#L917)):*

 Ugh.. Maybe there is no tray on other platforms. Otherwise we could have replaced this with:

```
return this.settings.gui.unpinnedWindow && !this.settings.gui.startMinimized;
```
</blockquote></details>

Yeah I changed it, if this code were to run on some other platform it would still default to the correct value.
___
*[`gui/src/main/user-interface.ts` line 267 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-N96L6Qu0jG_uvZ1SwMc:-N9km28V1T55ghAoO5d-:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0b3a73beab0bfb4bde68180eeb4917bf5397ec9b/gui/src/main/user-interface.ts#L267)):*

Mhm `default` case does exactly the same thing! Just remove `case 'linux'` altogether! :-) Also `default` case has parenthesis for some reason `{ }`
</blockquote></details>

Done.

pronebird (approved):
Perhaps more can be done to make the architecture more elegant, but perhaps not in one go. :lgtm: 

> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-:-NA5Owlt3_5-9Am29Bzn:b1o0zxc)* status: :shipit: complete! all files reviewed, all discussions resolved

raksooo (commented):
Yeah, there's definitely improvements and further refactoring to do in the future, but this is a good start that makes the remaining work easier to get started with.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3820#-:-NA8WVNVEqREDa6a53Xv:b-eykq3q)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3823, opened 2022-08-09
Improve account data updates

pronebird:
1. Throttle account data updates from UI with 10s cooldown when account has already expired, otherwise use 60s interval between requests.
2. Update account data as scene becomes active. If settings are on screen, do the regular update, if not then only update if account is close to expiry or already expired, otherwise do nothing. This is done this way to avoid spamming our backend when we don't really do anything with account expiry.
3. Add calls to update account data when navigating to account view and back to root settings. (See updated `SettingsNavigationController` and `SettingsNavigationControllerDelegate`)
4. Add additional method override to be able to tell when `CustomNavigationControler` is about to pop navigation item:

   ```swift
   func willPop(navigationItem: UINavigationItem)
   ```

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3823)


AndreasShortcut (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3823)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`ios/MullvadVPN/AccountDataThrottling.swift` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3823#-N9G9mun4iH2AbYYi7qg:-N9G9muo6vsMlWavGsWZ:b1qtgjq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3a5c3ec6514630f51a246ce0c50d883f203db982/ios/MullvadVPN/AccountDataThrottling.swift#L20)):*
> ```Swift
> 
>     /// Interval in days when account is considered to be close to expiry.
>     private static let closeToExpiryDays = 4
> ```

I think these don't have to be static? That way we don't have to use `Self.` when calling

PR #3825, opened 2022-08-10
Fix universal build after last npm audit fix

raksooo:
When we last ran `npm audit fix` some transitive dependency of `electron-builder` was updated which now fires the `afterPack` event for the universal packaging as well on macOS. This caused our build script to fail. This PR disables the code that checks that all resources exists for the universal packaging since it doesn't have any resources of its own.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3825)

### DIFF
--- a/gui/tasks/distribution.js
+++ b/gui/tasks/distribution.js
@@ -1,6 +1,7 @@
 const path = require('path');
 const fs = require('fs');
 const builder = require('electron-builder');
+const { Arch } = require('electron-builder');
 const parseSemver = require('semver/functions/parse');
 const { notarize } = require('electron-notarize');
 const { version } = require('../package.json');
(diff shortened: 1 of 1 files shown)


raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3825)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`gui/tasks/distribution.js` line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3825#-N96omDRAbnvwy3XrbEc:-N96u2mC1Uz7FjLRM8wu:bqivg41) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a78e5ed296961ce961581c69c94ead5e8f0ca124/gui/tasks/distribution.js#L53)):*


Please replace magic value with `Arch.universal`:

```
import { Arch } from 'electron-builder';
```

Then use `Arch.uniiversal` instead:

```
if (context.arch !== Arch.universal) {
  ... 
}
```

Documentation for context: https://www.electron.build/configuration/configuration.html#afterpackcontext

</blockquote></details>

Good idea. Done :+1:

PR #3846, opened 2022-08-17
Add a margin to the max MTU of the tunnel

Jontified:
Some users experienced issues when the new update increased the default MTU size for packets from 1380 to 1420. This is likely due to some program causing packets to exceed the device limit (usually 1500) and then fragmenting the packets. This PR reduces the mullvad interface MTU by a buffer of 60 bytes which means that even if some program causes a packet to be bigger as long as this increase is not bigger than 60 bytes it should not fragment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846)

### DIFF
--- a/talpid-core/src/tunnel/mod.rs
+++ b/talpid-core/src/tunnel/mod.rs
@@ -223,8 +223,16 @@ impl TunnelMonitor {
         })
     }
 
+    /// Set the MTU in the tunnel parameters based on the inputted device MTU and some
+    /// calculations. `peer_mtu` is the detected device MTU.
     #[cfg(any(target_os = "linux", target_os = "windows"))]
     fn set_mtu(params: &mut wireguard_types::TunnelParameters, peer_mtu: u16) {
+        // Some users experience fragmentation issues even when we take the interface MTU and
+        // subtract the header sizes. This is likely due to some program that they use which does
+        // not change the interface MTU but adds its own header onto the outgoing packets. For this
+        // reason we subtract some extra bytes from our MTU in order to give other programs some
+        // safety margin.
+        const MTU_SAFETY_MARGIN: u16 = 60;
         const IPV4_HEADER_SIZE: u16 = 20;
         const IPV6_HEADER_SIZE: u16 = 40;
         const WIREGUARD_HEADER_SIZE: u16 = 40;
@@ -234,8 +242,8 @@ impl TunnelMonitor {
                 true => IPV6_HEADER_SIZE,
             };
         // The largest peer MTU that we allow
-        const MAX_PEER_MTU: u16 = 1500;
-        // The minimum allowed MTU size for our tunnel in IPv6 is 1280
+        const MAX_PEER_MTU: u16 = 1500 - MTU_SAFETY_MARGIN;
+        // The minimum allowed MTU size for our tunnel in IPv6 is 1280 and 576 for IPv4
         const MIN_IPV4_MTU: u16 = 576;
         const MIN_IPV6_MTU: u16 = 1280;
         let min_mtu = match params.generic_options.enable_ipv6 {
@@ -248,6 +256,8 @@ impl TunnelMonitor {
         params.options.mtu = Some(tunnel_mtu);
     }
 
+    /// Detects the MTU of the device, calculates what the virtual device MTU should be and sets
+    /// that in the tunnel parameters.
     #[cfg(any(target_os = "linux", target_os = "windows"))]
     async fn assign_mtu(
         route_manager: &RouteManagerHandle,


Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`talpid-core/src/tunnel/mod.rs` line 233 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846#-N9fejV65EuMt4LyS8Ac:-N9fejV7C5d9prS-JS_3:b-gxvdgq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8c6c756539161241a64222c4ab2ea7cc44f207f0/talpid-core/src/tunnel/mod.rs#L233)):*
> ```XML
>         // reason we subtract some extra bytes from our MTU in order to give other programs some
>         // buffer room.
>         const EXTRA_MTU_BUFFER: u16 = 60;
> ```

I set this to be 60 for both IPv4 and IPv6. That means for IPv6 we'll have a default of 1360 and for IPv4 we'll have a default of 1380. We could make this protocol specific and have the result be 1380 for both but I don't see the reason for why we'd have different "buffer" sizes.
Also feel free to suggest a better name for the constant, I'm not sure if there is some better standard naming.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`talpid-core/src/tunnel/mod.rs` line 233 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846#-N9fejV65EuMt4LyS8Ac:-N9furj7DGxP4-9Q4V37:b-8db33l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8c6c756539161241a64222c4ab2ea7cc44f207f0/talpid-core/src/tunnel/mod.rs#L233)):*

I set this to be 60 for both IPv4 and IPv6. That means for IPv6 we'll have a default of 1360 and for IPv4 we'll have a default of 1380. We could make this protocol specific and have the result be 1380 for both but I don't see the reason for why we'd have different "buffer" sizes.
Also feel free to suggest a better name for the constant, I'm not sure if there is some better standard naming.
</blockquote></details>

Are we sure about the explanation above? Is it possible that the inferred MTU is simply wrong? If that's the case, it would arguably make more sense to just lower `MAX_PEER_MTU` to 1440.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846)* status: all files reviewed, 2 unresolved discussions (waiting on @Jontified)

___
*[`talpid-core/src/tunnel/mod.rs` line 248 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846#-N9g-ekIDRNppE_m10sV:-N9g-ekIDRNppE_m10sW:b-qpvjm2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8c6c756539161241a64222c4ab2ea7cc44f207f0/talpid-core/src/tunnel/mod.rs#L248)):*
> ```XML
>         const MIN_IPV4_MTU: u16 = 576;
>         const MIN_IPV6_MTU: u16 = 1280;
>         let min_mtu = match params.generic_options.enable_ipv6 {
> ```

I'm suspicious of this. Isn't what we care about the size of the outer IP header, rather than that of the encapsulated packet, which is what the above boolean is about?

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846)* status: all files reviewed, 2 unresolved discussions (waiting on @Jontified)

___
*[`talpid-core/src/tunnel/mod.rs` line 233 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846#-N9fejV65EuMt4LyS8Ac:-N9ju0Ws4pu014oAAFao:bnno14w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8c6c756539161241a64222c4ab2ea7cc44f207f0/talpid-core/src/tunnel/mod.rs#L233)):*

Are we sure about the explanation above? Is it possible that the inferred MTU is simply wrong? If that's the case, it would arguably make more sense to just lower `MAX_PEER_MTU` to 1440.
</blockquote></details>

In networking like this I think of a slab of memory to write data to when I hear "buffer". I think a better word here would be "headroom" or "safety margin" or something that does not overlap with other commonly used network code terms.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846#-:-N9k8t3pDQ-ZxEeushcQ:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`talpid-core/src/tunnel/mod.rs` line 227 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846#-N9k8axV0OYauJOcAP0h:-N9k8axV0OYauJOcAP0i:b8fr7hp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5fd085a277c69b49def4eb14f2c4e7b3e7ef118f/talpid-core/src/tunnel/mod.rs#L227)):*
> ```XML
> 
>     /// Set the MTU in the tunnel parameters based on the inputted device MTU and some
>     /// calculations. `peer_mtu` is the detected device MTU.
> ```

Thanks! :pray:

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3846)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

PR #3850, opened 2022-08-20
Add shortcuts

nchlscs:
PR adds setting to add, edit and delete shortcuts from the app.
It seems this is a more natural way to add shortcuts for most users rather than configure each of them manually via Shortcuts app on every device.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3850)

### DIFF
--- a/ios/CHANGELOG.md
+++ b/ios/CHANGELOG.md
@@ -32,6 +32,7 @@ Line wrap the file at 100 chars.                                              Th
 - Add ability to manage registered devices if too many devices detected during log-in.
 - Add intents: start VPN, stop VPN, reconnect VPN (acts as start VPN when the tunnel is down,
   otherwise picks new relay).
+- Add menu item to control shortcuts.
 
 ### Changed
 - When logged into an account with no time left, a new view is shown instead of account settings,
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -307,6 +307,9 @@
 		58FEEB46260A028D00A621A8 /* GeoJSON.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58FEEB45260A028D00A621A8 /* GeoJSON.swift */; };
 		58FEEB58260B662E00A621A8 /* AutomaticKeyboardResponder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58FEEB57260B662E00A621A8 /* AutomaticKeyboardResponder.swift */; };
 		58FF2C03281BDE02009EF542 /* SettingsManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58FF2C02281BDE02009EF542 /* SettingsManager.swift */; };
+		75FD0C2128B108570021E33E /* ShortcutsDataSource.swift in Sources */ = {isa = PBXBuildFile; fileRef = 75FD0C2028B108570021E33E /* ShortcutsDataSource.swift */; };
+		75FD0C2328B109860021E33E /* ShortcutsDataSourceDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 75FD0C2228B109860021E33E /* ShortcutsDataSourceDelegate.swift */; };
+		75FD0C2528B117D30021E33E /* ShortcutsViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 75FD0C2428B117D30021E33E /* ShortcutsViewController.swift */; };
 		E1187ABC289BBB850024E748 /* OutOfTimeViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1187ABA289BBB850024E748 /* OutOfTimeViewController.swift */; };
 		E1187ABD289BBB850024E748 /* OutOfTimeContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1187ABB289BBB850024E748 /* OutOfTimeContentView.swift */; };
 		E1187ABF289BE76F0024E748 /* RESTCreateApplePaymentResponse+Localization.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1187ABE289BE76F0024E748 /* RESTCreateApplePaymentResponse+Localization.swift */; };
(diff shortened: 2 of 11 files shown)


pronebird (changes requested):
> Reviewed 5 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3850)* status: 5 of 8 files reviewed, 2 unresolved discussions (waiting on @nchlscs)

___
*[`ios/MullvadVPN/ShortcutsDataSource.swift` line 92 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3850#-NA5H7hR6EB8OgKJD8NG:-NA5H7hR6EB8OgKJD8NH:b-ov3gfh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/427dfe6fc1dec0d4355b8e6787446a2a301166a2/ios/MullvadVPN/ShortcutsDataSource.swift#L92)):*
> ```Swift
>         let item = snapshot.itemForIndexPath(indexPath)!
>         switch item {
>         case .start, .reconnect, .stop:
> ```

You can skip this switch case. It makes no difference. All items are handled the same way in this controller.
___
*[`ios/MullvadVPN/ShortcutsDataSource.swift` line 98 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3850#-NA5L-ti0LoW5JZLTslo:-NA5L-ti0LoW5JZLTslp:b2apho3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/427dfe6fc1dec0d4355b8e6787446a2a301166a2/ios/MullvadVPN/ShortcutsDataSource.swift#L98)):*
> ```Swift
>             )
>             if let cell = cell as? SettingsCell {
>                 cell.titleLabel.text = item.title
> ```

FYI I had a short conversation with a colleague of mine, and he pointed out that it would be great to add a checkmark next to already registered shortcuts. I think this is a great effort and we should get this merged. Eventually, when we have time, I think we'll add `IconTick` image in a custom accessory and perhaps refresh the list of registered voice shortcuts when the app becomes active or after adding or removing shortcuts.
___
*[`ios/MullvadVPN/ShortcutsDataSource.swift` line 152 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3850#-NA5M18n5qdMOsgvqGLd:-NA5M18n5qdMOsgvqGLe:b-89jli9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/427dfe6fc1dec0d4355b8e6787446a2a301166a2/ios/MullvadVPN/ShortcutsDataSource.swift#L152)):*
> ```Swift
>                 "SHORTCUTS_NAME_STOP_VPN",
>                 tableName: "Shortcuts",
>                 value: "Stop VPN",
> ```

FYI I think last time I tried, Siri said that it cannot "Stop VPN"... Perhaps a blacklisted command. I wonder if you had success to stop VPN using that specific wording? :-)
___
*[`ios/MullvadVPN/ShortcutsViewController.swift` line 54 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3850#-NA5FdXQE-j4rMBr_AzX:-NA5FdXR-vvgp-9AbnXp:bhzak6m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/427dfe6fc1dec0d4355b8e6787446a2a301166a2/ios/MullvadVPN/ShortcutsViewController.swift#L54)):*
> ```Swift
>             DispatchQueue.main.async {
>                 guard let self = self else { return }
>                 let controller: UIViewController
> ```

It's getting a little long and hard to follow this closure especially because of `async` wrap that adds another level of indentation. Maybe the body of this closure could be moved into a separate method, i.e:

```swift
private let logger = Logger(label: "ShortcutsDataSource")

private func handleSelectShortcut(_ shortcut: INShortcut, item: ShortcutsDataSource.Item) {
  INVoiceShortcutCenter.shared.getAllVoiceShortcuts { [weak self] voiceShortcuts, error in
      DispatchQueue.main.async {
          if let error = error {
              self?.logger.log(chainedError: AnyChainedError(error), message: "Failed to fetch voice shortcuts.")
          } else {
              self?.didReceiveVoiceShortcuts(voiceShortcuts ?? [], targetShortcut: shortcut)
          }
      }
  }
}

private func didReceiveVoiceShortcuts(_ voiceShortcuts: [INVoiceShortcut], targetShortcut: INShortcut) {
  // the guts of that closure.
}
```

pronebird (commented):
> Reviewed 3 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3850)* status: all files reviewed, 2 unresolved discussions (waiting on @nchlscs)

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3850)* status: all files reviewed, 2 unresolved discussions (waiting on @nchlscs)

___
*[`ios/MullvadVPN/ShortcutsViewController.swift` line 54 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3850#-NA5FdXQE-j4rMBr_AzX:-NA5NlNd3iKdCEE9LKQl:bstx4o2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/427dfe6fc1dec0d4355b8e6787446a2a301166a2/ios/MullvadVPN/ShortcutsViewController.swift#L54)):*
> private let logger = Logger(label: "ShortcutsDataSource")

Sorry typo on my side, the label for logger would be the name of view controller.

nchlscs:
> You can skip this switch case. It makes no difference. All items are handled the same way in this controller.

Done. Initially, I’ve added this switch for scaleability in case there are new items that don’t represent shortcuts.

> FYI I had a short conversation with a colleague of mine, and he pointed out that it would be great to add a checkmark next to already registered shortcuts. I think this is a great effort and we should get this merged. Eventually, when we have time, I think we'll add IconTick image in a custom accessory and perhaps refresh the list of registered voice shortcuts when the app becomes active or after adding or removing shortcuts.

I thought the same way lately. I did refactor view controller and data source and extracted the shortcuts retrieving part into a new manager. It allows to observe registered shortcuts and update table view state accordingly. But I’ve hit a little thing with the tick icon. The origin image was too large and looked ridiculous, so I had 3 ways to tackle this: import another asset, change how the settings cell layouts images or use a dirty trick and resize the existing image. I chose the latter. See https://github.com/mullvad/mullvadvpn-app/pull/3850/commits/c45f058265c1fa3d4deadce375f7889c00ca50d3

> FYI I think last time I tried, Siri said that it cannot "Stop VPN"... Perhaps a blacklisted command. I wonder if you had success to stop VPN using that specific wording? :-)

Well, actually, I didn’t! I didn't even think about it, interesting notice. I never use Siri as a voice assistant and always run shortcuts via Spotlight or widget. These titles are simply used as default names for shortcuts. In that case, I guess, it won’t that bad to transform these lines into something like "Stop Mullvad VPN".

> It's getting a little long and hard to follow this closure especially because of async wrap that adds another level of indentation. Maybe the body of this closure could be moved into a separate method, i.e:

Refactored.

pronebird (approved):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3850)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3850#-:-NA9TTapAxVf_ULOROTC:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird:
Thanks! Merged. 👍

pronebird:
> Done. Initially, I’ve added this switch for scaleability in case there are new items that don’t represent shortcuts.

I realized that. It's a little premature as we don't really know how that controller will look like in the coming months.

>  The origin image was too large and looked ridiculous, so I had 3 ways to tackle this: import another asset, change how the settings cell layouts images or use a dirty trick and resize the existing image. I chose the latter.

That's fine. I've added a script to produce the right asset in https://github.com/mullvad/mullvadvpn-app/pull/3857.

> Well, actually, I didn’t! I didn't even think about it, interesting notice. I never use Siri as a voice assistant and always run shortcuts via Spotlight or widget. These titles are simply used as default names for shortcuts. In that case, I guess, it won’t that bad to transform these lines into something like "Stop Mullvad VPN".

Perhaps. "Disconnect VPN" should also work. The default vocal phrases are not set in stone yet and perhaps we'll revise them later on.

nchlscs:
> That's fine. I've added a script to produce the right asset in https://github.com/mullvad/mullvadvpn-app/pull/3857.

Ok, thanks

PR #3853, opened 2022-08-22
Add support for Linux ARM64

dlon:
This makes it possible to build the app for Linux ARM. It can be done either via cross-compilation or by building it on an ARM host (with the caveat that `grpc-tools` will refuse to generate code for the management interface .proto (https://github.com/mullvad/mullvadvpn-app/pull/2500) in the latter case).

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3853)

### DIFF
--- a/gui/tasks/distribution.js
+++ b/gui/tasks/distribution.js
@@ -12,6 +12,13 @@ const noAppleNotarization = process.argv.includes('--no-apple-notarization');
 const universal = process.argv.includes('--universal');
 const release = process.argv.includes('--release');
 
+const targetsIndex = process.argv.indexOf('--targets');
+let targets = null;
+
+if (targetsIndex !== -1) {
+  targets = process.argv[targetsIndex + 1];
+}
+
 const config = {
   appId: 'net.mullvad.vpn',
   copyright: 'Mullvad VPN AB',
@@ -147,17 +154,26 @@ const config = {
   },
 
   linux: {
-    target: ['deb', 'rpm'],
+    target: [
+      {
+        target: 'deb',
+        arch: getLinuxTargetArch(),
+      },
+      {
+        target: 'rpm',
+        arch: getLinuxTargetArch(),
+      },
+    ],
     artifactName: 'MullvadVPN-${version}_${arch}.${ext}',
     category: 'Network',
     icon: distAssets('icon.icns'),
     extraFiles: [{ from: distAssets('linux/mullvad-gui-launcher.sh'), to: '.' }],
     extraResources: [
-      { from: distAssets('mullvad-problem-report'), to: '.' },
-      { from: distAssets('mullvad-daemon'), to: '.' },
-      { from: distAssets('mullvad-setup'), to: '.' },
-      { from: distAssets('libtalpid_openvpn_plugin.so'), to: '.' },
-      { from: distAssets('binaries/x86_64-unknown-linux-gnu/openvpn'), to: '.' },
+      { from: distAssets(path.join(getLinuxTargetSubdir(), 'mullvad-problem-report')), to: '.' },
+      { from: distAssets(path.join(getLinuxTargetSubdir(), 'mullvad-daemon')), to: '.' },
+      { from: distAssets(path.join(getLinuxTargetSubdir(), 'mullvad-setup')), to: '.' },
+      { from: distAssets(path.join(getLinuxTargetSubdir(), 'libtalpid_openvpn_plugin.so')), to: '.' },
+      { from: distAssets(path.join('binaries', '${env.TARGET_TRIPLE}', 'openvpn')), to: '.' },
       { from: distAssets('linux/mullvad-daemon.service'), to: '.' },
     ],
   },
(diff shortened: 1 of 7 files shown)


pinkisemils (changes requested):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3853)* status: 4 of 8 files reviewed, 1 unresolved discussion (waiting on @dlon and @pinkisemils)

___
*[`build.sh` line 209 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3853#-NA7xw-n0DUYHHDFiT2J:-NA7xw-n0DUYHHDFiT2K:bfeckxq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/462c3d706d47ca54c24756fbe4d79c5b53248a13/build.sh#L209)):*
> ```Shell
>     local current_target=${1:-""}
>     local for_target_string=""
>     local stripbin="strip"
> ```

I presume this script wouldn't be supposed to be ran on anything but our own build machines? Because if it was running on an ARM host, then it would be nice if the strip binary could be specified via an environment variable or via some other way. Or the script could just detect that and use regular `strip` instead of the cross-compiled one. Or am I horribly misinterpreting what's going on here? :sweat_smile:

pinkisemils (commented):
Aside from the `strip` path, this  :lgtm_strong: !

> Reviewed 1 of 1 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3853#-:-NA7yzWW1JWh5QHW9SQi:bje3rrg)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3853)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`build.sh` line 209 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3853#-NA7xw-n0DUYHHDFiT2J:-NA8XG4lBhVrv9eTLlR3:bim5qqu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/462c3d706d47ca54c24756fbe4d79c5b53248a13/build.sh#L209)):*

I presume this script wouldn't be supposed to be ran on anything but our own build machines? Because if it was running on an ARM host, then it would be nice if the strip binary could be specified via an environment variable or via some other way. Or the script could just detect that and use regular `strip` instead of the cross-compiled one. Or am I horribly misinterpreting what's going on here? :sweat_smile: 
</blockquote></details>

Done. You don't have to specify `TARGETS` if host equals the target, but it should work regardless.

gui/tasks/distribution.js
```diff
@@ -328,6 +359,27 @@ function root(relativePath) {
...
+    if (targets === 'aarch64-unknown-linux-gnu') {
+      return 'arm64';
+    }
+    throw new Error(`Invalid or unknown target (only one may be specified)`);
+  }
+  // Use host architecture.
+  return undefined;
+}
+
+function getLinuxTargetSubdir() {
```
raksooo, line 373: For macOS we always provide a target when building the rust binaries resulting in them being placed in a target directory. This makes it possible to always just use the target triple in the path, e.g. `{ from: distAssets(path.join('${env.TARGET_TRIPLE}', 'mullvad')), to: '.' }`. What do you think about doing the same here?

This is the call in `build.sh` when no target is provided:
```
    if [[ "$(uname -s)" == "Darwin" ]]; then
        # Provide target for non-universal macOS builds to use the same output location as for
        # universal builds
        build "$ENV_TARGET"
    else
        build
    fi
```

gui/tasks/distribution.js
```diff
@@ -304,6 +320,21 @@ function packLinux() {
     targets: builder.Platform.LINUX.createTarget(),
     config: {
       ...config,
+      beforeBuild: (options) => {
+        switch (options.arch) {
+          case 'x64':
+            process.env.TARGET_TRIPLE = 'x86_64-unknown-linux-gnu';
```
raksooo, line 326: `env.TARGET_TRIPLE` doesn't seem to be used anywhere for Linux? I wouldn't mind using it as on macOS though.

Also would it be possible to extract this code and the very similar code for macOS to a function used for both?

gui/tasks/distribution.js
raksooo: After discussing this in the office David conviced me that this approach doesn't work due to file patterns not being supported in fpm arguments.

pinkisemils (approved):
> Reviewed 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3853)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`build.sh` line 214 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3853#-NA8k7KQDTQvWotpVeRW:-NA8k7KQDTQvWotpVeRX:b7o10l2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/af03e5720c0f205cb3bd017552b3078ab7d357dd/build.sh#L214)):*
> ```Shell
> 
>         local arch="$(uname -m)"
>         if [[ "$current_target" == "aarch64-unknown-linux-gnu" && "${arch,,}" != "aarch64" ]]; then
> ```

I think in this case the `arch` variable might well be inlined like `"$(uname -m)" != "aarch64"`.

pinkisemils (approved):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3853)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

PR #3855, opened 2022-08-23
Don't disconnect on GUI crash or when killed

raksooo:
This PR changes the behavior of the electron app to not disconnect when it's killed or crashes. The only way to disconnect on quit is when the user presses the quit-button.

This made it possible to remove the `--quit-without-disconnect` flag and to kill the app before the daemon on Linux and to not force kill it on Windows. Unfortunately we still need to keep the code that kills the previous app in `before-install.sh` and the corresponding part in `installer.nsh` to support upgrading from older versions of the app.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3855)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@ Line wrap the file at 100 chars.                                              Th
 - Reorganize settings into more logical categories.
 - Upgrade wireguard-go to 20220703234212 (Windows: v0.5.3).
 - Prune bridges far away from the selected relay.
+- Stay connected when desktop app is killed or crashes. The only situation where the app now
+  disconnects on quit is when the user presses the quit button.
 
 #### Windows
 - Remove dependency on `ipconfig.exe`. Call `DnsFlushResolverCache` to flush the DNS cache.
--- a/dist-assets/linux/before-install.sh
+++ b/dist-assets/linux/before-install.sh
@@ -11,6 +11,8 @@ if which systemctl &> /dev/null; then
     fi
 fi
 
+# This can be removed when 2022.4 is unsupported. That version is the last version where
+# before-remove.sh doesn't kill the GUI on upgrade.
 pkill -x "mullvad-gui" || true
 
 rm -f /var/cache/mullvad-vpn/relays.json
--- a/dist-assets/linux/before-remove.sh
+++ b/dist-assets/linux/before-remove.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -eu
 
+pkill -x "mullvad-gui" || true
+
 is_number_re='^[0-9]+$'
 # Check if we're running during an upgrade step on Fedora
 # https://fedoraproject.org/wiki/Packaging:Scriptlets#Syntax
@@ -17,8 +19,6 @@ fi
 # the user might've disabled or stopped the service themselves already
 systemctl stop mullvad-daemon.service || true
 systemctl disable mullvad-daemon.service || true
-
-pkill -x "mullvad-gui" || true
 
 /opt/Mullvad\ VPN/resources/mullvad-setup reset-firewall || echo "Failed to reset firewall"
 /opt/Mullvad\ VPN/resources/mullvad-setup remove-device || echo "Failed to remove device from account"
--- a/dist-assets/windows/installer.nsh
+++ b/dist-assets/windows/installer.nsh
@@ -1155,7 +1155,7 @@
 
 	Pop $FullUninstall
 
-	nsExec::Exec '"$INSTDIR\Mullvad VPN.exe" --quit-without-disconnect' $0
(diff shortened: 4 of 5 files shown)


hankolsen (approved):

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3855)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`dist-assets/linux/before-install.sh` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3855#-NA9ZJOo7tFJLggSfFaH:-NA9bu8h2qdtINwYEHFb:bbuey0v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d8223d38b38113633f1c759011230b0265e6849a/dist-assets/linux/before-install.sh#L14)):*

2022.4 has already been released. So the comment is not entirely correct.
</blockquote></details>

Assuming this change makes it into 2022.5, it will work correctly for all versions 2022.5+ including 2022.5. Therefore 2022.4 will be the last version that this doesn't work for. Or am I missing something? :slightly_smiling_face: 
___
*[`dist-assets/windows/installer.nsh` line 858 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3855#-NA9YISI9re4j6JZ60nZ:-NA9cNzy-XnEjmG5kjbD:b-fi63wf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d8223d38b38113633f1c759011230b0265e6849a/dist-assets/windows/installer.nsh#L858)):*

It looks like we skip this above if the previously installed version was <= 2021.2. So can't this be lowered to 2021.2?
</blockquote></details>

I removed the entire comment since it's not relevant due to that check. We could write a comment that we can remove the whole `customCheckAppRunning` macro when 2021.2 has been unsupported, but I think the code explains that pretty well and since we already have this check in place it feels like we shouldn't rush to remove it.

dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3855)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[`dist-assets/linux/before-install.sh` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3855#-NA9ZJOo7tFJLggSfFaH:-NA9eYc11bmbMEYPrQ_j:b34fr1b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d8223d38b38113633f1c759011230b0265e6849a/dist-assets/linux/before-install.sh#L14)):*
> That version [2022.4] is the first version where before-remove.sh kill the GUI on upgrade.

Won't 2022.5 be the first version where before-remove.sh will do that?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3855)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[`dist-assets/windows/installer.nsh` line 858 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3855#-NA9YISI9re4j6JZ60nZ:-NA9h4gY4VlnnAnXV3sl:brxp9sq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d8223d38b38113633f1c759011230b0265e6849a/dist-assets/windows/installer.nsh#L858)):*

I removed the entire comment since it's not relevant due to that check. We could write a comment that we can remove the whole `customCheckAppRunning` macro when 2021.2 has been unsupported, but I think the code explains that pretty well and since we already have this check in place it feels like we shouldn't rush to remove it.
</blockquote></details>

Sounds good.

raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3855)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`dist-assets/linux/before-install.sh` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3855#-NA9ZJOo7tFJLggSfFaH:-NA9tu6f8ZEoDQWbBYIf:b6a1jyu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d8223d38b38113633f1c759011230b0265e6849a/dist-assets/linux/before-install.sh#L14)):*

> That version [2022.4] is the first version where before-remove.sh kill the GUI on upgrade.

Won't 2022.5 be the first version where before-remove.sh will do that?
</blockquote></details>

Oh sorry, I misinterpreted your comment. Fixed!

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3855)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3882, opened 2022-08-31
Handle tunnel sleep cycle

pronebird:
1. Add condition to make sure that we don't attempt to evaluate connection to `.retryHeartbeatPing` when heartbeat is suspended.
2. Stop tunnel monitoring timer when tunnel process enters sleep and resume it on wake. This should also reset the `timeoutReference` and start counting the ping timeout relative to the time when the process woke up instead of the date stored from before entering sleep.


AndreasShortcut (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3882)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3893, opened 2022-09-02
Fix logout dialog flickering

raksooo:
This PR fixes the logout dialog switching to the text that should apply in the `confirm` stage when setting `logoutDialogStage` to `hidden`. It should show the text corresponding to the previous stage while closing. To solve this I separated the visibility value and the stage value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3893)

### DIFF
--- a/gui/src/renderer/components/Account.tsx
+++ b/gui/src/renderer/components/Account.tsx
@@ -42,11 +42,14 @@ interface IProps {
 }
 
 interface IState {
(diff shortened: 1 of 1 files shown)


hankolsen (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3893)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3906, opened 2022-09-06
Make dispatch workflow work with verify script

Jontified:
Dispatch workflow could previously not run the verification script
correctly since it could only get the appropriate commit depth when a
pull request was made. Change this so that a dispatched workflow has a
input for the commit depth which defaults to 50. This should make it
play ball for testing purposes which is when we need dispatch workflow
to function.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3906)

### DIFF
--- a/.github/workflows/verify-locked-down-signatures.yml
+++ b/.github/workflows/verify-locked-down-signatures.yml
@@ -38,7 +38,8 @@ jobs:
                 echo "git rev-list origin/master --count after: "
                 git rev-list origin/master --count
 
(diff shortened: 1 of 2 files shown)


faern (commented):
Please evaluate `git fetch --shallow-exclude=master` as a potential alternative.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3906#-:-NBM-rqnBrWJR-mxkOGx:b-8lvcb0)* status: 0 of 1 files reviewed, all discussions resolved

PR #3943, opened 2022-09-20
MUL22-02 for macOS - maintain blocking firewall rules during system shutdown. 

pinkisemils:
Since detecting a shutdown event on macOS would require us to add some bindings to `core-foundation`, for now on macOS the daemon will always leave blocking rules when exiting.


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3943)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3946, opened 2022-09-20
Update post quantum experiment V0 -> V1

faern:
Work in progress! Coming up with the new `PostQuantumSecure` proto definition for the next iteration of the post quantum secure tunnels.

The changes/improvements from the last/current iteration include:

* Change the used CME variant to `Classic-McEliece-460896f`. This has significantly smaller keys that take significantly less time to generate. Feedback from the people behind the CME algorithm tell us this is safe enough.
* Make the protocol allow multiple KEMs in the same key exchange request. This allows future apps to mix two types of PQ secure KEMs in case one of them proves to be weak over time.

### Why the CME variant with the `f` at the end?

The variant with the “f” at the end uses a semi-systematic matrix as part of the key generation process. Empirical benchmarking has shown that the variant with suffix "f" is slightly faster in its key-generation compared to the systematic variant (the one without the "f" suffix).

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)

### DIFF
--- a/talpid-tunnel-config-client/proto/tunnel_config.proto
+++ b/talpid-tunnel-config-client/proto/tunnel_config.proto
@@ -1,15 +1,3 @@
-//
-// If you need to (re)generate the gRPC code, see prerequisites
-//
-// 	https://grpc.io/docs/languages/go/quickstart/
-//
-// and then run
-//
-// 	protoc --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative tunnel_config.proto
-//
-// from this directory.
-//
-
 syntax = "proto3";
 
 option go_package = "github.com/mullvad/wg-manager/server/tuncfg";
@@ -19,7 +7,71 @@ package tunnel_config;
 service PostQuantumSecure {
   // PskExchangeExperimentalV0 uses the common API defined by LibOQS.  See:
   // https://github.com/open-quantum-safe/liboqs
+  // This endpoint is deprecated in favor for `PskExchangeExperimentalV1`. Please use that instead.
   rpc PskExchangeExperimentalV0(PskRequestExperimentalV0) returns (PskResponseExperimentalV0) {}
+
+  // Allows deriving a preshared key (PSK) using one or multiple PQ-secure key-encapsulation
+  // mechanisms (KEM). The preshared key is added to WireGuard's preshared-key field in a new
+  // ephemeral peer (PQ-peer). This makes the tunnel resistant towards attacks using
+  // quantum computers.
+  //
+  // The VPN server associates the PQ-peer with the peer who performed the exchange. Any
+  // already existing PQ-peer for the normal peer is replaced. Each normal peer can have
+  // at most one PQ-peer.
+  //
+  // The PQ-peer is mutually exclusive to the normal peer. The server keeps both peers in memory,
+  // but only one of them is loaded into WireGuard at any point in time. A handshake from the
+  // normal peer unloads the corresponding PQ-peer from WireGuard and vice versa.
+  //
+  // A new peer is negotiated for PQ to avoid a premature break of the tunnel used for negotiation.
+  // A tunnel would break prematurely if the preshared key is applied before the normal peer
+  // received the server's contribution to the KEM exchange. This cannot occur now because
+  // the client decides when to switch to the PQ-secure tunnel. This design also allows
+  // the client to switch back to using a non-PQ-secure tunnel at any point.
+  //
+  // The negotiated PQ-peer is ephemeral. The server gives no guarantees how long it will be
+  // valid and working. The client should negotiate a new PQ-peer every time it establishes a new
+  // tunnel to the server.
+  //
+  // The full exchange requires just a single request-response round trip between the VPN client
+  // and the VPN server.
+  //
+  // # Request-response format
+  //
+  // The request from the VPN client contains:
+  //   * `wg_pubkey` - The public key used by the current tunnel (that the request travels inside).
+  //   * `wg_psk_pubkey` - A newly generated ephemeral WireGuard public key for the PQ-peer.
+  //     The server will associate the derived PSK with this public key.
+  //   * `kem_pubkeys` - A list describing the KEM algorithms. Must have at least one entry.
+  //                     The same KEM must not be listed more than once. Each list item contains:
+  //     * `algorithm_name` - The name of the KEM, including which variant. Should be the same
+  //       name/format that `liboqs` uses.
+  //     * `key_data` - The client's public key for this KEM. Will be used by the server to
+  //       encapsulate the shared secret for this KEM.
+  //
+  // The response from the VPN server contains:
+  //   * `ciphertexts` - A list of the ciphertexts (the encapsulated shared secrets) for all
+  //     public keys in `kem_pubkeys` in the request, in the same order as in the request.
+  //
+  // # Deriving the WireGuard PSK
+  //
+  // The PSK to be used in WireGuard's preshared-key field is computed by XORing the resulting
+  // shared secrets of all the KEM algorithms. All currently supported and planned to be
+  // supported algorithms output 32 bytes, so this is trivial.
+  //
+  // Since the PSK provided to WireGuard is directly fed into a HKDF, it is not important that
+  // the entropy in the PSK is uniformly distributed. The actual keys used for encrypting the
+  // data channel will have uniformly distributed entropy anyway, thanks to the HKDF.
+  // But even if that was not true, since both CME and Kyber run SHAKE256 as the last step
+  // of their internal key derivation, the output they produce are uniformly distributed.
+  //
+  // If we later want to support another type of KEM that produce longer or shorter output,
+  // we can hash that secret into a 32 byte hash before proceeding to the XOR step.
+  //
+  // Mixing with XOR (A = B ^ C) is fine since nothing about A is revealed even if one of B or C
+  // is known. Both B *and* C must be known to compute any bit in A. This means all involved
+  // KEM algorithms must be broken before the PSK can be computed by an attacker.
+  rpc PskExchangeExperimentalV1(PskRequestExperimentalV1) returns (PskResponseExperimentalV1) {}
 }
 
 message PskRequestExperimentalV0 {
@@ -35,4 +87,19 @@ message KemPubkeyExperimentalV0 {
 
 message PskResponseExperimentalV0 {
   bytes ciphertext = 1;
+}
+
+message PskRequestExperimentalV1 {
+  bytes wg_pubkey = 1;
+  bytes wg_psk_pubkey = 2;
+  repeated KemPubkeyExperimentalV1 kem_pubkeys = 3;
+}
+
+message KemPubkeyExperimentalV1 {
+  string algorithm_name = 1;
+  bytes key_data = 2;
+}
+
+message PskResponseExperimentalV1 {
+  repeated bytes ciphertexts = 1;
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3232,6 +3232,7 @@ dependencies = [
  "rand 0.8.5",
  "serde",
  "x25519-dalek",
+ "zeroize",
 ]
 
 [[package]]
--- a/talpid-tunnel-config-client/Cargo.toml
+++ b/talpid-tunnel-config-client/Cargo.toml
@@ -15,7 +15,7 @@ tonic = "0.8"
 prost = "0.11"
 tower = "0.4"
 tokio = "1"
-classic-mceliece-rust = { version = "2.0.0", features = ["mceliece8192128f"] }
+classic-mceliece-rust = { version = "2.0.0", features = ["mceliece460896f"] }
 
 [dev-dependencies]
 tokio = { version = "1", features = ["rt-multi-thread"] }
--- a/talpid-tunnel-config-client/examples/psk-exchange.rs
+++ b/talpid-tunnel-config-client/examples/psk-exchange.rs
@@ -1,24 +1,25 @@
-use std::{
-    io,
-    net::{IpAddr, Ipv4Addr},
-};
+//! Example client implementing the quantum resistant tunnel PSK exchange.
+//! Useful to test this crate's implementation.
 
+use std::net::IpAddr;
 use talpid_types::net::wireguard::PublicKey;
 
 #[tokio::main]
 async fn main() {
-    println!("Make sure you're connected to a WireGuard peer and enter your public key: ");
+    let mut args = std::env::args().skip(1);
+    let tuncfg_server_ip: IpAddr = args
+        .next()
+        .expect("Give tuncfg server IP as first argument")
+        .parse()
+        .expect("tuncfg ip argument not a valid IP");
+    let pubkey_string = args
+        .next()
+        .expect("Give WireGuard public key as second argument");
+    let pubkey = PublicKey::from_base64(pubkey_string.trim()).expect("Invalid public key");
 
-    let mut pubkey_s = String::new();
-    io::stdin()
-        .read_line(&mut pubkey_s)
-        .expect("Failed to read from stdin");
-    let pubkey = PublicKey::from_base64(pubkey_s.trim()).expect("Invalid public key");
-
-    let (private_key, psk) =
-        talpid_tunnel_config_client::push_pq_key(IpAddr::V4(Ipv4Addr::new(10, 64, 0, 1)), pubkey)
-            .await
-            .unwrap();
+    let (private_key, psk) = talpid_tunnel_config_client::push_pq_key(tuncfg_server_ip, pubkey)
+        .await
+        .unwrap();
 
     println!("private key: {:?}", private_key);
     println!("psk: {:?}", psk);
--- a/talpid-tunnel-config-client/examples/tuncfg-server.rs
+++ b/talpid-tunnel-config-client/examples/tuncfg-server.rs
@@ -1,0 +1,84 @@
+//! A server implementation of the tuncfg PskExchangeExperimentalV1 RPC to test
+//! the client side implementation.
+
+#[allow(clippy::derive_partial_eq_without_eq)]
+mod proto {
+    tonic::include_proto!("tunnel_config");
+}
+use classic_mceliece_rust::{PublicKey, CRYPTO_PUBLICKEYBYTES};
+use proto::{
+    post_quantum_secure_server::{PostQuantumSecure, PostQuantumSecureServer},
+    PskRequestExperimentalV0, PskRequestExperimentalV1, PskResponseExperimentalV0,
+    PskResponseExperimentalV1,
+};
+use talpid_types::net::wireguard::PresharedKey;
+
+use tonic::{transport::Server, Request, Response, Status};
+
+#[derive(Debug, Default)]
+pub struct PostQuantumSecureImpl {}
+
+#[tonic::async_trait]
+impl PostQuantumSecure for PostQuantumSecureImpl {
+    async fn psk_exchange_experimental_v0(
+        &self,
+        _request: Request<PskRequestExperimentalV0>,
+    ) -> Result<Response<PskResponseExperimentalV0>, Status> {
+        unimplemented!("Use V1 instead");
+    }
+
+    async fn psk_exchange_experimental_v1(
+        &self,
+        request: Request<PskRequestExperimentalV1>,
+    ) -> Result<Response<PskResponseExperimentalV1>, Status> {
+        let mut rng = rand::thread_rng();
+        let request = request.into_inner();
+
+        println!("wg_pubkey: {:?}", request.wg_pubkey);
+        println!("wg_psk_pubkey: {:?}", request.wg_psk_pubkey);
+
+        // The ciphertexts that will be returned to the client
+        let mut ciphertexts = Vec::new();
+        // The final PSK that is computed by XORing together all the KEM outputs.
+        let mut psk_data = Box::new([0u8; 32]);
+
+        for kem_pubkey in request.kem_pubkeys {
+            println!("\tKEM algorithm: {}", kem_pubkey.algorithm_name);
+            let (ciphertext, shared_secret) = match kem_pubkey.algorithm_name.as_str() {
+                "Classic-McEliece-460896f" => {
+                    let key_data: [u8; CRYPTO_PUBLICKEYBYTES] =
+                        kem_pubkey.key_data.as_slice().try_into().unwrap();
+                    let public_key = PublicKey::from(&key_data);
+                    let (ciphertext, shared_secret) =
+                        classic_mceliece_rust::encapsulate_boxed(&public_key, &mut rng);
+                    (ciphertext.as_array().to_vec(), *shared_secret.as_array())
+                }
+                name => panic!("Unsupported KEM algorithm: {name}"),
+            };
+
+            ciphertexts.push(ciphertext);
+            println!("\tshared secret: {:?}", shared_secret);
+            for (psk_byte, shared_secret_byte) in psk_data.iter_mut().zip(shared_secret.iter()) {
+                *psk_byte ^= shared_secret_byte;
+            }
+        }
+
+        let psk = PresharedKey::from(psk_data);
+        println!("psk: {:?}", psk);
+        println!("==============================================");
+        Ok(Response::new(PskResponseExperimentalV1 { ciphertexts }))
+    }
+}
+
+#[tokio::main]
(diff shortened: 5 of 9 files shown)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`talpid-tunnel-config-client/examples/tuncfg-server.rs` line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NCQWwSU-KLRzXn4vZZ2:-NCU5m695e0ZBUOAHXYa:bssczk0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f28706a114bb696b99b89a7146827a8329ec862f/talpid-tunnel-config-client/examples/tuncfg-server.rs#L37)):*

The output should probably be written to `stdout`, not `stderr`.
</blockquote></details>

True! This is mostly for testing. But if we clean it up a bit we can merge it for sure. I'll change!
___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 22 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NCQZyif0G7GdmacZ7YR:-NCU5tUg0w0zPF7tZH00:bqjxwls) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f28706a114bb696b99b89a7146827a8329ec862f/talpid-tunnel-config-client/proto/tunnel_config.proto#L22)):*

Is the plan to keep supporting `V0` and deprecate/remove it at a later time?  Otherwise, maybe we could remove it.
</blockquote></details>

The app does not need to support `V0`. But the server should support it in a transition period. So users who use this experimental feature can seamlessly upgrade to apps using `V1` without having too much interruptions. There is no real reason to remove `V0` before/at the same time as creating `V1` IMO. Better to keep compatibility where it costs us nothing. We will of course remove all the experimental versions of this a while after deploying the stable RPC method to production.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 22 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NCQZyif0G7GdmacZ7YR:-NCUiShZ4K88oOdtVKXN:brtgpi9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f28706a114bb696b99b89a7146827a8329ec862f/talpid-tunnel-config-client/proto/tunnel_config.proto#L22)):*

The app does not need to support `V0`. But the server should support it in a transition period. So users who use this experimental feature can seamlessly upgrade to apps using `V1` without having too much interruptions. There is no real reason to remove `V0` before/at the same time as creating `V1` IMO. Better to keep compatibility where it costs us nothing. We will of course remove all the experimental versions of this a while after deploying the stable RPC method to production.
</blockquote></details>

Alright. The tunnel state machine will currently just retry if any kind of RPC error occurs. Maybe it would make sense to not do for deprecated functions.

faern:
Please note that the latest version adds kyber into the mix. This is **not intended to be merged**! This is just to showcase how the protocol can be used to mix multiple KEMs

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 8 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NCVB4DB3deFf3_5TBQy:-NCVB4DB3deFf3_5TBQz:bclucdb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7b7e13059c1d839d41d2dd8728235111ad648dd/talpid-tunnel-config-client/proto/tunnel_config.proto#L8)):*
> ```protobuf
> // and then run
> //
> // 	protoc --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative tunnel_config.proto
> ```

I don't think the server side should document things like this here in the proto file. Probably belongs in their internal documentation. Since the language used to implement the server is not relevant to the protocol itself.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: 1 of 8 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 22 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NCQZyif0G7GdmacZ7YR:-NCVCNvX4lOcGTueRBxX:bgc15lr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f28706a114bb696b99b89a7146827a8329ec862f/talpid-tunnel-config-client/proto/tunnel_config.proto#L22)):*

Alright. The tunnel state machine will currently just retry if any kind of RPC error occurs. Maybe it would make sense to not do for deprecated functions.
</blockquote></details>

It's fine. The current version is an experiment and we have not promised any stability or good behavior from the app for the users who opt into it :) Hopefully the vast majority have upgraded to an app using `V1` before we remove `V0` from the servers.

dlon (commented):
> Reviewed 8 of 8 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: all files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 101 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NCVQUA21SJvJyLC8YSR:-NCVQUA21SJvJyLC8YSS:b-fjgu21) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bab1afa11928fa1be31ae4d0522a37f6645b2c3a/talpid-tunnel-config-client/proto/tunnel_config.proto#L101)):*
> ```protobuf
> }
> 
> message PskResponseExperimentalV1 {
> ```

Do we need a way to tell the client which algorithms were rejected? For example, if some algorithm is removed, the client might want to stop using it.
___
*[`talpid-tunnel-config-client/src/kyber.rs` line 6 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NCVMo4P5Wbr0TTDVdCZ:-NCVMo4Q31o0FNnX2X_q:bvive54) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bab1afa11928fa1be31ae4d0522a37f6645b2c3a/talpid-tunnel-config-client/src/kyber.rs#L6)):*
Might be silly if the chance of failure is only `2^-142`, it wouldn't hurt to make this fallible. :)

Unwrapping is inconsistent with the following:

> ignore it and let the handshake fail in such cases, leading the app to just try again.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 101 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NCVQUA21SJvJyLC8YSR:-NCVTGNY7In-lWIuRhBq:by3uczv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bab1afa11928fa1be31ae4d0522a37f6645b2c3a/talpid-tunnel-config-client/proto/tunnel_config.proto#L101)):*

Do we need a way to tell the client which algorithms were rejected? For example, if some algorithm is removed, the client might want to stop using it.
</blockquote></details>

We don't intend to have any type of protocol negotiation. That's complex and it can open up for downgrade attacks. The server is supposed to support all KEMs used by all apps that we still support. If we need to stop supporting a particular KEM we need to wait until we stop supporting the apps that use it.
___
*[`talpid-tunnel-config-client/src/kyber.rs` line 6 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NCVMo4P5Wbr0TTDVdCZ:-NCVTY98DLYphpviGYOT:bh5p7i8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bab1afa11928fa1be31ae4d0522a37f6645b2c3a/talpid-tunnel-config-client/src/kyber.rs#L6)):*

Might be silly if the chance of failure is only `2^-142`, it wouldn't hurt to make this fallible. :)

Unwrapping is inconsistent with the following:

> ignore it and let the handshake fail in such cases, leading the app to just try again.
</blockquote></details>

If this was to be merged I would really make this way better. I see this as just PoC code showing that the protocol allows mixing multiple KEMs.

I'm not even sure we would be using `pqc_kyber` if we were to suport Kyber in the future.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 101 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NCVQUA21SJvJyLC8YSR:-NCWre0mCF_dqtcfbPuJ:b-x5ij3g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bab1afa11928fa1be31ae4d0522a37f6645b2c3a/talpid-tunnel-config-client/proto/tunnel_config.proto#L101)):*

We don't intend to have any type of protocol negotiation. That's complex and it can open up for downgrade attacks. The server is supposed to support all KEMs used by all apps that we still support. If we need to stop supporting a particular KEM we need to wait until we stop supporting the apps that use it.
</blockquote></details>

Alright, that makes sense.

talpid-tunnel-config-client/proto/tunnel_config.proto
```diff
@@ -19,7 +7,69 @@ package tunnel_config;
...
+  // mechanisms (KEM). The preshared key is added to WireGuard's preshared-key field in a new
+  // ephemeral peer (PQ-peer). This makes the tunnel resistant towards attacks using
+  // quantum computers.
+  //
+  // The VPN server associates the PQ-peer with the peer who performed the exchange. Any
+  // already existing PQ-peer for the normal peer is replaced. Each normal peer can have
+  // at most one PQ-peer.
+  //
+  // The PQ-peer is mutually exclusive to the normal peer. A handshake from the normal peer
+  // throws out the corresponding PQ-peer from WireGuard and vice versa.
```
mvd-ows, line 23: This could be clarified. When the PQ-peer first tries to handshake, the associated normal peer is unloaded from the WireGuard interface but remembered by `wg-manager`. If at a later time, the normal peer again tries to handshake, the PQ-peer is unloaded and remembered. A client can switch back and forth between the peers without having to renegotiate, for as long as the wg-manager instance stays alive.

talpid-tunnel-config-client/proto/tunnel_config.proto
```diff
@@ -19,7 +7,69 @@ package tunnel_config;
...
+  //
+  // A new peer is negotiated for PQ to avoid a premature break of the tunnel used for negotiation.
+  // A tunnel would break prematurely if the preshared key is applied before the normal peer
+  // received the server's contribution to the KEM exchange. This cannot occur now because
+  // the client decides when to switch to the PQ-secure tunnel. This design also allows
+  // the client to switch back to using a non-PQ-secure tunnel at any point.
+  //
+  // The negotiated PQ-peer is ephemeral. The server gives no guarantees how long it will be
+  // valid and working. The client should negotiate a new PQ-peer every time it establishes a new
+  // tunnel to the server.
```
mvd-ows, line 33: This would be preferred. I was told that the server, by-design, has to keep the PQ-peer alive for as long as possible.

faern: What does "this" refer to here? What's the change you are suggesting? The main point I'm trying to convey to client implementors here is that the PQ-peer is not persisted on the server. That the ephemeral peer could be lost at any time.

talpid-tunnel-config-client/proto/tunnel_config.proto
faern: I'll see if I can find a good wording. I don't want to mention `wg-manager` as the protocol is not specific to `wg-manager`. But I agree we can be a bit more clear on the fact that the server should hold on to both peers, but only have one loaded into WireGuard at any single point. And that swapping between them is possible.

mvd-ows: Alright

talpid-tunnel-config-client/proto/tunnel_config.proto
mvd-ows: "This" refers to the behavior outlined in the paragraph above. `The client should negotiate ... every time` makes total sense but is in contrast to a previous paragraph where it's encouraged to switch back-and-forth between a client's normal peer and its PQ-peer.

faern: I don't think we encourage anything. We just say it's possible. And it's true, isn't it? We don't guarantee anything about how long the pq-peer will exist. But for as long as it does exist, the client can switch back and forth between the two peers freely. Not that any implementation we have will, but it's possible.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: 1 of 10 files reviewed, 2 unresolved discussions (waiting on @dlon and @mvd-ows)

___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 23 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NCQPntGU_29LzDzYgHo-r7-23:-ND7Mb-51l5ZOij-Nn8h:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6ad6c6188ced11a6cea541ae14283c904bb8912c/talpid-tunnel-config-client/proto/tunnel_config.proto#L23)):*

Alright
</blockquote></details>

Done

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: 0 of 10 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[`talpid-tunnel-config-client/examples/psk-exchange.rs` line 10 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-ND8-95G8buBU40FN3NE:-ND8-95G8buBU40FN3NF:b-wwc25c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/examples/psk-exchange.rs#L10)):*
> ```XML
> async fn main() {
>     let mut args = std::env::args().skip(1);
>     let tuncfg_server_ip: IpAddr = args
> ```

It became more flexible if the IP is not hardcoded, but rather configurable. That way it can be used towards the `tuncfg-server` example in this crate on `127.0.0.1`
___
*[`talpid-tunnel-config-client/examples/psk-exchange.rs` line 17 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-ND8-FZ-6WwSlPMRTyYf:-ND8-FZ-6WwSlPMRTyYg:bplryxs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/examples/psk-exchange.rs#L17)):*
> ```XML
>     let pubkey_string = args
>         .next()
>         .expect("Give WireGuard public key as second argument");
> ```

I'm not sure if there was a specific reason we accepted this on stdin? It's just the pubkey, so nothing secret really. Having it on the argument makes it easier to run over and over since it's in your bash history.
___
*[`talpid-tunnel-config-client/src/classic_mceliece.rs` line 3 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-ND8-WueEJwM0UrEcDHF:-ND8-WueEJwM0UrEcDHG:buz5vtd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/src/classic_mceliece.rs#L3)):*
> ```XML
> use classic_mceliece_rust::{keypair_boxed, SharedSecret};
> 
> const STACK_SIZE: usize = 2 * 1024 * 1024;
> ```

This can probably be way smaller now, with the smaller CME variant. Maybe it works on the default stack even. But I'm not sure if it's worth ripping out the separate-thread solution.. I guess I should try this on Windows.

dlon (approved):
> Reviewed 8 of 9 files at r8, 1 of 1 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: all files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[`talpid-tunnel-config-client/examples/psk-exchange.rs` line 17 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-ND8-FZ-6WwSlPMRTyYf:-NDCQHmd9sefgIIVK7Zw:b-u497t1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/examples/psk-exchange.rs#L17)):*

I'm not sure if there was a specific reason we accepted this on stdin? It's just the pubkey, so nothing secret really. Having it on the argument makes it easier to run over and over since it's in your bash history.
</blockquote></details>

No reason other than that I found it convenient.
___
*[`talpid-tunnel-config-client/examples/tuncfg-server.rs` line 56 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NDC_WZV5tkJvpv_QHID:-NDC_WZV5tkJvpv_QHIE:b1qhler) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/examples/tuncfg-server.rs#L56)):*
> ```XML
>                     (ciphertext.as_array().to_vec(), *shared_secret.as_array())
>                 }
>                 // "Kyber1024" => {
> ```

Do we need to keep this?
___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 94 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NDCP9GJDL-0AIkyTGaM:-NDCP9GJDL-0AIkyTGaN:bdrbu27) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/proto/tunnel_config.proto#L94)):*
> ```protobuf
>   bytes wg_pubkey = 1;
>   bytes wg_psk_pubkey = 2;
>   repeated KemPubkeyExperimentalV1 kem_pubkeys = 3;
> ```

Should it be mentioned that this must fail if `len(kem_pubkeys) == 0`.
___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 103 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NDCPTAPD7hqe1RMShnN:-NDCPTAPD7hqe1RMShnO:b-i2o5si) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/proto/tunnel_config.proto#L103)):*
> ```protobuf
> 
> message PskResponseExperimentalV1 {
>   repeated bytes ciphertexts = 1;
> ```

Should it be pointed out that the client must reject a response if `len(ciphertexts) != len(kem_pubkeys)`?

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: 8 of 10 files reviewed, all discussions resolved (waiting on @dlon)

___
*[`talpid-tunnel-config-client/examples/tuncfg-server.rs` line 56 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NDC_WZV5tkJvpv_QHID:-NDCh-6q0Tij69_yF1pE:b-x2hg22) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/examples/tuncfg-server.rs#L56)):*

Do we need to keep this?
</blockquote></details>

Maybe not. I figured this was mostly a sandbox file anyway. But it's correct that even the kyber API might change before this goes live. So we can probably re-invent this code then. Removed!
___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 94 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NDCP9GJDL-0AIkyTGaM:-NDCdsDJEb_g6xWlf_uq:b-k7fn93) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/proto/tunnel_config.proto#L94)):*

Should it be mentioned that this must fail if `len(kem_pubkeys) == 0`.
</blockquote></details>

Good idea. Done
___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 103 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NDCPTAPD7hqe1RMShnN:-NDCcqF222D0i6qYtVFP:b6s4ufp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/proto/tunnel_config.proto#L103)):*

Should it be pointed out that the client must reject a response if `len(ciphertexts) != len(kem_pubkeys)`?
</blockquote></details>

We already write in the documentation for this field that it matches the `kem_pubkeys` list 1:1. See documentation on the RPC method.

What does reject a response mean here? If the length is different it won't be able to decapsulate all shared secrets, so it by definition will not be able to compute the PSK anyway. It feels redundant to write "If you can't compute the PSK, please don't". But maybe you have a formulation in mind that is helpful to an implementer?
___
*[`talpid-tunnel-config-client/src/classic_mceliece.rs` line 3 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-ND8-WueEJwM0UrEcDHF:-NDCdxPl2m-y38GHRMyf:bddvuhl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/src/classic_mceliece.rs#L3)):*

This can probably be way smaller now, with the smaller CME variant. Maybe it works on the default stack even. But I'm not sure if it's worth ripping out the separate-thread solution.. I guess I should try this on Windows.
</blockquote></details>

I tested this with a debug build on Windows. It still requires between 800k and 1M. So we should probably still do this in a separate thread with at least double that in stack size. Even if the stack usage might be lower in release builds.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`talpid-tunnel-config-client/proto/tunnel_config.proto` line 103 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3946#-NDCPTAPD7hqe1RMShnN:-NDDmqVo9H0hzU0-ZYAf:b4qd3oq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8e6470affe5ace89f63b5c12b633a879ddf3d8d9/talpid-tunnel-config-client/proto/tunnel_config.proto#L103)):*

We already write in the documentation for this field that it matches the `kem_pubkeys` list 1:1. See documentation on the RPC method.

What does reject a response mean here? If the length is different it won't be able to decapsulate all shared secrets, so it by definition will not be able to compute the PSK anyway. It feels redundant to write "If you can't compute the PSK, please don't". But maybe you have a formulation in mind that is helpful to an implementer?
</blockquote></details>

I guess a bad implementation might compute something like `PSK = xor(decapsulate_fn[i](k) for (i,k) in enumerate(kem_pubkeys))`. But I think you're right.

faern:
Merged! I forgot to add it to the changelog. But I did in a standalone commit after :see_no_evil:  https://github.com/mullvad/mullvadvpn-app/commit/065ef2f227b5c15749bad9ca8ba1f36260f2ca5c

PR #3961, opened 2022-09-26
Fix links for third-party translation contributions

dlon:
As above.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3961)


faern (approved):
Awesome! Thanks. I forgot this when we moved over to the form.

PR #3942, opened 2022-09-20
MUL22-02 for Windows - maintain blocking firewall rules during system shutdown.

dlon:
Previously, the daemon did keep blocking rules after exit if auto-connect was enabled. This means that there's a small leak window on shutdown if auto-connect (and always require VPN) is disabled. This PR changes the behavior of the daemon to always block on exit if the system is shutting down instead.

Related: #3940.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3942)

PR #3940, opened 2022-09-19
MUL22-02 for Linux - maintain blocking firewall rules during system shutdown.

pinkisemils:
Due to the ordering of shutdown events, the mullvad daemon will most often be shut down while network interfaces are still configured and working, which allows for a time window where arbitrary processes can send network traffic outside of the tunnel, even if the user never intended for this. 

To fix this, the daemon will not clear it's firewall rules during shutdown if the user didn't explicitly disconnect before the daemon is being stopped, unless `systemd` is reachable over D-Bus and it's reporting that the system is not shutting down.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3940)

PR #3966, opened 2022-09-26
Update security.md with regards to MUL22-02 (and MUL22-04)

pinkisemils:
Due to recent changes to the daemon to prevent early and late boot leaks, the security docs need to be updated to reflect the new behavior.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3966)

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -270,11 +270,19 @@ via unix domain sockets (UDS) on Linux and macOS and via named pipes on Windows.
 This management interface can be reached by any process running on the device.
 Locally running malicious programs are outside of the app's threat model.
 
-The service transitions to the [disconnected] state before exiting (i.e., normally when the OS is
-being shut down). In general, the last firewall policy is maintained when the service exits, and
-lost upon a reboot (except on Windows, see below). In other words, if the "Always require VPN"
-option is enabled, the blocking policy will be left intact when the daemon service stops.
-Otherwise, the system firewall will be reset to its original state.
+The `mullvad-daemon` transition to the [disconnected] state before exiting. To
+limit leaks during computer shutdown, it will maintain the blocking firewall
+rules upon exit in the following scenarios:
+- _Always require VPN_ is enabled
+- A user didn't explicitly request for the `mullvad-daemon` to be shut down and
+  either or both of the following are true
+    - The daemon is currently in one of the blocking states ([connected],
+      [connecting], or [error])
+    - _auto-connect_ is enabled
+
+In other cases, when the daemon process stops normally, firewall rules will be
+removed.
+
 
 ### Windows
 
@@ -284,6 +292,13 @@ the service has started back up again during boot, including before the BFE serv
 
 As with "Always require VPN", enabling "Auto-connect" in the service will cause it to
 enforce the blocking policy before being stopped.
+
+### Linux
+
+Due to the dependence on various other services, the `mullvad-daemon` is not
+started early enough to prevent leaks. To prevent this, another system unit is
+started during early boot that applies a blocking policy that persists until the
+`mullvad-daemon` is started.
 
 ## Desktop Electron GUI
 
--- a/docs/security.md
+++ b/docs/security.md
@@ -33,8 +33,8 @@ An app with permission to act as a VPN service can request to open a VPN tunnel 
 provide a set of IP networks it would like to have routed via itself. Doing so and specifying
 the routes `0/0` and `::0/0` forces all traffic to go via the app. That is what this app does both
 when it has a VPN tunnel up, but also when in a state where it would like to block all network
-traffic. Such as the [connecting], [disconnecting] and [error] states. In these states, all
-packets are simply dropped.
+traffic. Such as the [connecting], [disconnecting] and [error] states. In these states, all outgoing
+packets are simply dropped, but incoming traffic is still allowed due to the limitations of Android.
 
 ### iOS
 


faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3966)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @pinkisemils)

___
*[`docs/security.md` line 275 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3966#-NCyrYPj44VJ-xQixBwC:-NCyrYPj44VJ-xQixBwD:bh6l20w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3777e149528010fe65495b926812a732d464575/docs/security.md#L275)):*
"it will maintain the blocking policy if the currently applied policy is blocking" tells me very little, or is very confusing.

I think we can rewrite this entire paragraph to be clearer. Let's start by swapping the order of the first sentence.

> The service will transition to the [disconnected] state before exiting. To limit leaks during computer shutdown it will maintain the blocking firewall rules upon exit if `auto-connect` or "Always require VPN" is enabled, or if the service expect that it's being killed due to a computer shutdown.

Or something like that.
___
*[`docs/security.md` line 276 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3966#-NCyr5_z0T4SLj5oelFr:-NCyr5_z0T4SLj5oelFs:b-hcojlp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3777e149528010fe65495b926812a732d464575/docs/security.md#L276)):*
> ```Markdown
> [disconnected] state before exiting, but it will maintain the blocking policy if
> the currently applied policy is blocking or if the daemon should auto-connect on
> startup, unless it can detect that the shutdown was initiated by the user. If
> ```

I don't think we should use the word shutdown too much. It's too easy to confuse between daemon shutdown and computer shutdown. When we talk about the service being terminated I think we should avoid the term "shutdown" for this reason.
___
*[`docs/security.md` line 289 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3966#-NCyseVs9kq2naCBSBJ3:-NCyseVs9kq2naCBSBJ4:b-4zjr7l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3777e149528010fe65495b926812a732d464575/docs/security.md#L289)):*
> ```Markdown
> enforce the blocking policy before being stopped.
> 
> ### Linux
> ```

Nit. But empty line after header please :pray: 
___
*[`docs/security.md` line 290 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3966#-NCysznFAKjrCwv_KSko:-NCysznFAKjrCwv_KSkp:bfjmx8q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3777e149528010fe65495b926812a732d464575/docs/security.md#L290)):*
> ```Markdown
> 
> ### Linux
> Due to the dependence on various other services, the system service is not
> ```

Maybe replace "the system service" with "`mullvad-daemon`" to make it clear what is what, since the paragraph talks about two different system services.
___
*[`docs/security.md` line 291 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3966#-NCysme80wM-IHHes95l:-NCysme91AAE4CBDk39d:b-nll688) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3777e149528010fe65495b926812a732d464575/docs/security.md#L291)):*
> ```Markdown
> ### Linux
> Due to the dependence on various other services, the system service is not
> started early enough to prevent leaks. To aid this, another system unit is
> ```

"To aid this" is IMO unclear to me. But might be my English that is the problem. I would have written. "To prevent the leaks, we have another ..."

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3966)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[`docs/security.md` line 275 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3966#-NCyrYPj44VJ-xQixBwC:-NCz0luxBaIaxX55KYQS:bwoc7zc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3777e149528010fe65495b926812a732d464575/docs/security.md#L275)):*

"it will maintain the blocking policy if the currently applied policy is blocking" tells me very little, or is very confusing.

I think we can rewrite this entire paragraph to be clearer. Let's start by swapping the order of the first sentence.

> The service will transition to the [disconnected] state before exiting. To limit leaks during computer shutdown it will maintain the blocking firewall rules upon exit if `auto-connect` or "Always require VPN" is enabled, or if the service expect that it's being killed due to a computer shutdown.

Or something like that.
</blockquote></details>

I've improved it to refer to the daemon directly, 
___
*[`docs/security.md` line 276 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3966#-NCyr5_z0T4SLj5oelFr:-NCz15bF8YCR1fdOTGdL:bxuamka) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3777e149528010fe65495b926812a732d464575/docs/security.md#L276)):*

I don't think we should use the word shutdown too much. It's too easy to confuse between daemon shutdown and computer shutdown. When we talk about the service being terminated I think we should avoid the term "shutdown" for this reason.
</blockquote></details>

I agree, I've switched to _the daemon process stops_.

faern (approved):
:lgtm: Great :+1: For reference: This updates the security document to describe the fixes implemented for `MUL22-02`. https://github.com/mullvad/mullvadvpn-app/pull/3940, https://github.com/mullvad/mullvadvpn-app/pull/3942 and https://github.com/mullvad/mullvadvpn-app/pull/3943

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3966#-:-NCz2KCgCowG7YrP5_Su:b-yfcytq)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3986, opened 2022-09-29
Setting up a dockerfile for building the desktop app

faern:
Very very WIP! And I'm probably not the one who will complete this project. But I felt we needed to share this somewhere. I know @pinkisemils also has a WIP dockerfile.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 **/*
+!gui/package.json
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,8 @@
 # And add -e TARGETS="aarch64-unknown-linux-gnu" to build for ARM64
 
 # Debian 10 is the oldest supported distro. It has the oldest glibc that we support
-FROM debian:10.13-slim@sha256:557ee531b81ce380d012d83b7bb56211572e5d6088d3e21a3caef7d7ed7f718b
+# This checksum points to a 10.13-slim image.
+FROM debian@sha256:557ee531b81ce380d012d83b7bb56211572e5d6088d3e21a3caef7d7ed7f718b
 
 # === Define toolchain versions and paths ===
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,33 +1,55 @@
 # To build the image:
-# docker build . -t quay.io/mullvad/mullvadvpn-app-build
-# To push the image to Quay.io:
-# docker push quay.io/mullvad/mullvadvpn-app-build
-FROM debian:stretch@sha256:a5934d79acb9d1182ef5c747e23e462784f6345479e33b40c979fbe8dce5db40
-RUN apt update -y && apt install build-essential \
-	gcc \
-	libdbus-1-dev \
-	rpm \
-	binutils \
-	curl \
-	p7zip-full \
-	git -y
+# podman build . -t mullvadvpn-app-build
+#
+# To run the image and build the app you need to mount the app's source directory into the
+# container. You also probably want to mount in a directory for CARGO_HOME, so each container
+# does not need to start from scratch with cloning the crates.io index, download all
+# dependencies and building everything.
+#
+# podman run --rm \
+#     -v ~/.cargo:/root/.cargo:Z \
+#     -v .:/build:Z \
+#     mullvadvpn-app-build ./build.sh
 
+# Debian 10 is the oldest supported distro. It has the oldest glibc that we support
+FROM debian:10.13-slim@sha256:557ee531b81ce380d012d83b7bb56211572e5d6088d3e21a3caef7d7ed7f718b
+
+# === Define toolchain versions and paths ===
+
+ENV CARGO_HOME=/root/.cargo
+ENV CARGO_TARGET_DIR=/root/.cargo/target
+
+ENV GOLANG_VERSION 1.18.5
+ENV GOLANG_HASH 9e5de37f9c49942c601b191ac5fba404b868bfc21d446d6960acc12283d6e5f2
+
+# === Install/set up the image ===
+
+RUN apt-get update -y && apt-get install -y \
+    git \
+    curl \
+    gcc \
+    libdbus-1-dev \
+    rpm \
+    protobuf-compiler \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install latest stable Rust toolchain
+ENV PATH "/root/.cargo/bin:$PATH"
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
+    sh -s -- --default-toolchain stable --profile minimal -y
+
+ENV PATH /root/.volta/bin:$PATH
+# volta seemingly does not have a way to explicitly install the toolchain
+# versions from package.json, but `node --version` triggers an install
+COPY gui/package.json .
+RUN curl https://get.volta.sh | bash && node --version && rm package.json
 
 # Install golang
-ENV GOLANG_VERSION 1.18.5
 # Checksum from: https://go.dev/dl/
-ENV GOLANG_HASH 9e5de37f9c49942c601b191ac5fba404b868bfc21d446d6960acc12283d6e5f2
 RUN curl -Lo go.tgz https://go.dev/dl/go${GOLANG_VERSION}.linux-amd64.tar.gz && \
-	echo $(sha256sum go.tgz) && \
-	echo "${GOLANG_HASH} go.tgz" | sha256sum -c - && \
-	tar -C /usr/local -xzf go.tgz && \
-	rm go.tgz && \
-	rm -rf /var/lib/apt/lists/*
+    echo "${GOLANG_HASH} go.tgz" | sha256sum -c - && \
+    tar -C /usr/local -xzf go.tgz && \
+    rm go.tgz
+ENV PATH /usr/local/go/bin:$PATH
 
-ENV GOPATH /go
-ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
-RUN go version
-RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
-
-RUN mkdir /mvd
-CMD tail -f /dev/null
+WORKDIR /build
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,8 +21,8 @@ FROM debian:10.13-slim@sha256:557ee531b81ce380d012d83b7bb56211572e5d6088d3e21a3c
 
 ENV CARGO_TARGET_DIR=/root/.cargo/target
 
-ENV GOLANG_VERSION 1.18.5
-ENV GOLANG_HASH 9e5de37f9c49942c601b191ac5fba404b868bfc21d446d6960acc12283d6e5f2
+ENV GOLANG_VERSION=1.18.5 \
+    GOLANG_HASH=9e5de37f9c49942c601b191ac5fba404b868bfc21d446d6960acc12283d6e5f2
 
 # === Install/set up the image ===
 
@@ -41,7 +41,7 @@ RUN dpkg --add-architecture arm64 && apt-get update -y && apt-get install -y \
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
     sh -s -- --default-toolchain stable --profile minimal --target aarch64-unknown-linux-gnu -y
 
-ENV PATH "/root/.cargo/bin:$PATH"
+ENV PATH=/root/.cargo/bin:$PATH
 
 RUN echo '[target.aarch64-unknown-linux-gnu]\n\
 linker = "aarch64-linux-gnu-gcc"\n\
@@ -52,7 +52,7 @@ rustc-link-lib = ["dbus-1"]' > /root/.cargo/config.toml
 
 # === Volta for npm + node ===
 
-ENV PATH /root/.volta/bin:$PATH
+ENV PATH=/root/.volta/bin:$PATH
 # volta seemingly does not have a way to explicitly install the toolchain
 # versions from package.json, but `node --version` triggers an install
 COPY gui/package.json .
@@ -66,7 +66,7 @@ RUN curl -Lo go.tgz https://go.dev/dl/go${GOLANG_VERSION}.linux-amd64.tar.gz && 
     echo "${GOLANG_HASH} go.tgz" | sha256sum -c - && \
     tar -C /usr/local -xzf go.tgz && \
     rm go.tgz
-ENV PATH /usr/local/go/bin:$PATH
+ENV PATH=/usr/local/go/bin:$PATH
 
 
 WORKDIR /build
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,16 +7,18 @@
 # dependencies and building everything.
 #
 # podman run --rm \
-#     -v ~/.cargo:/root/.cargo:Z \
+#     -v /path/to/container_cache/target:/root/.cargo/target:Z \
+#     -v /path/to/container_cache/registry:/root/.cargo/registry:Z \
 #     -v .:/build:Z \
 #     mullvadvpn-app-build ./build.sh
+#
+# And add -e TARGETS="aarch64-unknown-linux-gnu" to build for ARM64
 
 # Debian 10 is the oldest supported distro. It has the oldest glibc that we support
 FROM debian:10.13-slim@sha256:557ee531b81ce380d012d83b7bb56211572e5d6088d3e21a3caef7d7ed7f718b
 
 # === Define toolchain versions and paths ===
 
-ENV CARGO_HOME=/root/.cargo
 ENV CARGO_TARGET_DIR=/root/.cargo/target
 
 ENV GOLANG_VERSION 1.18.5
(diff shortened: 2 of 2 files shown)


albin-mullvad (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[`Dockerfile` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NDgSTEm1XkxhDxbzgAc:-NDgSTEm1XkxhDxbzgAd:b-dbzwfu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3b2920d1d864cb72b3da87ff384cfb917a952de/Dockerfile#L10)):*
> ```Dockerfile
> #
> # podman run --rm \
> #     -v ~/.cargo:/root/.cargo:Z -e=CARGO_TARGET_DIR=/root/.cargo/target \
> ```

I suggest the following ENV variables as defaults in the dockerfile:

```
ENV CARGO_HOME=/root/.cargo
ENV CARGO_TARGET_DIR=/root/.cargo/target
```

Might perhaps be nice to do something similar for directories used by go and volta/npm as well. 



_Code quote:_
```Dockerfile
-e=CARGO_TARGET_DIR=/root/.cargo/target
```
___
*[`Dockerfile` line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NDgPqs12b0ktR3ZrMuk:-NDgPqs12b0ktR3ZrMul:b-8vewmh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3b2920d1d864cb72b3da87ff384cfb917a952de/Dockerfile#L11)):*
> ```Dockerfile
> # podman run --rm \
> #     -v ~/.cargo:/root/.cargo:Z -e=CARGO_TARGET_DIR=/root/.cargo/target \
> #     -v .:/build:Z -w /build \
> ```

I suggest putting `WORKDIR /build` as a default at the end of the dockerfile



_Code quote:_
```Dockerfile
 -w /build
```
___
*[`Dockerfile` line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NDgRxV21kqZQkyp8cqj:-NDgRxV21kqZQkyp8cqk:byzcsy2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3b2920d1d864cb72b3da87ff384cfb917a952de/Dockerfile#L37)):*
> ```Dockerfile
> ENV GOLANG_VERSION 1.18.5
> # Checksum from: https://go.dev/dl/
> ENV GOLANG_HASH 9e5de37f9c49942c601b191ac5fba404b868bfc21d446d6960acc12283d6e5f2
> ```

Would perhaps be nice to move the ENV variables to the top in order to provide a quick overview and make it easy to update the values.



_Code quote:_
```Dockerfile
ENV GOLANG_VERSION 1.18.5
# Checksum from: https://go.dev/dl/
ENV GOLANG_HASH 9e5de37f9c49942c601b191ac5fba404b868bfc21d446d6960acc12283d6e5f2
```

albin-mullvad:
I also suggest that we move the Dockerfile into a `docker` directory

faern:
> I also suggest that we move the Dockerfile into a docker directory

I'd argue we should be a bit more generic. Both because we'll likely not use docker specifically, rather podman, and because that folder would be of very narrow use. A more generic `build/` or similar folder could be used to absorb more files in this area.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad)

___
*[`Dockerfile` line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NDgPqs12b0ktR3ZrMuk:-NE0YvbiBgC1igZgYGH-:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3b2920d1d864cb72b3da87ff384cfb917a952de/Dockerfile#L11)):*

I suggest putting `WORKDIR /build` as a default at the end of the dockerfile
</blockquote></details>

Done.
___
*[`Dockerfile` line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NDgRxV21kqZQkyp8cqj:-NE0Z5wC6JfrLkdtzvpQ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3b2920d1d864cb72b3da87ff384cfb917a952de/Dockerfile#L37)):*

Would perhaps be nice to move the ENV variables to the top in order to provide a quick overview and make it easy to update the values.
</blockquote></details>

Done.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad)

___
*[`Dockerfile` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NDgSTEm1XkxhDxbzgAc:-NE0bx0L1jXU3Oakg6Rl:bse4keo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3b2920d1d864cb72b3da87ff384cfb917a952de/Dockerfile#L10)):*

I suggest the following ENV variables as defaults in the dockerfile:

```
ENV CARGO_HOME=/root/.cargo
ENV CARGO_TARGET_DIR=/root/.cargo/target
```

Might perhaps be nice to do something similar for directories used by go and volta/npm as well. 

</blockquote></details>

We automatically get `node_modules` mounted into the container by it being placed inside the repository. So each container does not need to download the full GUI dependency tree already. Not sure what else we want/can cache in that way?

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: all files reviewed, 3 unresolved discussions (waiting on @albin-mullvad)

albin-mullvad (commented):
Sure, perhaps `container`? `build/` is already used as an output for building **go**

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-:-NE18KQFBFg6BpR-c5QN:b-usovxe)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`Dockerfile` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NDgSTEm1XkxhDxbzgAc:-NE13j-x7s2nA3TEJ8co:b-4ihf3l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3b2920d1d864cb72b3da87ff384cfb917a952de/Dockerfile#L10)):*

We automatically get `node_modules` mounted into the container by it being placed inside the repository. So each container does not need to download the full GUI dependency tree already. Not sure what else we want/can cache in that way?
</blockquote></details>

Alright. I generally believe it's more convenient when dependencies/artifacts *can* be cached outside the repo, especially when building in a CI environment, since it makes cache management easier. Does the same apply to **go**?

albin-mullvad (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: all files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[`Dockerfile` line 33 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NE4dLd21xjPDc3VeuJF:-NE4dLd21xjPDc3VeuJG:biiazti) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ef79c13e733dd0132eab3774838c5d965479f2d9/Dockerfile#L33)):*
> ```Dockerfile
>     libdbus-1-dev \
>     rpm \
>     protobuf-compiler \
> ```

Were you able to build the daemon using this container? Otherwise we might need to add `libprotobuf-dev` as well.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: all files reviewed, 2 unresolved discussions (waiting on @albin-mullvad and @faern)

___
*[`Dockerfile` line 33 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NE4dLd21xjPDc3VeuJF:-NE5hlq74cgFsM44fZaK:b2n4e1z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ef79c13e733dd0132eab3774838c5d965479f2d9/Dockerfile#L33)):*

Were you able to build the daemon using this container? Otherwise we might need to add `libprotobuf-dev` as well.
</blockquote></details>

It didn't work on Debian 9 but on Debian 10, it believe.

faern (commented):
Maybe we can convince go to use another dir? A shame that such a small part of the entire app uses up such an awesome directory name.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-:-NE6LEni3tH4hoo3o6pF:b3okr5v)* status: all files reviewed, 2 unresolved discussions (waiting on @albin-mullvad)

___
*[`Dockerfile` line 33 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NE4dLd21xjPDc3VeuJF:-NE6L0j102p_bRkQh_n5:b-ec17ck) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ef79c13e733dd0132eab3774838c5d965479f2d9/Dockerfile#L33)):*

It didn't work on Debian 9 but on Debian 10, it believe.
</blockquote></details>

I'm able to build the entire app in this container.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: all files reviewed, 2 unresolved discussions (waiting on @albin-mullvad and @faern)

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: all files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @faern)

___
*[`Dockerfile` line 42 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NE_lieY5H6Lgmp48U8n:-NE_lieY5H6Lgmp48U8o:b-mssrci) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4fa750faee9bf75066e0a16cfaadd3336a23f93e/Dockerfile#L42)):*
> ```Dockerfile
> RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
>     sh -s -- --default-toolchain stable --profile minimal --target aarch64-unknown-linux-gnu -y
> ENV PATH "/root/.cargo/bin:$PATH"
> ```

Nit: If setting the PATH takes place after installing rustup, should the order of operations not be the same for Volta too?

pinkisemils (approved):
I'm sure `build` is where we intend to store intermediate artifacts used for building the daemon. But we currently only have a single intermediate artifact - `wireguard-go`. 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-:-NEa8QBP6CoPyEKDJ2Mr:b-u1am0q)* status: all files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @faern)

albin-mullvad (approved):
I suggest moving the volta path as suggested by emils, otherwise it :lgtm:

Potential cache improvement etc can be handled in separate PRs imo.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-:-NEeF-zv0SrGfr3_NS3B:b2a9r5y)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`Dockerfile` line 42 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NE_lieY5H6Lgmp48U8n:-NEfPZOH0re8G96LIgN-:b-xcu4cm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4fa750faee9bf75066e0a16cfaadd3336a23f93e/Dockerfile#L42)):*

Nit: If setting the PATH takes place after installing rustup, should the order of operations not be the same for Volta too?
</blockquote></details>

I knew someone would comment on this :joy: I think it's nicer to set it up after (any reason? Dunno, seems better to configure software after it is installed). But with Volta, the install was affected by it already being in the path. If I added it first, then the installer would stop touching `~/.bashrc` on install. Maybe it does not matter much, but there was a difference.

albin-mullvad (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: all files reviewed, 2 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[`Dockerfile` line 10 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NEfoxsADdaJcmgFmTpn:-NEfoxsADdaJcmgFmTpo:bo6pw5j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4fa750faee9bf75066e0a16cfaadd3336a23f93e/Dockerfile#L10)):*
> ```Dockerfile
> #
> # podman run --rm \
> #     -v ~/.cargo:/root/.cargo:Z \
> ```

Mounting the `CARGO_HOME` directory this way will cause the container to use mounted binaries rather than container binaries since `/root/.cargo/bin` already is part of the `PATH`.

Would it be sufficient to mount `~/.cargo/registry` and `~/.cargo/target`?



_Code quote:_
```Dockerfile
~/.cargo:/root/.cargo
```

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: all files reviewed, 2 unresolved discussions (waiting on @albin-mullvad and @pinkisemils)

___
*[`Dockerfile` line 10 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NEfoxsADdaJcmgFmTpn:-NEjSXYI8LtwOY77iR3r:b3i9894) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4fa750faee9bf75066e0a16cfaadd3336a23f93e/Dockerfile#L10)):*

Mounting the `CARGO_HOME` directory this way will cause the container to use mounted binaries rather than container binaries since `/root/.cargo/bin` already is part of the `PATH`.

Would it be sufficient to mount `~/.cargo/registry` and `~/.cargo/target`?
</blockquote></details>

Yes. I have also come to realize it's not a good idea to mount in all of `~/.cargo`. Working on a solution. We also don't need to keep `target` inside it, the output target could be anywhere. The registry is harder since you can't specify where cargo should store it.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: all files reviewed, 2 unresolved discussions (waiting on @albin-mullvad and @pinkisemils)

___
*[`Dockerfile` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NDgSTEm1XkxhDxbzgAc:-NEjTYkD6kRHZ5lWeB-f:bhzlv0b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e3b2920d1d864cb72b3da87ff384cfb917a952de/Dockerfile#L10)):*

Alright. I generally believe it's more convenient when dependencies/artifacts *can* be cached outside the repo, especially when building in a CI environment, since it makes cache management easier. Does the same apply to **go**?
</blockquote></details>

I agree. But supposedly, this can't be done for `node_modules` without switching from `npm` to `pnpm`. Which we might consider doing, but I think it's outside the scope of this PR.

albin-mullvad (commented):
We probably want to consider how we best can handle artifact/cache permissions. With the current setup, a non-root host user will have limited or no access. For a build container like this it might be sufficient to just use some `chown`/`chmod` as part of the build scripts, but there are ofc a range of ways to handle it. What's your thoughts?

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-:-NEjaDecE2BvAYruyb0m:b-pocicz)* status: all files reviewed, 1 unresolved discussion

pinkisemils (approved):
If you mean access to the directories that are passed through and then written to by root from within the container - this is generally a non-issue with podman since the UIDs are mapped so that on the _host_ all the IO is done with your normal user UIDs. Or at least that's been my experience.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-:-NEk29IyCsM1M2_JAZLI:bp7uj6c)* status: all files reviewed, 1 unresolved discussion (waiting on @albin-mullvad)

___
*[`Dockerfile` line 42 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NE_lieY5H6Lgmp48U8n:-NEk2bya7RL5ucU7TZM_:b-6vpziz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4fa750faee9bf75066e0a16cfaadd3336a23f93e/Dockerfile#L42)):*

This sadly does not work right now. If I move the `PATH` configuration to after the current `RUN` command the `node --version` part fails. Because it can't find node. I think I prefer setting it up front compared to having to split the installation into multiple `RUN` statements. What do you think?
</blockquote></details>

I think it's a nit and since it's a problem to move it around, let's leave the path reassignment as is now. It would be nicer to maintain consistency here and move the cargo path assignment up too then?

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @albin-mullvad and @pinkisemils)

___
*[`Dockerfile` line 52 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NEkAGNO-W0Pi3XWr1OI:-NEkAGNO-W0Pi3XWr1OJ:b-grhc67) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f57d9af4bb3d2ffa5c7d6782d911d175871f3d05/Dockerfile#L52)):*
> ```Dockerfile
> [target.aarch64-unknown-linux-gnu.dbus]\n\
> rustc-link-search = ["/usr/aarch64-linux-gnu/lib"]\n\
> rustc-link-lib = ["dbus-1"]' > /root/.cargo/config.toml
> ```

The alternative to this is to store the condig in the repo and move it into the container with `COPY`. But I preferred having the dockerfile self contained as much as possible. What's your thought? We still depend on `gui/package.json` externally.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: all files reviewed, 2 unresolved discussions (waiting on @albin-mullvad)

albin-mullvad (approved):
> If you mean access to the directories that are passed through and then written to by root from within the container - this is generally a non-issue with podman since the UIDs are mapped so that on the host all the IO is done with your normal user UIDs. Or at least that's been my experience.

Alright, sounds good, I have quite limited experience with podman compared to docker. I guess that means that we should be very clear about primarily supporting/using podman for building.

> Reviewed 1 of 2 files at r1, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-:-NEkDeaoCQn3FM2gxHua:b-q7sxtv)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`Dockerfile` line 52 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986#-NEkAGNO-W0Pi3XWr1OI:-NEkCWLg6YKra_pMu5D4:b-7qcwak) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f57d9af4bb3d2ffa5c7d6782d911d175871f3d05/Dockerfile#L52)):*

The alternative to this is to store the condig in the repo and move it into the container with `COPY`. But I preferred having the dockerfile self contained as much as possible. What's your thought? We still depend on `gui/package.json` externally.
</blockquote></details>

FYI we copy the file into the Android Dockerfile as that `config.toml` is quite much bigger atm:
https://github.com/mullvad/mullvadvpn-app/blob/master/android/docker/cargo-config.toml

albin-mullvad (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3986)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3999, opened 2022-10-05
Reset known location when connecting to a custom relay

dlon:
Previously, when connecting to a custom relay, the daemon would include details for the last non-custom relay when returning the location in tunnel state transitions and when responding to `GetCurrentLocation`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3999)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ Line wrap the file at 100 chars.                                              Th
 - Experimental: Upgrade the support for quantum-resistant WireGuard tunnels to a newer protocol.
 
 ### Fixed
+- Reset known location when connecting to a custom relay.
+
 #### Linux
 - Fix app crashing immediately when using some icon themes.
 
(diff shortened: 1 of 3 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3999)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[`mullvad-cli/src/format.rs` line 169 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3999#-NDbThqa-uxqzjD_3UdF:-NDbThqa-uxqzjD_3UdG:bg326tm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0889a579f5c636d4b6aa5a3be2df1d880d495813/mullvad-cli/src/format.rs#L169)):*
> ```XML
> }
> 
> fn format_endpoint(hostname: &str, protocol_enum: i32, addr: Option<&str>) -> String {
> ```

I think `hostname` should be optional. `addr` could be non-optional if we add a `verbose: bool` parameter instead.

PR #4023, opened 2022-10-11
Fix apt lists clean up in Dockerfile

PeterDaveHello:
Origin apt list clean up won't really work as they're not in the same Docker image layer.

With proper modification, we could get a tidier Docker image:

```
REPOSITORY  TAG         IMAGE ID       CREATED         SIZE
after       latest      09dd0aa54dbe   2 minutes ago   891MB
before      latest      685bb4ca12d5   5 minutes ago   908MB
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4023)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,8 @@ RUN apt update -y && apt install build-essential \
 	binutils \
 	curl \
 	p7zip-full \
-	git -y
+	git -y && \
+	rm -rf /var/lib/apt/lists/*
 
 
 # Install golang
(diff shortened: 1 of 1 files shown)


faern:
Thank you for this contributon. However, this Dockerfile is "dead". We have not used it at all for a very long time. We are actually gearing up to fix that and introduce new good containers for building the app. What you do in this PR is already part of my other PR: #3986.

PeterDaveHello:
Oh, sorry that I didn't notice.

PR #4029, opened 2022-10-11
Select a relay even if `include_in_country == false` for all matching relays

dlon:
If none of the relays are "included in the country", given all other constraints, then the relay selector will now select one of them anyway. Previously, the relay selector would find no matching relay in this case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4029)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ Line wrap the file at 100 chars.                                              Th
 
 ## [Unreleased]
 ### Fixed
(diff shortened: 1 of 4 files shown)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4029)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4140, opened 2022-11-11
trusted DNS servers option

avieth:
# Motivating use case

**This patch is intended to spark discussion about the feature.**

A user has a VPN connection for work which gives its own DNS server that is trusted by the user, and the local resolver is set up to only route requests for a particular domain to this server. They can now use `mullvad dns trusted set <address of resolver>` to make it work.

# What it does

Adds an option for trusted DNS servers. These will not be blocked on port 53 in the firewall.

# How it does it

Mostly copied and modified from the existing DNS settings.

# Remaining work

I don't believe it works properly when the trusted DNS is set while the tunnel is not in connected state. Will look into this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4140)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ Line wrap the file at 100 chars.                                              Th
 
 ### Changed
 - Update Electron from 19.0.13 to 21.1.1.
+- Add a trusted DNS option
 
 ### Security
 #### Windows
--- a/mullvad-cli/src/cmds/dns.rs
+++ b/mullvad-cli/src/cmds/dns.rs
@@ -13,7 +13,7 @@ impl Command for Dns {
 
     fn clap_subcommand(&self) -> clap::App<'static> {
         clap::App::new(self.name())
-            .about("Configure DNS servers to use when connected")
+            .about("Configure DNS servers and policy")
             .setting(clap::AppSettings::SubcommandRequiredElseHelp)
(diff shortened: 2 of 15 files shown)


dlon:
You might be able to achieve this using custom DNS already. The app ignores private IPs (except some special cases like `10.64.0.1`) when configuring the tunnel interface. So if your DNS server is on a private IP, you could probably just add that IP as well `10.64.0.1` to the custom DNS server list.

avieth:
> You might be able to achieve this using custom DNS already. The app ignores private IPs (except some special cases like `10.64.0.1`) when configuring the tunnel interface. So if your DNS server is on a private IP, you could probably just add that IP as well `10.64.0.1` to the custom DNS server list.

If I set mullvad's DNS server and also the work DNS server in `mullvad dns set custom` then the app will configure the system resolver with both servers for all domains, but that's not what I want.

With this patch, I can leave mullvad configured to trust the work DNS server, and then I can bring the work VPN up and down freely (openvpn, will configure resolved appropriately). I've been using this patch for a couple weeks now and it's been great.

faern:
I'm closing this due to being very stale and not really on our roadmap. I'd recommend to have future discussions about this in an issue if needed. Prefer posting in existing issues if there is one on the topic.

Issue #4133, opened 2022-11-08
Post-quantum secure multihop


jorik392:
### Issue description

Post-quantum secure support has expanded but it is not possible to use it in multihop yet. Please add this feature.

faern:
Yes, this is in the backlog. But it's just an experimental feature so far. We will fix it eventually, before we classify it as stable.

jackloomen:
Would it be possible to get a standalone binary for the key exchange? The WG PSK can then be manually manged or via scripts.

faern:
We currently don't provide one. But the protocol is public, and you can pretty easily build one yourself from https://github.com/mullvad/mullvadvpn-app/blob/master/talpid-tunnel-config-client/examples/psk-exchange.rs

jackloomen:
> We currently don't provide one. But the protocol is public, and you can pretty easily build one yourself from https://github.com/mullvad/mullvadvpn-app/blob/master/talpid-tunnel-config-client/examples/psk-exchange.rs

That was helpful. But there appears to not be a function for printing PSK's properly.
```
private key: ......
psk: PresharedKey([102, 48, 95, .... ]
```

faern:
> edit: It also seems like redoing the PQC PSK procedure while connected to a PQ-Peer does not work for some reason.

Yes, changing the PSK while using the PSK does not make sense. That will create a race condition where the server will start using the new PSK as soon as it's able to compute it. This will break the current tunnel (as the client now has the wrong PSK) and the response to the gRPC request will not even make it back to the client most of the time (based on timing)

MrChocolatine:
@jorik392 , looks like it's coming in the next stable release `2023.3`!

https://github.com/mullvad/mullvadvpn-app/releases/tag/2023.3-beta1

faern:
Yes. That's true. It's fixed for the upcoming release and is part of the current beta. So I'm closing this as done.

PR #4161, opened 2022-11-22
debug/display for PresharedKey

jackloomen:
https://github.com/mullvad/mullvadvpn-app/issues/4133

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4161)

### DIFF
--- a/talpid-types/src/net/wireguard.rs
+++ b/talpid-types/src/net/wireguard.rs
@@ -287,12 +287,6 @@ impl From<Box<[u8; 32]>> for PresharedKey {
 
 impl fmt::Debug for PresharedKey {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", &self)
-    }
-}
-
-impl fmt::Display for PresharedKey {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}", &base64::encode(self.as_bytes()))
     }
 }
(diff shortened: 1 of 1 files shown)


talpid-types/src/net/wireguard.rs
```diff
@@ -285,6 +285,18 @@ impl From<Box<[u8; 32]>> for PresharedKey {
     }
 }
 
+impl fmt::Debug for PresharedKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", &self)
+    }
+}
+
+impl fmt::Display for PresharedKey {
```
faern, line 294: I'm not sure we want this type to implement `Display`. The PSK is a secret and we don't want to accidentally print/log/store/show it. It implements `ZeroizeOnDrop` because it should not be spread into more memory/storage than strictly needed.

The `Debug` impl is fine. I think you can modify this PR to have the `Debug` impl print the base64 encoded version and then it's mergable.

faern (changes requested):

talpid-types/src/net/wireguard.rs
jackloomen: But Display is implemented for PrivateKey. Or is the PSK treated with even more importance?

jackloomen: Given that a PSK is per-peer and not centrally stored it does make sense. Changes made.

faern: Oh. With my above argument `PrivateKey` should not implement `Display`. I'm not sure why it does or if it has to for some other reason. But unless it's strictly needed I would prefer avoiding it.

faern (approved):

faern:
The CI failure is because of a compiler bug it seems. We have that in our other PRs as well. We'll get to merging this after we have resolved that I guess.

faern:
Sorry for the delay. Merged!

PR #3971, opened 2022-09-27
Refactor AuthFailed error

dlon:
This PR refactors the management interface to return structured errors instead of strings for `AuthFailed`. Failing to connect to a WireGuard relay when the account is out of time will now enter the error state due to `AuthFailed::EXPIRED_ACCOUNT`.

This PR does not touch `ErrorStateCause` in `talpid-types`, as accounts do not exist in that layer. Currently, the correct fix entails duplicating all tunnel states in `mullvad-types`. This would be somewhat pointless to do before the TSM has been refactored.

Related: #3965

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/3971)

PR #4162, opened 2022-11-23
Add transition on auth failed caused by account expiry

albin-mullvad:
This PR aims to adds an automatic transition to the out-of-time view if there's a tunnel state error caused by the account being expired. It also removes the expiry check scheduling in the connect view.

Depends on: #3971

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4162)

### DIFF
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/ConnectFragment.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/ConnectFragment.kt
@@ -44,6 +44,7 @@ import net.mullvad.mullvadvpn.util.JobTracker
 import net.mullvad.mullvadvpn.util.appVersionCallbackFlow
 import net.mullvad.mullvadvpn.util.callbackFlowFromNotifier
 import net.mullvad.mullvadvpn.viewmodel.ConnectViewModel
+import net.mullvad.talpid.tunnel.ErrorStateCause
 import org.joda.time.DateTime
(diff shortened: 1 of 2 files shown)


dlon (approved):
> Reviewed 1 of 2 files at r2, 5 of 5 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4162)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`android/app/src/main/kotlin/net/mullvad/talpid/tunnel/ErrorStateCause.kt` line 13 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4162#-NI7nIMG-7JYFphJ_diy:-NI7nIMG-7JYFphJ_diz:b-euuy9h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3d7526521fd852ad8c7019bd15217f4edf9e81d/android/app/src/main/kotlin/net/mullvad/talpid/tunnel/ErrorStateCause.kt#L13)):*
> ```Kotlin
>     class AuthFailed(private val reason: String?) : ErrorStateCause() {
>         fun isCausedByExpiredAccount(): Boolean {
>             return reason == AUTH_FAILED_REASON_EXPIRED_ACCOUNT
> ```

We should probably have parsed the string on the Rust side instead of passing it to the Android app. But this is alright for now.

PR #4167, opened 2022-11-23
Fix race condition when updating IPC target when recreating window

raksooo:
This PR fixes a race condition which caused the `IpcMainEventChannel` to not have a `webContents` after toggling "Unpinned window". The issue was that the `webContents` `destroyed` event was received after the new `webContents` had been set up and therefore removing it from `IpcMainEventChannel`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4167)


hankolsen (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4167)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4174, opened 2022-11-28
Fix animation glitch when pressing escape multiple times

raksooo:
This PR fixes the UI glitch that happened when pressing escape multiple time quickly. This was accomplished by:
* Preventing `onTransitionEnd` in `TransitionContainer` from firing twice for transitions where both current and next view has a CSS transition.
* Moving the transition information into the location state.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4174)

### DIFF
--- a/gui/src/renderer/app.tsx
+++ b/gui/src/renderer/app.tsx
@@ -182,9 +182,7 @@ export default class AppRenderer {
       this.reduxActions.userInterface.setMacOsScrollbarVisibility(visibility);
     });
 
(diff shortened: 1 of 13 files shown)


hankolsen (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4174)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4176, opened 2022-11-29
Improve network path status reporting

pronebird:
1. Wait for `NWPathMonitor` to report the network path update first before starting monitoring the tunnel. The reason for that is because `NWPathMonitor.currentPath` may not be set immediately and very often the tunnel monitor starts in the state where it thinks that connection is offline which is often times false. The new `pendingStart` state was added to account for that.
2. Wait for tunnel interface `utun` to appear in the list of network interfaces, before deducing if network connection is viable.
3. Improve check for presence of physical network interface by checking the interface type (wifi, cellular, or ethernet)

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4176)

### DIFF
--- a/ios/MullvadTypes/PacketTunnelStatus.swift
+++ b/ios/MullvadTypes/PacketTunnelStatus.swift
@@ -42,6 +42,7 @@ public struct PacketTunnelStatus: Codable, Equatable {
 
     /// Last performed device check.
     public var deviceCheck: DeviceCheck?
+
     /// Current relay.
     public var tunnelRelay: PacketTunnelRelay?
 
--- a/ios/PacketTunnel/TunnelMonitor/TunnelMonitor.swift
+++ b/ios/PacketTunnel/TunnelMonitor/TunnelMonitor.swift
@@ -437,30 +437,44 @@ final class TunnelMonitor: PingerDelegate {
     }
 
     private func handleNetworkPathUpdate(_ networkPath: Network.NWPath) {
-        let isReachable = isNetworkPathReachable(networkPath)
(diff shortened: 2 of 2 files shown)


ios/MullvadTypes/PacketTunnelStatus.swift
```diff
@@ -42,6 +42,7 @@ public struct PacketTunnelStatus: Codable, Equatable {
 
     /// Last performed device check.
     public var deviceCheck: DeviceCheck?
+
```
sajacl, line 45: FYI: Is this change made by swift format?

sajacl (approved):

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4176)* status: 0 of 2 files reviewed, all discussions resolved

___
*[`ios/MullvadTypes/PacketTunnelStatus.swift` line 45 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4176#-NI2DqrTIWUCWsnJcOCC-r1-45:-NI7RDW_7-VTe3o7aBx-:bj8a381) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/332f546fbd75848729d71158e6930405b9a4ad93/ios/MullvadTypes/PacketTunnelStatus.swift#L45)):*
<details><summary><i>Previously, sajacl (Sajad Vishkai) wrote…</i></summary><blockquote>

FYI: Is this change made by swift format?
</blockquote></details>

No. I myself added a newline between members.

PR #4177, opened 2022-11-29
Fix some scaling issues on Windows

raksooo:
This PR hopefully improves the scaling issues on Windows when or after a DPI/scale change. The specific case I was able to reproduce was that the window became taller for each time it was opened when it was unpinned and after changing the Windows scale factor. Hopefully this fixes more issues.

My guess is that `setPosition` sets the bounds but somehow has incorrect values for height and width after a scale/DPI change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4177)

### DIFF
--- a/gui/src/main/window-controller.ts
+++ b/gui/src/main/window-controller.ts
@@ -196,8 +196,9 @@ export default class WindowController {
 
   public updatePosition() {
     if (this.window) {
-      const { x, y } = this.windowPositioning.getPosition(this.window);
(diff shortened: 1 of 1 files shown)


hankolsen (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4177)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4187, opened 2022-11-30
Disable multi-window support and shortcuts on iPad

pronebird:
Alternative PR to https://github.com/mullvad/mullvadvpn-app/pull/4185. This one simply enables intents on iPhone only.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4187)


pronebird:
This POC works but the rumor is, split screen is coming to iPhones in the next iOS update. So to avoid adding a feature only to remove it sometime later, I suppose it would be better to go with #4185

PR #4208, opened 2022-12-12
Add netsh DNS manager

dlon:
Given the issues found on some Windows 10 builds, I've opted to keep it simple and just do the obvious thing. This will cause some complaints: (1) timeouts (i.e., the hosts file cannot be used for adblocking), and (2) the app depends on `dnscache`.

This PR adds a new Rust wrapper around netsh instead of bringing back the old WinDns one.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4208)


talpid-core/src/dns/windows.rs
```diff
@@ -0,0 +1,211 @@
...
+        if ipv6_index == 0 {
+            netsh_input.push_str(&create_netsh_flush_command(interface_index, IpVersion::V6));
+        }
+
+        run_netsh_with_timeout(netsh_input, NETSH_TIMEOUT)?;
+
+        Ok(())
+    }
+
+    fn reset(&mut self) -> Result<(), Error> {
```
dlon, line 114: Calling this is actively bad when we're just doing it prior to destroying the tunnel interface. Should we add another method that defaults to calling `reset` but that is no-op for most DNS monitor implementations? I'd want to use that whenever destroying the tunnel interface is sufficient to achieve the same result.

Jontified (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4208)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`talpid-core/src/dns/windows.rs` line 79 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4208#-NJK-Vgm6yqsreCTX27L:-NJK-Vgm6yqsreCTX27M:b-f0ipmq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a70475c73e78b55a590b53ff9f09f7b75c456775/talpid-core/src/dns/windows.rs#L79)):*
> ```XML
>         self.current_index = Some(interface_index);
> 
>         let mut ipv4_index = 0;
> ```

Shouldn't these be `bool` rather than integers? We only care if they are non-zero and we never decrement them.
___
*[`talpid-core/src/dns/windows.rs` line 114 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4208#-NJ6b3DACbLHwmI8RnZp-r1-114:-NJK3Zqx7eWkRJfdbR1I:bp8f9jp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a70475c73e78b55a590b53ff9f09f7b75c456775/talpid-core/src/dns/windows.rs#L114)):*

Calling this is actively bad when we're just doing it prior to destroying the tunnel interface. Should we add another method that defaults to calling `reset` but that is no-op for most DNS monitor implementations? I'd want to use that whenever destroying the tunnel interface is sufficient.
</blockquote></details>

This sounds like a pretty cheap solution to a somewhat real problem. So that sounds good to me.

Jontified (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4208#-:-NJKE26Z6YLne204T80I:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

dlon:
We should perhaps keep the old code and make this another module.

PR #4218, opened 2022-12-14
Don't block the default route listener while calling `REGISTER_IP_ADDRESSES`

dlon:
As above. The ioctl call can take some time to complete (mostly due to the transaction lock being blocked). If that occurs, the offline monitor will be affected as well. This PR simply queues up the calls on a different thread.

Another change that this PR makes is to simply log errors instead of forcing the tunnel state machine into the error state. There's no need to do so: failing to update the IPs cannot cause any leaks into the tunnel since the tunnel IP is static. From a security standpoint, moving to the error state is unnecessary.

Finally, also log any timeout that occurs on the split tunnel "request handling" thread.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4218)

### DIFF
--- a/talpid-core/src/split_tunnel/windows/event.rs
+++ b/talpid-core/src/split_tunnel/windows/event.rs
@@ -1,0 +1,213 @@
+//! Handle events dispatched from win-split-tunnel.
+//!
+//! It follows a typical inverted-call model, in which we request the next event, and block until
+//! such an event has been received, or until a quit event is signaled.
+
+use super::driver;
+use std::{
+    collections::HashMap,
+    io,
+    path::Path,
+    sync::{Arc, RwLock},
+    time::Duration,
+};
+use talpid_types::{split_tunnel::ExcludedProcess, ErrorExt};
+use talpid_windows::{io::Overlapped, sync::Event};
+use windows_sys::Win32::Foundation::ERROR_OPERATION_ABORTED;
+
+enum EventResult {
+    /// Result containing the next event.
+    Event(driver::EventId, driver::EventBody),
+    /// Quit event was signaled.
+    Quit,
+}
+
+const DRIVER_EVENT_BUFFER_SIZE: usize = 2048;
+
+/// Spawns an event loop thread that processes events from the driver service.
+pub fn spawn_listener(
+    handle: Arc<driver::DeviceHandle>,
+    excluded_processes: Arc<RwLock<HashMap<usize, ExcludedProcess>>>,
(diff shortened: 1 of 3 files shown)


Jontified (changes requested):
Like we talked about, it might be better to have the daemon block if there is a problem with the split tunneling rather than it silently failing. The ideal would be some way to notify users about non-security critical problems without blocking the tunnel but since that would require UI changes for now blocking seems preferable.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4218#-:-NJf8eOH8O05jt7-OFe3:bi818ff)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 891 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4218#-NJf9eN62pCJybRta78K:-NJf9eN62pCJybRta78L:bycm1z8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9bb5ebd0406b63cd649b474de7ba8a6632ddc45c/talpid-core/src/split_tunnel/windows/mod.rs#L891)):*
> ```XML
>                         )
>                     );
>                     maybe_send(TunnelCommand::Block(ErrorStateCause::SplitTunnelError));
> ```

Like we discussed keeping this is probably a good idea

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4218)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 891 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4218#-NJf9eN62pCJybRta78K:-NJu8rBR3NChXIoWIWaQ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9bb5ebd0406b63cd649b474de7ba8a6632ddc45c/talpid-core/src/split_tunnel/windows/mod.rs#L891)):*

Like we discussed keeping this is probably a good idea
</blockquote></details>

Done.

Jontified (dismissed):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4218#-:-NLM0DNJEa0KZXrDfIbr:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4219, opened 2022-12-14
Reduce the number of `REGISTER_IP_ADDRESSES` calls

dlon:
Follow-up to https://github.com/mullvad/mullvadvpn-app/pull/4218. This PR makes a couple of changes:
* Don't unregister tunnel IP in ST driver when reconnecting
* When entering the error state due to being offline, don't send an update.

Most of the time, this should lower the number of IOCTL calls from 2 to 0 when reconnecting, and from 2 to 1 when connecting if the tunnel IP has changed.

There is a small risk that some other network interface will take over the previous tunnel IP. If that happens, excluded connections will be incorrectly redirected to the default interface, though this is unlikely and is irrelevant from a security standpoint. As long as the tunnel IP is correct when there is a tunnel device, leaks are not possible.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219)

### DIFF
--- a/talpid-core/src/split_tunnel/windows/mod.rs
+++ b/talpid-core/src/split_tunnel/windows/mod.rs
@@ -108,7 +108,6 @@ pub struct SplitTunnel {
     quit_event: Arc<windows::Event>,
     excluded_processes: Arc<RwLock<HashMap<usize, ExcludedProcess>>>,
     _route_change_callback: Option<CallbackHandle>,
-    daemon_tx: Weak<mpsc::UnboundedSender<TunnelCommand>>,
     async_path_update_in_progress: Arc<AtomicBool>,
     route_manager: RouteManagerHandle,
 }
@@ -119,9 +118,19 @@ enum Request {
     Stop,
 }
 type RequestResponseTx = sync_mpsc::Sender<Result<(), Error>>;
-type RequestTx = sync_mpsc::Sender<(Request, RequestResponseTx)>;
+type RequestTx = sync_mpsc::Sender<(Request, Option<RequestResponseTx>)>;
 
 const REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+
+impl Request {
+    fn request_name(&self) -> &'static str {
+        match self {
+            Request::SetPaths(_) => "SetPaths",
+            Request::RegisterIps(_) => "RegisterIps",
+            Request::Stop => "Stop",
+        }
+    }
+}
 
 #[derive(Default, PartialEq, Clone)]
 struct InterfaceAddresses {
@@ -168,8 +177,12 @@ impl SplitTunnel {
     ) -> Result<Self, Error> {
         let excluded_processes = Arc::new(RwLock::new(HashMap::new()));
 
-        let (request_tx, handle) =
-            Self::spawn_request_thread(resource_dir, volume_update_rx, excluded_processes.clone())?;
+        let (request_tx, handle) = Self::spawn_request_thread(
+            resource_dir,
+            daemon_tx,
+            volume_update_rx,
+            excluded_processes.clone(),
+        )?;
 
         let (event_thread, quit_event) =
             Self::spawn_event_listener(handle, excluded_processes.clone())?;
@@ -180,7 +193,6 @@ impl SplitTunnel {
             event_thread: Some(event_thread),
             quit_event,
             _route_change_callback: None,
-            daemon_tx,
             async_path_update_in_progress: Arc::new(AtomicBool::new(false)),
             excluded_processes,
             route_manager,
@@ -387,6 +399,7 @@ impl SplitTunnel {
 
     fn spawn_request_thread(
         resource_dir: PathBuf,
+        daemon_tx: Weak<mpsc::UnboundedSender<TunnelCommand>>,
         volume_update_rx: mpsc::UnboundedReceiver<()>,
         excluded_processes: Arc<RwLock<HashMap<usize, ExcludedProcess>>>,
     ) -> Result<(RequestTx, Arc<driver::DeviceHandle>), Error> {
@@ -429,6 +442,7 @@ impl SplitTunnel {
             let mut previous_addresses = InterfaceAddresses::default();
 
             while let Ok((request, response_tx)) = rx.recv() {
+                let request_name = request.request_name();
                 let response = match request {
                     Request::SetPaths(paths) => {
                         let mut monitored_paths_guard = monitored_paths.lock().unwrap();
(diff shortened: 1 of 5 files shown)


Jontified (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon)

___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 115 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-NJfADwQBdR-qb7DpiG6:-NJfADwQBdR-qb7DpiG7:b-en7m4q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9bb5ebd0406b63cd649b474de7ba8a6632ddc45c/talpid-core/src/split_tunnel/windows/mod.rs#L115)):*
> ```XML
>     excluded_processes: Arc<RwLock<HashMap<usize, ExcludedProcess>>>,
>     _route_change_callback: Option<CallbackHandle>,
>     daemon_tx: Weak<mpsc::UnboundedSender<TunnelCommand>>,
> ```

Discussed in previous PR
___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 857 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-NJiieBQBEJkaswMd0mx:-NJiieBQBEJkaswMd0my:b7n4uje) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/100f9c6efd3050be8c9767c3633b2ae1e82b3669/talpid-core/src/split_tunnel/windows/mod.rs#L857)):*
> ```XML
>     }
> 
>     fn create_burst_guard(
> ```

Perhaps we'd like to create a generic implementation of a burst guard and use it for `taplid-routing/src/windows/default_route_monitor.rs` as well? I guess the abort handle is here to make cancelling the task a bit quicker? Just dropping the `tx` would do and would make it a bit simpler but the task would cancel only at the point were `rx.next()` returns `None`. Not saying it is better.
___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 875 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-NJjyFQO-as3Wz6wV4VG:-NJjyFQO-as3Wz6wV4VH:bpv8qtn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9bb5ebd0406b63cd649b474de7ba8a6632ddc45c/talpid-core/src/split_tunnel/windows/mod.rs#L875)):*
> ```XML
>                 Ok(None) => {
>                     log::warn!("Failed to obtain default route interface address");
>                     match address_family {
> ```

The only time that `get_up_address_for_interface` returns `Ok(None)` here is when the new default interface has lost internet connection? 

Previously we'd have cleared the ST drivers IP addresses but now we don't which means that if the old default interface still has connectivity we'll keep using that for the ST. The reason it is impossible for this to leak inside of the tunnel is that the old default interface can not have been the tunnel interface and the tunnel interface can not change to be the old default interface. The reason this is impossible to leak outside of the tunnel is that this is already leaking outside of the tunnel, that's the point.

Am I understanding correctly? If I am then that sounds reasonable in order to avoid having unnecessary calls to the ST driver.
___
*[`talpid-core/src/tunnel_state_machine/disconnected_state.rs` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-NJlS7b92FlC9VL3o04Z:-NJlS7b92FlC9VL3o04_:b-qoiqr3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/100f9c6efd3050be8c9767c3633b2ae1e82b3669/talpid-core/src/tunnel_state_machine/disconnected_state.rs#L55)):*
> ```XML
>     ) {
>         if shared_values.block_when_disconnected {
>             if let Err(error) = shared_values.split_tunnel.set_tunnel_addresses(None) {
> ```

General question here. Why do we reset the split tunnel when `block_when_disconnected` is set to true? I'm guessing blocking it would not be what we want since only vpn tunnel traffic should be blocked?
___
*[`talpid-types/src/tunnel.rs` line 120 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-NJlY2T30ZxodsLTVewJ:-NJlY2T30ZxodsLTVewK:b-k3gxbn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/100f9c6efd3050be8c9767c3633b2ae1e82b3669/talpid-types/src/tunnel.rs#L120)):*
> ```XML
>     #[cfg(target_os = "windows")]
>     pub fn prevents_split_tunneling(&self) -> bool {
>         matches!(self, Self::SplitTunnelError | Self::IsOffline)
> ```

If the error is `SplitTunnelError` we will not clear the split tunnel IPs, why? I thought this was because we want to be able to disconnect and reconnect to networks without having to spend time clearing and resetting the ST IPs. If there is an ST error isn't that a different scenario?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219)* status: all files reviewed, 5 unresolved discussions (waiting on @Jontified)

___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 115 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-NJfADwQBdR-qb7DpiG6:-NJuDL4c9-SZNlCHgbg2:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9bb5ebd0406b63cd649b474de7ba8a6632ddc45c/talpid-core/src/split_tunnel/windows/mod.rs#L115)):*

Discussed in previous PR
</blockquote></details>

Done.
___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 857 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-NJiieBQBEJkaswMd0mx:-NJuKJ9hCvbQICUZ8TrL:b-mo7p69) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/100f9c6efd3050be8c9767c3633b2ae1e82b3669/talpid-core/src/split_tunnel/windows/mod.rs#L857)):*

Perhaps we'd like to create a generic implementation of a burst guard and use it for `taplid-routing/src/windows/default_route_monitor.rs` as well? I guess the abort handle is here to make cancelling the task a bit quicker? Just dropping the `tx` would do and would make it a bit simpler but the task would cancel only at the point were `rx.next()` returns `None`. Not saying it is better.
</blockquote></details>

We should probably consider generalizing them, but the gain isn't massive.
___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 875 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-NJjyFQO-as3Wz6wV4VG:-NJuDs66Dd4UDDWO---8:bmwtcq4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9bb5ebd0406b63cd649b474de7ba8a6632ddc45c/talpid-core/src/split_tunnel/windows/mod.rs#L875)):*

The only time that `get_up_address_for_interface` returns `Ok(None)` here is when the new default interface has lost internet connection? 

Previously we'd have cleared the ST drivers IP addresses but now we don't which means that if the old default interface still has connectivity we'll keep using that for the ST. The reason it is impossible for this to leak inside of the tunnel is that the old default interface can not have been the tunnel interface and the tunnel interface can not change to be the old default interface. The reason this is impossible to leak outside of the tunnel is that this is already leaking outside of the tunnel, that's the point.

Am I understanding correctly? If I am then that sounds reasonable in order to avoid having unnecessary calls to the ST driver.
</blockquote></details>

That's correct.
___
*[`talpid-core/src/tunnel_state_machine/disconnected_state.rs` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-NJlS7b92FlC9VL3o04Z:-NJuCD70874XgvAcckbZ:b-e1tfmm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/100f9c6efd3050be8c9767c3633b2ae1e82b3669/talpid-core/src/tunnel_state_machine/disconnected_state.rs#L55)):*

General question here. Why do we reset the split tunnel when `block_when_disconnected` is set to true? I'm guessing blocking it would not be what we want since only vpn tunnel traffic should be blocked?
</blockquote></details>

What we're doing is really just setting the tunnel IP in win-split-tunnel to a reserved IP.

This prevents some other network device from stealing the IP of the last Mullvad interface, while keeping the device in the engaged state to allow some traffic (eg LAN) to leak outside the tunnel in excluded apps. We don't clear the tunnel IP because doing so would force win-split-tunnel out of the engaged state.

If we didn't use a reserved IP, and some interface took over the IP, its connections would be incorrectly redirected to the default interface. Not a vuln, since it would only affect excluded apps, but definitely unexpected.
___
*[`talpid-types/src/tunnel.rs` line 120 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-NJlY2T30ZxodsLTVewJ:-NJuIVhe-tJ7y4gWu43m:bv5h0l4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/100f9c6efd3050be8c9767c3633b2ae1e82b3669/talpid-types/src/tunnel.rs#L120)):*

If the error is `SplitTunnelError` we will not clear the split tunnel IPs, why? I thought this was because we want to be able to disconnect and reconnect to networks without having to spend time clearing and resetting the ST IPs. If there is an ST error isn't that a different scenario?
</blockquote></details>

Basically, it is likely to fail and stall, since changing the addrs was the reason for entering the error state in the first place. But perhaps it should try anyway.

Jontified (dismissed):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-:-NLM0eiY-T4ula0fjU0U:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 857 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4219#-NJiieBQBEJkaswMd0mx:-NLM0Rh-1MRWtrakfxQf:bdp2e6h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/100f9c6efd3050be8c9767c3633b2ae1e82b3669/talpid-core/src/split_tunnel/windows/mod.rs#L857)):*

We should probably consider generalizing them, but the gain isn't massive.
</blockquote></details>

Fair enough, OOS for this PR.

PR #4266, opened 2023-01-10
Fix window height being too small

raksooo:
This PR fixes an issue where the app isn't high enough. It seems to be improssible to set an exact height so the change here makes it default to be too big which prevents UI glitches at least.

I've also changed the `TunnelControl` `min-height` to accomodate for the IPv6 out-IP to not make the content above jump when the connection panel is expanded.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4266)


hankolsen (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4266)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4270, opened 2023-01-10
Run cargo udeps on Windows and macOS also

faern:
Run `cargo udeps` on Windows and macOS along with Linux. We had a bunch of dependencies there that were unused, so this seems justified.

I have not spent any time doing the same for Android. We probably should however. But I'll leave that for later as it has to be some cross compile thingy and I don't want to bother right now. This will at least make things significantly better than before.

I opted to "upgrade" the Linux job here to use our build container. To be consistent with the other Rust Linux build jobs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4270)

### DIFF
--- a/.github/workflows/rust-unused-dependencies.yml
+++ b/.github/workflows/rust-unused-dependencies.yml
@@ -8,47 +8,78 @@ on:
     # Check if requested manually from the Actions tab
     workflow_dispatch:
 jobs:
-    cargo-udeps:
-        # TODO: This does not capture and alert about platform specific dependencies on non-linux
-        # Since `cargo-udeps` seem to go pretty far with the compiling, i guess we need
-        # to run separate runners with different OSes to capture this aspect.
+    prepare-linux:
+      runs-on: ubuntu-latest
+      steps:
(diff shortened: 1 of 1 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4270)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`.github/workflows/rust-unused-dependencies.yml` line 59 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4270#-NLRFtOiAU4-Izhmwenm:-NLRFtOiAU4-Izhmwenn:b-4kxlhi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/986020982d6342ed41509b968e31918f260ba7d7/.github/workflows/rust-unused-dependencies.yml#L59)):*
> ```YAML
> 
> 
>     cargo-udeps-macos:
> ```

Maybe we could use a matrix here since the steps are identical for macOS and Windows.

Issue #4277, opened 2023-01-14
Installing ARM64 rpm reports wrong architecture - Fedora


deja-geek:
## Issue report

Operating system: Fedora 37 aarch64

App version: 2022.5.0

### Issue description

When attempting to install the ARM64 rpm on Fedora 37, rpm reports the rpm file is the wrong architecture

```
rpm -ivh MullvadVPN-2022.5_aarch64.rpm
Verifying...                          ################################# [100%]
Preparing...                          ################################# [100%]
        package mullvad-vpn-2022.5.0-1.arm64 is intended for a different architecture
```

```
rpm -qip MullvadVPN-2022.5_aarch64.rpm
Name        : mullvad-vpn
Version     : 2022.5.0
Release     : 1
Architecture: arm64
Install Date: (not installed)
Group       : default
Size        : 277237045
License     : GPL-3.0
Signature   : (none)
Source RPM  : mullvad-vpn-2022.5.0-1.src.rpm
Build Date  : Fri 14 Oct 2022 05:56:51 AM CDT
Build Host  : app-build-linux
Relocations : /
Packager    : Mullvad VPN <support@mullvad.net>
Vendor      : Mullvad VPN <support@mullvad.net>
URL         : https://github.com/mullvad/mullvadvpn-app#readme
Summary     : Mullvad VPN client
Description :
Mullvad VPN client
```

pinkisemils:
Thank you very much for reporting the issue. Rather embarrassingly, we hadn't tested for this, but the fix has turned out to be rather simple, and should be a part of the next release.

PR #4284, opened 2023-01-17
Fix aarch64 RPMs

pinkisemils:
Currently, the RPMs that are built for ARM specify the wrong architecture name (`arm64`), it should be `aarch64`, as seen in #4277. The end result is that the package won't install on user machines unless they force the installation. The fix is to pass the correct argument to `fpm`.


raksooo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4284#-:-NLzm4J-AKNMWSSQ5ayj:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4287, opened 2023-01-17
Fix deadlock in set_tunnel_addresses

dlon:
Previously, it was possible for a deadlock to occur in `set_tunnel_addresses` if it was called *and failed* at the same time as a route change occurred (i.e., if `split_tunnel_default_route_change_handler` was called at the right time). This PR fixes this by making sure that all mutexes are always released in the correct order.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4287)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ Line wrap the file at 100 chars.                                              Th
   Previously, the installer would abort.
 - Revert to using netsh for DNS config, as some Windows builds did not deal with changes correctly.
   `TALPID_DNS_MODULE` can be used to override this.
+- Fix deadlock that could occur when the default route changed while initializing split tunneling.
 
 ### Changed
 - Update Electron from 19.0.13 to 21.1.1.
--- a/talpid-core/src/split_tunnel/windows/mod.rs
+++ b/talpid-core/src/split_tunnel/windows/mod.rs
@@ -15,7 +15,7 @@ use std::{
     path::{Path, PathBuf},
     sync::{
         atomic::{AtomicBool, Ordering},
-        mpsc as sync_mpsc, Arc, Mutex, RwLock, Weak,
+        mpsc as sync_mpsc, Arc, Mutex, MutexGuard, RwLock, Weak,
     },
     time::Duration,
 };
@@ -639,7 +639,7 @@ impl SplitTunnel {
 
         self._route_change_callback = None;
         let moved_context_mutex = context_mutex.clone();
-        let mut context = context_mutex.lock().unwrap();
+        let context = context_mutex.lock().unwrap();
         let callback = self
             .runtime
             .block_on(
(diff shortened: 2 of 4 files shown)


Jontified (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4287)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 656 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4287#-NM32j5a2P2zdLWmsdum:-NM32j5a2P2zdLWmsdun:b-izan9h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6507270f38f24578d5a6771e47a652735b5f441e/talpid-core/src/split_tunnel/windows/mod.rs#L656)):*
> ```XML
>             )
>             .map(Some)
>             .map_err(|_| Error::RegisterRouteChangeCallback)?;
> ```

Like we discussed on Slack. If the callback is registered and a route event happens before the callback has been returned by this function and then the callback is dropped due to some error (for example a channel send failure) this deadlock could still happen.
This is a pretty amazingly rare occurrence but technically possible. I'm not sure exactly what the best solution for it is.
One way to prevent this is to send the lock back together with the callback after registering and dropping it when there is no possibility of failing, but that's a lot of extra weird complexity for a niched problem.

Jontified (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4287)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 656 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4287#-NM32j5a2P2zdLWmsdum:-NM3EWsn8u1vJUCrUOYC:by5zyri) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6507270f38f24578d5a6771e47a652735b5f441e/talpid-core/src/split_tunnel/windows/mod.rs#L656)):*

Like we discussed on Slack. If the callback is registered and a route event happens before the callback has been returned by this function and then the callback is dropped due to some error (for example a channel send failure) this deadlock could still happen.
This is a pretty amazingly rare occurrence but technically possible. I'm not sure exactly what the best solution for it is.
One way to prevent this is to send the lock back together with the callback after registering and dropping it when there is no possibility of failing, but that's a lot of extra weird complexity for a niched problem.
</blockquote></details>

After discussing this further and neither of us coming up with a realistic way for a channel error happening unless there is some other bug and there not being some obvious good solution to avoid this I'll approve.

Although perhaps you should make a comment about the potential deadlock in order to make debugging it easier in the future (so that people know that a channel failure here could trigger a deadlock in very specific circumstances).

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4287)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`talpid-core/src/split_tunnel/windows/mod.rs` line 656 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4287#-NM32j5a2P2zdLWmsdum:-NM3GEbhAOF8x0_v61Ed:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6507270f38f24578d5a6771e47a652735b5f441e/talpid-core/src/split_tunnel/windows/mod.rs#L656)):*

After discussing this further and neither of us coming up with a realistic way for a channel error happening unless there is some other bug and there not being some obvious good solution to avoid this I'll approve.

Although perhaps you should make a comment about the potential deadlock in order to make debugging it easier in the future (so that people know that a channel failure here could trigger a deadlock in very specific circumstances).
</blockquote></details>

Done.

Jontified (approved):
> Reviewed 3 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4287)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4322, opened 2023-01-31
Stricter cargo audit (deny warnings)

faern:
We do explicitly ignore `RUSTSEC-2021-0145`. But that CVE is only an info level warning, so just `cargo audit` did not trigger an error on it anyway. `--deny warnings` was needed for that. This PR makes our audit check stricter by denying all warnings.

This means that minor stuff such as unmaintained dependencies will trigger an error. That can be annoying. But I figure we can then explicitly add that CVE to this list when it happens. I think that's better than silently ignoring warnings. This way we will get flagged when something is unmaintained etc and we can act on it.

This PR fixes a warnings we had: Upgrade `bumpalo` to avoid [RUSTSEC-2022-0078](https://rustsec.org/advisories/RUSTSEC-2022-0078)

I also upgrade `env_logger`. Since we were mentioning that as a blocker for not ignoring `RUSTSEC-2021-0145` but `env_logger` has already fixed the issue, I figured we better upgrade. We are still pulling in `atty` against our will in a few places, but now they are at least fewer.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4322)

### DIFF
--- a/.github/workflows/cargo-audit.yml
+++ b/.github/workflows/cargo-audit.yml
@@ -37,9 +37,8 @@ jobs:
         #
         # RUSTSEC-2021-0145: The vulnerability affects custom global allocators,
         # so it should be safe to ignore it. Stop ignoring the warning once
-        # atty has been replaced in clap and env_logger:
+        # atty has been replaced in clap (when we upgrade to clap 4):
         # https://github.com/clap-rs/clap/pull/4249
-        # https://github.com/rust-cli/env_logger/pull/246
         run: |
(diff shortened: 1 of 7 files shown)


pinkisemils (approved):
> Reviewed 5 of 7 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4322)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4324, opened 2023-02-01
Fix boot issues in packet tunnel

pronebird:
This should fix the issue with the tunnel being stuck in connecting state indefinitely, presumably because it's waiting for path update from path monitor, which is again, presumably, not being dispatched. It was reported that rebooting device may cause this to happen.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4324)

### DIFF
--- a/ios/MullvadLogging/Logging.swift
+++ b/ios/MullvadLogging/Logging.swift
@@ -9,71 +9,84 @@
 import Foundation
 @_exported import Logging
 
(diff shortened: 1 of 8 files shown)


pronebird:
Closing in favor of https://github.com/mullvad/mullvadvpn-app/pull/4330 since we found that it was something else.

PR #4330, opened 2023-02-01
Handle tunnel issue after reboot

pronebird:
We received a few reports that the new TF beta remains stuck in connecting state after device reboot. After doing a bit of research I realized that on iOS file system and keychain are locked until device is first unlocked. We do explicitly specify that kind of strategy for keychain entries and it's implicit default for filesystem where we store logs and relays. This is of course unless specified otherwise for each individual file or via entitlements for the entire project and we do not do any of that.

Unlike UIKit (UIApplication), network extension doesn't seem to have any callbacks telling it when data protection is being lifted. Therefore for now the strategy is to exit the tunnel process immediately on failure when reading from keychain given that the error is permission related. This should trigger iOS to restart the tunnel at some point later and hopefully after device is first unlocked.

In this PR I also reworked a few more things such as:

1. Logging system initialization is now done via `LoggerBuilder` which makes it possible to catch file system error during configuration. Previously we'd ignore them, which we do in production anyway but it's still more accessible now than before when debugging the root cause of why logging system does not work.
2. `TextFileOutputStream` was switched to use serial `DispatchQueue`. TBH I have impression that global queues are concurrent, so that was changed to make sure we don't have races.
3. `DispatchIO` initializer was switched to the one that accepts file descriptor. That's because the convenience initializer accepting file path seems to either fail silently or open file lazily.
4. Show all packet tunnel errors in GUI banner. Previously only WireGuard error were displayed.
5. Debug only: add button to share problem report outside of our API. Useful during testing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330)


pinkisemils (commented):
> Reviewed 6 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadLogging/TextFileOutputStream.swift` line 46 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330#-NNCl9Xq7kYG89zU0grm:-NNCl9XrEDosVJb7gTVg:b7uoyiz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2f46a2fb8f299468da359918e64b777f6920b92/ios/MullvadLogging/TextFileOutputStream.swift#L46)):*
> ```Swift
> 
>         let fd = fileURL.path.withCString { path in
>             return Darwin.open(path, oflag, mode)
> ```

I presume it's OK to have a blocking syscall here since this isn't called from the UI thread ever, is it?
___
*[`ios/PacketTunnel/PacketTunnelProvider.swift` line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330#-NNClcQPBKvmQohPFZHZ:-NNClcQPBKvmQohPFZH_:bfmo7wc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2f46a2fb8f299468da359918e64b777f6920b92/ios/PacketTunnel/PacketTunnelProvider.swift#L94)):*
> ```Swift
>         let errors: [PacketTunnelErrorWrapper?] = [
>             wgError.flatMap { PacketTunnelErrorWrapper(error: $0) },
>             configurationError.flatMap { PacketTunnelErrorWrapper(error: $0) },
> ```

Is it not possible to just pass a function as a parameter to `flatMap` here ? 
Kind of like
```swift
configurationError.flatMap(PacketTunnelErrorWrapper)
```

Even if it is, this doesn't need to be changed if the current version follows swift conventions.

pinkisemils (approved):
> Reviewed 2 of 8 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330)* status: all files reviewed, 2 unresolved discussions (waiting on @pronebird)

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`ios/MullvadLogging/TextFileOutputStream.swift` line 46 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330#-NNCl9Xq7kYG89zU0grm:-NNCmf6u-niUkFP2L6wK:b-1f7lc1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2f46a2fb8f299468da359918e64b777f6920b92/ios/MullvadLogging/TextFileOutputStream.swift#L46)):*

I presume it's OK to have a blocking syscall here since this isn't called from the UI thread ever, is it?
</blockquote></details>

Reverted to use DispatchIO initializer that accepts file path. That should perhaps open file asynchronously in lazy manner.
___
*[`ios/PacketTunnel/PacketTunnelProvider.swift` line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330#-NNClcQPBKvmQohPFZHZ:-NNCqVIrEUUWGzYjO-_G:b-auxqk9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2f46a2fb8f299468da359918e64b777f6920b92/ios/PacketTunnel/PacketTunnelProvider.swift#L94)):*

Is it not possible to just pass a function as a parameter to `flatMap` here ? 
Kind of like
```swift
configurationError.flatMap(PacketTunnelErrorWrapper)
```

Even if it is, this doesn't need to be changed if the current version follows swift conventions.
</blockquote></details>

Ah Rust spoiled you. In Swift the equivalent would be:

```
configurationError.flatMap(PacketTunnelErrorWrapper.init(error:))
```

I prefer a closure with anonymous `$0` argument instead.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`ios/PacketTunnel/PacketTunnelProvider.swift` line 440 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330#-NNHY8fl9tcAHGa4WxAS:-NNHY8fl9tcAHGa4WxAT:b-3nh1dw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f4d4d426d5dfa4f5e015e22346b0adb7c275df0f/ios/PacketTunnel/PacketTunnelProvider.swift#L440)):*
> ```Swift
>         timer.activate()
> 
>         tunnelStartupFailureRecoveryTimer?.cancel()
> ```

Should the tunnel recovery timer be cancelled here? If so, why not call `cancelTunnelStartupFailureRecovery()` ?

pinkisemils (commented):
Unit tests are still failing, but I have tested this relatedly and it doesn't leak, nor does it fail to connect on boot - great job.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330#-:-NNHYWn75scHT6EgYQGT:b92y88s)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

pinkisemils (commented):
> Reviewed 3 of 4 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`ios/PacketTunnel/PacketTunnelProvider.swift` line 440 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330#-NNHY8fl9tcAHGa4WxAS:-NNLVLkaEMGD_GTHHdsM:b-7nuoiu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f4d4d426d5dfa4f5e015e22346b0adb7c275df0f/ios/PacketTunnel/PacketTunnelProvider.swift#L440)):*
> Should the tunnel recovery timer be cancelled here? 

I prefer to cancel the previously stored timer to make this function idempotent and to avoid running multiple timers if there is a mistake in my program.

> If so, why not call `cancelTunnelStartupFailureRecovery()`?

The subtle difference is that `cancelTunnelStartupFailureRecovery()` would also nullify `tunnelStartupFailureRecoveryTimer` after cancellation.

In this approach I simply replace the `tunnelStartupFailureRecoveryTimer` variable with new timer.

It's one line of code either way.

pronebird (commented):
Awesome :+1: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330#-:-NNLZZB4AZhi9oJ9cvwN:b-blkhx8)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

pinkisemils (approved):
:lgtm:, again! :sweat_smile: 

> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330#-:-NNLfKtn1b5Nn5Si3SzI:b-69vynx)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4330)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4341, opened 2023-02-08
Disable retry strategy for access tokens

pronebird:
Requests that require authorization already implement their own retry strategy. Hence access token requests should be a part of that strategy and not behave as each having their individual strategy.

Previously a request requiring authentication configured with 3 retry attempts would perform 9 calls (3 * 3) to obtain access token in the worst case. That's because each individual request to obtain access token was configured with default retry strategy (3 retry attempts). With this change exactly 3 attempts will be made in total.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4341)

### DIFF
--- a/ios/MullvadREST/RESTAPIProxy.swift
+++ b/ios/MullvadREST/RESTAPIProxy.swift
@@ -130,10 +130,7 @@ extension REST {
 
                     return requestBuilder.getRequest()
                 },
-                authorizationProvider: createAuthorizationProvider(
-                    accountNumber: accountNumber,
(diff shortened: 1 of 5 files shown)


pinkisemils (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4341#-:-NNls8t0D5pcbg5ru-f6:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4344, opened 2023-02-08
WIP: Run clippy on all desktop platforms

faern:
We have *lots* of clippy warnings because we have only run clippy in CI on Linux up until now.

We might not want to go and fix all of this in a big chunk. But it would be great if Rust developers try this out and tackle a few lints here and there until we are down to zero clippy issues.

We still don't run clippy on Android. We should add that also.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4344)

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -9,7 +9,10 @@ on:
   workflow_dispatch:
 jobs:
   clippy_check:
-    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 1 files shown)


dlon:
Closing this due to https://github.com/mullvad/mullvadvpn-app/pull/4642 and https://github.com/mullvad/mullvadvpn-app/pull/4611.

PR #4346, opened 2023-02-08
Add Clippy and the Windows toolchain in the Linux build container

faern:
This makes the container ~560 MiB larger (1.86 -> 2.45 GiB), but as a reward we can easily run `cargo check --target x86_64-pc-windows-gnu` and `cargo clippy --target x86_64-pc-windows-gnu` on our Linux dev machines, something at least I find pretty cumbersome at the moment.

This can potentially also help with #4344, allowing us to have a single job and container running clippy for both Linux and Windows at least, potentially speeding up the CI?

I have not done any deep research, but it seems hard to do the same thing for macOS sadly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4346)

PR #4349, opened 2023-02-09
Bump linux-container-image to mullvadvpn-app-build:20d7550f1

faern:
Update the Linux build container image to include #4346 and Rust 1.67

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4349)

### DIFF
--- a/.github/workflows/daemon.yml
+++ b/.github/workflows/daemon.yml
@@ -69,7 +69,9 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Checkout binaries submodule
-        run: git submodule update --init --depth=1 dist-assets/binaries
+        run: |
+          git config --global --add safe.directory '*'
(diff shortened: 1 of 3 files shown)


faern:
I think this is ready for review now. At least I'm OK with the hack around submodule checking out. I suppose the issue is that some directory is owned by the wrong user in the container. But any fix to solve that would also be at least one bash command, so it would not be shorter than this.

albin-mullvad (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4349#-:-NPYSVgVBYPCdYHa2p3y:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4353, opened 2023-02-10
Allow quantum-resistant tunnels (PQ) to work together with multihop

Jontified:
This PR makes it possible to use PQ while using wireguards multihop. It does this by modifying the firewall API and doing two PQ handshakes through the WG tunnel to establish the PQ PSKs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ Line wrap the file at 100 chars.                                              Th
   "off", but this might change in the future.
 - Update OpenVPN to 2.6.0 from 2.5.3.
 - Update OpenSSL to 1.1.1t from 1.1.1j.
+- Post-Quantum secure tunnels and multihop can now be used at the same time.
 
 #### Windows
 - Remove automatic fallback to wireguard-go. This is done as a first step before fully
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -197,10 +197,11 @@ metadata that might be useful.
 To establish a quantum-resistant tunnel, a pre-shared key (PSK) is derived using a quantum-safe
 key encapsulation mechanism (KEM) with the relay. This is achieved by initiating a regular
 WireGuard tunnel to the relay and deriving the PSK within the tunnel.
-The PSK is saved on the relay and the client, along with a new client generated ephemeral WireGuard
-key. Subsequently, a new tunnel is created using the new WireGuard key and the PSK, ensuring that
-the tunnel is quantum-resistant.
-See [this](../talpid-tunnel-config-client/proto/tunnel_config.proto) for more details on the protocol.
+The PSK is stored in memory on the relay and the client, along with a new client generated ephemeral
+WireGuard key. Subsequently, a new tunnel is created using the new WireGuard key and the PSK,
+ensuring that the tunnel is quantum-resistant.
+See the [protocol definition file](../talpid-tunnel-config-client/proto/tunnel_config.proto) for
+more details on the protocol.
 
 #### Quantum-resistant tunnels & Multihop
 
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -190,6 +190,30 @@ metadata that might be useful.
 
 ### Firewall integration
 
+### Connection logic
+
+#### Quantum-resistant tunnels
+
+To establish a quantum-resistant tunnel, a pre-shared key (PSK) is derived using a quantum-safe
+key encapsulation mechanism (KEM) with the relay. This is achieved by initiating a regular
+WireGuard tunnel to the relay and deriving the PSK within the tunnel.
+The PSK is saved on the relay and the client, along with a new client generated ephemeral WireGuard
+key. Subsequently, a new tunnel is created using the new WireGuard key and the PSK, ensuring that
+the tunnel is quantum-resistant.
+See [this](../talpid-tunnel-config-client/proto/tunnel_config.proto) for more details on the protocol.
+
+#### Quantum-resistant tunnels & Multihop
+
+To create a multihop tunnel where both hops are quantum resistant the client must negotiate a unique
+PSK with both the entry and the exit relay separately. It must use the same ephemeral WireGuard key
+on both relays since the end result (just as with regular multihop tunnels) is two peers on a
+single WireGuard interface, which can only have a single key for the local peer.
+
+The PSKs are established by first creating a regular multihop tunnel to the exit via the entry relay
+and negotiate a PSK with the exit. Then establish a regular tunnel to just the entry and negotiate a
+PSK with it. Lastly the client can set up a multihop tunnel using the new ephemeral WireGuard key
+and the two PSKs via the entry to the exit.
+
 ### Detecting device offline
 
 The tunnel state machine has an offline monitor that tries to detect when a device will certainly
--- a/mullvad-cli/src/cmds/relay.rs
+++ b/mullvad-cli/src/cmds/relay.rs
@@ -592,26 +592,6 @@ impl Relay {
         if let Some(entry) = matches.values_of("entry location") {
             wireguard_constraints.entry_location = parse_entry_location_constraint(entry);
             let use_multihop = wireguard_constraints.entry_location.is_some();
-            if use_multihop {
-                let quantum_resistant = rpc
-                    .get_settings(())
-                    .await?
-                    .into_inner()
-                    .tunnel_options
-                    .unwrap()
-                    .wireguard
-                    .unwrap()
-                    .quantum_resistant;
-                if quantum_resistant
-                    == Some(types::QuantumResistantState {
-                        state: i32::from(types::quantum_resistant_state::State::On),
-                    })
-                {
-                    return Err(Error::CommandFailed(
-                        "Quantum resistant tunnels do not work when multihop is enabled",
-                    ));
-                }
-            }
             wireguard_constraints.use_multihop = use_multihop;
         }
 
--- a/mullvad-cli/src/cmds/tunnel.rs
+++ b/mullvad-cli/src/cmds/tunnel.rs
@@ -251,27 +251,6 @@ impl Tunnel {
             _ => unreachable!("invalid PQ state"),
         };
         let mut rpc = new_rpc_client().await?;
-        let settings = rpc.get_settings(()).await?;
-        if quantum_resistant == types::quantum_resistant_state::State::On {
-            let multihop_is_enabled = settings
-                .into_inner()
-                .relay_settings
-                .unwrap()
-                .endpoint
-                .and_then(|endpoint| {
-                    if let types::relay_settings::Endpoint::Normal(settings) = endpoint {
-                        Some(settings.wireguard_constraints.unwrap().use_multihop)
-                    } else {
-                        None
-                    }
-                })
-                .unwrap_or(false);
-            if multihop_is_enabled {
-                return Err(Error::CommandFailed(
-                    "Quantum resistant tunnels do not work when multihop is enabled",
-                ));
-            }
-        }
         rpc.set_quantum_resistant_tunnel(types::QuantumResistantState {
             state: i32::from(quantum_resistant),
         })
--- a/talpid-core/Cargo.toml
+++ b/talpid-core/Cargo.toml
@@ -103,6 +103,7 @@ features = [
     "Win32_NetworkManagement_Ndis",
     "Win32_UI_Shell",
     "Win32_UI_WindowsAndMessaging",
+    "Win32_System_SystemInformation",
 ]
 
 [build-dependencies]
--- a/talpid-core/src/firewall/linux.rs
+++ b/talpid-core/src/firewall/linux.rs
@@ -576,8 +576,12 @@ impl<'a> PolicyBatch<'a> {
                             self.add_allow_tunnel_rules(&tunnel.interface)?;
                         }
                         AllowedTunnelTraffic::None => (),
-                        AllowedTunnelTraffic::Only(endpoint) => {
+                        AllowedTunnelTraffic::One(endpoint) => {
                             self.add_allow_in_tunnel_endpoint_rules(&tunnel.interface, endpoint)?;
+                        }
+                        AllowedTunnelTraffic::Two(endpoint1, endpoint2) => {
+                            self.add_allow_in_tunnel_endpoint_rules(&tunnel.interface, endpoint1)?;
+                            self.add_allow_in_tunnel_endpoint_rules(&tunnel.interface, endpoint2)?;
                         }
                     }
                     if *allow_lan {
--- a/talpid-core/src/firewall/macos.rs
+++ b/talpid-core/src/firewall/macos.rs
@@ -130,7 +130,7 @@ impl Firewall {
 
                 if let Some(tunnel) = tunnel {
                     rules.extend(
-                        self.get_allow_tunnel_rule(&tunnel.interface, allowed_tunnel_traffic)?
+                        self.get_allow_tunnel_rules(&tunnel.interface, allowed_tunnel_traffic)?
                             .into_iter(),
                     );
                 }
@@ -159,7 +159,7 @@ impl Firewall {
                 rules.append(&mut self.get_block_dns_rules()?);
 
                 rules.extend(
-                    self.get_allow_tunnel_rule(
+                    self.get_allow_tunnel_rules(
                         tunnel.interface.as_str(),
                         &AllowedTunnelTraffic::All,
                     )?
@@ -330,28 +330,55 @@ impl Firewall {
         Ok(vec![block_tcp_dns_rule, block_udp_dns_rule])
     }
 
-    fn get_allow_tunnel_rule(
+    fn base_rule(
         &self,
+        action: FilterRuleAction,
         tunnel_interface: &str,
-        allowed_traffic: &AllowedTunnelTraffic,
-    ) -> Result<Option<pfctl::FilterRule>> {
-        let mut rule_builder = self.create_rule_builder(FilterRuleAction::Pass);
-        let mut base_rule = rule_builder
+    ) -> pfctl::FilterRuleBuilder {
+        let mut rule_builder = self.create_rule_builder(action);
+        rule_builder
             .quick(true)
             .interface(tunnel_interface)
             .keep_state(pfctl::StatePolicy::Keep)
             .tcp_flags(Self::get_tcp_flags());
-        match allowed_traffic {
-            AllowedTunnelTraffic::Only(endpoint) => {
+        rule_builder
+    }
+
+    fn get_allow_tunnel_rules(
+        &self,
+        tunnel_interface: &str,
+        allowed_traffic: &AllowedTunnelTraffic,
+    ) -> Result<Vec<pfctl::FilterRule>> {
+        Ok(match allowed_traffic {
+            AllowedTunnelTraffic::One(endpoint) => {
+                let mut base_rule = &mut self.base_rule(FilterRuleAction::Pass, tunnel_interface);
                 let pfctl_proto = as_pfctl_proto(endpoint.protocol);
                 base_rule = base_rule.to(endpoint.address).proto(pfctl_proto);
+                vec![base_rule.build()?]
             }
-            AllowedTunnelTraffic::All => {}
+            AllowedTunnelTraffic::Two(endpoint1, endpoint2) => {
+                let mut rules = Vec::with_capacity(2);
+
+                let mut base_rule = &mut self.base_rule(FilterRuleAction::Pass, tunnel_interface);
+                let pfctl_proto = as_pfctl_proto(endpoint1.protocol);
+                base_rule = base_rule.to(endpoint1.address).proto(pfctl_proto);
+                rules.push(base_rule.build()?);
+
+                let mut base_rule = &mut self.base_rule(FilterRuleAction::Pass, tunnel_interface);
+                let pfctl_proto = as_pfctl_proto(endpoint2.protocol);
+                base_rule = base_rule.to(endpoint2.address).proto(pfctl_proto);
+                rules.push(base_rule.build()?);
+
+                rules
+            }
+            AllowedTunnelTraffic::All => {
+                let base_rule = &mut self.base_rule(FilterRuleAction::Pass, tunnel_interface);
+                vec![base_rule.build()?]
+            }
             AllowedTunnelTraffic::None => {
-                return Ok(None);
+                vec![]
             }
-        };
-        Ok(Some(base_rule.build()?))
+        })
     }
 
     fn get_allow_loopback_rules(&self) -> Result<Vec<pfctl::FilterRule>> {
--- a/talpid-core/src/firewall/windows.rs
+++ b/talpid-core/src/firewall/windows.rs
@@ -168,27 +168,52 @@ impl Firewall {
             ptr::null()
         };
 
-        let allowed_tun_ip;
-        let allowed_tunnel_endpoint =
-            if let AllowedTunnelTraffic::Only(endpoint) = allowed_tunnel_traffic {
-                allowed_tun_ip = widestring_ip(endpoint.address.ip());
-                Some(WinFwEndpoint {
-                    ip: allowed_tun_ip.as_ptr(),
-                    port: endpoint.address.port(),
-                    protocol: WinFwProt::from(endpoint.protocol),
-                })
-            } else {
-                None
-            };
+        // SAFETY: `allowed_tun_ips`, `entry_endpoint` and `exit_endpoint` must not be dropped until
+        // `WinFw_ApplyPolicyConnecting` has returned.
+        let mut allowed_tun_ips = [WideCString::new(), WideCString::new()];
+        let (entry_endpoint, exit_endpoint) = match allowed_tunnel_traffic {
+            AllowedTunnelTraffic::One(endpoint) => {
+                allowed_tun_ips[0] = widestring_ip(endpoint.address.ip());
+                (
+                    Some(WinFwEndpoint {
+                        ip: allowed_tun_ips[0].as_ptr(),
+                        port: endpoint.address.port(),
+                        protocol: WinFwProt::from(endpoint.protocol),
+                    }),
+                    None,
+                )
+            }
+            AllowedTunnelTraffic::Two(endpoint1, endpoint2) => {
+                allowed_tun_ips[0] = widestring_ip(endpoint1.address.ip());
+                let entry_endpoint = Some(WinFwEndpoint {
+                    ip: allowed_tun_ips[0].as_ptr(),
+                    port: endpoint1.address.port(),
+                    protocol: WinFwProt::from(endpoint1.protocol),
+                });
+                allowed_tun_ips[1] = widestring_ip(endpoint2.address.ip());
+                let exit_endpoint = Some(WinFwEndpoint {
+                    ip: allowed_tun_ips[1].as_ptr(),
+                    port: endpoint2.address.port(),
+                    protocol: WinFwProt::from(endpoint2.protocol),
+                });
+                (entry_endpoint, exit_endpoint)
+            }
+            AllowedTunnelTraffic::None | AllowedTunnelTraffic::All => (None, None),
+        };
+
         let allowed_tunnel_traffic = WinFwAllowedTunnelTraffic {
             type_: WinFwAllowedTunnelTrafficType::from(allowed_tunnel_traffic),
-            endpoint: allowed_tunnel_endpoint
+            entry_endpoint: entry_endpoint
+                .as_ref()
+                .map(|ep| ep as *const _)
+                .unwrap_or(ptr::null()),
+            exit_endpoint: exit_endpoint
                 .as_ref()
                 .map(|ep| ep as *const _)
                 .unwrap_or(ptr::null()),
         };
 
-        unsafe {
+        let res = unsafe {
             WinFw_ApplyPolicyConnecting(
                 winfw_settings,
                 &winfw_relay,
@@ -199,7 +224,14 @@ impl Firewall {
             )
             .into_result()
             .map_err(Error::ApplyingConnectingPolicy)
-        }
+        };
+        // SAFETY: All of these hold stack allocated memory which is pointed to by
+        // `allowed_tunnel_traffic` and must remain allocated until `WinFw_ApplyPolicyConnecting`
+        // has returned.
+        drop(allowed_tun_ips);
+        drop(entry_endpoint);
+        drop(exit_endpoint);
+        res
     }
 
     fn set_connected_state(
@@ -439,7 +471,8 @@ mod winfw {
     #[repr(C)]
     pub struct WinFwAllowedTunnelTraffic {
         pub type_: WinFwAllowedTunnelTrafficType,
-        pub endpoint: *const WinFwEndpoint,
+        pub entry_endpoint: *const WinFwEndpoint,
+        pub exit_endpoint: *const WinFwEndpoint,
     }
 
     #[repr(u8)]
@@ -447,7 +480,8 @@ mod winfw {
     pub enum WinFwAllowedTunnelTrafficType {
         None,
         All,
-        Only,
+        One,
+        Two,
     }
 
     impl From<&AllowedTunnelTraffic> for WinFwAllowedTunnelTrafficType {
@@ -455,7 +489,8 @@ mod winfw {
             match traffic {
                 AllowedTunnelTraffic::None => WinFwAllowedTunnelTrafficType::None,
                 AllowedTunnelTraffic::All => WinFwAllowedTunnelTrafficType::All,
-                AllowedTunnelTraffic::Only(..) => WinFwAllowedTunnelTrafficType::Only,
+                AllowedTunnelTraffic::One(..) => WinFwAllowedTunnelTrafficType::One,
+                AllowedTunnelTraffic::Two(..) => WinFwAllowedTunnelTrafficType::Two,
             }
         }
     }
--- a/talpid-core/src/firewall/windows.rs
+++ b/talpid-core/src/firewall/windows.rs
@@ -168,15 +168,16 @@ impl Firewall {
             ptr::null()
         };
 
-        // SAFETY: `allowed_tun_ips`, `entry_endpoint` and `exit_endpoint` must not be dropped until
+        // SAFETY: `endpoint1_ip`, `endpoint2_ip`, `endpoint1` and `endpoint2` must not be dropped until
         // `WinFw_ApplyPolicyConnecting` has returned.
-        let mut allowed_tun_ips = [WideCString::new(), WideCString::new()];
-        let (entry_endpoint, exit_endpoint) = match allowed_tunnel_traffic {
+        let mut endpoint1_ip = WideCString::new();
+        let mut endpoint2_ip = WideCString::new();
+        let (endpoint1, endpoint2) = match allowed_tunnel_traffic {
             AllowedTunnelTraffic::One(endpoint) => {
-                allowed_tun_ips[0] = widestring_ip(endpoint.address.ip());
+                endpoint1_ip = widestring_ip(endpoint.address.ip());
                 (
                     Some(WinFwEndpoint {
-                        ip: allowed_tun_ips[0].as_ptr(),
+                        ip: endpoint1_ip.as_ptr(),
                         port: endpoint.address.port(),
                         protocol: WinFwProt::from(endpoint.protocol),
                     }),
@@ -184,30 +185,30 @@ impl Firewall {
                 )
             }
             AllowedTunnelTraffic::Two(endpoint1, endpoint2) => {
-                allowed_tun_ips[0] = widestring_ip(endpoint1.address.ip());
-                let entry_endpoint = Some(WinFwEndpoint {
-                    ip: allowed_tun_ips[0].as_ptr(),
+                endpoint1_ip = widestring_ip(endpoint1.address.ip());
+                let endpoint1 = Some(WinFwEndpoint {
+                    ip: endpoint1_ip.as_ptr(),
                     port: endpoint1.address.port(),
                     protocol: WinFwProt::from(endpoint1.protocol),
                 });
-                allowed_tun_ips[1] = widestring_ip(endpoint2.address.ip());
-                let exit_endpoint = Some(WinFwEndpoint {
-                    ip: allowed_tun_ips[1].as_ptr(),
+                endpoint2_ip = widestring_ip(endpoint2.address.ip());
+                let endpoint2 = Some(WinFwEndpoint {
+                    ip: endpoint2_ip.as_ptr(),
                     port: endpoint2.address.port(),
                     protocol: WinFwProt::from(endpoint2.protocol),
                 });
-                (entry_endpoint, exit_endpoint)
+                (endpoint1, endpoint2)
             }
             AllowedTunnelTraffic::None | AllowedTunnelTraffic::All => (None, None),
         };
 
         let allowed_tunnel_traffic = WinFwAllowedTunnelTraffic {
             type_: WinFwAllowedTunnelTrafficType::from(allowed_tunnel_traffic),
-            entry_endpoint: entry_endpoint
+            endpoint1: endpoint1
                 .as_ref()
                 .map(|ep| ep as *const _)
                 .unwrap_or(ptr::null()),
-            exit_endpoint: exit_endpoint
+            endpoint2: endpoint2
                 .as_ref()
                 .map(|ep| ep as *const _)
                 .unwrap_or(ptr::null()),
@@ -228,9 +229,10 @@ impl Firewall {
         // SAFETY: All of these hold stack allocated memory which is pointed to by
         // `allowed_tunnel_traffic` and must remain allocated until `WinFw_ApplyPolicyConnecting`
         // has returned.
-        drop(allowed_tun_ips);
-        drop(entry_endpoint);
-        drop(exit_endpoint);
+        drop(endpoint1_ip);
+        drop(endpoint2_ip);
+        drop(endpoint1);
+        drop(endpoint2);
         res
     }
 
@@ -471,8 +473,8 @@ mod winfw {
     #[repr(C)]
     pub struct WinFwAllowedTunnelTraffic {
         pub type_: WinFwAllowedTunnelTrafficType,
-        pub entry_endpoint: *const WinFwEndpoint,
-        pub exit_endpoint: *const WinFwEndpoint,
+        pub endpoint1: *const WinFwEndpoint,
+        pub endpoint2: *const WinFwEndpoint,
     }
 
     #[repr(u8)]
--- a/talpid-core/src/tunnel/mod.rs
+++ b/talpid-core/src/tunnel/mod.rs
@@ -143,18 +143,7 @@ impl TunnelMonitor {
         let config = talpid_wireguard::config::Config::from_parameters(params)?;
         let monitor = talpid_wireguard::WireguardMonitor::start(
             config,
-            if params.options.quantum_resistant {
-                Some(
-                    params
-                        .connection
-                        .exit_peer
-                        .as_ref()
-                        .map(|peer| peer.public_key.clone())
-                        .unwrap_or_else(|| params.connection.peer.public_key.clone()),
-                )
-            } else {
-                None
-            },
+            params.options.quantum_resistant,
             log.as_deref(),
             args,
         )?;
--- a/talpid-tunnel-config-client/examples/psk-exchange.rs
+++ b/talpid-tunnel-config-client/examples/psk-exchange.rs
@@ -2,7 +2,7 @@
 //! Useful to test this crate's implementation.
 
 use std::net::IpAddr;
-use talpid_types::net::wireguard::PublicKey;
+use talpid_types::net::wireguard::{PrivateKey, PublicKey};
 
 #[tokio::main]
 async fn main() {
@@ -16,10 +16,15 @@ async fn main() {
         .next()
         .expect("Give WireGuard public key as second argument");
     let pubkey = PublicKey::from_base64(pubkey_string.trim()).expect("Invalid public key");
+    let private_key = PrivateKey::new_from_random();
 
-    let (private_key, psk) = talpid_tunnel_config_client::push_pq_key(tuncfg_server_ip, pubkey)
-        .await
-        .unwrap();
+    let psk = talpid_tunnel_config_client::push_pq_key(
+        tuncfg_server_ip,
+        pubkey,
+        private_key.public_key(),
+    )
+    .await
+    .unwrap();
 
     println!("private key: {private_key:?}");
     println!("psk: {psk:?}");
--- a/talpid-tunnel-config-client/src/lib.rs
+++ b/talpid-tunnel-config-client/src/lib.rs
@@ -1,5 +1,5 @@
 use std::{fmt, net::IpAddr};
-use talpid_types::net::wireguard::{PresharedKey, PrivateKey, PublicKey};
+use talpid_types::net::wireguard::{PresharedKey, PublicKey};
 use tonic::transport::Channel;
 use zeroize::Zeroize;
 
@@ -70,8 +70,8 @@ pub const CONFIG_SERVICE_PORT: u16 = 1337;
 pub async fn push_pq_key(
     service_address: IpAddr,
     wg_pubkey: PublicKey,
-) -> Result<(PrivateKey, PresharedKey), Error> {
-    let wg_psk_privkey = PrivateKey::new_from_random();
+    wg_psk_pubkey: PublicKey,
+) -> Result<PresharedKey, Error> {
     let (cme_kem_pubkey, cme_kem_secret) = classic_mceliece::generate_keys().await;
     let kyber_keypair = kyber::keypair(&mut rand::thread_rng());
 
@@ -79,7 +79,7 @@ pub async fn push_pq_key(
     let response = client
         .psk_exchange_v1(proto::PskRequestV1 {
             wg_pubkey: wg_pubkey.as_bytes().to_vec(),
-            wg_psk_pubkey: wg_psk_privkey.public_key().as_bytes().to_vec(),
+            wg_psk_pubkey: wg_psk_pubkey.as_bytes().to_vec(),
             kem_pubkeys: vec![
                 proto::KemPubkeyV1 {
                     algorithm_name: classic_mceliece::ALGORITHM_NAME.to_owned(),
@@ -127,7 +127,7 @@ pub async fn push_pq_key(
         shared_secret.zeroize();
     }
 
-    Ok((wg_psk_privkey, PresharedKey::from(psk_data)))
+    Ok(PresharedKey::from(psk_data))
 }
 
 /// Performs `dst = dst ^ src`.
--- a/talpid-types/src/net/mod.rs
+++ b/talpid-types/src/net/mod.rs
@@ -284,15 +284,21 @@ impl fmt::Display for AllowedEndpoint {
 pub enum AllowedTunnelTraffic {
     None,
     All,
-    Only(Endpoint),
+    One(Endpoint),
+    Two(Endpoint, Endpoint),
 }
 
 impl fmt::Display for AllowedTunnelTraffic {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
-        match *self {
+        match self {
             AllowedTunnelTraffic::None => "None".fmt(f),
             AllowedTunnelTraffic::All => "All".fmt(f),
-            AllowedTunnelTraffic::Only(endpoint) => endpoint.fmt(f),
+            AllowedTunnelTraffic::One(endpoint) => endpoint.fmt(f),
+            AllowedTunnelTraffic::Two(endpoint1, endpoint2) => {
+                endpoint1.fmt(f)?;
+                f.write_str(", ")?;
+                endpoint2.fmt(f)
+            }
         }
     }
 }
--- a/talpid-wireguard/src/config.rs
+++ b/talpid-wireguard/src/config.rs
@@ -6,7 +6,7 @@ use std::{
 use talpid_types::net::{obfuscation::ObfuscatorConfig, wireguard, GenericTunnelOptions};
 
 /// Config required to set up a single WireGuard tunnel
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub struct Config {
     /// Contains tunnel endpoint specific config
     pub tunnel: wireguard::TunnelConfig,
--- a/talpid-wireguard/src/lib.rs
+++ b/talpid-wireguard/src/lib.rs
@@ -9,12 +9,13 @@ use futures::future::{abortable, AbortHandle as FutureAbortHandle, BoxFuture, Fu
 use futures::{channel::mpsc, StreamExt};
 #[cfg(target_os = "linux")]
 use lazy_static::lazy_static;
+#[cfg(target_os = "android")]
+use std::borrow::Cow;
 #[cfg(target_os = "linux")]
 use std::env;
 #[cfg(windows)]
 use std::io;
 use std::{
-    borrow::Cow,
     convert::Infallible,
     net::IpAddr,
     path::Path,
@@ -245,11 +246,20 @@ impl WireguardMonitor {
         let endpoint_addrs: Vec<IpAddr> =
             config.peers.iter().map(|peer| peer.endpoint.ip()).collect();
 
+        let (close_obfs_sender, close_obfs_listener) = sync_mpsc::channel();
+        let obfuscator = args.runtime.block_on(maybe_create_obfuscator(
+            &mut config,
+            close_obfs_sender.clone(),
+        ))?;
+
         #[cfg(target_os = "windows")]
         let (setup_done_tx, setup_done_rx) = mpsc::channel(0);
         let tunnel = Self::open_tunnel(
             args.runtime.clone(),
+            #[cfg(target_os = "android")]
             &Self::patch_allowed_ips(&config, psk_negotiation),
+            #[cfg(not(target_os = "android"))]
+            &config,
             log_path,
             args.resource_dir,
             args.tun_provider.clone(),
@@ -260,12 +270,6 @@ impl WireguardMonitor {
         )?;
         let iface_name = tunnel.get_interface_name();
 
-        let (close_obfs_sender, close_obfs_listener) = sync_mpsc::channel();
-
-        let obfuscator = Arc::new(AsyncMutex::new(args.runtime.block_on(
-            maybe_create_obfuscator(&mut config, close_obfs_sender.clone()),
-        )?));
-
         #[cfg(target_os = "android")]
         if let Some(remote_socket_fd) = obfuscator.as_ref().map(|obfs| obfs.remote_socket_fd()) {
             // Exclude remote obfuscation socket or bridge
@@ -274,6 +278,8 @@ impl WireguardMonitor {
                 log::error!("Failed to exclude remote socket fd: {error}");
             }
         }
+
+        let obfuscator = Arc::new(AsyncMutex::new(obfuscator));
 
         let event_callback = Box::new(on_event.clone());
         let (pinger_tx, pinger_rx) = sync_mpsc::channel();
@@ -543,6 +549,7 @@ impl WireguardMonitor {
 
     /// Replace `0.0.0.0/0`/`::/0` with the gateway IPs when `gateway_only` is true.
     /// Used to block traffic to other destinations while connecting on Android.
+    #[cfg(target_os = "android")]
     fn patch_allowed_ips(config: &Config, gateway_only: bool) -> Cow<'_, Config> {
         if gateway_only {
             let mut patched_config = config.clone();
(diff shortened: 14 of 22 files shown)


Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: 0 of 19 files reviewed, 19 unresolved discussions

___
*[`talpid-core/src/firewall/macos.rs` line 373 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFMpG41Vi-caLLAReA:-NOFMpG41Vi-caLLAReB:b59ytm8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-core/src/firewall/macos.rs#L373)):*
> ```XML
>             }
>         };
>         Ok(rules)
> ```

We can return in the match instead
___
*[`talpid-core/src/firewall/windows.rs` line 172 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFNmri7EtP9ZTYdHlK:-NOFNmri7EtP9ZTYdHlL:b-tfveg2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-core/src/firewall/windows.rs#L172)):*
> ```XML
> 
>         // SAFETY: `allowed_tun_ips` and `allowed_tunnel_endpoints` must not be dropped until
>         // `WinFw_ApplyPolicyConnecting` is done with it
> ```

    until `WinFw_ApplyPolicyConnecting` has returned *



_Code quote:_
```XML
        // SAFETY: `allowed_tun_ips` and `allowed_tunnel_endpoints` must not be dropped until
        // `WinFw_ApplyPolicyConnecting` is done with it
```
___
*[`talpid-core/src/firewall/windows.rs` line 462 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFOVDc5J1rMvT3es1f:-NOFOVDc5J1rMvT3es1g:bp6roio) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-core/src/firewall/windows.rs#L462)):*
> ```XML
>         pub type_: WinFwAllowedTunnelTrafficType,
>         pub endpoints: *const WinFwEndpoint,
>         pub endpoints_length: usize,
> ```

Should this be pascalCase?
___
*[`talpid-core/src/tunnel/mod.rs` line 146 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFOcoJAn1CtPBWwfwi:-NOFOcoJAn1CtPBWwfwj:by8vr4h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-core/src/tunnel/mod.rs#L146)):*
> ```XML
>         let monitor = talpid_wireguard::WireguardMonitor::start(
>             config,
>             // TODO: We only need to use a bool here, the actual public key can be found in the
> ```

Remove this TODO
___
*[`talpid-tunnel-config-client/src/lib.rs` line 67 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFOv0W2Wk55NfZtFgP:-NOFOv0W2Wk55NfZtFgQ:b-myw90w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-tunnel-config-client/src/lib.rs#L67)):*
> ```XML
> 
>     let client = new_client(service_address).await;
>     let mut client = client?;
> ```

This can be removed
___
*[`talpid-tunnel-config-client/src/lib.rs` line 79 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFOsXz9OhT3yc-gMno:-NOFOsXz9OhT3yc-gMnp:b-myw90w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-tunnel-config-client/src/lib.rs#L79)):*
> ```XML
>         .await
>         .map_err(Error::GrpcError);
>     let response = response?;
> ```

This can be removed
___
*[`talpid-wireguard/src/lib.rs` line 239 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFP5iCC9QcO7HdQbQd:-NOFP5iDEhaw_qj46fjW:b-684e7m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-wireguard/src/lib.rs#L239)):*
> ```XML
>         args: TunnelArgs<'_, F>,
>     ) -> Result<WireguardMonitor> {
>         let on_event = args.on_event.clone();
> ```

Consider if necessary
___
*[`talpid-wireguard/src/lib.rs` line 248 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFPG7VDpSN88tNJtbP:-NOFPG7VDpSN88tNJtbQ:b-gxoo4m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-wireguard/src/lib.rs#L248)):*
> ```XML
>         let tunnel = Self::open_tunnel(
>             args.runtime.clone(),
>             &config,
> ```

Consider implications
___
*[`talpid-wireguard/src/lib.rs` line 274 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFPW58ACRXF-i-cp8A:-NOFPW58ACRXF-i-cp8B:bf08gv0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-wireguard/src/lib.rs#L274)):*
> ```XML
>             close_obfs_sender.clone(),
>         ))?;
>         let obfuscator = Arc::new(AsyncMutex::new(obfuscator));
> ```

Can be shortened
___
*[`talpid-wireguard/src/lib.rs` line 335 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFQ7DL6x4eXMJ6q52t:-NOFQ7DL6x4eXMJ6q52u:btgmhx3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-wireguard/src/lib.rs#L335)):*
> ```XML
>                     // Set up tunnel to entry do psk
>                     let mut entry_tun_config = config.clone();
>                     // Here we know there are exactly 2 peers, the last one is the exit
> ```

Should we do some verification that there are exactly 2 peers? Perhaps in the future we'll have more than 2 and weird stuff might happen here.
___
*[`talpid-wireguard/src/lib.rs` line 337 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFRLM8BIX8K830huDX:-NOFRLM8BIX8K830huDY:b-4mah9x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-wireguard/src/lib.rs#L337)):*
> ```XML
>                     // Here we know there are exactly 2 peers, the last one is the exit
>                     entry_tun_config.peers.pop();
>                     entry_tun_config
> ```

Consider implications of `Self::patch_allowed_ips`
___
*[`talpid-wireguard/src/lib.rs` line 501 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFRvk360waogs3VJLO:-NOFRvk360waogs3VJLP:bdq1n9g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-wireguard/src/lib.rs#L501)):*
> ```XML
> 
>     /// Reconfigures the tunnel to use the provided config while potentially modifying the config
>     /// and restarting the obfuscation provider. Returns the used config.
> ```

Returns the used config. -> Returns the new config used by the new tunnel*



_Code quote:_
```XML
Returns the used config.
```
___
*[`talpid-wireguard/src/lib.rs` line 610 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFS4v4-yplGV1qsrBG:-NOFS4v4-yplGV1qsrBH:b67f8wv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-wireguard/src/lib.rs#L610)):*
> ```XML
> 
>     async fn perform_psk_negotiation(
>         //obfuscation_handle: Arc<AsyncMutex<Option<ObfuscatorHandle>>>,
> ```

Cleanup
___
*[`talpid-wireguard/src/lib.rs` line 641 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFSFpbABPVk2PSBBFM:-NOFSFpbABPVk2PSBBFN:b67f8wv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-wireguard/src/lib.rs#L641)):*
> ```XML
> 
>         // Restart the obfuscation server
>         //let mut obfs_guard = obfuscation_handle.lock().await;
> ```

Cleanup
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 68 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFV00e1EVj4Oqc4lIW:-NOFV00e1EVj4Oqc4lIX:br4neg9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L68)):*
> ```cpp
>             {
>                 GUID guid;
>                 if (i == 0) {
> ```

Is there a nicer way to do this?
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 126 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFV6Jr6eQpfH63_Qn-:-NOFV6JsD9yXre5dTVVk:b67f8wv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L126)):*
> ```cpp
> 	return true;
> 
>     // TODO: Cleanup
> ```

Cleanup
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp` line 70 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFW85cDvCrcrWVAw01:-NOFW85cDvCrcrWVAw02:b-3xfwho) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp#L70)):*
> ```cpp
>             {
>                 GUID guid;
>                 if (i == 0) {
> ```

Can we improve?
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp` line 75 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFWHm2DjJCF0p-YsOA:-NOFWHm2DjJCF0p-YsOB:b-i5ofw5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp#L75)):*
> ```cpp
>                     guid = MullvadGuids::Filter_Baseline_PermitVpnTunnelService_Ipv4_Exit();
>                 } else {
>                     return false;
> ```

Should return error rather than false?
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp` line 129 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFXAj3CPDO_Q4krETX:-NOFXAj3CPDO_Q4krETY:b67na3p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp#L129)):*
> ```cpp
> 	return true;
> 
>     // TODO: Cleanup
> ```

Clenaup

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: 0 of 19 files reviewed, 20 unresolved discussions

___
*[`docs/architecture.md` line 197 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFY1aYD-AuUveWKHmV:-NOFY1aYD-AuUveWKHmW:b23xms5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/docs/architecture.md#L197)):*
> ```Markdown
> #### Post-quantum
> 
> A post-quantum (PQ) resistant tunnel is established by creating a pre-shared key (PSK)
> ```

This makes it sound like the PSK is created on the client, but the PSK is created through an exchange between the client and server. The client  only generates some key material and sends this to the server.

dlon (changes requested):
> Reviewed 1 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: 1 of 19 files reviewed, 21 unresolved discussions (waiting on @Jontified)

___
*[`talpid-core/src/firewall/windows.rs` line 462 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFOVDc5J1rMvT3es1f:-NOFk09l-6ebb2U79Tzy:bezgaf3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/talpid-core/src/firewall/windows.rs#L462)):*

Should this be pascalCase?
</blockquote></details>

We should be consistent, at least. But personally I think we should use underscores everywhere.
___
*[`talpid-core/src/firewall/windows.rs` line 175 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFjIvYEQT9idRWMjBv:-NOFjIvYEQT9idRWMjBw:b-8cnpoq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/900d1f224798613f9fe27b6f8ccce0811e3a5a5c/talpid-core/src/firewall/windows.rs#L175)):*
> ```XML
>         let mut allowed_tun_ips = [WideCString::new(), WideCString::new()];
>         let allowed_tunnel_endpoints = match allowed_tunnel_traffic {
>             AllowedTunnelTraffic::Only(endpoint) => {
> ```

It seems redundant to have `Only` when `Many` can also take a single endpoint. Maybe we only need `Many`/`Some`? Alternatively, `One` and `Two` or something, since the Windows impl only supports at most two endpoints anyway.
___
*[`talpid-core/src/firewall/windows.rs` line 185 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFjm5L1EfcyU-Iu-ID:-NOFjm5L1EfcyU-Iu-IE:b-8cgdbi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/900d1f224798613f9fe27b6f8ccce0811e3a5a5c/talpid-core/src/firewall/windows.rs#L185)):*
> ```XML
>             AllowedTunnelTraffic::Many(endpoints) => {
>                 Some(endpoints.into_iter().enumerate().map(|(i, endpoint)| {
>                     allowed_tun_ips[i] = widestring_ip(endpoint.address.ip());
> ```

This will panic if `endpoints` contains more than two endpoints.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 68 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFV00e1EVj4Oqc4lIW:-NOFkO891uNUuHENolzB:b-720q41) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L68)):*

Is there a nicer way to do this?
</blockquote></details>

I think it would probably be nicer if `PermitVpnTunnel` took one entry endpoint and one optional exit endpoint instead of a list.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp` line 70 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFW85cDvCrcrWVAw01:-NOFl1yt2tAC_RWCi20Y:b-otilcs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp#L70)):*

Can we improve?
</blockquote></details>

See comment above.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp` line 75 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFWHm2DjJCF0p-YsOA:-NOFl5EkDDuoxle_Y91E:b-35y3ba) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b12c62ccde225475acc1b51fe8cda493e78a91f/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp#L75)):*

Should return error rather than false?
</blockquote></details>

It would probably make sense to throw an exception here. Or make it impossible to pass in more than two endpoints in the first place.

dlon (commented):
> Reviewed 2 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: 3 of 19 files reviewed, 23 unresolved discussions (waiting on @Jontified)

___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 62 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOJPegvE5Z36clYSfzi:-NOJPegvE5Z36clYSfzj:bbm00u6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/900d1f224798613f9fe27b6f8ccce0811e3a5a5c/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L62)):*
> ```cpp
> 
> 		return objectInstaller.addFilter(filterBuilder, conditionBuilder);
>     } else {
> ```

Nit: The `else` is unnecessary.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp` line 64 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOJPTyE9xJC4UeyY-sz:-NOJPTyE9xJC4UeyY-t-:bbm00u6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/900d1f224798613f9fe27b6f8ccce0811e3a5a5c/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp#L64)):*
> ```cpp
> 
> 		return objectInstaller.addFilter(filterBuilder, conditionBuilder);
>     } else {
> ```

Nit: The `else` is unnecessary.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: 2 of 20 files reviewed, 6 unresolved discussions (waiting on @dlon)

___
*[`talpid-core/src/firewall/windows.rs` line 175 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFjIvYEQT9idRWMjBv:-NOOYVom4dhbTgnYsG5c:b-rwildz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/900d1f224798613f9fe27b6f8ccce0811e3a5a5c/talpid-core/src/firewall/windows.rs#L175)):*

It seems redundant to have `Only` when `Many` can also take a single endpoint. Maybe we only need `Many`/`Some`? Alternatively, `One` and `Two` or something, since the Windows impl only supports at most two endpoints anyway.
</blockquote></details>

I went with `One` and `Two` because as we discussed it offloads some of the potential runtime checks to compiletime and it is unlikely we will need to whitelist more than two endpoints in the future. The naming is still not yet set in stone.
___
*[`talpid-core/src/firewall/windows.rs` line 185 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOFjm5L1EfcyU-Iu-ID:-NOJJi-Y7sJIkqwiSO9R:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/900d1f224798613f9fe27b6f8ccce0811e3a5a5c/talpid-core/src/firewall/windows.rs#L185)):*

This will panic if `endpoints` contains more than two endpoints.
</blockquote></details>

Done.

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: 2 of 20 files reviewed, 7 unresolved discussions (waiting on @dlon)

___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp` line 61 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOOb40L31tdAP9av2nk:-NOOb40L31tdAP9av2nl:b-jmdbbr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/900d1f224798613f9fe27b6f8ccce0811e3a5a5c/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp#L61)):*
> ```cpp
> 		conditionBuilder.reset(FWPM_LAYER_ALE_AUTH_CONNECT_V6);
> 
> 		conditionBuilder.add_condition(ConditionInterface::Alias(m_tunnelInterfaceAlias));
> ```

Make all spaces into tabs in all C++ files

dlon (changes requested):
> Reviewed 1 of 19 files at r1, 2 of 17 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: 3 of 20 files reviewed, 4 unresolved discussions (waiting on @Jontified)

___
*[`talpid-core/src/firewall/windows.rs` line 173 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOPhcKV3jgipG5DBXBo:-NOPhcKV3jgipG5DBXBp:b9r0hn1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e07a2bc251ea6b25afdf5c15dcc7325bf0fd1ec3/talpid-core/src/firewall/windows.rs#L173)):*
> ```XML
>         // SAFETY: `allowed_tun_ips`, `entry_endpoint` and `exit_endpoint` must not be dropped until
>         // `WinFw_ApplyPolicyConnecting` has returned.
>         let mut allowed_tun_ips = [WideCString::new(), WideCString::new()];
> ```

There's no need to store these in an array anymore.
___
*[`talpid-core/src/firewall/windows.rs` line 222 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOPhnC-E-pMuamfbEDW:-NOPhnC0EJVYh03B0pE3:b-474ulr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e07a2bc251ea6b25afdf5c15dcc7325bf0fd1ec3/talpid-core/src/firewall/windows.rs#L222)):*
> ```XML
>         // `allowed_tunnel_traffic` and must remain allocated until `WinFw_ApplyPolicyConnecting`
>         // has returned.
>         drop(allowed_tun_ips);
> ```

Are we only manually dropping here to prevent accidents? Maybe we should not.
___
*[`talpid-types/src/net/mod.rs` line 297 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOPiRxx6fRqr4pZ_t4A:-NOPiRxx6fRqr4pZ_t4B:bio2rh7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e07a2bc251ea6b25afdf5c15dcc7325bf0fd1ec3/talpid-types/src/net/mod.rs#L297)):*
> ```XML
>             AllowedTunnelTraffic::All => "All".fmt(f),
>             AllowedTunnelTraffic::One(endpoint) => endpoint.fmt(f),
>             AllowedTunnelTraffic::Two(endpoint1, endpoint2) => {
> ```

Should the endpoints be separate by something, e.g. a comma?
___
*[`talpid-wireguard/src/lib.rs` line 313 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOPitJ42XLAz8oxrn1c:-NOPitJ42XLAz8oxrn1d:b2d68t7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d87248c1c052f6f6f338b60ff3ff284dd617bfd/talpid-wireguard/src/lib.rs#L313)):*
> ```XML
>                 AllowedTunnelTraffic::All
>             };
>             (on_event)(TunnelEvent::InterfaceUp(metadata.clone(), allowed_traffic)).await;
> ```

Why was this removed? Perhaps we purposely punched a hole before adding any routes here.

dlon (changes requested):
> Reviewed 1 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: 3 of 21 files reviewed, 9 unresolved discussions (waiting on @Jontified)

___
*[`talpid-wireguard/src/lib.rs` line 323 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOQAlnS2g3Nu1ddVr7w:-NOQAlnS2g3Nu1ddVr7x:b-k9ga03) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/talpid-wireguard/src/lib.rs#L323)):*
> ```XML
>                 .map_err(CloseMsg::SetupError)?;
> 
>             if psk_negotiation {
> ```

This should probably be moved to a function/method.
___
*[`talpid-wireguard/src/lib.rs` line 324 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOQ1XsM1Cd4j0EQCBSg:-NOQ1XsM1Cd4j0EQCBSh:b-h77bp1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/talpid-wireguard/src/lib.rs#L324)):*
> ```XML
> 
>             if psk_negotiation {
>                 let mut wg_psk_privkey = None;
> ```

You could just generate a key here. It does not need to be an option.
___
*[`talpid-wireguard/src/lib.rs` line 328 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOQBioDCs5L4ls3CJX5:-NOQBioDCs5L4ls3CJX6:bvnly8y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/talpid-wireguard/src/lib.rs#L328)):*
> ```XML
> 
>                 if config.peers.len() > 1 {
>                     assert_eq!(config.peers.len(), 2);
> ```

I'm a bit skeptical of the panicking here. I think we should return an error, even though it's supposed to be unreachable.
___
*[`talpid-wireguard/src/lib.rs` line 342 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOQ6w_m7OnDL6G600ek:-NOQ6w_m7OnDL6G600el:b4299ws) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/talpid-wireguard/src/lib.rs#L342)):*
> ```XML
>                         .allowed_ips
>                         .push(IpNetwork::new(IpAddr::V4(config.ipv4_gateway), 32).unwrap());
>                     let entry_tun_config =
> ```

This adds `10.64.0.1`. So calling `patch_allowed_ips` is unnecessary, as is `entry_tun_config.peers.pop();`.
___
*[`talpid-wireguard/src/lib.rs` line 375 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOQ90nWBufjeZo_5ehD:-NOQ90nWBufjeZo_5ehE:bu8er2t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/talpid-wireguard/src/lib.rs#L375)):*
> ```XML
>                 let wg_psk_privkey = wg_psk_privkey.unwrap_or_else(PrivateKey::new_from_random);
>                 // Set up tunnel to exit and exchange psk
>                 let exit_config = Self::patch_allowed_ips(&config, psk_negotiation);
> ```

All you need here is the original config, since it already swallows everything with `0.0.0.0/0`. Maybe it would make more sense to call `perform_psk_negotiation` for the exit first, so you'd only have to update the WireGuard config and firewall rules twice instead of three times?

I don't think we should use `patch_allowed_ips` at all here. It is intended for Android, and it makes no difference.

dlon (changes requested):
> Reviewed 1 of 19 files at r1, 1 of 1 files at r2, 9 of 17 files at r3, 2 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: 16 of 21 files reviewed, 20 unresolved discussions (waiting on @Jontified)

___
*[`docs/architecture.md` line 197 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk8u6y59BK833Mzf8k:-NOk8u6y59BK833Mzf8l:b-pnjgo6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/docs/architecture.md#L197)):*
> ```Markdown
> #### Post-quantum
> 
> A post-quantum (PQ) resistant tunnel is established by creating a pre-shared key (PSK)
> ```

This should probably be changed to either `quantum-resistant tunnel` or `post-quantum tunnel`.
___
*[`docs/architecture.md` line 201 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk96smEl6OtHpVgJUI:-NOk96snCvxV9dUmN8wc:b-b5qs8m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/docs/architecture.md#L201)):*
> ```Markdown
> the PSK is saved on the relay and on the client together with a new temporary wireguard
> key which the client generates. After this a new tunnel is created which uses the new WG key
> and the PSK, this new tunnel is PQ resistant.
> ```

"PQ safe" or "quantum resistant" (in my opinion).
___
*[`docs/architecture.md` line 208 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk9KLp-dvdpfhsuX_n:-NOk9KLp-dvdpfhsuX_o:b-otilcs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/docs/architecture.md#L208)):*
> ```Markdown
> temporary key and uses it to do a PSK exchange with the entry relay. Then, a normal tunnel is
> established from the entry to the exit relay and a new PSK is exchanged along with the temporary
> key. Finally, a new PQ resistant tunnel is created using the new WG key and the two PSKs to route
> ```

See comment above.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.h` line 35 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk4gQKBUBfWJxAgMwL:-NOk4gQKBUBfWJxAgMwM:b95inm7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.h#L35)):*
> ```objc
> 
> private:
> 	bool AddEndpointFilter(const std::optional<Endpoint> &endpoint, GUID ipv4Guid, GUID ipv6Guid, wfp::FilterBuilder &filterBuilder, IObjectInstaller& objectInstaller);
> ```

Could you change this to `IObjectInstaller &objectInstaller` for consistency?



_Code quote:_
```objc
IObjectInstaller& objectInstaller
```
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 27 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk5QYJ5G4gOIsddpWB:-NOk5QYJ5G4gOIsddpWC:b-jo8yo0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L27)):*
> ```cpp
> }
> 
> bool PermitVpnTunnel::AddEndpointFilter(const std::optional<Endpoint> &endpoint, GUID ipv4Guid, GUID ipv6Guid, wfp::FilterBuilder &filterBuilder, IObjectInstaller& objectInstaller)
> ```

I think the `wfp::FilterBuilder` should be constructed in this function instead of passed in as an argument. Right now, it's possible to pass in an incorrectly initialized filter builder. If you were to construct it here, the function would be completely self-contained. It's more cohesive.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 27 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk7P8mEt4Z7dB1nwUE:-NOk7P8mEt4Z7dB1nwUF:bky9gy7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L27)):*
> ```cpp
> }
> 
> bool PermitVpnTunnel::AddEndpointFilter(const std::optional<Endpoint> &endpoint, GUID ipv4Guid, GUID ipv6Guid, wfp::FilterBuilder &filterBuilder, IObjectInstaller& objectInstaller)
> ```

You could pass the GUIDs as `const GUID &...`.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 27 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk7agMDWY9m3SQ9QK5:-NOk7agMDWY9m3SQ9QK6:b-erq1fs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L27)):*
> ```cpp
> }
> 
> bool PermitVpnTunnel::AddEndpointFilter(const std::optional<Endpoint> &endpoint, GUID ipv4Guid, GUID ipv6Guid, wfp::FilterBuilder &filterBuilder, IObjectInstaller& objectInstaller)
> ```

`IObjectInstaller &objectInstaller`



_Code quote:_
```cpp
IObjectInstaller& objectInstaller
```
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 29 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk67g5CVIHU6YtdmfQ:-NOk67g5CVIHU6YtdmfR:b1lbcab) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L29)):*
> ```cpp
> bool PermitVpnTunnel::AddEndpointFilter(const std::optional<Endpoint> &endpoint, GUID ipv4Guid, GUID ipv6Guid, wfp::FilterBuilder &filterBuilder, IObjectInstaller& objectInstaller)
> {
> 	if (!endpoint.has_value() || endpoint.value().ip.type() == wfp::IpAddress::Ipv4)
> ```

Just my opinion, but I think we should store these conditions in booleans for the purpose of documentation, like we did previously. E.g.,

```
bool shouldAddV4Filter = ...;
```

What the condition is for is currently very non-obvious.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.h` line 8 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk4H1-3zpwuKl21C8D:-NOk4H1-3zpwuKl21C8E:b-lh65ro) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.h#L8)):*
> ```objc
> #include <libwfp/ipaddress.h>
> #include <string>
> #include <vector>
> ```

You should probably import `optional`, and `vector` is no longer needed.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.h` line 25 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk3bwpChQLskzKVNT2:-NOk3bwpChQLskzKVNT3:b95inm7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.h#L25)):*
> ```objc
> 
> private:
> 	bool AddEndpointFilter(const std::optional<PermitVpnTunnel::Endpoint> &endpoint, GUID ipv4Guid, GUID ipv6Guid, wfp::FilterBuilder &filterBuilder, IObjectInstaller& objectInstaller);
> ```

Could you change this to `IObjectInstaller &objectInstaller` for consistency?



_Code quote:_
```objc
IObjectInstaller&
```
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp` line 28 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk6ZYF5ChbSb-b9EFr:-NOk6ZYF5ChbSb-b9EFs:b-ta0y3n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp#L28)):*
> ```cpp
> }
> 
> bool PermitVpnTunnelService::AddEndpointFilter(const std::optional<PermitVpnTunnel::Endpoint> &endpoint, GUID ipv4Guid, GUID ipv6Guid, wfp::FilterBuilder &filterBuilder, IObjectInstaller& objectInstaller)
> ```

Everything I said about `PermitVpnTunnel` also applies here.

dlon (commented):
> Reviewed 4 of 17 files at r3, 1 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: all files reviewed, 20 unresolved discussions (waiting on @Jontified)

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: 15 of 21 files reviewed, 17 unresolved discussions (waiting on @dlon)

___
*[`talpid-core/src/firewall/windows.rs` line 222 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOPhnC-E-pMuamfbEDW:-NOUUMUZ57bIvI44MopL:btaqnuo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e07a2bc251ea6b25afdf5c15dcc7325bf0fd1ec3/talpid-core/src/firewall/windows.rs#L222)):*

Are we only manually dropping here to prevent accidents? Maybe we should not.
</blockquote></details>

Dropping here seems like a good idea to me because it causes compile errors if unsafe changes are made.
___
*[`talpid-wireguard/src/lib.rs` line 313 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOPitJ42XLAz8oxrn1c:-NOiQp4E4IumAhYWlx1u:bi8j6mp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d87248c1c052f6f6f338b60ff3ff284dd617bfd/talpid-wireguard/src/lib.rs#L313)):*

Why was this removed? Perhaps we purposely punched a hole before adding any routes here.
</blockquote></details>

Fixed after discussion.
___
*[`talpid-wireguard/src/lib.rs` line 323 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOQAlnS2g3Nu1ddVr7w:-NOiMfDgEIu-e25NrHN_:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/talpid-wireguard/src/lib.rs#L323)):*

This should probably be moved to a function/method.
</blockquote></details>

Done.
___
*[`talpid-wireguard/src/lib.rs` line 324 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOQ1XsM1Cd4j0EQCBSg:-NOiMg4t6z4m4jq4xdXi:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/talpid-wireguard/src/lib.rs#L324)):*

You could just generate a key here. It does not need to be an option.
</blockquote></details>

Done.
___
*[`talpid-wireguard/src/lib.rs` line 328 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOQBioDCs5L4ls3CJX5:-NOiSqnZ0rroZrXuBAxP:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/talpid-wireguard/src/lib.rs#L328)):*

I'm a bit skeptical of the panicking here. I think we should return an error, even though it's supposed to be unreachable.
</blockquote></details>

Done.
___
*[`talpid-wireguard/src/lib.rs` line 342 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOQ6w_m7OnDL6G600ek:-NOiT1eu-5njF5iLdHZT:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/talpid-wireguard/src/lib.rs#L342)):*

This adds `10.64.0.1`. So calling `patch_allowed_ips` is unnecessary, as is `entry_tun_config.peers.pop();`.
</blockquote></details>

Done.
___
*[`talpid-wireguard/src/lib.rs` line 375 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOQ90nWBufjeZo_5ehD:-NOspHFY2acVskL0rg3D:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/talpid-wireguard/src/lib.rs#L375)):*

All you need here is the original config, since it already swallows everything with `0.0.0.0/0`. Maybe it would make more sense to call `perform_psk_negotiation` for the exit first, so you'd only have to update the WireGuard config and firewall rules twice instead of three times?

I don't think we should use `patch_allowed_ips` at all here. It is intended for Android, and it makes no difference.
</blockquote></details>

Done.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.h` line 35 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk4gQKBUBfWJxAgMwL:-NOn9MIwCCeuDSGCCHc4:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.h#L35)):*

Could you change this to `IObjectInstaller &objectInstaller` for consistency?
</blockquote></details>

Done.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 27 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk5QYJ5G4gOIsddpWB:-NOn9tVt6lpWCSJyr6ue:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L27)):*

I think the `wfp::FilterBuilder` should be constructed in this function instead of passed in as an argument. Right now, it's possible to pass in an incorrectly initialized filter builder. If you were to construct it here, the function would be completely self-contained. It's more cohesive.
</blockquote></details>

Done.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 27 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk7P8mEt4Z7dB1nwUE:-NOnAC637ACzGBW4ozuy:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L27)):*

You could pass the GUIDs as `const GUID &...`.
</blockquote></details>

Done.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 27 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk7agMDWY9m3SQ9QK5:-NOnAEwn-bKAyPZOU8Q4:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L27)):*

`IObjectInstaller &objectInstaller`
</blockquote></details>

Done.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp` line 29 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk67g5CVIHU6YtdmfQ:-NOnAqteBFo90n3gtBr-:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnel.cpp#L29)):*

Just my opinion, but I think we should store these conditions in booleans for the purpose of documentation, like we did previously. E.g.,

```
bool shouldAddV4Filter = ...;
```

What the condition is for is currently very non-obvious.
</blockquote></details>

Done.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.h` line 8 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk4H1-3zpwuKl21C8D:-NOnAzPr1rf6Sc5YB4mv:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.h#L8)):*

You should probably import `optional`, and `vector` is no longer needed.
</blockquote></details>

Done.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.h` line 25 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk3bwpChQLskzKVNT2:-NOnB23sDzeA9sN_X5ok:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.h#L25)):*

Could you change this to `IObjectInstaller &objectInstaller` for consistency?
</blockquote></details>

Done.
___
*[`windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp` line 28 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk6ZYF5ChbSb-b9EFr:-NOnB2jw5wbu3gkDRRce:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/windows/winfw/src/winfw/rules/baseline/permitvpntunnelservice.cpp#L28)):*

Everything I said about `PermitVpnTunnel` also applies here.
</blockquote></details>

Done.

dlon (approved):
> Reviewed 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: all files reviewed, 5 unresolved discussions (waiting on @Jontified)

___
*[`talpid-wireguard/src/lib.rs` line 323 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOQAlnS2g3Nu1ddVr7w:-NOtHFwVBDWCzyv_v-Kn:btojbfd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/talpid-wireguard/src/lib.rs#L323)):*

Done.
</blockquote></details>

Great!
___
*[`talpid-wireguard/src/lib.rs` line 458 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOtH1fr2-IqzO7-HGUX:-NOtH1fs4eBrcg-vr_qG:b-7h9uhb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8089731dcb4e5a854f1988e14ac8b9f9a39c5ae9/talpid-wireguard/src/lib.rs#L458)):*
> ```XML
>         .await?;
> 
>         log::debug!("Successfully exchanged PSK with gateway peer");
> ```

Nit: Maybe "exit peer"?

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: all files reviewed, 4 unresolved discussions (waiting on @dlon)

___
*[`docs/architecture.md` line 197 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk8u6y59BK833Mzf8k:-NOtoVcR01fzFQ0S3GbH:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/docs/architecture.md#L197)):*

This should probably be changed to either `quantum-resistant tunnel` or `post-quantum tunnel`.
</blockquote></details>

Done.
___
*[`docs/architecture.md` line 201 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk96smEl6OtHpVgJUI:-NOtoWq0Bq8V6PxyqKNp:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/docs/architecture.md#L201)):*

"PQ safe" or "quantum resistant" (in my opinion).
</blockquote></details>

Done.
___
*[`docs/architecture.md` line 208 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOk9KLp-dvdpfhsuX_n:-NOtoXad2TyAvx712A6r:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7d1eac00dd638131495b9f086af9cc6138799e8c/docs/architecture.md#L208)):*

See comment above.
</blockquote></details>

Done.
___
*[`talpid-wireguard/src/lib.rs` line 458 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOtH1fr2-IqzO7-HGUX:-NOtLJ136sEFS-fSnE15:bxmpjkf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8089731dcb4e5a854f1988e14ac8b9f9a39c5ae9/talpid-wireguard/src/lib.rs#L458)):*

Nit: Maybe "exit peer"?
</blockquote></details>

The reason I said gateway peer is because it might be the entry and exit if there is no multihop, but if we usually call it the exit then I'll change it?

dlon (approved):
> Reviewed 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: all files reviewed, 2 unresolved discussions (waiting on @Jontified)

___
*[`docs/architecture.md` line 203 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOxehQKBm67EqThwEa5:-NOxehQKBm67EqThwEa6:bl3pe86) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2839828fb3e85b0ab3df4983d52431c8cff0d617/docs/architecture.md#L203)):*
> ```Markdown
> and the PSK, this new tunnel is quantum-resistant.
> 
> The specifics of how the PSK exchange is done can be found (here)[https://github.com/mullvad/mullvadvpn-app/blob/master/talpid-tunnel-config-client/proto/tunnel_config.proto]
> ```

The link is broken. Also, I think you could use a relative path: `../talpid-tunnel-config/proto ...`.
___
*[`talpid-wireguard/src/lib.rs` line 458 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOtH1fr2-IqzO7-HGUX:-NOxg3VH9Sy7bS-vyvby:b-x0katf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8089731dcb4e5a854f1988e14ac8b9f9a39c5ae9/talpid-wireguard/src/lib.rs#L458)):*

The reason I said gateway peer is because it might be the entry and exit if there is no multihop, but if we usually call it the exit then I'll change it?
</blockquote></details>

I would argue that "exit" is always correct. But if you disagree, maybe you could log different things after checking `config.peers.len()`.

dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: all files reviewed, 2 unresolved discussions (waiting on @Jontified)

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`talpid-wireguard/src/lib.rs` line 458 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOtH1fr2-IqzO7-HGUX:-NOyIHBW6xj1bTO5MTEz:be4mlrh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8089731dcb4e5a854f1988e14ac8b9f9a39c5ae9/talpid-wireguard/src/lib.rs#L458)):*

I would argue that "exit" is always correct. But if you disagree, maybe you could log different things after checking `config.peers.len()`.
</blockquote></details>

Exit is fine and correct. The last hop is always the exit no matter how many hops come before it.
___
*[`talpid-wireguard/src/lib.rs` line 419 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOyICD1AOJhD8jdhLEw:-NOyICD1AOJhD8jdhLEx:b-nqlids) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2839828fb3e85b0ab3df4983d52431c8cff0d617/talpid-wireguard/src/lib.rs#L419)):*
> ```XML
>             + Sync
>             + Clone
>             + 'static,
> ```

Nit, but this might be a case where putting the bounds in a `where` clause significantly clears up the function signature :see_no_evil:

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[`docs/architecture.md` line 198 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOyMW-BEgheUSghhrww:-NOyr-wN8ZpPfSHQGeih:b-7hzf34) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2839828fb3e85b0ab3df4983d52431c8cff0d617/docs/architecture.md#L198)):*

I'd like to have the text point out that the "cryptographic key exchange" is actually quantum-safe. And I think there would be value in including the well known abbreviation "KEM". The readme for `liboqs` says "quantum-safe key encapsulation mechanism (KEM)" and I think we should include that formulation here. How about phrasing this paragraph something like:

> A quantum-resistant tunnel is created by applying a pre-shared key (PSK) derived using a quantum-safe key encapsulation mechanism (KEM) to the encryption of the tunnel. This is done by establishing a regular WireGuard tunnel to the relay, derive the PSK in the tunnel and have it associated with an ephemeral WireGuard pubkey picked by the client, and then re-establish a new tunnel to the same relay but with the ephemeral WireGuard key and the PSK. Details about the protocol and the cryptographic strength are discussed in more details in the [protocol definition file]
</blockquote></details>

What do you think about the new paragraph?
___
*[`docs/architecture.md` line 203 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOxehQKBm67EqThwEa5:-NOyr2oYDn9OKcy1iL42:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2839828fb3e85b0ab3df4983d52431c8cff0d617/docs/architecture.md#L203)):*

The link is broken. Also, I think you could use a relative path: `../talpid-tunnel-config/proto ...`.
</blockquote></details>

Done.
___
*[`docs/architecture.md` line 209 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOyJlWv-jXbKJbqebx-:-NOysL2d4Ab5h27-DlcW:b-50mbmy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2839828fb3e85b0ab3df4983d52431c8cff0d617/docs/architecture.md#L209)):*

"generates a temporary WireGuard key and establishes a normal tunnel"

This can be confused as the temporary WireGuard key being used, which it is not. It is also not clear that this is a multihop tunnel. "normal" is a bit vague as it only refers to the non-pq aspect of it.

Can this paragraph(s) be written more concise as something like:

> To create a multihop tunnel where both hops are quantum resistant the client must negotiate a a unique PSK with both the entry and the exit relay separately. It must use the same ephemeral WireGuard key on both relays since the end result (just as with regular multihop tunnels) is two peers on a single WireGuard interface, which can only have a single key for the local peer.
>
> The PSKs are established by first creating a regular multihop tunnel to the exit via the entry relay and negotiate a PSK with the exit. Then establish a regular tunnel to just the entry and negotiate a PSK with it. Lastly the client can set up a multihop tunnel using the new ephemeral WireGuard key and the two PSKs via the entry to the exit.
</blockquote></details>

Yeah, that works really well.

dlon (approved):
> Reviewed 6 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: all files reviewed, 2 unresolved discussions (waiting on @faern)

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

___
*[`docs/architecture.md` line 198 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOyMW-BEgheUSghhrww:-NPHHNo34QXt5rQXHtVB:b-1gz5k1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2839828fb3e85b0ab3df4983d52431c8cff0d617/docs/architecture.md#L198)):*

What do you think about the new paragraph?
</blockquote></details>

I'm not a fan of using "saved on the relay and the client". Because we are very adamant about not persisting it or even letting it spread into too many places in RAM. The PSK is stored in memory only and treated as ephemeral on the client. On the server it's kept until reboot/restart of wg-manager but still only in memory.

Nitpicky, but using [this] as the link text is not my favorite option. It's used too often IMHO. I way prefer when the link text actually says what it links to. Which is why my example used [protocol definition file]

dlon (approved):
> Reviewed 1 of 2 files at r9, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: all files reviewed, 1 unresolved discussion (waiting on @Jontified)

Jontified (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

___
*[`docs/architecture.md` line 198 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353#-NOyMW-BEgheUSghhrww:-NPJrsc3AyQOBIWn6W9h:b-modxbt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2839828fb3e85b0ab3df4983d52431c8cff0d617/docs/architecture.md#L198)):*

I'm not a fan of using "saved on the relay and the client". Because we are very adamant about not persisting it or even letting it spread into too many places in RAM. The PSK is stored in memory only and treated as ephemeral on the client. On the server it's kept until reboot/restart of wg-manager but still only in memory.

Nitpicky, but using [this] as the link text is not my favorite option. It's used too often IMHO. I way prefer when the link text actually says what it links to. Which is why my example used [protocol definition file]
</blockquote></details>

I've updated to make it clear that it's stored only in memory on both the client and the server.
I've also updated the protocol definition file.

dlon (approved):
> Reviewed 1 of 3 files at r13, 1 of 1 files at r14, 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4353)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4356, opened 2023-02-13
Rename settings error variants

dlon:
`NoMatchingVersion` was used for any invalid settings format as long as it could be deserialized to JSON. So it has been renamed to `InvalidSettingsContent`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356)

### DIFF
--- a/mullvad-daemon/src/migrations/mod.rs
+++ b/mullvad-daemon/src/migrations/mod.rs
@@ -59,11 +59,11 @@ pub enum Error {
     #[error(display = "Failed to read the settings")]
     Read(#[error(source)] io::Error),
 
-    #[error(display = "Malformed settings")]
-    Parse(#[error(source)] serde_json::Error),
+    #[error(display = "Failed to deserialize settings")]
+    Deserialize(#[error(source)] serde_json::Error),
 
-    #[error(display = "Unable to read any version of the settings")]
-    NoMatchingVersion,
+    #[error(display = "Unexpected settings format")]
+    InvalidSettingsContent,
 
     #[error(display = "Unable to serialize settings to JSON")]
     Serialize(#[error(source)] serde_json::Error),
@@ -132,10 +132,10 @@ pub(crate) async fn migrate_all(
     let settings_bytes = fs::read(&path).await.map_err(Error::Read)?;
 
     let mut settings: serde_json::Value =
-        serde_json::from_reader(&settings_bytes[..]).map_err(Error::Parse)?;
+        serde_json::from_reader(&settings_bytes[..]).map_err(Error::Deserialize)?;
 
     if !settings.is_object() {
-        return Err(Error::NoMatchingVersion);
+        return Err(Error::InvalidSettingsContent);
     }
 
     let old_settings = settings.clone();
--- a/mullvad-daemon/src/migrations/v2.rs
+++ b/mullvad-daemon/src/migrations/v2.rs
@@ -27,7 +27,7 @@ pub fn migrate(settings: &mut serde_json::Value) -> Result<()> {
     {
         settings
             .as_object_mut()
-            .ok_or(Error::NoMatchingVersion)?
+            .ok_or(Error::InvalidSettingsContent)?
             .remove("show_beta_releases");
     }
 
@@ -55,7 +55,7 @@ pub fn migrate(settings: &mut serde_json::Value) -> Result<()> {
         settings["tunnel_options"]["wireguard"]["rotation_interval"] = serde_json::json!(new_ivl);
         settings["tunnel_options"]["wireguard"]
             .as_object_mut()
-            .ok_or(Error::NoMatchingVersion)?
+            .ok_or(Error::InvalidSettingsContent)?
             .remove("automatic_rotation");
     }
 
--- a/mullvad-daemon/src/migrations/v3.rs
+++ b/mullvad-daemon/src/migrations/v3.rs
@@ -61,7 +61,7 @@ pub fn migrate(settings: &mut serde_json::Value) -> Result<()> {
                 DnsState::Default
             };
             let addresses = if let Some(addrs) = options.get("addresses") {
-                serde_json::from_value(addrs.clone()).map_err(Error::Parse)?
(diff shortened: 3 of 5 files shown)


faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-daemon/src/migrations/v3.rs` line 64 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356#-NO8yNwk6K06FGY4Y0gp:-NO8yNwk6K06FGY4Y0gq:b-ebep6f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9d8bb299217f5ae2d8922edfb0db3f1d6f010cc0/mullvad-daemon/src/migrations/v3.rs#L64)):*
> ```XML
>             };
>             let addresses = if let Some(addrs) = options.get("addresses") {
>                 serde_json::from_value(addrs.clone()).map_err(Error::Deserialize)?
> ```

I think this should be `InvalidSettingsContent`. The input here is already a deserialized JSON `Value`. It just tries to map that value to a vector of IPs. It will fail if the value was not a list of IPs, which means the content was of an unexpected format (but not invalid JSON since it has already been parsed into `Value`.)

The same applies to all other uses of `Error::Deserialize` below. I think the only error that is actually a deserialization error is the one at `serde_json::from_reader`.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`mullvad-daemon/src/migrations/v3.rs` line 64 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356#-NO8yNwk6K06FGY4Y0gp:-NO96uLtBoPGMcNS5_-a:b-mlwb9j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9d8bb299217f5ae2d8922edfb0db3f1d6f010cc0/mullvad-daemon/src/migrations/v3.rs#L64)):*

I think this should be `InvalidSettingsContent`. The input here is already a deserialized JSON `Value`. It just tries to map that value to a vector of IPs. It will fail if the value was not a list of IPs, which means the content was of an unexpected format (but not invalid JSON since it has already been parsed into `Value`.)

The same applies to all other uses of `Error::Deserialize` below. I think the only error that is actually a deserialization error is the one at `serde_json::from_reader`.
</blockquote></details>

Done. Maybe we should still include the source?

pinkisemils (approved):
> Reviewed 2 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-daemon/src/migrations/v3.rs` line 64 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356#-NO8yNwk6K06FGY4Y0gp:-NO9E4Xp1duHWeOjFkfZ:bsd3ela) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9d8bb299217f5ae2d8922edfb0db3f1d6f010cc0/mullvad-daemon/src/migrations/v3.rs#L64)):*

Done. Maybe we should still include the source?
</blockquote></details>

Yeah, if we have a good source error I think it should be included. I'm not sure what these errors have in their `Display` impl, but probably does not hurt :+1:

However, I see that we do use `InvalidSettingsContent` in places where there is no source error. So if we go that route it would need to be two different error types.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`mullvad-daemon/src/migrations/v3.rs` line 64 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356#-NO8yNwk6K06FGY4Y0gp:-NO9P24f7n3uJAf_dwcx:b6rent8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9d8bb299217f5ae2d8922edfb0db3f1d6f010cc0/mullvad-daemon/src/migrations/v3.rs#L64)):*

Yeah, if we have a good source error I think it should be included. I'm not sure what these errors have in their `Display` impl, but probably does not hurt :+1:

However, I see that we do use `InvalidSettingsContent` in places where there is no source error. So if we go that route it would need to be two different error types.
</blockquote></details>

The `Display` impl outputs "unexpected ... on line X, column Y". This might confuse more than it enlightens, as it wouldn't be referring to the line and column in settings.json, just some json object. So we probably don't need to include it.

faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356#-:-NO9rAeI-WKtu8Cyb1Oe:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`mullvad-daemon/src/migrations/v3.rs` line 64 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4356#-NO8yNwk6K06FGY4Y0gp:-NO9r7cO7ZyHUPXOHRcg:b-p1i2jg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9d8bb299217f5ae2d8922edfb0db3f1d6f010cc0/mullvad-daemon/src/migrations/v3.rs#L64)):*

The `Display` impl outputs "unexpected ... on line X, column Y". This might confuse more than it enlightens, as it wouldn't be referring to the line and column in settings.json, just some json object. So we probably don't need to include it.
</blockquote></details>

Ok! Let's ignore it then!

PR #4360, opened 2023-02-14
Don't panic when the default directories cannot be obtained

dlon:
Previously, `mullvad-daemon` would panic if, for example, the path to the settings directory could not be obtained. This was problematic in the install script, since we one user reported that it would fail to install the system service, even though that directory was irrelevant.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4360)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ Line wrap the file at 100 chars.                                              Th
 #### Android
 - Fix adaptive app icon which previously had a displaced nose and some other oddities.
 
+#### Windows
+- Don't fail to show the mullvad-daemon help text if some of the default paths cannot be obtained.
+
 
 ## [2023.1-beta2] - 2023-02-06
 ### Fixed
--- a/mullvad-daemon/src/cli.rs
+++ b/mullvad-daemon/src/cli.rs
@@ -68,9 +68,9 @@ lazy_static::lazy_static! {
 
 ",
         mullvad_paths::get_default_resource_dir().display(),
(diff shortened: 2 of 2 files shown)


faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4360#-:-NOEotErCOlDS1Cimz2u:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-daemon/src/cli.rs` line 73 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4360#-NOEoeTY1rnE5iZMtRJr:-NOEoeTY1rnE5iZMtRJs:b-ki79hk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a97e619331198edfd94295d58c9d7bb293ac985b/mullvad-daemon/src/cli.rs#L73)):*
> ```XML
>         mullvad_paths::get_default_settings_dir().map(|dir| dir.display().to_string()).unwrap_or_else(|_| "N/A".to_string()),
>         mullvad_paths::get_default_cache_dir().map(|dir| dir.display().to_string()).unwrap_or_else(|_| "N/A".to_string()),
>         mullvad_paths::get_default_log_dir().map(|dir| dir.display().to_string()).unwrap_or_else(|_| "N/A".to_string()),
> ```

Nit and not blocking. But IMHO `.to_owned()` is more idiomatic. Since the source type is also a "string" of sort. `string.to_string()` is confusing. "To owned" more indicates the transformation that takes place. But this is nothing important really.

PR #4363, opened 2023-02-15
Replace OperationCompletion with plain Result<T, F>

pronebird:
Seemingly large PR is basically a lot of renaming. This PR replaces `OperationCompletion` (which is in essence a `Result<T, F>` + `.cancelled` variant) with `Result<T, Error>` where in case of cancellation operations return `OperationError.cancelled` error instead. 

`ResultOperation<Success>` no longer accepts failure type either and with that boundary gone it makes passing errors in operation subclasses a little more natural as the base class (`AsyncOperation`) already supports errors. 

This change also makes it possible to extend internal errors, for example `TransformOperation` expects input and if it's not provided it can return `OperationError.unsatisfiedRequirement` error instead, previously it used to return `OperationCompletion.cancelled` which is not exactly right.

Last but not least, it's kind of in line with how the language is designed where in structured concurrency `CancellationError` is returned in case of cancellation and try-catch expressions return type-erased `Error`, unless you cast to specific error type.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4363)


pinkisemils (changes requested):
> Reviewed 31 of 49 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4363)* status: 31 of 49 files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`ios/MullvadVPN/TunnelManager/StartTunnelOperation.swift` line 57 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4363#-NOQ3mNdC1WsLsbJpCiR:-NOQ3mNe1SdP20cbwlGM:bn2qqws) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6ee904bb0cfc02618e6e6eeccf4bc06ad4c3a2ba/ios/MullvadVPN/TunnelManager/StartTunnelOperation.swift#L57)):*
> ```Swift
> 
>                 makeTunnelProviderAndStartTunnel(selectorResult: selectorResult) { error in
>                     self.finish(result: error.map { .failure($0) } ?? .success(()))
> ```

I do not follow this bit
```
error.map { .failure($0) } ?? .success(())
```
The error is mapped via a closure, that's fair. But then it's coalesced - is the error not nil ,and therefore it cannot be mapped? Or is calling `map` on a nil error fine?

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4363)* status: 48 of 49 files reviewed, 1 unresolved discussion (waiting on @pronebird)

pinkisemils (approved):
> Reviewed 1 of 49 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4363)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4363#-:-NPMOWtg8ibgQ_07lzEA:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4363)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`ios/MullvadVPN/TunnelManager/StartTunnelOperation.swift` line 57 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4363#-NOQ3mNdC1WsLsbJpCiR:-NPN5MvZ2XxWH8qbtNEN:bud249d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6ee904bb0cfc02618e6e6eeccf4bc06ad4c3a2ba/ios/MullvadVPN/TunnelManager/StartTunnelOperation.swift#L57)):*

I do not follow this bit
```
error.map { .failure($0) } ?? .success(())
```
The error is mapped via a closure, that's fair. But then it's coalesced - is the error not nil ,and therefore it cannot be mapped? Or is calling `map` on a nil error fine?
</blockquote></details>

The `Error?` is mapped to `Result<Void, Error>` here. Since `error` is `Optional<Error>` (in short `Error?`), you can map `nil` which is basically yields `nil`, so the block does not execute because `Optional` variant is `.none`. Null coalescing operator (`??`) should then either return the mapped value or fallback to `Result<Void, Error>success(())`.

PR #4378, opened 2023-02-21
Fix timeout duration not fitting into 32-bit signed integer

raksooo:
The new timeout for showing account expiry notifications immediately when there is three days left had a too large duration. The duration needs to fit into a signed 32-bit integer which means that the maximum is around 24 days. I've changed the duration to max out at 24 days.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4378)


hankolsen (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4378)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4382, opened 2023-02-21
GUI setting for toggling quantum resistant connection

raksooo:
This PR adds the quantum resistant setting to the desktop GUI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4382)

### DIFF
--- a/gui/locales/messages.pot
+++ b/gui/locales/messages.pot
@@ -108,6 +108,7 @@ msgstr ""
 msgid "Connecting"
 msgstr ""
 
+#. Creating a secure connection that isn't breakable by quantum computers.
 msgid "CREATING QUANTUM SECURE CONNECTION"
 msgstr ""
 
@@ -187,6 +188,7 @@ msgstr ""
 msgid "Open URL"
 msgstr ""
 
+#. The connection is secure and isn't breakable by quantum computers.
 msgid "QUANTUM SECURE CONNECTION"
 msgstr ""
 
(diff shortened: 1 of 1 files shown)


gui/src/main/daemon-rpc.ts
```diff
@@ -437,6 +437,25 @@ export class DaemonRpc {
     await this.callNumber(this.client.setWireguardMtu, mtu);
   }
 
+  public async setWireguardQuantumResistant(quantumResistant?: boolean): Promise<void> {
+    const quantumResistantState = new grpcTypes.QuantumResistantState();
+    switch (quantumResistant) {
```
MrChocolatine, line 442: What about this shorter approach?

```suggestion
    const state = quantumResistant === true ? 'ON'
      : quantumResistant === false ? 'OFF'
        : 'AUTO';

    quantumResistantState.setState(grpcTypes.QuantumResistantState.State[state]);
```

hankolsen (approved):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4382)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

gui/src/main/daemon-rpc.ts
raksooo: I agree that the switch statement is a bit long and clunky but I'm not a huge fan of nested ternary expressions either. We do a lot of convertions to/from enums in this file and I opted for the switch statement to be more consistent with other conversions of bigger enums. I'll keep the switch statement for now.

I will hopefully spend some time refactoring and improving this file as a whole in a not too distant future and improving these convertions will be on my list of things to improve.

raksooo:
I forgot to push the last rebase and GitHub therefore considers it's closed, but it's actually merged.

PR #4386, opened 2023-02-22
Impersonate any process running as the system user instead of LSASS

dlon:
Fixes the issue of the daemon not starting when LSA protection is enabled. Changes:
* Impersonate any service instead of LSASS.
* Skip impersonation if it's not needed at all (i.e., the `mullvadvpn` service).
* Fall back on `%system32%\config\systemprofile\AppData\Local` if all else fails.


pinkisemils (changes requested):
> Reviewed 8 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: 8 of 12 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`windows/nsis-plugins/src/cleanup/cleaningops.cpp` line 79 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOuvbtDDOdmVNXxPuNv:-NOuvbtDDOdmVNXxPuNw:b-3xynlp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7479416268f10bddb8a3c31a64d4bf079ce9058a/windows/nsis-plugins/src/cleanup/cleaningops.cpp#L79)):*
> ```cpp
> 	if (Status::InsufficientBufferSize == result)
> 	{
> 		buffer.resize(bufferSize);
> ```

Should `bufferSize` not be doubled?

pinkisemils (changes requested):
> Reviewed 1 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: 9 of 12 files reviewed, 4 unresolved discussions (waiting on @dlon)

___
*[`mullvad-paths/src/windows.rs` line 64 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOuvvm55FUDAPZjah4m:-NOuvvm55FUDAPZjah4n:b3rqf1l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7479416268f10bddb8a3c31a64d4bf079ce9058a/mullvad-paths/src/windows.rs#L64)):*
> ```XML
> 
>     if current_user_is_system.map_err(|error| {
>         log::error!("is_local_system_user_token failed: {error}");
> ```

Instead of printing the error here, I think this if statement and the last 4 lines of the function could be merged into a match statement like so...
```rust
match is_local_system_user_token(current_token.0) {
    Ok(true) => get_known_folder_path(&FOLDERID_LocalAppData, KF_FLAG_DEFAULT, 0),
    Ok(false) =>     Err(io::Error::new(
        io::ErrorKind::Other,
        "current user is not SYSTEM",
    )),
    Err(err) =>  {
       log::error!(...);
       Err(err)
    }

}
```
___
*[`mullvad-paths/src/windows.rs` line 90 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOux0h-Brab92loqNA3:-NOux0h-Brab92loqNA4:b-3xrhij) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7479416268f10bddb8a3c31a64d4bf079ce9058a/mullvad-paths/src/windows.rs#L90)):*
> ```XML
> 
>         let status = unsafe {
>             OpenProcessToken(
> ```

Any reason not to wrap the token in a `Handle`? That way you can avoid having to worry about closing it, and have the logic be a bit clearer further down. 



_Code quote:_
```XML
drop(
```
___
*[`mullvad-paths/src/windows.rs` line 269 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOuxU_lBi-v1ifzyoEM:-NOuxU_lBi-v1ifzyoEN:bo8z6va) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7479416268f10bddb8a3c31a64d4bf079ce9058a/mullvad-paths/src/windows.rs#L269)):*
> ```XML
>         io::ErrorKind::NotFound,
>         "Could not find matching process",
>     ))
> ```

This could be shortened via `result.ok_or(io::Error::new(...))`

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: 7 of 13 files reviewed, 4 unresolved discussions (waiting on @pinkisemils)

___
*[`mullvad-nsis/src/lib.rs` line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOuyqc1-2ftuc2QtF5j:-NOxh-52DfCj7NVux1lv:b-bdzyaf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7479416268f10bddb8a3c31a64d4bf079ce9058a/mullvad-nsis/src/lib.rs#L53)):*

This statement should come before `return Status::InsufficientBufferSize;`
</blockquote></details>

Done. I'm glad you spotted that.
___
*[`mullvad-paths/src/windows.rs` line 64 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOuvvm55FUDAPZjah4m:-NOxgceO2vGmSGoKZVWx:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7479416268f10bddb8a3c31a64d4bf079ce9058a/mullvad-paths/src/windows.rs#L64)):*

Instead of printing the error here, I think this if statement and the last 4 lines of the function could be merged into a match statement like so...
```rust
match is_local_system_user_token(current_token.0) {
    Ok(true) => get_known_folder_path(&FOLDERID_LocalAppData, KF_FLAG_DEFAULT, 0),
    Ok(false) =>     Err(io::Error::new(
        io::ErrorKind::Other,
        "current user is not SYSTEM",
    )),
    Err(err) =>  {
       log::error!(...);
       Err(err)
    }

}
```
</blockquote></details>

Done.
___
*[`mullvad-paths/src/windows.rs` line 90 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOux0h-Brab92loqNA3:-NOxgndb3mun-tuYe4Uw:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7479416268f10bddb8a3c31a64d4bf079ce9058a/mullvad-paths/src/windows.rs#L90)):*

Any reason not to wrap the token in a `Handle`? That way you can avoid having to worry about closing it, and have the logic be a bit clearer further down. 
</blockquote></details>

Done.
___
*[`mullvad-paths/src/windows.rs` line 269 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOuxU_lBi-v1ifzyoEM:-NOxgoaPBFGnYxVX8KPy:bpgahgw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7479416268f10bddb8a3c31a64d4bf079ce9058a/mullvad-paths/src/windows.rs#L269)):*

This could be shortened via `result.ok_or(io::Error::new(...))` 
</blockquote></details>

How exactly? We don't have a result here. And storing the result and breaking out of the loop seems more complicated.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: 7 of 13 files reviewed, 5 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[`mullvad-paths/src/windows.rs` line 45 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOyQxuI8jFMQcUA5qdI:-NOyQxuI8jFMQcUA5qdJ:b-uqiuo5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0428112d84f33264de9ff57aedef32faf1459ae7/mullvad-paths/src/windows.rs#L45)):*
> ```XML
>     let result = get_appdata_as_system_user()
>         .or_else(|error| {
>             log::error!("get_appdata_as_system_user failed: {error}");
> ```

Will this not log "get_appdata_as_system_user failed: current user is not SYSTEM" every time this is called as a non-SYSTEM user? Since that's not an error path but rather a well known case where we will fall back, I don't think we should error-log this or not log it at all.

I think that either the function `get_appdata_as_system_user` should return `Result<Option<PathBuf>>` where `Ok(None)` means "nothing went wrong, but you are not the system user" OR, maybe even better, have the `is_local_system_user_token` call be done outside, and the responsibility of `get_appdata_as_system_user` is just to actually get the path and not check the permission.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: 7 of 13 files reviewed, 6 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[`mullvad-paths/src/windows.rs` line 52 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOyS-IL9ntOraRwEdF8:-NOyS-IM5UMtjIBG1w2v:b-vtfxva) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0428112d84f33264de9ff57aedef32faf1459ae7/mullvad-paths/src/windows.rs#L52)):*
> ```XML
>             infer_appdata_from_system()
>         });
>     if unsafe { RevertToSelf() } == 0 {
> ```

The `RevertToSelf` code is not paired with the code impersonating someone else. So IMO it becomes hard to see where the impersonation happens and validate that `RevertToSelf` is only called in relevant scenarios.

I think a better structure would be:

```rust
{
    impersonate();
    do_thing_as_other_user();
    RevertToSelf();
}
```
To have it be symmetrical and easy to match. It could be in a helper function that takes a closure and runs the closure as the impersonated user, reverts to self and returns the result of the closure.

pinkisemils (changes requested):
> Reviewed 1 of 12 files at r1, 2 of 4 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon)

___
*[`.github/workflows/daemon.yml` line 146 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOyQCsG5ts_qA7NaK4o:-NOyQCsHDpbwZRY9z5sx:bta5rup) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0428112d84f33264de9ff57aedef32faf1459ae7/.github/workflows/daemon.yml#L146)):*
> ```YAML
>         with:
>           toolchain: stable
>           target: i686-pc-windows-msvc
> ```

Will this still build x64 binaries of  the daemon?
___
*[`mullvad-paths/src/windows.rs` line 269 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOuxU_lBi-v1ifzyoEM:-NOxiBQ543kDdsGzDoJp:b-30uuzw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7479416268f10bddb8a3c31a64d4bf079ce9058a/mullvad-paths/src/windows.rs#L269)):*

How exactly? We don't have a result here. And storing the result and breaking out of the loop seems more complicated.
</blockquote></details>

```rust
let result = pid_bufffer.find_map(|process| -> Result<Option<Process>>{ ..}); 
```
Or something like that, I presume? But that might be too nitpicky here.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: all files reviewed, 5 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[`.github/workflows/daemon.yml` line 146 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOyQCsG5ts_qA7NaK4o:-NOy_Wy12C0ZoFRlZRde:b-gmk7wv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0428112d84f33264de9ff57aedef32faf1459ae7/.github/workflows/daemon.yml#L146)):*

Will this still build x64 binaries of  the daemon?
</blockquote></details>

Yes. This is added as an additional target.
___
*[`mullvad-paths/src/windows.rs` line 45 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOyQxuI8jFMQcUA5qdI:-NOy_5dS9d4Y-atU-Y0X:b-pm3ty) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0428112d84f33264de9ff57aedef32faf1459ae7/mullvad-paths/src/windows.rs#L45)):*

Will this not log "get_appdata_as_system_user failed: current user is not SYSTEM" every time this is called as a non-SYSTEM user? Since that's not an error path but rather a well known case where we will fall back, I don't think we should error-log this or not log it at all.

I think that either the function `get_appdata_as_system_user` should return `Result<Option<PathBuf>>` where `Ok(None)` means "nothing went wrong, but you are not the system user" OR, maybe even better, have the `is_local_system_user_token` call be done outside, and the responsibility of `get_appdata_as_system_user` is just to actually get the path and not check the permission.
</blockquote></details>

I'm tempted to log nothing here, as we already have logging in `get_appdata_as_system_user`. Would that work as well?
___
*[`mullvad-paths/src/windows.rs` line 52 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOyS-IL9ntOraRwEdF8:-NOyT23_AsYA_FZxbSKW:b-ozqo5q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0428112d84f33264de9ff57aedef32faf1459ae7/mullvad-paths/src/windows.rs#L52)):*

The `RevertToSelf` code is not paired with the code impersonating someone else. So IMO it becomes hard to see where the impersonation happens and validate that `RevertToSelf` is only called in relevant scenarios.

I think a better structure would be:

```rust
{
    impersonate();
    do_thing_as_other_user();
    RevertToSelf();
}
```
To have it be symmetrical and easy to match. It could be in a helper function that takes a closure and runs the closure as the impersonated user, reverts to self and returns the result of the closure.
</blockquote></details>

Agreed. If it's possible to revert the impersonation immediately after `OpenThreadToken` succeeds, it would be a lot more readable.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[`mullvad-paths/src/windows.rs` line 45 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOyQxuI8jFMQcUA5qdI:-NOygh4E-paXeEIj5sV8:b-e5dsox) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0428112d84f33264de9ff57aedef32faf1459ae7/mullvad-paths/src/windows.rs#L45)):*

I'm tempted to log nothing here, as we already have logging in `get_appdata_as_system_user`. Would that work as well?
</blockquote></details>

The logging is one thing. Yes, I think it's an improvement to not log. But I think that's only a minor part of my point here. The main point is that I think the different functions have hard to understand responsibilities.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[`mullvad-paths/src/windows.rs` line 45 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOyQxuI8jFMQcUA5qdI:-NOyitb27r_0KM8TigFS:b8yjana) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0428112d84f33264de9ff57aedef32faf1459ae7/mullvad-paths/src/windows.rs#L45)):*

The logging is one thing. Yes, I think it's an improvement to not log. But I think that's only a minor part of my point here. The main point is that I think the different functions have hard to understand responsibilities.
</blockquote></details>

The current impl of `get_appdata_as_system_user` looks more like it should be named `get_appdata_if_system_user`. Because it does not become the system user (the "as" part), but it gets the app data *if* it's the system user. And `get_appdata_as_admin` looks like what's it's actually doing is `get_appdata_as_system_user` since what it does is elevate to system user and then gets the app data path?

Can't this be implemented more like

```rust
let mut is_impersonating = false;
if !is_system_user() {
    is_impersonating = true;
    impersonate();
}
let appdata_path = get_known_folder_path(...).or_else(|error| {
    log::error!("Failed to get AppData path: {error}");
    infer_appdata_from_system()
});
if is_impersotating {
    RevertToSelf();
}
appdata_path
```

I'm probably simplifying too much here since this seems extremely complicated. But I'm convinced there is some code structure that would improve the readability somewhat.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: all files reviewed, 6 unresolved discussions (waiting on @dlon and @pinkisemils)

___
*[`mullvad-paths/src/windows.rs` line 129 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOykY9WBIkzBnOsZSlw:-NOykY9WBIkzBnOsZSlx:b46kx6b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0428112d84f33264de9ff57aedef32faf1459ae7/mullvad-paths/src/windows.rs#L129)):*
> ```XML
> 
> /// If all else fails, infer the AppData path from the system directory.
> fn infer_appdata_from_system() -> io::Result<PathBuf> {
> ```

"infer from system" reads to me like we make the system compute the app data path for us. But in reality we infer it from the **system directory**, which is a very different thing. I think the name of this function can be made clearer.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: all files reviewed, 6 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[`mullvad-paths/src/windows.rs` line 45 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOyQxuI8jFMQcUA5qdI:-NOyiu133r70qUimFGWp:bxw2t5m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0428112d84f33264de9ff57aedef32faf1459ae7/mullvad-paths/src/windows.rs#L45)):*

The current impl of `get_appdata_as_system_user` looks more like it should be named `get_appdata_if_system_user`. Because it does not become the system user (the "as" part), but it gets the app data *if* it's the system user. And `get_appdata_as_admin` looks like what's it's actually doing is `get_appdata_as_system_user` since what it does is elevate to system user and then gets the app data path?

Can't this be implemented more like

```rust
let mut is_impersonating = false;
if !is_system_user() {
    is_impersonating = true;
    impersonate();
}
let appdata_path = get_known_folder_path(...).or_else(|error| {
    log::error!("Failed to get AppData path: {error}");
    infer_appdata_from_system()
});
if is_impersotating {
    RevertToSelf();
}
appdata_path
```

I'm probably simplifying too much here since this seems extremely complicated. But I'm convinced there is some code structure that would improve the readability somewhat.
</blockquote></details>

I agree with the "if" part. "As" is ambiguous/misleading here. `get_appdata_as_admin` should really be called `get_appdata_if_admin` as well.

It can't be implemented like that. `RevertToSelf` is (confusingly) only paired with `ImpersonateSelf`. `SHGetKnownFolderPath` also impersonates the access token we pass in, but I doubt that we are supposed to revert this ourselves.

What we probably could do is obtain the token first (NULL if the current user is already system) and then call `get_known_folder_path`:

```rust
// hopefully we can just call `RevertSelf` in this function before returning
let thread_token = get_current_thread_token();

// token == null if current user is system, otherwise it enumerates the processes, etc.
let token = get_system_user_token(thread_token);

get_known_folder_path(...).or_else(|error| {
    log::error!("Failed to get AppData path: {error}");
    infer_appdata_from_system()
});
```

Do you find this any less confusing?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: 12 of 13 files reviewed, 5 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[`mullvad-nsis/src/lib.rs` line 3 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NP1vrkN9GSSOE6xfNms:-NP1vrkN9GSSOE6xfNmt:b-ikh83m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e2568b56e76a43dee9f69a82e7f98dbaeab788d8/mullvad-nsis/src/lib.rs#L3)):*
> ```XML
> #![cfg(all(target_arch = "x86", target_os = "windows"))]
> 
> // TODO: add logging
> ```

We probably don't really need logging. But if we do, the way to go might be to add a `mullvad_get_last_error` function and store it using thread-local storage that we update using `set_last_error(new_error: Box<&[u16]>)`, or something like that.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: all files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[`mullvad-paths/src/windows.rs` line 128 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NOzG08e24bLv3sRJzQJ:-NPHGpOA8JEcP1y7qmhg:b4c9lyr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6d2a2739905c4a645309165840a75484023dc6ff/mullvad-paths/src/windows.rs#L128)):*

I changed this to a panic. From docs:

> If RevertToSelf fails, your application continues to run in the context of the client, which is not appropriate. You should shut down the process if RevertToSelf fails.
</blockquote></details>

:thumbsup:

pinkisemils (commented):
> Reviewed 2 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: 12 of 13 files reviewed, 3 unresolved discussions (waiting on @faern)

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: 12 of 13 files reviewed, 1 unresolved discussion (waiting on @dlon and @pinkisemils)

___
*[`mullvad-nsis/src/lib.rs` line 7 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NPHrDoeB6bGJKto4gIf:-NPHrDoeB6bGJKto4gIg:b-ss0a80) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e2568b56e76a43dee9f69a82e7f98dbaeab788d8/mullvad-nsis/src/lib.rs#L7)):*
> ```XML
> use std::{os::windows::ffi::OsStrExt, panic::UnwindSafe, ptr};
> 
> macro_rules! check {
> ```

This macro is only used in a single place. Do we have known future usage of it? Otherwise I'd argue that it would be shorter and more easily readable to just inline this code in the single callsite.

pinkisemils (approved):
> Reviewed 1 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`mullvad-nsis/src/lib.rs` line 3 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NP1vrkN9GSSOE6xfNms:-NPIsjHa-r74N2tuQio7:b-88uo7e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e2568b56e76a43dee9f69a82e7f98dbaeab788d8/mullvad-nsis/src/lib.rs#L3)):*

We probably don't really need logging. But if we do, the way to go might be to add a `mullvad_get_last_error` function and storing it using thread-local storage that we update using `set_last_error(new_error: Box<&[u16]>)`, or something like that.
</blockquote></details>

This TODO comment is a big vague - either remove it and let's not have logging, or add logging, or at least elaborate the concerns as to why logging isn't added now and when should it be added.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: 12 of 13 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`mullvad-nsis/src/lib.rs` line 3 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NP1vrkN9GSSOE6xfNms:-NPIzayS-dffpvLXo4uf:bdhfidp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e2568b56e76a43dee9f69a82e7f98dbaeab788d8/mullvad-nsis/src/lib.rs#L3)):*

This TODO comment is a big vague - either remove it and let's not have logging, or add logging, or at least elaborate the concerns as to why logging isn't added now and when should it be added.
</blockquote></details>

Done. The only reason not to log is that I think it's not likely to valuable.

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`mullvad-nsis/src/lib.rs` line 3 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4386#-NP1vrkN9GSSOE6xfNms:-NPJ1SFt4Cl2Gy5bxJ9t:bg90yg3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e2568b56e76a43dee9f69a82e7f98dbaeab788d8/mullvad-nsis/src/lib.rs#L3)):*

Done. The only reason not to log is that I think it's not likely to valuable.
</blockquote></details>

I agree, and if doing a thing brings no value, let's not do said thing :smile:

PR #4388, opened 2023-02-23
Fix back action collision

raksooo:
This PR fixes two different issues with our back actions. The main issue was that back actions weren't unique (e.g. `history.pop`) causing the `removeBackAction` function to remove more actions than it should. We also need to wrap the back action in a function when setting the state since it would otherwise be called.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4388)

### DIFF
--- a/gui/src/renderer/components/KeyboardNavigation.tsx
+++ b/gui/src/renderer/components/KeyboardNavigation.tsx
@@ -12,8 +12,14 @@ interface IKeyboardNavigationProps {
 // Listens for and handles keyboard shortcuts
 export default function KeyboardNavigation(props: IKeyboardNavigationProps) {
   const history = useHistory();
-  const [backAction, setBackAction] = useState<BackActionFn>();
+  const [backAction, setBackActionImpl] = useState<BackActionFn>();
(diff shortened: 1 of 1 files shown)


hankolsen (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4388)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`gui/src/renderer/components/KeyboardNavigation.tsx` line 82 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4388#-NOyPPX1361zBWhTIMlh:-NOyPPX2AVaFTLQOV2Sp:btq0kyl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/578922b251624f15363ff10a6e4afb1e41c2e662/gui/src/renderer/components/KeyboardNavigation.tsx#L82)):*
> ```XML
>   }, []);
> 
>   // Each back action needs to be unique to make `removeBackAction` work. This is accomplished by
> ```

Love the describing comments!

PR #4389, opened 2023-02-25
Allow building without .git directory

cole-h:
Nixpkgs does not keep around .git directories by default since they are not totally deterministic. This new HEAD hash check will obviously fail without a .git.

To  that end, the `get_dev_suffix` function now returns an `Option<String>` and will not fail the build if there is no dev suffix.

---

I don't know if this needs a changelog entry, but I'd be happy to add one! Just let me know!


faern:
I like not depending no git to build in general :+1: But I'll have to review this a bit later.

cole-h:
(I noticed CI failed because of clippy, so I fixed that lint and force-pushed)

mullvad-version/build.rs
```diff
@@ -53,12 +53,14 @@ fn get_product_version(target: Target) -> String {
         .trim()
         .to_owned();
 
-    let dev_suffix = get_dev_suffix(target, &version);
-
-    format!("{version}{dev_suffix}")
+    if let Some(dev_suffix) = get_dev_suffix(target, &version) {
+        format!("{version}{dev_suffix}")
+    } else {
+        version.to_string()
```
faern, line 59: No need for `.to_string()` here, `version` is already a `String`

mullvad-version/build.rs
```diff
@@ -67,17 +69,18 @@ fn get_dev_suffix(target: Target, product_version: &str) -> String {
 
     // Get the git commit hashes for the latest release and current HEAD
     let product_version_commit_hash = git_rev_parse_commit_hash(&release_tag);
-    let current_head_commit_hash =
-        git_rev_parse_commit_hash("HEAD").expect("HEAD must have a commit hash");
+    let current_head_commit_hash = git_rev_parse_commit_hash("HEAD");
```
faern, line 72: You can just add a `?` here to short circuit and early-return `None` if we fail to get the hash for `HEAD`.

faern (changes requested):

cole-h:
Good catches all, thanks!

faern:
Thank you for your contribution! Hope this helps for your builds.

cole-h:
It will very much, thank you!

(By the way, I wouldn't be offended if you pushed a change fixing this to do the implicit returns -- I just didn't want to wait on / make you trigger another round of "accept and run" of GHA CI for new contributors :sweat_smile: )

PR #4392, opened 2023-02-27
Override the API hostname using the InetSocketAddress

dlon:
Previously, this could only be used to override the IP. This diff also takes the hostname from the `InetSocketAddress` when `api-override` is enabled.

It's not possible to specify both IP and hostname in `InetSocketAddress`, so the IP is always resolved from the hostname.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4392)


albin-mullvad (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4392)* status: 1 of 3 files reviewed, all discussions resolved

PR #4410, opened 2023-03-02
Fix CI

pronebird:
1. This PR fixes CI. Funny thing linter picked up dangling files that are no longer a part of project. Perhaps rebase issue of the past. AddressCache was moved to REST framework a while ago.

2. Disable `hoistTry` rule introduced in SwiftFormat 0.51. That's how it looks like:

    ```swift
    fn(try throwingFn())
    ```
    
    enforces `try` at the beginning of expression:
    
    ```swift
    try fn(throwingFn())
    ```

    I don't have a  very strong opinion if that should be enforced or not. One thing I know that `fn` is not throwing, but `throwingFn()` is, so logically I don't understand why moving `try` in front should be a thing, but Swift is notorious for all kind of dubious decisions in regards of syntax. Maybe folks doing a lot of async+await are tired of `await`-ing on each thing they pass into some function. Not sure. 🤷 

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4410)


pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4410)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4425, opened 2023-03-06
Drop Debian 10 support (Upgrade build container to Debian 11)

faern:
[Debian 10 reached end of life on 2022-09-10](https://wiki.debian.org/DebianReleases) (It's LTS-supported for another year or so, but that is not maintained by the Debian security team). We are technically held back by the old kernel and the old compilers on Debian 10. So it would be a good quality of life for developers to be able to upgrade. It will also allow simplifying the firewall rules which will affect all Linux users in a tiny but positive way (way less verbose output of `nft list ruleset`).

This means we stop supporting glibc <2.31. Both Debian 11 and Ubuntu 20.04 use 2.31. Before this PR we supported glibc 2.28 and newer.

I have tried building a container image from this new dockerfile. And I have built the Linux app in it, and it seems to work. I tried both x86 and arm64. Building the GUI tests seem to fail and I will debug that.

If any Debian 10 user lands on this PR I want you to know that you can still continue using version `2023.2` and `2023.3` for quite a while longer. And vanilla WireGuard and OpenVPN will always be available to you also. So this change does not block you from using the Mullvad VPN service!

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4425)

### DIFF
--- a/building/Dockerfile
+++ b/building/Dockerfile
@@ -14,9 +14,11 @@
 #
 # And add -e TARGETS="aarch64-unknown-linux-gnu" to build for ARM64
 
-# Debian 10 is the oldest supported distro. It has the oldest glibc that we support
-# This checksum points to a 10.13-slim image.
-FROM debian@sha256:557ee531b81ce380d012d83b7bb56211572e5d6088d3e21a3caef7d7ed7f718b
+# When building the app, we must link towards the oldest glibc version we want to support.
+# This is currently glibc 2.31 which is the version in Debian 11.
+# When updating this base image, to find the checksum, run: podman inspect <image> | jq '.[]["Digest"]'
+# This checksum points to a 11.6-slim image.
+FROM debian@sha256:77f46c1cf862290e750e913defffb2828c889d291a93bdd10a7a0597720948fc
 
 LABEL org.opencontainers.image.source=https://github.com/mullvad/mullvadvpn-app
 LABEL org.opencontainers.image.description="Mullvad VPN app Linux build container"


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4425)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`README.md` line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4425#-NProJ6R-xquiQYfJ3w0:-NProJ6SAJuPEY4XyGUX:bixur81) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/21080f8246decb596bec2011da88a7c2ea511fad/README.md#L42)):*
> ```Markdown
> | Linux (Ubuntu)| The two latest LTS releases and the latest non-LTS releases |
> | Linux (Fedora) | The versions that are not yet [EOL](https://fedoraproject.org/wiki/End_of_life) |
> | Linux (Debian) | 11 and newer    |
> ```

Maybe we should say "stable and newer"?

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4425)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`README.md` line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4425#-NProJ6R-xquiQYfJ3w0:-NPv5cL25NfGQFb3Or4E:bxdget) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/21080f8246decb596bec2011da88a7c2ea511fad/README.md#L42)):*

Maybe we should say "stable and newer"?
</blockquote></details>

For fast moving distros I do think we should use a terminology not needing updating for every release. Such as the one we have for Fedora. However, Debian moves incredibly slowly, and it's not like we will immediately drop Debian 11 support once Debian 12 lands. So I think for slow moving stuff it's more accurate and reliable to refer to the actual version.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4425)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4425)* status: :shipit: complete! all files reviewed, all discussions resolved

faern:
I removed the part about bumping the README. I think that better belongs to the later PR where we bump `building/linux-container-image.txt` to actually use the new Debian 11 images.

PR #4439, opened 2023-03-13
Add mold linker to container

faern:
Updates the Linux (and indirectly Android) build container to have the [`mold`](https://github.com/rui314/mold.git) linker installed in it. Then adjust the container helper scripts to be able to easily use it.

The reason for not using it [like recommended for Rust](https://github.com/rui314/mold#how-to-use) is because that requires me to install `clang`, which blows up the size of the container significantly. With this `LD_PRELOAD` hack way of running it it's much easier.

I have opted out of mold for the real builds. Partially because it does not cut down on build times anyway, and partially because it feels iffy to depend on this linker without knowing more about it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439)

### DIFF
--- a/building/Dockerfile
+++ b/building/Dockerfile
@@ -31,6 +31,9 @@ ENV CARGO_TARGET_DIR=/root/.cargo/target
 ENV GOLANG_VERSION=1.18.5 \
     GOLANG_HASH=9e5de37f9c49942c601b191ac5fba404b868bfc21d446d6960acc12283d6e5f2
 
+# The pinned commit has this solved: https://github.com/rui314/mold/issues/1003.
+ARG MOLD_COMMIT_HASH=c4722fe5aed96295837d9150b20ef8698c7a28db
+
 # === Install/set up the image ===
 
 RUN dpkg --add-architecture arm64 && apt-get update -y && apt-get install -y \
@@ -59,6 +62,21 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
 ENV PATH=/root/.cargo/bin:$PATH
 ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER="aarch64-linux-gnu-gcc" \
     PKG_CONFIG_SYSROOT_DIR_aarch64_unknown_linux_gnu=/usr/lib/aarch64-linux-gnu
+
+# === mold (fast linker) ===
+# Allows linking Rust binaries significantly faster.
+
+RUN apt-get update -y && \
+    apt-get install -y --mark-auto cmake libssl-dev zlib1g-dev gcc g++ && \
+    rm -rf /var/lib/apt/lists/* && \
+    git clone https://github.com/rui314/mold.git && \
+    mkdir mold/build && cd mold/build && \
+    git reset --hard "$MOLD_COMMIT_HASH" && \
+    cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=c++ .. && \
+    cmake --build . -j $(nproc) && \
+    cmake --install . && \
+    cd ../.. && rm -rf mold && \
+    apt-get autoremove -y
 
 # === Volta for npm + node ===
 
--- a/building/container-run.sh
+++ b/building/container-run.sh
@@ -3,12 +3,10 @@
 # Gives you a shell or runs a given command in the Android or Linux build container,
 # as designated by the *-container-image.txt files. Uses podman unless overridden using the
 # environment variable `CONTAINER_RUNNER`. Note that this script uses named
-# docker volumes that can be overridden using enviornment variables (see the
+# docker volumes that can be overridden using environment variables (see the
 # beginning of the script).
 #
 # Usage: $ container-run.sh <linux/android> [command ...]
-#
-# Not supplying any command gives you a shell in the container.
 
 set -eu
 
@@ -18,6 +16,9 @@ CARGO_REGISTRY_VOLUME_NAME=${CARGO_REGISTRY_VOLUME_NAME:-"cargo-registry"}
 GRADLE_CACHE_VOLUME_NAME=${GRADLE_CACHE_VOLUME_NAME:-"gradle-cache"}
 ANDROID_CREDENTIALS_DIR=${ANDROID_CREDENTIALS_DIR:-""}
 CONTAINER_RUNNER=${CONTAINER_RUNNER:-"podman"}
+# Don't use mold by default until we have published the images with mold in
+# and swapped our builds to use them.
+USE_MOLD=${USE_MOLD:-"false"}
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 REPO_DIR="$( cd "$SCRIPT_DIR/.." && pwd )"
@@ -45,6 +46,11 @@ case ${1-:""} in
         exit 1
 esac
 
+optional_mold=""
+if [[ "$USE_MOLD" == "true" ]]; then
+    optional_mold="mold -run"
+fi
+
 set -x
 exec "$CONTAINER_RUNNER" run --rm -it \
     -v "$REPO_DIR:$REPO_MOUNT_TARGET:Z" \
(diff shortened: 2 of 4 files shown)


faern (commented):
The way the scripts are set up right now, they should be mergable before we bump the official containers. Using this new `container-run.sh` on a container without mold should work fine if you just set `USE_MOLD=false` for now.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-:-NQQEUGi2LwYfIarbqIj:b-o5ri91)* status: 0 of 4 files reviewed, all discussions resolved

___
*[`building/container-run.sh` line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-NQQEK4j2kqc6ycHi2o5:-NQQEK4j2kqc6ycHi2o6:b-k6d5lh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f7e1c0e1bfbdee66c6abd81486f72a9ccbc568dd/building/container-run.sh#L11)):*
> ```Shell
> # Usage: $ container-run.sh <linux/android> [command ...]
> #
> # Not supplying any command gives you a shell in the container.
> ```

This simply was untrue. I think it broke back when we added the `bash -c` part. But I did not spend time on fixing it right now. Invoking it with an extra `bash` argument seemed not a big deal IMO

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`building/Dockerfile` line 46 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-NQQFb1DBDGQ5r2Nu3Ol:-NQQFb1E448hPfjTfXTH:b-jwri71) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/28814effd6929993ad9178a01d940f97567ef373/building/Dockerfile#L46)):*
> ```Dockerfile
>     gcc-mingw-w64-x86-64 \
>     # For building mold further down
>     cmake libssl-dev zlib1g-dev gcc g++ file \
> ```

I think it'd be better to install these packages all in one step when building mold.

albin-mullvad (changes requested):
Please elaborate briefly in the PR description why we want to use mold as it doesn't sound like it improves build times etc.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-:-NQQG_kz7ufQRyRVg9-6:b2x58mo)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`-- commits` line 5 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-NQQHw1W2Q_Mq8S9He4p):*

Shouldn't the default for `USE_MOLD` be `false` until we bump the container image version/tag? Otherwise this would brake on the build servers as they use an image version without `mold`, or am I missing something?



_Code quote:_
```
Use mold by default in container, but opt out with USE_MOLD=false
```

albin-mullvad (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`-- commits` line 5 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-NQQHw1W2Q_Mq8S9He4p):*

Shouldn't the default for `USE_MOLD` be `false` until we bump the container image version/tag? Otherwise this would brake on the build servers as they use an image version without `mold`, or am I missing something?
</blockquote></details>

Nvm, missed the bit where the build-server defaults to `false`. However, I would have preferred keeping as `false` until the new image were published

faern (commented):
When linking in release mode we have full LTO optimizations on. That takes up the majority of the linking time and essentially forces the linking to be single threaded anyway. So in release mode we don't save any time with `mold` indeed. However, in debug builds/builds without compiler optimization where we don't use LTO, we save a lot of time by using `mold`.

Some rudimentary benchmarking on my laptop shows we cut down linking from ~12s to ~4 seconds for all the Rust binaries together:

```
$ find . -iname "main.rs" -exec touch "{}" \; && time mold -run cargo build
...
real	0m3.716s
user	0m11.712s
sys	0m4.660s

$ find . -iname "main.rs" -exec touch "{}" \; && time cargo build
...
real	0m12.139s
user	0m24.675s
sys	0m5.975s
```

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-:-NQQKdvi002uH0XUPif1:b4q1mmw)* status: all files reviewed, 1 unresolved discussion (waiting on @albin-mullvad)

___
*[`-- commits` line 5 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-NQQHw1W2Q_Mq8S9He4p):*

Nvm, missed the bit where the build-server defaults to `false`. However, I would have preferred keeping as `false` until the new image were published
</blockquote></details>

Ok. Good point. Fixed

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @albin-mullvad and @pinkisemils)

___
*[`building/Dockerfile` line 46 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-NQQFb1DBDGQ5r2Nu3Ol:-NQQM4H1-W1iJxO-Kf1M:bppvxda) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/28814effd6929993ad9178a01d940f97567ef373/building/Dockerfile#L46)):*

I think it'd be better to install these packages all in one step when building mold.
</blockquote></details>

Then we need to duplicate the full `apt-get update -y && apt-get install -y ... && rm -rf /var/lib/apt/lists/*`. That's doable I guess. Do you still think that's better? If so, maybe we should consider if there are more things in this list that better belong to other chapters below.

albin-mullvad (changes requested):
Thanks for the clarification!

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-:-NQQO1kA6ZJB8gRHv44s:b-uv4tn)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`building/Dockerfile` line 46 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-NQQFb1DBDGQ5r2Nu3Ol:-NQQPpusEMLmLSHJXMAm:bjw0hc5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/28814effd6929993ad9178a01d940f97567ef373/building/Dockerfile#L46)):*

Then we need to duplicate the full `apt-get update -y && apt-get install -y ... && rm -rf /var/lib/apt/lists/*`. That's doable I guess. Do you still think that's better? If so, maybe we should consider if there are more things in this list that better belong to other chapters below.
</blockquote></details>

I believe it might be a good idea to do what @pinkisemils suggested, at least for `libssl-dev`, since the layer where it's installed will still include it regardless if it's removed in a later layer.

albin-mullvad (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`building/Dockerfile` line 68 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439#-NQU0AiF-6UuaqPwimhT:-NQU0AiF-6UuaqPwimhU:b3jlu4g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3e87cb782238aaea14cf2faab79f9c71e85bf642/building/Dockerfile#L68)):*
> ```Dockerfile
> # === mold (fast linker) ===
> # Allows linking Rust binaries significantly faster.
> # This requires some build time dependencies, they are installed further up.
> ```

Remove or update this comment.



_Code quote:_
```Dockerfile
This requires some build time dependencies, they are installed further up.
```

albin-mullvad (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4439)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4445, opened 2023-03-14
Replace mullvadmangle tables with a single table

dlon:
Older kernels do not support inet tables with chains whose base type is `route` or `nat`. For that reason, we currently have two separate sets of tables and chains for IPv4 and IPv6. As we're dropping support for older kernels, we can simplify these rules now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4445)

### DIFF
--- a/talpid-core/src/firewall/linux.rs
+++ b/talpid-core/src/firewall/linux.rs
@@ -64,12 +64,6 @@ lazy_static! {
     static ref OUT_CHAIN_NAME: CString = CString::new("output").unwrap();
     static ref FORWARD_CHAIN_NAME: CString = CString::new("forward").unwrap();
     static ref PREROUTING_CHAIN_NAME: CString = CString::new("prerouting").unwrap();
-
-    /// We need two separate tables for compatibility with older kernels (holds true for kernel
-    /// version 4.19 but not 5.6), where the base filter type may not be `nftnl::ChainType::Route`
-    /// or `nftnl::ChainType::Nat` for inet tables.
-    static ref MANGLE_TABLE_NAME_V4: CString = CString::new("mullvadmangle4").unwrap();
-    static ref MANGLE_TABLE_NAME_V6: CString = CString::new("mullvadmangle6").unwrap();
     static ref MANGLE_CHAIN_NAME: CString = CString::new("mangle").unwrap();
     static ref NAT_CHAIN_NAME: CString = CString::new("nat").unwrap();
 
@@ -101,12 +95,6 @@ pub struct Firewall {
     fwmark: u32,
 }
 
-struct FirewallTables {
-    main: Table,
-    mangle_v4: Table,
-    mangle_v6: Table,
-}
-
 impl Firewall {
     pub fn from_args(args: FirewallArguments) -> Result<Self> {
         Firewall::new(args.fwmark)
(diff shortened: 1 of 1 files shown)


talpid-core/src/firewall/linux.rs
```diff
@@ -65,11 +65,10 @@ lazy_static! {
...
 
-    /// We need two separate tables for compatibility with older kernels (holds true for kernel
-    /// version 4.19 but not 5.6), where the base filter type may not be `nftnl::ChainType::Route`
-    /// or `nftnl::ChainType::Nat` for inet tables.
-    static ref MANGLE_TABLE_NAME_V4: CString = CString::new("mullvadmangle4").unwrap();
-    static ref MANGLE_TABLE_NAME_V6: CString = CString::new("mullvadmangle6").unwrap();
+    /// NOTE: The base filter type may not be `nftnl::ChainType::Route` or `nftnl::ChainType::Nat`
+    /// for inet tables when running older kernels. It's been confirmed to work on 5.6, and not work
+    /// on 4.19.
+    static ref MANGLE_TABLE_NAME: CString = CString::new("mullvadmangle").unwrap();
```
dlon, line 71: We should remove the table completely and just use `mullvad` here.

faern: Yeah, if we can get away with fewer tables, we should do that!

faern (approved):
:lgtm: :partying_face: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4445#-:-NQZqMliA3s6d0uhDtQo:b-4x4msl)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`talpid-core/src/firewall/linux.rs` line 70 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4445#-NQZm4SB-to4ID4ApM-v:-NQZm4SB-to4ID4ApM-w:b-izjh8g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/17ea1c2e067948c9efb78708ae3a94bf8a439466/talpid-core/src/firewall/linux.rs#L70)):*
> ```XML
>     /// NOTE: The base filter type may not be `nftnl::ChainType::Route` or `nftnl::ChainType::Nat`
>     /// for inet tables when running older kernels. It's been confirmed to work on 5.6, and not work
>     /// on 4.19.
> ```

Why are we keeping the documentation for older kernels? This doc comment now moved from the `MANGLE_TABLE_NAME_V4` item to the `MANGLE_CHAIN_NAME` item.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4445)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`talpid-core/src/firewall/linux.rs` line 70 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4445#-NQZm4SB-to4ID4ApM-v:-NQZst4f96d37djLWhRr:b-qjtx3f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/17ea1c2e067948c9efb78708ae3a94bf8a439466/talpid-core/src/firewall/linux.rs#L70)):*

Why are we keeping the documentation for older kernels? This doc comment now moved from the `MANGLE_TABLE_NAME_V4` item to the `MANGLE_CHAIN_NAME` item.
</blockquote></details>

I've removed it.

PR #4443, opened 2023-03-14
Swap Linux builds over to Debian 11

faern:
Bumps the Linux container image to use the new Debian 11 based image. Also upgrades Rust from 1.67.1 to 1.68.0.

I'm not yet toggling `USE_MOLD` over to `true` by default since the Android images does not yet have mold installed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4443)

PR #4457, opened 2023-03-17
Use mold linker by default (USE_MOLD=true)

faern:
Follow up to #4443. Since both the Linux and Android builds now use the containers that have `mold` installed in them, we can enable it by default. This should speed up linking times by quite a lot *in debug builds*.

This will sadly not benefit the CI yet. The CI uses the containers, but are not executed through `container-run.sh` and mold is not the default linker in the containers. Fixing that is left as an exercise for the future.

Build servers still link with GNU ld by explicitly opting out with `USE_MOLD=false`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4457)

### DIFF
--- a/building/container-run.sh
+++ b/building/container-run.sh
@@ -16,9 +16,8 @@ CARGO_REGISTRY_VOLUME_NAME=${CARGO_REGISTRY_VOLUME_NAME:-"cargo-registry"}
 GRADLE_CACHE_VOLUME_NAME=${GRADLE_CACHE_VOLUME_NAME:-"gradle-cache"}
 ANDROID_CREDENTIALS_DIR=${ANDROID_CREDENTIALS_DIR:-""}
 CONTAINER_RUNNER=${CONTAINER_RUNNER:-"podman"}
-# Don't use mold by default until we have published the images with mold in
-# and swapped our builds to use them.
-USE_MOLD=${USE_MOLD:-"false"}
+# Use mold for linking by default. Build servers opt out of this and use GNU ld
+USE_MOLD=${USE_MOLD:-"true"}
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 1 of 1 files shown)


faern (commented):
Hmm. That would be a lot of places where we would need to add `mold -run`. Does not feel like it scales well? :thinking: I think that before embarking on such a journey we should check whether or not mold actually helps in CI at all, and if we can find a solution that does not involve manually adding it to very many `cargo build ...` invocations.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4457#-:-NQkDXoA8u-guJEjXylh:b-dv9pte)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (commented):
Unless you have a very easy way of doing this, I'd prefer to do that as a separate PR at some time. Just merging this still have the clear advantage that normal debug builds on developer machines will get faster.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4457#-:-NQkDtJ_6yzV2ul_TAH8:bz42297)* status: :shipit: complete! all files reviewed, all discussions resolved

albin-mullvad (approved):
Sure, but in that case I suggest updating the PR description since this won't impact the CI. Or does it still have some sort of CI impact? :thinking: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4457#-:-NQkEe0I0nu5TMjK5w2R:b3ftcdw)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (commented):
Sure! Fixed

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4457#-:-NQy2TOT7x8ULqQ0wAFN:b-l97ko5)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4495, opened 2023-03-27
Simplify settings persister

dlon:
This is only a suggestion, so feel free to shoot it down.

I've removed the setters and getters from `SettingsPersister`. Instead, settings are edited in a closure passed to `SettingsPersister::update`. If any changes are made in the closure, they're automatically saved to disk. I think there's no downside besides the fact that the settings must be cloned.

I've also made the file operations atomic. So the runtime settings state of the daemon should never change unless those changes are also successfully persisted.


faern (changes requested):
> Reviewed 11 of 14 files at r1, 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4495)* status: 13 of 14 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`mullvad-daemon/src/settings.rs` line 194 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4495#-NRc7tH3DAKlPFl33niq:-NRc7tH3DAKlPFl33nir:b-g23n5c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e9cbe633a5e85ce41974bbd83505e8450faebd1b/mullvad-daemon/src/settings.rs#L194)):*
> ```XML
> 
>         Self::save_inner(&self.path, &new_settings).await?;
>         self.settings = new_settings;
> ```

This will not store the new settings in `self.settings` if storing them to disk failed. I believe we have the opposite logic currently? If we invert it, then there is no need for `save_inner` and just `save` can be called directly.
___
*[`mullvad-fs/src/lib.rs` line 19 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4495#-NRbU1ih8uZIHWagzNDK:-NRbU1ih8uZIHWagzNDL:bt5x2cc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e9cbe633a5e85ce41974bbd83505e8450faebd1b/mullvad-fs/src/lib.rs#L19)):*
> ```XML
> impl AtomicFile {
>     pub async fn new<P: AsRef<Path>>(target_path: P) -> io::Result<Self> {
>         let target_path = PathBuf::from(target_path.as_ref());
> ```

Nit. If the first and only thing we want to do with the `target_path` argument is to convert it into a `PathBuf`I'd argue it would be nicer to make the type `P: Into<PathBuf>` if that works well with all the call sites.

faern (commented):
This is a pretty nice idea. It saves some code and some code even if it became a bit less explicit is still OK to read :+1:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4495#-:-NRcB0H77Luv50QB-zDS:bo98gn)* status: 13 of 14 files reviewed, 2 unresolved discussions (waiting on @dlon)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4495)* status: 13 of 14 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[`mullvad-daemon/src/settings.rs` line 194 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4495#-NRc7tH3DAKlPFl33niq:-NRcSIFBBnRfzYmORcCF:b-oet3al) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e9cbe633a5e85ce41974bbd83505e8450faebd1b/mullvad-daemon/src/settings.rs#L194)):*

This will not store the new settings in `self.settings` if storing them to disk failed. I believe we have the opposite logic currently? If we invert it, then there is no need for `save_inner` and just `save` can be called directly.
</blockquote></details>

The problem with the current code is that we update the fields even though the code that calls the setters fails if storing things to disk fails.

For example, if `set_obfuscation_settings` fails, then `on_set_obfuscation_settings` doesn't reconnect or call `event_listener.notify_settings()`. That's clearly incorrect.

That's why I think it makes sense to not store the new settings in memory if storing them on disk fails. As an alternative, we could make the function infallible and simply log I/O errors.

faern (approved):
> Reviewed 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4495)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-daemon/src/settings.rs` line 194 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4495#-NRc7tH3DAKlPFl33niq:-NRgmNJl4vrKnLNTXQG-:b-9v59zf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e9cbe633a5e85ce41974bbd83505e8450faebd1b/mullvad-daemon/src/settings.rs#L194)):*

The problem with the current code is that we update the fields even though the code that calls the setters fails if storing things to disk fails.

For example, if `set_obfuscation_settings` fails, then `on_set_obfuscation_settings` doesn't reconnect or call `event_listener.notify_settings()`. That's clearly incorrect.

That's why I think it makes sense to not store the new settings if storing them to disk fails. As an alternative, we could make the function infallible and simply log I/O errors.
</blockquote></details>

Ah. Yeah, this makes sense in a way. The new code here depends on a working settings persistence. We are no longer allowed to change in memory settings if the persistence fails. But that's OK. I have not seen a single error report where the storing of settings was not working. And it makes sense, as you say, to make sure everything is always in sync and actions related to changing a setting are performed when they are successfully changed only.

Let's keep this as is. Good improvement :+1:

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4495)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`mullvad-api/src/address_cache.rs` line 83 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4495#-NRhdnbL5Qur9JCVrlf5:-NRhdnbL5Qur9JCVrlf6:b-pjoper) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d630325c55920cb4c294812e40221fea7e6d7510/mullvad-api/src/address_cache.rs#L83)):*
> ```XML
>         };
> 
>         let mut file = mullvad_fs::AtomicFile::new(&**write_path)
> ```

Wow, this is interesting. Not optimal. But not going to block anything on this. Currently too lazy to pull this into my IDE to see what's going on :D

PR #4509, opened 2023-03-31
Drop header text in select location controller

pronebird:
1. Drop header text in select location controller
2. Drop solid background in navigation bar and configure it with transparent when scroll as the top edge, otherwise use default blur which looks more modern than solid fill.
3. Drop `UINavigationBar.titleLabelBottomInset` extension.
4. Switch `SelectLocationViewController` to `UITableViewController` -- no need to flash scrolls manually etc..
5. Refactor "select location" controller a bit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4509)


ios/MullvadVPN/View controllers/SelectLocation/SelectLocationViewController.swift
```diff
@@ -11,36 +11,26 @@ import MullvadTypes
 import RelayCache
 import UIKit
 
-final class SelectLocationViewController: UIViewController, UITableViewDelegate {
-    private var tableView: UITableView?
+final class SelectLocationViewController: UITableViewController {
```
rablador, line 14: Turning the VC into a UITableViewController will make it less flexible if we want to add subviews to it. For example, the iPad version of the upcoming search bar should probably be added as such.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4509)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadVPN/View controllers/SelectLocation/SelectLocationViewController.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4509#-NRtZdUJWa5Ifot-d46a-r1-14:-NS6-doH3iyTzkDFHDp3:b-9ax3b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e095a770f55b368fdae5d763bf696ba73de890d0/ios/MullvadVPN/View%20controllers/SelectLocation/SelectLocationViewController.swift#L14)):*

Turning the VC into a UITableViewController will make it less flexible if we want to add subviews to it. For example, the iPad version of the upcoming search bar should probably be added as such.
</blockquote></details>

What exactly is the challenge that you see in regards of search bar and why it cannot be overcome with table header or by configuring navigation item with search controller and providing appropriate configuration?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4509)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadVPN/View controllers/SelectLocation/SelectLocationViewController.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4509#-NRtZdUJWa5Ifot-d46a-r1-14:-NS6IBZL0uAX4l-JbKYV:bckhok1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e095a770f55b368fdae5d763bf696ba73de890d0/ios/MullvadVPN/View%20controllers/SelectLocation/SelectLocationViewController.swift#L14)):*

In my experience the table header is a little finnicky to work with, but maybe it's better since I last tried. As for the nav bar, we'll not be using it for anything else, so I thought it might be a little redundant. We don't really use much of the native functionality of the search controller either, so in practice a simple search bar at the top of the view would suffice.

Somewhat related is if we want to animate the bottom of the tableview alongside the keyboard (when shown/hidden). Maybe not relevant at this point, since we dismiss the keyboard on drag anyway, but I'm not sure how we'd accomplish that with an innate UITableViewController.
</blockquote></details>

Agree on table header. Speaking of handling keyboard. Table view is a subclass of scroll view so adjusting content insets has been the way ever since. It's only recently that we gained keyboard layout guide in iOS 15.0. Either way, I changed back to using vanilla `UIViewController`.

rablador (approved):
> Reviewed 6 of 6 files at r1, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4509)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4515, opened 2023-04-03
Improve AutoLayout constraints creation

pronebird:
This is a proposal to introduce `AutoLayoutAnchorsProtocol` and remove distinction between `UIView` and `UILayoutGuide`. Since both implement the same properties that provide access to AutoLayout anchors there is no need to handle them separately, instead we can add a protocol and mark them as conforming to it to remove that distinction and reuse the same code to create constraints.


rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4515)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4523, opened 2023-04-05
Persist next private key instead of creating it during each rotation

pronebird:
This PR persists next private key we're rotating to, so that even if we fail to rotate the key during current rotation task, we can pick up where we left during next rotation task.


rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4523)* status: :shipit: complete! all files reviewed, all discussions resolved

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4523)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4530, opened 2023-04-06
Wipe all settings on app reinstall

rablador:
To wipe the settings two conditions need to be satisfied.

- UserDefaults flag `hasFinishedFirstTimeLaunch == true`
- Keychain flag `previousAppVersion == true`

`hasFinishedFirstTimeLaunch` determines - as suggested in the issue description - whether a wipe is desirable or not. However, for it to go into effect we need the user to have previously been on a compatible app version, ie that has the UserDefaults flag.

Use cases:

- User does a clean install with no previous data. `hasFinishedFirstTimeLaunch` check succeeds and `previousAppVersion` check fail - no wipe.
- User updates from a version without the UserDefaults flag. `hasFinishedFirstTimeLaunch` check succeeds and `previousAppVersion` check fails - no wipe.
- User updates from a version with the UserDefaults flag. `hasFinishedFirstTimeLaunch` check fails and `previousAppVersion` check succeeds - no wipe.
- User reinstalls from a version with the UserDefaults flag. Both `hasFinishedFirstTimeLaunch` and `previousAppVersion` checks succeed - wipe.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)

### DIFF
--- a/ios/MullvadVPN/AppDelegate.swift
+++ b/ios/MullvadVPN/AppDelegate.swift
@@ -363,6 +363,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     }
 
     private func startInitialization(application: UIApplication) {
+        let wipeSettingsOperation = getWipeSettingsOperation()
+
         let loadTunnelStoreOperation = AsyncBlockOperation(dispatchQueue: .main) { operation in
             self.tunnelStore.loadPersistentTunnels { error in
                 if let error = error {
@@ -395,7 +397,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
                 migrationUIHandler.showMigrationError(error, completionHandler: finishHandler)
             }
         }
-        migrateSettingsOperation.addDependency(loadTunnelStoreOperation)
+        migrateSettingsOperation.addDependencies([wipeSettingsOperation, loadTunnelStoreOperation])
 
         let initTunnelManagerOperation = AsyncBlockOperation(dispatchQueue: .main) { operation in
             self.tunnelManager.loadConfiguration { error in
@@ -428,6 +430,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
         operationQueue.addOperations(
             [
+                wipeSettingsOperation,
                 loadTunnelStoreOperation,
                 migrateSettingsOperation,
                 initTunnelManagerOperation,
@@ -435,6 +438,36 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             ],
             waitUntilFinished: false
         )
+    }
+
+    /// Returns an operation that acts on two conditions:
+    /// 1. Has the app been launched at least once after install? (`FirstTimeLaunch.hasFinished`)
+    /// 2. Has the app - at some point in time - been updated from a version compatible with wiping settings?
+    /// (`SettingsManager.getShouldWipeSettings()`)
+    /// If (1) is `false` and (2) is `true`, we know that the app has been freshly installed/reinstalled and is
+    /// compatible, thus triggering a settings wipe.
+    private func getWipeSettingsOperation() -> AsyncBlockOperation {
+        return AsyncBlockOperation {
+            if !FirstTimeLaunch.hasFinished, SettingsManager.getShouldWipeSettings() {
+                if let deviceState = try? SettingsManager.readDeviceState(),
+                   let accountData = deviceState.accountData,
+                   let deviceData = deviceState.deviceData
+                {
+                    _ = self.devicesProxy.deleteDevice(
+                        accountNumber: accountData.number,
+                        identifier: deviceData.identifier,
+                        retryStrategy: .noRetry
+                    ) { _ in
+                        // Do nothing.
+                    }
+                }
+
+                SettingsManager.resetStore(completely: true)
+            }
+
+            FirstTimeLaunch.setHasFinished()
+            SettingsManager.setShouldWipeSettings()
+        }
     }
 
     // MARK: - StorePaymentManagerDelegate
--- a/ios/MullvadVPN/SettingsManager/SettingsManager.swift
+++ b/ios/MullvadVPN/SettingsManager/SettingsManager.swift
@@ -68,6 +68,23 @@ enum SettingsManager {
         }
     }
 
+    // MARK: - Should wipe settings
+
+    static func getShouldWipeSettings() -> Bool {
+        return (try? store.read(key: .shouldWipeSettings)) != nil
+    }
+
+    static func setShouldWipeSettings() {
+        do {
+            try store.write(Data(), for: .shouldWipeSettings)
+        } catch {
+            logger.error(
+                error: error,
+                message: "Failed to set should wipe settings."
+            )
+        }
+    }
+
     // MARK: - Settings
 
     static func readSettings() throws -> TunnelSettingsV2 {
@@ -153,6 +170,48 @@ enum SettingsManager {
         }
     }
 
+    /// Removes all legacy settings, device state and tunnel settings but keeps the last used
+    /// account number stored.
+    static func resetStore(completely: Bool = false) {
+        logger.debug("Reset store.")
+
+        do {
+            try store.delete(key: .deviceState)
+        } catch {
+            if (error as? KeychainError) != .itemNotFound {
+                logger.error(error: error, message: "Failed to delete device state.")
+            }
+        }
+
+        do {
+            try store.delete(key: .settings)
+        } catch {
+            if (error as? KeychainError) != .itemNotFound {
+                logger.error(error: error, message: "Failed to delete settings.")
+            }
+        }
+
+        if completely {
+            do {
+                try store.delete(key: .lastUsedAccount)
+            } catch {
+                if (error as? KeychainError) != .itemNotFound {
+                    logger.error(error: error, message: "Failed to delete last used account.")
+                }
+            }
+
+            do {
+                try store.delete(key: .shouldWipeSettings)
+            } catch {
+                if (error as? KeychainError) != .itemNotFound {
+                    logger.error(error: error, message: "Failed to delete should wipe settings.")
+                }
+            }
+        }
+
+        Self.deleteAllLegacySettings()
+    }
+
     // MARK: - Private
 
     private static func migrateLegacySettings(
@@ -210,30 +269,6 @@ enum SettingsManager {
         logger.error(error: error, message: "Encountered an unknown version.")
 
         throw error
-    }
-
-    /// Removes all legacy settings, device state and tunnel settings but keeps the last used
-    /// account number stored.
-    private static func resetStore() {
-        logger.debug("Reset store.")
-
-        do {
-            try store.delete(key: .deviceState)
-        } catch {
-            if (error as? KeychainError) != .itemNotFound {
-                logger.error(error: error, message: "Failed to delete device state.")
-            }
-        }
-
-        do {
-            try store.delete(key: .settings)
-        } catch {
-            if (error as? KeychainError) != .itemNotFound {
-                logger.error(error: error, message: "Failed to delete settings.")
-            }
-        }
-
-        Self.deleteAllLegacySettings()
     }
 
     // MARK: - Legacy settings
@@ -409,6 +444,7 @@ enum SettingsKey: String, CaseIterable {
     case settings = "Settings"
     case deviceState = "DeviceState"
     case lastUsedAccount = "LastUsedAccount"
+    case shouldWipeSettings = "shouldWipeSettings"
 }
 
 /// An error type describing a failure to read or parse settings version.
--- a/ios/CHANGELOG.md
+++ b/ios/CHANGELOG.md
@@ -25,6 +25,7 @@ Line wrap the file at 100 chars.                                              Th
 ## [Unreleased]
 ### Added
 - Add search functionality to location selection view.
+- Wipe all settings on app reinstall.
 
 ### Changed
 - Changed key rotation interval from 4 to 14 days.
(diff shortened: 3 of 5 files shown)


ios/MullvadVPN/AppDelegate.swift
```diff
@@ -437,6 +439,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         )
     }
 
+    private func checkWipeSettings() {
+        guard var previousAppVersion = try? SettingsManager.getPreviousAppVersion() else { return }
```
MrChocolatine, line 443: Looks like it can be a constant.

```suggestion
        guard let previousAppVersion = try? SettingsManager.getPreviousAppVersion() else { return }
```

pronebird (changes requested):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: all files reviewed, 2 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 445 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530#-NSLMpAd97HS0CjK9kFt:-NSLMpAd97HS0CjK9kFu:bg233r9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ae40f53af31410257c799ff9b102404e4a052acd/ios/MullvadVPN/AppDelegate.swift#L445)):*
> ```Swift
>         guard let previousAppVersion = try? SettingsManager.getPreviousAppVersion() else { return }
> 
>         let previousAppVersionIsCompatible = previousAppVersion >= FirstTimeLaunch.compatibleAppVersion
> ```

Ugh is it lexicographical comparison that I witness in the wild being used to decide whether we should wipe user settings? :-)

Second, why does this check exist? It would suffice to wipe settings only if user defaults do not contain a flag and keychain entry contains "previousAppVersion" version field. 

Two birds with one stone: existing users are safe since their keychain won't contain version field, reinstalls are handled if user removed the app since user defaults field is gone by then.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: all files reviewed, 2 unresolved discussions (waiting on @MrChocolatine and @pronebird)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 443 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530#gh-1159521061:-NSLRKUn3hrjg4LB-K8s:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/32fa562ffa5539d0b1a5dcda34cfcdfe0549c707/ios/MullvadVPN/AppDelegate.swift#L443)):*
<details><summary><i>Previously, MrChocolatine wrote…</i></summary><blockquote>

Looks like it can be a constant.

```suggestion
        guard let previousAppVersion = try? SettingsManager.getPreviousAppVersion() else { return }
```
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/AppDelegate.swift` line 445 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530#-NSLMpAd97HS0CjK9kFt:-NSLRMGZ4RRVhQ9tB9Sr:bf0kvka) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ae40f53af31410257c799ff9b102404e4a052acd/ios/MullvadVPN/AppDelegate.swift#L445)):*

Ugh is it lexicographical comparison that I witness in the wild being used to decide whether we should wipe user settings? :-)

Second, why does this check exist? It would suffice to wipe settings only if user defaults do not contain a flag and keychain entry contains "previousAppVersion" version field. 

Two birds with one stone: existing users are safe since their keychain won't contain version field, reinstalls are handled if user removed the app since user defaults field is gone by then.
</blockquote></details>

It is... 🙈 Didn't feel great adding it tbh, but seing how we'd be strict about versioning I thought it ...safe? I'll see myself out. And improve on the logic, as suggested, on the way.

ios/MullvadVPN/SettingsManager/SettingsManager.swift
```diff
@@ -409,6 +437,7 @@ enum SettingsKey: String, CaseIterable {
     case settings = "Settings"
     case deviceState = "DeviceState"
     case lastUsedAccount = "LastUsedAccount"
+    case isSettingsWipeCompatible = "IsSettingsWipeCompatible"
```
buggmagnet, line 440: Suggestion: Rename this "shouldWipeSettings"

buggmagnet (changes requested):

ios/MullvadVPN/SettingsManager/SettingsManager.swift
rablador: Agreed, much better.

ios/MullvadVPN/SettingsManager/SettingsManager.swift
```diff
@@ -153,6 +163,48 @@ enum SettingsManager {
...
 
+    /// Removes all legacy settings, device state and tunnel settings but keeps the last used
+    /// account number stored.
+    static func resetStore(completely: Bool = false) {
+        logger.debug("Reset store.")
+
+        do {
+            try store.delete(key: .deviceState)
+        } catch {
+            if (error as? KeychainError) != .itemNotFound {
```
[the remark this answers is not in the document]
rablador: I believe the checks are here in order to keep logs somewhat less crowded or misleading (since trying to delete a non-existing entry isn't really an error). What does @pronebird think?

ios/MullvadVPN/AppDelegate.swift
```diff
@@ -437,6 +439,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         )
     }
 
+    private func checkWipeSettings() {
+        guard let previousAppVersion = try? SettingsManager.getPreviousAppVersion() else { return }
+
+        let previousAppVersionIsCompatible = previousAppVersion >= FirstTimeLaunch.compatibleAppVersion
```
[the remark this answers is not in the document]
rablador: Not applicable anymore. Changes made as per the suggestions here: https://github.com/mullvad/mullvadvpn-app/pull/4530#pullrequestreview-1374724519

ios/MullvadVPN/SettingsManager/SettingsManager.swift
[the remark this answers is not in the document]
pronebird: Keychain throws errors when trying to delete an item that does not exist, so we ignore them here by checking against `.itemNotFound` error.

pronebird (commented):
> Reviewed 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: 1 of 5 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @MrChocolatine)

___
*[`ios/MullvadVPN/SettingsManager/SettingsManager.swift` line 174 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530#gh-1159762542:-NSl1kF8-pbRHIj-IQJ3:br3vbe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ddf43b92838e4c040020f2b7b1fea7667babb476/ios/MullvadVPN/SettingsManager/SettingsManager.swift#L174)):*

I just noticed something else that could go wrong here.

The call tree of this function looks like this
`didFinishLaunchingWithOptions` -> `startInitialization` -> `checkWipeSettings` -> `resetStore` -> ... -> `SecItemDelete` 

The documentation for that last function mentions 
```SecItemDelete blocks the calling thread, so it can cause your app’s UI to hang if called from the main thread. Instead, call SecItemDelete from a background dispatch queue or async function.```

Here we can unintentionally block the main thread, which on top of that is during the Application launch, which, in the worst case would prevent our app from starting (watch dog kills it if it takes too long to start).

Realistically, that shouldn't happen, because I don't see a scenario where `SecItemDelete` could block the main thread for too long.

But since we do several of those in a row, there's an increased risk.

It's out of scope to add thread safety here, but I think we should have a work item to look at making sure all keychain access operations are happening in a different thread than the main thread.

What do you think @pronebird and @rablador  ?
</blockquote></details>

As you said, realistically it shouldn't take too long, we're removing a few small entries from Keychain. I'd say let's keep this change for later. We could certainly run that as a block operation on a background queue but then we'd have to do the same with settings migration for consistency which then expands the scope of this task way too much.

**Review by pronebird — COMMENTED** [review]
> Reviewed 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @MrChocolatine)

**Review by rablador — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @MrChocolatine)

___
*[`ios/MullvadVPN/SettingsManager/SettingsManager.swift` line 174 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530#gh-1159762542:-NSpGKzN-y7eyKjPBPPI:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ddf43b92838e4c040020f2b7b1fea7667babb476/ios/MullvadVPN/SettingsManager/SettingsManager.swift#L174)):*

As you said, realistically it shouldn't take too long, we're removing a few small entries from Keychain. I'd say let's keep this change for later. We could certainly run that as a block operation on a background queue but then we'd have to do the same with settings migration for consistency which then expands the scope of this task way too much.
</blockquote></details>

Done.

**Review by buggmagnet — APPROVED** [review]
> Reviewed 1 of 3 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: all files reviewed, 5 unresolved discussions (waiting on @MrChocolatine)

**Review by pronebird — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: 3 of 5 files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @MrChocolatine, and @rablador)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 444 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530#-NT-C0jz5xgs5UXqXlnk:-NT-C0jz5xgs5UXqXlnl:biamowa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2f6e179cb07ae73eca7fe122e71b282d21a43586/ios/MullvadVPN/AppDelegate.swift#L444)):*
> ```Swift
> 
>     private func getWipeSettingsOperation() -> AsyncBlockOperation {
>         return AsyncBlockOperation { [weak self] in
> ```

Nit: you can retain `self`. It should not create any cycle as operation will release it upon completion.

**Review by pronebird — COMMENTED** [review]
> Reviewed 1 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @MrChocolatine, and @rablador)

**Review by pronebird — COMMENTED** [review]
> Dismissed @MrChocolatine and @buggmagnet from 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

File: ios/MullvadVPN/AppDelegate.swift [hunk 5]
```diff
@@ -437,6 +440,30 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         )
     }
 
+    private func getWipeSettingsOperation() -> AsyncBlockOperation {
```
buggmagnet, line 443: I think we should comment this function to explain in a few lines what are the expected conditions for a wipe, it took some time to test this properly, let's make sure our future selves profit from that time spent.

LGTM otherwise.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: 3 of 6 files reviewed, 1 unresolved discussion (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 443 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530#gh-1168316484:-NTE3bUW9GKCxR8gGF7b:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b1ec5a7e608e4a4f3b120c55aeb48497206cdfc5/ios/MullvadVPN/AppDelegate.swift#L443)):*

I think we should comment this function to explain in a few lines what are the expected conditions for a wipe, it took some time to test this properly, let's make sure our future selves profit from that time spent.

LGTM otherwise.
</blockquote></details>

Done.

pronebird (approved):
:lgtm:

> Reviewed 1 of 3 files at r6, 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530#-:-NTNsweWEypS1pv-2BDa:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

buggmagnet (approved):
> Reviewed 1 of 2 files at r5, 1 of 3 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: all files reviewed, 1 unresolved discussion

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4530)* status: all files reviewed, 1 unresolved discussion

PR #4531, opened 2023-04-06
Log settings summary when connecting

dlon:
For debugging purposes, log select settings whenever entering the connecting state.

The following settings are logged:
* mssfix/MTU
* WG over IPv6
* Multihop
* IPv6 (in tunnel)
* LAN sharing
* PQ
* Which content blockers are used and whether custom DNS is enabled
* Obfuscation state

> [mullvad_daemon][DEBUG] ovpn mssfix: unset, wg ip: any, wg mtu: 1420, multihop: off, ipv6 (tun): off, lan: on, pq: on, obfs: udp2tcp, dns: default

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ Line wrap the file at 100 chars.                                              Th
 
 ## [Unreleased]
 ### Added
+- Log select settings on each connection attempt.
+
 #### Android
 - Add DNS content blockers.
 
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -894,9 +894,12 @@ where
             }
         }
 
-        match tunnel_state {
+        match &tunnel_state {
             TunnelState::Disconnected => self.state.disconnected(),
-            TunnelState::Error(ref error_state) => {
+            TunnelState::Connecting { .. } => {
+                log::debug!("Settings: {}", self.settings.summary());
+            }
+            TunnelState::Error(error_state) => {
                 if error_state.is_blocking() {
                     log::info!(
                         "Blocking all network connections, reason: {}",
--- a/mullvad-daemon/src/settings.rs
+++ b/mullvad-daemon/src/settings.rs
@@ -1,11 +1,16 @@
 #[cfg(not(target_os = "android"))]
 use futures::TryFutureExt;
-use mullvad_types::settings::Settings;
+use mullvad_types::{
+    relay_constraints::{RelayConstraints, RelaySettings, WireguardConstraints},
+    settings::{DnsState, Settings},
+};
 use rand::Rng;
 use std::{
+    fmt::{self, Display},
     ops::Deref,
     path::{Path, PathBuf},
 };
+use talpid_core::firewall::is_local_address;
 use talpid_types::ErrorExt;
 use tokio::{
     fs,
@@ -194,6 +199,13 @@ impl SettingsPersister {
         self.settings = new_settings;
         Ok(true)
     }
+
+    /// Return a compact summary of important settings
+    pub fn summary(&self) -> SettingsSummary<'_> {
+        SettingsSummary {
+            settings: &self.settings,
+        }
+    }
 }
 
 impl Deref for SettingsPersister {
@@ -201,6 +213,127 @@ impl Deref for SettingsPersister {
 
     fn deref(&self) -> &Self::Target {
         &self.settings
+    }
+}
+
+/// A compact summary of important settings
+pub struct SettingsSummary<'a> {
+    settings: &'a Settings,
+}
+
+impl<'a> Display for SettingsSummary<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let bool_to_label = |state| {
+            if state {
+                "on"
+            } else {
+                "off"
+            }
+        };
+
+        let relay_settings = self.settings.get_relay_settings();
+
+        write!(f, "openvpn mssfix: ")?;
+        Self::fmt_option(f, self.settings.tunnel_options.openvpn.mssfix)?;
+        write!(f, ", wg mtu: ")?;
+        Self::fmt_option(f, self.settings.tunnel_options.wireguard.mtu)?;
+
+        if let RelaySettings::Normal(RelayConstraints {
+            wireguard_constraints: WireguardConstraints { ip_version, .. },
+            ..
+        }) = relay_settings
+        {
+            write!(f, ", wg ip version: {ip_version}")?;
+        }
+
+        let multihop = matches!(
+            relay_settings,
+            RelaySettings::Normal(RelayConstraints {
+                wireguard_constraints: WireguardConstraints {
+                    use_multihop: true,
+                    ..
+                },
+                ..
+            })
+        );
+
+        write!(
+            f,
+            ", multihop: {}, ipv6 (tun): {}, lan: {}, pq: {}, obfs: {}",
+            bool_to_label(multihop),
+            bool_to_label(self.settings.tunnel_options.generic.enable_ipv6),
+            bool_to_label(self.settings.allow_lan),
+            self.settings.tunnel_options.wireguard.quantum_resistant,
+            self.settings.obfuscation_settings.selected_obfuscation,
+        )?;
+
+        // Print DNS options
+
+        write!(f, ", dns: ")?;
+
+        match self.settings.tunnel_options.dns_options.state {
+            DnsState::Default => {
+                let mut content = vec![];
+                let default_options = &self.settings.tunnel_options.dns_options.default_options;
+
+                if default_options.block_ads {
+                    content.push("ads");
+                }
+                if default_options.block_trackers {
+                    content.push("trackers");
+                }
+                if default_options.block_malware {
+                    content.push("malware");
+                }
+                if default_options.block_adult_content {
+                    content.push("adult");
+                }
+                if default_options.block_gambling {
+                    content.push("gambling");
+                }
+                if content.is_empty() {
+                    content.push("default");
+                }
+                write!(f, "{}", content.join(" "))?;
+            }
+            DnsState::Custom => {
+                // NOTE: Technically inaccurate, as the gateway IP is a local IP but isn't treated as one.
+                let contains_local = self
+                    .settings
+                    .tunnel_options
+                    .dns_options
+                    .custom_options
+                    .addresses
+                    .iter()
+                    .any(is_local_address);
+                let contains_public = self
+                    .settings
+                    .tunnel_options
+                    .dns_options
+                    .custom_options
+                    .addresses
+                    .iter()
+                    .any(|addr| !is_local_address(addr));
+
+                match (contains_public, contains_local) {
+                    (true, true) => f.write_str("custom, public, local")?,
+                    (true, false) => f.write_str("custom, public")?,
+                    (false, false) => f.write_str("custom, no addrs")?,
+                    (false, true) => f.write_str("custom, local")?,
+                }
+            }
+        }
+        Ok(())
+    }
+}
+
+impl<'a> SettingsSummary<'a> {
+    fn fmt_option<T: Display>(f: &mut fmt::Formatter<'_>, val: Option<T>) -> fmt::Result {
+        if let Some(inner) = &val {
+            inner.fmt(f)
+        } else {
+            f.write_str("unset")
+        }
     }
 }
 
--- a/mullvad-types/src/relay_constraints.rs
+++ b/mullvad-types/src/relay_constraints.rs
@@ -31,6 +31,15 @@ pub enum Constraint<T: fmt::Debug + Clone + Eq + PartialEq> {
     Only(T),
 }
 
+impl<T: fmt::Debug + fmt::Display + Clone + Eq> fmt::Display for Constraint<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+        match self {
+            Constraint::Any => "any".fmt(f),
+            Constraint::Only(value) => fmt::Display::fmt(value, f),
+        }
+    }
+}
+
 impl<T: fmt::Debug + Clone + Eq + PartialEq> Constraint<T> {
     pub fn unwrap(self) -> T {
         match self {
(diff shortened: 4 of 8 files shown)


faern (changes requested):
> Reviewed 2 of 2 files at r1, 3 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531)* status: 5 of 8 files reviewed, 5 unresolved discussions (waiting on @dlon)

___
*[`mullvad-daemon/src/lib.rs` line 899 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLSOC65EhAND6xzHpv:-NSLSOC65EhAND6xzHpw:bji7qd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/mullvad-daemon/src/lib.rs#L899)):*
> ```XML
>         match &tunnel_state {
>             TunnelState::Disconnected => self.state.disconnected(),
>             TunnelState::Connecting { .. } => log::debug!("{}", self.settings.summary()),
> ```

I think we should somehow prefix this so it's clear from the log output that it's a settings summary. Currently it's a bit anonymous.
___
*[`mullvad-daemon/src/lib.rs` line 899 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLSpWk-aOcQAy-lqI4:-NSLSpWlDizHd-Q_ZPJv:blkceuz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/mullvad-daemon/src/lib.rs#L899)):*
> ```XML
>         match &tunnel_state {
>             TunnelState::Disconnected => self.state.disconnected(),
>             TunnelState::Connecting { .. } => log::debug!("{}", self.settings.summary()),
> ```

It would be awesome if we could avoid logging settings that are irrelevant for the current connection. Currently it logs WireGuard MTU even if the connection is going to be OpenVPN etc.

I think we should have access to the tunnel type that will come up via `endpoint.tunnel_type` inside the `TunnelState::Connecting` variant.

Then we can actually also get hold of a few actual values of things and not just the settings. For example the PQ stuff can currently print `auto`. But that forces us to look up what `auto` means in this particular release of the app. What if we make it so `auto` is random, then we don't know if the connection that is being established uses PQ or not.

It would be great if we to as large extent as possible could see the *values being used* rather than the settings constraining those values.
___
*[`mullvad-daemon/src/settings.rs` line 236 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLSAUh0hHnYV-OHeUk:-NSLSAUh0hHnYV-OHeUl:bv4ulsy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/mullvad-daemon/src/settings.rs#L236)):*
> ```XML
>         let relay_settings = self.settings.get_relay_settings();
> 
>         write!(f, "ovpn mssfix: ")?;
> ```

I don't think we should abbreviate OpenVPN as `ovpn`. I don't think we do this in any other place and I don't think we should start now.
___
*[`mullvad-types/src/relay_constraints.rs` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLQcPZBa6--HAEjh4p:-NSLQcPZBa6--HAEjh4q:b-4j6911) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/mullvad-types/src/relay_constraints.rs#L34)):*
> ```XML
> }
> 
> impl<T: fmt::Debug + fmt::Display + Clone + Eq> fmt::Display for Constraint<T> {
> ```

Why must `T` implement `Debug`? Should `Display` not be enough?
___
*[`talpid-wireguard/src/lib.rs` line 679 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLRja65JTAQI0Dk02i:-NSLRja65JTAQI0Dk02j:bnay9vy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/talpid-wireguard/src/lib.rs#L679)):*
> ```XML
>         #[cfg(windows)] setup_done_tx: mpsc::Sender<std::result::Result<(), BoxedError>>,
>     ) -> Result<Box<dyn Tunnel>> {
>         log::debug!("Tunnel MTU: {}", config.mtu);
> ```

Is this logged separately because we don't know the default MTU in the other place where you log the settings summary?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531)* status: 4 of 8 files reviewed, 5 unresolved discussions (waiting on @faern)

___
*[`mullvad-daemon/src/lib.rs` line 899 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLSpWk-aOcQAy-lqI4:-NSLUjfi5_oTLGmqY5Hs:ba4iljo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/mullvad-daemon/src/lib.rs#L899)):*
> It would be awesome if we could avoid logging settings that are irrelevant for the current connection.

I thought that initially, but the log is very short. That's why I think that logging everything is probably okay. Do you disagree?

> It would be great if we to as large extent as possible could see the values being used rather than the settings constraining those values.

Good point about printing actual values. I did consider adding `mtu` to `TunnelEndpoint`, though it feels a bit out of place there.

We're already logging many of the actual values, though, like whether PQ is on or off. I wonder to what extent it's useful to repeat them.
___
*[`mullvad-daemon/src/lib.rs` line 899 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLSOC65EhAND6xzHpv:-NSLfwbwEO5ydSGTflCN:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/mullvad-daemon/src/lib.rs#L899)):*

I think we should somehow prefix this so it's clear from the log output that it's a settings summary. Currently it's a bit anonymous.
</blockquote></details>

Done.
___
*[`mullvad-daemon/src/settings.rs` line 236 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLSAUh0hHnYV-OHeUk:-NSLXJuv7oPybDv0RHlz:b-z7bnpj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/mullvad-daemon/src/settings.rs#L236)):*

I don't think we should abbreviate OpenVPN as `ovpn`. I don't think we do this in any other place and I don't think we should start now.
</blockquote></details>

The relay list does, but I don't object to adding three characters.
___
*[`mullvad-types/src/relay_constraints.rs` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLQcPZBa6--HAEjh4p:-NSLd0rP3OMM8Y_YkWvn:b5pwiw7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/mullvad-types/src/relay_constraints.rs#L34)):*

Why must `T` implement `Debug`? Should `Display` not be enough?
</blockquote></details>

You're correct. There were many pointless bounds here. I've made all of them as general as possible.
___
*[`talpid-wireguard/src/lib.rs` line 679 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLRja65JTAQI0Dk02i:-NSLZBvEALXeGD-FcA6C:b-7mudbq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/talpid-wireguard/src/lib.rs#L679)):*

Is this logged separately because we don't know the default MTU in the other place where you log the settings summary?
</blockquote></details>

Correct. Since the actual default isn't static, we'll probably have to get the actual value from the

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531)* status: 4 of 8 files reviewed, 3 unresolved discussions

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531)* status: 4 of 8 files reviewed, 4 unresolved discussions (waiting on @dlon)

___
*[`mullvad-daemon/src/lib.rs` line 899 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLSpWk-aOcQAy-lqI4:-NSzvB_tDf00-cHn8UdF:b-iyu8ku) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/mullvad-daemon/src/lib.rs#L899)):*

> It would be awesome if we could avoid logging settings that are irrelevant for the current connection.

I thought that initially, but the log is very short. That's why I think that logging everything is probably okay. Do you disagree?

> It would be great if we to as large extent as possible could see the values being used rather than the settings constraining those values.

Good point about printing actual values. I did consider adding `mtu` to `TunnelEndpoint`, though it feels a bit out of place there.

We're already logging many of the actual values, though, like whether PQ is on or off. I wonder to what extent it's useful to repeat them.
</blockquote></details>

I see. Let's go with this for now then.
___
*[`mullvad-daemon/src/settings.rs` line 236 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLSAUh0hHnYV-OHeUk:-NSz8Fa90M4DjubNIlS3:bjjzpze) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/mullvad-daemon/src/settings.rs#L236)):*

The relay list does, but I don't object to adding three characters.
</blockquote></details>

I can't find it anywhere in our codebase. Where do you mean. Anyway, please add the characters
___
*[`mullvad-daemon/src/settings.rs` line 246 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSz9VFM92quznTaFlud:-NSz9VFM92quznTaFlue:b-xt3k39) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/da0ff1d0e18f58bddc6c858d3092f3f30deec564/mullvad-daemon/src/settings.rs#L246)):*
> ```XML
>         }) = relay_settings
>         {
>             write!(f, ", wg ip: {ip_version}")?;
> ```

I don't think we should be using "ip" as a label for "IP version". I think it can very easily be confused with an IP address. Seeing "wg ip: any" makes me only think "For WireGuard it can use any IP", which is not the correct thing here.
___
*[`mullvad-daemon/src/settings.rs` line 280 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSzslCGET-543a2jqvw:-NSzslCGET-543a2jqvx:b-9z2i95) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/da0ff1d0e18f58bddc6c858d3092f3f30deec564/mullvad-daemon/src/settings.rs#L280)):*
> ```XML
>                     .tunnel_options
>                     .dns_options
>                     .default_options
> ```

All of the `if`s here can be greatly shortened by assigning `self.settings.tunnel_options.dns_options.default_options` to a variable.
___
*[`mullvad-daemon/src/settings.rs` line 322 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSzszyW6-K1qnc7caE6:-NSzszyW6-K1qnc7caE7:bsduhkt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/da0ff1d0e18f58bddc6c858d3092f3f30deec564/mullvad-daemon/src/settings.rs#L322)):*
> ```XML
>                 }
> 
>                 write!(f, ", dns: {}", content.join(" "))?;
> ```

Looks like this will print `dns: default ads` if I have ad blocking enabled?

Would it not make sense to break out the `, dns: ` part from this (and below) and have this print `default` only if the list of blockers is empty?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531)* status: 4 of 8 files reviewed, 4 unresolved discussions (waiting on @faern)

___
*[`mullvad-daemon/src/settings.rs` line 236 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSLSAUh0hHnYV-OHeUk:-NT-7H0L4QX57zOfb0KM:beek41i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ab8fb96de515bfc7758c9dee8e4986ce47dad7b1/mullvad-daemon/src/settings.rs#L236)):*

I can't find it anywhere in our codebase. Where do you mean. Anyway, please add the characters
</blockquote></details>

Fixed. It's not in our code but in the relay list currently returned by the API.
___
*[`mullvad-daemon/src/settings.rs` line 246 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSz9VFM92quznTaFlud:-NT-7P696wjjRfhSNsMx:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/da0ff1d0e18f58bddc6c858d3092f3f30deec564/mullvad-daemon/src/settings.rs#L246)):*

I don't think we should be using "ip" as a label for "IP version". I think it can very easily be confused with an IP address. Seeing "wg ip: any" makes me only think "For WireGuard it can use any IP", which is not the correct thing here.
</blockquote></details>

Done.
___
*[`mullvad-daemon/src/settings.rs` line 280 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSzslCGET-543a2jqvw:-NT-7PrOBRnMPGLlWX2j:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/da0ff1d0e18f58bddc6c858d3092f3f30deec564/mullvad-daemon/src/settings.rs#L280)):*

All of the `if`s here can be greatly shortened by assigning `self.settings.tunnel_options.dns_options.default_options` to a variable.
</blockquote></details>

Done.
___
*[`mullvad-daemon/src/settings.rs` line 322 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4531#-NSzszyW6-K1qnc7caE6:-NT-7WoU1k2tXFM7LyUJ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/da0ff1d0e18f58bddc6c858d3092f3f30deec564/mullvad-daemon/src/settings.rs#L322)):*

Looks like this will print `dns: default ads` if I have ad blocking enabled?

Would it not make sense to break out the `, dns: ` part from this (and below) and have this print `default` only if the list of blockers is empty?
</blockquote></details>

Done.

PR #4532, opened 2023-04-06
Improve content blocker settings interface

rablador:
As the list for blocking content becomes longer, having too many toggles at once does not look good.

So the content blockers should feature a drop-down with all the blocking toggles.
Instead of informational dialogs on each toggle it will now feature a general dialog for all and a warning dialog for Malware only.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)

### DIFF
--- a/ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift
+++ b/ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift
@@ -12,6 +12,8 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
     PreferencesDataSource.Section,
     PreferencesDataSource.Item
 >, UITableViewDelegate {
+    typealias InfoButtonHandler = (PreferencesDataSource.Item) -> Void
+
     enum CellReuseIdentifiers: String, CaseIterable {
         case settingSwitch
         case dnsServer
@@ -30,11 +32,14 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
     }
 
     private enum HeaderFooterReuseIdentifiers: String, CaseIterable {
+        case contentBlockerHeader
         case customDNSFooter
         case spacer
 
         var reusableViewClass: AnyClass {
             switch self {
+            case .contentBlockerHeader:
+                return SettingsContentBlockersHeaderView.self
             case .customDNSFooter:
                 return SettingsStaticTextFooterView.self
             case .spacer:
@@ -43,8 +48,8 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
         }
     }
 
-    enum Section: String, Hashable {
-        case mullvadDNS
+    enum Section: String, Hashable, CaseIterable {
+        case contentBlockers
         case customDNS
     }
 
@@ -57,6 +62,10 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
         case useCustomDNS
         case addDNSServer
         case dnsServer(_ uniqueID: UUID)
+
+        static var contentBlockers: [Item] {
+            return [.blockAdvertising, .blockTracking, .blockMalware, .blockAdultContent, .blockGambling]
+        }
 
         var accessibilityIdentifier: String {
             switch self {
@@ -101,7 +110,9 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
 
     private var isEditing = false
 
-    private(set) var viewModel = PreferencesViewModel()
+    private(set) var viewModel = PreferencesViewModel() { didSet {
+        preferencesCellFactory.viewModel = viewModel
+    }}
     private(set) var viewModelBeforeEditing = PreferencesViewModel()
     private let preferencesCellFactory: PreferencesCellFactory
     private weak var tableView: UITableView?
@@ -142,7 +153,6 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
         updateSnapshot()
         reloadCustomDNSFooter()
 
-        updateCellFactory(with: viewModel)
         viewModel.customDNSDomains.forEach { entry in
             self.reload(item: .dnsServer(entry.identifier))
         }
@@ -160,7 +170,6 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
             viewModel = mergedViewModel
         }
 
-        updateCellFactory(with: viewModel)
         updateSnapshot()
         reloadCustomDNSFooter()
     }
@@ -225,7 +234,6 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
         let removedEntry = viewModel.customDNSDomains.remove(at: sourceIndex)
         viewModel.customDNSDomains.insert(removedEntry, at: destinationIndex)
 
-        updateCellFactory(with: viewModel)
         updateSnapshot()
     }
 
@@ -236,16 +244,29 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
     }
 
     func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
-        return tableView.dequeueReusableHeaderFooterView(
-            withIdentifier: HeaderFooterReuseIdentifiers.spacer.rawValue
-        )
+        let sectionIdentifier = snapshot().sectionIdentifiers[section]
+
+        switch sectionIdentifier {
+        case .contentBlockers:
+            let view = tableView
+                .dequeueReusableHeaderFooterView(
+                    withIdentifier: HeaderFooterReuseIdentifiers.contentBlockerHeader.rawValue
+                ) as! SettingsContentBlockersHeaderView
+            configureContentBlockersHeader(view)
+            return view
+
+        case .customDNS:
+            return tableView.dequeueReusableHeaderFooterView(
+                withIdentifier: HeaderFooterReuseIdentifiers.spacer.rawValue
+            )
+        }
     }
 
     func tableView(_ tableView: UITableView, viewForFooterInSection section: Int) -> UIView? {
         let sectionIdentifier = snapshot().sectionIdentifiers[section]
 
         switch sectionIdentifier {
-        case .mullvadDNS:
+        case .contentBlockers:
             return nil
 
         case .customDNS:
@@ -260,14 +281,22 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
     }
 
     func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
-        return UIMetrics.sectionSpacing
+        let sectionIdentifier = snapshot().sectionIdentifiers[section]
+
+        switch sectionIdentifier {
+        case .contentBlockers:
+            return UITableView.automaticDimension
+
+        case .customDNS:
+            return UIMetrics.sectionSpacing
+        }
     }
 
     func tableView(_ tableView: UITableView, heightForFooterInSection section: Int) -> CGFloat {
         let sectionIdentifier = snapshot().sectionIdentifiers[section]
 
         switch sectionIdentifier {
-        case .mullvadDNS:
+        case .contentBlockers:
             return 0
 
         case .customDNS:
@@ -347,25 +376,30 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
     }
 
     private func updateSnapshot(animated: Bool = false, completion: (() -> Void)? = nil) {
-        var snapshot = NSDiffableDataSourceSnapshot<Section, Item>()
+        var newSnapshot = NSDiffableDataSourceSnapshot<Section, Item>()
 
-        snapshot.appendSections([.mullvadDNS, .customDNS])
-        snapshot.appendItems(
-            [.blockAdvertising, .blockTracking, .blockMalware, .blockAdultContent, .blockGambling],
-            toSection: .mullvadDNS
-        )
-        snapshot.appendItems([.useCustomDNS], toSection: .customDNS)
+        newSnapshot.appendSections(Section.allCases)
+
+        let oldSnapshot = snapshot()
+        if oldSnapshot.indexOfSection(.contentBlockers) != nil {
+            newSnapshot.appendItems(
+                oldSnapshot.itemIdentifiers(inSection: .contentBlockers),
+                toSection: .contentBlockers
+            )
+        }
+
+        newSnapshot.appendItems([.useCustomDNS], toSection: .customDNS)
 
         let dnsServerItems = viewModel.customDNSDomains.map { entry in
             return Item.dnsServer(entry.identifier)
         }
-        snapshot.appendItems(dnsServerItems, toSection: .customDNS)
+        newSnapshot.appendItems(dnsServerItems, toSection: .customDNS)
 
         if isEditing, viewModel.customDNSDomains.count < DNSSettings.maxAllowedCustomDNSDomains {
-            snapshot.appendItems([.addDNSServer], toSection: .customDNS)
+            newSnapshot.appendItems([.addDNSServer], toSection: .customDNS)
         }
 
-        apply(snapshot, completion: completion)
+        apply(newSnapshot, completion: completion)
     }
 
     private func reload(item: Item) {
@@ -374,10 +408,6 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
         {
             preferencesCellFactory.configureCell(cell, item: item, indexPath: indexPath)
         }
-    }
-
-    func updateCellFactory(with viewModel: PreferencesViewModel) {
-        preferencesCellFactory.viewModel = viewModel
     }
 
     private func setBlockAdvertising(_ isEnabled: Bool) {
@@ -478,7 +508,6 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
         let newDNSEntry = DNSServerEntry(address: "")
         viewModel.customDNSDomains.append(newDNSEntry)
 
-        updateCellFactory(with: viewModel)
         updateSnapshot(animated: true) { [weak self] in
             if oldViewModel.customDNSPrecondition != self?.viewModel.customDNSPrecondition {
                 self?.reloadCustomDNSFooter()
@@ -516,7 +545,6 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
 
         viewModel.customDNSDomains.remove(at: entryIndex)
 
-        updateCellFactory(with: viewModel)
         updateSnapshot(animated: true) { [weak self] in
             if oldViewModel.customDNSPrecondition != self?.viewModel.customDNSPrecondition {
                 self?.reloadCustomDNSFooter()
@@ -525,14 +553,46 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
     }
 
     private func reloadCustomDNSFooter() {
-        updateCellFactory(with: viewModel)
         reload(item: .useCustomDNS)
 
         let sectionIndex = snapshot().indexOfSection(.customDNS)!
-        if let reusableView = tableView?
-            .footerView(forSection: sectionIndex) as? SettingsStaticTextFooterView
-        {
-            configureFooterView(reusableView)
+
+        UIView.performWithoutAnimation {
+            tableView?.performBatchUpdates({
+                if let reusableView = tableView?.footerView(forSection: sectionIndex) as? SettingsStaticTextFooterView {
+                    configureFooterView(reusableView)
+                }
+            })
+        }
+    }
+
+    private func configureContentBlockersHeader(_ reusableView: SettingsContentBlockersHeaderView) {
+        reusableView.titleLabel.text = NSLocalizedString(
+            "BLOCK_ADS_CELL_LABEL",
+            tableName: "Preferences",
+            value: "DNS content blockers",
+            comment: ""
+        )
+
+        reusableView.infoButtonHandler = { [weak self] in
+            if let self = self {
+                self.delegate?.preferencesDataSource(self, didPressInfoButton: nil)
+            }
+        }
+
+        reusableView.didCollapseHandler = { [weak self] headerView in
+            guard let self = self else { return }
+
+            var snapshot = self.snapshot()
+
+            if headerView.isExpanded {
+                snapshot.deleteItems(Item.contentBlockers)
+            } else {
+                snapshot.appendItems(Item.contentBlockers, toSection: .contentBlockers)
+            }
+
+            headerView.isExpanded.toggle()
+            self.apply(snapshot, animatingDifferences: true)
         }
     }
 
@@ -579,5 +639,9 @@ extension PreferencesDataSource: PreferencesCellEventHandler {
         inputString: String
     ) -> Bool {
         return handleDNSEntryChange(with: identifier, inputString: inputString)
+    }
+
+    func didPressInfoButton(for item: Item) {
+        delegate?.preferencesDataSource(self, didPressInfoButton: item)
     }
 }
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -359,6 +359,7 @@
 		58FEEB58260B662E00A621A8 /* AutomaticKeyboardResponder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58FEEB57260B662E00A621A8 /* AutomaticKeyboardResponder.swift */; };
 		58FF2C03281BDE02009EF542 /* SettingsManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58FF2C02281BDE02009EF542 /* SettingsManager.swift */; };
 		7A09C98129D99215000C2CAC /* String+FuzzyMatch.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A09C98029D99215000C2CAC /* String+FuzzyMatch.swift */; };
+		7A7AD28F29DEDB1C00480EF1 /* SettingsContentBlockersHeaderView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A7AD28E29DEDB1C00480EF1 /* SettingsContentBlockersHeaderView.swift */; };
 		7A7AD28D29DC677800480EF1 /* FirstTimeLaunch.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A7AD28C29DC677800480EF1 /* FirstTimeLaunch.swift */; };
 		7A818F1F29F0305800C7F0F4 /* RootConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A818F1E29F0305800C7F0F4 /* RootConfiguration.swift */; };
 		7AD2DA1529DC4EB900250737 /* UISearchBar+Appearance.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AD2DA1429DC4EB900250737 /* UISearchBar+Appearance.swift */; };
@@ -945,6 +946,7 @@
 		58FEEB57260B662E00A621A8 /* AutomaticKeyboardResponder.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AutomaticKeyboardResponder.swift; sourceTree = "<group>"; };
 		58FF2C02281BDE02009EF542 /* SettingsManager.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SettingsManager.swift; sourceTree = "<group>"; };
 		7A09C98029D99215000C2CAC /* String+FuzzyMatch.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "String+FuzzyMatch.swift"; sourceTree = "<group>"; };
+		7A7AD28E29DEDB1C00480EF1 /* SettingsContentBlockersHeaderView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SettingsContentBlockersHeaderView.swift; sourceTree = "<group>"; };
 		7A7AD28C29DC677800480EF1 /* FirstTimeLaunch.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = FirstTimeLaunch.swift; sourceTree = "<group>"; };
 		7A818F1E29F0305800C7F0F4 /* RootConfiguration.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RootConfiguration.swift; sourceTree = "<group>"; };
 		7AD2DA1429DC4EB900250737 /* UISearchBar+Appearance.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "UISearchBar+Appearance.swift"; sourceTree = "<group>"; };
@@ -1276,17 +1278,19 @@
 		583FE01829C19709006E85F9 /* Settings */ = {
 			isa = PBXGroup;
 			children = (
-				5864AF0029C7879B005B0CD9 /* SettingsCellFactory.swift */,
-				58CCA01122424D11004F3011 /* SettingsViewController.swift */,
-				5867770F290975E8006F721F /* SettingsInteractorFactory.swift */,
-				58677711290976FB006F721F /* SettingsInteractor.swift */,
-				582BB1AE229566420055B6EF /* SettingsCell.swift */,
-				584D26C1270C8542004EA533 /* SettingsStaticTextFooterView.swift */,
-				58ACF64A26553C3F00ACE4B7 /* SettingsSwitchCell.swift */,
+				582BB1B2229574F40055B6EF /* SettingsAccountCell.swift */,
 				5819C2162729595500D6EC38 /* SettingsAddDNSEntryCell.swift */,
+				582BB1AE229566420055B6EF /* SettingsCell.swift */,
+				5864AF0029C7879B005B0CD9 /* SettingsCellFactory.swift */,
+				7A7AD28E29DEDB1C00480EF1 /* SettingsContentBlockersHeaderView.swift */,
 				58EE2E38272FF814003BFF93 /* SettingsDataSource.swift */,
 				58EE2E39272FF814003BFF93 /* SettingsDataSourceDelegate.swift */,
 				584D26C5270C8741004EA533 /* SettingsDNSTextCell.swift */,
+				58677711290976FB006F721F /* SettingsInteractor.swift */,
+				5867770F290975E8006F721F /* SettingsInteractorFactory.swift */,
+				584D26C1270C8542004EA533 /* SettingsStaticTextFooterView.swift */,
+				58ACF64A26553C3F00ACE4B7 /* SettingsSwitchCell.swift */,
+				58CCA01122424D11004F3011 /* SettingsViewController.swift */,
 			);
 			path = Settings;
 			sourceTree = "<group>";
@@ -2715,6 +2719,7 @@
 				58607A4D2947287800BC467D /* AccountExpiryInAppNotificationProvider.swift in Sources */,
 				06410E07292D108E00AFC18C /* SettingsStore.swift in Sources */,
 				586A950D290125F0007BAF2B /* PresentAlertOperation.swift in Sources */,
+				7A7AD28F29DEDB1C00480EF1 /* SettingsContentBlockersHeaderView.swift in Sources */,
 				5878F50229CDB989003D4BE2 /* ChangeLogCoordinator.swift in Sources */,
 				58B93A1326C3F13600A55733 /* TunnelState.swift in Sources */,
 				58B26E262943522400D5980C /* NotificationProvider.swift in Sources */,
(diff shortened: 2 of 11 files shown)


rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 0 of 32 files reviewed, 2 unresolved discussions

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NSuKDE58WfK5A4oKB_A:-NSuKDE58WfK5A4oKB_B:bwg5kbq) (no related file):*
Many images were updated when running the asset conversion script. Only IconInfo is new.

___
___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesViewController.swift` line 57 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NSuIbWa2HaIc9IMFUQY:-NSuIbWbCa70ly0ePv-T:b-l1hvek) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/742c652ee665876ce01071a8cd10fe45efa8526c/ios/MullvadVPN/View%20controllers/Preferences/PreferencesViewController.swift#L57)):*
> ```Swift
>     override func viewWillAppear(_ animated: Bool) {
>         super.viewWillAppear(animated)
>         tableView
> ```

Adding the header view in viewDidLoad generates an error about tableview being told to layout its visible cells and other contents without being in the view hierarchy. Adding it here instead seems to help.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 0 of 32 files reviewed, 4 unresolved discussions

___
*[`ios/MullvadVPN/View controllers/Settings/SettingsCell.swift` line 123 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NSyfvvpD18PywS0dl9G:-NSyfvvq74oOYmAwKu5g:b-aoom5e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bcce1731cc5d4c7893b7e9d98506019f9dca02c7/ios/MullvadVPN/View%20controllers/Settings/SettingsCell.swift#L123)):*
> ```Swift
>             infoButton.leadingAnchor.constraint(
>                 equalTo: titleLabel.trailingAnchor,
>                 constant: -UIMetrics.interButtonSpacing
> ```

Perhaps a more correct way would be to let the button be native size and then adjust the touch area (`point(inside...)`), but simply moving the button to the left and "over" the title works pretty well too. Suggestions?
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsContentBlockersHeaderView.swift` line 84 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NSyg-pjDVuXTGYMIak5:-NSyg-pjDVuXTGYMIak6:b-aoom5e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bcce1731cc5d4c7893b7e9d98506019f9dca02c7/ios/MullvadVPN/View%20controllers/Settings/SettingsContentBlockersHeaderView.swift#L84)):*
> ```Swift
>             infoButton.leadingAnchor.constraint(
>                 equalTo: titleLabel.trailingAnchor,
>                 constant: -UIMetrics.interButtonSpacing
> ```

Perhaps a more correct way would be to let the button be native size and then adjust the touch area (`point(inside...)`), but simply moving the button to the left and "over" the title works pretty well too. Suggestions?

buggmagnet (changes requested):
> Reviewed 26 of 32 files at r2, 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 27 of 33 files reviewed, 6 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesCellFactory.swift` line 110 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTDUe9k0jUvcRaokYMC:-NTDUe9k0jUvcRaokYMD:b-9wlq71) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aa32847adffc5c77ba57829f0b4e853afac8d3b6/ios/MullvadVPN/View%20controllers/Preferences/PreferencesCellFactory.swift#L110)):*
> ```Swift
>             cell.accessibilityHint = nil
>             cell.applySubCellStyling()
>             cell.setInfoButtonIsVisible(true)
> ```

Isn't it a bit overkill to have a method just for that when doing `infoButton.isHidden = false` is less code and achieves the same result ?
___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesViewController.swift` line 57 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NSuIbWa2HaIc9IMFUQY:-NTDSm4ZD1W54Wbz0t3R:b-3a1hj3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/742c652ee665876ce01071a8cd10fe45efa8526c/ios/MullvadVPN/View%20controllers/Preferences/PreferencesViewController.swift#L57)):*

Adding the header view in viewDidLoad generates an error about tableview being told to layout its visible cells and other contents without being in the view hierarchy. Adding it here instead seems to help.
</blockquote></details>

That's because we do update the cell views in the `viewDidLoad` method. 
If you set the `tableHeaderView` after the call to `dataSource?.update(from: interactor.dnsSettings)` in `viewDidLoad` the warning is not issued anymore.
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsCell.swift` line 123 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTDDhgV8oSVOpbcoaGf:-NTDDhgV8oSVOpbcoaGg:bniyr17) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aa32847adffc5c77ba57829f0b4e853afac8d3b6/ios/MullvadVPN/View%20controllers/Settings/SettingsCell.swift#L123)):*
> ```Swift
>             infoButton.leadingAnchor.constraint(
>                 equalTo: titleLabel.trailingAnchor,
>                 constant: -UIMetrics.interButtonSpacing
> ```

Small pet peeve of mine, but I think it's a code smell when we have negative values in layout constraints constants.
If we swap the order of the constraint like so
```
titleLabel.trailingAnchor.constraint(equalTo: infoButton.leadingAnchor, constant: UIMetrics.interButtonSpacing)
```

We should be able to get rid of that `-` sign. It's easy to miss, and it makes it harder to reason about values when you have to deal with both positive and negative values.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 27 of 33 files reviewed, 6 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesCellFactory.swift` line 110 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTDUe9k0jUvcRaokYMC:-NTE3upDApJHiqzCSppQ:b-cc7mem) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aa32847adffc5c77ba57829f0b4e853afac8d3b6/ios/MullvadVPN/View%20controllers/Preferences/PreferencesCellFactory.swift#L110)):*

Isn't it a bit overkill to have a method just for that when doing `infoButton.isHidden = false` is less code and achieves the same result ?
</blockquote></details>

I don't disagree, but since infoButton is private I thought it unnecessary to expose if just for this one thing.
___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesViewController.swift` line 57 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NSuIbWa2HaIc9IMFUQY:-NTE4cfSAT39Xorm6mQY:b3t1g35) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/742c652ee665876ce01071a8cd10fe45efa8526c/ios/MullvadVPN/View%20controllers/Preferences/PreferencesViewController.swift#L57)):*

That's because we do update the cell views in the `viewDidLoad` method. 
If you set the `tableHeaderView` after the call to `dataSource?.update(from: interactor.dnsSettings)` in `viewDidLoad` the warning is not issued anymore.
</blockquote></details>

Ah, that's better.



_Code quote:_
```Swift
tableView.tableHeaderView
```
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsCell.swift` line 123 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTDDhgV8oSVOpbcoaGf:-NTE4ldGAClJqgPqeMEc:b-a1tfos) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aa32847adffc5c77ba57829f0b4e853afac8d3b6/ios/MullvadVPN/View%20controllers/Settings/SettingsCell.swift#L123)):*

Small pet peeve of mine, but I think it's a code smell when we have negative values in layout constraints constants.
If we swap the order of the constraint like so
```
titleLabel.trailingAnchor.constraint(equalTo: infoButton.leadingAnchor, constant: UIMetrics.interButtonSpacing)
```

We should be able to get rid of that `-` sign. It's easy to miss, and it makes it harder to reason about values when you have to deal with both positive and negative values.

</blockquote></details>

I don't agree on this one. To me the direction - or context - is more important than the value. In this case the context is the `infoButton` and it's relation to others, ie we're positioning the `infoButton` in relation to the label rather than the opposite.

A better example is when we have a single view constrained to a superview, with margins. To me the subview is the thing being positioned, so the direction should be subview anchors -> superview anchors. It makes little sense constraining the superview to the subview.

I know that the Apple docs suggest switching the order to eliminate negative values, so I'm probably not in the majority here. It's fine to change it according to your suggestion, but I thought I'd give my perspective on it at least. :)

buggmagnet (commented):
> Reviewed 2 of 32 files at r2, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 30 of 33 files reviewed, 6 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPN/View controllers/Settings/SettingsCell.swift` line 123 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTDDhgV8oSVOpbcoaGf:-NTIK9MO2K-8OqAPn0jh:b-4rxn02) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aa32847adffc5c77ba57829f0b4e853afac8d3b6/ios/MullvadVPN/View%20controllers/Settings/SettingsCell.swift#L123)):*

I don't agree on this one. To me the direction - or context - is more important than the value. In this case the context is the `infoButton` and it's relation to others, ie we're positioning the `infoButton` in relation to the label rather than the opposite.

A better example is when we have a single view constrained to a superview, with margins. To me the subview is the thing being positioned, so the direction should be subview anchors -> superview anchors. It makes little sense constraining the superview to the subview.

I know that the Apple docs suggest switching the order to eliminate negative values, so I'm probably not in the majority here. It's fine to change it according to your suggestion, but I thought I'd give my perspective on it at least. :)
</blockquote></details>

Apple also says that "You should choose the solution that most clearly describes your intent".
And that being consistent is better than being right. 


My point of view, is that we think "left to right".
However, reading that constraint, we take a leading anchor (left), constrain it to a trailing anchor (right) but then we go in the opposite direction (negative constant). 

In both cases (negative or positive constant value), we are constraining two items in relation to each other, I think it's incorrect to say that we're placing `infoButton` in relation to `titleLabel` rather than the opposite, because this constraint directly affects `titleLabel` too.

Let's leave this for now as it is because ultimately it's not that important. But for future decisions, we should decide to a style and commit to it. Ideally, a style that everyone is happy with !

ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift
```diff
@@ -249,14 +267,22 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
     }
 
     func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
-        return UIMetrics.sectionSpacing
+        let sectionIdentifier = snapshot().sectionIdentifiers[section]
+
+        switch sectionIdentifier {
+        case .contentBlockers:
+            return UITableView.automaticDimension
```
mojganii, line 274: I think maybe it's better to set fixed value for header. automatic dimension caused overhead while the header is not going to be grown up dynamically

ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift
```diff
@@ -336,25 +362,30 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
...
-        var snapshot = NSDiffableDataSourceSnapshot<Section, Item>()
+        var newSnapshot = NSDiffableDataSourceSnapshot<Section, Item>()
 
-        snapshot.appendSections([.mullvadDNS, .customDNS])
-        snapshot.appendItems(
-            [.blockAdvertising, .blockTracking, .blockMalware, .blockAdultContent, .blockGambling],
-            toSection: .mullvadDNS
-        )
-        snapshot.appendItems([.useCustomDNS], toSection: .customDNS)
+        newSnapshot.appendSections([.contentBlockers, .customDNS])
```
mojganii, line 367: If I were you,I would conform CaseIterable protocol to have access all cases by .allCases.It makes much easier for us to keep future changes simpler

ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift
```diff
@@ -525,6 +556,43 @@ final class PreferencesDataSource: UITableViewDiffableDataSource<
...
+            if let self = self {
+                self.delegate?.preferencesDataSource(self, didPressInfoButton: nil)
+            }
+        }
+
+        reusableView.didCollapseHandler = { [weak self] headerView in
+            guard let self = self else { return }
+
+            var snapshot = self.snapshot()
+            let contentBlockers: [Item] = [
```
mojganii, line 577: maybe it's better to group the options based on their section.

mojganii (commented):
take a look at these comments and let me know what you think!

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 30 of 33 files reviewed, 9 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift` line 274 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#gh-1169765271:-NTJ5qZg9vRXvZ8xX9gb:b-a3pf6r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bbc54bf51c50ebf7d94c638eef59eff494701afc/ios/MullvadVPN/View%20controllers/Preferences/PreferencesDataSource.swift#L274)):*

I think maybe it's better to set fixed value for header. automatic dimension caused overhead while the header is not going to be grown up dynamically
</blockquote></details>

I see your point, but there are two things at play here. The content blocker header is a real header, while the dns settings header is not. It's actually a regular cell. So, in order to sync the heights I made the content blocker header work the same way as the cell, ie height being determined by the size of the label. 

The optimal solution might be to turn the dns header into a real header as well, but that would require some rework of the whole data source content structure.
___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift` line 367 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#gh-1169770850:-NTJ8VB97PLAAwL_hEyX:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bbc54bf51c50ebf7d94c638eef59eff494701afc/ios/MullvadVPN/View%20controllers/Preferences/PreferencesDataSource.swift#L367)):*

If I were you,I would conform CaseIterable protocol to have access all cases by .allCases.It makes much easier for us to keep future changes simpler
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift` line 577 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#gh-1169777426:-NTJ98Wd5XyNSV3Gn1gD:bk3cn80) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bbc54bf51c50ebf7d94c638eef59eff494701afc/ios/MullvadVPN/View%20controllers/Preferences/PreferencesDataSource.swift#L577)):*

maybe it's better to group the options based on their section.
</blockquote></details>

I'd rather not change how Item is structured (would require rework of data source), but I created a static var for content blockers on it.
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsCell.swift` line 123 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTDDhgV8oSVOpbcoaGf:-NTJ7uHOATdBelsorChm:b-gqdwow) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aa32847adffc5c77ba57829f0b4e853afac8d3b6/ios/MullvadVPN/View%20controllers/Settings/SettingsCell.swift#L123)):*

Apple also says that "You should choose the solution that most clearly describes your intent".
And that being consistent is better than being right. 


My point of view, is that we think "left to right".
However, reading that constraint, we take a leading anchor (left), constrain it to a trailing anchor (right) but then we go in the opposite direction (negative constant). 

In both cases (negative or positive constant value), we are constraining two items in relation to each other, I think it's incorrect to say that we're placing `infoButton` in relation to `titleLabel` rather than the opposite, because this constraint directly affects `titleLabel` too.

Let's leave this for now as it is because ultimately it's not that important. But for future decisions, we should decide to a style and commit to it. Ideally, a style that everyone is happy with !
</blockquote></details>

Agreed, we should decide on something that we all can stick to.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 29 of 33 files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NSuKDE58WfK5A4oKB_A:-NTP1sNRDs4M0FHn8Ndn:b-qntprj) (no related file):*

Many images were updated when running the asset conversion script. Only IconInfo is new.
</blockquote></details>

Are there actual changes or metadata noise? If it's metadata noise then I'd only commit new asset and discaard the rest.

___
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsCell.swift` line 123 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NSyfvvpD18PywS0dl9G:-NTP2TpY0w5PeI0c6365:bi12mcd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bcce1731cc5d4c7893b7e9d98506019f9dca02c7/ios/MullvadVPN/View%20controllers/Settings/SettingsCell.swift#L123)):*

Perhaps a more correct way would be to let the button be native size and then adjust the touch area (`point(inside...)`), but simply moving the button to the left and "over" the title works pretty well too. Suggestions?
</blockquote></details>

Similar to other instance we have here. As long as it does not break the UI it's fine to keep it this way for now.
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsCell.swift` line 123 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTDDhgV8oSVOpbcoaGf:-NTOxr1Q4R-7tMgwY8Id:bnrw8wl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aa32847adffc5c77ba57829f0b4e853afac8d3b6/ios/MullvadVPN/View%20controllers/Settings/SettingsCell.swift#L123)):*

Agreed, we should decide on something that we all can stick to.
</blockquote></details>

Personally, I flip first and second views to avoid negative values but I wouldn't be blocking this PR because of this. It's also possible to add some sort of helper to our builder to pin edges of two individual views, which then would flip it for us to eliminate negative values.

Same could be applied to  `pinEdges()` family of methods in autolayout builder which currently produce negative constants. It's probably easier to debug UI when all values are positive.
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsContentBlockersHeaderView.swift` line 84 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NSyg-pjDVuXTGYMIak5:-NTOzyw07VuqadNhefZ2:bdeady5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bcce1731cc5d4c7893b7e9d98506019f9dca02c7/ios/MullvadVPN/View%20controllers/Settings/SettingsContentBlockersHeaderView.swift#L84)):*

Perhaps a more correct way would be to let the button be native size and then adjust the touch area (`point(inside...)`), but simply moving the button to the left and "over" the title works pretty well too. Suggestions?
</blockquote></details>

You could also implement `func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView?`, but that requires a bit of math and coordinate conversion. I am just not sure it's needed since you fixed that issue with constraints, but I can imagine it may become an issue with introduction of Dynamic Type. Maybe leave it as is for now?

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 29 of 33 files reviewed, 9 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPN/View controllers/Settings/SettingsContentBlockersHeaderView.swift` line 124 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTP8H0h6WrKxO3oKZiI:-NTP8H0i51Ski1tsoXdZ:bru5tym) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/49b1bf5a161c0868edbb9ffe50feb494ae41890a/ios/MullvadVPN/View%20controllers/Settings/SettingsContentBlockersHeaderView.swift#L124)):*
> ```Swift
>             ? NSLocalizedString(
>                 "SELECT_LOCATION_COLLAPSE_ACCESSIBILITY_ACTION",
>                 tableName: "SelectLocation",
> ```

Accessibility labels are copied from somewhere else but not updated to reflect that we're expanding a list of content blockers and not "locations"

pronebird (commented):
> Reviewed 2 of 32 files at r2, 1 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 32 of 33 files reviewed, 9 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 8 of 33 files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @mojganii, and @pronebird)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NSuKDE58WfK5A4oKB_A:-NTP5R1D2uNck2zWKRpl:bhotci7) (no related file):*

Are there actual changes or metadata noise? If it's metadata noise then I'd only commit new asset and discaard the rest.
</blockquote></details>

I think is was discussed in chat that this is just noise. Will discard the rest.

___
___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesCellFactory.swift` line 110 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTDUe9k0jUvcRaokYMC:-NTSWVGbBELaB_IJkfch:b44tpx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aa32847adffc5c77ba57829f0b4e853afac8d3b6/ios/MullvadVPN/View%20controllers/Preferences/PreferencesCellFactory.swift#L110)):*

For future reference:

We could introduce a model and let the cell configure itself based on provided data. It's a bit more work but it eliminates the need to configure each cell within a huge switch+case. It also eliminates quite a bit of code repetition. For example:

```swift
struct ContentBlockerItemConfiguration {
  var title: String // titleLabel
  var isOn: Bool
  var accessibilityHint: String?
  // etc...
}

extension SettingsSwitchCell {
  func update(from model: ContentBlockerItemConfiguration) {
      titleLabel.text = model.title
      // etc..
  }
}
```
</blockquote></details>

👍
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsContentBlockersHeaderView.swift` line 84 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NSyg-pjDVuXTGYMIak5:-NTP4t-IDZHqZZnwLnj6:b-gtfq9f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bcce1731cc5d4c7893b7e9d98506019f9dca02c7/ios/MullvadVPN/View%20controllers/Settings/SettingsContentBlockersHeaderView.swift#L84)):*

You could also implement `func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView?`, but that requires a bit of math and coordinate conversion. I am just not sure it's needed since you fixed that issue with constraints, but I can imagine it may become an issue with introduction of Dynamic Type. Maybe leave it as is for now?
</blockquote></details>

Yeah, let's see if we need to change it later. Then we could create a helper for it or something.
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsContentBlockersHeaderView.swift` line 124 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTP8H0h6WrKxO3oKZiI:-NTSXDnXAAJEVCoXhBrR:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/49b1bf5a161c0868edbb9ffe50feb494ae41890a/ios/MullvadVPN/View%20controllers/Settings/SettingsContentBlockersHeaderView.swift#L124)):*

Accessibility labels are copied from somewhere else but not updated to reflect that we're expanding a list of content blockers and not "locations"
</blockquote></details>

Done.



_Code quote:_
```Swift
ELECT_LOCATION_COLLAPSE_ACCESSIBILITY_ACTIO
```

pronebird (commented):
> Reviewed 2 of 32 files at r2, 13 of 26 files at r10.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 20 of 33 files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/CHANGELOG.md` line 35 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTTU7zC7ggu2EQryzzl:-NTTU7zC7ggu2EQryzzm:b-gj1d5m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c64f89ba382740b20c532564bad7c62e95346bce/ios/CHANGELOG.md#L35)):*
> ```Markdown
>   propagation to relays.
> ```

Nit: We do have two newlines between unreleased and released blocks. So maybe just discard changes you have made to reduce the diff

pronebird (commented):
> Reviewed 1 of 1 files at r5, 2 of 26 files at r10.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: 22 of 33 files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

pronebird (approved):
:lgtm:

> Reviewed 4 of 5 files at r11.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-:-NTYViXd1L0lgVsivbu8:bnfp4nl)* status: 23 of 34 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @mojganii)

buggmagnet (approved):
> Reviewed 10 of 26 files at r10, 1 of 5 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: all files reviewed, 7 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift` line 66 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTmXbgZ5vp4WcSof2pk:-NTmXbgZ5vp4WcSof2pl:bx6kz6r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7f165c053e65ab30f800ba1c9647e8b87f7b5ec3/ios/MullvadVPN/View%20controllers/Preferences/PreferencesDataSource.swift#L66)):*
> ```Swift
>         case dnsServer(_ uniqueID: UUID)
> 
>         static var contentBlockers: [Item] {
> ```

I think that instead of doing this, we should reuse `DNSBlockingOptions`if possible
This way, whenever we modify those settings, we only need to change one place in the codebase

It would also likely simplify `PreferencesViewModel` and all the settings related code

But I think we should do this in a different pull request, not this one.
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsContentBlockersHeaderView.swift` line 18 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTm1Pj27FdllTvk0k-G:-NTm1Pj27FdllTvk0k-H:b6ykoj0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7f165c053e65ab30f800ba1c9647e8b87f7b5ec3/ios/MullvadVPN/View%20controllers/Settings/SettingsContentBlockersHeaderView.swift#L18)):*
> ```Swift
>         let titleLabel = UILabel()
>         titleLabel.translatesAutoresizingMaskIntoConstraints = false
>         titleLabel.font = .systemFont(ofSize: 17)
> ```

I counted 22 instances of `.systemFont(ofSize: 17)` in the project so far.
I propose that we have a constant in the `UIMetrics` class instead. 
As it is, I'm guessing `17` is usually what we want for a title label, but the search results show that sometimes, it's a text label, or a detail text label
with this font size. So the value `17` is ambiguous here, is it for titles, something else ?

We should probably engage the UX team as well to define that. 

No necessary need to change that, but let's schedule some sync in the near future so we can define that.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: all files reviewed, 7 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift` line 66 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTmXbgZ5vp4WcSof2pk:-NTmauQX4wHjfw_XSh9V:by0a7g6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7f165c053e65ab30f800ba1c9647e8b87f7b5ec3/ios/MullvadVPN/View%20controllers/Preferences/PreferencesDataSource.swift#L66)):*

I think that instead of doing this, we should reuse `DNSBlockingOptions`if possible
This way, whenever we modify those settings, we only need to change one place in the codebase

It would also likely simplify `PreferencesViewModel` and all the settings related code

But I think we should do this in a different pull request, not this one.
</blockquote></details>

The diffable data source currently expects all rows to be of type `Item`, so either we'd have to make `DNSBlockingOptions` conform to some protocol that can be shared with the other items in the table (`useCustomDNS`, `addDNSServer`, `dnsServer`), or introduce some sort of mapping. But as you said, this should be explored in another PR.
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsContentBlockersHeaderView.swift` line 18 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTm1Pj27FdllTvk0k-G:-NTm_c_Q96z-3wQ1_UkU:b-th7k5t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7f165c053e65ab30f800ba1c9647e8b87f7b5ec3/ios/MullvadVPN/View%20controllers/Settings/SettingsContentBlockersHeaderView.swift#L18)):*

I counted 22 instances of `.systemFont(ofSize: 17)` in the project so far.
I propose that we have a constant in the `UIMetrics` class instead. 
As it is, I'm guessing `17` is usually what we want for a title label, but the search results show that sometimes, it's a text label, or a detail text label
with this font size. So the value `17` is ambiguous here, is it for titles, something else ?

We should probably engage the UX team as well to define that. 

No necessary need to change that, but let's schedule some sync in the near future so we can define that.
</blockquote></details>

This will be taken care of in another PR handling dynamic type (https://linear.app/mullvad/issue/IOS-64/use-dynamic-type).

In dynamic type, 17 is the size for the `.body` text style.

buggmagnet (approved):
> Reviewed 1 of 1 files at r7, 14 of 26 files at r10, 4 of 5 files at r11.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: all files reviewed, 7 unresolved discussions (waiting on @mojganii)

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: all files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPN/Supporting Files/Assets.xcassets/IconInfo.imageset/IconInfo.pdf` line 0 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTs33T480QZnQeq7bdH:-NTs33T480QZnQeq7bdI:b560rc2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/80b8a3f36a86c9290d67e50386b78555af5107cb/ios/MullvadVPN/Supporting%20Files/Assets.xcassets/IconInfo.imageset/IconInfo.pdf#L0)):*
Re-run the export script please.
___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift` line 274 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#gh-1169765271:-NTs22Z288hyq10GeEiZ:buiv43z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bbc54bf51c50ebf7d94c638eef59eff494701afc/ios/MullvadVPN/View%20controllers/Preferences/PreferencesDataSource.swift#L274)):*

I see your point, but there are two things at play here. The content blocker header is a real header, while the dns settings header is not. It's actually a regular cell. So, in order to sync the heights I made the content blocker header work the same way as the cell, ie height being determined by the size of the label. 

The optimal solution might be to turn the dns header into a real header as well, but that would require some rework of the whole data source content structure.
</blockquote></details>

I don't think we can have sections with zero items. So can we please turn the header into individual cell?
___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift` line 66 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTmXbgZ5vp4WcSof2pk:-NTs0-NcERnoqowFlNHw:b5gdo2p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/80b8a3f36a86c9290d67e50386b78555af5107cb/ios/MullvadVPN/View%20controllers/Preferences/PreferencesDataSource.swift#L66)):*

The diffable data source currently expects all rows to be of type `Item`, so either we'd have to make `DNSBlockingOptions` conform to some protocol that can be shared with the other items in the table (`useCustomDNS`, `addDNSServer`, `dnsServer`), or introduce some sort of mapping. But as you said, this should be explored in another PR.
</blockquote></details>

`DNSBlockingOptions` is an `OptionSet` so I don't think it will make it easier. 
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsContentBlockersHeaderView.swift` line 18 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-NTm1Pj27FdllTvk0k-G:-NTs-dZREQAyIV5t_dxZ:b-uik7rz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/80b8a3f36a86c9290d67e50386b78555af5107cb/ios/MullvadVPN/View%20controllers/Settings/SettingsContentBlockersHeaderView.swift#L18)):*

This will be taken care of in another PR handling dynamic type (https://linear.app/mullvad/issue/IOS-64/use-dynamic-type).

In dynamic type, 17 is the size for the `.body` text style.
</blockquote></details>

As Jon said. Dynamic type will take care of this. It's imminent and we don't need to do work that will be wiped weeks to months later.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: all files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPN/View controllers/Preferences/PreferencesDataSource.swift` line 274 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#gh-1169765271:-NTs4V8B1bioAwxzHyxn:b-9udkck) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bbc54bf51c50ebf7d94c638eef59eff494701afc/ios/MullvadVPN/View%20controllers/Preferences/PreferencesDataSource.swift#L274)):*

I don't think we can have sections with zero items. So can we please turn the header into individual cell?
</blockquote></details>

Scratch that.

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532#-:-NTs7cxm2i0_2T16yMIo:bnfp4nl)* status: all files reviewed, 8 unresolved discussions (waiting on @mojganii and @rablador)

pronebird (commented):
> Reviewed 2 of 2 files at r12.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4532)* status: all files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

PR #4560, opened 2023-04-14
Work around fragmentation issues during PSK exchange with exit node

dlon:
This PR prevents packets from being fragmented in the tunnel when exchanging the PSK with the exit endpoint by setting a maximum segment size for the tunnel config client TCP socket. Connecting when both multihop and PQ are enabled should be faster in general. The timeouts for the PSK exchange were also made more pessimistic, since the payload is fairly large.


Jontified (approved):
> Reviewed 4 of 5 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4560)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4563, opened 2023-04-14
Prevent close notification close handler from running when closing by category

raksooo:
Sometimes the notification dot doesn't disappear when it should. I haven't been able to reproduce it consistently but I think this will solve the issue. The problem was that the close handler added the notification to the dismissed notifications when it should just have been removed from all lists.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4563)


hankolsen (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4563)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4565, opened 2023-04-14
Prevent using system-wide proxy settings in URLSession requests

pronebird:
URLSession configuration is initialized with option that enable it to use system wide proxy configuration. We should disable that to prevent users from proxying our API requests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4565)

### DIFF
--- a/ios/CHANGELOG.md
+++ b/ios/CHANGELOG.md
@@ -31,6 +31,9 @@ Line wrap the file at 100 chars.                                              Th
 - Delay tunnel reconnection after a WireGuard private key rotates. Accounts for latency in key
   propagation to relays.
 
+### Fixed
+- Configure URLSession to ignore system proxy configuration.
+
 
 ## [2023.1] - 2023-03-21
 ### Added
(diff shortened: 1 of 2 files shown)


pinkisemils:
We should hold off on making such a change until we've evaluated if this is a real issue. Even though users can use a proxy to intercept our traffic, I'm not sure if it's a real issue.

pinkisemils:
Is it possible to infer if a user is using a global proxy setting? If so, we should think about how to make use of this. One user might not be aware that their requests are being proxied, whereas another user might depend on it - ideally we'd be able to notice that there is a system wide proxy and then stop using shadowsocks. This isn't anything we should focus on just now, we can do this after our shadowsocks changes have been released.

pronebird:
@pinkisemils probably via CFNetwork, we could fetch system proxy settings via `CFNetworkCopySystemProxySettings()` or `CFNetworkCopyProxiesForURL()` which also query which system proxy settings will be used to access a given URL.

PR #4593, opened 2023-04-19
Enable pointer events on the new view when transitioning

raksooo:
Currently we prevent pointer events on both views when transitioning. This causes some issues on macOS where scrolling is disabled until you stop scrolling if you started scrolling on a view with `pointer-events: none`. This means that scrolling will be disabled until the user stops scrolling. I've solved this by enabling pointer events on the new view since I can only see reasons to prevent it for the previous view.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4593)

### DIFF
--- a/gui/src/renderer/components/TransitionContainer.tsx
+++ b/gui/src/renderer/components/TransitionContainer.tsx
@@ -39,17 +39,11 @@ interface IState {
   nextItemTransition?: Partial<IItemStyle>;
 }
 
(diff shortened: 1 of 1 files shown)


hankolsen (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4593)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4602, opened 2023-04-21
Ensure all keychain operations are done on a background queue

rablador:
We run operations that will block the main thread during app startup, which means we can get killed by the startup watch dog.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4602)

### DIFF
--- a/ios/MullvadVPN/SettingsManager/KeychainSettingsStore.swift
+++ b/ios/MullvadVPN/SettingsManager/KeychainSettingsStore.swift
@@ -13,6 +13,7 @@ import Security
 class KeychainSettingsStore: SettingsStore {
     let serviceName: String
     let accessGroup: String
+    let dispatchQueue = DispatchQueue(label: "KeychainSettingsStore.dispatchqueue")
 
     init(serviceName: String, accessGroup: String) {
         self.serviceName = serviceName
@@ -20,15 +21,27 @@ class KeychainSettingsStore: SettingsStore {
     }
 
     func read(key: SettingsKey) throws -> Data {
-        try readItemData(key)
+        try readDataInBackground { [weak self] in
+            guard let self = self else { throw KeychainError.itemNotFound }
+
+            return try self.readItemData(key)
+        }
     }
 
     func write(_ data: Data, for key: SettingsKey) throws {
-        try addOrUpdateItem(key, data: data)
+        try mutateDataInBackground { [weak self] in
+            guard let self = self else { return }
+
+            return try self.addOrUpdateItem(key, data: data)
+        }
     }
 
     func delete(key: SettingsKey) throws {
-        try deleteItem(key)
+        try mutateDataInBackground { [weak self] in
+            guard let self = self else { return }
+
+            return try self.deleteItem(key)
+        }
     }
 
     private func addItem(_ item: SettingsKey, data: Data) throws {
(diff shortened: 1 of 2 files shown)


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4602)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @rablador)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4602#-NTYzFteC_K58k_ES1h4:-NTYzFteC_K58k_ES1h5:behi2y1) (no related file):*
Is there any need for that right now? A lot of calls to settings manager are done on background queues anyway and `asyncAndWait` would simply block current thread while executing work on background thread. So if you call that method from main thread it's basically the same thing as calling it from main thread except now main thread waits for secondary thread to do the work. Please explore documentation for documented C counterpart of this Swift function: https://developer.apple.com/documentation/dispatch/3191901-dispatch_async_and_wait

___

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4602)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @pronebird)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4602#-NTYzFteC_K58k_ES1h4:-NTZCsKYBbOxzxku_ziW:b-fjfsqq) (no related file):*

Is there any need for that right now? A lot of calls to settings manager are done on background queues anyway and `asyncAndWait` would simply block current thread while executing work on background thread. So if you call that method from main thread it's basically the same thing as calling it from main thread except now main thread waits for secondary thread to do the work. Please explore documentation for documented C counterpart of this Swift function: https://developer.apple.com/documentation/dispatch/3191901-dispatch_async_and_wait
</blockquote></details>

I picked it up since the task has medium prio in Linear. Wouldn't say it's very critical though.

As for `asyncAndWait`, I clearly misinterpreted a conversation we had here earlier today. I was under the impression that sending the execution to a separate thread would - while still blocking - offload the main thread. I didn't want to introduce an asynchronous call chain for reading/writing settings, so this is where it ended up, but I guess this does very little for our cause.

___

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4602)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @pronebird)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4602#-NTYzFteC_K58k_ES1h4:-NTeVuuU8eyW7LdzMI84:btkpmcr) (no related file):*

I picked it up since the task has medium prio in Linear. Wouldn't say it's very critical though.

As for `asyncAndWait`, I clearly misinterpreted a conversation we had here earlier today. I was under the impression that sending the execution to a separate thread would - while still blocking - offload the main thread. I didn't want to introduce an asynchronous call chain for reading/writing settings, so this is where it ended up, but I guess this does very little for our cause.
</blockquote></details>

Anyway, let's scrap that. Although the task specifies all keychain operations, it originates from the startup flow with possibly multiple delete operations (`resetStore`). If we just want to alleviate that, can't we let `getWipeSettingsOperation` run on a background queue and call it a day?

___

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4602)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @pronebird)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4602#-NTYzFteC_K58k_ES1h4:-NTnC_0TAF2IMIPW6FWl:bjragt3) (no related file):*

`getWipeSettingsOperation` creates aa block operation does not receive any `dispatchQueue` in constructor, which means that operation it will automatically create a new background queue and execute everything on it You can test it by putting a breakpoint inside the body and see the execution context. 

Tunnel manager interacts with the settings store from background queue for the most part, so I think we should be using keychain from background almost everywhere if not entirely everywhere. I haven't verified each and every call but I assume tunnel manager is the primary consumer.
</blockquote></details>

Ah, yes, the AsyncOperation init does indeed create a new queue. I guess it's not so bad to let each caller to decide for themselves which queue to access settings from, rather than forcing it from `KeychainSettingsStore`. Should I close this issue then?

___

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4602)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @pronebird)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4602#-NTYzFteC_K58k_ES1h4:-NTwYCAACjzkPAE8BylP:b-4pkxh4) (no related file):*

I think this should be closed. There are two approaches that can be taken, either force call site to ensure execution context, or force the object itself to define execution context, usually by dispatching blocks on internal queue, which in turn causes the whole API to become async, meaning more completion handlers and indirection. In most cases it's probably easier to run things on a background queue and work with the class in a synchronous fashion.
</blockquote></details>

Agreed, will close it now.

___

PR #4605, opened 2023-04-21
Backport block based handler for UIBarButtonItem

pronebird:
Quality of life improvement for developers.  This PR adds `UIBarButtonItem` with block handlers on iOS. Keeping it draft because I am not entirely sure we need it nor do I have much time to spread too thin on other things right now. Sadly block based (`UIAction`) handlers are only available since iOS 14. So this is somewhat an attempt to ease pain of using Objective-C selectors in the meanwhile.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4605)

### DIFF
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -119,6 +119,7 @@
 		586168692976F6BD00EF8598 /* DisplayError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 586168682976F6BD00EF8598 /* DisplayError.swift */; };
 		586250BB29E6F8F300F4B521 /* OperationCancellationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 586250BA29E6F8F300F4B521 /* OperationCancellationTests.swift */; };
 		5862805422428EF100F5A6E1 /* TranslucentButtonBlurView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5862805322428EF100F5A6E1 /* TranslucentButtonBlurView.swift */; };
+		5864211F29F04CED00822139 /* UIBarButtonItem+Blocks.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5864211E29F04CED00822139 /* UIBarButtonItem+Blocks.swift */; };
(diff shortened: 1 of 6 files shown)


rablador (commented):
Tested and works well 👍

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4605#-:-NT_PwI_6YYirCQmLwjq:bhzdx1t)* status: :shipit: complete! all files reviewed, all discussions resolved

buggmagnet (changes requested):
Check out this new feature called [@backdeployed](https://github.com/apple/swift-evolution/blob/main/proposals/0376-function-back-deployment.md) coming in Swift 5.8.

With it, you can make new feature retroactively available on older OSes.

It looks like it's exactly what we need to solve this, and I would suggest using it if we can instead of adding more dependencies on the ObjC runtime.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4605#-:-NToR_Cb41-Sd_-0V_3R:bnlbngo)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (commented):
@buggmagnet  that's great but that won't work for a few reasons:

1. Our CI will not be able to compile our app, unless we switch to the latest Xcode. Swift 5.8 is available in Xcode 14.3.
2. The following code snippet, while it looks great and easy to use is impossible to implement because `UIAction`s do not publicly expose the action handler (see https://gist.github.com/paulz/91af10b2655d4b9d1ac34495e2ab357d).
3. There is nowhere where we can put this `UIMenu` either since there is no support for that in iOS before 14.

```swift
extension UIBarButtonItem {
    @backDeployed(before: iOS 14.0)
    public convenience init(
        systemItem: UIBarButtonItem.SystemItem,
        primaryAction: UIAction? = nil,
        menu: UIMenu? = nil
    ) {
        /* ??? */
    }
}
```

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4605#-:-NTrTyxyC-nkxWrTLwrC:b-wr2s7b)* status: :shipit: complete! all files reviewed, all discussions resolved

mojganii (approved):

buggmagnet (approved):
Fair enough, thanks for looking into that deeper than I did. I want to take more time to look into that in the future, but I don't want to block this either. :shipit:   !

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4605#-:-NTxgQQh0zGqU2CsCFuu:b-gp9bjf)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4607, opened 2023-04-24
Use shadowsocks proxies to reach the API

pinkisemils:
This is not intended to be merged to `main`, but we do want to make a release with these changes - it'd be great to get these changes sanity checked for review.

Instead of passing a proxy config to URLSession to get it to use our proxy, instead we've chosen to have a port-forwarded proxy running locally and have a custom transport that rewrites request URLs when in use. To get something out sooner, rather than later, the changes currently would always use a proxy and start a new proxy per request. This is intentional, but ultimately not the desired behavior, so we'll be looking to address that later.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)


pronebird (commented):
> Reviewed 24 of 24 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 24 files reviewed, all discussions resolved (waiting on @buggmagnet)

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 24 files reviewed, 2 unresolved discussions (waiting on @buggmagnet, @pinkisemils, and @rablador)

___
*[`ios/MullvadVPN/TransportMonitor/TransportMonitor.swift` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmLQ9Q7OVadhxhl6ls:-NTmLQ9RAOd76DE0aIAt:b-29y7ys) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadVPN/TransportMonitor/TransportMonitor.swift#L32)):*
> ```Swift
>         let shadowSocksBridgeRelay = cachedRelays.relays.bridge.relays.randomElement()!
> 
>         let shadowSocksURLSession = REST.makeURLSession()
> ```

Create URLSession in `init()` and then use it or share the same with `URLSessionTransport`, `urlSession` is public property on `URLSessionTransport` so you can use it. Creating entirely new URLSession for each request is probably very expensive given that session runs a task queue internally.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 24 files reviewed, 2 unresolved discussions (waiting on @buggmagnet, @pinkisemils, @pronebird, and @rablador)

___
*[`ios/MullvadREST/URLSessionTransport.swift` line 74 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmJvOK0cmhL3UeVPVF:-NTmMwgE1c6ahn79sERk:bpji1cl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadREST/URLSessionTransport.swift#L74)):*

Is it ok to start proxy on request rather than on init?
</blockquote></details>

Nevermind the accidental code quote...

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 24 files reviewed, 3 unresolved discussions (waiting on @buggmagnet, @pinkisemils, @pronebird, and @rablador)

___
*[`ios/MullvadREST/URLSessionTransport.swift` line 74 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmJvOK0cmhL3UeVPVF:-NTmT1mg6X5nvDlxHOCF:b44tpx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadREST/URLSessionTransport.swift#L74)):*

The SS transport is only used once since it's being created each time, so I didn't bother to comment on that, but yes, ideally we want it to start in `init()`
</blockquote></details>

👍

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 24 files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @pinkisemils, and @rablador)

___
*[`ios/MullvadREST/HttpProxy.swift` line 2 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmvyvM9kvO3T5uY_UM:-NTmvyvM9kvO3T5uY_UN:bgif22e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadREST/HttpProxy.swift#L2)):*
> ```Swift
> //
> //  ShadowSocksProxy.swift
> ```

Nit: Maybe rename the file to `ShadowSocksProxy.swift` since it has the class named `ShadowSocksProxy` in it

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 24 files reviewed, 5 unresolved discussions (waiting on @pinkisemils, @pronebird, and @rablador)

___
*[`ios/MullvadREST/URLSessionTransport.swift` line 38 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmhEGR7rLnpO5zYk3L:-NTmxxOC83Nd27Nt4NFi:b-fjzy1t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadREST/URLSessionTransport.swift#L38)):*

Can you please give it a name that reflects that we're using shadowsock. This is reflected in app logs and would help in debugging.
</blockquote></details>

That was definitely an oversight on my part, thanks for catching that !
___
*[`ios/MullvadREST/URLSessionTransport.swift` line 77 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmOOPp3LIlL7cyOTzG:-NTmZTKk9BltyRMJle5H:bfn24ea) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadREST/URLSessionTransport.swift#L77)):*

`URLRequest` in Swift is a value type so to clone it you can simply do the following to clone it and you should be able to mutate it by defining it as aa variable with `var`:

```
var clonedRequest = request
```
</blockquote></details>

You're absolutely right, I'm not sure anymore why I went through all that hassle.
___
*[`ios/MullvadREST/URLSessionTransport.swift` line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmU0rU36oVc-_hhMB2:-NTmcajTCnN8bO55st2t:b7bdyxw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadREST/URLSessionTransport.swift#L80)):*

A lot of force unwrapping here. I'd simply use `flatMap`, i.e:

```swift
var rewrittenURLRequest = request
rewrittenURLRequest.url = request.url.flatMap { url in
    var components = URLComponents(url: url, resolvingAgainstBaseURL: false)
    components?.host = "127.0.0.1"
    components?.port = Int(shadowSocksProxy.localPort())
    return components?.url
}
```

I don't anticipate any of calls here to fail, but using optional chaining here helps us to avoid panics in case if errors do happen. Worst case URLSession with throw badURL error or something like that but it should not crash the app.

</blockquote></details>

That's a great suggestion, thanks ! 
___
*[`ios/MullvadVPN/TransportMonitor/TransportMonitor.swift` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmLQ9Q7OVadhxhl6ls:-NTmwt0W0g2RmeK9AdAn:b14z6i2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadVPN/TransportMonitor/TransportMonitor.swift#L32)):*

Create URLSession in `init()` and then use it or share the same with `URLSessionTransport`, `urlSession` is public property on `URLSessionTransport` so you can use it. Creating entirely new URLSession for each request is probably very expensive given that session runs a task queue internally.
</blockquote></details>

The reason we went with this design was because we needed to change the bridge relay every now and then. 
You're right that it's an expensive operation, and it is not the final design intended. 

There is definitely a better way to reuse the session that uses the shadow socks proxy instance, I just didn't get around it yet. 
I will take some time to look into that and come with a better solution in the near future.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 24 files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @pinkisemils, and @rablador)

___
*[`ios/MullvadREST/BridgingHeader.h` line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmxlld2M2bVeLlQTCX:-NTmxlld2M2bVeLlQTCY:btog06w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadREST/BridgingHeader.h#L15)):*
> ```objc
> #endif /* BridgingHeader_h */
> 
> #include "shadowsocks-proxy/include/shadowsocks.h"
> ```

I'd replace that with private module map since bridging headers aren't supported in frameworks as per: https://developer.apple.com/documentation/swift/importing-objective-c-into-swift#Import-Code-Within-a-Framework-Target

See this commit that attempts to leverage modulemap to bridge shadowsocks_proxy:

https://github.com/mullvad/mullvadvpn-app/commit/7446d3ae2f7d841a1831e7bde6fa37d1e7dada95#diff-f29dfe1a0ce589ae099027f56ea82375f45099dbb7203cf78a6110dc340e6a22

buggmagnet (commented):
> Reviewed 1 of 3 files at r2, 2 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 3 of 25 files reviewed, 6 unresolved discussions (waiting on @pinkisemils, @pronebird, and @rablador)

___
*[`ios/MullvadREST/BridgingHeader.h` line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmxlld2M2bVeLlQTCX:-NTn2y1pEPNsNfbkHXWp:bvflrdd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadREST/BridgingHeader.h#L15)):*

I'd replace that with private module map since bridging headers aren't supported in frameworks as per: https://developer.apple.com/documentation/swift/importing-objective-c-into-swift#Import-Code-Within-a-Framework-Target

See this commit that attempts to leverage modulemap to bridge shadowsocks_proxy:

https://github.com/mullvad/mullvadvpn-app/commit/7446d3ae2f7d841a1831e7bde6fa37d1e7dada95#diff-f29dfe1a0ce589ae099027f56ea82375f45099dbb7203cf78a6110dc340e6a22
</blockquote></details>

This was discussed offline, and we cherry picked the suggestion into this branch

pronebird (commented):
> Reviewed 2 of 3 files at r2, 2 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 3 of 25 files reviewed, 1 unresolved discussion (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/TransportMonitor/TransportMonitor.swift` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmLQ9Q7OVadhxhl6ls:-NTn1lGpCE7hF5DfRUkG:bbyjbqx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadVPN/TransportMonitor/TransportMonitor.swift#L32)):*

The reason we went with this design was because we needed to change the bridge relay every now and then. 
You're right that it's an expensive operation, and it is not the final design intended. 

There is definitely a better way to reuse the session that uses the shadow socks proxy instance, I just didn't get around it yet. 
I will take some time to look into that and come with a better solution in the near future.
</blockquote></details>

I understand the intention in regards of creating new transport each time. That's expected, but we should reuse existing url session when creating new shadow socks transport.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 3 of 25 files reviewed, 1 unresolved discussion (waiting on @buggmagnet and @rablador)

pronebird (commented):
> Reviewed 1 of 4 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 3 of 25 files reviewed, 1 unresolved discussion (waiting on @buggmagnet and @rablador)

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 3 of 25 files reviewed, 1 unresolved discussion (waiting on @buggmagnet, @pinkisemils, and @rablador)

___
*[`ios/MullvadREST/URLSessionTransport.swift` line 74 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmJvOK0cmhL3UeVPVF:-NTnNhMT5kP14LyCp2Ey:bjv8lxx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadREST/URLSessionTransport.swift#L74)):*

👍
</blockquote></details>

Oh boy, I realized you asked the opposite of what I thought. We should start proxy in `init()` to avoid force unwrapping `shadowSocksProxy`. This way even if you create a transport and never send any requests, it should release `shadowSocksProxy` on `deinit`. However in practice we always use transport after creating it. In the long run, keeping this transport immutable is also good for thread safety as if in the future we decide to share transports, then we don't have to use thread synchronization to access `shadowSocksProxy`.

rablador (commented):
> Reviewed 3 of 12 files at r6, 7 of 10 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 26 files reviewed, 2 unresolved discussions (waiting on @buggmagnet, @pinkisemils, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/TransportMonitor/TransportMonitor.swift` line 28 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NUQswt36dFsMFZS_9_K:-NUQswt36dFsMFZS_9_L:bsk41j6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a74711f02f3799ea84c00147aa19239fe3aefb4/ios/MullvadVPN/TransportMonitor/TransportMonitor.swift#L28)):*
> ```Swift
>         urlSessionTransport = REST.URLSessionTransport(urlSession: REST.makeURLSession())
>     }
>     
> ```

Trailing spaces.

rablador (commented):
> Reviewed 1 of 10 files at r8, 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 26 files reviewed, 2 unresolved discussions (waiting on @buggmagnet, @pinkisemils, @pronebird, and @rablador)

pronebird (commented):
> Reviewed 1 of 10 files at r8.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 26 files reviewed, 2 unresolved discussions (waiting on @buggmagnet, @pinkisemils, and @rablador)

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 26 files reviewed, 2 unresolved discussions (waiting on @pinkisemils, @pronebird, and @rablador)

___
*[`ios/MullvadREST/URLSessionTransport.swift` line 60 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTnSP2c4EvhgGJE23Js:-NU_bT6D6L9DgEqIDQ6Y:b-6eju9c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/149b678f1da4c82283282f825671397c054faa1e/ios/MullvadREST/URLSessionTransport.swift#L60)):*

I concur, let's encapsulate the shadowsocks resource in it's own class to ensure it doesn't leak.
</blockquote></details>

I don't think that's a good idea. 
Relying on the `deinit` to run code is just asking for trouble.
That means you get undefined behaviour as soon as you get retain cycles. 
I'm well aware of the irony of this statement as it's precisely what I did here.

@pinkisemils `ShadowSocksProxy` is already encapsulated in its own class. 

The problem here is that we don't have a well defined way to call `stop` on Shadowsocks, that's why I tied it 
to the lifetime of `URLSessionShadowSocksTransport`.

Because we are using the model that a `RESTTransport` object is used for each request that is sent, 
this is more or less guaranteed to run. Moving this logic in the `deinit` of `ShadowsocksProxy` instead would just move the problem to a different place, not solve it.

One way of solving this would be to use the `URLSessionTaskDelegate` protocol, instead of using the block based API as we currently do, but this is a big refactoring that wouldn't add benefits.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 26 files reviewed, 2 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadREST/URLSessionTransport.swift` line 60 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTnSP2c4EvhgGJE23Js:-NU_ykWuBp8gvyXmMZGx:b-cnufv8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/149b678f1da4c82283282f825671397c054faa1e/ios/MullvadREST/URLSessionTransport.swift#L60)):*

I don't think that's a good idea. 
Relying on the `deinit` to run code is just asking for trouble.
That means you get undefined behaviour as soon as you get retain cycles. 
I'm well aware of the irony of this statement as it's precisely what I did here.

@pinkisemils `ShadowSocksProxy` is already encapsulated in its own class. 

The problem here is that we don't have a well defined way to call `stop` on Shadowsocks, that's why I tied it 
to the lifetime of `URLSessionShadowSocksTransport`.

Because we are using the model that a `RESTTransport` object is used for each request that is sent, 
this is more or less guaranteed to run. Moving this logic in the `deinit` of `ShadowsocksProxy` instead would just move the problem to a different place, not solve it.

One way of solving this would be to use the `URLSessionTaskDelegate` protocol, instead of using the block based API as we currently do, but this is a big refactoring that wouldn't add benefits.
</blockquote></details>

Currently, we already rely on `deinit` here. I think using `deinit` here is the right call, but it should be encapsulated in `ShadowSocksProxy` instead of here.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 26 files reviewed, 1 unresolved discussion (waiting on @buggmagnet, @pronebird, and @rablador)

rablador (commented):
> Reviewed 17 of 20 files at r12, 5 of 6 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 43 files reviewed, 2 unresolved discussions (waiting on @buggmagnet, @pinkisemils, @pronebird, and @rablador)

rablador (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-:-NV8sjQ84LxTVinQA5Wt:bnfp4nl)* status: 0 of 43 files reviewed, 2 unresolved discussions (waiting on @buggmagnet, @pinkisemils, @pronebird, and @rablador)

pronebird (commented):
> Reviewed 1 of 2 files at r11, 16 of 20 files at r12, 5 of 6 files at r13.
> Dismissed @pinkisemils from a discussion.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 43 files reviewed, 1 unresolved discussion (waiting on @buggmagnet, @pinkisemils, and @rablador)

___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 33 at r13](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NV96NYS1zjM5uv-TTUr:-NV96NYS1zjM5uv-TTUs:bhbmc35) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fcf9fe117f12461c7024717adc42aeaafec772fe/ios/MullvadRESTTests/TransportStrategyTests.swift#L33)):*
> ```Swift
> 
>         // Serialize the strategy and reload it from memory to simulate an application restart
>         let encodedRawStrategy = try JSONEncoder().encode(strategy)
> ```

FYI  `Codable` is auto-synthesized for `TransportStrategy` since we don't provide custom implementation. I don't know how much value is in that really, but then we also test how mod operator works, so perhaps I shouldn't question these tests at all?
___
*[`ios/MullvadTypes/Cancellable.swift` line 20 at r13](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NV91ADVCuD4tpQXUSQO:-NV91ADWAt7NCtNc-doL:b7x2f9d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fcf9fe117f12461c7024717adc42aeaafec772fe/ios/MullvadTypes/Cancellable.swift#L20)):*
> ```Swift
> ///
> /// Can be used to implement the concept of `no-op` or clean up tasks when an action has been cancelled.
> public struct AnyCancellable: Cancellable {
> ```

FYI  This needs to be rebased from `main` because the type was already introduced before.

pronebird (commented):
> Reviewed 1 of 20 files at r12.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 43 files reviewed, 1 unresolved discussion (waiting on @buggmagnet, @pinkisemils, and @rablador)

rablador (approved):
> Reviewed 1 of 28 files at r14, 4 of 10 files at r16, 23 of 24 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 43 files reviewed, 1 unresolved discussion (waiting on @buggmagnet, @pinkisemils, @pronebird, and @rablador)

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 43 files reviewed, 1 unresolved discussion (waiting on @buggmagnet, @pinkisemils, @pronebird, and @rablador)

rablador (approved):
> Reviewed 1 of 27 files at r18, 4 of 10 files at r20, 22 of 23 files at r21, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 43 files reviewed, 1 unresolved discussion (waiting on @buggmagnet, @pinkisemils, @pronebird, and @rablador)

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 43 files reviewed, 1 unresolved discussion (waiting on @pronebird and @rablador)

___
*[`ios/MullvadREST/URLSessionTransport.swift` line 60 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTnSP2c4EvhgGJE23Js:-NVY6FSs9_G8DTIkylnw:bmkwdk2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/149b678f1da4c82283282f825671397c054faa1e/ios/MullvadREST/URLSessionTransport.swift#L60)):*

There are currently two problems with this implementation:

1. Based on my understanding after talking to @pinkisemils , `ShadowSocksProxy` would create two SS sockets if you call `start()` twice on it which would lead to resource socket and memory leak.
2. `ShadowSocksProxy` - allocates the resource but does not free it when it goes away.

So in my opinion `ShadowSocksProxy` should take a better care of resources that it allocates and guard against invalid usage that would lead to leaks. 

`init()` and `deinit()` are two things that are guaranteed and `ShadowSocksProxy` should implement wrapper in one of the following ways:

1. Use RAII approach where `init` allocates a resource and `deinit` frees it.
2. Use thread-safe approach to prevent double call to `start()` or use RAAII approach where init starts proxy and deinit stops it (see point number 1). Right now `URLSessionShadowSocksTransport` is created for each request but that may not be the case a year from now, so ideally this code should somewhat handle that nicely..

> Because we are using the model that a `RESTTransport` object is used for each request that is sent,  
this is more or less guaranteed to run. Moving this logic in the `deinit` of `ShadowsocksProxy` instead would just move the problem to a different place, not solve it.
> 
> One way of solving this would be to use the `URLSessionTaskDelegate` protocol, instead of using the block based API as we currently do, but this is a big refactoring that wouldn't add benefits.

I don't know what `URLSessionTaskDelegate` has to do with any of this. The transport is retained for the duration of the task and `ShadowSocksProxy` is retained by transport. Who calls deinit does not matter all that much except it's nice when the object managing the resource does that instead of delegating it to something else. If I were you , I'd just change  `ShadowSocksProxy` to RAII approach and have absolutely thread safe object to work with:

```swift
final class ShadowSocksProxy {
   init(...) {
       start()
   }
   deinit {
       stop()
   }
   var localPort: UInt16 { .. }
}
```
</blockquote></details>

ShadowSocksProxy is now a state lock to make `start` and `stop` idempotent, and is calling `stop` in its `deinit` method.
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 33 at r13](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NV96NYS1zjM5uv-TTUr:-NVY6zvq16AeDSN2HBh5:b-jjxd1m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fcf9fe117f12461c7024717adc42aeaafec772fe/ios/MullvadRESTTests/TransportStrategyTests.swift#L33)):*

FYI  `Codable` is auto-synthesized for `TransportStrategy` since we don't provide custom implementation. I don't know how much value is in that really, but then we also test how mod operator works, so perhaps I shouldn't question these tests at all?
</blockquote></details>

I believe this discussion has been solved offline with @pinkisemils.
___
*[`ios/MullvadVPN/AppDelegate.swift` line 12 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NUXBa2q987JmJgJjn4w:-NVYEV208DtUsow33r5z:b-ci6h52) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0a74711f02f3799ea84c00147aa19239fe3aefb4/ios/MullvadVPN/AppDelegate.swift#L12)):*

Added by mistake?
</blockquote></details>

Must've automatically imported it when I was testing the waters and forgot about it, thanks for catching that !
___
*[`ios/MullvadVPN/TransportMonitor/TransportMonitor.swift` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NTmLQ9Q7OVadhxhl6ls:-NVYEbTG6EaOEGHGEiz5:bp6eyi2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1b4e4e8bfa434fbbdb0b37123156a93d4d888819/ios/MullvadVPN/TransportMonitor/TransportMonitor.swift#L32)):*

I understand the intention in regards of creating new transport each time. That's expected, but we should reuse existing url session when creating new shadow socks transport.
</blockquote></details>

We reuse the existing URLSession now
___
*[`ios/MullvadTypes/Cancellable.swift` line 20 at r13](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NV91ADVCuD4tpQXUSQO:-NVYEjp_AcYIQcv8j3Qw:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fcf9fe117f12461c7024717adc42aeaafec772fe/ios/MullvadTypes/Cancellable.swift#L20)):*

FYI  This needs to be rebased from `main` because the type was already introduced before.
</blockquote></details>

Done

rablador (approved):
> Reviewed 1 of 1 files at r22, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 43 files reviewed, 1 unresolved discussion (waiting on @buggmagnet, @pronebird, and @rablador)

rablador (approved):
> Reviewed 6 of 6 files at r23, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 45 files reviewed, 1 unresolved discussion (waiting on @buggmagnet, @pronebird, and @rablador)

pronebird (commented):
> Reviewed 4 of 11 files at r16, 10 of 23 files at r17, 3 of 23 files at r21.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 45 files reviewed, all discussions resolved (waiting on @buggmagnet and @rablador)

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-:-NVZxlDJ1XXRzEebbb8-:bnfp4nl)* status: 0 of 45 files reviewed, all discussions resolved (waiting on @buggmagnet and @rablador)

pronebird (approved):
> Reviewed 1 of 23 files at r17, 1 of 28 files at r18, 3 of 11 files at r20, 17 of 23 files at r21, 6 of 6 files at r23, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 45 files reviewed, all discussions resolved (waiting on @buggmagnet, @pronebird, and @rablador)

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 45 files reviewed, all discussions resolved (waiting on @buggmagnet, @pronebird, and @rablador)

dlon (approved):
> Reviewed 3 of 3 files at r24, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 45 files reviewed, 3 unresolved discussions (waiting on @buggmagnet, @dlon, @pinkisemils, @pronebird, and @rablador)

___
*[`ios/MullvadREST/shadowsocks-proxy/Cargo.toml` line 11 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NV_PwhvE4wRR9WYEslt:-NV_PwhvE4wRR9WYEslu:b-eho7dn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/Cargo.toml#L11)):*
> ```TOML
> 
> [dependencies]
> shadowsocks-service = { git = "https://github.com/mullvad/shadowsocks-rust", rev = "8f6afd081a9440fff2dda565908eddc27a3295f1",  features = [ "local", "stream-cipher", "local-http", "local-tunnel" ] }
> ```

These changes should be upstreamed.
___
*[`ios/MullvadREST/shadowsocks-proxy/src/ffi.rs` line 24 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaLiby-wdADGymxTAP:-NVaLiby-wdADGymxTAQ:b-7yjqlm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/src/ffi.rs#L24)):*
> ```XML
> #[no_mangle]
> pub unsafe extern "C" fn start_shadowsocks_proxy(
>     addr: *const u8,
> ```

It would probably be more readable/conventional to pass in something like `*const sockaddr_in` here (instead of `addr`, `addr_len`, and `port`). You'd be able to parse that with `socket2::SockAddr`.

It might not be worth doing just to make the interface slightly nicer.
___
*[`ios/MullvadREST/shadowsocks-proxy/src/ffi.rs` line 27 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaLMmW1-fc0iIPQ-yB:-NVaLMmW1-fc0iIPQ-yC:b-opttco) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/src/ffi.rs#L27)):*
> ```XML
>     addr_len: usize,
>     port: u16,
>     password: *const u8,
> ```

Nit: This could just be a null-terminated string. `CStr::from_ptr(password).to_str()`
___
*[`ios/MullvadREST/shadowsocks-proxy/src/ffi.rs` line 29 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaSJ4r3Y2e1TV3I1wr:-NVaSJ4r3Y2e1TV3I1ws:b54r34n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/src/ffi.rs#L29)):*
> ```XML
>     password: *const u8,
>     password_len: usize,
>     cipher: *const u8,
> ```

Nit: This could just be a null-terminated string.
___
*[`ios/MullvadREST/shadowsocks-proxy/src/ffi.rs` line 32 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaIvYy0SGA2atxAEeD:-NVaIvYy0SGA2atxAEeE:b722qg4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/src/ffi.rs#L32)):*
> ```XML
>     cipher_len: usize,
>     proxy_config: *mut ProxyHandle,
> ) -> i32 {
> ```

Nit: This seems awkward compared to just returning a `*mut ProxyHandle` on success (and null on error), given that there are no specific error codes.
___
*[`ios/MullvadREST/shadowsocks-proxy/src/ffi.rs` line 126 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaWGYBENWBHuGmAGtH:-NVaWGYBENWBHuGmAGtI:bw18hxf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/src/ffi.rs#L126)):*
> ```XML
> unsafe fn parse_str(data: *const u8, len: usize) -> Option<String> {
>     // SAFETY: data pointer must be valid for the size of len
>     let bytes = unsafe { std::slice::from_raw_parts(data, len) };
> ```

The inner `unsafe` is superfluous.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 45 files reviewed, 3 unresolved discussions (waiting on @buggmagnet, @dlon, @pinkisemils, @pronebird, and @rablador)

___
*[`Cargo.toml` line 5 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaeHJnE2lW8my7wDz-:-NVaeHJnE2lW8my7wDz0:b5k1jih) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/Cargo.toml#L5)):*
> ```TOML
> members = [
>     "android/translations-converter",
>     "ios/MullvadREST/shadowsocks-proxy",
> ```

Is there any way to make this not part of the same workspace?

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 47 files reviewed, 3 unresolved discussions (waiting on @buggmagnet, @dlon, @pronebird, and @rablador)

___
*[`Cargo.toml` line 5 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaeHJnE2lW8my7wDz-:-NVcq51YBjzDHfRGt-tt:bnkhck6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/Cargo.toml#L5)):*

Is there any way to make this not part of the same workspace?
</blockquote></details>

We want it to be a part of this workspace.
___
*[`ios/MullvadREST/shadowsocks-proxy/Cargo.toml` line 11 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NV_PwhvE4wRR9WYEslt:-NVcq7vBBNHOh-RUCGiv:b70ipke) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/Cargo.toml#L11)):*

These changes should be upstreamed.
</blockquote></details>

They have been upstreamed just now, but I do also believe that they broke the build. Let's update this dependency once it's released.
___
*[`ios/MullvadREST/shadowsocks-proxy/src/ffi.rs` line 24 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaLiby-wdADGymxTAP:-NVcuamb-aqryqdhTx13:b-x97lec) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/src/ffi.rs#L24)):*

It would probably be more readable/conventional to pass in something like `*const sockaddr_in` here (instead of `addr`, `addr_len`, and `port`). You'd be able to parse that with `socket2::SockAddr`.

It might not be worth doing just to make the interface slightly nicer.
</blockquote></details>

It's easier to pass bytes like this than it is to create a `sockaddr` on the Swift side, AFAIK.
___
*[`ios/MullvadREST/shadowsocks-proxy/src/ffi.rs` line 27 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaLMmW1-fc0iIPQ-yB:-NVcqDgz5NH1q2wg18RL:bmmzu8x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/src/ffi.rs#L27)):*

Nit: This could just be a null-terminated string. `CStr::from_ptr(password).to_str()`
</blockquote></details>

It's swift, so it's easier to not have a C-string, as then the string needs

dlon (approved):
> Reviewed 6 of 10 files at r25, 8 of 8 files at r26, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 48 files reviewed, 1 unresolved discussion (waiting on @buggmagnet, @pinkisemils, @pronebird, and @rablador)

___
*[`ios/MullvadREST/shadowsocks-proxy/src/ffi.rs` line 24 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaLiby-wdADGymxTAP:-NVq0_WV4evE0d9qa2Bt:b-4glgam) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/src/ffi.rs#L24)):*

It's easier to pass bytes like this than it is to create a `sockaddr` on the Swift side, AFAIK.
</blockquote></details>

You could also define a struct that contains a `family`, a `port`, and a `union { ipv4: [u8; 4], ipv6: [u8; 8] }`.

...I don't mean to nitpick too hard, especially if it just complicates things on the Swift side. Feel free to ignore this.
___
*[`ios/MullvadREST/shadowsocks-proxy/src/ffi.rs` line 27 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaLMmW1-fc0iIPQ-yB:-NVq12Xg5-1PPFK9J9Z9:bq7pojg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/src/ffi.rs#L27)):*

It's swift, so it's easier to not have a C-string, as then the string needs 
</blockquote></details>

Found this but I don't know if it's usable: https://developer.apple.com/documentation/swift/string/utf8cstring

Feel free to ignore this, again, since it's not important.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 48 files reviewed, 1 unresolved discussion (waiting on @dlon, @pinkisemils, @pronebird, and @rablador)

___
*[`ios/MullvadREST/shadowsocks-proxy/src/ffi.rs` line 29 at r24](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607#-NVaSJ4r3Y2e1TV3I1wr:-NW1IE_m-t3jrKeP85ez:b-q6ie0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5d1120d3acf0bcbf6db64cf2f1cb6ea814576fbc/ios/MullvadREST/shadowsocks-proxy/src/ffi.rs#L29)):*

Nit: This could just be a null-terminated string.
</blockquote></details>

@dlon  When you pass a swift `String` object to rust, it should be [automatically converted to a null-terminated string](https://developer.apple.com/documentation/swift/calling-functions-with-pointer-parameters)

dlon (approved):
> Reviewed 1 of 1 files at r27, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4607)* status: 0 of 48 files reviewed, 1 unresolved discussion (waiting on @buggmagnet, @pinkisemils, @pronebird, and @rablador)

PR #4611, opened 2023-04-24
Fix errors returned by Clippy on Windows

dlon:
That and a few warnings that remained when I ran it using the msvc toolchain. I've also added Windows to the workflow for Clippy.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4611)

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -9,7 +9,10 @@ on:
   workflow_dispatch:
 jobs:
   clippy_check:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
@@ -29,11 +32,13 @@ jobs:
           override: true
 
       - name: Install build dependencies
+        if: matrix.os == 'ubuntu-latest'
         run: |
           sudo apt-get update
           sudo apt-get install libdbus-1-dev
 
       - name: Clippy check
+        shell: bash
         run: |-
           export RUSTFLAGS="--deny warnings"
           source env.sh
--- a/mullvad-cli/src/cmds/split_tunnel/windows.rs
+++ b/mullvad-cli/src/cmds/split_tunnel/windows.rs
@@ -21,7 +21,7 @@ impl Command for SplitTunnel {
                     .arg(
                         clap::Arg::new("policy")
                             .required(true)
-                            .possible_values(&["on", "off"]),
+                            .possible_values(["on", "off"]),
                     ),
             )
             .subcommand(clap::App::new("get").about("Display the split tunnel status"))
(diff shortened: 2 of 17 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4611)* status: 0 of 17 files reviewed, all discussions resolved

___
*[`talpid-core/src/split_tunnel/windows/driver.rs` line 935 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4611#-NTnnDX79DTt0TEVdjHE:-NTnnDX79DTt0TEVdjHF:b-ssxn0m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c826091de45ae9af387472d24ecebb7ad5600fad/talpid-core/src/split_tunnel/windows/driver.rs#L935)):*
> ```XML
>         INFINITE,
>     );
>     let signaled_index = if result >= WAIT_OBJECT_0 && result < WAIT_OBJECT_0 + objects_len {
> ```

This is considered an error because `WAIT_OBJECT_0 == 0`. `objects_len` is unsigned, so of course it will always be at least 0.

I'm not sure if I agree with the lint, since the actual value of the constant could be considered an implementation detail. That said, it will in reality always be defined as 0, so I've changed it.
___
*[`talpid-core/src/split_tunnel/windows/path_monitor.rs` line 536 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4611#-NTnnk0V0O0-f39RC7Bd:-NTnnk0V0O0-f39RC7Be:b-l9ss89) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c826091de45ae9af387472d24ecebb7ad5600fad/talpid-core/src/split_tunnel/windows/path_monitor.rs#L536)):*
> ```XML
>         cmd_rx: &sync_mpsc::Receiver<PathMonitorCommand>,
>     ) -> bool {
>         while let Some(cmd) = cmd_rx.try_iter().next() {
> ```

This is considered an error because all cases return from the function. It is impossible for it to actually loop. I think the idea was to handle all commands in the "queue" before retrieving the next I/O packet. It should work without the loop, since every command in `PathMonitorHandle` sends an I/O packet anyway. So I've just made it explicitly not loop.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4611)* status: 16 of 17 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`talpid-core/src/split_tunnel/windows/path_monitor.rs` line 532 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4611#-NTrEF4m5Sq9mVgnwK3y:-NTrRVz2-LHea08wA7tm:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c2d8cbdf478811277bfcbe610ac1b564220f5873/talpid-core/src/split_tunnel/windows/path_monitor.rs#L532)):*

Should probably be renamed to `process_command` or similar when it's a single command only.
</blockquote></details>

Done.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4611)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4614, opened 2023-04-25
Fix DPI in asset conversion script IOS-135

pronebird:
- Fix DPI at 72 to produce assets of the exact same size as it's defined in SVG files.
- Refactor `system()` call to `rsvg_convert()` to avoid repetition.
- Remove `retina_scale_suffix()` that is no longer used.
- Regenerate all assets. This should auto-fix those assets generated recently that have smaller size than expected (i.e 18px vs. 24px) and should also fix timezone which should be using UTC instead of local timezone, which was a bug in rsvg-convert a little while ago.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4614)


rablador (approved):
> Reviewed 24 of 24 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4614)* status: :shipit: complete! all files reviewed, all discussions resolved

buggmagnet (approved):
:lgtm:
Great work ! 

> Reviewed 24 of 24 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4614#-:-NTrrova5upvOtXztgN4:blqxyci)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4619, opened 2023-04-26
Spawn shadowsocks for ios 114

buggmagnet:
I would like to get early feedback on this tentative implementation for having shadow socks proxied requests in the packet tunnel

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)


buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 0 of 13 files reviewed, 2 unresolved discussions

___
*[`ios/PacketTunnel/TunnelTransportProvider.swift` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTw__Py-bgGTJCdbNP7:-NTw__Py-bgGTJCdbNP8:b-q0mfg4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/PacketTunnel/TunnelTransportProvider.swift#L30)):*
> ```Swift
>     }
> 
>     func selectNextTransport() {
> ```

Most of this code was copy / pasted from `TransportMonitor` 
I want to find a way to refactor this to have a shared implementation for creating a shadow socks transport layer and avoid duplication
___
*[`ios/TunnelProviderMessaging/ProxyURLRequest.swift` line 18 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwa2P28Fg6adZl7P6d:-NTwa2P28Fg6adZl7P6e:b-ik85qk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/ProxyURLRequest.swift#L18)):*
> ```Swift
>     public let httpBody: Data?
>     public let httpHeaders: [String: String]?
>     public let useAlternativeTransport: Bool
> ```

Right now, if this is ever set, both the UI process and the network extensions are going to use Shadow Socks. This is not the intended final behavior, but this should work for the moment.

pronebird (changes requested):
> Reviewed 4 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 4 of 15 files reviewed, 3 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift` line 33 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwc7K85NZFmubAH1pd:-NTwc7K97i09cj9ED-dl:b-zel34y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b25214aa4e9556608a84b13ea16e645117d15f8/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift#L33)):*
> ```Swift
>             urlSession: REST.makeURLSession(),
>             dispatchQueue: dispatchQueue,
>             transportProvider: { nil }
> ```

FYI This will break the app on simulator, you should still supply a provider here. Since we don't have packet tunnel on simulator, this class is basically mocking it.
___
*[`ios/MullvadVPN/TransportMonitor/PacketTunnelTransport.swift` line 41 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwk-RSD_auNXHbwMj2:-NTwk-RT22ew51z-0Czc:bqbs7ow) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/MullvadVPN/TransportMonitor/PacketTunnelTransport.swift#L41)):*
> ```Swift
>         guard let proxyRequest = proxyRequest else {
>             completion(nil, nil, URLError(.badURL))
>             return Operation()
> ```

What was wrong with a throwing flow that precisely described what went wrong and booted you out of this function without allocating dummy `Operation` in absence of better way to express what happened? 

Worst case you could have opt-in for `Cancellable?` to indicate that nothing happened.

Precisely the thing that I had to battle with last year is now being reinvented again for no good reason.
___
*[`ios/TunnelProviderMessaging/ProxyURLRequest.swift` line 18 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwa2P28Fg6adZl7P6d:-NTweD008xhoyIlPbfGP:by5xs76) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/ProxyURLRequest.swift#L18)):*

Right now, if this is ever set, both the UI process and the network extensions are going to use Shadow Socks. This is not the intended final behavior, but this should work for the moment.
</blockquote></details>

If you think of transports this way, you have three of them on UI side:

1. URLSession
2. Shadowsocks
3. Packet tunnel transport

The deciding function would look like this:

```swift
func decide() -> RESTTransport? {
  if shouldUsePacketTunnel() {
      return packetTunnelTransport
  } else if shouldUseURLSession() {
      return urlSessionTransport
  } else {
      return shadowSocksTransport
  }
}
```

On the packet tunnel side it's the same except there is no packet tunnel transport.

So in that regard it would be much simpler to communicate failures back to transport and let it pick what to use next instead of trying to synchronize both processes this way.

pronebird (changes requested):
> Reviewed 2 of 13 files at r1, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 7 of 15 files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/PacketTunnel/TunnelTransportProvider.swift` line 33 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwkjdqDFTWJ3ECwszN:-NTwkjdqDFTWJ3ECwszO:bjq6ivx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/PacketTunnel/TunnelTransportProvider.swift#L33)):*
> ```Swift
>         do {
>             let cachedRelays = try relayCache.read()
>             let shadowSocksConfiguration = cachedRelays.relays.bridge.shadowsocks.filter { $0.protocol == "tcp" }
> ```

Can you move that into `RelaySelector` refactoring the function that I had added last week?

pronebird (changes requested):
> Reviewed 1 of 13 files at r1, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 9 of 15 files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/TunnelProviderMessaging/URLRequestProxy.swift` line 59 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwme9_BGE_Cyo2I_Kq:-NTwme9_BGE_Cyo2I_Kr:baoezax) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/URLRequestProxy.swift#L59)):*
> ```Swift
>                 // All tasks should have unique identifiers, but if not, cancel the task scheduled
>                 // earlier.
>                 let oldTask = self.addRequest(identifier: proxyRequest.id, task: task as! URLSessionDataTask)
> ```

Force cast

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 9 of 15 files reviewed, 7 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/TunnelProviderMessaging/URLRequestProxy.swift` line 39 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwo-TB9atky6tEY-Od:-NTwo-TB9atky6tEY-Oe:boeytya) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/URLRequestProxy.swift#L39)):*
> ```Swift
>         completionHandler: @escaping (ProxyURLResponse) -> Void
>     ) {
>         dispatchQueue.async { [weak self] in
> ```

Why does it have to be weak?
___
*[`ios/TunnelProviderMessaging/URLRequestProxy.swift` line 53 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwmwuY00dYCPaKVlYE:-NTwmwuY00dYCPaKVlYF:bb661s7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/URLRequestProxy.swift#L53)):*
> ```Swift
>                 let response = ProxyURLResponse(data: data, response: response, error: error)
> 
>                 _ = self.removeRequest(identifier: proxyRequest.id)
> ```

`removeRequest()` will be called on a queue where the completion handler is invoked which will likely cause race condition given that we mutate `proxiedRequests` here.

pronebird (commented):
> Reviewed 2 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 11 of 15 files reviewed, 7 unresolved discussions (waiting on @buggmagnet)

pronebird (commented):
> Reviewed 3 of 13 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: all files reviewed, 7 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/TunnelProviderMessaging/ProxyURLRequest.swift` line 18 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwa2P28Fg6adZl7P6d:-NTwqfAn9xq3cavhenDj:b-swvdhn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/ProxyURLRequest.swift#L18)):*

If you think of transports this way, you have three of them on UI side:

1. URLSession
2. Shadowsocks
3. Packet tunnel transport

The deciding function would look like this:

```swift
func decide() -> RESTTransport? {
  if shouldUsePacketTunnel() {
      return packetTunnelTransport
  } else if shouldUseURLSession() {
      return urlSessionTransport
  } else {
      return shadowSocksTransport
  }
}
```

On the packet tunnel side it's the same except there is no packet tunnel transport.

So in that regard it would be much simpler to communicate failures back to transport and let it pick what to use next instead of trying to synchronize both processes this way.
</blockquote></details>

Actually for the packet tunnel side, as discussed later today. We can keep things the same way, i.e simple proxy.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: all files reviewed, 8 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/TunnelProviderMessaging/URLRequestProxy.swift` line 44 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwsR5O9BFd3MkLxn8k:-NTwsR5O9BFd3MkLxn8l:bv3fj7e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b25214aa4e9556608a84b13ea16e645117d15f8/ios/TunnelProviderMessaging/URLRequestProxy.swift#L44)):*
> ```Swift
>             // Instruct the Packet Tunnel to try to reach the API via the local shadow socks proxy instance if needed
>             let transportProvider = transportProvider()
>             if proxyRequest.useAlternativeTransport {
> ```

I understand why we have transport provider here, but it would be easier to understand if we had something like that:

```swift
let transport = proxyReqest.useShadowsocks ? makeShadowsocksTranasport() : urlSessionTransport
````

That would avoid calling `selectNextTransport()` and ultimately enable UI to decide which transport it wants.

**Review by pronebird — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 8 of 16 files reviewed, 8 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/TransportMonitor/PacketTunnelTransport.swift` line 41 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwk-RSD_auNXHbwMj2:-NTxdGdR3HIoXTtC7xuz:beburg1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/MullvadVPN/TransportMonitor/PacketTunnelTransport.swift#L41)):*
> What does a successful completion handler mean if the call threw ?
> Or if the completion handler has an error set but it also threw ?

It means that the transport couldn't initiate the request. Similar to how provider messaging works on iOS except they don't return errors in completion handler because the protocol is different, they simply swallow the completion handler and forget about it:

```swift
func sendProviderMessage(
    _ messageData: Data,
    responseHandler: ((Data?) -> Void)? = nil
) throws
```

https://developer.apple.com/documentation/networkextension/netunnelprovidersession/1406409-sendprovidermessage 

If you want to keep it this way I'd suggest to introduce `AnyCancellable` and return it with empty block so that we do not create dummy `Operation`, something along these lines:

```swift
public final class AnyCancellable: Cancellable {
    private let block: () -> Void

    public init(block: @escaping () -> Void) {
        self.block = block
    }

    public func cancel() {
        block()
    }
}
```

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 9 of 17 files reviewed, 6 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/PacketTunnel/TunnelTransportProvider.swift` line 33 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwkjdqDFTWJ3ECwszN:-NTxjBq9CWeyBFWQzB0_:bf239oz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/PacketTunnel/TunnelTransportProvider.swift#L33)):*

I think I missed that, I will look into it, thanks for the reminder
</blockquote></details>

Amazing. Btw, if you want you can perhaps combine two of them into one function that returns a tuple:

```swift
let (ssConfig, ssRelay) = RelaySelector.getShadowSocksSomething(relays: cachedRelays.relays)
```
___
*[`ios/TunnelProviderMessaging/URLRequestProxy.swift` line 53 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwmwuY00dYCPaKVlYE:-NTxkYiC9I3YEJ83O0vD:b-u5rxoi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/URLRequestProxy.swift#L53)):*

I was wondering about this, `URLSession` uses an internal dispatch queue, which should schedule properly back to our own dispatch queue.
I've modified the code so that this shouldn't be a problem anymore
</blockquote></details>

I don't like that `removeRequest()` makes assumption about its execution context. This breaks isolation. Also `URLSession` runs on its own queue and not our own AFAIK. 

The code has been reshuffled with no gain and I don't understand why we have weak reference when using dispatch queue. Please revert and add additional logic associated with transport, the rest was perfectly fine.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 9 of 17 files reviewed, 6 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/TunnelProviderMessaging/URLRequestProxy.swift` line 39 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwo-TB9atky6tEY-Od:-NTxnrAM0cuidyb1zK0l:b4z4xnn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/URLRequestProxy.swift#L39)):*

We need explicit capture semantics because `transportProvider()` is an `@escaping` closure.
That being said, I forgot to re weakify `self` in the below request.
</blockquote></details>

We don't need weak ref inside `dispatchQueue`. We only need it inside completion handler to transport as this one may be executing for a relatively extended time.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 9 of 17 files reviewed, 6 unresolved discussions (waiting on @pronebird)

___
*[`ios/TunnelProviderMessaging/URLRequestProxy.swift` line 53 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwmwuY00dYCPaKVlYE:-NU0RKRy4_tpX6zGg8nW:b1fvn7c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/URLRequestProxy.swift#L53)):*

I don't like that `removeRequest()` makes assumption about its execution context. This breaks isolation. Also `URLSession` runs on its own queue and not our own AFAIK. 

The code has been reshuffled with no gain and I don't understand why we have weak reference when using dispatch queue. Please revert and add additional logic associated with transport, the rest was perfectly fine.
</blockquote></details>

Fair enough. It's just that the original code had 2 calls to `dispatchQueue.async` which makes the execution hard to follow.
As you pointed out, `URLSession` uses an internal dispatch queue, which means that the first call to `dispatchQueue.async` was unnecessary. 

The weak reference in the dispatch queue was an error based on my wrong recollection of memory semantics in this case, apologies for that.
___
*[`ios/TunnelProviderMessaging/URLRequestProxy.swift` line 44 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwsR5O9BFd3MkLxn8k:-NU0T8amCfmktHtdYX1s:bo4cs3x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b25214aa4e9556608a84b13ea16e645117d15f8/ios/TunnelProviderMessaging/URLRequestProxy.swift#L44)):*

I understand why we have transport provider here, but it would be easier to understand if we had something like that:

```swift
let transport = proxyReqest.useShadowsocks ? makeShadowsocksTranasport() : urlSessionTransport
````

That would avoid calling `selectNextTransport()` and ultimately enable UI to decide which transport it wants.
</blockquote></details>

I like that idea, I'll see how to integrate it.
However, I will integrate this idea in IOS-133 as this is a larger refactoring than needed for this PR

**Review by buggmagnet — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 9 of 17 files reviewed, 6 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN/TransportMonitor/PacketTunnelTransport.swift` line 41 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwk-RSD_auNXHbwMj2:-NU0Unih9MR3bSlxPkbt:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/MullvadVPN/TransportMonitor/PacketTunnelTransport.swift#L41)):*

> What does a successful completion handler mean if the call threw ?
> Or if the completion handler has an error set but it also threw ?

It means that the transport couldn't initiate the request. Similar to how provider messaging works on iOS except they don't return errors in completion handler because the protocol is different, they simply swallow the completion handler and forget about it:

```swift
func sendProviderMessage(
    _ messageData: Data,
    responseHandler: ((Data?) -> Void)? = nil
) throws
```

https://developer.apple.com/documentation/networkextension/netunnelprovidersession/1406409-sendprovidermessage 

If you want to keep it this way I'd suggest to introduce `AnyCancellable` and return it with empty block so that we do not create dummy `Operation`, something along these lines:

```swift
public final class AnyCancellable: Cancellable {
    private let block: () -> Void

    public init(block: @escaping () -> Void) {
        self.block = block
    }

    public func cancel() {
        block()
    }
}
```
</blockquote></details>

Done.
___
*[`ios/PacketTunnel/TunnelTransportProvider.swift` line 33 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwkjdqDFTWJ3ECwszN:-NU0bcnZ0ApJz8eC0GFp:b-3isml) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/PacketTunnel/TunnelTransportProvider.swift#L33)):*

Amazing. Btw, if you want you can perhaps combine two of them into one function that returns a tuple:

```swift
let (ssConfig, ssRelay) = RelaySelector.getShadowSocksSomething(relays: cachedRelays.relays)
```
</blockquote></details>

In general I'm not a big fan of tuples, unless it makes a piece of code easier to reason about. 
I think there's little benefit here. 
However, I'm thinking of having a static method on `URLSessionShadowSocksTransport` to allow for the easy creation of a shadow socks transport instance, given an instance of a `RelayCache`
However,  I can't do that because that would intertwine dependencies between `RelayCache` and `MullvadREST` 

I'll mull over it and try to come up with a better alternative
___
*[`ios/TunnelProviderMessaging/URLRequestProxy.swift` line 39 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwo-TB9atky6tEY-Od:-NU0cQNO6Z0c6RtuQk0u:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/URLRequestProxy.swift#L39)):*

We don't need weak ref inside `dispatchQueue`. We only need it inside completion handler to transport as this one may be executing for a relatively extended time.
</blockquote></details>

Done.

pronebird (changes requested):
> Reviewed 2 of 6 files at r5, 1 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 12 of 18 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/TunnelProviderMessaging/URLRequestProxy.swift` line 53 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwmwuY00dYCPaKVlYE:-NU0vuHfCJXj7VBBjt59:b-mkrt4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/URLRequestProxy.swift#L53)):*

Fair enough. It's just that the original code had 2 calls to `dispatchQueue.async` which makes the execution hard to follow.
As you pointed out, `URLSession` uses an internal dispatch queue, which means that the first call to `dispatchQueue.async` was unnecessary. 

The weak reference in the dispatch queue was an error based on my wrong recollection of memory semantics in this case, apologies for that.
</blockquote></details>

Each public method in this class runs on a `dispatchQueue` for synchronization purposes so that we could mutate `proxiedRequests` in orderly manner without creating race conditions, in short it looks as such: 

```swift
func sendRequest() {
  dispatchQueue.async { /* do work */ }
}

func cancelRequest() {
  dispatchQueue.async { /* do work */ }
}
```

What we have now is this:

```swift
func sendRequest() {
  /* do work */
}

func cancelRequest() {
  dispatchQueue.async { /* do work */ }
}
```

This will work for as long as `sendRequest()` is invoked from the same `dispatchQueue`, which is probably the case with packet tunnel but I don't think it's the case with simulator where we use the same class to run requests but probably a different queue.

So to avoid making assumptions here, you'd probably want to stick with the original design, i.e dispatch work on the given queue in `send Request()`, so the anatomy of it would probably look as such:

```swift
func sendRequest() {
  dispatchQueue.async {
       /* prepare transport */
      transport.sendRequest(.., completion:  { [weak self] data, response, error in
          guard let self = self else { return }

          /* executing on URLSession's queue, so move to our internal queue */
          dispatchQueue.async {
              /* handle response */
          }
      })
  }
}
```

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 12 of 18 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/PacketTunnel/TunnelTransportProvider.swift` line 33 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwkjdqDFTWJ3ECwszN:-NU13ome2CySInnktQKc:b-qsmm1o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/PacketTunnel/TunnelTransportProvider.swift#L33)):*

In general I'm not a big fan of tuples, unless it makes a piece of code easier to reason about. 
I think there's little benefit here. 
However, I'm thinking of having a static method on `URLSessionShadowSocksTransport` to allow for the easy creation of a shadow socks transport instance, given an instance of a `RelayCache`
However,  I can't do that because that would intertwine dependencies between `RelayCache` and `MullvadREST` 

I'll mull over it and try to come up with a better alternative

</blockquote></details>

You can return a struct instead, no problem with that. I feel that both calls are connected so it would make sense to return both SS config and bridge relay together. But up to you. This is more of an FYI, I think it could keep things shorter on consumer side.

pronebird (commented):
> Reviewed 1 of 3 files at r4, 2 of 6 files at r5, 1 of 2 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 17 of 18 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

pronebird (commented):
> Reviewed 1 of 2 files at r11.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 17 of 18 files reviewed, 1 unresolved discussion

rablador (changes requested):
> Reviewed 7 of 13 files at r1, 1 of 2 files at r3, 2 of 3 files at r4, 4 of 6 files at r5, 2 of 2 files at r6, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/TransportMonitor/TransportMonitor.swift` line 25 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NUQucyx5KZt-aUTzxsa:-NUQucyx5KZt-aUTzxsb:b6uobkg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/12887c977d7560f577ee62fdce63eaef473ffbe7/ios/MullvadVPN/TransportMonitor/TransportMonitor.swift#L25)):*
> ```Swift
> 
>     // MARK: -
> ```

Nit: Remove MARK?
___
*[`ios/MullvadVPN/TransportMonitor/TransportMonitor.swift` line 55 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NURAwTm7ckU34prQiFw:-NURAwTm7ckU34prQiFx:bf0yhz1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/12887c977d7560f577ee62fdce63eaef473ffbe7/ios/MullvadVPN/TransportMonitor/TransportMonitor.swift#L55)):*
> ```Swift
>     }
> 
>     public func selectNextTransport() {
> ```

I think it's sometimes a little hard to follow what is generally applicable and what is specifically so. Below the generic `RESTTransportProvider` protocol provides a function for selecting a new transport, yet we always expect it to deal with shadow socks. Could `selectNewTransport` ever be intended for some other transport?
___
*[`ios/PacketTunnel/TunnelTransportProvider.swift` line 31 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NURFYuL2U579TdM2YBZ:-NURFYuL2U579TdM2YB_:bf5yg3f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/12887c977d7560f577ee62fdce63eaef473ffbe7/ios/PacketTunnel/TunnelTransportProvider.swift#L31)):*
> ```Swift
>     }
> 
>     func selectNextTransport() {
> ```

Same thoughts as with `selectNextTransport` in `TransportMonitor`.
___
*[`ios/PacketTunnel/TunnelTransportProvider.swift` line 33 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NURFi29CAHwZraeiRFY:-NURFi29CAHwZraeiRFZ:bdulb85) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/12887c977d7560f577ee62fdce63eaef473ffbe7/ios/PacketTunnel/TunnelTransportProvider.swift#L33)):*
> ```Swift
>     func selectNextTransport() {
>         do {
>             let cachedRelays = try relayCache.read()
> ```

This code block is very similar to what dynamic var `shadowSocksTransport` produces in `TransportMonitor`. Can they be combined and extracted for more efficient reuse?

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: all files reviewed, 4 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPN/TransportMonitor/TransportMonitor.swift` line 55 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NURAwTm7ckU34prQiFw:-NURIaksAeBZh6NjauG1:brgvee) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/12887c977d7560f577ee62fdce63eaef473ffbe7/ios/MullvadVPN/TransportMonitor/TransportMonitor.swift#L55)):*

I think it's sometimes a little hard to follow what is generally applicable and what is specifically so. Below the generic `RESTTransportProvider` protocol provides a function for selecting a new transport, yet we always expect it to deal with shadow socks. Could `selectNewTransport` ever be intended for some other transport?
</blockquote></details>

That part evolved a lot and is already gone from the next piece of work on Shadow Socks. 
___
*[`ios/PacketTunnel/TunnelTransportProvider.swift` line 33 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NURFi29CAHwZraeiRFY:-NURJkVkAEN0Z7-Z5xiM:b-gbg4ar) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/12887c977d7560f577ee62fdce63eaef473ffbe7/ios/PacketTunnel/TunnelTransportProvider.swift#L33)):*

This code block is very similar to what dynamic var `shadowSocksTransport` produces in `TransportMonitor`. Can they be combined and extracted for more efficient reuse?
</blockquote></details>

My first attempt ended up in circular dependencies.
I will take a shot at this when working on `Use a more optimal bridge selection algorithm`
___
*[`ios/TunnelProviderMessaging/URLRequestProxy.swift` line 53 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwmwuY00dYCPaKVlYE:-NU2Q9VIAtBJjWdlu-aN:b-nbgi3b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/TunnelProviderMessaging/URLRequestProxy.swift#L53)):*

Each public method in this class runs on a `dispatchQueue` for synchronization purposes so that we could mutate `proxiedRequests` in orderly manner without creating race conditions, in short it looks as such: 

```swift
func sendRequest() {
  dispatchQueue.async { /* do work */ }
}

func cancelRequest() {
  dispatchQueue.async { /* do work */ }
}
```

What we have now is this:

```swift
func sendRequest() {
  /* do work */
}

func cancelRequest() {
  dispatchQueue.async { /* do work */ }
}
```

This will work for as long as `sendRequest()` is invoked from the same `dispatchQueue`, which is probably the case with packet tunnel but I don't think it's the case with simulator where we use the same class to run requests but probably a different queue.

So to avoid making assumptions here, you'd probably want to stick with the original design, i.e dispatch work on the given queue in `send Request()`, so the anatomy of it would probably look as such:

```swift
func sendRequest() {
  dispatchQueue.async {
       /* prepare transport */
      transport.sendRequest(.., completion:  { [weak self] data, response, error in
          guard let self = self else { return }

          /* executing on URLSession's queue, so move to our internal queue */
          dispatchQueue.async {
              /* handle response */
          }
      })
  }
}
```
</blockquote></details>

This has been reverted to the original design.

pinkisemils (approved):
> Reviewed 7 of 13 files at r1, 1 of 2 files at r3, 2 of 3 files at r4, 4 of 6 files at r5, 2 of 2 files at r6, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: all files reviewed, 2 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPN/TransportMonitor/TransportMonitor.swift` line 25 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NUQucyx5KZt-aUTzxsa:-NUWiOWc513Pq-AKgQC2:ba6vdb2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/12887c977d7560f577ee62fdce63eaef473ffbe7/ios/MullvadVPN/TransportMonitor/TransportMonitor.swift#L25)):*

Nit: Remove MARK?
</blockquote></details>

I'd like to keep it for now if you don't mind.
___
*[`ios/PacketTunnel/TunnelTransportProvider.swift` line 33 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NTwkjdqDFTWJ3ECwszN:-NUWinam1ghpfiwRDuW8:bubt34y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5a08874d8269a7c442de6633500f55dd4226246/ios/PacketTunnel/TunnelTransportProvider.swift#L33)):*

You can return a struct instead, no problem with that. I feel that both calls are connected so it would make sense to return both SS config and bridge relay together. But up to you. This is more of an FYI, I think it could keep things shorter on consumer side.
</blockquote></details>

I think that for now, I want to keep them separate, and rework that in the later ticket `IOS-138` 
___
*[`ios/PacketTunnel/TunnelTransportProvider.swift` line 31 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NURFYuL2U579TdM2YBZ:-NUWiU8n8kJDIrCCGbE7:bokuim1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/12887c977d7560f577ee62fdce63eaef473ffbe7/ios/PacketTunnel/TunnelTransportProvider.swift#L31)):*

Same thoughts as with `selectNextTransport` in `TransportMonitor`.
</blockquote></details>

This was solved in a discussion in another PR.

pronebird (commented):
> Reviewed 1 of 2 files at r11.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

pinkisemils (commented):
> Reviewed 4 of 8 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 16 of 20 files reviewed, all discussions resolved (waiting on @pronebird and @rablador)

pronebird (commented):
> Reviewed 8 of 8 files at r12.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (changes requested):
> Reviewed 4 of 5 files at r14.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/MullvadREST/ShadowSocksProxy.swift` line 34 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NV3rEk--ZkrNkrqWtsQ:-NV3rEk--ZkrNkrqWtsR:b-56xbfm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6890a290bef55c3638148ab3cdd7e393f6d7e64a/ios/MullvadREST/ShadowSocksProxy.swift#L34)):*
> ```Swift
>     /// - Returns: The local port for the shadow socks proxy when it has started, 0 otherwise.
>     public func localPort() -> UInt16 {
>         proxyConfig.port
> ```

Has to be locked then too as `start()` mutate it.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 21 of 22 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/ShadowSocksProxy.swift` line 33 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NV3rp4A5DYrXtw-2okq:-NV3rp4A5DYrXtw-2okr:brrzs3j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6890a290bef55c3638148ab3cdd7e393f6d7e64a/ios/MullvadREST/ShadowSocksProxy.swift#L33)):*
> ```Swift
>     ///
>     /// - Returns: The local port for the shadow socks proxy when it has started, 0 otherwise.
>     public func localPort() -> UInt16 {
> ```

Nit: this could be defined as a property, i.e `var localPort: UInt16`

pronebird (commented):
> Reviewed 1 of 5 files at r14.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`ios/MullvadREST/ShadowSocksProxy.swift` line 34 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4619#-NV3rEk--ZkrNkrqWtsQ:-NV4FCvd8JXOk9xQyTp_:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6890a290bef55c3638148ab3cdd7e393f6d7e64a/ios/MullvadREST/ShadowSocksProxy.swift#L34)):*

Has to be locked then too as `start()` mutate it.
</blockquote></details>

Done.

PR #4620, opened 2023-04-26
Fix view state when disconnecting from out-of-time view

rablador:
When in the out-of-time view, the app will fail to update its view state on disconnect, blocking the user from accessing the add time button.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620)

### DIFF
--- a/ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeContentView.swift
+++ b/ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeContentView.swift
@@ -105,6 +105,13 @@ class OutOfTimeContentView: UIView {
         fatalError("init(coder:) has not been implemented")
     }
 
+    func enableDisconnectButton(_ enabled: Bool, animated: Bool) {
+        disconnectButton.isEnabled = enabled
+        UIView.animate(withDuration: animated ? 0.25 : 0) {
+            self.disconnectButton.alpha = enabled ? 1 : 0
+        }
+    }
+
     // MARK: - Private Functions
 
     func setUpSubviews() {
--- a/ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift
+++ b/ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift
@@ -7,20 +7,27 @@
 //
 
 import Foundation
+import MullvadLogging
 import MullvadREST
 import MullvadTypes
 import Operations
 import StoreKit
 
+/// Interval used for periodic polling account updates.
+private let accountUpdateTimerInterval: TimeInterval = 60
+
 final class OutOfTimeInteractor {
     private let storePaymentManager: StorePaymentManager
     private let tunnelManager: TunnelManager
 
-    var didReceivePaymentEvent: ((StorePaymentEvent) -> Void)?
-    var didReceiveTunnelStatus: ((TunnelStatus) -> Void)?
-
     private var tunnelObserver: TunnelObserver?
     private var paymentObserver: StorePaymentObserver?
+
+    private let logger = Logger(label: "OutOfTimeInteractor")
+    private var accountUpdateTimer: DispatchSourceTimer?
+
+    var didReceivePaymentEvent: ((StorePaymentEvent) -> Void)?
+    var didReceiveTunnelStatus: ((TunnelStatus) -> Void)?
 
     init(storePaymentManager: StorePaymentManager, tunnelManager: TunnelManager) {
         self.storePaymentManager = storePaymentManager
@@ -80,5 +87,22 @@ final class OutOfTimeInteractor {
             with: productIdentifiers,
             completionHandler: completionHandler
         )
+    }
+
+    func startAccountUpdateTimer() {
+        logger.debug(
+            "Start polling account updates every \(accountUpdateTimerInterval) second(s)."
+        )
+
+        let timer = DispatchSource.makeTimerSource(queue: .main)
+        timer.setEventHandler { [weak self] in
+            self?.tunnelManager.updateAccountData()
+        }
+
+        accountUpdateTimer?.cancel()
+        accountUpdateTimer = timer
+
+        timer.schedule(wallDeadline: .now() + accountUpdateTimerInterval, repeating: accountUpdateTimerInterval)
+        timer.activate()
     }
 }
--- a/ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeViewController.swift
+++ b/ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeViewController.swift
@@ -99,6 +99,7 @@ class OutOfTimeViewController: UIViewController, RootContainment {
 
         interactor.didReceiveTunnelStatus = { [weak self] tunnelStatus in
             self?.setNeedsHeaderBarStyleAppearanceUpdate()
+            self?.applyViewState()
         }
 
         if StorePaymentManager.canMakePayments {
@@ -106,6 +107,8 @@ class OutOfTimeViewController: UIViewController, RootContainment {
         } else {
             productState = .cannotMakePurchases
         }
+
+        interactor.startAccountUpdateTimer()
     }
 
     // MARK: - Private
@@ -136,8 +139,8 @@ class OutOfTimeViewController: UIViewController, RootContainment {
         purchaseButton.isEnabled = productState.isReceived && isInteractionEnabled && !tunnelState
             .isSecured
         contentView.restoreButton.isEnabled = isInteractionEnabled
-        contentView.disconnectButton.isEnabled = tunnelState.isSecured
-        contentView.disconnectButton.alpha = tunnelState.isSecured ? 1 : 0
+
+        contentView.enableDisconnectButton(tunnelState.isSecured, animated: true)
 
         if tunnelState.isSecured {
             contentView.setBodyLabelText(
@@ -341,6 +344,7 @@ class OutOfTimeViewController: UIViewController, RootContainment {
     }
 
     @objc private func handleDisconnect(_ sender: Any) {
+        contentView.disconnectButton.isEnabled = false
         interactor.stopTunnel()
     }
 }


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeViewController.swift` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU2GI3UBlrwAkcQ9zZl:-NU2GI3UBlrwAkcQ9zZm:b-8cu4a1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4271de7db101a457ba34df08f6e70eb73e79554f/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeViewController.swift#L17)):*
> ```Swift
> 
> /// Interval used for periodic polling account updates.
> private let accountUpdateTimerInterval: TimeInterval = 60
> ```

How about moving this into interactor and simply calling methods to start or stop polling account data from here?
___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeViewController.swift` line 123 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU2G8oqDZFc4WgoxV7R:-NU2G8oqDZFc4WgoxV7S:b8b3ojv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4271de7db101a457ba34df08f6e70eb73e79554f/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeViewController.swift#L123)):*
> ```Swift
> 
>     deinit {
>         cancelAccountUpdateTimer()
> ```

I think `DispatchSourceTimer` invalidates itself on deinit (unlike `Timer`)
___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeViewController.swift` line 186 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU2H1LX59xsYtUt4dfV:-NU2H1LX59xsYtUt4dfW:bt7nitw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4271de7db101a457ba34df08f6e70eb73e79554f/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeViewController.swift#L186)):*
> ```Swift
> 
>         contentView.disconnectButton.isEnabled = tunnelState.isSecured
>         UIView.animate(withDuration: 0.25) {
> ```

FYI I'd move that into `OutOfTimeContentView` instead and call a function to enable disconnect button with or without animation.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeViewController.swift` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU2GI3UBlrwAkcQ9zZl:-NU3kY-dDkVscRvkxIRn:b-j58sim) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4271de7db101a457ba34df08f6e70eb73e79554f/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeViewController.swift#L17)):*

How about moving this into interactor and simply calling methods to start or stop polling account data from here?
</blockquote></details>

It crossed my mind, but I wasn't sure if interactor should be more of a mediator of function calls rather than keeping track of things itself. It does clean things up a bit though.
___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeViewController.swift` line 123 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU2G8oqDZFc4WgoxV7R:-NU3kS8_Cz0JWxAoOKka:bb77uci) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4271de7db101a457ba34df08f6e70eb73e79554f/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeViewController.swift#L123)):*

I think `DispatchSourceTimer` invalidates itself on deinit (unlike `Timer`)
</blockquote></details>

Yep!

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620)* status: 0 of 3 files reviewed, 6 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 28 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU6AuYWECXSEnXeYCPQ:-NU6AuYWECXSEnXeYCPR:b-3uuxg7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L28)):*
> ```Swift
>     private let logger = Logger(label: "OutOfTimeInteractor")
>     private var accountUpdateTimer: DispatchSourceTimer?
>     private var isPolling = false
> ```

Nit: What are we polling here ?

The way we use this boolean is to have idempotency in `startAccountUpdateTimer` 
I think a name like `updateTimerIsRunning` would be more explicit. 
___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 94 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU66BsX-puPyx2VXpzY:-NU66BsX-puPyx2VXpzZ:b-33yhao) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L94)):*
> ```Swift
> 
>     func startAccountUpdateTimer() {
>         guard !isPolling else { return }
> ```

There are some potential race conditions here.

Say we have thread A and thread B running here this function.
L94-95: This method is not thread safe, thread A could be pre-empted at L94 before the if check, whilst thread B runs L95.
Now we have 2 dispatch source timers running.

In practice, I haven't checked how we use `OutOfTimeInteractor`, so we could be safe, but I think it's still worth being said.
___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 95 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU6BpZh8kTpX5xvQtNv:-NU6BpZh8kTpX5xvQtNw:bnr5w9c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L95)):*
> ```Swift
>     func startAccountUpdateTimer() {
>         guard !isPolling else { return }
>         isPolling = true
> ```

With this current design, we never set `isPolling` to `false` so this will run exactly once.

I'm not sure whether that's the intended design, but in that case, it sounds like we would want to cancel `accountUpdateTimer` in `stopTunnel` and set this flag to false there as well.

Please let me know if I misunderstood this. 
___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 106 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU6AKXW8Ls0VqGAmd48:-NU6AKXW8Ls0VqGAmd49:b-gqn0i4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L106)):*
> ```Swift
>         }
>         timer.schedule(wallDeadline: .now() + accountUpdateTimerInterval, repeating: accountUpdateTimerInterval)
>         timer.activate()
> ```

We have another potential race condition that's slightly more dangerous.
As per the docs, `The system may fire a timer sooner than the value in the wallDeadline parameter.`

Once again, in practice I think it's unlikely to happen as the `accountUpdateTimerInterval` is quite large (60 seconds)
If the timers were shorter however, that could have bad consequences.

As a general practice, I would advise to cancel the previous timer before activating the new one.

I would suggest moving `timer.activate()` at the end of this function, and calling `accountUpdateTimer?.cancel()` before that happens

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 94 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU66BsX-puPyx2VXpzY:-NU6mFxL8WFPbj-pBdXx:bv5rgn8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L94)):*

There are some potential race conditions here.

Say we have thread A and thread B running here this function.
L94-95: This method is not thread safe, thread A could be pre-empted at L94 before the if check, whilst thread B runs L95.
Now we have 2 dispatch source timers running.

In practice, I haven't checked how we use `OutOfTimeInteractor`, so we could be safe, but I think it's still worth being said.
</blockquote></details>

We use interactor exclusively from main thread, so there is no need to worry about any other thread in here.
___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 106 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU6AKXW8Ls0VqGAmd48:-NU6mQcu06YS-2ZhoGf0:b87nv56) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L106)):*

We have another potential race condition that's slightly more dangerous.
As per the docs, `The system may fire a timer sooner than the value in the wallDeadline parameter.`

Once again, in practice I think it's unlikely to happen as the `accountUpdateTimerInterval` is quite large (60 seconds)
If the timers were shorter however, that could have bad consequences.

As a general practice, I would advise to cancel the previous timer before activating the new one.

I would suggest moving `timer.activate()` at the end of this function, and calling `accountUpdateTimer?.cancel()` before that happens
</blockquote></details>

Both timer and this function run on main queue. So the timer will either fire before execution of this function or after. It's really a deeper discussion on how libdispatch handles cancellation and how timers work.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 28 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU6AuYWECXSEnXeYCPQ:-NU72Bzz9FwJJtzK9En4:b-pbdqz3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L28)):*

Nit: What are we polling here ?

The way we use this boolean is to have idempotency in `startAccountUpdateTimer` 
I think a name like `updateTimerIsRunning` would be more explicit. 
</blockquote></details>

Agreed.
___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 94 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU66BsX-puPyx2VXpzY:-NU6y37k0pievCwlUtdh:b-s89c7w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L94)):*

We use interactor exclusively from main thread, so there is no need to worry about any other thread in here.
</blockquote></details>

I assume no change is needed then?
___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 95 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU6BpZh8kTpX5xvQtNv:-NU733YXBrjjsLCHxgr5:b-w95xfv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L95)):*

With this current design, we never set `isPolling` to `false` so this will run exactly once.

I'm not sure whether that's the intended design, but in that case, it sounds like we would want to cancel `accountUpdateTimer` in `stopTunnel` and set this flag to false there as well.

Please let me know if I misunderstood this. 
</blockquote></details>

There was previously a function to cancel the timer in which the isPolling would change. That was removed since the timer will cancel itself when the viewcontroller (along with this class) is deinitialized. As such, a cancel function timer was no longer needed, although iPolling was not removed along with it. I'll add the cancel function back again to make things more complete (and clear) for future use.
___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 106 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU6AKXW8Ls0VqGAmd48:-NU74RMQ9oWQf5ysiO4G:bajnyn8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L106)):*

But as a general rule of thumb, I agree.
</blockquote></details>

No change, then?

buggmagnet (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 95 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU6BpZh8kTpX5xvQtNv:-NUQlMTp76fnSgR8UwgN:b-ubdbv7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L95)):*

There was previously a function to cancel the timer in which the isPolling would change. That was removed since the timer will cancel itself when the viewcontroller (along with this class) is deinitialized. As such, a cancel function timer was no longer needed, although iPolling was not removed along with it. I'll add the cancel function back again to make things more complete (and clear) for future use.
</blockquote></details>

We discussed this offline and decided to remove the `cancelAccountUpdateTimer` function as long as it's not in use
___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 106 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU6AKXW8Ls0VqGAmd48:-NUQlUl44NVIli2_EDbv:b-1g6dw8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L106)):*

No change, then?
</blockquote></details>

nit: I would personally have preferred a change just for the sake of enforcing good habits, and for the sake of consistently having them even if it doesn't matter here

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 95 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU6BpZh8kTpX5xvQtNv:-NUQrb5z7fU53qbFHIxn:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L95)):*

We discussed this offline and decided to remove the `cancelAccountUpdateTimer` function as long as it's not in use
</blockquote></details>

Done.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/View controllers/OutOfTime/OutOfTimeInteractor.swift` line 106 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-NU6AKXW8Ls0VqGAmd48:-NUXB7Y82OmMuN6LciWf:b82yoa9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c17dbf6875b613d0f6e3e4540e0fb647ffec639/ios/MullvadVPN/View%20controllers/OutOfTime/OutOfTimeInteractor.swift#L106)):*

nit: I would personally have preferred a change just for the sake of enforcing good habits, and for the sake of consistently having them even if it doesn't matter here
</blockquote></details>

Seems to be addressed

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4620#-:-NUay5gh-1HH68bXDdDm:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4623, opened 2023-04-26
Add cancellable initializer in operations IOS-151

pronebird:
1. Introduce new initializer in operations `(AsyncOperation) -> Cancellable`. Replaces previously available `addCancellationBlock()`.
2. Remove `setExecutionBlock()` on block-based operations. Pass execution block via initializer instead.
3. Remove `addCancellationBlock()` on block-based operations. Use initializer that returns `Cancellable` instead to wire up cancellation.
4. Add `AnyCancellable` object which is a block-based cancellation token. It's similar to Combine's `AnyCancellable` except it does not call `cancel()` on `deinit`. This is in part inspired by conversation with @buggmagnet today. Struct or class -- picked class to avoid copies with structs which seems reasonable in this case as this class is immutable.
5. Add more tests for operations
6. Fix bug where giving initial input to `TransformOperation` would result into `OperationError.unsatisfiedRequirement` error. Clearly shouldn't happen when input is given.
7. `Result+Extensions` were moved to `MullvadTypes` so they can be used elsewhere.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)


rablador (changes requested):
> Reviewed 15 of 19 files at r1, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)* status: all files reviewed, 3 unresolved discussions (waiting on @pronebird)

___
*[`ios/Operations/TransformOperation.swift` line 1 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NUQj3SP-jMLT9YtMn-T:-NUQj3SP-jMLT9YtMn-U:b-9cd9ig) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/Operations/TransformOperation.swift#L1)):*
> ```Swift
> //
> ```

Why do we accept optional values for `input` when `main()` fails any operation where `input` is nil?
___
*[`ios/OperationsTests/AsyncBlockOperationTests.swift` line 76 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NUQYusXEfleP73cHeLK:-NUQYusXEfleP73cHeLL:b-v3abu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncBlockOperationTests.swift#L76)):*
> ```Swift
>         operationQueue.addOperation(operation)
> 
>         waitForExpectations(timeout: 100)
> ```

100 seems a little excessive. Typo?
___
*[`ios/OperationsTests/TransformOperationTests.swift` line 52 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NUQiNaNCeehgGNfCdCA:-NUQiNaO2sf_7kRScUOp:b-k52dng) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/TransformOperationTests.swift#L52)):*
> ```Swift
>     }
> 
>     func testCancellableTaskBlockTranasformOperation() {
> ```

"Tranasform"



_Code quote:_
```Swift
Tranasform
```

buggmagnet (changes requested):
> Reviewed 15 of 19 files at r1, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)* status: all files reviewed, 6 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 862 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURfoG550aMQbKBlMN-:-NURfoG550aMQbKBlMN0:bs0f0qz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/MullvadVPN.xcodeproj/project.pbxproj#L862)):*
> ```
> 		58ACF64E26567A7100ACE4B7 /* CustomSwitchContainer.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CustomSwitchContainer.swift; sourceTree = "<group>"; };
> 		58AEEF642344A36000C9BBD5 /* KeychainError.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = KeychainError.swift; sourceTree = "<group>"; };
> 		58AFC99429F96F7B000829DE /* AsyncBlockOperationTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AsyncBlockOperationTests.swift; sourceTree = "<group>"; };
> ```

I think the files are ordered alphabetically in the Xcode project navigator, I know I've been guilty of this, but let's try to keep that order, it looks good
___
*[`ios/OperationsTests/AsyncBlockOperationTests.swift` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURnV-F3rTirqNy81gK:-NURnV-F3rTirqNy81gL:bvmcy8b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncBlockOperationTests.swift#L15)):*
> ```Swift
> final class AsyncBlockOperationTests: XCTestCase {
>     let operationQueue = AsyncOperationQueue()
> ```

nit: I want to be a bit pedantic here.
We are missing a test that checks that the `finish` method defined in `AsyncOperation` gets called when the `executionBlock` gets deallocated before the `main` from `AsyncBlockOperation` gets called.
___
*[`ios/OperationsTests/AsyncResultBlockOperationTests.swift` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURpmeg-SEbRL1wqOeu:-NURpmeg-SEbRL1wqOev:b-skps6h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncResultBlockOperationTests.swift#L24)):*
> ```Swift
> 
>         operation.onFinish { op, error in
>             XCTAssertEqual(op.result?.value, true)
> ```

I've been bitten in the past where calling `XCTAssert*` in async closures would crash the entire `xctest` process instead of failing 
Maybe it's been fixed, but instead I would suggest using an `expectation` and validating expectations at the end of the test.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)* status: 16 of 19 files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 862 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURfoG550aMQbKBlMN-:-NUVm-NlBNbtEe0-xwBP:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/MullvadVPN.xcodeproj/project.pbxproj#L862)):*

I think the files are ordered alphabetically in the Xcode project navigator, I know I've been guilty of this, but let's try to keep that order, it looks good
</blockquote></details>

Done.
___
*[`ios/Operations/TransformOperation.swift` line 1 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NUQj3SP-jMLT9YtMn-T:-NUVjcZu1rTSpnxXCH0-:b-ep8bjq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/Operations/TransformOperation.swift#L1)):*

Why do we accept optional values for `input` when `main()` fails any operation where `input` is nil?
</blockquote></details>

To support injections, i.e passing output of operation A to the input of operation B. See `OperationInputInjectionTests`
___
*[`ios/OperationsTests/AsyncBlockOperationTests.swift` line 76 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NUQYusXEfleP73cHeLK:-NUVkn1m297pCjdW2aS9:b-9fw9p9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncBlockOperationTests.swift#L76)):*

100 seems a little excessive. Typo?
</blockquote></details>

Typo . Fixed.
___
*[`ios/OperationsTests/AsyncResultBlockOperationTests.swift` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURpmeg-SEbRL1wqOeu:-NUVkt5u5HRYLyB4mXQX:b-r2mafq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncResultBlockOperationTests.swift#L24)):*

I've been bitten in the past where calling `XCTAssert*` in async closures would crash the entire `xctest` process instead of failing 
Maybe it's been fixed, but instead I would suggest using an `expectation` and validating expectations at the end of the test.
</blockquote></details>

It works as expected based on my observation.
___
*[`ios/OperationsTests/TransformOperationTests.swift` line 52 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NUQiNaNCeehgGNfCdCA:-NUVlity3LhCFK-gOrmY:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/TransformOperationTests.swift#L52)):*

"Tranasform"
</blockquote></details>

Done.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)* status: 16 of 19 files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/OperationsTests/AsyncBlockOperationTests.swift` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURnV-F3rTirqNy81gK:-NUVmxGy7jaL4fANcbYN:bvs9tgr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncBlockOperationTests.swift#L15)):*

nit: I want to be a bit pedantic here.
We are missing a test that checks that the `finish` method defined in `AsyncOperation` gets called when the `executionBlock` gets deallocated before the `main` from `AsyncBlockOperation` gets called.


</blockquote></details>

For operations that finish automatically after running a block (aka "synchronous") we have a test called `testSynchronousBlockOperation`. Based on documentation `completionBlock` would be called once `isFinished = true`, so `finish()` should be called for that to happen.

Asynchronous operations call `finish()` explicitly.

If I misunderstood your question, then would you mind to elaborate a bit more?

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)* status: 16 of 19 files reviewed, 6 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/OperationsTests/AsyncBlockOperationTests.swift` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURnV-F3rTirqNy81gK:-NUW3UHs3Yu_IenvB2Xv:b-vz3w50) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncBlockOperationTests.swift#L15)):*

For operations that finish automatically after running a block (aka "synchronous") we have a test called `testSynchronousBlockOperation`. Based on documentation `completionBlock` would be called once `isFinished = true`, so `finish()` should be called for that to happen.

Asynchronous operations call `finish()` explicitly.

If I misunderstood your question, then would you mind to elaborate a bit more?
</blockquote></details>

We have the following construct for the `main` function of `AsyncBlockOperation`
```
override public func main() {
        let block = executionBlock
        executionBlock = nil

        if let block = block {
            block(self)
        } else {
            finish()
        }
    }
```

The `else { finish() }` will never get executed as far as I can tell, it's dead code.

When constructing an `AsyncBlockOperation` we keep a strong reference to the `block` passed in the constructor.
So when unwrapping it with the `if let block = block` this will always succeed, since we have a strong reference to it.
Therefore, we always execute that block, and the else branch`finish()` is never implicitly called.

The following test proves that 
```
    func testBlockOperationDoesNotFinish() throws {
        
        let expectation = expectation(description: "test")
        let operation: AsyncBlockOperation = AsyncBlockOperation { _ in  expectation.fulfill() }
        operationQueue.addOperation(operation)
        waitForExpectations(timeout: 1)
        
        XCTAssertTrue(operation.isFinished)
        
    }
```

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)* status: 14 of 19 files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/OperationsTests/AsyncBlockOperationTests.swift` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURnV-F3rTirqNy81gK:-NUWEztTCIgCkwU5ySP6:bqovz0s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncBlockOperationTests.swift#L15)):*

We have the following construct for the `main` function of `AsyncBlockOperation`
```
override public func main() {
        let block = executionBlock
        executionBlock = nil

        if let block = block {
            block(self)
        } else {
            finish()
        }
    }
```

The `else { finish() }` will never get executed as far as I can tell, it's dead code.

When constructing an `AsyncBlockOperation` we keep a strong reference to the `block` passed in the constructor.
So when unwrapping it with the `if let block = block` this will always succeed, since we have a strong reference to it.
Therefore, we always execute that block, and the else branch`finish()` is never implicitly called.

The following test proves that 
```
    func testBlockOperationDoesNotFinish() throws {
        
        let expectation = expectation(description: "test")
        let operation: AsyncBlockOperation = AsyncBlockOperation { _ in  expectation.fulfill() }
        operationQueue.addOperation(operation)
        waitForExpectations(timeout: 1)
        
        XCTAssertTrue(operation.isFinished)
        
    }
```
</blockquote></details>

Good spot. It's a dead code. `ResultBlockOperation` does not do that and I'll make sure to align implementations

Otherwise It's by contract that if you create an operation that accepts itself as an argument, that you have to `finish()` it manually.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)* status: 14 of 19 files reviewed, 6 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/OperationsTests/AsyncBlockOperationTests.swift` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURnV-F3rTirqNy81gK:-NUW692rBBzI7Or4oq1P:bdvegjf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncBlockOperationTests.swift#L15)):*

Good spot. It's a dead code. `ResultBlockOperation` does not do that and I'll make sure to align implementations

Otherwise It's by contract that if you create an operation that accepts itself as an argument, that you have to `finish()` it manually.
</blockquote></details>

Sounds fair enough, but in that case I would like to request that we document that, or add a check somewhere that will end in a `fatalError` if that isn't the case. Reading the code from `AsyncBlockOperation` doesn't suggest that, or the finish function itself, it can easily be misued.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)* status: 13 of 19 files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/OperationsTests/AsyncBlockOperationTests.swift` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURnV-F3rTirqNy81gK:-NUWMJTf3worZQuBCYyA:bnzo8c5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncBlockOperationTests.swift#L15)):*

Sounds fair enough, but in that case I would like to request that we document that, or add a check somewhere that will end in a `fatalError` if that isn't the case. Reading the code from `AsyncBlockOperation` doesn't suggest that, or the finish function itself, it can easily be misued. 
</blockquote></details>

I've added assertions. 

Ideally `fatalError` and forced unwrapping should be avoided as we do not want the app crash in the middle of I/O or any other critical task that can be happening in parallel.

Would it be clearer if instead of `Operation` we passed a `finish` block instead? 

```swift
typealias FinishHandler = (Error?) -> Void

init(..., block: (_ finish: FinishHandler) -> Void)
```

I think we never really check for cancellation inside of block operations as `AsyncOperation` handles that early on before it starts + we have cancellable initializer now.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)* status: 13 of 19 files reviewed, 6 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/OperationsTests/AsyncBlockOperationTests.swift` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURnV-F3rTirqNy81gK:-NUWOyuODZGzqL1X0YwI:bmp7qes) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncBlockOperationTests.swift#L15)):*
> Ideally `fatalError` and forced unwrapping should be avoided
Agreed, I was citing that as an example because I consider "not using the provided object according to the contract a serious programming error" and as such, the consequences for not respecting the API should be serious.

> Would it be clearer if instead of `Operation` we passed a `finish` block instead?
I think so yes, if we can enforce the correct usage via the API all the better. 
I can easily see myself misuse an API because I didn't read the documentation correctly. 
If the API itself forces the correct usage, then I cannot make the mistake and everyone wins.

That being said, if it's too cumbersome to implement, or results in code bloating, maybe the asserts would be good enough for now ?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)* status: 13 of 19 files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/OperationsTests/AsyncBlockOperationTests.swift` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURnV-F3rTirqNy81gK:-NUWQqSyCGKYtrABNDQ4:b3jrdth) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncBlockOperationTests.swift#L15)):*

> Ideally `fatalError` and forced unwrapping should be avoided
Agreed, I was citing that as an example because I consider "not using the provided object according to the contract a serious programming error" and as such, the consequences for not respecting the API should be serious.

> Would it be clearer if instead of `Operation` we passed a `finish` block instead?
I think so yes, if we can enforce the correct usage via the API all the better. 
I can easily see myself misuse an API because I didn't read the documentation correctly. 
If the API itself forces the correct usage, then I cannot make the mistake and everyone wins.

That being said, if it's too cumbersome to implement, or results in code bloating, maybe the asserts would be good enough for now ?
</blockquote></details>

I'll look into that. Thanks for your feedback.

buggmagnet (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623)* status: 4 of 22 files reviewed, 5 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/OperationsTests/AsyncBlockOperationTests.swift` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4623#-NURnV-F3rTirqNy81gK:-NUX7JHAEh-QCtlV-V4i:b-66o7xy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cb857165e517e6e48d71766e8163e6016c495ec3/ios/OperationsTests/AsyncBlockOperationTests.swift#L15)):*

Finished by giving block operations that run asynchronously a closure to finish operation:

Block operations receive `(Error?) -> Void`
Result block operations receive `(Result<Output, Error>) -> Void`
Transform operations receive `(Input, (Result<Output, Error>) -> Void) -> Void`
</blockquote></details>

Sounds good to me :thumbsup:  !

PR #4629, opened 2023-04-27
Fix in-app notification equality and action button color

pronebird:
1. This PR reverts `Equatable` implementation for  `InAppNotificationDescriptor`. The same notification can be displayed many times with different attributes so checking notifications by `identifier` only does not always work. This proved to be the case with `RegisteredDeviceInAppNotification` which didn't disappear on logout and then didn't update after logging with new account. Maybe it's an edge case? In any case, please only call `invalidate()` on notification providers when external events result into new content being produced.
2. `InAppNotificationAction` implements `Equatable` but ignore block handler because there is no easy way to achieve that, but on other side we always produce new closures so it's pointless either way. In that regard let's focus equality check for properties that affect visual content representation and leave handlers out of it.
3. Instead of providing `UIImage`, `InAppNotifcationAction` now accepts a enum type (`InAppNotificationActionType`). More or less we're going to use a fixed list of actions so it's ok if we keep them limited and control graphics in one place and not in each individual provider.
4. Fix tint color for action button.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4629)

### DIFF
--- a/ios/MullvadVPN/Notifications/InAppNotificationDescriptor.swift
+++ b/ios/MullvadVPN/Notifications/InAppNotificationDescriptor.swift
@@ -10,7 +10,7 @@ import Foundation
 import UIKit.UIImage
 
 /// Struct describing in-app notification.
-struct InAppNotificationDescriptor {
+struct InAppNotificationDescriptor: Equatable {
     /// Notification identifier.
     var identifier: String
 
@@ -23,19 +23,21 @@ struct InAppNotificationDescriptor {
     /// Notification body.
     var body: NSAttributedString
 
-    /// Notification action
+    /// Notification action.
     var action: InAppNotificationAction?
 }
 
-extension InAppNotificationDescriptor: Equatable {
-    static func == (lhs: InAppNotificationDescriptor, rhs: InAppNotificationDescriptor) -> Bool {
-        lhs.identifier == rhs.identifier
-    }
-}
-
-struct InAppNotificationAction {
+/// Type describing a specific in-app notification action.
+struct InAppNotificationAction: Equatable {
+    /// Image assigned to action button.
     var image: UIImage?
+
+    /// Action handler for button.
     var handler: (() -> Void)?
+
+    static func == (lhs: InAppNotificationAction, rhs: InAppNotificationAction) -> Bool {
+        return lhs.image == rhs.image
+    }
 }
 
 enum NotificationBannerStyle {
--- a/ios/MullvadVPN/Notifications/NotificationManager.swift
+++ b/ios/MullvadVPN/Notifications/NotificationManager.swift
@@ -204,9 +204,7 @@ final class NotificationManager: NotificationProviderDelegate {
             if let replaceNotificationDescriptor = notificationProvider.notificationDescriptor {
                 newNotificationDescriptors = notificationProviders
                     .compactMap { notificationProvider -> InAppNotificationDescriptor? in
-                        if replaceNotificationDescriptor.identifier == notificationProvider
-                            .identifier
-                        {
+                        if replaceNotificationDescriptor.identifier == notificationProvider.identifier {
                             return replaceNotificationDescriptor
                         } else {
                             return inAppNotificationDescriptors.first { descriptor in
--- a/ios/MullvadVPN/Notifications/UI/NotificationBannerView.swift
+++ b/ios/MullvadVPN/Notifications/UI/NotificationBannerView.swift
@@ -89,7 +89,7 @@ final class NotificationBannerView: UIView {
 
     var actionHandler: InAppNotificationAction? {
         didSet {
-            actionButton.setImage(actionHandler?.image, for: .normal)
+            actionButton.setImage(actionHandler?.type.image, for: .normal)
             actionButton.addTarget(self, action: #selector(didPress), for: .touchUpInside)
         }
     }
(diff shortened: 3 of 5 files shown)


buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4629)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN/Notifications/InAppNotificationDescriptor.swift` line 33 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4629#-NU5rd0T3AtNfx2TMSSG:-NU5rd0T3AtNfx2TMSSH:b-67tlct) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/00e4cccf08c20c95d66a00981857f418d3b317ba/ios/MullvadVPN/Notifications/InAppNotificationDescriptor.swift#L33)):*
> ```Swift
> enum InAppNotificationActionType {
>     /// Action type represented by an image that depicts a cross tilted by 45 degrees.
>     case close
> ```

Do we plan to have more than one `InAppNotificationActionType` ? 
Otherwise this feels a bit overkill. 

Also right now there's no link between the image that's forcedly returning a `close` image, and the action.

I would suggest to change the image getter.
Check `self ` in a switch statement and return the appropriate image depending on the type of action instead.

```
    var image: UIImage? {
        switch self {
        case .close: return UIImage(named: "IconCloseSml")
        }
    }
```

There's no `default` handling in the above because there's only one case, but you get the picture
___
*[`ios/MullvadVPN/Notifications/InAppNotificationDescriptor.swift` line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4629#-NU5mXdEEDgZlqESUslh:-NU5mXdEEDgZlqESUsli:b-uxnwfl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/00e4cccf08c20c95d66a00981857f418d3b317ba/ios/MullvadVPN/Notifications/InAppNotificationDescriptor.swift#L37)):*
> ```Swift
>     /// Returns an image associated with the corresponding action type.
>     var image: UIImage? {
>         return UIImage(named: "IconCloseSml")?.withRenderingMode(.automatic)
> ```

The `renderingMode` is already `automatic` [by default](https://developer.apple.com/documentation/uikit/uiimage/1624122-renderingmode). 
You can remove `?.withRenderingMode(.automatic)`

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4629)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/Notifications/InAppNotificationDescriptor.swift` line 33 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4629#-NU5rd0T3AtNfx2TMSSG:-NU66kjCAHSCZxPojf24:bfdy79o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/00e4cccf08c20c95d66a00981857f418d3b317ba/ios/MullvadVPN/Notifications/InAppNotificationDescriptor.swift#L33)):*

Do we plan to have more than one `InAppNotificationActionType` ? 
Otherwise this feels a bit overkill. 

Also right now there's no link between the image that's forcedly returning a `close` image, and the action.

I would suggest to change the image getter.
Check `self ` in a switch statement and return the appropriate image depending on the type of action instead.

```
    var image: UIImage? {
        switch self {
        case .close: return UIImage(named: "IconCloseSml")
        }
    }
```

There's no `default` handling in the above because there's only one case, but you get the picture
</blockquote></details>

Yes I think there will be more but I have reverted this back to `UIImage`. We can introduce new type when we have more kinds of actions
___
*[`ios/MullvadVPN/Notifications/InAppNotificationDescriptor.swift` line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4629#-NU5mXdEEDgZlqESUslh:-NU6HMuG9qG4qqVXlDnp:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/00e4cccf08c20c95d66a00981857f418d3b317ba/ios/MullvadVPN/Notifications/InAppNotificationDescriptor.swift#L37)):*

The `renderingMode` is already `automatic` [by default](https://developer.apple.com/documentation/uikit/uiimage/1624122-renderingmode). 
You can remove `?.withRenderingMode(.automatic)`

</blockquote></details>

Done.

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4629#-:-NU6cqmC5cmEYGRHcl52:bnfp4nl)* status: 0 of 6 files reviewed, 2 unresolved discussions

buggmagnet (commented):
> Reviewed 2 of 5 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4629)* status: all files reviewed, 2 unresolved discussions

buggmagnet (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4629)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4652, opened 2023-05-04
Rework the address cache IOS-149

buggmagnet:
This PR vastly simplifies the `AddressCache` class whilst keeping the rest of the architecture of the app intact to minimize changes.

As we are trying to deprecate port-forwarded addresses, we now ignore all but the first result when querying the `api-addrs` endpoint.
A new filter method has been added to do just that.

### DIFF
--- a/ios/MullvadREST/AddressCache.swift
+++ b/ios/MullvadREST/AddressCache.swift
@@ -61,44 +61,15 @@ extension REST {
         public func getCurrentEndpoint() -> AnyIPEndpoint {
             nslock.lock()
             defer { nslock.unlock() }
-            return cachedAddresses.endpoints.first!
+            return cachedAddresses.endpoints.first ?? REST.defaultAPIEndpoint
         }
 
         public func selectNextEndpoint(_ failedEndpoint: AnyIPEndpoint) -> AnyIPEndpoint {
             nslock.lock()
             defer { nslock.unlock() }
 
-            var currentEndpoint = cachedAddresses.endpoints.first!
-
-            guard failedEndpoint == currentEndpoint else {
-                return currentEndpoint
-            }
-
-            cachedAddresses.endpoints.removeFirst()
-            cachedAddresses.endpoints.append(failedEndpoint)
-
-            if isReadOnly {
-                refreshAddresses()
-            }
-
-            currentEndpoint = cachedAddresses.endpoints.first!
-
-            logger.debug(
-                "Failed to communicate using \(failedEndpoint). Next endpoint: \(currentEndpoint)"
-            )
-
-            if !isReadOnly {
-                do {
-                    try writeToDisk()
-                } catch {
-                    logger.error(
-                        error: error,
-                        message: "Failed to write address cache after selecting next endpoint."
-                    )
-                }
-            }
-
-            return currentEndpoint
+            // This function currently acts as a convoluted no-op. It will be soon deleted.
+            return getCurrentEndpoint()
         }
 
         public func setEndpoints(_ endpoints: [AnyIPEndpoint]) {
@@ -111,23 +82,7 @@ extension REST {
 
             if Set(cachedAddresses.endpoints) == Set(endpoints) {
                 cachedAddresses.updatedAt = Date()
-            } else {
-                // Shuffle new endpoints
-                var newEndpoints = endpoints.shuffled()
-
-                // Move current endpoint to the top of the list
-                let currentEndpoint = cachedAddresses.endpoints.first!
-                if let index = newEndpoints.firstIndex(of: currentEndpoint) {
-                    newEndpoints.remove(at: index)
-                    newEndpoints.insert(currentEndpoint, at: 0)
-                }
-
-                cachedAddresses = CachedAddresses(
-                    updatedAt: Date(),
-                    endpoints: newEndpoints
-                )
             }
-
             if !isReadOnly {
                 do {
                     try writeToDisk()
@@ -167,9 +122,7 @@ extension REST {
                 cachedAddresses = readResult.cachedAddresses
 
             case .bundle:
-                var addresses = readResult.cachedAddresses
-                addresses.endpoints.shuffle()
-                cachedAddresses = addresses
+                cachedAddresses = readResult.cachedAddresses
 
                 if !isReadOnly {
                     logger.debug("Persist address list read from bundle.")
@@ -289,27 +242,6 @@ extension REST {
             }
 
             return try result!.get()
-        }
-
-        private func refreshAddresses() {
-            do {
-                let readResult = try readFromCacheLocation()
-                var newCachedAddresses = readResult.cachedAddresses
-
-                guard Set(newCachedAddresses.endpoints) != Set(cachedAddresses.endpoints)
-                else { return }
-
-                // Move current endpoint to the top of the list
-                let currentEndpoint = cachedAddresses.endpoints.first!
-                if let index = newCachedAddresses.endpoints.firstIndex(of: currentEndpoint) {
-                    newCachedAddresses.endpoints.remove(at: index)
-                    newCachedAddresses.endpoints.insert(currentEndpoint, at: 0)
-                }
-
-                cachedAddresses = newCachedAddresses
-            } catch {
-                logger.error(error: error, message: "Failed to refresh address cache from disk.")
-            }
         }
     }
 
--- a/ios/MullvadREST/RESTAPIProxy.swift
+++ b/ios/MullvadREST/RESTAPIProxy.swift
@@ -37,10 +37,7 @@ extension REST {
                 )
             }
 
-            let responseHandler = REST.defaultResponseHandler(
-                decoding: [AnyIPEndpoint].self,
-                with: responseDecoder
-            )
+            let responseHandler = REST.apiFilterResponseHandler(with: responseDecoder)
 
             return addOperation(
                 name: "get-api-addrs",
--- a/ios/MullvadREST/RESTResponseHandler.swift
+++ b/ios/MullvadREST/RESTResponseHandler.swift
@@ -7,6 +7,7 @@
 //
 
 import Foundation
+import MullvadTypes
 
 protocol RESTResponseHandler {
     associatedtype Success
@@ -56,6 +57,33 @@ extension REST {
             if HTTPStatus.isSuccess(response.statusCode) {
                 return .decoding {
                     try decoder.decode(type, from: data)
+                }
+            } else {
+                return .unhandledResponse(
+                    try? decoder.decode(
+                        ServerErrorResponse.self,
+                        from: data
+                    )
+                )
+            }
+        }
+    }
+
+    /// A response handler that parses a JSON response that produces an array of `AnyIPEndpoint`s
+    ///
+    /// All results but the first will be ignored.
+    /// - Parameter decoder: A JSON decoder
+    /// - Returns: An array containing the first endpoint found in `data`, or a default value in case none were
+    /// found.
+    static func apiFilterResponseHandler(
+        with decoder: JSONDecoder
+    ) -> AnyResponseHandler<[AnyIPEndpoint]> {
+        return AnyResponseHandler { response, data in
+            if HTTPStatus.isSuccess(response.statusCode) {
+                return .decoding {
+                    guard let result = try decoder.decode([AnyIPEndpoint].self, from: data).first
+                    else { return [defaultAPIEndpoint] }
+                    return [result]
                 }
             } else {
                 return .unhandledResponse(


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/AddressCache.swift` line 122 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NU_wIY6C7q4WD-sc76a:-NU_wIY6C7q4WD-sc76b:b-ps4w1v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e1686bc21f91d35a7ed0bf1d9eb7cd3aa3f275f/ios/MullvadREST/AddressCache.swift#L122)):*
> ```Swift
>             switch readResult.source {
>             case .disk:
>                 cachedAddresses = readResult.cachedAddresses
> ```

Seems like both `.bundle` and `.disk`  cases assign cached addresses, so maybe this can be simplified to:

```swift
cachedAddresses = readResult.cachedAddresses
if readResult.source == .bundle, !isReadonly {
  // persist to disk cache
}
```
___
*[`ios/MullvadREST/AddressCache.swift` line 294 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NU_z_fjDefkbwY_T7hH:-NU_z_fjDefkbwY_T7hI:bzeyh20) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b8091d32497d52171ffc3888094d52856d96b194/ios/MullvadREST/AddressCache.swift#L294)):*
> ```Swift
>         }
> 
>         private func refreshAddresses() {
> ```

This code was used to reload address cache in packet tunnel where we don't have "tracker" class as it's main process only. So perhaps in readonly mode we should always read file from disk, i.e:

```swift
func getCurrentEndpoint() -> AnyIPEndpoint {
 if isReadOnly {
       // read cache same way we do during initialization
  } else {
      nslock.lock()
      defer { nslock.unlock() }
      return cachedAddresses.endpoints.first ?? REST.defaultAPIEndpoint
  }
}
```
___
*[`ios/MullvadREST/RESTResponseHandler.swift` line 85 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NUa0T9PAj3ZkpjEatjv:-NUa0T9PAj3ZkpjEatjw:bfl0qye) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e1686bc21f91d35a7ed0bf1d9eb7cd3aa3f275f/ios/MullvadREST/RESTResponseHandler.swift#L85)):*
> ```Swift
>                 return .decoding {
>                     guard let result = try decoder.decode([AnyIPEndpoint].self, from: data).first
>                     else { return [defaultAPIEndpoint] }
> ```

Yuck why does API client replace the server response? This is what consumer should do?

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NUa2Bid7PgDdSjhExfX:-NUa2Bie526J02cKcMcm:bjwkimr) (no related file):*
I looked into ip-address-cache a bit and I think we should simply ship one address.

The cURL line in `rest-prebuild.sh` can be simplified to:

```
curl https://api.mullvad.net/app/v1/api-addrs -s | jq '.[0]' > ip-address-cache.txt 
```

`jq` is almost a standard tool these days so I think it’s reasonable to expect it installed on build server and developer machines. Reading and shipping one .txt file with a single address is a low effort but would align with how it’s done on desktop and simplify things a bit on iOS side too along with removing confusion if anyone has question in regards of why we have many addresses but use only one. (edit

___

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: all files reviewed, 4 unresolved discussions (waiting on @pronebird)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NUa2Bid7PgDdSjhExfX:-NUa_3U83_Xxkr9a24UZ:b-supd3o) (no related file):*

The right command would be with `-r` flag:

```
curl https://api.mullvad.net/app/v1/api-addrs -s | jq '.[0]' -r > ip-address-cache.txt 
```
</blockquote></details>

I went with the following command 
```
curl https://api.mullvad.net/app/v1/api-addrs -s | cut -d'"' -f 2 | sed -e 's/\(.*\)/["\1"]/g' > "$API_IP_ADDRESS_LIST_FILE"
```

To avoid needing a dependency on `jq`
More comments on the line itself in the `rest-prebuild.sh` script 

___
___
*[`ios/MullvadREST/AddressCache.swift` line 122 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NU_wIY6C7q4WD-sc76a:-NUac7pO2Ro6iFMosP7f:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e1686bc21f91d35a7ed0bf1d9eb7cd3aa3f275f/ios/MullvadREST/AddressCache.swift#L122)):*

Seems like both `.bundle` and `.disk`  cases assign cached addresses, so maybe this can be simplified to:

```swift
cachedAddresses = readResult.cachedAddresses
if readResult.source == .bundle, !isReadonly {
  // persist to disk cache
}
```
</blockquote></details>

Done.
___
*[`ios/MullvadREST/AddressCache.swift` line 294 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NU_z_fjDefkbwY_T7hH:-NUagt231ZZnwtCV3248:bcqm9de) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b8091d32497d52171ffc3888094d52856d96b194/ios/MullvadREST/AddressCache.swift#L294)):*

This code was used to reload address cache in packet tunnel where we don't have "tracker" class as it's main process only. So perhaps in readonly mode we should always read file from disk, i.e:

```swift
func getCurrentEndpoint() -> AnyIPEndpoint {
 if isReadOnly {
       // read cache same way we do during initialization
  } else {
      nslock.lock()
      defer { nslock.unlock() }
      return cachedAddresses.endpoints.first ?? REST.defaultAPIEndpoint
  }
}
```
</blockquote></details>

That's a very good point, I'll reimport some of the cache update logic from the original code.
___
*[`ios/MullvadREST/RESTResponseHandler.swift` line 85 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NUa0T9PAj3ZkpjEatjv:-NUagzeU1vRpinZQDVc1:bgvwcew) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e1686bc21f91d35a7ed0bf1d9eb7cd3aa3f275f/ios/MullvadREST/RESTResponseHandler.swift#L85)):*

Yuck why does API client replace the server response? This is what consumer should do?
</blockquote></details>

This was discussed offline. 
The conclusion is that the client only cares about the first endpoint returned by the API and discards the rest.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @buggmagnet)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NUa2Bid7PgDdSjhExfX:-NUapCuMCwwrVPSArRc6:b-bu8cga) (no related file):*

I went with the following command 
```
curl https://api.mullvad.net/app/v1/api-addrs -s | cut -d'"' -f 2 | sed -e 's/\(.*\)/["\1"]/g' > "$API_IP_ADDRESS_LIST_FILE"
```

To avoid needing a dependency on `jq`
More comments on the line itself in the `rest-prebuild.sh` script 
</blockquote></details>

`jq` would parse JSON properly. But I don't mind how you do that. See the discussion on Slack for more info.

What I think is sensible to do is to store a TXT file instead of JSON and then read it as such in the address cache:


```swift
let data = try Data(contentsOf: "<api-ip-address.txt file from bundle>")

guard let endpoint = String(data: data, encoding: .utf8)
    .flatMap { AnyIPAddress($0) } else { throw CannotReadDefaultIPFromFile() }
```

___
___
*[`ios/rest-prebuild.sh` line 23 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NUan5438NkDVKNPSVHQ:-NUan5438NkDVKNPSVHR:b-6bq2cd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ead3a4086b869b5ee593e2657fe03953bd91eae2/ios/rest-prebuild.sh#L23)):*
> ```Shell
>   # Take only the first returned endpoint, and ignore the rest
>   # The sed command wraps the output of `cut` with square brackets and double quotes to keep the JSON valid
>   curl https://api.mullvad.net/app/v1/api-addrs -s | cut -d'"' -f 2 | sed -e 's/\(.*\)/["\1"]/g' > "$API_IP_ADDRESS_LIST_FILE"
> ```

Maybe change `api-ip-address.json` to `api-ip-address.txt` and store single IP in it. It should be much simpler to load text file than parsing JSON just to extract a single IP from it.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @buggmagnet)

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet)

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NUetY2l9_VRpnqx-q2l:-NUetY2l9_VRpnqx-q2m:b9kxix6) (no related file):*
As we discussed it on Slack today. Let's remove pre bundled addresses and rely on static one in the code.

___

pronebird (changes requested):
> Reviewed 3 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: 3 of 8 files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/AddressCache.swift` line 31 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV-JiIy4f04gAxwCm0E:-NV-JiIy4f04gAxwCm0F:bh95gxi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadREST/AddressCache.swift#L31)):*
> ```Swift
> 
>         /// The name of the cache file on disk
>         internal static let cacheFileName = "api-ip-address.txt"
> ```

It should perhaps remain called `api-ip-address.json`.
___
*[`ios/MullvadREST/AddressCache.swift` line 132 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV-OEnG-p-oybdiCIJh:-NV-OEnG-p-oybdiCIJi:b-xh65u9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadREST/AddressCache.swift#L132)):*
> ```Swift
>         /// If no cache file is present, a default API endpoint will be selected instead
>         private func initCache() {
>             cacheLock.lock()
> ```

There is no need to lock this method as `init()` is atomic whichever way you look at it. You may as well move the body of this function into `init()`
___
*[`ios/MullvadREST/AddressCache.swift` line 197 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV-SUEu7wdH55oUbySa:-NV-SUEu7wdH55oUbySb:b-87rh75) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadREST/AddressCache.swift#L197)):*
> ```Swift
>         ///   - accessor: The `accessor` closure used by `NSFileCoordinator`
>         /// - Returns: `nil`, or `Swift.Error` if an error happened when accessing the cache file
>         private func accessCache(for action: CacheAction, accessor: (URL) -> Void) -> Swift.Error? {
> ```

This function provides very little convenience over what we had before, but introduces `CacheAction` entity which is ultimately a boolean flag.

Maybe we should consider adding small extension on file coordinator to make its API a bit less C-ish looking, for example:

```swift
extension NSFileCoordinator {
    func coordinate<R>(readingItemAt itemURL: URL, options: ReadingOptions = [], accessor: (URL) throws -> R) throws -> R {
        var error: NSError?
        var result: Result<R, Error> = .failure(NSError(domain: NSCocoaErrorDomain, code: NSFileReadUnknownError))

        coordinate(readingItemAt: itemURL, options: options, error: &error) { url in
            result = Result { try accessor(url) }
        }

        if let error {
            throw error
        }

        return try result.get()
    }
}
```

i.e this looks a bit less messy:

```swift
let fileCoordinator = NSFileCoordinator(filePresenter: nil)

do {
    let value = try fileCoordinator.coordinate(readingItemAt: URL(filePath: "/tmp/file.txt"), options: []) { file in
        return 0
    }
    print(value)
} catch {
    print(error)
}
```

pronebird (commented):
> Reviewed 5 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: all files reviewed, 5 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadREST/AddressCache.swift` line 206 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV-kJHQ0gk0XhheaZbb:-NV-kJHQ0gk0XhheaZbc:bxidynh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadREST/AddressCache.swift#L206)):*
> ```Swift
>                 fileCoordinator.coordinate(
>                     readingItemAt: cacheFileURL,
>                     options: [],
> ```

I forgot to add back the `.withoutChanges` option here

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadRESTTests/AddressCacheTests.swift` line 209 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV4A-dgC1sV3W3oqEWY:-NV4A-dgC1sV3W3oqEWZ:bdckjhn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadRESTTests/AddressCacheTests.swift#L209)):*
> ```Swift
> }
> 
> class AddressCacheFilePresenter: NSObject, NSFilePresenter {
> ```

Feels heavy. What do you think about making an abstract reader and writer for cache, create the one that uses `NSFileCoordinator` in production and provide a memory-based mock in development?

```swift
protocol FileReaderWriter {
   func read() throws -> Data
   func write(_ bytes: Data) throws
}
```

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadRESTTests/AddressCacheTests.swift` line 209 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV4A-dgC1sV3W3oqEWY:-NV4AzZA6yntEecZpZe9:b2eakjl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadRESTTests/AddressCacheTests.swift#L209)):*

Feels heavy. What do you think about making an abstract reader and writer for cache, create the one that uses `NSFileCoordinator` in production and provide a memory-based mock in development?

```swift
protocol FileReaderWriter {
   func read() throws -> Data
   func write(_ bytes: Data) throws
}
```
</blockquote></details>

Sorry I meant to mock file I/O in test environment using memory cache, and then use the file coordinator based class otherwise. The interface can be different, but something that does not oblige us to use file system in test environment.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: 7 of 8 files reviewed, 4 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadREST/AddressCache.swift` line 31 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV-JiIy4f04gAxwCm0E:-NV3gD2q0isZiIiDwswN:bewnj97) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadREST/AddressCache.swift#L31)):*

It should perhaps remain called `api-ip-address.json`.
</blockquote></details>

Yes definitely
___
*[`ios/MullvadREST/AddressCache.swift` line 132 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV-OEnG-p-oybdiCIJh:-NV3gVATD2moLeBMvYy3:biqbuhu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadREST/AddressCache.swift#L132)):*

There is no need to lock this method as `init()` is atomic whichever way you look at it. You may as well move the body of this function into `init()`
</blockquote></details>

Agreed about the lock statement, I will remove it.
As for moving it into the `init` I think I want to keep it separated. It helps reasoning about what it does in context that's separate from `init` 
___
*[`ios/MullvadREST/AddressCache.swift` line 197 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV-SUEu7wdH55oUbySa:-NV8eunQ2EnF7yXS9qqW:bs8td29) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadREST/AddressCache.swift#L197)):*

This function provides very little convenience over what we had before, but introduces `CacheAction` entity which is ultimately a boolean flag.

Maybe we should consider adding small extension on file coordinator to make its API a bit less C-ish looking, for example:

```swift
extension NSFileCoordinator {
    func coordinate<R>(readingItemAt itemURL: URL, options: ReadingOptions = [], accessor: (URL) throws -> R) throws -> R {
        var error: NSError?
        var result: Result<R, Error> = .failure(NSError(domain: NSCocoaErrorDomain, code: NSFileReadUnknownError))

        coordinate(readingItemAt: itemURL, options: options, error: &error) { url in
            result = Result { try accessor(url) }
        }

        if let error {
            throw error
        }

        return try result.get()
    }
}
```

i.e this looks a bit less messy:

```swift
let fileCoordinator = NSFileCoordinator(filePresenter: nil)

do {
    let value = try fileCoordinator.coordinate(readingItemAt: URL(filePath: "/tmp/file.txt"), options: []) { file in
        return 0
    }
    print(value)
} catch {
    print(error)
}
```
</blockquote></details>

This was discussed offline.
I'm deciding to not use this for now, and revisit it later when we look at adding tests for `RelayCache` because it also uses `FileCoordinator`. 

There is definitely merit in having a nicer API for using `FileCoordinator`.
___
*[`ios/MullvadRESTTests/AddressCacheTests.swift` line 209 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV4A-dgC1sV3W3oqEWY:-NV8b_0m3HSFsVMhBtTT:b-ujirqk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadRESTTests/AddressCacheTests.swift#L209)):*

Sorry I meant to mock file I/O in test environment using memory cache, and then use the file coordinator based class otherwise. The interface can be different, but something that does not oblige us to use file system in test environment.
</blockquote></details>

My original intention was to do exactly that. But it quickly got messy because suddenly I had to find a way to execute the `accessor` closures that do the actual work from the mocks, and then mock the URL passed to them etc...

This was definitely the simplest approach I could think of that didn't change too much the original code and that offered an acceptable tradeoff in terms of complexity and testability.
Maybe I'm missing something obvious here.

You bring a very good point about doing I/O in the test environment using memory caches instead of disk caches.
This brings the equally great point of "we're here to test our code, not Apple's APIs".

But in my experience, using I/O in CI is not bad as long as you're not doing crazy things.
Creating a temporary folder should be fine most of the time. And sometimes it's inevitable when it comes to testing caching.

I don't see anymore the test environment as something pristine that should never be doing I/O or even network.
As long as it's correctly controlled, it brings the testing game to the next level IMHO.

The bonus of having a dedicated `AddressCacheFilePresenter` is that we can do further simulation of the network extension and the main process running in parallel and fighting for cache access, if we wanted to test for deadlocks here.
That was a lot of extra work for this, and I decided to leave it for later down the road.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/AddressCache.swift` line 197 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV-SUEu7wdH55oUbySa:-NV8thMZ-c0t9niVZ8mo:b8u2yvc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadREST/AddressCache.swift#L197)):*

This was discussed offline.
I'm deciding to not use this for now, and revisit it later when we look at adding tests for `RelayCache` because it also uses `FileCoordinator`. 

There is definitely merit in having a nicer API for using `FileCoordinator`.
</blockquote></details>

I fail to see how this is an improvement over what we had before and hence don't see it as being necessary. 

What was wrong with write/read cache functions and we it has to be moved into `accessCache()`. It also returns error which is not Swift-way of doing things and rather objective-c.

Result is repeated twice now too.

```swift
let unknownWriteError = NSError(domain: NSPOSIXErrorDomain, code: NSFileWriteUnknownError)
var result: Result<Void, Swift.Error> = .failure(unknownWriteError)
```

So either do it right in the first time or leave it unchanged because it does not help it to reason about code easier, not has any impact on the outcome.
___
*[`ios/MullvadRESTTests/AddressCacheTests.swift` line 209 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV4A-dgC1sV3W3oqEWY:-NV8pEx_96BBVb4vbzAA:b-r352op) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadRESTTests/AddressCacheTests.swift#L209)):*

My original intention was to do exactly that. But it quickly got messy because suddenly I had to find a way to execute the `accessor` closures that do the actual work from the mocks, and then mock the URL passed to them etc...

This was definitely the simplest approach I could think of that didn't change too much the original code and that offered an acceptable tradeoff in terms of complexity and testability.
Maybe I'm missing something obvious here.

You bring a very good point about doing I/O in the test environment using memory caches instead of disk caches.
This brings the equally great point of "we're here to test our code, not Apple's APIs".

But in my experience, using I/O in CI is not bad as long as you're not doing crazy things.
Creating a temporary folder should be fine most of the time. And sometimes it's inevitable when it comes to testing caching.

I don't see anymore the test environment as something pristine that should never be doing I/O or even network.
As long as it's correctly controlled, it brings the testing game to the next level IMHO.

The bonus of having a dedicated `AddressCacheFilePresenter` is that we can do further simulation of the network extension and the main process running in parallel and fighting for cache access, if we wanted to test for deadlocks here.
That was a lot of extra work for this, and I decided to leave it for later down the road.

</blockquote></details>

`NSFilePresenter` is a very exotic thing and is not the simplest approach. JSON serialization which is an implementation detail is also used in quite a few places.

Why can't we wrap the entire persistence and serialization layer into a simple type instead, something along these lines:

```swift
struct FileReaderWriter {
    init(cacheDirectory: URL, filename: String)

    // for testing
    internal init(inMemoryCache initialValue: CacheAddresses)

    func read() throws -> CachedAddresses 
    func write(_ cachedAddresses: CachedAddresses) throws
}
```

This would enable you to test logic of address cache and not how file coordinator works.

There is currently 239 lines of code (LOC) in `AddressCache.swift` and 236 LOC in `AddressCacaheTests.swift`. This is an overkill. I would not even bother to write tests for such class but if you insist then at least make it simpler to test so that we don't spend more efforts on testing the code than the tested code itself.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadRESTTests/AddressCacheTests.swift` line 209 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV4A-dgC1sV3W3oqEWY:-NV9IIQ-A9k0ibCqN-67:bshl0bo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75b52da486cb3fb4fccacdd48fc0bc718279d3a7/ios/MullvadRESTTests/AddressCacheTests.swift#L209)):*

`NSFilePresenter` is a very exotic thing and is not the simplest approach. JSON serialization which is an implementation detail is also used in quite a few places.

Why can't we wrap the entire persistence and serialization layer into a simple type instead, something along these lines:

```swift
struct FileReaderWriter {
    init(cacheDirectory: URL, filename: String)

    // for testing
    internal init(inMemoryCache initialValue: CacheAddresses)

    func read() throws -> CachedAddresses 
    func write(_ cachedAddresses: CachedAddresses) throws
}
```

This would enable you to test logic of address cache and not how file coordinator works.

There is currently 239 lines of code (LOC) in `AddressCache.swift` and 236 LOC in `AddressCacaheTests.swift`. This is an overkill. I would not even bother to write tests for such class but if you insist then at least make it simpler to test so that we don't spend more efforts on testing the code than the tested code itself.
</blockquote></details>

It has been discussed offlie earlier today, but I don't necessarily see a problem with a test containing more LOC than its counterpart. It all depends on what is tested and the complexity of it. I think a mocked I/O would have been preferred, but now that time has been spent on building and tuning things this way I would be hard-pressed to say they should be rewritten.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: 6 of 9 files reviewed, 1 unresolved discussion (waiting on @pronebird and @rablador)

___
*[`ios/MullvadREST/AddressCache.swift` line 77 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV8uMhKBIAfYOb80jUu:-NVAAhuO2zpWncxDNrsH:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7846a9963cb054526c20404aa5b6b3117584576b/ios/MullvadREST/AddressCache.swift#L77)):*

Nit: Missing line spacing above.
</blockquote></details>

Done

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652)* status: 6 of 9 files reviewed, all discussions resolved (waiting on @rablador)

___
*[`ios/MullvadREST/AddressCache.swift` line 88 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV8ua5C3paFHs7HYyg1:-NVEON7RE5F2Bq1VW8v3:b-gpwze3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7846a9963cb054526c20404aa5b6b3117584576b/ios/MullvadREST/AddressCache.swift#L88)):*

https://english.stackexchange.com/questions/326475/on-hard-drive-versus-in-memory#326487

This is valid english as far as I'm aware
</blockquote></details>

"on-disk" cache I think but that's fine.

pronebird (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-:-NVEeKbe5dVWPQsfa5UO:b-uwkowd)* status: :shipit: complete! all files reviewed, all discussions resolved

rablador (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-:-NVEhfh-0NNQcLXPnwMH:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`ios/MullvadREST/AddressCache.swift` line 88 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4652#-NV8ua5C3paFHs7HYyg1:-NVEgz4F4PYgemLbrV-o:bx464zs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7846a9963cb054526c20404aa5b6b3117584576b/ios/MullvadREST/AddressCache.swift#L88)):*

"on-disk" cache I think but that's fine.
</blockquote></details>

buggmagnet, I misunderstood that "on" was not part of "on disk". Nevermind then.

PR #4654, opened 2023-05-04
Fix date interval formatting and update tests

pronebird:
When querying multiple date components, `Calendar` tends to collapse to the largest unit. That's the problem because for dates < 2 years we'd only display the number of days that's less than a year, after computing number of whole years. So when formatting a date that's exactly 365 days, we'd print 0 days instead because the largest unit a year in that case.

This PR also clamps the `end` date so that we don't have to handle "0 days" as a special case, instead when `start > end`, the `end = start`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4654)


rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4654)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4642, opened 2023-05-02
Fix Clippy warnings on macOS and run in CI

faern:
Basically a follow up to #4611. Fixes all complaints Clippy had on our Rust code for the macOS target. Also adds macOS into the automatic Clippy check in Github actions to prevent further regressions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4642)

PR #4679, opened 2023-05-11
Fix android clippy warnings

faern:
Follow up to #4642 and #4611. Yet another platform where we more carefully check for idiomatic Rust usage.

I can't make it run `cargo clippy` on the entire workspace. At least not for now. Way too much code in the workspace that simply does not build on Android right now. Some of them are probably not even worth fixing. Might be simpler and more correct to just check the relevant crates (like this) instead.

Some commits has nothing with clippy to do. I just found some opportunities for small refactors and improvements. I saw that we did something *very* unsafe in a not unsafe function. I fixed that and documented it. I hope I got the safety contract correctly specified. Do you know of any reason it was using the arbitrary lifetime `'a` decided by the caller instead of just going with `'static` like I did here now?

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4679)


dlon (commented):
> Reviewed 8 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4679)* status: 8 of 9 files reviewed, all discussions resolved

dlon (approved):
> Reviewed 6 of 7 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4679)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`talpid-routing/src/unix/macos.rs` line 303 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4679#-NWC-NXmDBROtsuqOwmK:-NWC7UHG4-j_9hS3VIHl:bbgdqos) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/34602bd09a3ecef49fb76f1379ae6ae54f93b704/talpid-routing/src/unix/macos.rs#L303)):*

The error type here is different now. Instead of being a `crate::Error` it's now a `crate::imp::PlatformError`. But the only place it seems used is to log the error. And the extra context level of the crate error did not really add anything valuable IMO
</blockquote></details>

Makes sense. Anyway, this will no longer exist after the routing changes.
___
*[`talpid-routing/src/unix/mod.rs` line 2 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4679#-NWC-FkrCpqUPr2L0F0D:-NWC6Ae-7z0jYv5A2eJ3:bplttsy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d5d2bd1d62ec6d52faef631b4c827641fe438728/talpid-routing/src/unix/mod.rs#L2)):*

I have no idea why this was here. Windows does not even include this module.
</blockquote></details>

I'm guessing it was overlooked when this was refactored. :shrug:

PR #4686, opened 2023-05-15
Rework macOS routing

dlon:
Some applications rely on the default route for connectivity instead of using the routing table, which fails because we currently don't create a default route for the tunnel device. The main change in this PR is to replace the default route with a tunnel route and keep track of any changes of the "best interface" to use (using the network service order).

Closes DES-227.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)


pinkisemils (commented):
> Reviewed 1 of 25 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 1 of 28 files reviewed, all discussions resolved

pinkisemils (changes requested):
> Reviewed 5 of 25 files at r2, 1 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 7 of 28 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`talpid-routing/src/unix.rs` line 2 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NVoQ6yH4FI-60sHGr20:-NVoQ6yH4FI-60sHGr21:ba6zr09) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/unix.rs#L2)):*
> ```XML
> #![cfg_attr(target_os = "android", allow(dead_code))]
> #![cfg_attr(target_os = "windows", allow(dead_code))]
> ```

Can this not be changed to conditionally include `crate::Route` only on macOS and linux?
___
*[`talpid-routing/src/macos/data.rs` line 20 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NVoQsjyAVqFWrqH3lhW:-NVoQsjyAVqFWrqH3lhX:b-teaa41) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/data.rs#L20)):*
> ```XML
> }
> 
> impl RouteMessage {
> ```

This impl block is large, maybe `RouteMessage` should be split into 3 - one inner type, one type that is _received_, and one type that'd be used for sending. Not certain if the extra ceremony is a hard requirement, but it might improve the readability of this particular module a lot :thinking:

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 7 of 28 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`talpid-routing/src/macos/mod.rs` line 255 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NVoVCJ_1nXkr3dH4oE5:-NVoVCJ_1nXkr3dH4oE6:b-o98j1c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/mod.rs#L255)):*
> ```XML
>                 }
> 
>                 // TODO: may need to add back tunnel route if it does not exist
> ```

Should this `TODO` not be removed? I don't think we should be anticipating BOFHs removing tunnel routes, should we?
___
*[`talpid-routing/src/macos/mod.rs` line 347 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NVoVSYP0e9nn_OnJ0eU:-NVoVSYP0e9nn_OnJ0eV:b-lmt2kc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/mod.rs#L347)):*
> ```XML
>         // As long as the relay route has a way of reaching the internet, we'll want to add a tunnel
>         // route for both IPv4 and IPv6.
>         // NOTE: This is incorrect. We're assuming that any "default destination" is used for
> ```

I wonder if it'd be better to always supply both the IPv4 and the IPv6 relay address for the tunnel traffic here.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 7 of 28 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[`talpid-routing/src/macos/mod.rs` line 450 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NVoWxg7CyhS9rxmrMz1:-NVoWxg7CyhS9rxmrMz2:baxuomo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/mod.rs#L450)):*
> ```XML
>         let ifscope_index = if should_be_ifscoped {
>             let n = default_route.interface_index();
>             if n == 0 {
> ```

`interface_index`would be a better name, IMO.

pinkisemils (changes requested):
> Reviewed 1 of 25 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 8 of 28 files reviewed, 5 unresolved discussions (waiting on @dlon)

___
*[`talpid-routing/src/macos/routing_socket.rs` line 36 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NW1eTG98n6xIYxKoYd0:-NW1eTG98n6xIYxKoYd1:bqokflp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/routing_socket.rs#L36)):*
> ```XML
> type Result<T> = std::result::Result<T, Error>;
> 
> /// Wraps a PF_ROUTE socket and returns
> ```

This docstring of mine is not helpful.

Should go something like:
```
/// Wraps a `PF_ROUTE` socket, keeps track of sent message IDs, and facilitates sending and receiving [route socket messages](#RouteMessage)
```
___
*[`talpid-routing/src/macos/routing_socket.rs` line 100 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NW1fKvCD55wDP6nhnrH:-NW1fKvCD55wDP6nhnrI:biurf1m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/routing_socket.rs#L100)):*
> ```XML
>     }
> 
>     /// Will panic if the message length ends up overflowing an i32.
> ```

I don't believe this comment is valid either - I can't see it panicking for that. But it should.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 8 of 28 files reviewed, 6 unresolved discussions (waiting on @dlon)

___
*[`talpid-wireguard/src/lib.rs` line 929 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NW1owFv-ANEA8epmaB_:-NW1owFv-ANEA8epmaBa:b69dig8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-wireguard/src/lib.rs#L929)):*
> ```XML
>         }
> 
>         #[cfg(any(target_os = "linux", target_os = "macos", target_os = "android"))]
> ```

Does this just mean `#[cfg(not(windows))]` ?  Same as on line 918.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 8 of 28 files reviewed, 6 unresolved discussions (waiting on @pinkisemils)

___
*[`talpid-routing/src/macos/mod.rs` line 255 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NVoVCJ_1nXkr3dH4oE5:-NW9-qhlARaowH84BjjN:b-2yrng9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/mod.rs#L255)):*

Should this `TODO` not be removed? I don't think we should be anticipating BOFHs removing tunnel routes, should we?
</blockquote></details>

Removed. I could swear I'd seen this occur once in a while before, but I suspect it was due to some other bug that was fixed.
___
*[`talpid-routing/src/macos/mod.rs` line 450 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NVoWxg7CyhS9rxmrMz1:-NW91XMu1sdi2VnNtYdn:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/mod.rs#L450)):*

`interface_index`would be a better name, IMO.
</blockquote></details>

Done.
___
*[`talpid-wireguard/src/lib.rs` line 929 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NW1owFv-ANEA8epmaB_:-NW95O7k77wTUxnfuNvz:byvyago) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-wireguard/src/lib.rs#L929)):*

Does this just mean `#[cfg(not(windows))]` ?  Same as on line 918.
</blockquote></details>

Yes. Done.
___
*[`talpid-routing/src/macos/routing_socket.rs` line 36 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NW1eTG98n6xIYxKoYd0:-NW91zvo0b38yaQS3EYH:b-psav52) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/routing_socket.rs#L36)):*

This docstring of mine is not helpful.

Should go something like:
```
/// Wraps a `PF_ROUTE` socket, keeps track of sent message IDs, and facilitates sending and receiving [route socket messages](#RouteMessage)
```
</blockquote></details>

That was actually my unhelpful docstring. ;)

Done.
___
*[`talpid-routing/src/macos/routing_socket.rs` line 100 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NW1fKvCD55wDP6nhnrH:-NW92f8N6t1NBeujeVFn:bt5erxl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/routing_socket.rs#L100)):*

I don't believe this comment is valid either - I can't see it panicking for that. But it should. 
</blockquote></details>

Is this not referring to `RouteMessage::payload` panicking if `rtm_msglen` is too small to contain the payload? If so, that one does panic, but (1) this docstring belongs to `RouteMessage::payload`, and (2) `rtm_msglen` is a `c_ushort`, which I think is just 16 bits.

If it's referring to the sequence number, I don't see the point of panicking if it overflows.

pinkisemils (changes requested):
> Reviewed 8 of 25 files at r2, 1 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 17 of 28 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`talpid-routing/src/macos/routing_socket.rs` line 100 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NW1fKvCD55wDP6nhnrH:-NWBkhY5EQlFDqTw3iJE:bf0w1hq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/routing_socket.rs#L100)):*

Is this not referring to `RouteMessage::payload` panicking if `rtm_msglen` is too small to contain the payload? If so, that one does panic, but (1) this docstring belongs to `RouteMessage::payload`, and (2) `rtm_msglen` is a `c_ushort`, which I think is just 16 bits.

If it's referring to the sequence number, I don't see the point of panicking if it overflows.
</blockquote></details>

It won't panic since it's a wrapping addition here. It might be a good idea to add some more sanity checks here, but I can't really point out which _size_ variables should be checked against what here :/

Jontified (changes requested):
> Reviewed 21 of 25 files at r2, 3 of 5 files at r3, 1 of 1 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: all files reviewed, 26 unresolved discussions (waiting on @dlon)

___
*[`talpid-core/src/offline/macos.rs` line 7 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWMJbrvE3_pzTjNQxmN:-NWMJbrvE3_pzTjNQxmO:b81ysma) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-core/src/offline/macos.rs#L7)):*
> ```XML
> //! Currently, this functionality is implemented by using `route monitor -n` to observe routing
> //! table changes and then use the CLI once more to query if there exists a default route.
> //! Generally, it is assumed that a machine is online if there exists a route to a public IP
> ```

Perhaps it would be cool to explain why this is the criteria for being considered online?
___
*[`talpid-core/src/offline/macos.rs` line 75 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWMEFFxBEIyv0HBHLu-:-NWMEFFxBEIyv0HBHLu0:bflz9r0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-core/src/offline/macos.rs#L75)):*
> ```XML
>         Ok((v4_route, v6_route)) => (v4_route.is_some(), v6_route.is_some()),
>         Err(error) => {
>             log::warn!("Failed to initialize offline monitor: {error}");
> ```

It's not immediately obvious to me why "Failing to initialize offline monitor" does not return an `Err` and then sets the connectivity on both routes to `true`. Could we clarify in a comment?
___
*[`talpid-core/src/offline/macos.rs` line 155 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWMPJo25VMYV3gBdMIw:-NWMPJo25VMYV3gBdMIx:b2ovldh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-core/src/offline/macos.rs#L155)):*
> ```XML
>                     );
> 
>                     let _ = notify_tx.unbounded_send(!new_connectivity);
> ```

Why are we sending `!new_connectivity` here? Shouldn't we be sending `new_connectivity`?
___
*[`talpid-routing/src/unix.rs` line 200 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWM4mlL9tzKtYgqKxTx:-NWM4mlL9tzKtYgqKxTy:b-8mw5ac) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3821d62c16eda5141f7b4e5eba582d90966f4e2/talpid-routing/src/unix.rs#L200)):*
> ```XML
> 
> impl RouteManager {
>     /// Constructs a RouteManager and applies the required routes.
> ```

Update this comment, we are no longer applying required routes.
___
*[`talpid-routing/src/macos/mod.rs` line 76 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWpqOQvBogfavSvH6mQ:-NWpqOQvBogfavSvH6mR:b-ds11x5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L76)):*
> ```XML
>         Ok(Self {
>             routing_table,
>             default_destinations: HashSet::new(),
> ```

Could this be renamed? `non_tunnel_destinations`? Perhaps also a comment to explain what it is?
If we end up renaming this we should also rename `apply_default_destinations` in the same file.
___
*[`talpid-routing/src/macos/mod.rs` line 305 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWblcjY3s8uhr5qCKng:-NWblcjZ3SRGphrdLKWZ:bbum7e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L305)):*
> ```XML
>     async fn update_best_default_route(&mut self, family: interface::Family) -> Result<()> {
>         let best_route = interface::get_best_default_route(&mut self.routing_table, family).await;
>         // FIXME: remove logging
> ```

Unfixed comment?
___
*[`talpid-routing/src/macos/mod.rs` line 346 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWpwE5L5hYoG-EMKng7:-NWpwE5L5hYoG-EMKng8:b-mgj3fs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L346)):*
> ```XML
>         // As long as the relay route has a way of reaching the internet, we'll want to add a tunnel
>         // route for both IPv4 and IPv6.
>         // NOTE: This is incorrect. We're assuming that any "default destination" is used for
> ```

What happens if this assumption is wrong?
___
*[`talpid-routing/src/macos/mod.rs` line 354 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWpwRYm8YOet7QooXS3:-NWpwRYm8YOet7QooXS4:b-2i11i0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L354)):*
> ```XML
>                 (v4 || route.is_ipv4(), v6 || route.is_ipv6())
>             });
>         let relay_route_is_valid = (v4_conn && self.v4_default_route.is_some())
> ```

What is valid trying to represent here?
___
*[`talpid-routing/src/macos/mod.rs` line 444 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWfaDw90S4L4Ehlm7Xk:-NWfaDw90S4L4Ehlm7Xl:b-8n0omw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L444)):*
> ```XML
>         }
> 
>         // FIXME: remove logging
> ```

Dead comment
___
*[`talpid-routing/src/macos/mod.rs` line 457 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWpxZulAX7qWuzQQkbH:-NWpxZulAX7qWuzQQkbI:bytm23t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L457)):*
> ```XML
>         };
>         let new_route = default_route.clone().set_ifscope(interface_index);
>         let old_route = std::mem::replace(default_route, new_route);
> ```

This might cause inconsistencies with the routing table since the following 2 calls might fail. I'm not sure how far we want to go to prevent these types of inconsistencies because they arguably could happen all over the place?
___
*[`talpid-routing/src/macos/mod.rs` line 476 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWputXo45rTh_g9SfLo:-NWputXo45rTh_g9SfLp:bdczrqj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L476)):*
> ```XML
>             .map_err(Error::AddRoute)?;
> 
>         let destination = RouteDestination::try_from(&route).map_err(Error::InvalidData)?;
> ```

If this fails we'll have an inconsistency with the routing table. If we want to avoid this inconsistency we should either assert that the `try_from` can't fail here or roll back the route table.

Perhaps the best way to deal with this is to instead of building a large roll-back module or crashing instead we just log that potentially the routing module is inconsistent with the actual routing table.
This goes for the other instances of inconsistency comments.
___
*[`talpid-routing/src/macos/watch.rs` line 62 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWbfrjo-KneN1_5Q_CF:-NWbfrjo-KneN1_5Q_CG:blji0yt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/watch.rs#L62)):*
> ```XML
> 
>         match msg {
>             Ok(RouteSocketMessage::AddRoute(_route)) => Ok(()),
> ```

Check `_route.errno()`?
___
*[`talpid-routing/src/macos/data.rs` line 88 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWaTtg443_zNBGd3iep:-NWaTtg443_zNBGd3ieq:bnh7u8h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L88)):*
> ```XML
>             Some(None) => true,
>             Some(Some(addr)) => {
>                 if let Some(netmask_addr) = addr.as_sockaddr_in() {
> ```

Use `saddr_to_ipv4()`? and get rid of clone?
___
*[`talpid-routing/src/macos/data.rs` line 288 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWabi3WA8438b78_O-I:-NWabi3WA8438b78_O-J:b6chf1n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L288)):*
> ```XML
> 
>         let mut sockaddrs = self.route_addrs().collect::<Vec<_>>();
>         sockaddrs.sort_by_key(|saddr| saddr.address_flag());
> ```

Could we explain why we need to sort here?
___
*[`talpid-routing/src/macos/data.rs` line 296 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWabReRBTRv_HFlAUbN:-NWabReRBTRv_HFlAUbO:b-2gep37) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L296)):*
> ```XML
>         let payload_len: usize = payload_bytes.iter().map(Vec::len).sum();
> 
>         let rtm_msglen = (payload_len + std::mem::size_of::<super::data::rt_msghdr>())
> ```

Use `ROUTE_MESSAGE_HEADER_SIZE`
___
*[`talpid-routing/src/macos/data.rs` line 377 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWbkv9EEG5wT3AAmVxC:-NWbkv9EEG5wT3AAmVxD:b-rc9dvc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L377)):*
> ```XML
>             self.route_flags.insert(RouteFlag::RTF_IFSCOPE);
>         } else {
>             // self.interface_index = iface_index;
> ```

Dead comment? This should be uncommented I think?
___
*[`talpid-routing/src/macos/data.rs` line 442 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWaeUY52Y9MQ9L5-9O0:-NWaeUY52Y9MQ9L5-9O1:btmbk88) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L442)):*
> ```XML
>         if msg_len > buffer.len() {
>             return Err(Error::BufferTooSmall(
>                 "Mesage is shorter than it's msg_len indicates",
> ```

Message*
___
*[`talpid-routing/src/macos/data.rs` line 561 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWagRma0NSYGFsgEO7y:-NWagRma0NSYGFsgEO7z:b-hu2eng) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L561)):*
> ```XML
> 
>         match rt_msghdr_short::from_bytes(buffer) {
>             Some(header) if header.is_type(libc::RTM_ADD) => route_message(Self::AddRoute, buffer),
> ```

nit: inconsistent formatting
___
*[`talpid-routing/src/macos/data.rs` line 668 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWagjzsC8kQS09GoCFU:-NWagjzsC8kQS09GoCFV:b-me4961) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L668)):*
> ```XML
> }
> 
> impl Interface {
> ```

Why 2 implementations of `Interface`?
___
*[`talpid-routing/src/macos/data.rs` line 695 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWaguKu4kDVBgWM4eBX:-NWaguKu4kDVBgWM4eBY:bpaeeio) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L695)):*
> ```XML
>     /// All enum values of address flags can be iterated via `flag <<= 1`, starting from 1.
>     /// See https://www.manpagez.com/man/4/route/.
>     // #[derive(Clone, Copy, PartialOrd)]
> ```

Commented out derive
___
*[`talpid-routing/src/macos/data.rs` line 719 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWags2n5KzC7XiBj8Vc:-NWags2n5KzC7XiBj8Vd:bpaeeio) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L719)):*
> ```XML
>     /// Types of routing messages
>     /// See https://www.manpagez.com/man/4/route/.
>     // #[derive(Clone, Copy, PartialOrd)]
> ```

Commented out derive
___
*[`talpid-routing/src/macos/data.rs` line 759 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWMItu0AAdCXYo28J1Q:-NWMItu0AAdCXYo28J1R:b-8n0omw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L759)):*
> ```XML
>     /// Routing message flags
>     /// See https://www.manpagez.com/man/4/route/.
>     // #[derive(Clone, Copy, PartialOrd)]
> ```

Dead comment
___
*[`talpid-routing/src/macos/data.rs` line 892 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWakfu1D54HfsYSGPEg:-NWakfu1D54HfsYSGPEh:b-sj8ucn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L892)):*
> ```XML
>         match self.inner() {
>             None => vec![0u8; 4],
>             Some(addr) => {
> ```

Add invariant for the case where `addr.len() < 4`? Should be `unreachable!()`
___
*[`talpid-routing/src/macos/data.rs` line 904 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWaspgC01QIyyPHs4Th:-NWaspgC01QIyyPHs4Ti:b-4dseiu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L904)):*
> ```XML
>                         addr.as_ptr() as *const _,
>                         buffer.as_mut_ptr(),
>                         addr.len() as usize,
> ```

reuse `len`
___
*[`talpid-routing/src/macos/data.rs` line 1097 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWav3eeExf6ZR-cH-ku:-NWav3eeExf6ZR-cH-kv:b-2gep37) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L1097)):*
> ```XML
> impl rt_msghdr {
>     pub fn from_bytes(buf: &[u8]) -> Result<Self> {
>         if buf.len() >= std::mem::size_of::<rt_msghdr>() {
> ```

Use `ROUTE_MESSAGE_HEADER_SIZE`
___
*[`talpid-routing/src/macos/routing_socket.rs` line 108 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWbdPao9WntFLq_Fe37:-NWbdPao9WntFLq_Fe38:b3pp9zp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/routing_socket.rs#L108)):*
> ```XML
>         let (header, payload) = message.payload(msg_type, seq, self.own_pid);
>         let mut msg_buffer = vec![0u8; header.rtm_msglen.into()];
> ```

Consider adding an assert for making sure that `msg_buffer.len()` is bigger than `size_of::<super::data::rt_msghdr>()`?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 0 of 41 files reviewed, 24 unresolved discussions (waiting on @Jontified and @pinkisemils)

___
*[`talpid-core/src/offline/macos.rs` line 75 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWMEFFxBEIyv0HBHLu-:-NXAVTqq2vErhst5ScSd:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-core/src/offline/macos.rs#L75)):*

It's not immediately obvious to me why "Failing to initialize offline monitor" does not return an `Err` and then sets the connectivity on both routes to `true`. Could we clarify in a comment?
</blockquote></details>

Done.
___
*[`talpid-core/src/offline/macos.rs` line 155 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWMPJo25VMYV3gBdMIw:-NXAQz-HCbOMGhxSBwuG:b-kkiw5l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-core/src/offline/macos.rs#L155)):*

Why are we sending `!new_connectivity` here? Shouldn't we be sending `new_connectivity`?
</blockquote></details>

Because the channel wants to know if we're offline, not online. So it should be negated.
___
*[`talpid-routing/src/unix.rs` line 2 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NVoQ6yH4FI-60sHGr20:-NXATO0QDMm5BzQFrpEr:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/unix.rs#L2)):*

Can this not be changed to conditionally include `crate::Route` only on macOS and linux?
</blockquote></details>

Done.
___
*[`talpid-routing/src/unix.rs` line 200 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWM4mlL9tzKtYgqKxTx:-NXAVXeFAj828JrNAlqD:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3821d62c16eda5141f7b4e5eba582d90966f4e2/talpid-routing/src/unix.rs#L200)):*

Update this comment, we are no longer applying required routes.
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 88 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWaTtg443_zNBGd3iep:-NXARDQj1XHw4Jw8hr4S:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L88)):*

Use `saddr_to_ipv4()`? and get rid of clone?
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 288 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWabi3WA8438b78_O-I:-NXAREG68IkDMKp2mOmh:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L288)):*

Could we explain why we need to sort here?
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 296 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWabReRBTRv_HFlAUbN:-NXARYJZ1FVizlFazFHE:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L296)):*

Use `ROUTE_MESSAGE_HEADER_SIZE`
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 377 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWbkv9EEG5wT3AAmVxC:-NXARZ1t41qs_RxHnEY0:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L377)):*

Dead comment? This should be uncommented I think?
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 442 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWaeUY52Y9MQ9L5-9O0:-NXAR_Uv97soveHohe7G:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L442)):*

Message*
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 668 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWagjzsC8kQS09GoCFU:-NXARbxn4J8IUYSkbUAh:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L668)):*

Why 2 implementations of `Interface`?
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 695 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWaguKu4kDVBgWM4eBX:-NXARcey8WdzGsEjXIDF:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L695)):*

Commented out derive
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 719 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWags2n5KzC7XiBj8Vc:-NXARcxF35KFlKnX4igz:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L719)):*

Commented out derive
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 759 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWMItu0AAdCXYo28J1Q:-NXARdI3BKrY9bONtaCI:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L759)):*

Dead comment
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 892 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWakfu1D54HfsYSGPEg:-NXARdwbDzaq9ShaSgL3:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L892)):*

Add invariant for the case where `addr.len() < 4`? Should be `unreachable!()`
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 904 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWaspgC01QIyyPHs4Th:-NXARf7O5g7si6Rxydek:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L904)):*

reuse `len`
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/data.rs` line 1097 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWav3eeExf6ZR-cH-ku:-NXARg4O0s1rNkO0qTrJ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/data.rs#L1097)):*

Use `ROUTE_MESSAGE_HEADER_SIZE`
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/mod.rs` line 347 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NVoVSYP0e9nn_OnJ0eU:-NW915NLCylSNJbHw1jn:b-ow2ppm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/mod.rs#L347)):*

I wonder if it'd be better to always supply both the IPv4 and the IPv6 relay address for the tunnel traffic here.
</blockquote></details>

Does that make any sense if there is no IPv6 connectivity?
___
*[`talpid-routing/src/macos/routing_socket.rs` line 100 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NW1fKvCD55wDP6nhnrH:-NXATFD-A0Ba4IcjHaTJ:b-eouddr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/81cd99efd6cdae2aac121798d754c82d59a1d3eb/talpid-routing/src/macos/routing_socket.rs#L100)):*

It won't panic since it's a wrapping addition here. It might be a good idea to add some more sanity checks here, but I can't really point out which _size_ variables should be checked against what here :/ 
</blockquote></details>

I've removed the comment. I intentionally added the wrapping, because reusing numbers caused weird behavior at times, but it looks correct to me.
___
*[`talpid-routing/src/macos/routing_socket.rs` line 108 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWbdPao9WntFLq_Fe37:-NXAPuk1EIFef6EHVgEH:b-y3agxl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/routing_socket.rs#L108)):*

Consider adding an assert for making sure that `msg_buffer.len()` is bigger than `size_of::<super::data::rt_msghdr>()`?
</blockquote></details>

That's already guaranteed by `RouteMessage::payload`, so I think it's unnecessary.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 0 of 41 files reviewed, 24 unresolved discussions (waiting on @Jontified and @pinkisemils)

___
*[`talpid-routing/src/macos/mod.rs` line 76 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWpqOQvBogfavSvH6mQ:-NXAbet75FHpcww0rQ_G:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L76)):*

Could this be renamed? `non_tunnel_destinations`? Perhaps also a comment to explain what it is?
If we end up renaming this we should also rename `apply_default_destinations` in the same file.
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/mod.rs` line 346 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWpwE5L5hYoG-EMKng7:-NXAW82LBxC1VmWT4yAX:bf28o27) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L346)):*

What happens if this assumption is wrong?
</blockquote></details>

The default route will be replaced with a tunnel route, even though there's no relay route. I don't think that the assumption is wrong, though. It would be incorrect to not add a relay route.
___
*[`talpid-routing/src/macos/mod.rs` line 354 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWpwRYm8YOet7QooXS3:-NXAW0jx5UkI3SqWZBUa:bnxudxz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L354)):*

What is valid trying to represent here?
</blockquote></details>

The comment above explains it.
___
*[`talpid-routing/src/macos/mod.rs` line 457 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWpxZulAX7qWuzQQkbH:-NXAVn5n37eA7JFmL4la:bvogtyk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L457)):*

This might cause inconsistencies with the routing table since the following 2 calls might fail. I'm not sure how far we want to go to prevent these types of inconsistencies because they arguably could happen all over the place?
</blockquote></details>

I think it will be inconsistent no matter how you do it, unless you add rollback like you said. But that's a pretty big change and we haven't seen a need for that so far. And even that could fail.

If you delete the old route here, but adding a new route fails, then the only way to make it consistent again is to add back the deleted route.
___
*[`talpid-routing/src/macos/mod.rs` line 476 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWputXo45rTh_g9SfLo:-NXAXC5G96cbOz-H1oJz:bxw6eig) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/mod.rs#L476)):*

If this fails we'll have an inconsistency with the routing table. If we want to avoid this inconsistency we should either assert that the `try_from` can't fail here or roll back the route table.

Perhaps the best way to deal with this is to instead of building a large roll-back module or crashing instead we just log that potentially the routing module is inconsistent with the actual routing table.
This goes for the other instances of inconsistency comments.
</blockquote></details>

Fixed by figuring running `try_from` before adding the route.

pinkisemils (commented):
> Reviewed 5 of 29 files at r2, 1 of 5 files at r3, 1 of 1 files at r4, 10 of 27 files at r9, 2 of 6 files at r10.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 12 of 41 files reviewed, 22 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`talpid-routing/src/macos/routing_socket.rs` line 108 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWbdPao9WntFLq_Fe37:-NXAcn_G7pKVJgwXe_76:b-div75o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/routing_socket.rs#L108)):*

That's already guaranteed by `RouteMessage::payload`, so I think it's unnecessary.
</blockquote></details>

I think this assertion should be documented here as a `// SAFETY: ..` comment.

pinkisemils (changes requested):
> Reviewed 2 of 27 files at r9, 2 of 6 files at r10, 1 of 1 files at r11.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 17 of 41 files reviewed, 25 unresolved discussions (waiting on @dlon and @Jontified)

___
*[`talpid-routing/src/unix/macos/mod.rs` line 296 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NXAeyzK4N8-gfFuo5oy:-NXAeyzK4N8-gfFuo5oz:bdqbgfg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/47cbc854c5351ad8f920756d16a2cacb4b67420c/talpid-routing/src/unix/macos/mod.rs#L296)):*
> ```XML
>         }
> 
>         self.update_best_default_route(if route.is_ipv4() {
> ```

I think this could be improved by moving the inner expression out and assigning it to a variable.
```rust
let ip_version = if route.is_ipv4() { interface::Family::V4 } else { interface::Family::V6 };
self.update_best_default_route(ip_version);
```
Like that ^, but with `rustfmt`. 
___
*[`talpid-routing/src/unix/macos/mod.rs` line 514 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NXAfUpB0jqLLl90iNeG:-NXAfUpB0jqLLl90iNeH:b-tu39hj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/47cbc854c5351ad8f920756d16a2cacb4b67420c/talpid-routing/src/unix/macos/mod.rs#L514)):*
> ```XML
> 
> fn v4_default() -> IpNetwork {
>     IpNetwork::new(Ipv4Addr::UNSPECIFIED.into(), 0).unwrap()
> ```

This code is duplicated in `interface.rs`. `v4_default` and `v6_default` could be lifted further up, but it should be consolidated.
___
*[`talpid-routing/src/unix/macos/watch.rs` line 12 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NXAgf-UEFVjnVkix9xK:-NXAgf-UEFVjnVkix9xL:bpmkpb6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/47cbc854c5351ad8f920756d16a2cacb4b67420c/talpid-routing/src/unix/macos/watch.rs#L12)):*
> ```XML
> pub enum Error {
>     #[error(display = "Failed to resolve interface name to index")]
>     ResolveInterfaceName(nix::Error),
> ```

I don't believe this error type is used anywhere.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 28 of 41 files reviewed, 25 unresolved discussions (waiting on @Jontified and @pinkisemils)

___
*[`talpid-routing/src/unix/macos/mod.rs` line 296 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NXAeyzK4N8-gfFuo5oy:-NXAmJnK0EVGE1kqJ1om:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/47cbc854c5351ad8f920756d16a2cacb4b67420c/talpid-routing/src/unix/macos/mod.rs#L296)):*

I think this could be improved by moving the inner expression out and assigning it to a variable.
```rust
let ip_version = if route.is_ipv4() { interface::Family::V4 } else { interface::Family::V6 };
self.update_best_default_route(ip_version);
```
Like that ^, but with `rustfmt`. 
</blockquote></details>

Done.
___
*[`talpid-routing/src/unix/macos/mod.rs` line 514 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NXAfUpB0jqLLl90iNeG:-NXAmMW37eFW8peAeaH-:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/47cbc854c5351ad8f920756d16a2cacb4b67420c/talpid-routing/src/unix/macos/mod.rs#L514)):*

This code is duplicated in `interface.rs`. `v4_default` and `v6_default` could be lifted further up, but it should be consolidated.
</blockquote></details>

Done.
___
*[`talpid-routing/src/unix/macos/watch.rs` line 12 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NXAgf-UEFVjnVkix9xK:-NXAmZEWCiP45wBShcS2:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/47cbc854c5351ad8f920756d16a2cacb4b67420c/talpid-routing/src/unix/macos/watch.rs#L12)):*

I don't believe this error type is used anywhere.
</blockquote></details>

Done.
___
*[`talpid-routing/src/macos/routing_socket.rs` line 108 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-NWbdPao9WntFLq_Fe37:-NXAmPXM9U6CpKdyy-va:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d452aa9487473da4f5b39e88a8126963ab9727ae/talpid-routing/src/macos/routing_socket.rs#L108)):*

I think this assertion should be documented here as a `// SAFETY: ..` comment.
</blockquote></details>

Done.

pinkisemils (approved):
> Reviewed 2 of 36 files at r6, 2 of 37 files at r7, 4 of 27 files at r9, 3 of 5 files at r12, 1 of 1 files at r13, 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: all files reviewed, 22 unresolved discussions (waiting on @dlon and @Jontified)

pinkisemils (approved):
> Reviewed 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: all files reviewed, 22 unresolved discussions (waiting on @dlon and @Jontified)

Jontified (approved):
:lgtm:

> Reviewed 4 of 36 files at r6, 10 of 37 files at r7, 19 of 27 files at r9, 4 of 6 files at r10, 3 of 5 files at r12, 1 of 1 files at r13, 1 of 1 files at r14, 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686#-:-NXBAY_R1TI77MV0tLZY:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (commented):
> Reviewed 4 of 36 files at r6, 10 of 37 files at r7, 27 of 27 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4686)* status: 14 of 41 files reviewed, all discussions resolved

PR #4689, opened 2023-05-15
Increase the touch target for the cross button in the banner view

mojganii:
- Make action button size bigger
- Make the condition for the showing device info easier

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4689)

### DIFF
--- a/ios/MullvadVPN/Coordinators/App/ApplicationCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/App/ApplicationCoordinator.swift
@@ -708,9 +708,9 @@ final class ApplicationCoordinator: Coordinator, Presenting, RootContainerViewCo
 
         switch deviceState {
         case let .loggedIn(accountData, _):
-            let shouldHideDeviceInfo = previousDeviceState == .loggedOut
+            let showsDeviceInfo = previousDeviceState != .loggedOut
             updateOutOfTimeTimer(accountData: accountData)
-            updateView(deviceState: deviceState, showDeviceInfo: !shouldHideDeviceInfo)
+            updateView(deviceState: deviceState, showDeviceInfo: showsDeviceInfo)
 
             // Handle transition to and from expired state.
             let accountWasExpired = previousDeviceState.accountData?.isExpired ?? false
--- a/ios/MullvadVPN/Notifications/UI/NotificationBannerView.swift
+++ b/ios/MullvadVPN/Notifications/UI/NotificationBannerView.swift
@@ -10,7 +10,7 @@ import UIKit
 
 final class NotificationBannerView: UIView {
     private static let indicatorViewSize = CGSize(width: 12, height: 12)
-    private static let buttonSize = CGSize(width: 18, height: 18)
+    private static let buttonSize = CGSize(width: 32, height: 32)
 
     private let backgroundView: UIVisualEffectView = {
         let effect = UIBlurEffect(style: .dark)
@@ -63,10 +63,11 @@ final class NotificationBannerView: UIView {
         return view
     }()
 
-    private let actionButton: UIButton = {
-        let button = UIButton(type: .system)
+    private let actionButton: CustomButton = {
+        let button = CustomButton()
         button.tintColor = UIColor.InAppNotificationBanner.actionButtonColor
         button.translatesAutoresizingMaskIntoConstraints = false
+        button.imageAlignment = .leadingFixed
         return button
     }()
 


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4689)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/Coordinators/App/ApplicationCoordinator.swift` line 711 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4689#-NVV6Qqq7Sa2bKJ2YJW9:-NVV6Qqq7Sa2bKJ2YJWA:b-4eu9o1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5570852a05fcb0c46e7de36cef8151462f3be614/ios/MullvadVPN/Coordinators/App/ApplicationCoordinator.swift#L711)):*
> ```Swift
>         switch deviceState {
>         case let .loggedIn(accountData, _):
>             let showsDeviceInfo = previousDeviceState != .loggedOut
> ```

Shouldn't we only show device info once user clicked "(x)" button in notification banner? 

Otherwise any update to account data could trigger situation where `previousDeviceState` is `loggedIn`, i.e:

```swift
previousDeviceState == .loggedIn(<old account data>, _) && deviceState == .loggedIn(<new account data>, _)
```

Ideally it should be:

```swift
var showsDeviceInfo: Bool = true

private func deviceStateDidChange(_ deviceState: DeviceState, previousDeviceState: DeviceState) {
    if deviceState.isLoggedIn, previousDeviceState == .loggedOut {
        // Device is created!   
        // Transitioning from loggedOut -> loggedIn state
        showsDeviceInfo = false
    } else if deviceState == .loggedOut || deviceState == .revoked {
        // Device is destroyed at this point!
        showsDeviceInfo = true
    }

    // redacted code
}

func notificationManager(_ manager: NotificationManager, didReceiveResponse response: NotificationResponse) {
    switch response.providerIdentifier {
        // redacted code

        case .registeredDeviceInAppNotification:
            showsDeviceInfo = true

            updateView(deviceState: tunnelManager.deviceState)

        default: return
    }
}
```
___
*[`ios/MullvadVPN/Notifications/UI/NotificationBannerView.swift` line 67 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4689#-NVV4n7D30umY8rl82GV:-NVV4n7D30umY8rl82GW:b-zgmnea) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5570852a05fcb0c46e7de36cef8151462f3be614/ios/MullvadVPN/Notifications/UI/NotificationBannerView.swift#L67)):*
> ```Swift
> 
>     private let actionButton: CustomButton = {
>         let button = CustomButton()
> ```

I don't exactly know what role the CustomButton plays here.

But I spoke about that with Matilda earlier today. My first suggestion was to change hit area detection, i.e with subclass:

```swift
class NotificationActionButton: UIButton {
    override func point(inside point: CGPoint, with event: UIEvent?) -> Bool {
        return bounds.insetBy(dx: -10, dy: -10).contains(point)
    }
}
```

Matilda on the other side used constraints and set button size to 44x44 and then offset the button by 14pt in horizontal and vertical direction which is a working solution as well.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4689)* status: all files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN/Coordinators/App/ApplicationCoordinator.swift` line 711 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4689#-NVV6Qqq7Sa2bKJ2YJW9:-NVVQTi8-KYoztHJxeND:b-xdva5n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5570852a05fcb0c46e7de36cef8151462f3be614/ios/MullvadVPN/Coordinators/App/ApplicationCoordinator.swift#L711)):*

Shouldn't we only show device info once user clicked "(x)" button in notification banner? 

Otherwise any update to account data could trigger situation where `previousDeviceState` is `loggedIn`, i.e:

```swift
previousDeviceState == .loggedIn(<old account data>, _) && deviceState == .loggedIn(<new account data>, _)
```

Ideally it should be:

```swift
var showsDeviceInfo: Bool = true

private func deviceStateDidChange(_ deviceState: DeviceState, previousDeviceState: DeviceState) {
    if deviceState.isLoggedIn, previousDeviceState == .loggedOut {
        // Device is created!   
        // Transitioning from loggedOut -> loggedIn state
        showsDeviceInfo = false
    } else if deviceState == .loggedOut || deviceState == .revoked {
        // Device is destroyed at this point!
        showsDeviceInfo = true
    }

    // redacted code
}

func notificationManager(_ manager: NotificationManager, didReceiveResponse response: NotificationResponse) {
    switch response.providerIdentifier {
        // redacted code

        case .registeredDeviceInAppNotification:
            showsDeviceInfo = true

            updateView(deviceState: tunnelManager.deviceState)

        default: return
    }
}
```
</blockquote></details>

I think there is no need to make the function for showing device info
___
*[`ios/MullvadVPN/Notifications/UI/NotificationBannerView.swift` line 67 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4689#-NVV4n7D30umY8rl82GV:-NVVQihy3I8jfgDpMUXm:b-g8w3ko) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5570852a05fcb0c46e7de36cef8151462f3be614/ios/MullvadVPN/Notifications/UI/NotificationBannerView.swift#L67)):*

I don't exactly know what role the CustomButton plays here.

But I spoke about that with Matilda earlier today. My first suggestion was to change hit area detection, i.e with subclass:

```swift
class NotificationActionButton: UIButton {
    override func point(inside point: CGPoint, with event: UIEvent?) -> Bool {
        return bounds.insetBy(dx: -10, dy: -10).contains(point)
    }
}
```

Matilda on the other side used constraints and set button size to 44x44 and then offset the button by 14pt in horizontal and vertical direction which is a working solution as well.
</blockquote></details>

I strongly believe  that it's not a good idea. maybe it can solve our problem for now but when we put another UIControl beside this button. we are making the touch area mix up.what is the problem with the button size?
moreover,I used custom button for taking advantage of imageAlignment.but Matilda consists on using something like you mentioned in the PR comment.but I think it caused overlapping between text and button.

PR #4709, opened 2023-05-19
Add alert if user disconnects when there's no connectivity

rablador:
If the app has no network access and the user disconnects the tunnel, the "Secure my connection" and "Switch location" buttons will be disabled until network access is restored. This puts the user in a vulnerable state and leaking traffic. To make sure the user understands this and does not accidentally disconnect, a new alert will be shown before disconnecting the tunnel.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709)

### DIFF
--- a/ios/MullvadVPN/Coordinators/App/TunnelCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/App/TunnelCoordinator.swift
@@ -11,6 +11,7 @@ import UIKit
 class TunnelCoordinator: Coordinator {
     private let tunnelManager: TunnelManager
     private let controller: TunnelViewController
+    private let alertPresenter = AlertPresenter()
 
     private var tunnelObserver: TunnelObserver?
 
@@ -31,6 +32,10 @@ class TunnelCoordinator: Coordinator {
         controller.shouldShowSelectLocationPicker = { [weak self] in
             self?.showSelectLocationPicker?()
         }
+
+        controller.shouldShowCancelTunnelAlert = { [weak self] in
+            self?.showCancelTunnelAlert()
+        }
     }
 
     func start() {
@@ -50,5 +55,46 @@ class TunnelCoordinator: Coordinator {
         let deviceState = tunnelManager.deviceState
 
         controller.setMainContentHidden(!deviceState.isLoggedIn, animated: animated)
+    }
+
+    private func showCancelTunnelAlert() {
+        let alertController = UIAlertController(
+            title: nil,
+            message: NSLocalizedString(
+                "CANCEL_TUNNEL_ALERT_MESSAGE",
+                tableName: "Main",
+                value: "If you disconnect now, you won’t be able to secure your connection until the device is online.",
+                comment: ""
+            ),
+            preferredStyle: .alert
+        )
+
+        alertController.addAction(
+            UIAlertAction(
+                title: NSLocalizedString(
+                    "CANCEL_TUNNEL_ALERT_DISCONNECT_ACTION",
+                    tableName: "Main",
+                    value: "Disconnect",
+                    comment: ""
+                ),
+                style: .destructive,
+                handler: { [weak self] _ in
+                    self?.tunnelManager.stopTunnel()
+                }
+            )
+        )
+
+        alertController.addAction(
+            UIAlertAction(
+                title: NSLocalizedString(
+                    "CANCEL_TUNNEL_ALERT_CANCEL_ACTION",
+                    tableName: "Main",
+                    value: "Cancel",
+                    comment: ""
+                ), style: .cancel
+            )
+        )
+
+        alertPresenter.enqueue(alertController, presentingController: rootViewController)
     }
 }
--- a/ios/MullvadVPN/View controllers/Tunnel/TunnelControlView.swift
+++ b/ios/MullvadVPN/View controllers/Tunnel/TunnelControlView.swift
@@ -187,7 +187,7 @@ final class TunnelControlView: UIView {
             NSLocalizedString(
                 "CANCEL_BUTTON_TITLE",
                 tableName: "Main",
-                value: "Cancel",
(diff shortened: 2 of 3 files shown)


buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709#-:-NVnr6twCWuHhrWMpYjb:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709)* status: 0 of 2 files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709#-NW1xGF84u_-TwXO6MDc:-NW1xGF84u_-TwXO6MDd:bh62do3) (no related file):*
FYI

1. If user re-enters the app from push notification, we may attempt to present account view above alert which probably wouldn't work. Ideally we should somehow make router aware of alerts when they are presented and perhaps make them exclusive. This would certainly spill into some architectural changes.
2. It would be better to handle navigation within coordinator instead of presenting alert view controller from within view controller itself. While some part of the code still does things the old fashion way, it's certainly something I thought we should move away from and in favor of using coordinators.

___

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709)* status: 0 of 2 files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709#-NW1xGF84u_-TwXO6MDc:-NW2Iavb5OxIw8GD2uU2:bqzovcl) (no related file):*

FYI

1. If user re-enters the app from push notification, we may attempt to present account view above alert which probably wouldn't work. Ideally we should somehow make router aware of alerts when they are presented and perhaps make them exclusive. This would certainly spill into some architectural changes.
2. It would be better to handle navigation within coordinator instead of presenting alert view controller from within view controller itself. While some part of the code still does things the old fashion way, it's certainly something I thought we should move away from and in favor of using coordinators.
</blockquote></details>

1. Should we perhaps defer this to another issue?
2. Alright, so delegate alert presentation to the coordinator and present on `rootViewController`?

___

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709)* status: 0 of 3 files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709#-NW1xGF84u_-TwXO6MDc:-NW7Km7u3C_LfrG4nbND:b-nktoji) (no related file):*

1. Should we perhaps defer this to another issue?
2. Alright, so delegate alert presentation to the coordinator and present on `rootViewController`?
</blockquote></details>

1. Perhaps but we definitely should add a task to handle alerts in some way.
2. Yes.

___

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

mojganii (commented):
when the app runs and there is no connectivity we saw in app banner that tells to user doesn't has access to the network and the relevant buttons are disabled that looks great but when I connect the network the app still stays on disabled mode and it doesn't update UI and dismiss the banner.I think you should fix this part.
Moreover, maybe it's not the scope of this task but something like this goes for tunnel as well.I connect via tunnel and I disconnect the internet and the app shows me I have a secure connection while there is no connectivity.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709#-:-NWBs6VMBszSt1jvYIST:bvb7bu4)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

mojganii (changes requested):
Please, have a look at my comment above.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709#-:-NWBzHGy3aynJCqNrlEC:blb81ae)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

rablador (commented):
First issue: I haven't been able to reproduce this myself. Are you testing on simulator or device?
Second issue: This is by design. If you connect to the tunnel first, then lose connection, you'll still be protected (secure "connection") unless you disconnect (which is the reason for this new popup).

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709#-:-NWC_GFn2ASzMiSYKVPF:b-byrvie)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709#-NW1xGF84u_-TwXO6MDc:-NWCZTcsC5gir182aZHr:b-896fix) (no related file):*

1. Perhaps but we definitely should add a task to handle alerts in some way.
2. Yes.
</blockquote></details>

Done.

___

mojganii (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709#-:-NWGqnty9cPKs26CfiDS:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4709)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4738, opened 2023-05-25
Only build the Rust artifacts we need on desktop

faern:
I have realized that our `build.sh` for building the desktop app builds more Rust code than needed. It builds the entire workspace. Historically this has not really mattered much at all since basically all Rust crates/code was used on desktop anyway. But that is becoming less and less true. For example the Android `translations-converter` is one binary we don't need. And recently added is the `ios/MullvadREST/shadowsocks-proxy` which we don't need on desktop. The dependencies overlap a lot, so the actual **compile** part is not shortened a lot by not building them. But each binary that we need to **link** cost significant amount of time.

By specifying just what we need to build here, we save a bunch of time. On my laptop, when building in debug mode, the `cargo build ...` part goes down from ~1m50s to ~1m10s.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4738)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -214,7 +214,17 @@ function build {
     if [[ -n $current_target ]]; then
         cargo_target_arg+=(--target="$current_target")
     fi
-    cargo build "${cargo_target_arg[@]}" "${CARGO_ARGS[@]}"
+    local cargo_crates_to_build=(
+        -p mullvad-daemon --bin mullvad-daemon
+        -p mullvad-cli --bin mullvad
+        -p mullvad-setup --bin mullvad-setup
+        -p mullvad-problem-report --bin mullvad-problem-report
+        -p talpid-openvpn-plugin --lib
+    )
+    if [[ ("$(uname -s)" == "Linux") ]]; then
+        cargo_crates_to_build+=(-p mullvad-exclude --bin mullvad-exclude)
+    fi
+    cargo build "${cargo_target_arg[@]}" "${CARGO_ARGS[@]}" "${cargo_crates_to_build[@]}"
 
     ################################################################################
     # Move binaries to correct locations in dist-assets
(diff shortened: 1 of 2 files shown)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4738)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`mullvad-daemon/Cargo.toml` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4738#-NWIFjTu6e2L-n_Tfwoi:-NWIFjTu6e2L-n_Tfwoj:b3ki9gu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e4ba63420e509c74f17de7174d11ddba6ee3cb1/mullvad-daemon/Cargo.toml#L12)):*
> ```TOML
> [features]
> # Allow the API server to use to be configured
> api-override = ["mullvad-api/api-override"]
> ```

This was required since `cargo build --feature api-override --package mullvad-daemon` would complain that no selected package specified this feature. We do the exact same thing in `mullvad-jni` already, so nothing new here really. We just make sure to properly expose and forward the feature from the binary crate that will ultimately use it.

pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4738#-:-NWIKk8b1TX1d493G3Xx:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

MarkusPettersson98 (approved):
I also see an improvement to the build times for debug mode: from ~1m9s to ~1m2s. So while it is not as significant of an improvement on my machine, it is a welcome optimization!

Well done! :lgtm:

> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4738#-:-NWIJwkQ9YZHGIIZmqDt:b-qtbfkd)* status: all files reviewed, 1 unresolved discussion

PR #4743, opened 2023-05-26
simplify rpm package

### LINKED ISSUE #2242: Add OpenSUSE package
It would be nice that if there is a official OpenSUSE version. OpenSUSE is a a well-known, very established Linux distro and has been around since 2005. Its Enterprise edition, SUSE, existed since 2000. OpenSUSE uses RPM package so it's pretty similar to Fedora. However, during the installation on my OpenSUSE Tumbleweed using the Fedora package, I encountered two issues:

1. The installer says dependency `dbus-libs` cannot be found. This warning can be ignored since OpenSUSE already installs this dependency actually by default, but just with a different naming convention.
2. Startup on boot doesn't work and requires manual fixing. One needs to edit `mullvad-daemon.service` with this command input: `sudo nano /opt/Mullvad\ VPN/resources/mullvad-daemon.service`, then change `Wants=network.target` to `Wants=network-online.target`. You can see the detailed discussion [here](https://forums.opensuse.org/showthread.php/537394-Can-t-get-service-to-start-on-boot).

Having an officially supported OpenSUSE package should be a quick job, since the OpenSUSE package is pretty similar to Fedora package with a few changes.

cornfeedhobo:
This commit simplifies the RPM package to not include dependency
declarations that are likely to be different across distributions.

Fixes #2242 

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4743)

### DIFF
--- a/gui/tasks/distribution.js
+++ b/gui/tasks/distribution.js
@@ -166,12 +166,15 @@ const config = {
     artifactName: 'MullvadVPN-${version}_${arch}.${ext}',
     category: 'Network',
     icon: distAssets('icon.icns'),
-    extraFiles: [{ from: distAssets('linux/mullvad-gui-launcher.sh'), to: '.' }],
+    extraFiles: [
+      { from: distAssets('linux/mullvad-gui-launcher.sh'), to: '.' },
+      { from: distAssets(path.join(getLinuxTargetSubdir(), 'mullvad-daemon')), to: '.' },
+      { from: distAssets(path.join(getLinuxTargetSubdir(), 'libtalpid_openvpn_plugin.so')), to: '.' },
+      { from: distAssets(path.join('binaries', '${env.TARGET_TRIPLE}', 'openvpn')), to: '.' },
+    ],
     extraResources: [
       { from: distAssets(path.join(getLinuxTargetSubdir(), 'mullvad-problem-report')), to: '.' },
       { from: distAssets(path.join(getLinuxTargetSubdir(), 'mullvad-setup')), to: '.' },
-      { from: distAssets(path.join(getLinuxTargetSubdir(), 'libtalpid_openvpn_plugin.so')), to: '.' },
-      { from: distAssets(path.join('binaries', '${env.TARGET_TRIPLE}', 'openvpn')), to: '.' },
     ],
   },
 
@@ -202,6 +205,7 @@ const config = {
 
   rpm: {
     fpm: [
+      '--no-depends',
       // Prevents RPM from packaging build-id metadata, some of which is the
       // same across all electron-builder applications, which causes package
       // conflicts
@@ -227,7 +231,6 @@ const config = {
     ],
     afterInstall: distAssets('linux/after-install.sh'),
     afterRemove: distAssets('linux/after-remove.sh'),
-    depends: ['libXScrnSaver', 'libnotify', 'dbus-libs'],
   },
 };
 


building/Dockerfile
```diff
@@ -10,7 +10,7 @@
 #     -v $CARGO_TARGET_VOLUME_NAME:/cargo-target:Z \
 #     -v $CARGO_REGISTRY_VOLUME_NAME:/root/.cargo/registry:Z \
 #     -v /path/to/repository_root:/build:Z \
-#     mullvadvpn-app-build ./build.sh
+#     mullvad-app-build ./build.sh
```
faern, line 13: Why this change? We do use `mullvadvpn-app-build` in all other places.

gui/tasks/distribution.js
```diff
@@ -227,7 +231,6 @@ const config = {
     ],
     afterInstall: distAssets('linux/after-install.sh'),
     afterRemove: distAssets('linux/after-remove.sh'),
-    depends: ['libXScrnSaver', 'libnotify', 'dbus-libs'],
```
faern, line 230: This is not an improvement for Fedora users, but rather a problem. If the dependencies are not specified a Fedora user can install the app, but it might now work properly. That will generate a lot of frustrated emails to our support.

cornfeedhobo: @faern you are doing this at the cost of other distros, and as stressed in the related issue, this is already the accepted norm for debian - why screw one flavor and not all of them? Why play favorites?

building/Dockerfile
cornfeedhobo: good catch. this is a typo.

gui/tasks/distribution.js
raksooo: For reference these dependencies were added here https://github.com/mullvad/mullvadvpn-app/pull/289 and here https://github.com/mullvad/mullvadvpn-app/pull/489. I have no idea if any of them aren't needed anymore.

faern: > The needed dependencies are popular and likely already installed for most desktop environments

That's not how dependencies work. A package should specify what it needs to run. Not what it needs to run *that is likely not installed*.

> Fedora isn't the only rpm distro around and it's users should not be given special treatment

Just because someone produce an RPM does not mean they intend to support every platform using that format. We currently have the capacity to support Fedora, and we made the decision to support Fedora. That does not mean we have the capacity to do QA for every RPM based distro.

cornfeedhobo: @faern Unless there is a plan to revamp the build system to consider distro-specific dependencies, or fix the deb's lack of dependencies, this entire concern feels arbitrary. I've been packaging this for over a decade now and fully understand how dependencies work, but the reality is that your team has tightly coupled your build process to electron and thus are locked into certain limitations - this is one of them.

faern: > as demonstrated by their absence in debian

The reason they are not in the .deb is probably because it has never been an issue. Both .rpm and .deb started out without many dependencies at all. Then we added them as we found out it caused issues. So the only reason we added it to the rpm is likely because users hit issues, but Debian/Ubuntu users did not. A very pragmatic approach.

However. This was added in 2018 and probably not revised since then. I see now that a minimal install of Fedora 38 comes with `dbus-libs` installed. So it's as you say, very unlikely that any recent Fedora install does not have this library. It was probably different in 2018.

I'm still not sure about removing `libXScrnSaver` and `libnotify`. I will need to discuss with @raksooo about those. But are those also problematic to OpenSUSE or is it just `dbus-libs`?


> I dug into this general topic a little further and found that there are other problematic dependencies that would require per-distro specification that mullvad has accepted is beyond the purview of these packages. Given that this is already an accepted reality

That's a different situation. Tray icons on Linux is a terrible ecosystem right now. And there are multiple ways of doing it. So we don't force any solution on the user since they might want another solution. With `dbus-libs` there is no alternative (for Fedora) so we simply installed the one required package.

cornfeedhobo: @faern Thanks for taking a deeper looking this. I appreciate it.

> are those also problematic to OpenSUSE or is it just dbus-libs?

AFAIK, it's only `dbus-libs`.

My personal take would be remove all dependencies and instead document them, that way the approach to dependency resolution is always the same - read the docs.

If I really wanted to figure this out, I'd probably try to to get the build system to create per-distro packages using a map of distro dependencies, but that also seems like a lot of work to maintain for a small team.

faern: People don't read docs. We'll get flooded by emails from users who fail to start the app. I'm OK with accepting this PR if we only remove `dbus-libs`

> If I really wanted to figure this out, I'd probably try to to get the build system to create per-distro packages using a map of distro dependencies, but that also seems like a lot of work to maintain for a small team.

If we can avoid it we don't want more artifacts. Because it blows up almost exponentially (two architectures etc) and we need to redesign the download page to have more download buttons etc. It all just escalates.

gui/tasks/distribution.js
```diff
@@ -166,12 +166,15 @@ const config = {
...
+      { from: distAssets('linux/mullvad-gui-launcher.sh'), to: '.' },
+      { from: distAssets(path.join(getLinuxTargetSubdir(), 'mullvad-daemon')), to: '.' },
+      { from: distAssets(path.join(getLinuxTargetSubdir(), 'libtalpid_openvpn_plugin.so')), to: '.' },
+      { from: distAssets(path.join('binaries', '${env.TARGET_TRIPLE}', 'openvpn')), to: '.' },
+    ],
     extraResources: [
       { from: distAssets(path.join(getLinuxTargetSubdir(), 'mullvad-problem-report')), to: '.' },
       { from: distAssets(path.join(getLinuxTargetSubdir(), 'mullvad-setup')), to: '.' },
-      { from: distAssets(path.join(getLinuxTargetSubdir(), 'libtalpid_openvpn_plugin.so')), to: '.' },
-      { from: distAssets(path.join('binaries', '${env.TARGET_TRIPLE}', 'openvpn')), to: '.' },
```
faern, line 174: Why are these moved?

gui/tasks/distribution.js
```diff
@@ -166,12 +166,15 @@ const config = {
     artifactName: 'MullvadVPN-${version}_${arch}.${ext}',
     category: 'Network',
     icon: distAssets('icon.icns'),
-    extraFiles: [{ from: distAssets('linux/mullvad-gui-launcher.sh'), to: '.' }],
+    extraFiles: [
+      { from: distAssets('linux/mullvad-gui-launcher.sh'), to: '.' },
+      { from: distAssets(path.join(getLinuxTargetSubdir(), 'mullvad-daemon')), to: '.' },
```
faern, line 171: Won't this install `mullvad-daemon` in two places? Further down we install it to `/usr/bin/` also.

gui/tasks/distribution.js
secretmango: as a Fedora user I agree. Your instructions are already a few commands, just include the command to install the needed packages and you are done.

faern:
I'm closing this for two reasons. The first one is that it's stale. There are questions and feedback that has not been processed. And secondly we do not want to make it harder to install the app on our *supported distros*. We do not support OpenSUSE.

PR #4745, opened 2023-05-30
Implement text formatter for account number and voucher code

mojganii:
This formatter formates strings based on these configuration  

-  allowed inputs in two modes that includes `numeric` or `alphanumeric`,
-  separates string with delimiter character
-  group by the string based on size

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)


pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-:-NWgVF1B27BgB-1SBmld:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 271 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWlFMT32n2QCc3Iquq2:-NWlFMT47UyDVQWsFRpB:bd7sx8j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L271)):*
> ```Swift
>     private class func isLetter(_ character: Character) -> Bool {
>         switch character {
>         case "a" ... "z", "A" ... "Z":
> ```

This doens't cover characters outside the ascii range.
We should use `CharacterSet` instead. 
I haven't tried it, but this should do the same thing
```
    private class func isLetter(_ character: Character) -> Bool {
        character.unicodeScalars.allSatisfy(CharacterSet.letters.contains)
    }
```
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 278 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWlFoYfDR3EuPS5U-ZQ:-NWlFoYg-GMUJuuowKEm:b5eznf6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L278)):*
> ```Swift
>     }
> 
>     private class func isNumber(_ character: Character) -> Bool {
> ```

Same for this 
```
    private class func isNumber(_ character: Character) -> Bool {
        character.unicodeScalars.allSatisfy(CharacterSet.decimalDigits.contains)
    }
```

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 4 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPNTests/MaskedInputFormatterTests.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpsi3O5wefMVs8zcAe:-NWpsi3O5wefMVs8zcAf:b-9h81e6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPNTests/MaskedInputFormatterTests.swift#L14)):*
> ```Swift
> 
> class MaskedInputFormatterTests: XCTestCase {
>     private static func makeConfiguration() -> MaskedInputFormatter.Configuration {
> ```

This doesn't need to be static.
Also since it's a helper it's not important to see what it does, I suggest we move this at the end of the file instead, it's what we do in the other tests.
___
*[`ios/MullvadVPNTests/MaskedInputFormatterTests.swift` line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWprm-3C3gGA3XHCRF4:-NWprm-3C3gGA3XHCRF5:b-v308rk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPNTests/MaskedInputFormatterTests.swift#L24)):*
> ```Swift
> 
>     func testInitialValue() {
>         let input = MaskedInputFormatter(
> ```

We repeat this line in almost every single test, let's use the `setup` function to do that
```
    var input: MaskedInputFormatter!
    
    override func setUpWithError() throws {
        try super.setUpWithError()
        
        input = MaskedInputFormatter(string: accountNumber, configuration: makeConfiguration())
    }

```

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 7 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPNTests/MaskedInputFormatterTests.swift` line 108 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWq34OA4jXZvWBEYt85:-NWq34OA4jXZvWBEYt86:bmkbfvj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPNTests/MaskedInputFormatterTests.swift#L108)):*
> ```Swift
>             configuration: Self.makeConfiguration()
>         )
>         guard let range = input.formattedString.range(withOffset: 5, length: 0) else {
> ```

Nit: Missing space above.
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NWpyEpn3FV12B1ia_M2:b-c2ck8g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*
> ```Swift
>             // Apply cap on number of groups of characters that can be entered.
>             if configuration.maxGroups > 0 {
>                 let numGroups = reparsedString.count / Int(configuration.groupSize)
> ```

Can .groupSize ever be 0 here? If so, perhaps best to wrap it in max(.groupSize, 1)
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 269 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpz0g86c_VesFYtT6c:-NWpz0g86c_VesFYtT6d:bjkxvn9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L269)):*
> ```Swift
>     }
> 
>     private class func isLetter(_ character: Character) -> Bool {
> ```

Is there a reason for .isLetter and .isNumber to be class functions?
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 271 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWlFMT32n2QCc3Iquq2:-NWq-PZ7D6ZQ1WNbQWyR:b-56le04) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L271)):*

This doens't cover characters outside the ascii range.
We should use `CharacterSet` instead. 
I haven't tried it, but this should do the same thing
```
    private class func isLetter(_ character: Character) -> Bool {
        character.unicodeScalars.allSatisfy(CharacterSet.letters.contains)
    }
```
</blockquote></details>

Can't we just use Swift's native isLetter?
https://developer.apple.com/documentation/swift/character/isletter
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 278 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWlFoYfDR3EuPS5U-ZQ:-NWpzZdNBtlZIdGAO2v7:b3wy4rm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L278)):*

Same for this 
```
    private class func isNumber(_ character: Character) -> Bool {
        character.unicodeScalars.allSatisfy(CharacterSet.decimalDigits.contains)
    }
```
</blockquote></details>

Can't we just use Swift's native isNumber?
https://developer.apple.com/documentation/swift/character/isnumber

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: 4 of 6 files reviewed, 6 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 269 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpz0g86c_VesFYtT6c:-NX9QfUBASAA1tPdavcD:botj4d0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L269)):*

Is there a reason for .isLetter and .isNumber to be class functions?
</blockquote></details>

+1 I don't think they need to be 
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 31 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NX9R0z89ck2DH4kdTq7:-NX9R0z89ck2DH4kdTq8:b9im5gs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2628e3834e2cd743ad497d56b30b63a8df3f7f62/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L31)):*
> ```Swift
>         var groupSeparator: GroupSeparator
> 
>         /// The character length of each group of characters.
> ```

`The length of each group of characters.`



_Code quote:_
```Swift
        /// The character length of each group of characters.
```

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: 5 of 6 files reviewed, 5 unresolved discussions (waiting on @mojganii and @pronebird)

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 5 unresolved discussions (waiting on @mojganii)

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: 3 of 6 files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadVPNTests/MaskedInputFormatterTests.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpsi3O5wefMVs8zcAe:-NWvHW-t4sQw5aEgeLkd:bejgd1m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPNTests/MaskedInputFormatterTests.swift#L14)):*

This doesn't need to be static.
Also since it's a helper it's not important to see what it does, I suggest we move this at the end of the file instead, it's what we do in the other tests.
</blockquote></details>

it can be property of the class and be named `subject under test`. but I want to have the fresh object for every test case that's why I put this in `setup` and `tearDown`
___
*[`ios/MullvadVPNTests/MaskedInputFormatterTests.swift` line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWprm-3C3gGA3XHCRF4:-NWvHqes9es46VXEvSC_:b-7tdf7d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPNTests/MaskedInputFormatterTests.swift#L24)):*

We repeat this line in almost every single test, let's use the `setup` function to do that
```
    var input: MaskedInputFormatter!
    
    override func setUpWithError() throws {
        try super.setUpWithError()
        
        input = MaskedInputFormatter(string: accountNumber, configuration: makeConfiguration())
    }

```
</blockquote></details>

agree
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NXAmbBd27ABtMLpo_yR:b-cf7k6u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*

Can .groupSize ever be 0 here? If so, perhaps best to wrap it in max(.groupSize, 1)
</blockquote></details>

It's meaningless to be set zero.I removed the default value
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 269 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpz0g86c_VesFYtT6c:-NX9diHX4SqutOXe-gm0:b-y6r3h8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L269)):*

+1 I don't think they need to be 
</blockquote></details>

I agree there is no need to be function let alone to de defined static
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 271 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWlFMT32n2QCc3Iquq2:-NWvEFS712zSp7ztsnYY:brqpjsi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L271)):*

Can't we just use Swift's native isLetter?
https://developer.apple.com/documentation/swift/character/isletter
</blockquote></details>

Why unicode?
formatter works in 2 different types.it means the scope is clear.do you think we have any text that it's in unicode encoding and we are going to format this with this configuration?

```
   private class func isLetter(_ character: Character) -> Bool {
        character.isASCII && character.isLetter
    }
    


```
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 278 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWlFoYfDR3EuPS5U-ZQ:-NWvHMzg5MooounzGaYc:bzhs4lo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L278)):*

Can't we just use Swift's native isNumber?
https://developer.apple.com/documentation/swift/character/isnumber
</blockquote></details>

the above comment goes for this
```
    private class func isNumber(_ character: Character) -> Bool {
        character.isASCII && character.isNumber

    }
```
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 31 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NX9R0z89ck2DH4kdTq7:-NX9eaWQ-N4gxGBUVil9:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2628e3834e2cd743ad497d56b30b63a8df3f7f62/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L31)):*

`The length of each group of characters.`
</blockquote></details>

Done.

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: 3 of 6 files reviewed, 5 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadVPNTests/MaskedInputFormatterTests.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpsi3O5wefMVs8zcAe:-NXCej0p33BMN2sNN5F7:b-iztwjd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPNTests/MaskedInputFormatterTests.swift#L14)):*

it can be property of the class and be named `subject under test`. but I want to have the fresh object for every test case that's why I put this in `setup` and `tearDown`
</blockquote></details>

This is almost good, but according to the recently [introduced guidelines](https://google.github.io/swift/#initializers-1)
We should not call `.init` directly. 

If we already have the type here, we should be able to directly write

`private let configuration  = MaskedInputFormatter.Configuration(`

Also, we don't change the configuration after creating it, so we can declare it as a `let` variable instead. 
___
*[`ios/MullvadVPNTests/MaskedInputFormatterTests.swift` line 13 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXCjpQJ3lnI1IwN_8fz:-NXCjpQJ3lnI1IwN_8g-:bgykcjy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fe7d945b479efb3b90dd318edff27f87c4afc009/ios/MullvadVPNTests/MaskedInputFormatterTests.swift#L13)):*
> ```Swift
> class MaskedInputFormatterTests: XCTestCase {
>     private let accountNumber = "12345678"
>     private var subjectUnderTest: MaskedInputFormatter!
> ```

nit:

To be honest, I'm not a huge fan of the name `Subject under test` for a couple of reasons: 
* It doesn't tell me what is the subject (it's non descriptive of what is being tested)
* It doesn't convey information about the context 

And the biggest one, is the following logic. Imagine we name all our test subjects "subject under test" in all our tests.
Suddently, we can't use the "Rename all" functionality of the IDE because it would find hundreds of matches in the entire project.

Personally I think `inputFormatter` would be a more descriptive name.

I'm putting this as a `nit` because I don't want to block the PR just for this, but I would prefer that we use a more descriptive name.
___
*[`ios/MullvadVPNTests/MaskedInputFormatterTests.swift` line 23 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXCiIGo8EECphHHEb_3:-NXCiIGo8EECphHHEb_4:b-jbwzor) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fe7d945b479efb3b90dd318edff27f87c4afc009/ios/MullvadVPNTests/MaskedInputFormatterTests.swift#L23)):*
> ```Swift
> 
>     override func setUp() {
>         subjectUnderTest = .init(
> ```

Same remark about `.init` here, let's replace this with 
`subjectUnderTest = MaskedInputFormatter(`
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 31 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NX9R0z89ck2DH4kdTq7:-NXCdHXs0aHrPwjG9QyT:b-f4yqaz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2628e3834e2cd743ad497d56b30b63a8df3f7f62/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L31)):*

Done.
</blockquote></details>

Actually my suggestion was slightly incorrect, I wanted to suggest `The size of each group of characters` to be in line with the name `groupSize` 

I think that `The  length of characters of  each group.`  conveys the wrong information about this.

pronebird (commented):
> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPNTests/MaskedInputFormatterTests.swift` line 13 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXCjpQJ3lnI1IwN_8fz:-NXFvf8D1iV5swSdcwZL:b-34p1rq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fe7d945b479efb3b90dd318edff27f87c4afc009/ios/MullvadVPNTests/MaskedInputFormatterTests.swift#L13)):*

nit:

To be honest, I'm not a huge fan of the name `Subject under test` for a couple of reasons: 
* It doesn't tell me what is the subject (it's non descriptive of what is being tested)
* It doesn't convey information about the context 

And the biggest one, is the following logic. Imagine we name all our test subjects "subject under test" in all our tests.
Suddently, we can't use the "Rename all" functionality of the IDE because it would find hundreds of matches in the entire project.

Personally I think `inputFormatter` would be a more descriptive name.

I'm putting this as a `nit` because I don't want to block the PR just for this, but I would prefer that we use a more descriptive name.
</blockquote></details>

I second that. `subjectUnderTest` is heavy. It could have been just `subject` but I don't mind `inputFormatter` either.

@buggmagnet  you can rename individual properties/classnames with right click > refactor > rename...

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPNTests/MaskedInputFormatterTests.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpsi3O5wefMVs8zcAe:-NXJhznC0gsMSkNtYKL5:bb7f2mf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPNTests/MaskedInputFormatterTests.swift#L14)):*

This is almost good, but according to the recently [introduced guidelines](https://google.github.io/swift/#initializers-1)
We should not call `.init` directly. 

If we already have the type here, we should be able to directly write

`private let configuration  = MaskedInputFormatter.Configuration(`

Also, we don't change the configuration after creating it, so we can declare it as a `let` variable instead. 


</blockquote></details>

done
___
*[`ios/MullvadVPNTests/MaskedInputFormatterTests.swift` line 23 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXCiIGo8EECphHHEb_3:-NXJmfC-BpXS1utyt0EF:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fe7d945b479efb3b90dd318edff27f87c4afc009/ios/MullvadVPNTests/MaskedInputFormatterTests.swift#L23)):*

Same remark about `.init` here, let's replace this with 
`subjectUnderTest = MaskedInputFormatter(`
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NXJnUKv96gDG7rmYGVE:bb7f2mf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*

It's meaningless to be set zero.I removed the default value
</blockquote></details>

done
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 31 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NX9R0z89ck2DH4kdTq7:-NXJn1Up6yNUHAzRABmY:b3dq89e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2628e3834e2cd743ad497d56b30b63a8df3f7f62/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L31)):*

Actually my suggestion was slightly incorrect, I wanted to suggest `The size of each group of characters` to be in line with the name `groupSize` 

I think that `The  length of characters of  each group.`  conveys the wrong information about this.
</blockquote></details>

`The size of each group.` this comment I think is better because bringing the characters would be redundancy when the name of the class is been changed to `InputTextFormatter`

pronebird (changes requested):
> Reviewed 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPN/Classes/InputTextFormatter.swift` line 136 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXK8bNS4LKXvs9T8jvN:-NXK8bNS4LKXvs9T8jvO:b-isfd1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b678c4d9062d53ec6a965d3ce00af61e72883087/ios/MullvadVPN/Classes/InputTextFormatter.swift#L136)):*
> ```Swift
> 
>             // Add separator between the groups of digits
>             if numDigits > 0, numDigits % Int(configuration.groupSize) == 0 {
> ```

Same here potential division by zero
___
*[`ios/MullvadVPN/Classes/InputTextFormatter.swift` line 151 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXK6YMuBgJOsUL_J8fz:-NXK6YMvEYq0TbeVwxfL:bdvu60r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b678c4d9062d53ec6a965d3ce00af61e72883087/ios/MullvadVPN/Classes/InputTextFormatter.swift#L151)):*
> ```Swift
>         }
> 
>         if case let AllowedInput.alphanumeric(isUpperCase) = configuration.allowedInput,
> ```

We don't need to prefix with `AllowedInput` as it's inferred from `configuration.allowedInput`, i.e:

```swift
if case let .alphanumeric(isUpperCase) = configuration.allowedInput {
```

You could also match against specific case without extracting `isUpperCase`:

```swift
if case . alphanumeric(true) = configuration.allowedInput {
 // uppercase text
}
```
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NXK813DBKduoU8Nkcp0:b-sm3893) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*

done
</blockquote></details>

Division by zero is possible. So ideally what Jon said, if not too much work we should simply cap it and then use it within the scope:

```swift
let groupSize = max(configuration.groupSize, 1)
```

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 7 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPN/Classes/InputTextFormatter.swift` line 263 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXK93FZBb5OSbEzL9wB:-NXK93FZBb5OSbEzL9wC:bco1ofb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b678c4d9062d53ec6a965d3ce00af61e72883087/ios/MullvadVPN/Classes/InputTextFormatter.swift#L263)):*
> ```Swift
>         switch configuration.allowedInput {
>         case .numeric:
>             return character.isASCII && character.isNumber
> ```

Nit: 

You could avoid repetition by checking for ASCII at the top.

```
guard character.isASCII else { return false }
```

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @mojganii)

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/Classes/InputTextFormatter.swift` line 136 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXK8bNS4LKXvs9T8jvN:-NXL1_etANFubafXCqW4:b-ncoups) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b678c4d9062d53ec6a965d3ce00af61e72883087/ios/MullvadVPN/Classes/InputTextFormatter.swift#L136)):*

Same here potential division by zero
</blockquote></details>

same as above description 
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NXL1D-n5qpQmeTfaEuO:b-46tcib) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*

Division by zero is possible. So ideally what Jon said, if not too much work we should simply cap it and then use it within the scope:

```swift
let groupSize = max(configuration.groupSize, 1)
```
</blockquote></details>

if the user is able to set this value to zero then the formatter will be meaningless.so I think the app should be crashed with assertion

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @rablador)

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NXLtzvF68XLp7JslI2C:b-lv8uc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*

if the user is able to set this value to zero then the formatter will be meaningless.so I think the app should be crashed with assertion 
</blockquote></details>

I don't think it's good practice to let the app crash by design that way. If `groupSize` is never meant to be zero I think we should at least have an explicit assertion (eg. `assertionFailure`) for it, preferrably as close to its assignment as possible.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/Classes/InputTextFormatter.swift` line 151 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXK6YMuBgJOsUL_J8fz:-NXPRFAa6LhRgV68qmJH:b-7rl0jq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b678c4d9062d53ec6a965d3ce00af61e72883087/ios/MullvadVPN/Classes/InputTextFormatter.swift#L151)):*

We don't need to prefix with `AllowedInput` as it's inferred from `configuration.allowedInput`, i.e:

```swift
if case let .alphanumeric(isUpperCase) = configuration.allowedInput {
```

You could also match against specific case without extracting `isUpperCase`:

```swift
if case . alphanumeric(true) = configuration.allowedInput {
 // uppercase text
}
```
</blockquote></details>

done!
___
*[`ios/MullvadVPN/Classes/InputTextFormatter.swift` line 263 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXK93FZBb5OSbEzL9wB:-NXL9HwG0gNVDe0aHon9:bt0cksq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b678c4d9062d53ec6a965d3ce00af61e72883087/ios/MullvadVPN/Classes/InputTextFormatter.swift#L263)):*

Nit: 

You could avoid repetition by checking for ASCII at the top.

```
guard character.isASCII else { return false }
```
</blockquote></details>

the old way appeal to me
___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NXPQNQd4dNrXjNASD_b:b-j2tly6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*

I don't think it's good practice to let the app crash by design that way. If `groupSize` is never meant to be zero I think we should at least have an explicit assertion (eg. `assertionFailure`) for it, preferrably as close to its assignment as possible.
</blockquote></details>

I'm fan of assertion because it's meaningless to be set zero but for this time I get the max size between group size and one.maybe it's better. to define a unified rule for invalid values

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NXPd-1m1sXXFYBUb3Bj:bp0g4sv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*

I'm fan of assertion because it's meaningless to be set zero but for this time I get the max size between group size and one.maybe it's better. to define a unified rule for invalid values 
</blockquote></details>

I see your point about 0 being meaningless, but I don't want the app to unintentionally crash. `assert(groupSize > 0, Message)` is good enough to me as an alternative.

pronebird (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NXPZsVhCV3nM751IXM0:bhbnca9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*

I see your point about 0 being meaningless, but I don't want the app to unintentionally crash. `assert(groupSize > 0, Message)` is good enough to me as an alternative.
</blockquote></details>

First of all it's unacceptable for the application to crash in most cases because:

1. It's hard to trace and fix bugs based on crash logs.
2. Crash may affect data integrity or cause data loss. Any interrupted input or output to disk or elsewhere could leave the app in inconsistent state.
3. It's much easier to understand `assert()` than a crash in the middle of something.
4. It's better to have a broken UI than a broken app because at least people can file a bug report to us!

So why not simply add assertion at the top of this function and move on? 

```swift
assert(configuration.groupSize > 0, "Group size must be greater than zero")
```` 

Anyone who'll attempt to configure formatter with zero size group will be left with a meaningful message instead of a mysterious stack trace.

I can imagine that there are ways to validate configuration and all that but I think this will lead to over-engineering of what appears to be a formatter used in limited context and configured solely by developer and not user:

1. Account input: 4 digit groups separated by blank space ` `
2. Voucher input: 4 digit/letter groups separated by `-`

All that said I think assertion early on in this function should be sufficient enough.

**Review by pronebird — COMMENTED** [review]
> Dismissed @buggmagnet from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

**Review by mojganii — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NXQuW8z80QJJqQgm0El:buuby0a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*

First of all it's unacceptable for the application to crash in most cases because:

1. It's hard to trace and fix bugs based on crash logs.
2. Crash may affect data integrity or cause data loss. Any interrupted input or output to disk or elsewhere could leave the app in inconsistent state.
3. It's much easier to understand `assert()` than a crash in the middle of something.
4. It's better to have a broken UI than a broken app because at least people can file a bug report to us!

So why not simply add assertion at the top of this function and move on? 

```swift
assert(configuration.groupSize > 0, "Group size must be greater than zero")
```` 

Anyone who'll attempt to configure formatter with zero size group will be left with a meaningful message instead of a mysterious stack trace.

I can imagine that there are ways to validate configuration and all that but I think this will lead to over-engineering of what appears to be a formatter used in limited context and configured solely by developer and not user:

1. Account input: 4 digit groups separated by blank space ` `
2. Voucher input: 4 digit/letter groups separated by `-`

All that said I think assertion early on in this function should be sufficient enough.
</blockquote></details>

@rablador @pronebird I think we all of us agree to use `assert` function for preventing from invalid value.so for the next time I will add assretion for input values.by the way, my  current changes check if the group size is bigger than zero then it goes for formatting.therefore I think the PR is mergeable. 
I didn't mean the crash is acceptable I meant we have to crash by `assert` function to give user the information the value should be bigger than zero.

**Review by pronebird — COMMENTED** [review]
> Dismissed @rablador from a discussion.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NXR6Z-_9tXoYUSblzFI:b-ipwoce) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*

@rablador @pronebird I think we all of us agree to use `assert` function for preventing from invalid value.so for the next time I will add assretion for input values.by the way, my  current changes check if the group size is bigger than zero then it goes for formatting.therefore I think the PR is mergeable. 
I didn't mean the crash is acceptable I meant we have to crash by `assert` function to give user the information the value should be bigger than zero.
</blockquote></details>

`assert()` is acceptable in case like this I believe. 

Either way, I just pulled your PR and I must have overlooked the check for `configuration.groupSize > 0` and the other one that led to division by zero in mod operator.  

I think the way you handle it is the most optimal, so sorry for rumbling about nothing.

In short the way I understand it:

1. `maxGroups == 0`  means as many groups as you want. Which is great.
2. `groupSize == 0` means that no separators, i.e it's all one big group.

All checks out and I am merging this. Good work! :+1:

**Review by pronebird — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: all files reviewed, 1 unresolved discussion (waiting on @mojganii)

___
*[`ios/MullvadVPN/Classes/InputTextFormatter.swift` line 125 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXRCXNAA4K9IqflrI4q:-NXRCXNAA4K9IqflrI4r:b-91smnt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/91b480d0c5b3562c715bcfa831ea1459415a580b/ios/MullvadVPN/Classes/InputTextFormatter.swift#L125)):*
> ```Swift
>             // Apply cap on number of groups of characters that can be entered.
>             if configuration.maxGroups > 0 {
>                 let numGroups = reparsedString.count / Int(configuration.groupSize)
> ```

Uh oh one last thing, since you divide by `configuration.groupSize` in here, perhaps modify the check for `groupSize` too.

```swift
configuration.maxGroups > 0, configuration.groupSize > 0
```

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`ios/MullvadVPN/Classes/InputTextFormatter.swift` line 125 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NXRCXNAA4K9IqflrI4q:-NXRD6bR5qsvakHszDdW:beqcvle) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/91b480d0c5b3562c715bcfa831ea1459415a580b/ios/MullvadVPN/Classes/InputTextFormatter.swift#L125)):*

Uh oh one last thing, since you divide by `configuration.groupSize` in here, perhaps modify the check for `groupSize` too.

```swift
configuration.maxGroups > 0, configuration.groupSize > 0
```
</blockquote></details>

You know I think I'll do that quickly right now and merge it to avoid stalling it for much longer.

pronebird (approved):

rablador (approved):
> Reviewed 1 of 1 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`ios/MullvadVPN/Classes/MaskedInputFormatter.swift` line 127 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4745#-NWpyEpn3FV12B1ia_M1:-NXRGEr8BvYpDhsOg7SU:b12fhce) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9ce9a9e419bd7caca109139d2639c7c059f2c78/ios/MullvadVPN/Classes/MaskedInputFormatter.swift#L127)):*

`assert()` is acceptable in case like this I believe. 

Either way, I just pulled your PR and I must have overlooked the check for `configuration.groupSize > 0` and the other one that led to division by zero in mod operator.  

I think the way you handle it is the most optimal, so sorry for rumbling about nothing.

In short the way I understand it:

1. `maxGroups == 0`  means as many groups as you want. Which is great.
2. `groupSize == 0` means that no separators, i.e it's all one big group.

All checks out and I am merging this. Good work! :+1: 
</blockquote></details>

Oh, sorry, I missed that check too! 🙈 Got hung up on the division... 🫠

PR #4746, opened 2023-05-30
Verify that important files have not been accidentally removed

MarkusPettersson98:
Verify that we do not remove any file we deem important. If any such file is removed, they should also be removed from the list this CI job uses, and as such we raise a flag.

Update the CI job to also check that all files in the list of important files still exists in the working directory/at the HEAD where it's being executed from.

Hopefully this PR will solve DES-216.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746)

### DIFF
--- a/ci/verify-locked-down-signatures.sh
+++ b/ci/verify-locked-down-signatures.sh
@@ -2,8 +2,8 @@
 set -eu
 shopt -s nullglob
 
-SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-readonly SCRIPT_DIR
+readonly SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+readonly REPO_DIR="$SCRIPT_DIR/.."
 
 # In the CI environment we would like to import trusted public keys from a file,
 # but not in our build environment
@@ -51,26 +51,38 @@ fi
 # correctly. This is done by replacing all new-lines with a `;`
 readonly SEPARATOR=';'
 locked_down_paths=$(\
-    < "$SCRIPT_DIR/../.github/workflows/verify-locked-down-signatures.yml" tr '\n' $SEPARATOR \
+    < "$REPO_DIR/.github/workflows/verify-locked-down-signatures.yml" tr '\n' $SEPARATOR \
     | sed "s/.*paths:$SEPARATOR\(\(\s*-\s[a-zA-Z\/\.-]*$SEPARATOR\)*\).*/\1/" \
     | tr $SEPARATOR '\n' \
     | awk '{print $2}')
 
+
+
 unsigned_commits_exist=0
+important_file_was_removed=0
 for locked_path in $locked_down_paths; do
+    echo "Checking $locked_path"
+
     locked_path_commit_hashes=$(git rev-list --oneline "$whitelisted_commit"..HEAD \
-        "$SCRIPT_DIR/../$locked_path" | awk '{print $1}')
+        "$REPO_DIR/$locked_path" | awk '{print $1}')
 
     for commit in $locked_path_commit_hashes; do
+        echo -e "\tin $commit.."
         if ! git verify-commit "$commit" 2> /dev/null; then
             echo "Commit $commit which changed $locked_path is not signed."
             unsigned_commits_exist=1
         fi
     done
+
+    # Check if important file has been removed.
+    if [[ ! -e "$REPO_DIR/$locked_path" ]]; then
+        echo "$locked_path was removed. If this was intentional, remove it from `verify-locked-down-signatures.yml`."
+        important_file_was_removed=1
+    fi
 done
 
-if [[ $unsigned_commits_exist == 0 ]]; then
-    echo "SUCCESS: Could not find any unsigned commits which modified a locked down path"
+if [[ "$unsigned_commits_exist" != 0 || "$important_file_was_removed" != 0 ]]; then
+    exit 1
 fi
 
-exit $unsigned_commits_exist
+echo "SUCCESS: Could not find any offenses to locked down paths"


raksooo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746)* status: all files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`ci/verify-locked-down-signatures.sh` line 81 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746#-NWqNQWt2_74ckU0PX11:-NWqNQWt2_74ckU0PX12:b-p2sp7e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94853cfde27b7042bd211a41a8a464c37e9fd1d6/ci/verify-locked-down-signatures.sh#L81)):*
> ```Shell
>     echo "These important file no longer exist:"
>     for removed_important_file in ${removed_important_files[*]}; do
>         echo -e "\t- $removed_important_file"
> ```

Have you thought about doing this the same way as the non-signed changes check, to print this message where the check is performed? I'm not a huge fan of arrays in bash and I think it would make the code a bit simpler.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[`ci/verify-locked-down-signatures.sh` line 81 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746#-NWqNQWt2_74ckU0PX11:-NX9SD1l-nb_xMPBWfQd:bt5x6n8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94853cfde27b7042bd211a41a8a464c37e9fd1d6/ci/verify-locked-down-signatures.sh#L81)):*

Have you thought about doing this the same way as the non-signed changes check, to print this message where the check is performed? I'm not a huge fan of arrays in bash and I think it would make the code a bit simpler.
</blockquote></details>

What's wrong with arrays? We liberally use them in many other scripts. We don't have any precedence in this repository for avoiding them as far as I know.

But with that being said, I agree that this script could indeed be a bit simpler by just printing directly when finding an offense, storing just the fact that we found an error and exit with that code later. Also mostly for keeping the code more consistent with itself.

Staying with a single `exit` command at the end would be preferable IMO. Also, currently the script may be printing `SUCCESS: ...` when in reality it exits with an error. That might be confusing.

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[`ci/verify-locked-down-signatures.sh` line 81 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746#-NWqNQWt2_74ckU0PX11:-NX9c-q62yuEq8JjpyBX:buqgtjv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94853cfde27b7042bd211a41a8a464c37e9fd1d6/ci/verify-locked-down-signatures.sh#L81)):*

What's wrong with arrays? We liberally use them in many other scripts. We don't have any precedence in this repository for avoiding them as far as I know.

But with that being said, I agree that this script could indeed be a bit simpler by just printing directly when finding an offense, storing just the fact that we found an error and exit with that code later. Also mostly for keeping the code more consistent with itself.

Staying with a single `exit` command at the end would be preferable IMO. Also, currently the script may be printing `SUCCESS: ...` when in reality it exits with an error. That might be confusing.
</blockquote></details>

Thanks for the nuanced discussion!

I removed the array approach in favor of printing offenses as they are found, as this is a more direct approach.

One thing in the latest revision: combining the results of `$unsigned_commits_exist` and `$important_file_was_removed`inline was a bit cumbersome and code-golfy imho, so I instead chose to branch on the boolean expression "Nothing went wrong verifying locked down paths". What are your opinion on this? The branchless exit command at the end of the script no longer exists, but on the other hand there exists other branches in the script which also exits with an error code in case something was off. :blush:

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746)* status: all files reviewed, 2 unresolved discussions (waiting on @MarkusPettersson98 and @raksooo)

___
*[`ci/verify-locked-down-signatures.sh` line 81 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746#-NWqNQWt2_74ckU0PX11:-NX9fZvM-0lP_DqACWp0:b-v5cr0o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94853cfde27b7042bd211a41a8a464c37e9fd1d6/ci/verify-locked-down-signatures.sh#L81)):*

Thanks for the nuanced discussion!

I removed the array approach in favor of printing offenses as they are found, as this is a more direct approach.

One thing in the latest revision: combining the results of `$unsigned_commits_exist` and `$important_file_was_removed`inline was a bit cumbersome and code-golfy imho, so I instead chose to branch on the boolean expression "Nothing went wrong verifying locked down paths". What are your opinion on this? The branchless exit command at the end of the script no longer exists, but on the other hand there exists other branches in the script which also exits with an error code in case something was off. :blush: 
</blockquote></details>

This code structure works! I think this is better than before :ok_hand: 

My only nitpick now is that I think we can join the two `SUCCESS ...`-messages into one. There is no reason to print them separately. The original reason the success-message was added in the first place was to easier verify that the script had been running at all. Since if a successful script execution yielded *no output* it was easy to miss if something caused it to not run at all. So sort of a canary. If we just have a single message telling the user that no offenses were found, that serves the same purpose.
___
*[`ci/verify-locked-down-signatures.sh` line 73 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746#-NX9emG9BYf2hJhXW13Q:-NX9emG9BYf2hJhXW13R:b5zp2cn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dbac252bb2a11df92d49b554eff9c95a968dafaa/ci/verify-locked-down-signatures.sh#L73)):*
> ```Shell
> 
>     # Check if important file has been removed.
>     if [[ ! -e "$SCRIPT_DIR/../$locked_path" ]]; then
> ```

I would maybe opt for computing `REPO_DIR="$SCRIPT_DIR/.."` at the top of the script. Since when it's baked in further down like this, finding this if the script is ever moved is going to be way harder. I know we already had it like this before you got here. But having it also duplicated adds to the argument of moving it out to a variable.

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @faern and @raksooo)

___
*[`ci/verify-locked-down-signatures.sh` line 81 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746#-NWqNQWt2_74ckU0PX11:-NX9wzTQ-UrWSrKKlJhO:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94853cfde27b7042bd211a41a8a464c37e9fd1d6/ci/verify-locked-down-signatures.sh#L81)):*

This code structure works! I think this is better than before :ok_hand: 

My only nitpick now is that I think we can join the two `SUCCESS ...`-messages into one. There is no reason to print them separately. The original reason the success-message was added in the first place was to easier verify that the script had been running at all. Since if a successful script execution yielded *no output* it was easy to miss if something caused it to not run at all. So sort of a canary. If we just have a single message telling the user that no offenses were found, that serves the same purpose.
</blockquote></details>

Done.
___
*[`ci/verify-locked-down-signatures.sh` line 73 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746#-NX9emG9BYf2hJhXW13Q:-NX9x0HaDdXsCdz9jaMu:b-e10khe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dbac252bb2a11df92d49b554eff9c95a968dafaa/ci/verify-locked-down-signatures.sh#L73)):*

I would maybe opt for computing `REPO_DIR="$SCRIPT_DIR/.."` at the top of the script. Since when it's baked in further down like this, finding this if the script is ever moved is going to be way harder. I know we already had it like this before you got here. But having it also duplicated adds to the argument of moving it out to a variable.
</blockquote></details>

Nice suggestion! Refactored this in a separate commit :blush:

faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746#-:-NX9xmMZDcfXd3DcNYzb:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

___
*[`ci/verify-locked-down-signatures.sh` line 81 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746#-NWqNQWt2_74ckU0PX11:-NX9xk770Zr6dm2S_4wJ:b-83sjoh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/94853cfde27b7042bd211a41a8a464c37e9fd1d6/ci/verify-locked-down-signatures.sh#L81)):*

Done.
</blockquote></details>

Nice!

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4746)* status: all files reviewed, 1 unresolved discussion (waiting on @raksooo)

PR #4763, opened 2023-06-05
Cache the last used shadowsocks configuration and reuse it subsequently

buggmagnet:
This PR introduce changes needed for the changes to come in `IOS-165`
Namely, enabling the caching of a shadowsocks configuration so that we don't always take a random one.

**Important**
The code in this PR is valid, however the behaviour is *not guaranteed to always work*.
This relates to how the shadowsocks configuration is cached in the `TransportProvider` class.

It will cache the first shadowsocks configuration object found at random, and will not try to change it afterwards.
The reason for this is because this PR would have been twice the size otherwise.

However, I took the opportunity to do some refactoring.
There is now a `Caching` protocol that is reused in `AddressCache`, `RelayCache` and the new `ShadowsocksConfigurationCache`.
The situation is not ideal, but considering that the `RelayCacheTracker` already updates a relay list on disk, I didn't want to put the shadowsock configuration in the `CachedRelays`.
Fortunately, the `Caching` protocol makes this very easy to implement.


pronebird (changes requested):
> Reviewed 7 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763)* status: 7 of 16 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadTypes/Cache.swift` line 2 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXJnIqw2vvy7KGhUyZj:-NXJnIqw2vvy7KGhUyZk:bjlqxb5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f34fc9fb65fca817b01ef45a86f64409db8a4d75/ios/MullvadTypes/Cache.swift#L2)):*
> ```Swift
> //
> //  Cache.swift
> ```

FYI The file is called `Cache.swift` although it defines a `Caching` protocol. Seems a bit of out of sync although not critical.
___
*[`ios/MullvadTypes/Cache.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXJnQrv9POIGOUNLspf:-NXJnQrv9POIGOUNLspg:b-ib2o97) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f34fc9fb65fca817b01ef45a86f64409db8a4d75/ios/MullvadTypes/Cache.swift#L14)):*
> ```Swift
> ///
> /// Uses `JSONDecoder` for reading and `JSONEncoder` for writing. `CacheType` must conform to `Codable`
> public protocol Caching<CacheType> where CacheType: Codable {
> ```

Protocols are tempting but I'd use a small struct to implement that. The benefit is that it doesn't require classes conforming to this protocol to expose I/O methods to the outside and therefore leaking internal implementation detail. 

For instance `ShadowsocksConfigurationCache` exposes a `configuration` property which performs serialization and I/O using methods defined on `Caching` protocol. A new developer joins the team and he needs to work with that cache, he then proceeds to call `try shadowsocksCache.readFromDisk()` directly instead of using the `configuration` property. 

This is why I think using composition is better as it prevents leaking implementation detail, i.e:

```swift
struct FileCacheCoordinator<CacheType: Codable> {
  init(fileURL: URL)
  func write(_ object: CacheType) throws
  func read() throws -> CacheType 
}

class ShadowsocksCache {
   private let cache: FileCacheCoordinator<SSCacheType>
   init(location: URL) {
      cache = FileCacheCoordinator(...)
   }
   var configuration: SSCacheType { get { ... } set { ... } }
}
```
___
*[`ios/MullvadVPN/AppDelegate.swift` line 57 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXJuy9T26co80JttzI_:-NXJuy9T26co80JttzIa:b-d1blu7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f34fc9fb65fca817b01ef45a86f64409db8a4d75/ios/MullvadVPN/AppDelegate.swift#L57)):*
> ```Swift
> 
>         let containerURL = FileManager.default
>             .containerURL(forSecurityApplicationGroupIdentifier: ApplicationConfiguration.securityGroupIdentifier)!
> ```

FYI I like that we only fetch container path once. Marvelous.
___
*[`ios/MullvadVPN/AppDelegate.swift` line 62 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXJw7dbAE6jAMopdZrg:-NXJw7dbAE6jAMopdZrh:b-sujn0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f34fc9fb65fca817b01ef45a86f64409db8a4d75/ios/MullvadVPN/AppDelegate.swift#L62)):*
> ```Swift
>             canWriteToCache: true, cacheFolder: containerURL
>         )
>         addressCache.initCache()
> ```

Nit: maybe we should rename it to `load()`? `Cache` repeats twice in the name and `init` is somewhat ambiguous as it coincides with initializers in Swift.

pronebird (commented):
> Reviewed 6 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763)* status: 13 of 16 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

pronebird (commented):
Refactored the original PR:

1. Replaced `Caching` protocol in favor of `FileCache<Content: Codable>` type as I prefer composition in this case, vs. protocols which when implemented for public types also expose methods defined in protocol which in this case are only meant for internal consumption.
2. Introduced `MockFileCache` when testing `AddressCache` and `RelayCache` which enables us to emulate file cache but avoids dealing with file coordination API for Codable APIs and instead focus on testing behavior.
3. Moved `ShadowsocksConfigurationCache` and `ShadowsocksConfiguration` into `MullvadTransport`. Ideally I think all of that should be within REST framework but I am not going to refactor any of this right now.
4. Reworked `ShadowsocksConfiguration` to provide `read()` and `write()` methods as opposed to property accessor as it seems cleaner in intent and also errors are being propagated meaning we can log them or react to them if we want to.
5. Removed all tests that relied on implementation details such as that a certain configuration of Codable is used to serialize data or heavy reliance on file coordinator APIs. Instead `FileCache` tests test persistence layer while all other types of Cache only test their respective behavior.
6. `URLSessionShadowsocksTransport` was moved into separate file.
7. `RelayCache.initCache()` was renamed to `loadFromFile()` to avoid ambiguity with common initializers.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-:-NXo5nE22gOtxp8Ski07:b-ins0vb)* status: 0 of 22 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

mojganii (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763)* status: 0 of 22 files reviewed, all discussions resolved (waiting on @pronebird)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 62 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXJw7dbAE6jAMopdZrg:-NXLUyn67_cKbc-mvOYV:b-1l8vom) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f34fc9fb65fca817b01ef45a86f64409db8a4d75/ios/MullvadVPN/AppDelegate.swift#L62)):*

Nit: maybe we should rename it to `load()`? `Cache` repeats twice in the name and `init` is somewhat ambiguous as it coincides with initializers in Swift.
</blockquote></details>

`load` or `read`

rablador (changes requested):
> Reviewed 19 of 21 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/AddressCache.swift` line 28 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXoVCMm8bHjoLr20l-Y:-NXoVCMn6iz-am5_tEVG:bb02i2j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/585656d83af399b5df521ef4490213f61d9d46bd/ios/MullvadREST/AddressCache.swift#L28)):*
> ```Swift
> 
>         /// Whether address cache can be written to.
>         private let canWriteToCache: Bool
> ```

When do we want read only?
___
*[`ios/MullvadTransport/ShadowsocksConfigurationCache.swift` line 3 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXoWHAe1FStq0561c4i:-NXoWHAf8_8ePPm5Vi7F:buscj8d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/585656d83af399b5df521ef4490213f61d9d46bd/ios/MullvadTransport/ShadowsocksConfigurationCache.swift#L3)):*
> ```Swift
> //
> //  ShadowsocksConfigurationCache.swift
> //  MullvadTypes
> ```

MullvadTransport
___
*[`ios/MullvadTransport/TransportProvider.swift` line 63 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXoXI8J9tK3O6hyzYVc:-NXoXI8J9tK3O6hyzYVd:b-yv8y2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/585656d83af399b5df521ef4490213f61d9d46bd/ios/MullvadTransport/TransportProvider.swift#L63)):*
> ```Swift
>             return try shadowsocksCache.read()
>         } catch {
>             logger.error(error: error, message: "Failed to read shadowsocks cache.")
> ```

This is not strictly an error I think. In this context, trying to read from cache only means that we grab what we have or create a new one. We should probably change the log level to info or something.

Although I think the function is fine as it is, one could also make the cache read seem less critical by pure semantics. No place for logs though, but we also don't need to catch -> rethrow. 🤷‍♂️

return try? shadowsocksCache.read ?? try makeNewShadowsocksConfiguration()
___
*[`ios/MullvadVPNTests/MockFileCache.swift` line 13 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXomxi61F0cpmiElpc2:-NXomxi61F0cpmiElpc3:brn59wq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/585656d83af399b5df521ef4490213f61d9d46bd/ios/MullvadVPNTests/MockFileCache.swift#L13)):*
> ```Swift
> 
> /// File cache implementation that simulates file state and uses internal lock to synchronize access to it.
> final class MockFileCache<Content: Codable & Equatable>: FileCacheProtocol {
> ```

Very nice to have a mock cache for tests 👍
___
*[`ios/RelayCache/RelayCache.swift` line 47 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXorY18EVXT_DAqSFLy:-NXorY19Enwpc7CJjHuD:bwwhemi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/585656d83af399b5df521ef4490213f61d9d46bd/ios/RelayCache/RelayCache.swift#L47)):*
> ```Swift
>     /// Read pre-bundled relays file from disk.
>     private func readPrebundledRelays() throws -> CachedRelays {
>         guard let prebundledRelaysFileURL = Bundle(for: Self.self).url(forResource: "relays", withExtension: "json")
> ```

Nit: "relay" is used in multiple places. To avoid misspelling or unintended name changes, perhaps it'd be good to create a constant or enum for it.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadREST/AddressCache.swift` line 28 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXoVCMm8bHjoLr20l-Y:-NXp-iHAC9DFW3Ys1Ojw:bhjyysv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/585656d83af399b5df521ef4490213f61d9d46bd/ios/MullvadREST/AddressCache.swift#L28)):*

When do we want read only?
</blockquote></details>

It's in read-only mode in the packet tunnel and in r/w mode in the main app which keeps address cache up-to-date with periodic refresh.
___
*[`ios/MullvadTransport/ShadowsocksConfigurationCache.swift` line 3 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXoWHAe1FStq0561c4i:-NXp0j7L1FKySYWc_gxe:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/585656d83af399b5df521ef4490213f61d9d46bd/ios/MullvadTransport/ShadowsocksConfigurationCache.swift#L3)):*

MullvadTransport
</blockquote></details>

Done
___
*[`ios/MullvadTransport/TransportProvider.swift` line 63 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXoXI8J9tK3O6hyzYVc:-NXp0qVN4y9Q-J0wkMcw:bwtvq3b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/585656d83af399b5df521ef4490213f61d9d46bd/ios/MullvadTransport/TransportProvider.swift#L63)):*

This is not strictly an error I think. In this context, trying to read from cache only means that we grab what we have or create a new one. We should probably change the log level to info or something.

Although I think the function is fine as it is, one could also make the cache read seem less critical by pure semantics. No place for logs though, but we also don't need to catch -> rethrow. 🤷‍♂️

return try? shadowsocksCache.read ?? try makeNewShadowsocksConfiguration()
</blockquote></details>

I am afraid a combo of `try?` + `try` separated by `??` is not going to compile so instead I kept it try-catchy but removed the line that logged the error. Not a big fan of ignoring errors but I agree that it's not the end of the world if ignore the one originating from here because we gonna log the one when attempting to persist the new config into SS cache.
___
*[`ios/RelayCache/RelayCache.swift` line 47 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXorY18EVXT_DAqSFLy:-NXp2a7S24zLdo9ED4qk:bfwdu46) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/585656d83af399b5df521ef4490213f61d9d46bd/ios/RelayCache/RelayCache.swift#L47)):*

Nit: "relay" is used in multiple places. To avoid misspelling or unintended name changes, perhaps it'd be good to create a constant or enum for it.
</blockquote></details>

Those are two different files and I'd rather avoid making it easy to rename both of them at once because the latter file is a part of Xcode project and generated by script and so renaming it in Swift code is not enough. This should be better documented I suppose.

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763)* status: :shipit: complete! all files reviewed, all discussions resolved

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`ios/MullvadVPNTests/MockFileCache.swift` line 13 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4763#-NXomxi61F0cpmiElpc2:-NXp5M_j-6aemrHyYOIG:bwrdlxq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/585656d83af399b5df521ef4490213f61d9d46bd/ios/MullvadVPNTests/MockFileCache.swift#L13)):*

Very nice to have a mock cache for tests 👍
</blockquote></details>

I really hope Marco likes that too.

PR #4764, opened 2023-06-05
Enable reusing the shadowsocks connection once it is established

buggmagnet:
**Important**
This PR is meant to be merged after `IOS-185` which introduced caching a Shadowsocks configuration object to be reused.

This PR enables reuse of the Shadowsocks session to avoid switching shadowsocks relay for each API call. 

There will be conflicts with the `TransportProvider` file because both `IOS-185` and this PR touch this file in clashing ways. 

I have created the `merge-ios-185-in-ios-165` branch for this effect

commit `18dd7a850` should contain all the necessary changes. 
It also introduces a needed change from both combined works: making sure that when a shadowsocks connection is failed, the cache can be bypassed via the `skipShadowsocksCache` flag.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)


pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: all files reviewed, 12 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/RESTNetworkOperation.swift` line 198 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWOsKj7qUYWFjJHmc-:-NXWOsKj7qUYWFjJHmc0:bzgwr1p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadREST/RESTNetworkOperation.swift#L198)):*
> ```Swift
>             dispatchPrecondition(condition: .onQueue(dispatchQueue))
> 
>             if let urlError = error as? URLError, urlError.code == .cancelled {
> ```

Foundation implements pattern matching `~=(T.Code, T)`, so we could simply if-case instead of unpacking the code:

```swift
if case URLError.cancelled = error {
    // ..
}
```

Old article: https://figure.ink/2017/07/24/update-matching-nserrors.html
___
*[`ios/MullvadREST/RESTNetworkOperation.swift` line 200 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWQDJI3c64iWxk7Nt9:-NXWQDJJEsdPJUkVvKtc:b-ljs2uc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadREST/RESTNetworkOperation.swift#L200)):*
> ```Swift
>             if let urlError = error as? URLError, urlError.code == .cancelled {
>                 finish(result: .failure(OperationError.cancelled))
>                 return
> ```

This code is much simpler now so we could simply avoid explicit `return` and go with if-else flow which is more explicit:

```swift
if case URLError.cancelled = error {
  // then
} else {
  // otherwise
}
```
___
*[`ios/MullvadTransport/TransportProvider.swift` line 69 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWVE4A-110BwdiMXL8:-NXWVE4A-110BwdiMXL9:b-f3qkov) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L69)):*
> ```Swift
>             return shadowsocksTransport
>         } catch {
>             logger.error(error: error)
> ```

Nit: would be great to add a context to error, `message: "Failed to create shadowsocks transport."`
___
*[`ios/MullvadTransport/TransportProvider.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWUjpE1Lvv1IFxKZr8:-NXWUjpE1Lvv1IFxKZr9:bvw490n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L78)):*
> ```Swift
>     // MARK: RESTTransport implementation
> 
>     public var name: String { currentTransport?.name ?? "TransportProvider" }
> ```

Understandably a fallback string which we may never see in logs, but it would probably make sense to return empty string then?
___
*[`ios/MullvadTransport/TransportProvider.swift` line 91 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWVxQ68amgh1XZQDzr:-NXWVxQ68amgh1XZQDzs:b1sf9x9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L91)):*
> ```Swift
>         let currentStrategy = transportStrategy
>         guard let transport = makeTransport() else { return AnyCancellable() }
>         let transportSwitchErrors: [URLError.Code] = [
> ```

It reads as the list of errors that should cause the switch, but in reality it's the opposite -- the list of errors to ignore.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 101 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWT96QDPnfRgRAs_3t:-NXWT96QDPnfRgRAs_3u:b-zdircw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L101)):*
> ```Swift
>             -> Void = { [weak self] data, response, maybeError in
>                 guard let self else { return }
>                 if let error = maybeError as? URLError,
> ```

This part that handles the switch can be moved into a separate method to make it a bit easier to read.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWWfaT14G85mW4oAGf:-NXWWfaT14G85mW4oAGg:b-gb76v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L106)):*
> ```Swift
>                     parallelRequestsMutex.lock()
>                     // Guarantee that the transport strategy switches mode only once when parallel requests fail at
>                     // the same time.
> ```

I'll have to poke around but I don't really follow this part in regards how the switch occurs only when all parallel requests fail at the same time. 
___
*[`ios/MullvadTransport/TransportProvider.swift` line 120 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWRboCAaFih8IXyPg7:-NXWRboCAaFih8IXyPg8:b-k6n8if) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L120)):*
> ```Swift
> 
>     func makeTransport() -> RESTTransport? {
>         if currentTransport == nil {
> ```

Should be private because if called outside it would circumvent the mutex lock you use before calling this function from `sendRequest()`
___
*[`ios/MullvadTransport/TransportProvider.swift` line 125 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWTupz29CvPvJ37OTW:-NXWTupz29CvPvJ37OTX:b-hhml8g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L125)):*
> ```Swift
>                 currentTransport = shadowsocksTransport()
>             case .useURLSession:
>                 currentTransport = transport()
> ```

`transport()` basically returns `urlSessionTransport` and so I think we could remove that function and assign `urlSessionTransport` directly.
___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWN7t67nF8LoiXpgUE:-NXWN7t67nF8LoiXpgUF:b-mwpa7b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift#L14)):*
> ```Swift
> import MullvadTypes
> 
> final class SimulatorTunnelTransportProvider: RESTTransport {
> ```

I think we could feed `URLSessionTransport` directly on simulator and avoid having this class which basically acts as a wrapper.
___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWMYRsBO454qKKFDbR:-NXWMYRsBO454qKKFDbS:bvo1q5g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift#L25)):*
> ```Swift
>     }
> 
>     func sendRequest(_ request: URLRequest, completion: @escaping (Data?, URLResponse?, Error?) -> Void) -> MullvadTypes
> ```

Can ditch `MullvadTypes.` in here
___
*[`ios/MullvadVPN/TransportMonitor/TransportMonitor.swift` line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWNbTlCzF7J5CSKjh6:-NXWNbTlCzF7J5CSKjh7:bxlo814) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadVPN/TransportMonitor/TransportMonitor.swift#L34)):*
> ```Swift
>     var name: String { selectTransport(transportProvider).name }
> 
>     func sendRequest(_ request: URLRequest, completion: @escaping (Data?, URLResponse?, Error?) -> Void) -> MullvadTypes
> ```

We could discard `MullvadTypes.` prefix.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: 8 of 12 files reviewed, 11 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadREST/RESTNetworkOperation.swift` line 198 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWOsKj7qUYWFjJHmc-:-NXtBDO19t_R5dgul7kD:bp0da83) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadREST/RESTNetworkOperation.swift#L198)):*

Foundation implements pattern matching `~=(T.Code, T)`, so we could simply if-case instead of unpacking the code:

```swift
if case URLError.cancelled = error {
    // ..
}
```

Old article: https://figure.ink/2017/07/24/update-matching-nserrors.html
</blockquote></details>

I love the suggestion, thanks !
___
*[`ios/MullvadREST/RESTNetworkOperation.swift` line 200 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWQDJI3c64iWxk7Nt9:-NXtByBS3OBHw2rp8oo5:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadREST/RESTNetworkOperation.swift#L200)):*

This code is much simpler now so we could simply avoid explicit `return` and go with if-else flow which is more explicit:

```swift
if case URLError.cancelled = error {
  // then
} else {
  // otherwise
}
```
</blockquote></details>

Done.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 69 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWVE4A-110BwdiMXL8:-NXtD0LpB5IW5eCiTHFV:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L69)):*

Nit: would be great to add a context to error, `message: "Failed to create shadowsocks transport."`
</blockquote></details>

Done.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWUjpE1Lvv1IFxKZr8:-NXtD4S314malGGugqEB:b-j8wbxb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L78)):*

Understandably a fallback string which we may never see in logs, but it would probably make sense to return empty string then?
</blockquote></details>

I'm on the fence about this. Ultimately,  I'm thinking about removing the `name` since it doesn't make sense anymore, I left it as part of the transition to this new model of doing requests, but it seems to not have much use now.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 91 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWVxQ68amgh1XZQDzr:-NXtdIc39e1xMnMxTUBl:bb8d2nx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L91)):*

It reads as the list of errors that should cause the switch, but in reality it's the opposite -- the list of errors to ignore.
</blockquote></details>

I agree, I struggle a bit with the double negation but after spending over an hour trying to come up with something more fluent, the best I could do was renaming it `nonSwitchErrorCodes`

@mojganii  @rablador  if you have suggestion I'm all ears.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 101 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWT96QDPnfRgRAs_3t:-NXtjQ2SAt4pCLZXb4VI:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L101)):*

This part that handles the switch can be moved into a separate method to make it a bit easier to read.
</blockquote></details>

Done.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWWfaT14G85mW4oAGf:-NXtjQWLDmzspXSvBHEB:b-8m5927) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L106)):*

I'll have to poke around but I don't really follow this part in regards how the switch occurs only when all parallel requests fail at the same time. 
</blockquote></details>

How it works is that each request takes a copy of the current strategy when the request was made

So now we have request A and B that have a copy of the same `currentStrategy` object.
If both requests fail at the same time, only one of them will pass the `currentStrategy == transportStrategy` check, because of two things
* Only one failure at a time can run this check because of the `parallelRequestsMutex` lock
* Once the first failure has been acknowledged (via the call to `transportStrategy.didFail()`), the second check will fail because the `currentStrategy` is not equal anymore to `transportStrategy` -- one will have its `connectionAttempts` set to `0`, and the other set to `1`

Hope that makes it clearer
___
*[`ios/MullvadTransport/TransportProvider.swift` line 120 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWRboCAaFih8IXyPg7:-NXtxnp37ls8Cr0Gwu3b:b-6ddzo3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L120)):*

Should be private because if called outside it would circumvent the mutex lock you use before calling this function from `sendRequest()`
</blockquote></details>

It wouldn't be exported since this resides in `MullvadTransport` and thus cannot be called outside of this framework because of the default `internal` access level.

However, the intent is clearer by declaring this `private`, so it's still good to add it
___
*[`ios/MullvadTransport/TransportProvider.swift` line 125 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWTupz29CvPvJ37OTW:-NXtywbhBZp37-4evYs9:b-2idik2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L125)):*

`transport()` basically returns `urlSessionTransport` and so I think we could remove that function and assign `urlSessionTransport` directly.
</blockquote></details>

Maybe it makes more sense to do that now, since we don't need to expose as much `transport` vs `shadowsocksTransport` 

I will think about it
___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWMYRsBO454qKKFDbR:-NXtz9DXCeptWGYdv-r9:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift#L25)):*

Can ditch `MullvadTypes.` in here
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/TransportMonitor/TransportMonitor.swift` line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWNbTlCzF7J5CSKjh6:-NXtzCk_89mP8_rLSE9b:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadVPN/TransportMonitor/TransportMonitor.swift#L34)):*

We could discard `MullvadTypes.` prefix.
</blockquote></details>

Done.

pronebird (changes requested):
> Reviewed 3 of 4 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWUjpE1Lvv1IFxKZr8:-NY2Nqv2C82e9F845JYu:bcevxo1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L78)):*

I'm on the fence about this. Ultimately,  I'm thinking about removing the `name` since it doesn't make sense anymore, I left it as part of the transition to this new model of doing requests, but it seems to not have much use now.
</blockquote></details>

We'd still need to log which transport/way of communication is used. 

It does seem however odd that `TransportProvider` is a `RESTTransport`. It shouldn't be in my opinion. Instead it should maintain transport strategy and listen for failures coming from each returned transport and then decide whether it should switch to new transport or not. It would be a little more elegant solution, although would require transports to provide some way to tell the transport provider when requests begin and end.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 91 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWVxQ68amgh1XZQDzr:-NY2tVG1CECanyobsqNA:b-fapxru) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L91)):*

I agree, I struggle a bit with the double negation but after spending over an hour trying to come up with something more fluent, the best I could do was renaming it `nonSwitchErrorCodes`

@mojganii  @rablador  if you have suggestion I'm all ears.

</blockquote></details>

We had something similar in the past:

```swift
extension URLError {
  /// Returns true if the error is network related. 
  var isNetworkError: Bool { get }
}
````
___
*[`ios/MullvadTransport/TransportProvider.swift` line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWWfaT14G85mW4oAGf:-NY2X_r5CGTK6PhdZ01j:bhqh1p0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L106)):*

How it works is that each request takes a copy of the current strategy when the request was made

So now we have request A and B that have a copy of the same `currentStrategy` object.
If both requests fail at the same time, only one of them will pass the `currentStrategy == transportStrategy` check, because of two things
* Only one failure at a time can run this check because of the `parallelRequestsMutex` lock
* Once the first failure has been acknowledged (via the call to `transportStrategy.didFail()`), the second check will fail because the `currentStrategy` is not equal anymore to `transportStrategy` -- one will have its `connectionAttempts` set to `0`, and the other set to `1`

Hope that makes it clearer
</blockquote></details>

It does. Thanks. 

So in essence `TransportStrategy` is basically a struct that holds `connectionAttempts: UInt`, i.e:

```swift
struct TransportStrategy: Equatable {
  var connectionAttempts: UInt = 0

  mutating func didFail() {
    connectionAttempts += 1
    connectionAttempts = connectionAttempts.isMultiple(of: 3) ? 0 : connectionAttempts
  }
}
```

Given that `connectionAttempts` is simply a counter that wraps as it reaches `3`, and unless I misunderstand the logic, it seems probable that a request that's being stuck for a while, could pass the equality test after the actual strategy wraps around, and cause the transport to switch inadvertently, i.e:

- Request 1 sent with `TransportStrategy(connectionAttempts: 0)`
- Request 2 sent with `TransportStrategy(connectionAttempts: 0)` // Request that takes long to timeout
- Request 1 fails
- Request 3 sent with `TransportStrategy(connectionAttempts: 1)`
- Request 3 fails
- Request 4 sent with `TransportStrategy(connectionAttempts: 2)`
- Request 4 fails. Strategy wraps around to `TransportStrategy(connectionAttempts: 0)`
- Request 2 fails, `TransportStrategy(connectionAttempts: 0)` passes equality check // !!!

It would be better to design a system that's resilient to this kind of issues and is clear in intent. To me it's not obvious that `TransportStrategy` behaves as a group token for requests. Addition of new stored properties to it could inadvertently cause the whole algorithm to break too because `Equatable` will automatically compare all stored properties to determine equality.

What I would suggest is to simply group parallel requests by storing them in array or set. A simple large enough `counter` could be sufficient enough to reference each individual request, for example:

```swift
class TransportProvider {
    let lock = NSLock()
    var counter = UInt64.zero // Can even be UInt16/32, should still suffice imo.
    var activeRequests = Set<UInt64>()
    var transportStrategy = TransportStrategy()

    func sendRequest(...) {
        let requestID = lock.withLock {
            // Craft request identifier.
            let (partial, isOverfow) = self.counter.addingReportingOverflow(1)

            if isOverfow {
                self.counter = 0
            } else {
                self.counter = partial
            }

            // Add request into group
            self.activeRequests.insert(self.counter)

            return self.counter
        }

        sendRequest(..) {
            lock.withLock {
                // Check if request is within the group, or otherwise the first to fail.
                if self.activeRequests.contains(requestID) {
                    // Fail the whole group. Those removed requests when finished will do nothing to transport strategy
                    self.activeRequests.removeAll() 
                    self.transportStrategy.didFail()
                }
            }
        }
    }
}
```

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: all files reviewed, 5 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWWfaT14G85mW4oAGf:-NY335U5EdZ_jBAN64eH:b8kmjjc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L106)):*

It does. Thanks. 

So in essence `TransportStrategy` is basically a struct that holds `connectionAttempts: UInt`, i.e:

```swift
struct TransportStrategy: Equatable {
  var connectionAttempts: UInt = 0

  mutating func didFail() {
    connectionAttempts += 1
    connectionAttempts = connectionAttempts.isMultiple(of: 3) ? 0 : connectionAttempts
  }
}
```

Given that `connectionAttempts` is simply a counter that wraps as it reaches `3`, and unless I misunderstand the logic, it seems probable that a request that's being stuck for a while, could pass the equality test after the actual strategy wraps around, and cause the transport to switch inadvertently, i.e:

- Request 1 sent with `TransportStrategy(connectionAttempts: 0)`
- Request 2 sent with `TransportStrategy(connectionAttempts: 0)` // Request that takes long to timeout
- Request 1 fails
- Request 3 sent with `TransportStrategy(connectionAttempts: 1)`
- Request 3 fails
- Request 4 sent with `TransportStrategy(connectionAttempts: 2)`
- Request 4 fails. Strategy wraps around to `TransportStrategy(connectionAttempts: 0)`
- Request 2 fails, `TransportStrategy(connectionAttempts: 0)` passes equality check // !!!

It would be better to design a system that's resilient to this kind of issues and is clear in intent. To me it's not obvious that `TransportStrategy` behaves as a group token for requests. Addition of new stored properties to it could inadvertently cause the whole algorithm to break too because `Equatable` will automatically compare all stored properties to determine equality.

What I would suggest is to simply group parallel requests by storing them in array or set. A simple large enough `counter` could be sufficient enough to reference each individual request, for example:

```swift
class TransportProvider {
    let lock = NSLock()
    var counter = UInt64.zero // Can even be UInt16/32, should still suffice imo.
    var activeRequests = Set<UInt64>()
    var transportStrategy = TransportStrategy()

    func sendRequest(...) {
        let requestID = lock.withLock {
            // Craft request identifier.
            let (partial, isOverfow) = self.counter.addingReportingOverflow(1)

            if isOverfow {
                self.counter = 0
            } else {
                self.counter = partial
            }

            // Add request into group
            self.activeRequests.insert(self.counter)

            return self.counter
        }

        sendRequest(..) {
            lock.withLock {
                // Check if request is within the group, or otherwise the first to fail.
                if self.activeRequests.contains(requestID) {
                    // Fail the whole group. Those removed requests when finished will do nothing to transport strategy
                    self.activeRequests.removeAll() 
                    self.transportStrategy.didFail()
                }
            }
        }
    }
}
```
</blockquote></details>

Thanks for taking the time to find a different approach. The example you have is valid, however, it ignores a very crucial point of how this works.

We only `resetTransportMatching(currentStrategy)` when the network request failed in certain circumstances, namely our direct API calls are blocked, or forcibly timed out via ISP interference, or otherwise.

Realistically, if all requests time out, then the connection is just bad, and this won't make it worse or better.

In the best case scenario, we never have to run the `failureCompletionHandler` because requests work immediately.
If the direct API calls are blocked, but shadowsocks isn't, then it should work as long as the connection is stable.

What I'm trying to say is: There has to be a reason why a request fails. If both direct API calls, and shadowsocks are blocked, we cannot do anything anyway.

In your example, all requests fail, so if that happened in a real situation, neither the solution you propose, or the solution we have here would improve things.

Ultimately we don't need to track requests, we just need to guarantee that they have a valid mode of transport.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWWfaT14G85mW4oAGf:-NY3pE5eAtYjfpXrGSwF:b-g8qx4r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L106)):*
> Thanks for taking the time to find a different approach. The example you have is valid, however, it ignores a very crucial point of how this works.

YW. The error handling part is omitted intentionally because that's not my primary focus here and to keep it short. Error checks the way you have them would stay the same.

> In your example, all requests fail, so if that happened in a real situation, neither the solution you propose, or the solution we have here would improve things.

As described in step-by-step scenario above, collisions are likely with current solution, given that we have a counter wrapping after every 3rd consecutive failure attempt.

So as alternative, you could remove wrapping in `TransportStrategy` and only wrap on actual overflow to reduce the likelihood of collision to nil:

```swift
public mutating func didFail() {
    let (partial, isOverflow) = connectionAttempts.addingReportingOverflow(1)
    connectionAttempts = isOverflow ? 0 : partial
}
```

> Ultimately we don't need to track requests, we just need to guarantee that they have a valid mode of transport.

In essence you already track requests by passing a copy of current `TransportStrategy`to failure handler, which holds a counter inside (`connectionAttempts`). This counter basically represents a request group.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift` line 22 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NY45Rw10UYpWlUTqhuq:-NY45Rw10UYpWlUTqhur:b-n71rcb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a06210522a9cf4786f30f46446aceb2697f79a86/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift#L22)):*
> ```Swift
> 
>     var name: String {
>         "SimulatorTunnelTransportProvider"
> ```

transport != provider.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: 9 of 16 files reviewed, 6 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWWfaT14G85mW4oAGf:-NYHwyRM-iC2cCKAe9Oi:b59rlm4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L106)):*

> Thanks for taking the time to find a different approach. The example you have is valid, however, it ignores a very crucial point of how this works.

YW. The error handling part is omitted intentionally because that's not my primary focus here and to keep it short. Error checks the way you have them would stay the same.

> In your example, all requests fail, so if that happened in a real situation, neither the solution you propose, or the solution we have here would improve things.

As described in step-by-step scenario above, collisions are likely with current solution, given that we have a counter wrapping after every 3rd consecutive failure attempt.

So as alternative, you could remove wrapping in `TransportStrategy` and only wrap on actual overflow to reduce the likelihood of collision to nil:

```swift
public mutating func didFail() {
    let (partial, isOverflow) = connectionAttempts.addingReportingOverflow(1)
    connectionAttempts = isOverflow ? 0 : partial
}
```

> Ultimately we don't need to track requests, we just need to guarantee that they have a valid mode of transport.

In essence you already track requests by passing a copy of current `TransportStrategy`to failure handler, which holds a counter inside (`connectionAttempts`). This counter basically represents a request group.

</blockquote></details>

After giving it more thought, I really like that second suggestion. 
With all due respect, I think your first suggestion was too heavy and I was reticent, but this is a very clever use of overflowing. 

Fun fact, it turns out that `Int.max` *is* a multiple of 3 !
So we need to skip it and go directly to 1 when overflowing, I've added a test to catch that. 
___
*[`ios/MullvadTransport/TransportProvider.swift` line 125 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWTupz29CvPvJ37OTW:-NYHy2MM7z4wThSn6FIq:bq27owj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L125)):*

Maybe it makes more sense to do that now, since we don't need to expose as much `transport` vs `shadowsocksTransport` 

I will think about it
</blockquote></details>

In my head it seemed more complicated than this, but after the refactoring it was as easy as you pointed out, thanks !
___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWN7t67nF8LoiXpgUE:-NYHywzHB8gUn2RNVzUY:b-70d9ov) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift#L14)):*

I think we could feed `URLSessionTransport` directly on simulator and avoid having this class which basically acts as a wrapper.
</blockquote></details>

Yes indeed, will remove it !
___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift` line 22 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NY45Rw10UYpWlUTqhuq:-NYHzGdbCsw7oOOsdAmR:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a06210522a9cf4786f30f46446aceb2697f79a86/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelTransportProvider.swift#L22)):*

transport != provider.
</blockquote></details>

Done.

pronebird (commented):
> Reviewed 7 of 7 files at r4, 1 of 1 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: 17 of 18 files reviewed, 2 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: 17 of 18 files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 91 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWVxQ68amgh1XZQDzr:-NYMhIaNByFD9z1POzFA:b54i14a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L91)):*

We had something similar in the past:

```swift
extension URLError {
  /// Returns true if the error is network related. 
  var isNetworkError: Bool { get }
}
````
</blockquote></details>

@pronebird's opinion is more make sense.specially the codes are in the same category
___
*[`ios/MullvadTransport/TransportProvider.swift` line 32 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMdk7zERpxoNopcPYk:-NYMdk7zERpxoNopcPYl:bdb7ypw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L32)):*
> ```Swift
>         addressCache: REST.AddressCache,
>         shadowsocksCache: ShadowsocksConfigurationCache,
>         transportStrategy: TransportStrategy = .init()
> ```

based on google code style,Explicitly calling .init(...) is allowed only when the receiver of the call is a metatype variable.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 45 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMedxk9WYXqZO6w51l:-NYMedxk9WYXqZO6w51m:b-muiklt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L45)):*
> ```Swift
>     // MARK: RESTTransport implementation
> 
>     private func shadowsocksTransport() -> RESTTransport? {
> ```

Nit: it could be variable and maybe `shadowSocks`
___
*[`ios/MullvadTransport/TransportProvider.swift` line 64 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMgPTRAILdlw2zpDzF:-NYMgPTRAILdlw2zpDzG:b-kduxww) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L64)):*
> ```Swift
> 
>     /// Returns the last used shadowsocks configuration, otherwise a new randomized configuration.
>     private func shadowsocksConfiguration() throws -> ShadowsocksConfiguration {
> ```

Nit : it could be variable as well
```
  private var shadowSocksconfiguration : ShadowsocksConfiguration {
        get throws {
            do {
                return try shadowsocksCache.read()
            } catch {
                // There is no previous configuration either if this is the first time this code ran
                // Or because the previous shadowsocks configuration was invalid, therefore generate a new one.
                return try makeNewShadowsocksConfiguration()
            }
        }
    }
```
___
*[`ios/MullvadTransport/TransportProvider.swift` line 127 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMhajpARwSPcWu1jrd:-NYMhajpARwSPcWu1jre:b-72l0av) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L127)):*
> ```Swift
>                 guard let self else { return }
> 
>                 if let error = maybeError as? URLError,
> ```

isn't `requestError` more descriptive?

**Review by buggmagnet — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: 16 of 18 files reviewed, 6 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWUjpE1Lvv1IFxKZr8:-NYMaCOD3V1Pm1FUA9rO:b-iooi4y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L78)):*
> Instead it should maintain transport strategy and listen for failures coming from each returned transport and then decide whether it should switch to new transport or not.

It does all of that, the only reason it's a `RESTTransport` is to hide the abstractions of the transport selection mechanism really.

Currently we will actually hit that `name` when a transport provider isn't set just yet, because we select the transport when making a request, but we log the transport name before doing that, in `RESTNetworkOperation.didReceiveURLRequest` like so

```swift
        private func didReceiveURLRequest(_ restRequest: REST.Request, endpoint: AnyIPEndpoint) {
            dispatchPrecondition(condition: .onQueue(dispatchQueue))
            let transport = // ... Get the transport here

            logger.debug(
                """
                Send request to \(restRequest.pathTemplate.templateString) via \(endpoint) \
                using \(transport.name).
                """
            )
```

I'm okay with not having it set for a debug message, as we are guaranteed to always have a valid transport for actual requests

This is what it currently looks like in debug messages

```
2023-06-20 09:08:31.035092+0200 MullvadVPN[67193:20342986] [REST.NetworkOperation] name=get-relays.1 Send request to /app/v1/relays via 45.83.223.196:443 using TransportProvider.
2023-06-20 09:08:31.044740+0200 MullvadVPN[67193:20342985] [TunnelManager] Refresh tunnel status for new tunnel.
2023-06-20 09:08:31.144780+0200 MullvadVPN[67193:20348061] [REST.NetworkOperation] name=get-access-token.3 Send request to /auth/v1/token via 45.83.223.196:443 using url-session.
```
___
*[`ios/MullvadTransport/TransportProvider.swift` line 91 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWVxQ68amgh1XZQDzr:-NYMj0AY5z0OS0924EVL:buc57lo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L91)):*

@pronebird's opinion is more make sense.specially the codes are in the same category
</blockquote></details>

```swift
private extension URLError {
    /// Whether the transport selection should be reset.
    ///
    /// `true` if the network request
    ///  * Was not cancelled
    ///  * Was not done during a phone call
    ///  * Was made when internet connection was available
    ///  * Was made in a context with data roaming, but international roaming was turned off
    var shouldResetNetworkTransport: Bool {
        code != .cancelled &&
            code != .notConnectedToInternet &&
            code != .internationalRoamingOff &&
            code != .callIsActive
    }
}
```

I went for this solution, I think it's quite clear now, and hides the ugly "not this, not that, and not that either" logic that was confusing to read. 
___
*[`ios/MullvadTransport/TransportProvider.swift` line 127 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMhajpARwSPcWu1jrd:-NYMjF034DLfBxIB2bte:bb2zetg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L127)):*

isn't `requestError` more descriptive?
</blockquote></details>

The problem here is that we don't know if it's a request error (wrong API call, server is unavailable etc...) or a network error (network is down, time out etc...) , so in a sense, `requestError` would be technically incorrect.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: 16 of 18 files reviewed, 7 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 32 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMdk7zERpxoNopcPYk:-NYMqb9j9KfIa3VIwtDa:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L32)):*

based on google code style,Explicitly calling .init(...) is allowed only when the receiver of the call is a metatype variable.
</blockquote></details>

Done
___
*[`ios/MullvadTransport/TransportProvider.swift` line 45 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMedxk9WYXqZO6w51l:-NYMqa89CoT9VzUPLxWu:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L45)):*

Nit: it could be variable and maybe `shadowSocks`
</blockquote></details>

Done
___
*[`ios/MullvadTransport/TransportProvider.swift` line 64 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMgPTRAILdlw2zpDzF:-NYMoozT0nYqKmuRnQBZ:bejbnmq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L64)):*

Nit : it could be variable as well
```
  private var shadowSocksconfiguration : ShadowsocksConfiguration {
        get throws {
            do {
                return try shadowsocksCache.read()
            } catch {
                // There is no previous configuration either if this is the first time this code ran
                // Or because the previous shadowsocks configuration was invalid, therefore generate a new one.
                return try makeNewShadowsocksConfiguration()
            }
        }
    }
```
</blockquote></details>

There is an implicit assumption that when you get a variable from an object, there's no cost associated to it.
However, if it's a function call, that implies there is a cost, so you think more carefully about the potential call of the cost.

Using a variable also means that you could write nonsensical things such as 

`shadowsocksConfiguration = ...` 
Even if it results in a compiler error, the syntax is still valid. 
With a function, it's clear that it's not something that's meant to be assigned, and that it would be invalid syntax to do so, leading to a clearer usage.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 156 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMlsG20Er2grtipwWM:-NYMqfyj7nsbs_0i95kF:bdci18z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L156)):*

Why?
</blockquote></details>

It's not a recursive mutex, so we would deadlock here
___
*[`ios/RelaySelector/RelaySelector.swift` line 29 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMmKymA4ADkxmUFDyq:-NYMqxqz4AjqooQ10hSC:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/RelaySelector/RelaySelector.swift#L29)):*

Expilcitly checking for `== true` seems a little redundant here.
</blockquote></details>

Done

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: all files reviewed, 5 unresolved discussions (waiting on @mojganii and @pronebird)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 127 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMhajpARwSPcWu1jrd:-NYMyP7wDyuc6AjQWOmg:b-ynveow) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L127)):*

The problem here is that we don't know if it's a request error (wrong API call, server is unavailable etc...) or a network error (network is down, time out etc...) , so in a sense, `requestError` would be technically incorrect.
</blockquote></details>

That's how interpreted it as well. but do we really need to be explicit about it? Can't it just be `error` since we check its existence no matter the intent?

mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 127 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMhajpARwSPcWu1jrd:-NYN-7uK0kQ_jFppDPua:b-ksb5av) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L127)):*

That's how interpreted it as well. but do we really need to be explicit about it? Can't it just be `error` since we check its existence no matter the intent?
</blockquote></details>

I think it covers all of types error.you've made a request and you've gotten an error is related to that request.I agree with @rablador as well .I mean the `maybeError` is not suitable for a variable name.It is just my opinion.I need to say that and share this

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: all files reviewed, 5 unresolved discussions (waiting on @mojganii and @pronebird)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 91 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWVxQ68amgh1XZQDzr:-NYN3UT99RGGJEnZ0HFf:boeykz5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L91)):*

I think the "none" part of the word in a phrase like `nonSwitchErrorCodes.contains(error.code) == false` makes it hard to read. Like previous reviewers I also think that this could be remedied by supplying a helper function/variable.
</blockquote></details>

Please let me know if the alternative I provided is satisfying
___
*[`ios/MullvadTransport/TransportProvider.swift` line 127 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMhajpARwSPcWu1jrd:-NYN1goO9YBzo6x6fQJ5:b-kzswm4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L127)):*

I think it covers all of types error.you've made a request and you've gotten an error is related to that request.I agree with @rablador as well .I mean the `maybeError` is not suitable for a variable name.It is just my opinion.I need to say that and share this
</blockquote></details>

The reason the error is called `maybeError` is to highlight the fact it's an optional.
It's an old habit I have from the time we couldn't do things like `if let error` 

It feels like this is just nitpicking at this point, can we resolve this discussion ?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 64 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMgPTRAILdlw2zpDzF:-NYN62oGBnkfdBbKhTSW:b-5angmc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L64)):*

There is an implicit assumption that when you get a variable from an object, there's no cost associated to it.
However, if it's a function call, that implies there is a cost, so you think more carefully about the potential call of the cost.

Using a variable also means that you could write nonsensical things such as 

`shadowsocksConfiguration = ...` 
Even if it results in a compiler error, the syntax is still valid. 
With a function, it's clear that it's not something that's meant to be assigned, and that it would be invalid syntax to do so, leading to a clearer usage.
</blockquote></details>

What's the benefit of it being a property vs. function?

mojganii (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @pronebird)

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: 17 of 18 files reviewed, 3 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWUjpE1Lvv1IFxKZr8:-NYN8xao-hRpEZAXlLzP:bhe2bi3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L78)):*

> Instead it should maintain transport strategy and listen for failures coming from each returned transport and then decide whether it should switch to new transport or not.

It does all of that, the only reason it's a `RESTTransport` is to hide the abstractions of the transport selection mechanism really.

Currently we will actually hit that `name` when a transport provider isn't set just yet, because we select the transport when making a request, but we log the transport name before doing that, in `RESTNetworkOperation.didReceiveURLRequest` like so

```swift
        private func didReceiveURLRequest(_ restRequest: REST.Request, endpoint: AnyIPEndpoint) {
            dispatchPrecondition(condition: .onQueue(dispatchQueue))
            let transport = // ... Get the transport here

            logger.debug(
                """
                Send request to \(restRequest.pathTemplate.templateString) via \(endpoint) \
                using \(transport.name).
                """
            )
```

I'm okay with not having it set for a debug message, as we are guaranteed to always have a valid transport for actual requests

This is what it currently looks like in debug messages

```
2023-06-20 09:08:31.035092+0200 MullvadVPN[67193:20342986] [REST.NetworkOperation] name=get-relays.1 Send request to /app/v1/relays via 45.83.223.196:443 using TransportProvider.
2023-06-20 09:08:31.044740+0200 MullvadVPN[67193:20342985] [TunnelManager] Refresh tunnel status for new tunnel.
2023-06-20 09:08:31.144780+0200 MullvadVPN[67193:20348061] [REST.NetworkOperation] name=get-access-token.3 Send request to /auth/v1/token via 45.83.223.196:443 using url-session.
```
</blockquote></details>

I think this whole new issue with logging transport is a result of wrapping a group of transports behind single transport. We could have a better separation here and decouple transport from provider while maintaining the existing consistency, i.e the actual transport is logged when it's available. 

For example to capture requests and responses you could introduce a tiny wrapper around individual transports, i.e:

```swift
private struct RESTTransportWrapper<Context>: RESTTransport {
    private let transport: RESTTransport

    var onSend: (() -> Context)?
    var onReceive: ((Context, Data?, URLResponse?, Error?) -> Void)?

    init(transport: RESTTransport) {
        self.transport = transport
    }

    var name: String {
        return transport.name
    }

    func sendRequest(_ request: URLRequest, completion: @escaping (Data?, URLResponse?, Error?) -> Void) -> Cancellable {
        let context = onSend?()

        return transport.sendRequest(request) { [self] data, res, error in
            if let context {
                self.onReceive?(context, data, res, error)
            }

            completion(data, res, error)
        }
    }
}
```

and then simply add a `RESTTransportProvider ` protocol that I think we had before:

```swift
final class TransportProvider: RESTTransportProvider {
    public func getTransport() -> RESTTransport? {
        let transport = makeTransport()

        return transport.flatMap {
            var wrapper = RESTTransportWrapper(transport: transport)
            wrapper.onSend = { [self] in
                return self.transportStrategy
            }
            wrapper.onReceive = { [self] transportStrategy, data, res, err in
                // handle response
            }
        }
    }
   // etc..
}
```

Or something along those lines. Don't take my examples as verbatim solution, I am only sketching something on-fly to demonstrate what I think could suit us better.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 91 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWVxQ68amgh1XZQDzr:-NYN6s6NEv-bTZdXvBeJ:b-ra74ya) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L91)):*

Please let me know if the alternative I provided is satisfying
</blockquote></details>

Fine by me.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: 17 of 18 files reviewed, 3 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWUjpE1Lvv1IFxKZr8:-NYNDP5Z5o7giSw05zLE:b-z7nqmq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L78)):*

I think this whole new issue with logging transport is a result of wrapping a group of transports behind single transport. We could have a better separation here and decouple transport from provider while maintaining the existing consistency, i.e the actual transport is logged when it's available. 

For example to capture requests and responses you could introduce a tiny wrapper around individual transports, i.e:

```swift
private struct RESTTransportWrapper<Context>: RESTTransport {
    private let transport: RESTTransport

    var onSend: (() -> Context)?
    var onReceive: ((Context, Data?, URLResponse?, Error?) -> Void)?

    init(transport: RESTTransport) {
        self.transport = transport
    }

    var name: String {
        return transport.name
    }

    func sendRequest(_ request: URLRequest, completion: @escaping (Data?, URLResponse?, Error?) -> Void) -> Cancellable {
        let context = onSend?()

        return transport.sendRequest(request) { [self] data, res, error in
            if let context {
                self.onReceive?(context, data, res, error)
            }

            completion(data, res, error)
        }
    }
}
```

and then simply add a `RESTTransportProvider ` protocol that I think we had before:

```swift
final class TransportProvider: RESTTransportProvider {
    public func getTransport() -> RESTTransport? {
        let transport = makeTransport()

        return transport.flatMap {
            var wrapper = RESTTransportWrapper(transport: transport)
            wrapper.onSend = { [self] in
                return self.transportStrategy
            }
            wrapper.onReceive = { [self] transportStrategy, data, res, err in
                // handle response
            }
        }
    }
   // etc..
}
```

Or something along those lines. Don't take my examples as verbatim solution, I am only sketching something on-fly to demonstrate what I think could suit us better.
</blockquote></details>

Forgot to mention provider protocol:

```swift
public protocol RESTTransportProvider {
    func getTransport() -> RESTTransport?
}
```

We also use `transportProvider: () -> RESTTransport?` elsewhere which could be replaced with `RESTTransportProvider` if desired for consistency.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: 17 of 18 files reviewed, 3 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 127 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMhajpARwSPcWu1jrd:-NYNE-Nh44tyLI9SDq_g:b-pudqmq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L127)):*

The reason the error is called `maybeError` is to highlight the fact it's an optional.
It's an old habit I have from the time we couldn't do things like `if let error` 

It feels like this is just nitpicking at this point, can we resolve this discussion ?
</blockquote></details>

I agree with Jon and Mojgan. That was raised with me before and I believe we don't need to be explicit about optionals in such way that we have to prefix variables with "maybe" because everything here is optional. 

```swift
    func sendRequest(
        _ request: URLRequest,
        completion: @escaping (Data?, URLResponse?, Error?) -> Void
    ) -> Cancellable
```

It also begs the question whether passing three optionals back is a good way to communicate network response. I know URLSession is designed that way but we could make it nicer since we wrap it anyway. 

Not that long ago Marco mentioned structs with all optionals as being an anti-pattern and so perhaps functions with all optionals could be considered as falling into the same category of things to avoid. But this is something to think about and not address right away.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: 17 of 18 files reviewed, 3 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 127 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMhajpARwSPcWu1jrd:-NYNbjpDB9RkOs_bw3Hl:b-donjyg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L127)):*
I've renamed `maybeError` to `error` after discussing offline with @mojganii.
She was satisfied with that solution.

> But this is something to think about and not address right away.

Agreed

pronebird (commented):
> Reviewed 1 of 2 files at r9, 1 of 1 files at r10, 1 of 1 files at r11.
> Dismissed @mojganii from a discussion.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`ios/MullvadTransport/TransportProvider.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NXWUjpE1Lvv1IFxKZr8:-NYO8auqDYnYAejkZZT_:b-z2rntu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/decf3fdfb5b1ff06ceb9ac629a77bcd5869f153b/ios/MullvadTransport/TransportProvider.swift#L78)):*

Forgot to mention provider protocol:

```swift
public protocol RESTTransportProvider {
    func getTransport() -> RESTTransport?
}
```

We also use `transportProvider: () -> RESTTransport?` elsewhere which could be replaced with `RESTTransportProvider` if desired for consistency.
</blockquote></details>

Agreed to address that in a separate PR.
___
*[`ios/MullvadTransport/TransportProvider.swift` line 64 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4764#-NYMgPTRAILdlw2zpDzF:-NYO8muW9dRh9V_r0r9b:b-rxvnwa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d545e5689b53c91f7ab82f578bae6da7d5965c6/ios/MullvadTransport/TransportProvider.swift#L64)):*

What's the benefit of it being a property vs. function?
</blockquote></details>

Apologize for being so abrupt, but I believe we can merge this PR as is, since we have a bigger fish to fry, more important changes to do.

PR #4777, opened 2023-06-09
Fix bug when user presses logged out button

mojganii:
when user taps the logout button.the app still stays on logged-in state.The problem was that the logout button calls `logOut` instead of `doLogout`.
now when user taps the logged out button, we unset the device then we call the callbacks


rablador (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4777#-:-NXUtADO6fAl5Ma2kHP7:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4789, opened 2023-06-13
Implement redeeming voucher UI part ios 193

mojganii:
in the account flow, user is able to add more credit through voucher code.
- The voucher code can be a combination of ascii letter and number with 16 length that is grouped by `-` in 4 character
- The `Redeem` button shouldn't be enabled until the above condition is met.
- when user enters a valid voucher code and presses `Redeem` button.the service is revoked and if it's successful we show the `successful view` with the days are been added to the account.once user presses `Got it!` the view will be disappeared and the app stays on account page and each part of the app that shows remained time should be updated with the new expiry date.
- when the response of service wan't successful.we stay at the redeem voucher code  view and show the error message under `voucher text field`.when it comes to error handling, if user redeems an used voucher, the app should show `This voucher code has already been used.` and when the code is invalid should show `Voucher code is invalid.` otherwise we show server response message

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)

### DIFF
--- a/ios/MullvadREST/RESTError.swift
+++ b/ios/MullvadREST/RESTError.swift
@@ -113,6 +113,8 @@ extension REST {
         public static let deviceNotFound = ServerResponseCode(rawValue: "DEVICE_NOT_FOUND")
         public static let serviceUnavailable = ServerResponseCode(rawValue: "SERVICE_UNAVAILABLE")
         public static let tooManyRequests = ServerResponseCode(rawValue: "TOO_MANY_REQUESTS")
+        public static let invalidVoucher = ServerResponseCode(rawValue: "INVALID_VOUCHER")
+        public static let usedVoucher = ServerResponseCode(rawValue: "VOUCHER_USED")
 
         public let rawValue: String
         public init(rawValue: String) {
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -411,7 +411,13 @@
 		E1187ABD289BBB850024E748 /* OutOfTimeContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1187ABB289BBB850024E748 /* OutOfTimeContentView.swift */; };
 		E158B360285381C60002F069 /* String+AccountFormatting.swift in Sources */ = {isa = PBXBuildFile; fileRef = E158B35F285381C60002F069 /* String+AccountFormatting.swift */; };
 		E1FD0DF528AA7CE400299DB4 /* StatusActivityView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E1FD0DF428AA7CE400299DB4 /* StatusActivityView.swift */; };
+		F028A5492A336E8500C0CAA3 /* VoucherTextField.swift in Sources */ = {isa = PBXBuildFile; fileRef = F028A5482A336E8500C0CAA3 /* VoucherTextField.swift */; };
+		F028A54B2A3370FA00C0CAA3 /* RedeemVoucherContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F028A54A2A3370FA00C0CAA3 /* RedeemVoucherContentView.swift */; };
+		F028A56A2A34D4E700C0CAA3 /* RedeemVoucherViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F028A5692A34D4E700C0CAA3 /* RedeemVoucherViewController.swift */; };
+		F028A56C2A34D8E600C0CAA3 /* RedeemVoucherSucceededViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F028A56B2A34D8E600C0CAA3 /* RedeemVoucherSucceededViewController.swift */; };
+		F028A56E2A34DCC600C0CAA3 /* RedeemVoucherInteractor.swift in Sources */ = {isa = PBXBuildFile; fileRef = F028A56D2A34DCC600C0CAA3 /* RedeemVoucherInteractor.swift */; };
 		F03580252A13842C00E5DAFD /* IncreasedHitButton.swift in Sources */ = {isa = PBXBuildFile; fileRef = F03580242A13842C00E5DAFD /* IncreasedHitButton.swift */; };
+		F041CD562A38B0B7001B703B /* RedeemVoucherCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = F041CD552A38B0B7001B703B /* RedeemVoucherCoordinator.swift */; };
 		F07BF2582A26112D00042943 /* InputTextFormatterTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F07BF2572A26112D00042943 /* InputTextFormatterTests.swift */; };
 		F07BF2622A26279100042943 /* RedeemVoucherOperation.swift in Sources */ = {isa = PBXBuildFile; fileRef = F07BF2612A26279100042943 /* RedeemVoucherOperation.swift */; };
 		F07CFF2029F2720E008C0343 /* RegisteredDeviceInAppNotificationProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = F07CFF1F29F2720E008C0343 /* RegisteredDeviceInAppNotificationProvider.swift */; };
@@ -1133,7 +1139,13 @@
 		E1187ABB289BBB850024E748 /* OutOfTimeContentView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = OutOfTimeContentView.swift; sourceTree = "<group>"; };
 		E158B35F285381C60002F069 /* String+AccountFormatting.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "String+AccountFormatting.swift"; sourceTree = "<group>"; };
 		E1FD0DF428AA7CE400299DB4 /* StatusActivityView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StatusActivityView.swift; sourceTree = "<group>"; };
+		F028A5482A336E8500C0CAA3 /* VoucherTextField.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = VoucherTextField.swift; sourceTree = "<group>"; };
+		F028A54A2A3370FA00C0CAA3 /* RedeemVoucherContentView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RedeemVoucherContentView.swift; sourceTree = "<group>"; };
+		F028A5692A34D4E700C0CAA3 /* RedeemVoucherViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RedeemVoucherViewController.swift; sourceTree = "<group>"; };
+		F028A56B2A34D8E600C0CAA3 /* RedeemVoucherSucceededViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RedeemVoucherSucceededViewController.swift; sourceTree = "<group>"; };
+		F028A56D2A34DCC600C0CAA3 /* RedeemVoucherInteractor.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RedeemVoucherInteractor.swift; sourceTree = "<group>"; };
 		F03580242A13842C00E5DAFD /* IncreasedHitButton.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = IncreasedHitButton.swift; sourceTree = "<group>"; };
+		F041CD552A38B0B7001B703B /* RedeemVoucherCoordinator.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RedeemVoucherCoordinator.swift; sourceTree = "<group>"; };
 		F07BF2572A26112D00042943 /* InputTextFormatterTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = InputTextFormatterTests.swift; sourceTree = "<group>"; };
 		F07BF2612A26279100042943 /* RedeemVoucherOperation.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RedeemVoucherOperation.swift; sourceTree = "<group>"; };
 		F07CFF1F29F2720E008C0343 /* RegisteredDeviceInAppNotificationProvider.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RegisteredDeviceInAppNotificationProvider.swift; sourceTree = "<group>"; };
@@ -1457,6 +1469,7 @@
 				583FE01B29C19786006E85F9 /* OutOfTime */,
 				583FE01A29C19777006E85F9 /* Preferences */,
 				583FE01929C19760006E85F9 /* ProblemReport */,
+				F028A5472A336E1900C0CAA3 /* RedeemVoucher */,
 				583FE01C29C19793006E85F9 /* RevokedDevice */,
 				583FE01729C196F3006E85F9 /* SelectLocation */,
 				583FE01829C19709006E85F9 /* Settings */,
@@ -2159,6 +2172,19 @@
 				A9467E8A2A2E0317000DC21F /* ShadowsocksConfigurationCache.swift */,
 			);
 			path = MullvadTransport;
+			sourceTree = "<group>";
+		};
+		F028A5472A336E1900C0CAA3 /* RedeemVoucher */ = {
+			isa = PBXGroup;
+			children = (
+				F028A54A2A3370FA00C0CAA3 /* RedeemVoucherContentView.swift */,
+				F041CD552A38B0B7001B703B /* RedeemVoucherCoordinator.swift */,
+				F028A56D2A34DCC600C0CAA3 /* RedeemVoucherInteractor.swift */,
+				F028A56B2A34D8E600C0CAA3 /* RedeemVoucherSucceededViewController.swift */,
+				F028A5692A34D4E700C0CAA3 /* RedeemVoucherViewController.swift */,
+				F028A5482A336E8500C0CAA3 /* VoucherTextField.swift */,
+			);
+			path = RedeemVoucher;
 			sourceTree = "<group>";
 		};
 /* End PBXGroup section */
@@ -2923,6 +2949,7 @@
 				58BA693123EADA6A009DC256 /* SimulatorTunnelProvider.swift in Sources */,
 				587B753B2666467500DEF7E9 /* NotificationBannerView.swift in Sources */,
 				58B993B12608A34500BA7811 /* LoginContentView.swift in Sources */,
+				F041CD562A38B0B7001B703B /* RedeemVoucherCoordinator.swift in Sources */,
 				5878A27529093A310096FC88 /* StorePaymentEvent.swift in Sources */,
 				7A7AD28D29DC677800480EF1 /* FirstTimeLaunch.swift in Sources */,
 				58B26E2A2943545A00D5980C /* NotificationManagerDelegate.swift in Sources */,
@@ -2934,6 +2961,7 @@
 				58F185AA298A3E3E00075977 /* TunnelCoordinator.swift in Sources */,
 				F03580252A13842C00E5DAFD /* IncreasedHitButton.swift in Sources */,
 				58F8AC0E25D3F8CE002BE0ED /* ProblemReportReviewViewController.swift in Sources */,
+				F028A56E2A34DCC600C0CAA3 /* RedeemVoucherInteractor.swift in Sources */,
 				5878A27129091CF20096FC88 /* AccountInteractor.swift in Sources */,
 				5878F4FC29CDA2E4003D4BE2 /* ChangeLogViewController.swift in Sources */,
 				068CE5742927B7A400A068BB /* Migration.swift in Sources */,
@@ -2961,6 +2989,7 @@
 				58F2E14C276A61C000A79513 /* RotateKeyOperation.swift in Sources */,
 				5871FB96254ADE4E0051A0A4 /* ConsolidatedApplicationLog.swift in Sources */,
 				068CE57029278F5300A068BB /* MigrationFromV1ToV2.swift in Sources */,
+				F028A54B2A3370FA00C0CAA3 /* RedeemVoucherContentView.swift in Sources */,
 				58FEEB58260B662E00A621A8 /* AutomaticKeyboardResponder.swift in Sources */,
 				5846227326E22A160035F7C2 /* StorePaymentObserver.swift in Sources */,
 				58F2E146276A2C9900A79513 /* StopTunnelOperation.swift in Sources */,
@@ -3020,6 +3049,7 @@
 				5891BF1C25E3E3EB006D6FB0 /* Bundle+ProductVersion.swift in Sources */,
 				5878A26F2907E7E00096FC88 /* ProblemReportInteractor.swift in Sources */,
 				7AE47E522A17972A000418DA /* CustomAlertViewController.swift in Sources */,
+				F028A56A2A34D4E700C0CAA3 /* RedeemVoucherViewController.swift in Sources */,
 				58E11188292FA11F009FCA84 /* SettingsMigrationUIHandler.swift in Sources */,
 				58CAFA002983FF0200BE19F7 /* LoginInteractor.swift in Sources */,
 				5807E2C02432038B00F5FF30 /* String+Split.swift in Sources */,
@@ -3062,8 +3092,10 @@
 				58FB865526E8BF3100F188BC /* StorePaymentManagerError.swift in Sources */,
 				58FD5BF42428C67600112C88 /* InAppPurchaseButton.swift in Sources */,
 				587D9676288989DB00CD8F1C /* NSLayoutConstraint+Helpers.swift in Sources */,
+				F028A56C2A34D8E600C0CAA3 /* RedeemVoucherSucceededViewController.swift in Sources */,
 				58293FAE2510CA58005D0BB5 /* ProblemReportViewController.swift in Sources */,
 				5878F50429CDC547003D4BE2 /* ChangeLogContentView.swift in Sources */,
+				F028A5492A336E8500C0CAA3 /* VoucherTextField.swift in Sources */,
 				58B9EB152489139B00095626 /* RESTError+Display.swift in Sources */,
 				587B753F2668E5A700DEF7E9 /* NotificationContainerView.swift in Sources */,
 				58F2E144276A13F300A79513 /* StartTunnelOperation.swift in Sources */,
--- a/ios/MullvadVPN/AppDelegate.swift
+++ b/ios/MullvadVPN/AppDelegate.swift
@@ -81,7 +81,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             tunnelStore: tunnelStore,
             relayCacheTracker: relayCacheTracker,
             accountsProxy: accountsProxy,
-            devicesProxy: devicesProxy
+            devicesProxy: devicesProxy,
+            apiProxy: apiProxy
         )
 
         storePaymentManager = StorePaymentManager(
--- a/ios/MullvadVPN/Classes/InputTextFormatter.swift
+++ b/ios/MullvadVPN/Classes/InputTextFormatter.swift
@@ -27,7 +27,7 @@ class InputTextFormatter: NSObject, UITextFieldDelegate, UITextPasteDelegate {
         var groupSize: UInt8
 
         /// Maximum number of groups of characters allowed.
-        var maxGroups: UInt
+        var maxGroups: UInt8
     }
 
     var configuration: Configuration {
@@ -259,11 +259,12 @@ class InputTextFormatter: NSObject, UITextFieldDelegate, UITextPasteDelegate {
     }
 
     private func isAllowed(_ character: Character) -> Bool {
+        guard character.isASCII else { return false }
         switch configuration.allowedInput {
         case .numeric:
-            return character.isASCII && character.isNumber
+            return character.isNumber
         case .alphanumeric:
-            return character.isASCII && character.isLetter
+            return character.isLetter || character.isNumber
         }
     }
 }
--- a/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift
@@ -48,6 +48,36 @@ final class AccountCoordinator: Coordinator, Presentable, Presenting {
 }
 
 extension AccountCoordinator: AccountViewControllerDelegate {
+    func accountViewController(
+        _ controller: AccountViewController,
+        didRequestRoutePresentation route: AccountsNavigationRoute
+    ) {
+        switch route {
+        case .redeemVoucher:
+            let coordinator = RedeemVoucherCoordinator(
+                navigationController: CustomNavigationController(),
+                interactor: RedeemVoucherInteractor(tunnelManager: interactor.tunnelManager)
+            )
+            coordinator.didFinish = { redeemVoucherCoordinator in
+                redeemVoucherCoordinator.dismiss(animated: true)
+            }
+            coordinator.didCancel = { redeemVoucherCoordinator in
+                redeemVoucherCoordinator.dismiss(animated: true)
+            }
+
+            coordinator.start()
+            presentChild(
+                coordinator,
+                animated: true,
+                configuration: ModalPresentationConfiguration(
+                    preferredContentSize: UIMetrics.RedeemVoucher.preferredContentSize,
+                    modalPresentationStyle: .custom,
+                    transitioningDelegate: FormSheetTransitioningDelegate()
+                )
+            )
+        }
+    }
+
     func accountViewControllerDidFinish(_ controller: AccountViewController) {
         didFinish?(self, .none)
     }
--- a/ios/MullvadVPN/Coordinators/App/ApplicationCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/App/ApplicationCoordinator.swift
@@ -408,7 +408,7 @@ final class ApplicationCoordinator: Coordinator, Presenting, RootContainerViewCo
         NotificationCenter.default.addObserver(
             self,
             selector: #selector(formSheetControllerWillChangeFullscreenPresentation(_:)),
-            name: FormsheetPresentationController.willChangeFullScreenPresentation,
+            name: FormSheetPresentationController.willChangeFullScreenPresentation,
             object: secondaryNavigationContainer
         )
     }
@@ -419,7 +419,7 @@ final class ApplicationCoordinator: Coordinator, Presenting, RootContainerViewCo
     private func removeSecondaryContextPresentationStyleObserver() {
         NotificationCenter.default.removeObserver(
             self,
-            name: FormsheetPresentationController.willChangeFullScreenPresentation,
+            name: FormSheetPresentationController.willChangeFullScreenPresentation,
             object: secondaryNavigationContainer
         )
     }
--- a/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift
+++ b/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift
@@ -1,5 +1,5 @@
 //
-//  FormsheetPresentationController.swift
+//  FormSheetPresentationController.swift
 //  MullvadVPN
 //
 //  Created by pronebird on 18/02/2023.
@@ -8,19 +8,15 @@
 
 import UIKit
 
-private let dimmingViewOpacity: CGFloat = 0.5
-private let presentedViewCornerRadius: CGFloat = 8
-private let animationDuration: TimeInterval = 0.5
-
 /**
  Custom implementation of a formsheet presentation controller.
  */
-class FormsheetPresentationController: UIPresentationController {
+class FormSheetPresentationController: UIPresentationController {
     /**
      Name of notification posted when fullscreen presentation changes, including during initial presentation.
      */
     static let willChangeFullScreenPresentation = Notification
-        .Name(rawValue: "FormsheetPresentationControllerWillChangeFullScreenPresentation")
+        .Name(rawValue: "FormSheetPresentationControllerWillChangeFullScreenPresentation")
 
     /**
      User info key passed along with `willChangeFullScreenPresentation` notification that contains boolean value that
@@ -33,9 +29,11 @@ class FormsheetPresentationController: UIPresentationController {
      */
     private var lastKnownIsInFullScreen: Bool?
 
+    private var keyboardResponder: AutomaticKeyboardResponder?
+
     private let dimmingView: UIView = {
         let dimmingView = UIView()
-        dimmingView.backgroundColor = .black
+        dimmingView.backgroundColor = UIMetrics.DimmingView.backgroundColor
         return dimmingView
     }()
 
@@ -55,6 +53,11 @@ class FormsheetPresentationController: UIPresentationController {
     var isInFullScreenPresentation: Bool {
         return useFullScreenPresentationInCompactWidth &&
             traitCollection.horizontalSizeClass == .compact
+    }
+
+    override init(presentedViewController: UIViewController, presenting presentingViewController: UIViewController?) {
+        super.init(presentedViewController: presentedViewController, presenting: presentingViewController)
+        addKeyboardResponder()
     }
 
     override var frameOfPresentedViewInContainerView: CGRect {
@@ -88,11 +91,11 @@ class FormsheetPresentationController: UIPresentationController {
         dimmingView.alpha = 0
         containerView?.addSubview(dimmingView)
 
-        presentedView?.layer.cornerRadius = presentedViewCornerRadius
+        presentedView?.layer.cornerRadius = UIMetrics.DimmingView.cornerRadius
         presentedView?.clipsToBounds = true
 
         let revealDimmingView = {
-            self.dimmingView.alpha = dimmingViewOpacity
+            self.dimmingView.alpha = UIMetrics.DimmingView.opacity
         }
 
         if let transitionCoordinator = presentingViewController.transitionCoordinator {
@@ -151,21 +154,44 @@ class FormsheetPresentationController: UIPresentationController {
             userInfo: [Self.isFullScreenUserInfoKey: NSNumber(booleanLiteral: currentIsInFullScreen)]
         )
     }
+
+    private func addKeyboardResponder() {
+        guard let presentedView else { return }
+        keyboardResponder = AutomaticKeyboardResponder(
+            targetView: presentedView,
+            handler: { [weak self] view, adjustment in
+                guard let self,
+                      let containerView,
+                      !isInFullScreenPresentation else { return }
+                let frame = view.frame
+                let bottomMarginFromKeyboard = adjustment > 0 ? UIMetrics.sectionSpacing : 0
+                view.frame = CGRect(
+                    origin: CGPoint(
+                        x: frame.origin.x,
+                        y: containerView.bounds.midY - presentedViewController.preferredContentSize
+                            .height * 0.5 - adjustment - bottomMarginFromKeyboard
+                    ),
+                    size: frame.size
+                )
+                view.layoutIfNeeded()
+            }
+        )
+    }
 }
 
-class FormsheetTransitioningDelegate: NSObject, UIViewControllerTransitioningDelegate {
+class FormSheetTransitioningDelegate: NSObject, UIViewControllerTransitioningDelegate {
     func animationController(
         forPresented presented: UIViewController,
         presenting: UIViewController,
         source: UIViewController
     ) -> UIViewControllerAnimatedTransitioning? {
-        return FormsheetPresentationAnimator()
+        return FormSheetPresentationAnimator()
     }
 
     func animationController(forDismissed dismissed: UIViewController)
         -> UIViewControllerAnimatedTransitioning?
     {
-        return FormsheetPresentationAnimator()
+        return FormSheetPresentationAnimator()
     }
 
     func presentationController(
@@ -173,18 +199,18 @@ class FormsheetTransitioningDelegate: NSObject, UIViewControllerTransitioningDel
         presenting: UIViewController?,
         source: UIViewController
     ) -> UIPresentationController? {
-        return FormsheetPresentationController(
+        return FormSheetPresentationController(
             presentedViewController: presented,
             presenting: source
         )
     }
 }
 
-class FormsheetPresentationAnimator: NSObject, UIViewControllerAnimatedTransitioning {
+class FormSheetPresentationAnimator: NSObject, UIViewControllerAnimatedTransitioning {
     func transitionDuration(using transitionContext: UIViewControllerContextTransitioning?)
         -> TimeInterval
     {
-        return (transitionContext?.isAnimated ?? true) ? animationDuration : 0
+        return (transitionContext?.isAnimated ?? true) ? UIMetrics.FormSheetTransition.duration : 0
     }
 
     func animateTransition(using transitionContext: UIViewControllerContextTransitioning) {
@@ -212,8 +238,8 @@ class FormsheetPresentationAnimator: NSObject, UIViewControllerAnimatedTransitio
         if transitionContext.isAnimated {
             UIView.animate(
                 withDuration: duration,
-                delay: 0,
-                options: [.curveEaseInOut],
+                delay: UIMetrics.FormSheetTransition.delay,
+                options: UIMetrics.FormSheetTransition.animationOptions,
                 animations: {
                     destinationView.frame = transitionContext.finalFrame(for: destinationController)
                 },
@@ -238,8 +264,8 @@ class FormsheetPresentationAnimator: NSObject, UIViewControllerAnimatedTransitio
         if transitionContext.isAnimated {
             UIView.animate(
                 withDuration: duration,
-                delay: 0,
-                options: [.curveEaseInOut],
+                delay: UIMetrics.FormSheetTransition.delay,
+                options: UIMetrics.FormSheetTransition.animationOptions,
                 animations: {
                     sourceView.frame = initialFrame
                 },
--- a/ios/MullvadVPN/Presentation controllers/SecondaryContextPresentationController.swift
+++ b/ios/MullvadVPN/Presentation controllers/SecondaryContextPresentationController.swift
@@ -12,7 +12,7 @@ import UIKit
  This is a presentation controller class used for presentation of secondary navigation context
  in application coordinator.
  */
-class SecondaryContextPresentationController: FormsheetPresentationController {
+class SecondaryContextPresentationController: FormSheetPresentationController {
     override func presentationTransitionWillBegin() {
         super.presentationTransitionWillBegin()
 
@@ -49,7 +49,7 @@ class SecondaryContextPresentationController: FormsheetPresentationController {
     }
 }
 
-class SecondaryContextTransitioningDelegate: FormsheetTransitioningDelegate {
+class SecondaryContextTransitioningDelegate: FormSheetTransitioningDelegate {
     override func presentationController(
         forPresented presented: UIViewController,
         presenting: UIViewController?,
--- a/ios/MullvadVPN/TunnelManager/TunnelManager.swift
+++ b/ios/MullvadVPN/TunnelManager/TunnelManager.swift
@@ -48,6 +48,7 @@ final class TunnelManager: StorePaymentObserver {
     private let relayCacheTracker: RelayCacheTracker
     private let accountsProxy: REST.AccountsProxy
     private let devicesProxy: REST.DevicesProxy
+    private let apiProxy: REST.APIProxy
 
     private let logger = Logger(label: "TunnelManager")
     private var nslock = NSRecursiveLock()
@@ -82,13 +83,15 @@ final class TunnelManager: StorePaymentObserver {
         tunnelStore: TunnelStore,
         relayCacheTracker: RelayCacheTracker,
         accountsProxy: REST.AccountsProxy,
-        devicesProxy: REST.DevicesProxy
+        devicesProxy: REST.DevicesProxy,
+        apiProxy: REST.APIProxy
     ) {
         self.application = application
         self.tunnelStore = tunnelStore
         self.relayCacheTracker = relayCacheTracker
         self.accountsProxy = accountsProxy
         self.devicesProxy = devicesProxy
+        self.apiProxy = apiProxy
         self.operationQueue.name = "TunnelManager.operationQueue"
         self.operationQueue.underlyingQueue = internalQueue
 
@@ -398,6 +401,34 @@ final class TunnelManager: StorePaymentObserver {
         )
 
         operationQueue.addOperation(operation)
+    }
+
+    func redeemVoucher(
+        _ voucherCode: String,
+        completion: ((Result<REST.SubmitVoucherResponse, Error>) -> Void)? = nil
+    ) -> Cancellable {
+        let operation = RedeemVoucherOperation(
+            dispatchQueue: internalQueue,
+            interactor: TunnelInteractorProxy(self),
+            voucherCode: voucherCode,
+            apiProxy: apiProxy
+        )
+
+        operation.completionQueue = .main
+        operation.completionHandler = completion
+
+        operation.addObserver(
+            BackgroundObserver(
+                application: application,
+                name: "Redeem voucher",
+                cancelUponExpiration: true
+            )
+        )
+
+        operation.addCondition(MutuallyExclusive(category: OperationCategory.deviceStateUpdate.category))
+
+        operationQueue.addOperation(operation)
+        return operation
     }
 
     func updateDeviceData(_ completionHandler: ((Error?) -> Void)? = nil) {
--- a/ios/MullvadVPN/UI appearance/UIMetrics.swift
+++ b/ios/MullvadVPN/UI appearance/UIMetrics.swift
@@ -21,6 +21,23 @@ enum UIMetrics {
         /// Spacing between views in container (main view) in `CustomAlertViewController`
         static let containerSpacing: CGFloat = 16
     }
+
+    enum DimmingView {
+        static let opacity: CGFloat = 0.5
+        static let cornerRadius: CGFloat = 8
+        static let backgroundColor: UIColor = .black
+    }
+
+    enum FormSheetTransition {
+        static let duration: TimeInterval = 0.5
+        static let delay: TimeInterval = .zero
+        static let animationOptions: UIView.AnimationOptions = [.curveEaseInOut]
+    }
+
+    enum RedeemVoucher {
+        static let cornerRadius = 8.0
+        static let preferredContentSize = CGSize(width: 292, height: 263)
+    }
 }
 
 extension UIMetrics {
--- a/ios/MullvadVPN/View controllers/Account/AccountContentView.swift
+++ b/ios/MullvadVPN/View controllers/Account/AccountContentView.swift
@@ -28,6 +28,19 @@ class AccountContentView: UIView {
         return button
     }()
 
+    let redeemVoucherButton: AppButton = {
+        let button = AppButton(style: .success)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.accessibilityIdentifier = "redeemVoucherButton"
+        button.setTitle(NSLocalizedString(
+            "REDEEM_VOUCHER_BUTTON_TITLE",
+            tableName: "Account",
+            value: "Redeem voucher",
+            comment: ""
+        ), for: .normal)
+        return button
+    }()
+
     let logoutButton: AppButton = {
         let button = AppButton(style: .danger)
         button.translatesAutoresizingMaskIntoConstraints = false
@@ -74,7 +87,7 @@ class AccountContentView: UIView {
 
     lazy var buttonStackView: UIStackView = {
         let stackView =
-            UIStackView(arrangedSubviews: [purchaseButton, restorePurchasesButton, logoutButton])
+            UIStackView(arrangedSubviews: [redeemVoucherButton, purchaseButton, restorePurchasesButton, logoutButton])
         stackView.translatesAutoresizingMaskIntoConstraints = false
         stackView.axis = .vertical
         stackView.spacing = UIMetrics.interButtonSpacing
--- a/ios/MullvadVPN/View controllers/Account/AccountInteractor.swift
+++ b/ios/MullvadVPN/View controllers/Account/AccountInteractor.swift
@@ -14,7 +14,7 @@ import StoreKit
 
 final class AccountInteractor {
     private let storePaymentManager: StorePaymentManager
-    private let tunnelManager: TunnelManager
+    let tunnelManager: TunnelManager
 
     var didReceivePaymentEvent: ((StorePaymentEvent) -> Void)?
     var didReceiveDeviceState: ((DeviceState) -> Void)?
--- a/ios/MullvadVPN/View controllers/Account/AccountViewController.swift
+++ b/ios/MullvadVPN/View controllers/Account/AccountViewController.swift
@@ -13,9 +13,17 @@ import Operations
 import StoreKit
 import UIKit
 
+enum AccountsNavigationRoute {
+    case redeemVoucher
+}
+
 protocol AccountViewControllerDelegate: AnyObject {
     func accountViewControllerDidFinish(_ controller: AccountViewController)
     func accountViewControllerDidLogout(_ controller: AccountViewController)
+    func accountViewController(
+        _ controller: AccountViewController,
+        didRequestRoutePresentation route: AccountsNavigationRoute
+    )
 }
 
 class AccountViewController: UIViewController {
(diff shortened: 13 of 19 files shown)


mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: 14 of 21 files reviewed, 9 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift` line 60 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXopOTCEDJ-O-Zqtysg:-NXp-g7N36UgwZD9npyb:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift#L60)):*

Nit: formatter went haywire in here. Could you please manually fix such instances. `interactor` and `.tunnelManager` could easily fit on the same line
</blockquote></details>

Done
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 2 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXotduR-KCdjcToeUd4:-NXp082q6xTmzIv5YYiN:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L2)):*

Consider renaming the file too.
</blockquote></details>

Done
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 10 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXotBkq-i2ulQp3HLJy:-NXp-pZvATPLFD-oqGSo:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L10)):*

Consider moving it to the bottom of this file as it's a private extension and anyone opening this file would have to start by reading a private extension instead of focusing on a public type.
</blockquote></details>

Done
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 172 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXorHn2EyBazrpU4ZTR:-NXp16_ZB-tnJZSKpfpA:badm1vi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L172)):*

Pretty sure that `frameOfPresentedViewInContainerView` should reflect the adjustment. Does it work properly with device rotation or multi tasking on iPad?
</blockquote></details>

we support `portrait` orientaion on iPhone, there is no need to worry about rotation on iPhone.please,correct me if I'm wrong.
when it comes to iPad.I've tested multi tasking I didn't see any problem.
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 176 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXorv19AJUFAhg5Xx-0:-NXp1ZXCETBivMG7mtWy:b-o2061c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L176)):*

Should it re-position the view when `useFullScreenPresentationInCompactWidth = true`?
</blockquote></details>

Nope,I'll check the value `isInFullScreenPresentation` before making adjustment.
___
*[`ios/MullvadVPN/View controllers/Account/AccountInteractor.swift` line 16 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXouL8VEk4fMqiCcS--:-NXp0mJX5o_AlIzY1OOQ:bx1oil0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9be10ce8dea8cd373a1302bad5014dc19a765a0e/ios/MullvadVPN/View%20controllers/Account/AccountInteractor.swift#L16)):*

Private types should not be exposed. Interactor should handle business logic and only expose a minimal interface the view controllers need to function.
</blockquote></details>

I need to pass `Tunnel Manager` to next interactor.it's data passing object and I'm not able to maintain the object.but I've made a mistake and I shouldn't have made storePaymentManager exposable.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/AccountRedeemVoucherController.swift` line 21 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXowXgF8cHBiT4XDulB:-NXp5FWl0BSIcIIbLJla:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/View%20controllers/RedeemVoucher/AccountRedeemVoucherController.swift#L21)):*

Instead of managing children from parent view controller, we should consider making this a coordinator. Then we could also present it from parent coordinator providing custom presentation configuration etc.


```swift
class RedeemVoucherCoordinator: Coordinator, Presentable {
   let navigationController: UINavigationController?

   init(navigationController: UINavigationController = .init()) {
        self.navigationController = nvaigationController
   }
  
   func start() {
        // begin the flow
   }

   var presentedViewController: UIViewController {
        return navigationController
   }
}
```
</blockquote></details>

Done
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/AccountRedeemVoucherController.swift` line 21 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXox2oyCjZ9Ckq0BcuD:-NXpBAxt3AwrX_XQiAC2:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/View%20controllers/RedeemVoucher/AccountRedeemVoucherController.swift#L21)):*

`UINavigationControllerDelegate` seems to be unused.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/AccountRedeemVoucherController.swift` line 56 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXoz67k9_SSuRT7IYo7:-NXpBdgS2PnWE_LqCl7X:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/View%20controllers/RedeemVoucher/AccountRedeemVoucherController.swift#L56)):*

You can omit `MullvadREST.`
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherInteractor.swift` line 24 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXozF0MAGJDHKPKRBA2:-NXpBs7c7RoLTFzqcTlI:b38t83m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherInteractor.swift#L24)):*

Nit: cold probably fit on the same line
</blockquote></details>

it's formatter preferences

pronebird (changes requested):
> Reviewed 3 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: 17 of 21 files reviewed, 2 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift` line 61 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXpJUmkEDYo2Pkoerno:-NXpJUmkEDYo2Pkoernp:bgid2o2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift#L61)):*
> ```Swift
>                 interactor: RedeemVoucherInteractor(tunnelManager: interactor.tunnelManager)
>             )
>             coordinator.start()
> ```

`AccountCoordinator` implements `Presenting` so the parent could initiate modal presentation, i.e:

```swift
// Will push controller into navigation context and start the flow
coordinator.start()

// Present navigation controller modally, optionally providing presentation configuration
self.presentChild(coordinator, animated: true, configuration: ModalPresentationConfiguration(...))
```

`AccountCoordiantor` should then maintain strong reference to `coordinator` until it's dismissed (via `coordinator.dismiss(animated:)`
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 172 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXorHn2EyBazrpU4ZTR:-NXpIdx493I9ZxsZSGZz:bdimuzn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L172)):*

we support `portrait` orientaion on iPhone, there is no need to worry about rotation on iPhone.please,correct me if I'm wrong.
when it comes to iPad.I've tested multi tasking I didn't see any problem.
</blockquote></details>

iPad supports rotation but if you reckon it works then fine by me
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherInteractor.swift` line 24 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXozF0MAGJDHKPKRBA2:-NXpJ0jC3u3g5ynuUtCV:ban46ew) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherInteractor.swift#L24)):*

it's formatter preferences 
</blockquote></details>

It has issues when breaking down long lines so I usually make sure it does not create a deep indentation like this where there is enough space to fit the whole line of code.

pronebird (changes requested):
> Reviewed 2 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: 19 of 21 files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherCoordinator.swift` line 32 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXpLWy5-1WLORdweSxr:-NXpLWy6BKx2bFmoJl-V:buqlqnq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherCoordinator.swift#L32)):*
> ```Swift
>         let controller = AccountRedeemVoucherController(interactor: interactor)
> 
>         navigationController.present(controller, animated: true)
> ```

Hmmm I thought we were pushing redeem voucher into navigation controller or have I misunderstood this?

pronebird (commented):
> Reviewed 2 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 7 unresolved discussions (waiting on @mojganii)

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: 14 of 22 files reviewed, 18 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift` line 61 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXpJUmkEDYo2Pkoerno:-NXtcRgxC6DffvP-qJbU:b-7rl0jr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift#L61)):*

`AccountCoordinator` implements `Presenting` so the parent could initiate modal presentation, i.e:

```swift
// Will push controller into navigation context and start the flow
coordinator.start()

// Present navigation controller modally, optionally providing presentation configuration
self.presentChild(coordinator, animated: true, configuration: ModalPresentationConfiguration(...))
```

`AccountCoordiantor` should then maintain strong reference to `coordinator` until it's dismissed (via `coordinator.dismiss(animated:)`
</blockquote></details>

done 
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 279 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtbNiN608m5JeuCUbr:-NXu7cBCBh5OxE2Lfe27:b-ffiht1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L279)):*

It's nice to semantically have everything under UIMetrics, but in my opinion collecting all margins and such in one place is the main reason for it. Putting it into an extension means spreading it out over many files, which sort of defeats the purpose.

There are more instances of this in the PR.
</blockquote></details>

I disagree to put all of them at the same file when we are using them once.it depends on the domain.if it's using in different places we should put it in the main file otherwise we can put it at the file with the restricted access level.the goal of UIMetrics is name spacing as I understood 
___
*[`ios/MullvadVPN/View controllers/Account/AccountContentView.swift` line 33 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtcS5WAtmuzUsqpY2k:-NXtukou5kaMxCkRCJOJ:bb7f2mf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/Account/AccountContentView.swift#L33)):*

Views inside stackviews don't implicitly need to set `translatesAutoresizingMaskIntoConstraints`.

There are more instances of this in the PR.
</blockquote></details>

done
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherContentView.swift` line 5 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtfsvW16m7rA9fUwuZ:-NXu8ihaAzcDEoszm3wd:bl95zby) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherContentView.swift#L5)):*

Author is Mojgan? :)
</blockquote></details>

other people worked on it before I've been taking over the task and polishing some places
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherContentView.swift` line 169 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXth0tt3U0lBaMtavV7:-NXu99hqEbzHj8aCz-M3:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherContentView.swift#L169)):*

"isRedeemButtonEnabled" reads better I think.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherContentView.swift` line 247 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXti3ZKE_AcGddc6y3y:-NXuAL2G9ZezanZozGdJ:b-2hwweg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherContentView.swift#L247)):*

You can/should use `view.addConstrainedSubviews` instead, along with functions like `view.pinEdges` for layouting constraints. 
</blockquote></details>

I prefer to use the classic way.is the extension fully tested?
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherCoordinator.swift` line 32 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXpLWy5-1WLORdweSxr:-NXuAk1f7p5xqy1UyVqV:b-mk5v4x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherCoordinator.swift#L32)):*

Hmmm I thought we were pushing redeem voucher into navigation controller or have I misunderstood this?
</blockquote></details>

let's take a look at the latest changes
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift` line 5 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtjPng7YMDHs3GnQYo:-NXuAxhM1fXdcdIsa162:brdy8ye) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift#L5)):*

Author is Mojgan?
</blockquote></details>

Same as before.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift` line 102 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtllMRAUGF5A_GIrHD:-NXuB3gT1kzz_9yzBC8o:bjquptk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift#L102)):*

You can/should use `view.addConstrainedSubviews` instead, along with functions like `view.pinEdges` for layouting constraints. 
</blockquote></details>

same as before
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherViewController.swift` line 57 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtmPrW4ZoMwUEAkp3f:-NXuBPF0BQmfQYgS_PZw:bo2lfrt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherViewController.swift#L57)):*

It doesn't seem like contentView does anything bad if we try to enable it while already enabled. If it would, however, can't we put the failsafe inside contentView instead (ie, closer to the source)?
</blockquote></details>

Nope,in content view based on this flag makes decision to show or hide keyboard.so we cannot put the code there.moreover, `viewWillAppear` often calls multiple times in view life cycle  hierarchy  that's why I need to be sure it's the first time and I'll make the textfield ready to input
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherViewController.swift` line 73 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtmihf9R2u9zqe64NG:-NXuChNP5NML2mjAr7iu:bjquptk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherViewController.swift#L73)):*

You can/should use `view.addConstrainedSubviews` instead, along with functions like `view.pinEdges` for layouting constraints. 
</blockquote></details>

same as before
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/VoucherTextField.swift` line 13 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtmuy03HXW9CA89O_M:-NXuDwLvEYmK6O-tJLwb:b-convqf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/VoucherTextField.swift#L13)):*

Not sure if this is better, but you could calculate the length to be sure it's never out of sync with group size / max groups:
</blockquote></details>

the algorithm works when the valid length is multiple of group size.when the minimum valid length  of voucher code would be 9 character  but the length of  voucher  code can be 19, the algorithm doesn't work.we are using `voucherLength` to check if the current code meets our exception or not.
by the way,In our case I can use you snip code but based on the case it's not always true                     
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/VoucherTextField.swift` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtiw5t9nLBnma2p8AQ:-NXuO09xA8FKvHtr3zWD:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/VoucherTextField.swift#L24)):*

"isVoucherLengthSatisfied" reads better I think.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/AccountRedeemVoucherController.swift` line 21 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXpO090CeZ9wBQQodnu:-NXtccPX69J5AOc6whSE:ba3dfob) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/AccountRedeemVoucherController.swift#L21)):*

This controller behaves as coordinator. Instead flow management should go into coordinator that you have introduced, including perhaps delegate, which can be replaced with one or two closures.
</blockquote></details>

Removed
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/AccountRedeemVoucherController.swift` line 46 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXpMJDQDbQ20Bt-WkDW:-NXtciGb60560ozHH2uB:bb7f2mf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/AccountRedeemVoucherController.swift#L46)):*

Agreed, seems more in line with the current coordinator pattern.
</blockquote></details>

done
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/AccountRedeemVoucherController.swift` line 53 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtf_ACBV3of1YD_pBC:-NXtvq4R0GKWmlrx4GQS:bb7f2mf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/AccountRedeemVoucherController.swift#L53)):*

Nit: "didSucceed"
</blockquote></details>

done
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/AccountRedeemVoucherController.swift` line 59 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXpLnpy0Xuao-N9bDPI:-NXtdAIZD9f5ccClvlhr:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/AccountRedeemVoucherController.swift#L59)):*

The point of introducing coordinator was to make sure that it's responsible for creating new child controllers and pushing them into navigation stack in response to delegate calls.
</blockquote></details>

Done
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/AccountRedeemVoucherController.swift` line 69 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXpOeSI-JmTpvVEJh6i:-NXu20Vu4q5ZDfbnnsJ8:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/AccountRedeemVoucherController.swift#L69)):*

Parent coordinator should dismiss child.. Controllers should not dismiss themselves etc../
</blockquote></details>

Done.

rablador (changes requested):
> Reviewed 7 of 8 files at r4, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 13 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift` line 62 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NY25mOL08TjkDdP9Ymn:-NY25mOL08TjkDdP9Ymo:b-4atgty) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62d52237f7ed3b1561af101851b14c5ed5e1ed02/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift#L62)):*
> ```Swift
>             )
>             coordinator.didFinish = { [weak self] coordinator in
>                 self?.dismiss(animated: true)
> ```

More of a question here, but should one dismiss on self or on the coordinator? I know that both works. @pronebird 
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 279 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtbNiN608m5JeuCUbr:-NY26-r56nVLK3utbQzv:bg20ru8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L279)):*

I disagree to put all of them at the same file when we are using them once.it depends on the domain.if it's using in different places we should put it in the main file otherwise we can put it at the file with the restricted access level.the goal of UIMetrics is name spacing as I understood 
</blockquote></details>

Ok, that's a fair point, but on the other hand, if we don't collect them centrally, how do we know if there's anything to reuse when similar metrics are needed?
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherContentView.swift` line 247 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXti3ZKE_AcGddc6y3y:-NY26ov4Ev33F2UA-7TD:b-t0mpq4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherContentView.swift#L247)):*

I prefer to use the classic way.is the extension fully tested?
</blockquote></details>

It's not covered by tests I think, but it's getting more and more use. I think this is more a matter of reducing boiler plate and standardising how we handle autolayout.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherViewController.swift` line 57 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtmPrW4ZoMwUEAkp3f:-NY29W0w7H_APagwkGkr:bcx155t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherViewController.swift#L57)):*

Nope,in content view based on this flag makes decision to show or hide keyboard.so we cannot put the code there.moreover, `viewWillAppear` often calls multiple times in view life cycle  hierarchy  that's why I need to be sure it's the first time and I'll make the textfield ready to input
</blockquote></details>

I was thinking that a similar function could be inside contentView instead of the property setter, thereby letting outside callers simply try to enable without doing checks. But it's not important, more a suggestion.

pronebird (changes requested):
> Reviewed 5 of 8 files at r4, 3 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 8 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift` line 62 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NY25mOL08TjkDdP9Ymn:-NY2xOIf3UojSTROXrK5:b2ogekp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62d52237f7ed3b1561af101851b14c5ed5e1ed02/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift#L62)):*

More of a question here, but should one dismiss on self or on the coordinator? I know that both works. @pronebird 
</blockquote></details>

My preference is usually to put this responsibility on parent. Parent presents, parent dismisses its child. However, shouldn't this call do `coordinator.dismiss(true)` instead?
___
*[`ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift` line 65 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NY2y3R71yz9Ar_DZn7a:-NY2y3R71yz9Ar_DZn7b:b12w4az) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62d52237f7ed3b1561af101851b14c5ed5e1ed02/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift#L65)):*
> ```Swift
>             }
>             coordinator.didCancel = { [weak self] coordinator in
>                 self?.dismiss(animated: true)
> ```

Shouldn't this call do `coordinator.dismiss(animated: true)` instead?
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 279 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtbNiN608m5JeuCUbr:-NY36Ddf7RUBA2sf0JH3:byfu4vh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L279)):*

Ok, that's a fair point, but on the other hand, if we don't collect them centrally, how do we know if there's anything to reuse when similar metrics are needed?
</blockquote></details>

This refactoring seems to be out of scope of this PR and I always ask you guys and girls to keep each PR as scoped and focused on solving a particular issue as possible. Because each change draws scrutiny, comments, opinions, etc... However keep the change since you've already implemented it.

Historically the point of `UIMetrics` and `UIColor+Pallete` is not only reusability but also customization point for the entire UI. To make it easier to adjust colors and margins etc... This happened before with desktop where margins, sizes etc. were adjusted once. I think it's much easier to do that when certain metrics are concentrated in one place than when they are spread across the entire codebase.

Frankly I think `UIColor+Pallete` and `UIMetrics` seem to be connected quite a bit, so perhaps we could combine both of them in the future the way Mojgan did it here. 
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 159 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NY3EJqqCEdwz_XF8_2Y:-NY3EJqqCEdwz_XF8_2Z:b7b6p8f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62d52237f7ed3b1561af101851b14c5ed5e1ed02/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L159)):*
> ```Swift
> 
>     private func addKeyboardResponder() {
>         keyboardResponder = .init(targetView: presentedView!, handler: { [weak self] view, adjustment in
> ```

Force unwrapping.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherContentView.swift` line 247 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXti3ZKE_AcGddc6y3y:-NY2zjSMAhjJbs2HUJd_:b-npbupn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherContentView.swift#L247)):*

It's not covered by tests I think, but it's getting more and more use. I think this is more a matter of reducing boiler plate and standardising how we handle autolayout.
</blockquote></details>

FYI

It's not fully tested but we're working on improving it and we could certainly add tests for it or improve the interface to reduce amount of boilerplate. More people involved would also help to improve and test it. 

What you prefer is certainly a valid way to build constraints but eventually you'll realize that you repeat a lot of code and that many things can be done faster with helper tools. Such as this is why you use stack views. Some of the most common constraints are the ones where you need to pin views to edges.

I don't insist on hard introduction into our constraint builder but I think eventually we'll have to reduce amount of boilerplate and that seems to be one way to do so.

A lot of code here can be replaced with:

```
voucherCodeStackView.pinEdgesToSuperviewMargins(.all().excluding(.bottom))
actionsStackView.pinEdgesToSuperviewMargins(.all().excluding(.top))
```

`view.addConstrainedSubviews()` also makes sure to set `translatesAutoresizingMaskIntoConstraints = false` for you.

pronebird (changes requested):
> Reviewed 1 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 9 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherSucceededCoordinator.swift` line 11 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NY3HkAJ9jlb4-gNDUTc:-NY3HkAJ9jlb4-gNDUTd:b-fjaavr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62d52237f7ed3b1561af101851b14c5ed5e1ed02/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherSucceededCoordinator.swift#L11)):*
> ```Swift
> import Foundation
> import UIKit
> ```

This coordinator is mostly empty and not doing much of coordination. Unless I miss something, I think we can merge it into ` RedeemVoucherCoordinator`.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: 14 of 22 files reviewed, 9 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift` line 65 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NY2y3R71yz9Ar_DZn7a:-NY3GxRgARjo4C1zBZhT:b-896fjx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62d52237f7ed3b1561af101851b14c5ed5e1ed02/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift#L65)):*

Shouldn't this call do `coordinator.dismiss(animated: true)` instead?
</blockquote></details>

Done
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherContentView.swift` line 247 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXti3ZKE_AcGddc6y3y:-NY3Hy5b4oI_HIynyRrJ:birdvb2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherContentView.swift#L247)):*

FYI

It's not fully tested but we're working on improving it and we could certainly add tests for it or improve the interface to reduce amount of boilerplate. More people involved would also help to improve and test it. 

What you prefer is certainly a valid way to build constraints but eventually you'll realize that you repeat a lot of code and that many things can be done faster with helper tools. Such as this is why you use stack views. Some of the most common constraints are the ones where you need to pin views to edges.

I don't insist on hard introduction into our constraint builder but I think eventually we'll have to reduce amount of boilerplate and that seems to be one way to do so.

A lot of code here can be replaced with:

```
voucherCodeStackView.pinEdgesToSuperviewMargins(.all().excluding(.bottom))
actionsStackView.pinEdgesToSuperviewMargins(.all().excluding(.top))
```

`view.addConstrainedSubviews()` also makes sure to set `translatesAutoresizingMaskIntoConstraints = false` for you.
</blockquote></details>

I got your point.before we were using `SnapKit` for avoiding repetition.but let me use this extension when it's fully tested.

pronebird (changes requested):
> Reviewed 1 of 4 files at r5, 7 of 8 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 8 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadVPN/UI appearance/UIMetrics.swift` line 31 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NY4IM9k97vLUMrho9_T:-NY4IM9k97vLUMrho9_U:bjp3ryd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/UI%20appearance/UIMetrics.swift#L31)):*
> ```Swift
>     }
> 
>     enum Transition {
> ```

`Transition` is a very generic word as it can refer to any kind of transition. Maybe `FormSheetTransition`?

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: 20 of 22 files reviewed, 8 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift` line 62 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NY25mOL08TjkDdP9Ymn:-NYHWdzIENi1udFEY7Wl:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62d52237f7ed3b1561af101851b14c5ed5e1ed02/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift#L62)):*

My preference is usually to put this responsibility on parent. Parent presents, parent dismisses its child. However, shouldn't this call do `coordinator.dismiss(true)` instead?
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 279 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtbNiN608m5JeuCUbr:-NYHWkV_8541hkIIdjRo:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L279)):*

This refactoring seems to be out of scope of this PR and I always ask you guys and girls to keep each PR as scoped and focused on solving a particular issue as possible. Because each change draws scrutiny, comments, opinions, etc... However keep the change since you've already implemented it.

Historically the point of `UIMetrics` and `UIColor+Pallete` is not only reusability but also customization point for the entire UI. To make it easier to adjust colors and margins etc... This happened before with desktop where margins, sizes etc. were adjusted once. I think it's much easier to do that when certain metrics are concentrated in one place than when they are spread across the entire codebase.

Frankly I think `UIColor+Pallete` and `UIMetrics` seem to be connected quite a bit, so perhaps we could combine both of them in the future the way Mojgan did it here. 

</blockquote></details>

Done.
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 159 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NY3EJqqCEdwz_XF8_2Y:-NYHYUjO6_nNtXoUuhgk:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62d52237f7ed3b1561af101851b14c5ed5e1ed02/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L159)):*

Force unwrapping.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/UI appearance/UIMetrics.swift` line 31 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NY4IM9k97vLUMrho9_T:-NYHWzBZBY6CRYI5q_bJ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/UI%20appearance/UIMetrics.swift#L31)):*

`Transition` is a very generic word as it can refer to any kind of transition. Maybe `FormSheetTransition`?
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherInteractor.swift` line 24 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXozF0MAGJDHKPKRBA2:-NYHXDXy-AxyUmXItXUn:b4mpq3u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherInteractor.swift#L24)):*

It has issues when breaking down long lines so I usually make sure it does not create a deep indentation like this where there is enough space to fit the whole line of code.
</blockquote></details>

when I change this CI/CD pipe line doesn't pass swift format step.let it stay with current format.

buggmagnet (changes requested):
> Reviewed 6 of 20 files at r1, 1 of 6 files at r2, 2 of 7 files at r3, 1 of 8 files at r4, 1 of 4 files at r5, 5 of 8 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: 20 of 22 files reviewed, 12 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift` line 62 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NY25mOL08TjkDdP9Ymn:-NYHPvXB8Cm7fiq-CVQv:b-ig5rtp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62d52237f7ed3b1561af101851b14c5ed5e1ed02/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift#L62)):*
Just my 2 cents regarding dismissing in general, if we take a look at [UIKit dismissal on view controllers](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621505-dismiss) it says the following
> The presenting view controller is responsible for dismissing the view controller it presented. If you call this method on the presented view controller itself, UIKit asks the presenting view controller to handle the dismissal.

We should try to follow that philosophy if possible because that means we can assume the same functioning model as UIKit in regards to dismissal, which in turns makes it easier to reason about our own code
___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 159 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHNwxU7s61SEX8OExr:-NYHNwxU7s61SEX8OExs:bs3kdwd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L159)):*
> ```Swift
> 
>     private func addKeyboardResponder() {
>         keyboardResponder = .init(targetView: presentedView!, handler: { [weak self] view, adjustment in
> ```

Please only use `.init` when the [receiver is a metatype](https://google.github.io/swift/#initializers-1)

The same goes for line 165
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherContentView.swift` line 247 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXti3ZKE_AcGddc6y3y:-NYHSx8DEdXF7PgxDPEi:b7fkd1t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherContentView.swift#L247)):*

I got your point.before we were using `SnapKit` for avoiding repetition.but let me use this extension when it's fully tested. 
</blockquote></details>

@mojganii  I understand your reticense in refusing to use untested code, but the point of having that extension is not just about preferences, it's also about having a uniform code. 

I'm sure @pronebird thought really hard about how to improve the situation with regards to Autolayout, and we should not let his efforts go to waste "because it's untested" otherwise, we wouldn't be making any progress.

If you have genuine reasons to not use the proposed extensions, please make a case for it, otherwise I would like to ask kindly that you reconsider the suggestions made.
By not wanting to do that, not only you make it harder for the next person that will read the code, but it also creates unnecessary tension in the team, and nobody wins in this situation.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherCoordinator.swift` line 16 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHWGRJ7lSvfNp4EHkb:-NYHWGRJ7lSvfNp4EHkc:b-jmur8w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherCoordinator.swift#L16)):*
> ```Swift
>     private let navigationController: UINavigationController
>     private let viewController: RedeemVoucherViewController
>     var didFinish: ((RedeemVoucherCoordinator) -> Void)?
> ```

I looked at where `didFinish` and `didCancel` were used, and I'm not sure why we have those. 

The only place I found was in `AccountCoordinator` and we do the following 

```
            coordinator.didFinish = { redeemVoucherCoordinator in
                redeemVoucherCoordinator.dismiss(animated: true)
            }
            coordinator.didCancel = { redeemVoucherCoordinator in
                redeemVoucherCoordinator.dismiss(animated: true)
            }
```

Unless I'm misunderstanding something, we don't need to do that, and we could do the following instead

```
extension RedeemVoucherCoordinator: RedeemVoucherViewControllerDelegate {
    func redeemVoucherDidSucceed(
        _ controller: RedeemVoucherViewController,
        with response: REST.SubmitVoucherResponse
    ) {
        let viewController = RedeemVoucherSucceededViewController(timeAddedComponents: response.dateComponents)
        viewController.delegate = self
        navigationController.pushViewController(viewController, animated: true)
    }

    func redeemVoucherDidCancel(_ controller: RedeemVoucherViewController) {
        dismiss(animated: true)
    }
}

extension RedeemVoucherCoordinator: RedeemVoucherSucceededViewControllerDelegate {
    func redeemVoucherSucceededViewControllerDidFinish(_ controller: RedeemVoucherSucceededViewController) {
        dismiss(animated: true)
    }
}

```
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift` line 142 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHZAiQ7bFkpPSPUP9m:-NYHZAiQ7bFkpPSPUP9n:b-twlli9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift#L142)):*
> ```Swift
> }
> 
> private extension DateComponents {
> ```

This shouldn't be an extension on `DateComponents` because we don't reuse it anywhere else, and it's just needed here.

I suggest turning this into a function on `RedeemVoucherSucceededViewController` instead
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/VoucherTextField.swift` line 54 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHYIbB3r52n_8lJO6g:-NYHYIbB3r52n_8lJO6h:biyr9l2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/View%20controllers/RedeemVoucher/VoucherTextField.swift#L54)):*
> ```Swift
>         replacementString string: String
>     ) -> Bool {
>         return inputFormatter.textField(
> ```

We can omit `return` here

pronebird (commented):
> Reviewed 1 of 2 files at r8, 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 10 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherInteractor.swift` line 24 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXozF0MAGJDHKPKRBA2:-NYHiK0z7byfHssItnbj:bebk72p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherInteractor.swift#L24)):*

when I change this CI/CD pipe line doesn't pass swift format step.let it stay with current format.
</blockquote></details>

Try re-formatting code after your change. Usually when the string fits on the same line, then it will keep it untouched. If it breaks it on multiple lines after re-formatting then leave it the way the formatter likes it. Otherwise, I am not blocking PR because of formatting.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift` line 142 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHZAiQ7bFkpPSPUP9m:-NYHjHbZ2nM7WcUZzD_S:bdr6n9b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift#L142)):*

This shouldn't be an extension on `DateComponents` because we don't reuse it anywhere else, and it's just needed here.

I suggest turning this into a function on `RedeemVoucherSucceededViewController` instead
</blockquote></details>

I don't disagree, but note that it's a private extension so it will not be visible outside of this file.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/VoucherTextField.swift` line 54 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHYIbB3r52n_8lJO6g:-NYHjf8D0Puik0dfvZXG:b-jedylm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/View%20controllers/RedeemVoucher/VoucherTextField.swift#L54)):*

We can omit `return` here
</blockquote></details>

I think this has to be addressed with formatter, instead of case-by-case, because it's not critical (the code is functional with or without return statement) but it stalls the review.

rablador (commented):
> Reviewed 5 of 8 files at r6, 1 of 1 files at r7, 1 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 10 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherCoordinator.swift` line 16 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHWGRJ7lSvfNp4EHkb:-NYHlP-j9rIutl_zRf8L:b-9ebcfl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherCoordinator.swift#L16)):*

I looked at where `didFinish` and `didCancel` were used, and I'm not sure why we have those. 

The only place I found was in `AccountCoordinator` and we do the following 

```
            coordinator.didFinish = { redeemVoucherCoordinator in
                redeemVoucherCoordinator.dismiss(animated: true)
            }
            coordinator.didCancel = { redeemVoucherCoordinator in
                redeemVoucherCoordinator.dismiss(animated: true)
            }
```

Unless I'm misunderstanding something, we don't need to do that, and we could do the following instead

```
extension RedeemVoucherCoordinator: RedeemVoucherViewControllerDelegate {
    func redeemVoucherDidSucceed(
        _ controller: RedeemVoucherViewController,
        with response: REST.SubmitVoucherResponse
    ) {
        let viewController = RedeemVoucherSucceededViewController(timeAddedComponents: response.dateComponents)
        viewController.delegate = self
        navigationController.pushViewController(viewController, animated: true)
    }

    func redeemVoucherDidCancel(_ controller: RedeemVoucherViewController) {
        dismiss(animated: true)
    }
}

extension RedeemVoucherCoordinator: RedeemVoucherSucceededViewControllerDelegate {
    func redeemVoucherSucceededViewControllerDidFinish(_ controller: RedeemVoucherSucceededViewController) {
        dismiss(animated: true)
    }
}

```
</blockquote></details>

Isn't this touching on the same idea further up in the review that parent controllers should dismiss their children? I seems we might not need two callbacks leading to the same thing though.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 10 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherCoordinator.swift` line 16 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHWGRJ7lSvfNp4EHkb:-NYHo02f0rR7bRfUda26:bci8e7u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherCoordinator.swift#L16)):*

Isn't this touching on the same idea further up in the review that parent controllers should dismiss their children? I seems we might not need two callbacks leading to the same thing though.
</blockquote></details>

We can replace two callbacks with one, for example only leave `didFinish()` and call it in either case.

However since the parent coordinator `AccountCoordinator` presents the `RedeemVoucherCoordinator` modally, it should also be responsible for dismissing it.

Same principle in relation to UIKit was mentioned by @buggmagnet in the review for some other PR.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 10 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift` line 159 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHNwxU7s61SEX8OExr:-NYH_kr37u5vWSaVdF65:bq2lusx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/Presentation%20controllers/FormsheetPresentationController.swift#L159)):*

Please only use `.init` when the [receiver is a metatype](https://google.github.io/swift/#initializers-1)

The same goes for line 165
</blockquote></details>

I forgot.done
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherContentView.swift` line 247 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXti3ZKE_AcGddc6y3y:-NYHaWz_7vP3wPxVzxRn:baguaa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherContentView.swift#L247)):*

@mojganii  I understand your reticense in refusing to use untested code, but the point of having that extension is not just about preferences, it's also about having a uniform code. 

I'm sure @pronebird thought really hard about how to improve the situation with regards to Autolayout, and we should not let his efforts go to waste "because it's untested" otherwise, we wouldn't be making any progress.

If you have genuine reasons to not use the proposed extensions, please make a case for it, otherwise I would like to ask kindly that you reconsider the suggestions made.
By not wanting to do that, not only you make it harder for the next person that will read the code, but it also creates unnecessary tension in the team, and nobody wins in this situation.
</blockquote></details>

It doesn't mean I want not to consider the his efforts and my comment doesn't mean I don't want to use as I mentioned `let me replace when it's fully tested`.it's about UI breaking.I don't think so using extension for the important part like UI without fully tested is a protest.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherCoordinator.swift` line 16 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHWGRJ7lSvfNp4EHkb:-NYHaVOX8CcIfjG8QK7o:b-l306pm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherCoordinator.swift#L16)):*

I looked at where `didFinish` and `didCancel` were used, and I'm not sure why we have those. 

The only place I found was in `AccountCoordinator` and we do the following 

```
            coordinator.didFinish = { redeemVoucherCoordinator in
                redeemVoucherCoordinator.dismiss(animated: true)
            }
            coordinator.didCancel = { redeemVoucherCoordinator in
                redeemVoucherCoordinator.dismiss(animated: true)
            }
```

Unless I'm misunderstanding something, we don't need to do that, and we could do the following instead

```
extension RedeemVoucherCoordinator: RedeemVoucherViewControllerDelegate {
    func redeemVoucherDidSucceed(
        _ controller: RedeemVoucherViewController,
        with response: REST.SubmitVoucherResponse
    ) {
        let viewController = RedeemVoucherSucceededViewController(timeAddedComponents: response.dateComponents)
        viewController.delegate = self
        navigationController.pushViewController(viewController, animated: true)
    }

    func redeemVoucherDidCancel(_ controller: RedeemVoucherViewController) {
        dismiss(animated: true)
    }
}

extension RedeemVoucherCoordinator: RedeemVoucherSucceededViewControllerDelegate {
    func redeemVoucherSucceededViewControllerDidFinish(_ controller: RedeemVoucherSucceededViewController) {
        dismiss(animated: true)
    }
}

```
</blockquote></details>

as a few before comments, the current preferences, the parent is responsible for starting and ending the flow 
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift` line 142 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHZAiQ7bFkpPSPUP9m:-NYHdJfV0tjblUbm8nM9:bqloxf7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift#L142)):*

I don't disagree, but note that it's a private extension so it will not be visible outside of this file.
</blockquote></details>

it's with private access level.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 10 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherInteractor.swift` line 24 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXozF0MAGJDHKPKRBA2:-NYHueymEJJy3FrfbGLe:b7aigzr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b62cca4be3219fb53cf484b5293a2fbe841e9cea/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherInteractor.swift#L24)):*

Try re-formatting code after your change. Usually when the string fits on the same line, then it will keep it untouched. If it breaks it on multiple lines after re-formatting then leave it the way the formatter likes it. Otherwise, I am not blocking PR because of formatting.
</blockquote></details>

I did the instruction you've mentioned.but formatter is eager to break it into multiple lines.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/VoucherTextField.swift` line 54 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHYIbB3r52n_8lJO6g:-NYHtfVK5hVvDgNXPGsL:bgnb3iu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/View%20controllers/RedeemVoucher/VoucherTextField.swift#L54)):*

I think this has to be addressed with formatter, instead of case-by-case, because it's not critical (the code is functional with or without return statement) but it stalls the review.
</blockquote></details>

I agree.there are different places with different format.It should be addressed by formatter

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: 19 of 22 files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherContentView.swift` line 247 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXti3ZKE_AcGddc6y3y:-NYIbex8D0KTSKavY9ZG:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherContentView.swift#L247)):*

It doesn't mean I want not to consider the his efforts and my comment doesn't mean I don't want to use as I mentioned `let me replace when it's fully tested`.it's about UI breaking.I don't think so using extension for the important part like UI without fully tested is a protest.
</blockquote></details>

Done
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift` line 102 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtllMRAUGF5A_GIrHD:-NYIbpf87HxbPPe_V2Bh:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherSucceededViewController.swift#L102)):*

same as before
</blockquote></details>

Done
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherViewController.swift` line 73 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NXtmihf9R2u9zqe64NG:-NYIbujP9r7YZ8E98AwB:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b3af3c690ed50853e5fd08f6e4a4c08f71e7bba7/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherViewController.swift#L73)):*

same as before
</blockquote></details>

Done

buggmagnet (approved):
> Reviewed 1 of 2 files at r8, 1 of 1 files at r9, 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 6 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/VoucherTextField.swift` line 54 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-NYHYIbB3r52n_8lJO6g:-NYIczKUAaAwJNkbGsi1:bkpz1cc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/41df1ade144805aab4224efc27e061a023e80983/ios/MullvadVPN/View%20controllers/RedeemVoucher/VoucherTextField.swift#L54)):*

I agree.there are different places with different format.It should be addressed by formatter
</blockquote></details>

It's just good housekeeping. We don't need to wait to have a formatter to have good coding practices.

rablador (commented):
> Reviewed 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @mojganii, and @pronebird)

rablador (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-:-NYNQ40z9VYaCtLbupzO:bnfp4nl)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @mojganii, and @pronebird)

pronebird (commented):
> Dismissed @rablador from a discussion.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @mojganii)

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789#-:-NYNZeSCCj6kFGJpm1kW:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @mojganii)

pronebird (approved):
> Reviewed 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @mojganii)

buggmagnet (approved):
> Reviewed 1 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4789)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii and @rablador)

PR #4796, opened 2023-06-15
When trying to revoke a device from the `Too many devices view`, it doesn't happen 

mojganii:
When the number of device is exceeded and we are trying to revoke a device from the `Too many devices view`,nothing happens.
after checking the code I figured out we are removing the buttons handlers in `viewDidDisappear` while after dismissing the controller(the `viewDidDisappear ` will be called by view lifecycle sooner than dismiss's `completion `) we are looking for them at `completion`, therefore It causes the button callbacks never be accessible.
for solving the problem I did this step :

- removing handlers at `dismiss` callback

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796)

### DIFF
--- a/ios/MullvadVPN/Classes/CustomAlertViewController.swift
+++ b/ios/MullvadVPN/Classes/CustomAlertViewController.swift
@@ -94,11 +94,6 @@ class CustomAlertViewController: UIViewController {
         }
     }
 
-    override func viewDidDisappear(_ animated: Bool) {
-        super.viewDidDisappear(animated)
-        handlers.removeAll()
-    }
-
     func addAction(title: String, style: ActionStyle, handler: (() -> Void)? = nil) {
         // The presence of a button should reset any custom button margin to default.
         containerView.directionalLayoutMargins.bottom = UIMetrics.CustomAlert.containerMargins.bottom
@@ -183,8 +178,9 @@ class CustomAlertViewController: UIViewController {
             if let handler = handlers[button] {
                 handler()
             }
-
             didDismiss?()
+            didDismiss = nil
+            handlers.removeAll()
         }
     }
 }
--- a/ios/MullvadVPN/Operations/PresentAlertOperation.swift
+++ b/ios/MullvadVPN/Operations/PresentAlertOperation.swift
@@ -37,8 +37,8 @@ final class PresentAlertOperation: AsyncOperation {
     }
 
     override func main() {
-        alertController.didDismiss = {
-            self.finish()
+        alertController.didDismiss = { [weak self] in
+            self?.finish()
         }
 
         presentingController.present(alertController, animated: true) {


rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796)* status: :shipit: complete! all files reviewed, all discussions resolved

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796)* status: all files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/Classes/CustomAlertViewController.swift` line 45 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796#-NY2Rwbq9HarWdt16Dy1:-NY2Rwbq9HarWdt16Dy2:bsk36ar) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2ea49484e3207a17c92e6abf204efdb6f6be95c/ios/MullvadVPN/Classes/CustomAlertViewController.swift#L45)):*
> ```Swift
>     }
> 
>     var didDismiss: (() -> Void)?
> ```

It turns out that the way we use this closure creates a retain cycle. 

In `PresentAlertOperation` we do the following 

```
        alertController.didDismiss = { // self is implicitly strongly captured here
            self.finish()
        }
```

so now, the `CustomAlertViewController.didDismiss` holds a strong reference to `PresentAlertOperation` 
At the same time, `PresentAlertOperation` holds a strong reference to `CustomAlertViewController` (via its constructor)

We should do 2 things here

* set `didDismiss` to nil in `didTapButton` after doing `handlers.removeAll()` 
* Capture `self` weakly in `PresentAlertOperation.main` to avoid creating a retain cycle

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/Classes/CustomAlertViewController.swift` line 177 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796#-NY29pcBEjecpu-PxOWu:-NY2TVQ23Ht_Y27BTqPT:bhur5pu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2ea49484e3207a17c92e6abf204efdb6f6be95c/ios/MullvadVPN/Classes/CustomAlertViewController.swift#L177)):*

There's no need to use `unowned` here 
`CustomAlertViewController` is not retaining the closure, so there's no retain cycle that should happen.

In general, we should not have to use `unowned` as it's error prone and can lead to crashes.
[The docs](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/automaticreferencecounting/#Weak-and-Unowned-References) are quite explicit about this 

> define a capture as a weak reference when the captured reference may become nil at some point in the future.

Which is the case here, the `CustomAlertViewController` will become nil very soon, therefore capture it as a weak reference.

We should only ever use `unowned` if the reference will never become nil.
</blockquote></details>

The strong retainment was brought up during original review of the component here: https://reviewable.io/reviews/mullvad/mullvadvpn-app/4718#-NXKBWqqC3BISY5dqo6O
However, after discussion on standup today I'm thinking we might have misunderstood each other then.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/Classes/CustomAlertViewController.swift` line 45 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796#-NY2Rwbq9HarWdt16Dy1:-NY2iiJ_EnlCPORyGKxS:bqjuqlb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2ea49484e3207a17c92e6abf204efdb6f6be95c/ios/MullvadVPN/Classes/CustomAlertViewController.swift#L45)):*

It turns out that the way we use this closure creates a retain cycle. 

In `PresentAlertOperation` we do the following 

```
        alertController.didDismiss = { // self is implicitly strongly captured here
            self.finish()
        }
```

so now, the `CustomAlertViewController.didDismiss` holds a strong reference to `PresentAlertOperation` 
At the same time, `PresentAlertOperation` holds a strong reference to `CustomAlertViewController` (via its constructor)

We should do 2 things here

* set `didDismiss` to nil in `didTapButton` after doing `handlers.removeAll()` 
* Capture `self` weakly in `PresentAlertOperation.main` to avoid creating a retain cycle


</blockquote></details>

I agree to make the self `weak` for preventing from probably retain cycle,I've done it before but I forgot to commit .moreover, there is no need to make dismiss `nil`.if we have any retain cycles for this object making the pointer `nil` doesn't help because ARC doesn't make it release.
___
*[`ios/MullvadVPN/Classes/CustomAlertViewController.swift` line 177 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796#-NY29pcBEjecpu-PxOWu:-NY37za3DZYywx13DpaM:by9nhl8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2ea49484e3207a17c92e6abf204efdb6f6be95c/ios/MullvadVPN/Classes/CustomAlertViewController.swift#L177)):*

The strong retainment was brought up during original review of the component here: https://reviewable.io/reviews/mullvad/mullvadvpn-app/4718#-NXKBWqqC3BISY5dqo6O
However, after discussion on standup today I'm thinking we might have misunderstood each other then.
</blockquote></details>

this closure is `Non-escaping`.thats why it's not a good idea to consider `weak` or `unowned` for that.unless we care about delayed deallocation.I remove this

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796)* status: 1 of 2 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/Classes/CustomAlertViewController.swift` line 177 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796#-NY29pcBEjecpu-PxOWu:-NY3WuJ1Cjt3ipOpTKwk:b-o0b6dv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2ea49484e3207a17c92e6abf204efdb6f6be95c/ios/MullvadVPN/Classes/CustomAlertViewController.swift#L177)):*

this closure is `Non-escaping`.thats why it's not a good idea to consider `weak` or `unowned` for that.unless we care about delayed deallocation.I remove this
</blockquote></details>

Temporary retain. No problem here. This closure must be escaping too regardless what the docs say.
___
*[`ios/MullvadVPN/Classes/CustomAlertViewController.swift` line 181 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796#-NY3UaLT0xG1QwgSISe4:-NY3UaLT0xG1QwgSISe5:b-n1nfzb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ccf1e5999292f07bc0f506372a5982fc18a19944/ios/MullvadVPN/Classes/CustomAlertViewController.swift#L181)):*
> ```Swift
>                 handler()
>             }
>             self.didDismiss?()
> ```

Please set `didDismiss = nil` after calling it.
___
*[`ios/MullvadVPN/Operations/PresentAlertOperation.swift` line 41 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796#-NY3Uhye49Mh9ZDOc3WE:-NY3Uhye49Mh9ZDOc3WF:bneep11) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ccf1e5999292f07bc0f506372a5982fc18a19944/ios/MullvadVPN/Operations/PresentAlertOperation.swift#L41)):*
> ```Swift
>     override func main() {
>         alertController.didDismiss = { [weak self] in
> ```

Nit: No need for `guard let self`. Just call `self?.finish()` on a weak ref. Since you only have one statement in here, then it would be equivalent but with less effort.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadVPN/Classes/CustomAlertViewController.swift` line 45 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796#-NY2Rwbq9HarWdt16Dy1:-NYI_gvSE8qG58uBbOFl:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2ea49484e3207a17c92e6abf204efdb6f6be95c/ios/MullvadVPN/Classes/CustomAlertViewController.swift#L45)):*

I agree to make the self `weak` for preventing from probably retain cycle,I've done it before but I forgot to commit .moreover, there is no need to make dismiss `nil`.if we have any retain cycles for this object making the pointer `nil` doesn't help because ARC doesn't make it release.
</blockquote></details>

Done
___
*[`ios/MullvadVPN/Classes/CustomAlertViewController.swift` line 177 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796#-NY29pcBEjecpu-PxOWu:-NYIZ00D9NG4ptZCz_qn:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d2ea49484e3207a17c92e6abf204efdb6f6be95c/ios/MullvadVPN/Classes/CustomAlertViewController.swift#L177)):*

Temporary retain. No problem here. This closure must be escaping too regardless what the docs say.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/Classes/CustomAlertViewController.swift` line 181 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796#-NY3UaLT0xG1QwgSISe4:-NY3hZoOD12proRhIVmJ:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ccf1e5999292f07bc0f506372a5982fc18a19944/ios/MullvadVPN/Classes/CustomAlertViewController.swift#L181)):*

Please set `didDismiss = nil` after calling it.
</blockquote></details>

Done

buggmagnet (approved):
> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @pronebird)

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4796#-:-NYId3iN9wdCHFW7i4U-:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4811, opened 2023-06-21
Fix argument to `read` in build.sh that did not work

faern:
The arguments here were in the wrong order, so the read would not work. Since this has been like this since forever, obviously we don't execute this code path very much :sweat_smile: But now I fix it so it's possible to do `./build.sh --sign` without running it from `ci/buildserver-build.sh`

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4811)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -104,7 +104,7 @@ if [[ "$SIGN" == "true" ]]; then
             exit 1
         fi
         if [[ -z ${CSC_KEY_PASSWORD-} ]]; then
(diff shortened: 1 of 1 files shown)


raksooo (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4811)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4818, opened 2023-06-22
Update swiftformat to enforce Google's swift rules

buggmagnet:
This PR formalizes our use of Google's swift styleguide via the `swiftformat` tool.

After discussing some potentially controversial points with the team, I have synthesized the guidelines into what will be in this PR.

### *How to look at this PR ?*

Each commit will enable a set of rules and will show how the rule affected the codebase.
Therefore, please look at each commit individually to understand how a given rule will affect the codebase.

The last commit enables all the rules. As it turns out, we were doing a decent job already, so not much will be changed overall.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4818)

### DIFF
--- a/ios/.swiftformat
+++ b/ios/.swiftformat
@@ -6,10 +6,77 @@
 
 # format options
 --indent 4
+# Google suggests 100, but Mullvad historically uses 120
 --maxwidth 120
+# The cardinal rules of Google Swift style for line-wrapping are:
+  # If the entire declaration, statement, or expression fits on one line, then do that.
+  # All elements must fit on the same line, or each element must be on its own line.
 --wraparguments before-first
 --wrapcollections before-first
 --wrapternary before-operators
 --redundanttype inferred
 --ifdef no-indent
---disable initCoderUnavailable, redundantReturn, unusedArguments, redundantRawValues, preferKeyPath, extensionAccessControl, hoistTry
+# Use K&R brace style, which is what Apple also uses
+--allman false
+# Use consistent ordering for member modifiers.
+--modifierorder
+--disable                         \
+initCoderUnavailable,             \
+unusedArguments,                  \
+redundantRawValues,               \
+preferKeyPath,                    \
+extensionAccessControl,           \
+hoistTry,                         \
+# Do not reorganize declarations
+organizeDeclarations,             \
+# Do not reorder switch case statements
+sortedSwitchCases,                \
+# Do not add // MARK: ClassName before every type
+markTypes
+
+
+--rules                           \
+blankLinesAroundMark,             \
+blankLinesAtEndOfScope,           \
+blankLinesAtStartOfScope,         \
+blankLinesBetweenScopes,          \
+braces,                           \
+consecutiveBlankLines,            \
+consecutiveSpaces,                \
+duplicateImports,                 \
+elseOnSameLine,                   \
+emptyBraces,                      \
+enumNamespaces,                   \
+extensionAccessControl,           \
+hoistPatternLet,                  \
+indent,                           \
+leadingDelimiters,                \
+linebreakAtEndOfFile,             \
+redundantInit,                    \
+redundantParens,                  \
+redundantPattern,                 \
+redundantRawValues,               \
+redundantType,                    \
+redundantVoidReturnType,          \
+semicolons,                       \
+sortedImports,                    \
+spaceAroundBraces,                \
+spaceAroundBrackets,              \
+spaceAroundComments,              \
+spaceAroundGenerics,              \
+spaceAroundOperators,             \
+spaceAroundParens,                \
+spaceInsideBraces,                \
+spaceInsideBrackets,              \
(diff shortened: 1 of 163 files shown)


pronebird (changes requested):
> Reviewed 1 of 150 files at r1, 1 of 32 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4818)* status: 1 of 167 files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/.swiftformat` line 1 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4818#-NYX_1oB17fR-ni81Fe2:-NYX_1oB17fR-ni81Fe3:b-73k1h1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/611dababf675ff065c091cf33a1885781f74478b/ios/.swiftformat#L1)):*
> ```
> # file options
> ```

It's hard to follow very verbose configs, they need a lot of explanation etc.. so I'd suggest to focus on what we want different from defaults.

Most of the rules here are _on by default_ as documented in:

https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md

Also the use of code comments could be minimized. `Rules.md` is a good source of knowledge and we don't need to reiterate what's already documented for us. 

Side note: historically Mullvad used 100 char per line, and it's only recently that iOS team started using 120 because Swift is very verbose. Such comments can be omitted in my opinion or moved to git commit message or PR description.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4818)* status: 1 of 167 files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`ios/.swiftformat` line 1 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4818#-NYX_1oB17fR-ni81Fe2:-NYXwOO19TbczHvmMf2k:b80c6ap) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/611dababf675ff065c091cf33a1885781f74478b/ios/.swiftformat#L1)):*

It's hard to follow very verbose configs, they need a lot of explanation etc.. so I'd suggest to focus on what we want different from defaults.

Most of the rules here are _on by default_ as documented in:

https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md

Also the use of code comments could be minimized. `Rules.md` is a good source of knowledge and we don't need to reiterate what's already documented for us. 

Side note: historically Mullvad used 100 char per line, and it's only recently that iOS team started using 120 because Swift is very verbose. Such comments can be omitted in my opinion or moved to git commit message or PR description.
</blockquote></details>

It's default _today_ but that doesn't mean it will not change in the future. 
If `swiftformat` decides tomorrow to change defaults, we will have to spend time figuring out why. 
This way, we keep control of the changes, and we can update the format when we need to.

The comment about the number of lines is serving as a justification about why we diverge from Google's styleguide.
It's much nicer to find the explanation directly in the source file rather than looking for the commit that made the change and look for the commit message.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4818)* status: 0 of 167 files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/.swiftformat` line 1 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4818#-NYX_1oB17fR-ni81Fe2:-NYYWqDv06Jgh5GEBgDq:b5stu85) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/611dababf675ff065c091cf33a1885781f74478b/ios/.swiftformat#L1)):*
> It's default _today_ but that doesn't mean it will not change in the future.  
If `swiftformat` decides tomorrow to change defaults, we will have to spend time figuring out why.  
This way, we keep control of the changes, and we can update the format when we need to.

I understand your intention. The odds of this happening are low and community has been maintaining sane defaults for the past years.

Higher risk is appearance of new rules and old ones being deprecated/removed but thankfully our CI will break first when either of this happens as it pulls the latest swiftformat on each run.

I put a message on slack with a short list of advantages/disadvantages of each approach, at least the way I see them.

mojganii (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4818#-:-NYs3ewu8KoubZcffuwx:bnfp4nl)* status: 0 of 167 files reviewed, 1 unresolved discussion (waiting on @buggmagnet and @pronebird)

pinkisemils (commented):
> Reviewed 1 of 150 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4818)* status: 0 of 167 files reviewed, 1 unresolved discussion (waiting on @buggmagnet and @pronebird)

PR #4820, opened 2023-06-22
Handle pubkey-in-use error IOS-204

pronebird:
We use retry strategy when creating new device during log-in. In some cases it's possible that the server managed to process the request but the client hasn't been able to receive the response (timeout error). When that happens, the next retry attempt usually results in `pubkey-in-use` error since the same key that we attempt to push is already assigned to the device that had been created during the previous attempt.

In order to recover from that error, this PR introduces additional fetch request that attempts to find the previously created device by public key.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4820)

### DIFF
--- a/ios/MullvadVPN/TunnelManager/SetAccountOperation.swift
+++ b/ios/MullvadVPN/TunnelManager/SetAccountOperation.swift
@@ -12,6 +12,7 @@ import MullvadREST
 import MullvadTypes
 import Operations
 import class WireGuardKitTypes.PrivateKey
+import class WireGuardKitTypes.PublicKey
 
 enum SetAccountAction {
     /// Set new account.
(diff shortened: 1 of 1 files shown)


buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4820#-:-NYrhXad6gvfqpj7gDqG:bnfp4nl)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN/TunnelManager/SetAccountOperation.swift` line 358 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4820#-NYrgQzJ582JulD0J9zg:-NYrgQzJ582JulD0J9zh:b-41s3kc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/40d616308aafdc97508bbc3ca28a96dbc0ad52a2/ios/MullvadVPN/TunnelManager/SetAccountOperation.swift#L358)):*
> ```Swift
>                     if let error = result.error as? REST.Error, error.compareErrorCode(.publicKeyInUse) {
>                         self.findDevice(accountNumber: accountNumber, publicKey: privateKey.publicKey) { result in
>                             let result = result.flatMap { device -> Result<NewDevice, Error> in
> ```

nit: This can be `device in` 
___
*[`ios/MullvadVPN/TunnelManager/SetAccountOperation.swift` line 377 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4820#-NYrO4ej42sUJlziNuhC:-NYrO4ej42sUJlziNuhD:b-1fvakv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/40d616308aafdc97508bbc3ca28a96dbc0ad52a2/ios/MullvadVPN/TunnelManager/SetAccountOperation.swift#L377)):*
> ```Swift
> 
>     /// Find device by public key in the list of devices registered on server. The result passed to `completion` handler
>     /// may contain `nil` if such device is not found for some reason.
> ```

nit: Let's remove `for some reason`.
It doesn't add anything of value to the comment.

PR #4828, opened 2023-06-26
Do not consider removing a revoked device to be an error.

MarkusPettersson98:
Sometimes when removing a device when uninstalling the app, a `404` is returned if the device has already been revoked from the account. However, removing a non-existent device should not be considered an error if the end goal _is_ to not have the device on the account. That goal has clearly already been achieved.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4828)

### DIFF
--- a/mullvad-setup/src/main.rs
+++ b/mullvad-setup/src/main.rs
@@ -1,5 +1,5 @@
 use clap::Parser;
-use mullvad_api::{self, proxy::ApiConnectionMode};
+use mullvad_api::{self, proxy::ApiConnectionMode, DEVICE_NOT_FOUND};
 use mullvad_management_interface::MullvadProxyClient;
 use mullvad_types::version::ParsedAppVersion;
 use std::{path::PathBuf, process, str::FromStr, time::Duration};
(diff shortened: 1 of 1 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4828)* status: all files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`mullvad-setup/src/main.rs` line 189 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4828#-NYrXyY_6t-aOnHMAIlV:-NYrX_yr2mb69ynKxxLR:b-d5klxt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0af4858b9ef63522b546108d5443636a67187da0/mullvad-setup/src/main.rs#L189)):*
> ```XML
>         match device_removal {
>             Ok(_) => Ok(()),
>             Err(mullvad_api::rest::Error::ApiError(_status, code)) if code == DEVICE_NOT_FOUND => {
> ```

I wonder if it might also make sense to not consider this an error in `mullvad-management-interface` and the device manager.

E.g., when removing a device, the gRPC server should not return an error if the device has already been removed.

Also, `UpdateDevice` probably does not need to an error when the device has been removed.

PR #4837, opened 2023-06-27
Fix inability to use router/gateway as the custom WireGuard endpoint

dlon:
This is currently not possible because the `PF_ROUTE` socket demurs when a route is added if its destination IP equals its gateway.

The use case is pretty strange, but we do rely on it for testing. So let's quick fix it by ignoring such routes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4837)

### DIFF
--- a/talpid-routing/src/unix/macos/watch.rs
+++ b/talpid-routing/src/unix/macos/watch.rs
@@ -48,6 +48,16 @@ impl RoutingTable {
     }
 
     pub async fn add_route(&mut self, message: &RouteMessage) -> Result<()> {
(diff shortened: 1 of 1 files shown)


pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4837#-:-NYxzG1s2v8K502y5zpN:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4844, opened 2023-06-28
Add Linus' new gpg key

faern:
Adding my shiny new GPG pubkey to this repo. Allows me to sign stuff with the new key and have the CI trust me. And is a trusted source of app team key material.

The key file itself contains the public key (duh) and also signatures from my old code signing key + Mullvads code signing key. This should allow you to import it and trust it without manual verification of the fingerprint.

The commit adding the key file is signed with my old (current) GPG key.

The key file has a new naming scheme compared to before. Because we need a way to change developers keys without immediately removing the old key immediately. I can't guarantee I'll be able to sign every single commit with the new key right away. And even if I could, the old key is not untrusted (at least not yet). I feel like `$name-$keyid.pub` is a pretty good naming standard. That's the format I use for key files in my guide on how to generate new keys also.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4844)


raksooo (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4844)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4849, opened 2023-06-29
Add CI job for vendoring Rust dependencies

### LINKED ISSUE #4848: Cargo Vendor fails because of duplicating source of dependency package versions
## Issue report

Operating system: openSUSE Tumbleweed

App version: 2023.4

### Issue description

I am the packager of mullvadvpn on openSUSE Tumbleweed (https://build.opensuse.org/package/show/home:nuklly/mullvadvpn). The build environment has no Internet access, so I have to run `cargo vendor` command to package all the dependency packages. This works fine on 2023.3.

However, in 2023.4, when I run the command `cargo vendor`, it fails with the following message:

```
error: failed to sync

Caused by:
  found duplicate version of package `shadowsocks v1.15.3` vendored from two sources:

  	source 1: registry `crates-io`
  	source 2: https://github.com/mullvad/shadowsocks-rust?rev=c45980bb22d0d50ac888813c59a1edf0cff14a36#c45980bb
```

I am wondering if we can fix this issue from mullvad side?

faern:
Add a CI job for checking if `cargo vendor` works. Since it's part of cargo upstream there is no need to install anything except the Rust toolchain. Vendoring is nothing we use at Mullvad ourselves. But it can help third party package maintainers to build and package stuff.

My main motivation for adding this is the dependency tree cleanliness checks it implicitly imposes. Vendoring will fail if a dependency comes from multiple sources. Something we have had an issue with in for example version `2023.4` (see linked issue)

I think it's enough to run this on a single platform. It looks like `cargo vendor` on Linux also vendors macOS and Windows specific dependencies. So hopefully this very simple CI job should be enough.

Fixes DES-264
Fixes #4848

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4849)

### DIFF
--- a/.github/workflows/cargo-vendor.yml
+++ b/.github/workflows/cargo-vendor.yml
@@ -1,0 +1,28 @@
+---
+# The reason why we check for vendorability is not because we at Mullvad usually vendor
+# dependencies ourselves. But it can help some third party packagers of this project.
+# It also is a sanity check on our dependency tree. Vendoring will fail if a single
+# dependency has multiple sources: https://github.com/mullvad/mullvadvpn-app/issues/4848
+name: Rust - Vendor dependencies
+on:
+  pull_request:
+    paths:
+      - .github/workflows/cargo-vendor.yml
+      - Cargo.lock
+      - '**/Cargo.toml'
+  workflow_dispatch:
+jobs:
+  cargo-vendor:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4849)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4856, opened 2023-06-29
Enable Shadowsocks bridges selection via RelaySelector

buggmagnet:
This PR introduces the ability for `RelaySelector` to select a Shadowsocks bridge that is filtered using the User's prefered relay chosen via the UI.

Due to how we propagate `relayConstraints` in the application, as long as the users do not choose a relay, a Shadowsocks bridge close to Sweden will be chosen by default.

I tried to implement filtering by Haversine distance (as the desktop clients do) but I decided to do that in a future work instead as to keep the PR impact minimal for `RelaySelector` 

How to test ?

This will only affect Shadowsocks bridges, so it's not immediately obvious how to test this. 
I would recommend forcing Shadowsocks transports to be used by default (set the `connectionAttempts` to `1` in `TransportStrategy`) and then select a Relay.

Due to how we cache the current Shadowsocks connection and selected endpoint at the moment, this is more likely to have an impact when the user experiences poor connectivity (more chances to switch transports in that case).

To that effect, you could disable the Shadowsocks caching for testing purposes and break on the selected relay to check whether it's close to what was chosen in the UI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856)

### DIFF
--- a/ios/MullvadREST/ServerRelaysResponse.swift
+++ b/ios/MullvadREST/ServerRelaysResponse.swift
@@ -48,30 +48,6 @@ extension REST {
         public let ipv6AddrIn: IPv6Address
         public let publicKey: Data
         public let includeInCountry: Bool
-
-        public init(
-            hostname: String,
-            active: Bool,
-            owned: Bool,
-            location: String,
-            provider: String,
-            weight: UInt64,
-            ipv4AddrIn: IPv4Address,
-            ipv6AddrIn: IPv6Address,
-            publicKey: Data,
-            includeInCountry: Bool
-        ) {
-            self.hostname = hostname
-            self.active = active
-            self.owned = owned
-            self.location = location
-            self.provider = provider
-            self.weight = weight
-            self.ipv4AddrIn = ipv4AddrIn
-            self.ipv6AddrIn = ipv6AddrIn
-            self.publicKey = publicKey
-            self.includeInCountry = includeInCountry
-        }
     }
 
     public struct ServerWireguardTunnels: Codable, Equatable {
@@ -79,18 +55,6 @@ extension REST {
         public let ipv6Gateway: IPv6Address
         public let portRanges: [[UInt16]]
         public let relays: [ServerRelay]
-
-        public init(
-            ipv4Gateway: IPv4Address,
-            ipv6Gateway: IPv6Address,
-            portRanges: [[UInt16]],
-            relays: [REST.ServerRelay]
-        ) {
-            self.ipv4Gateway = ipv4Gateway
-            self.ipv6Gateway = ipv6Gateway
-            self.portRanges = portRanges
-            self.relays = relays
-        }
     }
 
     public struct ServerShadowsocks: Codable, Equatable {
@@ -98,38 +62,16 @@ extension REST {
         public let port: UInt16
         public let cipher: String
         public let password: String
-
-        public init(protocol: String, port: UInt16, cipher: String, password: String) {
-            self.protocol = `protocol`
-            self.port = port
-            self.cipher = cipher
-            self.password = password
-        }
     }
 
     public struct ServerBridges: Codable, Equatable {
         public let shadowsocks: [ServerShadowsocks]
         public let relays: [BridgeRelay]
-
-        public init(shadowsocks: [REST.ServerShadowsocks], relays: [BridgeRelay]) {
-            self.shadowsocks = shadowsocks
-            self.relays = relays
-        }
     }
 
     public struct ServerRelaysResponse: Codable, Equatable {
         public let locations: [String: ServerLocation]
         public let wireguard: ServerWireguardTunnels
         public let bridge: ServerBridges
-
-        public init(
-            locations: [String: REST.ServerLocation],
-            wireguard: REST.ServerWireguardTunnels,
-            bridge: ServerBridges
-        ) {
-            self.locations = locations
-            self.wireguard = wireguard
-            self.bridge = bridge
-        }
     }
 }
--- a/ios/MullvadTransport/TransportProvider.swift
+++ b/ios/MullvadTransport/TransportProvider.swift
@@ -23,19 +23,30 @@ public final class TransportProvider: RESTTransport {
 
     private var currentTransport: RESTTransport?
     private let parallelRequestsMutex = NSLock()
+    private var relayConstraints = RelayConstraints()
+    private let constraintsUpdater: RelayConstraintsUpdater
 
     public init(
         urlSessionTransport: URLSessionTransport,
         relayCache: RelayCache,
         addressCache: REST.AddressCache,
         shadowsocksCache: ShadowsocksConfigurationCache,
-        transportStrategy: TransportStrategy = .init()
+        transportStrategy: TransportStrategy = .init(),
+        constraintsUpdater: RelayConstraintsUpdater
     ) {
         self.urlSessionTransport = urlSessionTransport
         self.relayCache = relayCache
         self.addressCache = addressCache
         self.shadowsocksCache = shadowsocksCache
         self.transportStrategy = transportStrategy
+        self.constraintsUpdater = constraintsUpdater
+        constraintsUpdater.onNewConstraints = { [weak self] newConstraints in
+            self?.parallelRequestsMutex.lock()
+            defer {
+                self?.parallelRequestsMutex.unlock()
+            }
+            self?.relayConstraints = newConstraints
+        }
     }
 
     // MARK: -
@@ -75,10 +86,13 @@ public final class TransportProvider: RESTTransport {
     /// Returns a randomly selected shadowsocks configuration.
     private func makeNewShadowsocksConfiguration() throws -> ShadowsocksConfiguration {
         let cachedRelays = try relayCache.read()
-        let bridgeAddress = RelaySelector.getShadowsocksRelay(relays: cachedRelays.relays)?.ipv4AddrIn
-        let bridgeConfiguration = RelaySelector.getShadowsocksTCPBridge(relays: cachedRelays.relays)
+        let bridgeConfiguration = RelaySelector.shadowsocksTCPBridge(from: cachedRelays.relays)
+        let closestRelay = RelaySelector.closestShadowsocksRelayConstrained(
+            by: relayConstraints,
+            in: cachedRelays.relays
+        )
 
-        guard let bridgeAddress, let bridgeConfiguration else { throw POSIXError(.ENOENT) }
+        guard let bridgeAddress = closestRelay?.ipv4AddrIn, let bridgeConfiguration else { throw POSIXError(.ENOENT) }
 
         let newConfiguration = ShadowsocksConfiguration(
             bridgeAddress: bridgeAddress,
--- a/ios/MullvadTypes/RelayConstraints.swift
+++ b/ios/MullvadTypes/RelayConstraints.swift
@@ -8,6 +8,18 @@
 
 import Foundation
 
+public protocol ConstraintsPropagation {
+    var onNewConstraints: ((RelayConstraints) -> Void)? { get set }
+}
+
+public class RelayConstraintsUpdater: ConstraintsPropagation {
+    public var onNewConstraints: ((RelayConstraints) -> Void)?
+
+    public init(onNewConstraints: ((RelayConstraints) -> Void)? = nil) {
+        self.onNewConstraints = onNewConstraints
+    }
+}
+
 public struct RelayConstraints: Codable, Equatable, CustomDebugStringConvertible {
     public var location: RelayConstraint<RelayLocation>
 
--- a/ios/MullvadVPN/AppDelegate.swift
+++ b/ios/MullvadVPN/AppDelegate.swift
@@ -40,6 +40,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     private(set) var relayCacheTracker: RelayCacheTracker!
     private(set) var storePaymentManager: StorePaymentManager!
     private var transportMonitor: TransportMonitor!
+    private var relayConstraintsObserver: TunnelBlockObserver!
 
     // MARK: - Application lifecycle
 
@@ -85,6 +86,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             apiProxy: apiProxy
         )
 
+        let constraintsUpdater = RelayConstraintsUpdater()
+        relayConstraintsObserver = TunnelBlockObserver(didUpdateTunnelSettings: { _, settings in
+            constraintsUpdater.onNewConstraints?(settings.relayConstraints)
+        })
+
         storePaymentManager = StorePaymentManager(
             application: application,
             queue: .default(),
@@ -98,8 +104,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             urlSessionTransport: urlSessionTransport,
             relayCache: relayCache,
             addressCache: addressCache,
-            shadowsocksCache: shadowsocksCache
+            shadowsocksCache: shadowsocksCache,
+            constraintsUpdater: constraintsUpdater
         )
+
+        tunnelManager.addObserver(relayConstraintsObserver)
 
         transportMonitor = TransportMonitor(
             tunnelManager: tunnelManager,
--- a/ios/MullvadVPNTests/RelayCacheTests.swift
+++ b/ios/MullvadVPNTests/RelayCacheTests.swift
@@ -6,7 +6,7 @@
 //  Copyright © 2023 Mullvad VPN AB. All rights reserved.
 //
 
-import MullvadREST
+@testable import MullvadREST
 import MullvadTransport
 @testable import RelayCache
 import XCTest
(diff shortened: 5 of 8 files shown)


pronebird (changes requested):
> Reviewed 5 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856)* status: 5 of 8 files reviewed, 6 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/ServerRelaysResponse.swift` line 51 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZ9qxcvBzLxyE3Hz5UC:-NZ9qxcvBzLxyE3Hz5UD:bpdugs0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/MullvadREST/ServerRelaysResponse.swift#L51)):*
> ```Swift
>         public let weight: UInt64
>         public let includeInCountry: Bool
>         // Unused, therefore not included in `CodingKeys`
> ```

 Please avoid using `Codbale` this way and adding unused fields into concrete types. 

Use generics and protocols instead if you need to filter by common properties of a group of structs while preserving concrete types, i.e:

```
protocol CommonType {
    var weight: UInt64 { get set }
}

struct TypeA: CommonType, Codable {
    var weight: UInt64
}

struct TypeB: CommonType, Codable {
    var weight: UInt64
}


func filter<T>(_ values: [T]) -> [T] where T: CommonType {
    return values.filter { s in
        return s.weight < 2
    }
}

let values: [TypeA] = [TypeA(weight: 1), TypeA(weight: 2)]
let filtered = filter(values)

```
___
*[`ios/MullvadTransport/TransportProvider.swift` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZ9pZcA4eFF2qjW23LR:-NZ9pZcA4eFF2qjW23LS:b4kxefu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/MullvadTransport/TransportProvider.swift#L26)):*
> ```Swift
>     private let parallelRequestsMutex = NSLock()
>     private var relayConstraints = RelayConstraints(location: .any)
> ```

This will lead to transport using `any` constraint until user changes location. This will happen on each run from what I understand. Default constraint we have is bound to Sweden. 

This code would not suffer this problem if you used a block to provide relay constraints `() -> RelayConstraints`.
___
*[`ios/MullvadTypes/RelayConstraints.swift` line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZ9oJgcEI5m6F2QJahT:-NZ9oJgcEI5m6F2QJahU:b-b68zxk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/MullvadTypes/RelayConstraints.swift#L15)):*
> ```Swift
> }
> 
> public class RelayConstraintsUpdater: ConstraintsPropagation {
> ```

This abstraction still looks foreign to me and should be replaced with `() -> RelayConstraints`. Keep it simple.
___
*[`ios/MullvadVPN/AppDelegate.swift` line 108 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZ9n0I_BjHVHHvnJV-S:-NZ9n0I_BjHVHHvnJV-T:b-3oir9l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/MullvadVPN/AppDelegate.swift#L108)):*
> ```Swift
>             addressCache: addressCache,
>             shadowsocksCache: shadowsocksCache,
>             constraintsUpdater: constraintsUpdater
> ```

Should be replaced with `() -> RelayConstraints`, i.e: 

```swift
constraintsProvider: { tunnelManager.settings.relayConstraints }
```
___
*[`ios/PacketTunnel/PacketTunnelProvider.swift` line 153 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZ9navB-6NQbOsD9SvO:-NZ9navB-6NQbOsD9SvP:b-kmbt7l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/PacketTunnel/PacketTunnelProvider.swift#L153)):*
> ```Swift
>             addressCache: addressCache,
>             shadowsocksCache: shadowsocksCache,
>             constraintsUpdater: constraintsUpdater
> ```

Could be replaced with or something along these lines:

```
constraintsProvider: {
  self.dispatchQueue.sync {
    // read and store it in makeConfiguration
    return self.relayConstraints
  }
}
```
___
*[`ios/RelaySelector/RelaySelector.swift` line 82 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZ9lo1S8fWtOlAx2m0X:-NZ9lo1S8fWtOlAx2m0Y:blqsnm0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/RelaySelector/RelaySelector.swift#L82)):*
> ```Swift
>         return RelaySelectorResult(
>             endpoint: endpoint,
>             relay: relayWithLocation.relay as! REST.ServerRelay,
> ```

Force unwrap does not look right. Please use generics and/or protocols to negotiate the actual type with compiler.

pronebird (commented):
> Reviewed 3 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet)

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856)* status: all files reviewed, 7 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/RelaySelector/RelaySelector.swift` line 210 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZAX-tv5wdIhZYhbQ7r:-NZAX-tv5wdIhZYhbQ7s:bw4u2lj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/RelaySelector/RelaySelector.swift#L210)):*
> ```Swift
>     }
> 
>     private static func mapRelayBridgesIn(_ response: REST.ServerRelaysResponse) -> [RelayWithLocation] {
> ```

You could merge two of those mapXXX functions into one, for example:


```swift
protocol AnyRelayProtocol {
    var hostname: String { get }
    var location: String { get }
    var weight: UInt64 { get }
    var active: Bool { get }
    var includeInCountry: Bool { get }
}

extension REST.ServerRelay: AnyRelayProtocol {}
extension REST.BridgeRelay: AnyRelayProtocol {}

private struct RelayWithLocation<T: AnyRelayProtocol> {
    let relay: T
    let location: Location
}
```

Then mapping each individual representation of either bridge or wg relay can be done using generic function:

```swift
private static func mapAnyRelay<T: AnyRelayProtocol>(
    relays: [T],
    locations: [String: REST.ServerLocation]
) -> [RelayWithLocation<T>] {
    relays.compactMap { relay in
        guard let serverLocation = locations[relay.location] else { return nil }
        return makeRelayWithLocationFrom(serverLocation, relay: relay)
    }
}

/* Can be called with:

let serverResponse: REST.ServerRelaysResponse = ...

// Bridges
mapAnyRelay(relays: serverResponse.bridge.relays, locations: serverResponse.locations)

// WG relays
mapAnyRelay(relays: serverResponse.wireguard.relays, locations: serverResponse.locations)

*/
```

Following the same example the rest of functions can be generic too and used interchangeably, i.e:

```swift
static func pickRandomRelay<T: AnyRelayProtocol>(relays: [RelayWithLocation<T>]) -> RelayWithLocation<T>?

static func makeRelayWithLocationFrom<T: AnyRelayProtocol>(
    _ serverLocation: REST.ServerLocation,
    relay: T
) -> RelayWithLocation<T>?

static func applyConstraints<T: AnyRelayProtocol>(
    _ constraints: RelayConstraints,
    relays: [RelayWithLocation<T>]
) -> [RelayWithLocation<T>]
```

For the most part only signature change is needed to make it work. This is the beauty of protocols/generics.

pronebird (dismissed):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856)* status: all files reviewed, 7 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/RelaySelector/RelaySelector.swift` line 210 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZAX-tv5wdIhZYhbQ7r:-NZAZNa-0SelmmfBcEIf:bhd8agc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/RelaySelector/RelaySelector.swift#L210)):*

You could merge two of those mapXXX functions into one, for example:


```swift
protocol AnyRelayProtocol {
    var hostname: String { get }
    var location: String { get }
    var weight: UInt64 { get }
    var active: Bool { get }
    var includeInCountry: Bool { get }
}

extension REST.ServerRelay: AnyRelayProtocol {}
extension REST.BridgeRelay: AnyRelayProtocol {}

private struct RelayWithLocation<T: AnyRelayProtocol> {
    let relay: T
    let location: Location
}
```

Then mapping each individual representation of either bridge or wg relay can be done using generic function:

```swift
private static func mapAnyRelay<T: AnyRelayProtocol>(
    relays: [T],
    locations: [String: REST.ServerLocation]
) -> [RelayWithLocation<T>] {
    relays.compactMap { relay in
        guard let serverLocation = locations[relay.location] else { return nil }
        return makeRelayWithLocationFrom(serverLocation, relay: relay)
    }
}

/* Can be called with:

let serverResponse: REST.ServerRelaysResponse = ...

// Bridges
mapAnyRelay(relays: serverResponse.bridge.relays, locations: serverResponse.locations)

// WG relays
mapAnyRelay(relays: serverResponse.wireguard.relays, locations: serverResponse.locations)

*/
```

Following the same example the rest of functions can be generic too and used interchangeably, i.e:

```swift
static func pickRandomRelay<T: AnyRelayProtocol>(relays: [RelayWithLocation<T>]) -> RelayWithLocation<T>?

static func makeRelayWithLocationFrom<T: AnyRelayProtocol>(
    _ serverLocation: REST.ServerLocation,
    relay: T
) -> RelayWithLocation<T>?

static func applyConstraints<T: AnyRelayProtocol>(
    _ constraints: RelayConstraints,
    relays: [RelayWithLocation<T>]
) -> [RelayWithLocation<T>]
```

For the most part only signature change is needed to make it work. This is the beauty of protocols/generics.
</blockquote></details>

Commit with the above changes: https://github.com/mullvad/mullvadvpn-app/commit/f44c79cbd8b4641cbbdd33a4c2e1e2755f3093df

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856)* status: all files reviewed, 7 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZ9pZcA4eFF2qjW23LR:-NZPbl853PELee_vp-02:beyacys) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/MullvadTransport/TransportProvider.swift#L26)):*
I will reply to this comment as most of your other comments are saying the same thing (don't use a protocol, use a closure instead).

> This will lead to transport using `any` constraint until user changes location.

When I wrote this, I hadn't realized we defaulted to Sweden. I guess that's fair, but I will argue that this default is not necessarily the best choice for our users, as it penalizes our users that are not based in Europe. But that's beyond the point  of this so I will change that.

> This code would not suffer this problem if you used a block to provide relay constraints `() -> RelayConstraints`.

That is not correct. `TunnelManager` has the following default
`private var _tunnelSettings = TunnelSettingsV2()` 

It's only updated when we call `scheduleSettingsUpdate` which itself is called when `setRelayConstraints` is called.

So the outcome would be exactly the same.
And lastly, to answer to another comment

> This abstraction still looks foreign to me and should be replaced with `() -> RelayConstraints`. Keep it simple.

It's not because it looks foreign to you that it's not a good choice, or "not simple".
Using closures to provide arguments is generally a design smell. Here's a very crude example that illustrates why.

```swift
func myApi(_ param1: () -> String, param2: () -> Int, param3: () -> Bool) { }

```

Now the callsite looks like this
```swift

myApi {
    <#code#>
} param2: {
    <#code#>
} param3: {
    <#code#>
}

```

There are many instances where a closure as a parameter makes sense (usually in the form of a callback to do async processing). However,`TransportProvider` is not one of those.
___
*[`ios/RelaySelector/RelaySelector.swift` line 210 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZAX-tv5wdIhZYhbQ7r:-NZPbWbN07D37Wvp6R-7:bcpn9ev) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/RelaySelector/RelaySelector.swift#L210)):*

Commit with the above changes: https://github.com/mullvad/mullvadvpn-app/commit/f44c79cbd8b4641cbbdd33a4c2e1e2755f3093df
</blockquote></details>

Thanks for the suggestion, this is ultimately what I wanted to do, but you expressed it in a nicer way.
I'll cherry pick your commit for that fix.

rablador (commented):
> Reviewed 8 of 8 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856)* status: all files reviewed, 7 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZ9pZcA4eFF2qjW23LR:-NZQ72LM755G8W45TQUv:bgo6yuc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/MullvadTransport/TransportProvider.swift#L26)):*
> There are many instances where a closure as a parameter makes sense (usually in the form of a callback to do async processing). [...]

I (generally) agree with this. And while I also think that the intent is better expressed with the explicit `RelayConstraintsUpdater`, it IS nice to get rid of the default constraints. It might not matter right now, if TunnelSettingsV2 is initialised with the default values anyway, but it's not exactly futureproof.

buggmagnet (commented):
> Dismissed @pronebird from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856)* status: 7 of 8 files reviewed, 4 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/MullvadTransport/TransportProvider.swift` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZ9pZcA4eFF2qjW23LR:-NZ_yyNcBxjLU9vcMMaf:b-tph529) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/MullvadTransport/TransportProvider.swift#L26)):*

> There are many instances where a closure as a parameter makes sense (usually in the form of a callback to do async processing). [...]

I (generally) agree with this. And while I also think that the intent is better expressed with the explicit `RelayConstraintsUpdater`, it IS nice to get rid of the default constraints. It might not matter right now, if TunnelSettingsV2 is initialised with the default values anyway, but it's not exactly futureproof.
</blockquote></details>

Yes, we need a better design for the future, but we will have to do with that for now.
___
*[`ios/RelaySelector/RelaySelector.swift` line 82 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-NZ9lo1S8fWtOlAx2m0X:-NZa--LP4c6LPUSe_kH_:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/073ff6afae15ed838931999a95c7db7688e88e7d/ios/RelaySelector/RelaySelector.swift#L82)):*

Force unwrap does not look right. Please use generics and/or protocols to negotiate the actual type with compiler.
</blockquote></details>

Done.

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856)* status: all files reviewed, 4 unresolved discussions (waiting on @pronebird)

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856)* status: all files reviewed, 4 unresolved discussions (waiting on @pronebird)

mojganii (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856#-:-NZf3QBg4tsXYg-9VlqR:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @pronebird)

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4856)* status: all files reviewed, 4 unresolved discussions (waiting on @pronebird)

PR #4866, opened 2023-07-03
Upgrade `x25519-dalek` to `2.0.0-rc.3`

faern:
I just saw that this dependency had some updates. Nothing we really need. But it's nicer to be on a release candidate compared to a `pre` release. We can also rely on the library to pick a suitable CSPRNG instead of having us do that.

The zeroize part is not related to the upgrade. I just felt it was nice to add. It's good practice to clear private keys from memory as much as possible.

This upgrade removes the clamping of the private key material used for WireGuard. See [`x25519-dalek` changelog](https://github.com/dalek-cryptography/x25519-dalek/blob/main/CHANGELOG.md#200-rc3). We don't believe this should hurt in any way:
* Public keys generated with the new code will still be clamped in the pubkey generation code
* Old private keys read from `device.json` will still work, they are clamped and the new code can still take in clamped arrays
* New keys generated by the upgraded code will not be clamped. So the base64 string in `device.json` will not be clamped. But this should hopefully not hurt in any way. Users should not access this key manually anyway
* We have checked that the WireGuard in the Linux kernel, WireGuardNT and wireguard-go all clamp the private key when it receives it. So we don't need to clamp it for them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4866)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -603,9 +603,9 @@ checksum = "e496a50fda8aacccc86d7529e2c1e0892dbd0f898a6b5645b5561b89c3210efa"
 
 [[package]]
 name = "cpufeatures"
-version = "0.2.5"
+version = "0.2.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "28d997bd5e24a5928dd43e46dc529867e207907fe0b239c3477d924f7f2ca320"
+checksum = "03e69e28e9f7f77debdedbaafa2866e1de9ba56df55a8bd7cfc724c25a09987c"
 dependencies = [
  "libc",
 ]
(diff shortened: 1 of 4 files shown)


pinkisemils (approved):
:lgtm:

> Reviewed 3 of 3 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4866#-:-NZR7kbB7rsvS5QTmgh-:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4872, opened 2023-07-05
Move tests into test plans

rablador:
The MullvadVPNScreenshot test target only works in the CI environment, or with an install of MullvadVPN that has never been interacted with. 

It's also configured to always run, which takes a lot of time (only to fail), which creates a frustrating development experience.

We should switch to using [test plans](https://developer.apple.com/documentation/xcode/organizing-tests-to-improve-feedback) instead so that we can have a test plan that does not run UI automated tests by default when we are doing daily work on our local machines.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4872)

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -6,6 +6,7 @@ on:
       - .github/workflows/ios.yml
       - ios/.swiftformat
       - ios/**/*.swift
+      - ios/**/*.xctestplan
   workflow_dispatch:
 jobs:
   check-formatting:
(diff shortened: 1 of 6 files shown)


buggmagnet (approved):
:lgtm:

> Reviewed 5 of 5 files at r1, 1 of 1 files at r2, 1 of 2 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4872#-:-N_sVb8p5WLEH8vocgD7:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4876, opened 2023-07-06
Upgrade duct and os_pipe

faern:
Upgrades `duct` to latest version. This makes it use `os_pipe ^1` and `shared_child ^1` instead of `0.x` versions.

Then upgrade `os_pipe` as direct dependency also, to end up with the same version as duct (latest). Keeps our dependency tree tidier and brings us up to the latest versions.

Also found we had one unused instance of `os_pipe`

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4876)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -907,13 +907,13 @@ dependencies = [
 
 [[package]]
 name = "duct"
-version = "0.13.5"
+version = "0.13.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0fc6a0a59ed0888e0041cf708e66357b7ae1a82f1c67247e1f93b5e0818f7d8d"
+checksum = "37ae3fc31835f74c2a7ceda3aeede378b0ae2e74c8f1c36559fcc9ae2a4e7d3e"
 dependencies = [
  "libc",
  "once_cell",
- "os_pipe",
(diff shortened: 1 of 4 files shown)


faern (commented):
Yeah maybe. But ultimately we should just use `tokio::process::Command` instead. We use the OpenVPN process handle asynchronously anyway. We already are very very nasty by spawning the process in a blocking way in an async function I see now: https://github.com/mullvad/mullvadvpn-app/blob/b65d04b9c4704caad7c068e93b27032e1880382b/talpid-openvpn/src/lib.rs#L469

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4876#-:-NZjHbor4QMAvELvE8kI:b-lazhjo)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (approved):
Probably. We've just shoved async parts into it, but the whole of `openvpn` could use refactoring. Not too important right now, though.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4876#-:-NZjzFuk7hS1fG1GFl5C:bsdvoyi)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4886, opened 2023-07-10
Use full CLI path in GUI tests

dlon:
The test runner starts before `PATH` has been updated, so let's not rely on `PATH`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4886)


dlon:
Let's make the user set the correct environment instead.

PR #4897, opened 2023-07-13
Add filtering by Haversine distance

buggmagnet:
This PR implements filtering Shadowsocks bridges by using Haversine distance.

It works almost like the Rust version of it, except for the fact that the Rust version has a `filter_map` function to either take the first 5 relays it find or all the relays with a haversine distance < 1500km.

The iOS version takes all the relays that satisfy the distance < 1500km and then takes the first 5 out of that list
because it doesn't have a native `filter_map` option and that I didn't want to introduce extra complexity.

This matters most when the user already has selected a Relay in the GUI which will create the appropriate `RelayConstraint` that lets us filter which Shadowsocks relay to use for getting a relay closer to what the user wished for.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4897)


pinkisemils (changes requested):
> Reviewed 4 of 7 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4897)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/RelaySelector/RelaySelector.swift` line 68 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4897#-N_IRf9N3IFNg25X_TbU:-N_IRf9N3IFNg25X_TbV:bxe2v3k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ec8ad0ccff2ac5245f0a05c4de017f4def9a2970/ios/RelaySelector/RelaySelector.swift#L68)):*
> ```Swift
>         }.filter {
>             $0.distance <= maximumBridgeDistance
>         }.prefix(5)
> ```

If all bridges are further away than the maximum distance, won't this imply that no bridge will be selected?

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4897)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`ios/RelaySelector/RelaySelector.swift` line 68 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4897#-N_IRf9N3IFNg25X_TbU:-N_rUTmA5K3Wrt8GwKBu:bndtjcx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ec8ad0ccff2ac5245f0a05c4de017f4def9a2970/ios/RelaySelector/RelaySelector.swift#L68)):*

If all bridges are further away than the maximum distance, won't this imply that no bridge will be selected?
</blockquote></details>

Almost, at L81 we check that if there was no randomly selected relay from the weight function (implying anything in L53-L80 failed), we source a random relay from the `filteredRelays` which are guaranteed to contain at least 1 element at this point

pinkisemils (approved):
> Reviewed 1 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4897)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4901, opened 2023-07-14
Set package fields in workspace and inherit in packages

faern:
When researching denying `rustc` lints in a config file instead of repeating `#![deny(rust_2018_idioms)]` in every `lib.rs` and `main.rs` (that is a nightly feature and probably nothing we can use right now) I stumbled upon the `[workspace.package]` feature of Cargo. I had completely missed this feature! Apparently we can define a lot of common fields directly in the top level `Cargo.toml` and have all packages underneath inherit the value.

With this feature we run a slightly lower risk of forgetting `publish = false` in new crates. We can also bump the Rust edition in a single place instead of a million places.

It's a bit verbose to have to declare `<field>.workspace = true` for every field in the packages. I wished it automatically inherited everything not explicitly overridden in the fields. But it is what it is. And since Rust 1.71.0 `cargo new` automatically inserts these inherit fields when creating new packages. So makes it harder to forget setting some of them.

This is a draft PR because I did not finish migrating all packages to inherit fields. I wanted to first publish this and get some feedback on the feature overall.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4901)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4227,7 +4227,7 @@ dependencies = [
 
 [[package]]
 name = "translations-converter"
-version = "0.1.0"
+version = "0.0.0"
 dependencies = [
  "derive_more",
  "htmlize",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,11 @@
+[workspace.package]
+version = "0.0.0"
+authors = ["Mullvad VPN"]
+repository = "https://github.com/mullvad/mullvadvpn-app/"
+license = "GPL-3.0"
+edition = "2021"
+publish = false
+
 [workspace]
 resolver = "2"
 members = [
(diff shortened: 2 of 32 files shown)


faern:
One potentially really nice thing we can do with a related feature: `[workspace.dependencies]`. Declare dependencies on the workspace level. We have a bunch of dependencies we depend on in a large set of packages. But we need to declare the same version over and over again. Making upgrades tedious and also running the risk of depending on multiple versions. With `[workspace.dependencies]` we can probably declare for example our `tonic` dependency only once.

dlon (approved):
> Reviewed 12 of 12 files at r1, 20 of 20 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4901)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4902, opened 2023-07-14
Use workspace dependencies for Rust

faern:
Trying out the `[workspace.dependencies]` functionality of Cargo. It's pretty neat to be able to declare dependencies at the workspace level. Then we only have a single place to bump the version when we upgrade dependencies.

You could argue that we are almost never going to bump tokio. And when we do, it will be such a large re-write anyway that this "optimization" is not important. And that's correct. But there are plenty of other crates that we do upgrade fairly regularly that might benefit more from this. So maybe starting with `tokio` was wrong. For `tonic` it makes more sense. Because we will need to upgrade it (there is already `0.9` out), and there are a bunch of related crates that goes with it. So there is one place to bump all the five crates.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4902)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,12 @@ members = [
 
 [workspace.dependencies]
 tokio = { version = "1.8" }
+# Tonic and related crates
+tonic = "0.8"
+tonic-build = { version = "0.8", default-features = false }
+tower = "0.4"
+prost = "0.11"
+prost-types = "0.11"
 
 [profile.release]
 opt-level = 3
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,9 @@ members = [
     "tunnel-obfuscation",
 ]
 
+[workspace.dependencies]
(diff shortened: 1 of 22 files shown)


dlon (approved):
:lgtm:

> Reviewed 22 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4902#-:-N_JNdI26V89_VHfelyO:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

faern:
So when should one use workspace dependencies? Surely not all dependencies should go there! I guess different organizations can use this a bit differently. But my thought is something along the lines of:
* Use workspace dependencies for big/important dependencies where you don't want different packages in the workspace to (accidentally?) use different versions.
* Still specify `features` in the individual packages that need that feature

PR #4914, opened 2023-07-20
Implement account deletion

mojganii:
This PR aims to enable users to delete their accounts, giving them more control over the data they have shared with the service. To achieve this, several steps have been implemented:

1. Addition of an `Account Delete` button on the `Account` view to facilitate the account deletion process.

2. Implementation of a warning message to inform the user about the consequences of deleting their account. This step ensures users are fully aware of the actions they are taking.

3. User verification is required to delete the account. To prevent accidental deletions, users are asked to input the four last digits of their account number. This ensures that the user is making a deliberate decision and minimizes the chances of accidental deletions.

4. Service revocation is performed to notify the backend about the user's decision to delete their account.

5. After the account has been successfully removed, the following actions are taken:

   - Unset VPN configuration to ensure the user's data is no longer accessible through any VPN connection.

   - Cancel all ongoing operations associated with the user's account.

   - Remove the user's account number from the system to protect their privacy.

   - Redirect the user to a `logged out` state, indicating that the account deletion process was successful.

6. In case of any failure during the process, appropriate error messages are displayed to the user to inform them of the issue.

Overall, this PR enhances the user experience by providing a secure and deliberate process for account deletion, safeguarding user data and privacy.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)

### DIFF
--- a/ios/CHANGELOG.md
+++ b/ios/CHANGELOG.md
@@ -24,6 +24,7 @@ Line wrap the file at 100 chars.                                              Th
 ## [Unreleased]
 ### Added
 - Allow redeeming vouchers in account view.
+- Allow deleting account in account view.
 
 ## [2023.3 - 2023-07-15]
 ### Added
--- a/ios/MullvadREST/RESTAccountsProxy.swift
+++ b/ios/MullvadREST/RESTAccountsProxy.swift
@@ -82,6 +82,48 @@ extension REST {
                 completionHandler: completion
             )
         }
+
+        public func deleteAccount(
+            accountNumber: String,
+            retryStrategy: RetryStrategy,
+            completion: @escaping CompletionHandler<Void>
+        ) -> Cancellable {
+            let requestHandler = AnyRequestHandler(createURLRequest: { endpoint, authorization in
+                var requestBuilder = try self.requestFactory.createRequestBuilder(
+                    endpoint: endpoint,
+                    method: .delete,
+                    pathTemplate: "accounts/me"
+                )
+                requestBuilder.setAuthorization(authorization)
+                requestBuilder.addValue(accountNumber, forHTTPHeaderField: "Mullvad-Account-Number")
+
+                return requestBuilder.getRequest()
+            }, authorizationProvider: createAuthorizationProvider(accountNumber: accountNumber))
+
+            let responseHandler = AnyResponseHandler { response, data -> ResponseHandlerResult<Void> in
+                let statusCode = HTTPStatus(rawValue: response.statusCode)
+
+                switch statusCode {
+                case let statusCode where statusCode.isSuccess:
+                    return .success(())
+                default:
+                    return .unhandledResponse(
+                        try? self.responseDecoder.decode(
+                            ServerErrorResponse.self,
+                            from: data
+                        )
+                    )
+                }
+            }
+
+            return addOperation(
+                name: "delete-my-account",
+                retryStrategy: retryStrategy,
+                requestHandler: requestHandler,
+                responseHandler: responseHandler,
+                completionHandler: completion
+            )
+        }
     }
 
     public struct NewAccountData: Decodable {
--- a/ios/MullvadREST/RESTRequestFactory.swift
+++ b/ios/MullvadREST/RESTRequestFactory.swift
@@ -119,6 +119,13 @@ extension REST {
             )
         }
 
+        mutating func addValue(_ value: String, forHTTPHeaderField: String) {
+            restRequest.urlRequest.addValue(
+                value,
+                forHTTPHeaderField: forHTTPHeaderField
+            )
+        }
+
         func getRequest() -> REST.Request {
             restRequest
         }
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -438,12 +438,18 @@
 		F07C0A072A52DA64009825CA /* SetupAccountCompletedCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = F07C0A062A52DA64009825CA /* SetupAccountCompletedCoordinator.swift */; };
 		F07CFF2029F2720E008C0343 /* RegisteredDeviceInAppNotificationProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = F07CFF1F29F2720E008C0343 /* RegisteredDeviceInAppNotificationProvider.swift */; };
 		F0C2AEFD2A0BB5CC00986207 /* NotificationProviderIdentifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0C2AEFC2A0BB5CC00986207 /* NotificationProviderIdentifier.swift */; };
+		F0C6FA812A66E23300F521F0 /* DeleteAccountOperation.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0C6FA802A66E23300F521F0 /* DeleteAccountOperation.swift */; };
 		F0E3618B2A4ADD2F00AEEF2B /* WelcomeContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0E3618A2A4ADD2F00AEEF2B /* WelcomeContentView.swift */; };
 		F0E8CC032A4C753B007ED3B4 /* WelcomeViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0E8CC022A4C753B007ED3B4 /* WelcomeViewController.swift */; };
 		F0E8CC052A4CC88F007ED3B4 /* WelcomeCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0E8CC042A4CC88F007ED3B4 /* WelcomeCoordinator.swift */; };
 		F0E8CC0A2A4EE127007ED3B4 /* SetupAccountCompletedContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0E8CC092A4EE127007ED3B4 /* SetupAccountCompletedContentView.swift */; };
 		F0E8CC0C2A4EE672007ED3B4 /* SetupAccountCompletedViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0E8CC0B2A4EE672007ED3B4 /* SetupAccountCompletedViewController.swift */; };
 		F0E8E4BB2A56C9F100ED26A3 /* WelcomeInteractor.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0E8E4BA2A56C9F100ED26A3 /* WelcomeInteractor.swift */; };
+		F0E8E4C12A602CCB00ED26A3 /* AccountDeletionContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0E8E4C02A602CCB00ED26A3 /* AccountDeletionContentView.swift */; };
+		F0E8E4C32A602E0D00ED26A3 /* AccountDeletionViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0E8E4C22A602E0D00ED26A3 /* AccountDeletionViewModel.swift */; };
+		F0E8E4C52A60499100ED26A3 /* AccountDeletionViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0E8E4C42A60499100ED26A3 /* AccountDeletionViewController.swift */; };
+		F0E8E4C72A604CBE00ED26A3 /* AccountDeletionCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0E8E4C62A604CBE00ED26A3 /* AccountDeletionCoordinator.swift */; };
+		F0E8E4C92A604E7400ED26A3 /* AccountDeletionInteractor.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0E8E4C82A604E7400ED26A3 /* AccountDeletionInteractor.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -1213,12 +1219,18 @@
 		F07C0A062A52DA64009825CA /* SetupAccountCompletedCoordinator.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SetupAccountCompletedCoordinator.swift; sourceTree = "<group>"; };
 		F07CFF1F29F2720E008C0343 /* RegisteredDeviceInAppNotificationProvider.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RegisteredDeviceInAppNotificationProvider.swift; sourceTree = "<group>"; };
 		F0C2AEFC2A0BB5CC00986207 /* NotificationProviderIdentifier.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NotificationProviderIdentifier.swift; sourceTree = "<group>"; };
+		F0C6FA802A66E23300F521F0 /* DeleteAccountOperation.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DeleteAccountOperation.swift; sourceTree = "<group>"; };
 		F0E3618A2A4ADD2F00AEEF2B /* WelcomeContentView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = WelcomeContentView.swift; sourceTree = "<group>"; };
 		F0E8CC022A4C753B007ED3B4 /* WelcomeViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = WelcomeViewController.swift; sourceTree = "<group>"; };
 		F0E8CC042A4CC88F007ED3B4 /* WelcomeCoordinator.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = WelcomeCoordinator.swift; sourceTree = "<group>"; };
 		F0E8CC092A4EE127007ED3B4 /* SetupAccountCompletedContentView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SetupAccountCompletedContentView.swift; sourceTree = "<group>"; };
 		F0E8CC0B2A4EE672007ED3B4 /* SetupAccountCompletedViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SetupAccountCompletedViewController.swift; sourceTree = "<group>"; };
 		F0E8E4BA2A56C9F100ED26A3 /* WelcomeInteractor.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = WelcomeInteractor.swift; sourceTree = "<group>"; };
+		F0E8E4C02A602CCB00ED26A3 /* AccountDeletionContentView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountDeletionContentView.swift; sourceTree = "<group>"; };
+		F0E8E4C22A602E0D00ED26A3 /* AccountDeletionViewModel.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountDeletionViewModel.swift; sourceTree = "<group>"; };
+		F0E8E4C42A60499100ED26A3 /* AccountDeletionViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountDeletionViewController.swift; sourceTree = "<group>"; };
+		F0E8E4C62A604CBE00ED26A3 /* AccountDeletionCoordinator.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountDeletionCoordinator.swift; sourceTree = "<group>"; };
+		F0E8E4C82A604E7400ED26A3 /* AccountDeletionInteractor.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountDeletionInteractor.swift; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
 /* Begin PBXFrameworksBuildPhase section */
@@ -1501,6 +1513,7 @@
 		5823FA5726CE4A4100283BF8 /* TunnelManager */ = {
 			isa = PBXGroup;
 			children = (
+				F0C6FA802A66E23300F521F0 /* DeleteAccountOperation.swift */,
 				588527B1276B3F0700BAA373 /* LoadTunnelConfigurationOperation.swift */,
 				58F2E147276A307400A79513 /* MapConnectionStatusOperation.swift */,
 				F07BF2612A26279100042943 /* RedeemVoucherOperation.swift */,
@@ -1547,9 +1560,10 @@
 		583FE01629C196E8006E85F9 /* View controllers */ = {
 			isa = PBXGroup;
 			children = (
-				F0E8E4B92A55593300ED26A3 /* CreationAccount */,
 				583FE02029C1A0B1006E85F9 /* Account */,
+				F0E8E4BF2A602C7D00ED26A3 /* AccountDeletion */,
 				5878F4FA29CDA2D4003D4BE2 /* ChangeLog */,
+				F0E8E4B92A55593300ED26A3 /* CreationAccount */,
 				583FE01D29C197C1006E85F9 /* DeviceList */,
 				583FE02529C1AD0E006E85F9 /* Launch */,
 				583FE02129C1A0F4006E85F9 /* Login */,
@@ -1931,11 +1945,11 @@
 			isa = PBXGroup;
 			children = (
 				7AF0419D29E957EB00D492DD /* AccountCoordinator.swift */,
+				F0E8E4C62A604CBE00ED26A3 /* AccountDeletionCoordinator.swift */,
 				F07C0A042A52D4C3009825CA /* AccountRedeemingVoucherCoordinator.swift */,
 				58BBB39629717E0C00C8DB7C /* ApplicationCoordinator.swift */,
 				5893C6FB29C311E9009090D1 /* ApplicationRouter.swift */,
 				5878F50129CDB989003D4BE2 /* ChangeLogCoordinator.swift */,
-				F07C0A062A52DA64009825CA /* SetupAccountCompletedCoordinator.swift */,
 				58CAF9F92983E0C600BE19F7 /* LoginCoordinator.swift */,
 				583FE00D29C0D586006E85F9 /* OutOfTimeCoordinator.swift */,
 				5847D58C29B7740F008C3808 /* RevokedCoordinator.swift */,
@@ -1943,6 +1957,7 @@
 				587C92FD2986E28100FB9664 /* SelectLocationCoordinator.swift */,
 				58C3F4FA296C3AD500D72515 /* SettingsCoordinator.swift */,
 				F041CD552A38B0B7001B703B /* SettingsRedeemVoucherCoordinator.swift */,
+				F07C0A062A52DA64009825CA /* SetupAccountCompletedCoordinator.swift */,
 				587C92FF2986E2B600FB9664 /* TermsOfServiceCoordinator.swift */,
 				58F185A9298A3E3E00075977 /* TunnelCoordinator.swift */,
 				F0E8CC042A4CC88F007ED3B4 /* WelcomeCoordinator.swift */,
@@ -2342,6 +2357,17 @@
 				F0E361892A4ADCF500AEEF2B /* Welcome */,
 			);
 			path = CreationAccount;
+			sourceTree = "<group>";
+		};
+		F0E8E4BF2A602C7D00ED26A3 /* AccountDeletion */ = {
+			isa = PBXGroup;
+			children = (
+				F0E8E4C02A602CCB00ED26A3 /* AccountDeletionContentView.swift */,
+				F0E8E4C82A604E7400ED26A3 /* AccountDeletionInteractor.swift */,
+				F0E8E4C42A60499100ED26A3 /* AccountDeletionViewController.swift */,
+				F0E8E4C22A602E0D00ED26A3 /* AccountDeletionViewModel.swift */,
+			);
+			path = AccountDeletion;
 			sourceTree = "<group>";
 		};
 /* End PBXGroup section */
@@ -3199,6 +3225,7 @@
 				5846227126E229F20035F7C2 /* StoreSubscription.swift in Sources */,
 				58421030282D8A3C00F24E46 /* UpdateAccountDataOperation.swift in Sources */,
 				F0E8CC052A4CC88F007ED3B4 /* WelcomeCoordinator.swift in Sources */,
+				F0E8E4C92A604E7400ED26A3 /* AccountDeletionInteractor.swift in Sources */,
 				58FF2C03281BDE02009EF542 /* SettingsManager.swift in Sources */,
 				5803B4B02940A47300C23744 /* TunnelConfiguration.swift in Sources */,
 				587EB672271451E300123C75 /* PreferencesViewModel.swift in Sources */,
@@ -3247,6 +3274,7 @@
 				58F2E14C276A61C000A79513 /* RotateKeyOperation.swift in Sources */,
 				5871FB96254ADE4E0051A0A4 /* ConsolidatedApplicationLog.swift in Sources */,
 				068CE57029278F5300A068BB /* MigrationFromV1ToV2.swift in Sources */,
+				F0E8E4C52A60499100ED26A3 /* AccountDeletionViewController.swift in Sources */,
 				F028A54B2A3370FA00C0CAA3 /* RedeemVoucherContentView.swift in Sources */,
 				58FEEB58260B662E00A621A8 /* AutomaticKeyboardResponder.swift in Sources */,
 				5846227326E22A160035F7C2 /* StorePaymentObserver.swift in Sources */,
@@ -3303,6 +3331,8 @@
 				582BB1AF229566420055B6EF /* SettingsCell.swift in Sources */,
 				58F3C0A4249CB069003E76BE /* HeaderBarView.swift in Sources */,
 				5864AF0829C78849005B0CD9 /* CellFactoryProtocol.swift in Sources */,
+				F0C6FA812A66E23300F521F0 /* DeleteAccountOperation.swift in Sources */,
+				F0E8E4C72A604CBE00ED26A3 /* AccountDeletionCoordinator.swift in Sources */,
 				F07CFF2029F2720E008C0343 /* RegisteredDeviceInAppNotificationProvider.swift in Sources */,
 				587A01FC23F1F0BE00B68763 /* SimulatorTunnelProviderHost.swift in Sources */,
 				58CAF9F82983D36800BE19F7 /* Coordinator.swift in Sources */,
@@ -3344,10 +3374,12 @@
 				586891CD29D452E4002A8278 /* SafariCoordinator.swift in Sources */,
 				58ACF64F26567A7100ACE4B7 /* CustomSwitchContainer.swift in Sources */,
 				58EE2E3A272FF814003BFF93 /* SettingsDataSource.swift in Sources */,
+				F0E8E4C12A602CCB00ED26A3 /* AccountDeletionContentView.swift in Sources */,
 				58B26E1E2943514300D5980C /* InAppNotificationDescriptor.swift in Sources */,
 				58421032282E42B000F24E46 /* UpdateDeviceDataOperation.swift in Sources */,
 				06410E04292D0F7100AFC18C /* SettingsParser.swift in Sources */,
 				5878A27D2909657C0096FC88 /* RevokedDeviceInteractor.swift in Sources */,
+				F0E8E4C32A602E0D00ED26A3 /* AccountDeletionViewModel.swift in Sources */,
 				58677710290975E9006F721F /* SettingsInteractorFactory.swift in Sources */,
 				58B26E282943527300D5980C /* SystemNotificationProvider.swift in Sources */,
 				58CCA01222424D11004F3011 /* SettingsViewController.swift in Sources */,
--- a/ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift
+++ b/ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift
@@ -13,7 +13,6 @@ class AutomaticKeyboardResponder {
     weak var targetView: UIView?
     private let handler: (UIView, CGFloat) -> Void
 
-    private var showsKeyboard = false
     private var lastKeyboardRect: CGRect?
 
     private let logger = Logger(label: "AutomaticKeyboardResponder")
@@ -28,58 +27,74 @@ class AutomaticKeyboardResponder {
         NotificationCenter.default.addObserver(
             self,
             selector: #selector(keyboardWillChangeFrame(_:)),
-            name: UIWindow.keyboardWillChangeFrameNotification,
+            name: UIResponder.keyboardWillChangeFrameNotification,
             object: nil
         )
         NotificationCenter.default.addObserver(
             self,
             selector: #selector(keyboardWillShow(_:)),
-            name: UIWindow.keyboardWillShowNotification,
+            name: UIResponder.keyboardWillShowNotification,
             object: nil
         )
         NotificationCenter.default.addObserver(
             self,
-            selector: #selector(keyboardDidHide(_:)),
-            name: UIWindow.keyboardDidHideNotification,
+            selector: #selector(keyboardWillHide(_:)),
+            name: UIResponder.keyboardWillHideNotification,
             object: nil
         )
     }
 
     func updateContentInsets() {
         guard let keyboardRect = lastKeyboardRect else { return }
-
-        adjustContentInsets(keyboardRect: keyboardRect)
+        adjustContentInsets(convertedKeyboardFrameEnd: keyboardRect)
     }
 
     // MARK: - Keyboard notifications
 
     @objc private func keyboardWillShow(_ notification: Notification) {
-        showsKeyboard = true
-
         addPresentationControllerObserver()
-        handleKeyboardNotification(notification)
     }
 
-    @objc private func keyboardDidHide(_ notification: Notification) {
-        showsKeyboard = false
+    @objc private func keyboardWillHide(_ notification: Notification) {
         presentationFrameObserver = nil
     }
 
     @objc private func keyboardWillChangeFrame(_ notification: Notification) {
-        guard showsKeyboard else { return }
-
         handleKeyboardNotification(notification)
     }
 
     // MARK: - Private
 
     private func handleKeyboardNotification(_ notification: Notification) {
-        guard let keyboardFrameValue = notification
-            .userInfo?[UIWindow.keyboardFrameEndUserInfoKey] as? NSValue else { return }
+        guard let userInfo = notification.userInfo,
+              let targetView else { return }
+        // In iOS 16.1 and later, the keyboard notification object is the screen the keyboard appears on.
+        if #available(iOS 16.1, *) {
+            guard let screen = notification.object as? UIScreen,
+                  // Get the keyboard’s frame at the end of its animation.
+                  let keyboardFrameEnd = userInfo[UIResponder.keyboardFrameEndUserInfoKey] as? CGRect else { return }
 
-        lastKeyboardRect = keyboardFrameValue.cgRectValue
+            // Use that screen to get the coordinate space to convert from.
+            let fromCoordinateSpace = screen.coordinateSpace
 
-        adjustContentInsets(keyboardRect: keyboardFrameValue.cgRectValue)
+            // Get your view's coordinate space.
+            let toCoordinateSpace: UICoordinateSpace = targetView
+
+            // Convert the keyboard's frame from the screen's coordinate space to your view's coordinate space.
+            let convertedKeyboardFrameEnd = fromCoordinateSpace.convert(keyboardFrameEnd, to: toCoordinateSpace)
+
+            lastKeyboardRect = convertedKeyboardFrameEnd
+
+            adjustContentInsets(convertedKeyboardFrameEnd: convertedKeyboardFrameEnd)
+        } else {
+            guard let keyboardValue = notification.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue
+            else { return }
+            let keyboardFrameEnd = keyboardValue.cgRectValue
+            let convertedKeyboardFrameEnd = targetView.convert(keyboardFrameEnd, from: targetView.window)
+            lastKeyboardRect = convertedKeyboardFrameEnd
+
+            adjustContentInsets(convertedKeyboardFrameEnd: convertedKeyboardFrameEnd)
+        }
     }
 
     private func addPresentationControllerObserver() {
@@ -100,7 +115,7 @@ class AutomaticKeyboardResponder {
                 guard let self,
                       let keyboardFrameValue = lastKeyboardRect else { return }
 
-                adjustContentInsets(keyboardRect: keyboardFrameValue)
+                adjustContentInsets(convertedKeyboardFrameEnd: keyboardFrameValue)
             }
         )
     }
@@ -112,7 +127,6 @@ class AutomaticKeyboardResponder {
             responder = responder?.next
             return responder
         }
-
         return iterator.first { $0 is UIViewController } as? UIViewController
     }
 
@@ -152,16 +166,24 @@ class AutomaticKeyboardResponder {
         return presentationStyle == .formSheet
     }
 
-    private func adjustContentInsets(keyboardRect: CGRect) {
-        guard let targetView, let superview = targetView.superview else { return }
+    private func adjustContentInsets(convertedKeyboardFrameEnd: CGRect) {
+        guard let targetView else { return }
 
-        // Compute the target view frame within screen coordinates
-        let screenRect = superview.convert(targetView.frame, to: nil)
+        // Get the safe area insets when the keyboard is offscreen.
+        var bottomOffset = targetView.safeAreaInsets.bottom
 
-        // Find the intersection between the keyboard and the view
-        let intersection = keyboardRect.intersection(screenRect)
+        // Get the intersection between the keyboard's frame and the view's bounds to work with the
+        // part of the keyboard that overlaps your view.
+        let viewIntersection = targetView.bounds.intersection(convertedKeyboardFrameEnd)
 
-        handler(targetView, intersection.height)
+        // Check whether the keyboard intersects your view before adjusting your offset.
+        if !viewIntersection.isEmpty {
+            // Adjust the offset by the difference between the view's height and the height of the
+            // intersection rectangle.
+            bottomOffset = targetView.bounds.maxY - viewIntersection.minY
+        }
+
+        handler(targetView, bottomOffset)
     }
 }
 
--- a/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/App/AccountCoordinator.swift
@@ -11,6 +11,7 @@ import UIKit
 enum AccountDismissReason: Equatable {
     case none
     case userLoggedOut
+    case accountDeletion
 }
 
 enum AddedMoreCreditOption: Equatable {
@@ -70,6 +71,8 @@ final class AccountCoordinator: Coordinator, Presentable, Presenting {
             logOut()
         case .navigateToVoucher:
             navigateToRedeemVoucher()
+        case .navigateToDeleteAccount:
+            navigateToDeleteAccount()
         }
     }
 
@@ -92,9 +95,35 @@ final class AccountCoordinator: Coordinator, Presentable, Presenting {
             coordinator,
             animated: true,
             configuration: ModalPresentationConfiguration(
-                preferredContentSize: UIMetrics.RedeemVoucher.preferredContentSize,
-                modalPresentationStyle: .custom,
-                transitioningDelegate: FormSheetTransitioningDelegate()
+                preferredContentSize: UIMetrics.SettingsRedeemVoucher.preferredContentSize,
+                modalPresentationStyle: .formSheet
+            )
+        )
+    }
+
+    private func navigateToDeleteAccount() {
+        let coordinator = AccountDeletionCoordinator(
+            navigationController: CustomNavigationController(),
+            interactor: AccountDeletionInteractor(tunnelManager: interactor.tunnelManager)
+        )
+
+        coordinator.start()
+        coordinator.didCancel = { accountDeletionCoordinator in
+            accountDeletionCoordinator.dismiss(animated: true)
+        }
+
+        coordinator.didFinish = { accountDeletionCoordinator in
+            accountDeletionCoordinator.dismiss(animated: true) {
+                self.didFinish?(self, .userLoggedOut)
+            }
+        }
+
+        presentChild(
+            coordinator,
+            animated: true,
+            configuration: ModalPresentationConfiguration(
+                preferredContentSize: UIMetrics.AccountDeletion.preferredContentSize,
+                modalPresentationStyle: .formSheet
             )
         )
     }
--- a/ios/MullvadVPN/Coordinators/App/AccountDeletionCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/App/AccountDeletionCoordinator.swift
@@ -1,0 +1,47 @@
+//
+//  AccountDeletionCoordinator.swift
+//  MullvadVPN
+//
+//  Created by Mojgan on 2023-07-13.
+//  Copyright © 2023 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+final class AccountDeletionCoordinator: Coordinator, Presentable {
+    private let navigationController: UINavigationController
+    private let interactor: AccountDeletionInteractor
+
+    var didCancel: ((AccountDeletionCoordinator) -> Void)?
+    var didFinish: ((AccountDeletionCoordinator) -> Void)?
+
+    var presentedViewController: UIViewController {
+        navigationController
+    }
+
+    init(
+        navigationController: UINavigationController,
+        interactor: AccountDeletionInteractor
+    ) {
+        self.navigationController = navigationController
+        self.interactor = interactor
+    }
+
+    func start() {
+        navigationController.navigationBar.isHidden = true
+        let viewController = AccountDeletionViewController(interactor: interactor)
+        viewController.delegate = self
+        navigationController.pushViewController(viewController, animated: true)
+    }
+}
+
+extension AccountDeletionCoordinator: AccountDeletionViewControllerDelegate {
+    func deleteAccountDidSucceed(controller: AccountDeletionViewController) {
+        didFinish?(self)
+    }
+
+    func deleteAccountDidCancel(controller: AccountDeletionViewController) {
+        didCancel?(self)
+    }
+}
--- a/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift
+++ b/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift
@@ -29,8 +29,6 @@ class FormSheetPresentationController: UIPresentationController {
      */
     private var lastKnownIsInFullScreen: Bool?
 
-    private var keyboardResponder: AutomaticKeyboardResponder?
-
     private let dimmingView: UIView = {
         let dimmingView = UIView()
         dimmingView.backgroundColor = UIMetrics.DimmingView.backgroundColor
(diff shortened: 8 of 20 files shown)


pinkisemils (changes requested):
> Reviewed 3 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 3 of 17 files reviewed, 2 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadREST/RESTAccountsProxy.swift` line 120 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_mzpsE8GqwkYsLIFXZ:-N_mzpsE8GqwkYsLIFX_:b-jwuttj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e310c1e6fdeff94f718e6e26d1c8bde40532189/ios/MullvadREST/RESTAccountsProxy.swift#L120)):*
> ```Swift
> 
>             return addOperation(
>                 name: "delete-my-account",
> ```

NIT: This could just be `delete-account`.
___
*[`ios/MullvadREST/RESTRequestFactory.swift` line 122 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_mzzcz-Dnj958iFeVg:-N_mzzcz-Dnj958iFeVh:btcdnlg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e310c1e6fdeff94f718e6e26d1c8bde40532189/ios/MullvadREST/RESTRequestFactory.swift#L122)):*
> ```Swift
>         }
> 
>         mutating func addValue(_ forHTTPHeaderField: String, value: String) {
> ```

I think this function should be called `insertHeader` or `addHeader`, as it wasn't obvious to me that a header was being added here.

pinkisemils (changes requested):
> Reviewed 9 of 17 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 12 of 17 files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift` line 22 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_n0qGq-OPXy5hATK8w:-N_n0qGq-OPXy5hATK8x:ba0abu9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e310c1e6fdeff94f718e6e26d1c8bde40532189/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionInteractor.swift#L22)):*
> ```Swift
>                 "INVALID_ACCOUNT_NUMBER",
>                 tableName: "Account",
>                 value: "The four last digits of the account number is invalid.",
> ```

_"Last four digits of the account number are incorrect"_ would be better, I think.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 12 of 17 files reviewed, 4 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 467 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_n04lcANpmPMsK6w4k:-N_n04lcANpmPMsK6w4l:b-z179k4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e310c1e6fdeff94f718e6e26d1c8bde40532189/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L467)):*
> ```Swift
>         )
> 
>         operation.addCondition(MutuallyExclusive(category: OperationCategory.deviceStateUpdate.category))
> ```

I take it this is so that the tunnel manager won't try to update the device state whilst an account deletion is taking place, right?

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 12 of 17 files reviewed, 4 unresolved discussions (waiting on @pinkisemils)

___
*[`ios/MullvadREST/RESTAccountsProxy.swift` line 120 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_mzpsE8GqwkYsLIFXZ:-N_n6wXjETMeuu2Qoelu:b-h3p3cf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e310c1e6fdeff94f718e6e26d1c8bde40532189/ios/MullvadREST/RESTAccountsProxy.swift#L120)):*

NIT: This could just be `delete-account`.
</blockquote></details>

for keeping the naming consistency I've gotten inspiration the pattern we've chosen for other operations on `AccountsProxy`
___
*[`ios/MullvadREST/RESTRequestFactory.swift` line 122 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_mzzcz-Dnj958iFeVg:-N_n84ma35Bck8TtjLi4:bhs4h88) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e310c1e6fdeff94f718e6e26d1c8bde40532189/ios/MullvadREST/RESTRequestFactory.swift#L122)):*

I think this function should be called `insertHeader` or `addHeader`, as it wasn't obvious to me that a header was being added here.
</blockquote></details>

I've borrowed the name `Apple` used for adding header to `URLRequest`.

https://developer.apple.com/documentation/foundation/urlrequest/2011522-addvalue
___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 467 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_n04lcANpmPMsK6w4k:-N_n8jyr5YaLzd4fdgi_:buketii) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e310c1e6fdeff94f718e6e26d1c8bde40532189/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L467)):*

I take it this is so that the tunnel manager won't try to update the device state whilst an account deletion is taking place, right?
</blockquote></details>

Indeed
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift` line 22 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_n0qGq-OPXy5hATK8w:-N_n9AQ81Xrv7DNlrJ1x:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e310c1e6fdeff94f718e6e26d1c8bde40532189/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionInteractor.swift#L22)):*

_"Last four digits of the account number are incorrect"_ would be better, I think.
</blockquote></details>

Done.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 15 of 17 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`ios/CHANGELOG.md` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_nM4MG5uCFIpfIeJfa:-N_nWTr889bHaKVJ7gcT:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6e91a3c0fe18863e45c92eb167121d857006b522/ios/CHANGELOG.md#L34)):*

I've updated the changelog since all entries here except the one that is being added now are a part of the app that just was released. So you should rebase this and move the entry into the unreleased part.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 467 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_n04lcANpmPMsK6w4k:-N_nWUbm-H-NIvWtfcgk:bp3aovt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2e310c1e6fdeff94f718e6e26d1c8bde40532189/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L467)):*

There was no need to remove this, I am OK with these operations being mutually exclusive :sweat_smile: 
</blockquote></details>

Reverted!

pinkisemils (approved):
> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: :shipit: complete! all files reviewed, all discussions resolved

buggmagnet (changes requested):
> Reviewed 5 of 17 files at r1, 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: all files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadREST/RESTAccountsProxy.swift` line 89 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_rZTaK3vPsprdJhDBY:-N_rZTaK3vPsprdJhDBZ:bc7r3q4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8ddb37d1bea7b56998d0fe6018f832d1915ee6a/ios/MullvadREST/RESTAccountsProxy.swift#L89)):*
> ```Swift
>             accountNumber: String,
>             retryStrategy: RetryStrategy,
>             completion: @escaping CompletionHandler<Bool>
> ```

This doesn't need to be `CompletionHandler<Bool>` it can be `CompletionHandler<Void>` instead, the boolean value doesn't add any information since `.success` already tells us that the operation worked. The same goes for the `DeleteAccountOperation` and in the `TunnelManager` 

Here's what the patch looks like after all the changes (Unfortunately I can't attach a diff file, apologies in advance for the seeminly big code blob here)

```swift

diff --git a/ios/MullvadREST/RESTAccountsProxy.swift b/ios/MullvadREST/RESTAccountsProxy.swift
index 4b1fa12a0..42669a794 100644
--- a/ios/MullvadREST/RESTAccountsProxy.swift
+++ b/ios/MullvadREST/RESTAccountsProxy.swift
@@ -86,7 +86,7 @@ extension REST {
         public func deleteAccount(
             accountNumber: String,
             retryStrategy: RetryStrategy,
-            completion: @escaping CompletionHandler<Bool>
+            completion: @escaping CompletionHandler<Void>
         ) -> Cancellable {
             let requestHandler = AnyRequestHandler(createURLRequest: { endpoint, authorization in
                 var requestBuilder = try self.requestFactory.createRequestBuilder(
@@ -100,12 +100,12 @@ extension REST {
                 return requestBuilder.getRequest()
             }, authorizationProvider: createAuthorizationProvider(accountNumber: accountNumber))
 
-            let responseHandler = AnyResponseHandler { response, data -> ResponseHandlerResult<Bool> in
+            let responseHandler = AnyResponseHandler { response, data -> ResponseHandlerResult<Void> in
                 let statusCode = HTTPStatus(rawValue: response.statusCode)
 
                 switch statusCode {
                 case let statusCode where statusCode.isSuccess:
-                    return .success(true)
+                    return .success(())
                 default:
                     return .unhandledResponse(
                         try? self.responseDecoder.decode(
diff --git a/ios/MullvadVPN/TunnelManager/DeleteAccountOperation.swift b/ios/MullvadVPN/TunnelManager/DeleteAccountOperation.swift
index 906228c91..3230e9fde 100644
--- a/ios/MullvadVPN/TunnelManager/DeleteAccountOperation.swift
+++ b/ios/MullvadVPN/TunnelManager/DeleteAccountOperation.swift
@@ -12,7 +12,7 @@ import MullvadREST
 import MullvadTypes
 import Operations
 
-class DeleteAccountOperation: ResultOperation<Bool> {
+class DeleteAccountOperation: ResultOperation<Void> {
     private let logger = Logger(label: "\(DeleteAccountOperation.self)")
 
     private let accountNumber: String
@@ -36,8 +36,8 @@ class DeleteAccountOperation: ResultOperation<Bool> {
             completion: { [weak self] result in
                 self?.dispatchQueue.async {
                     switch result {
-                    case let .success(isDeleted):
-                        self?.finish(result: .success(isDeleted))
+                    case .success:
+                        self?.finish(result: .success(()))
                     case let .failure(error):
                         self?.logger.error(
                             error: error,
diff --git a/ios/MullvadVPN/TunnelManager/TunnelManager.swift b/ios/MullvadVPN/TunnelManager/TunnelManager.swift
index f45cb121f..6ace5425a 100644
--- a/ios/MullvadVPN/TunnelManager/TunnelManager.swift
+++ b/ios/MullvadVPN/TunnelManager/TunnelManager.swift
@@ -433,7 +433,7 @@ final class TunnelManager: StorePaymentObserver {
 
     func deleteAccount(
         accountNumber: String,
-        completion: ((Result<Bool, Error>) -> Void)? = nil
+        completion: ((Result<Void, Error>) -> Void)? = nil
     ) -> Cancellable {
         let operation = DeleteAccountOperation(
             dispatchQueue: internalQueue,
@@ -449,7 +449,7 @@ final class TunnelManager: StorePaymentObserver {
                     self?.operationQueue.cancelAllOperations()
                     self?.wipeAllUserData()
                     self?.setDeviceState(.loggedOut, persist: true)
-                    completion?(.success(true))
+                    completion?(.success(()))
                 }
             case let .failure(error):
                 completion?(.failure(error))
diff --git a/ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift b/ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift
index f818054d3..d0dfd7458 100644
--- a/ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift	
+++ b/ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift	
@@ -48,7 +48,7 @@ class AccountDeletionInteractor {
         }
     }
 
-    func delete(accountNumber: String, completionHandler: @escaping (Result<Bool, Error>) -> Void) -> Cancellable {
+    func delete(accountNumber: String, completionHandler: @escaping (Result<Void, Error>) -> Void) -> Cancellable {
         return tunnelManager.deleteAccount(accountNumber: accountNumber, completion: completionHandler)
     }
 }

```
___
*[`ios/MullvadREST/RESTRequestFactory.swift` line 122 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_rWP6yE0WyU8r-4Q-y:-N_rWP6yE0WyU8r-4Q-z:b56a8xi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8ddb37d1bea7b56998d0fe6018f832d1915ee6a/ios/MullvadREST/RESTRequestFactory.swift#L122)):*
> ```Swift
>         }
> 
>         mutating func addValue(_ forHTTPHeaderField: String, value: String) {
> ```

If we are copying Apple's API names, then this should be named
`addValue(_ value: String, forHTTPHeaderField field: String)`

Here it's confusing which one is the HTTP header field and which one is the value otherwise.
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 65 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_rn4Px6s1S4iqrsY4u:-N_rn4Px6s1S4iqrsY4v:b-3rusg6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8ddb37d1bea7b56998d0fe6018f832d1915ee6a/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L65)):*
> ```Swift
>             tableName: "Account",
>             value: """
>             This logs out all devices using this account and all VPN access will be denied even if there is time left on the account. Enter the last 4 digits of the account number and hit OK if you really want to delete the account:
> ```

nit: Do we need the `:` at the end ?

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 12 of 18 files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadREST/RESTAccountsProxy.swift` line 89 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_rZTaK3vPsprdJhDBY:-N_sAay1-5n_QYquD4w3:b-5i9qxr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8ddb37d1bea7b56998d0fe6018f832d1915ee6a/ios/MullvadREST/RESTAccountsProxy.swift#L89)):*

This doesn't need to be `CompletionHandler<Bool>` it can be `CompletionHandler<Void>` instead, the boolean value doesn't add any information since `.success` already tells us that the operation worked. The same goes for the `DeleteAccountOperation` and in the `TunnelManager` 

Here's what the patch looks like after all the changes (Unfortunately I can't attach a diff file, apologies in advance for the seeminly big code blob here)

```swift

diff --git a/ios/MullvadREST/RESTAccountsProxy.swift b/ios/MullvadREST/RESTAccountsProxy.swift
index 4b1fa12a0..42669a794 100644
--- a/ios/MullvadREST/RESTAccountsProxy.swift
+++ b/ios/MullvadREST/RESTAccountsProxy.swift
@@ -86,7 +86,7 @@ extension REST {
         public func deleteAccount(
             accountNumber: String,
             retryStrategy: RetryStrategy,
-            completion: @escaping CompletionHandler<Bool>
+            completion: @escaping CompletionHandler<Void>
         ) -> Cancellable {
             let requestHandler = AnyRequestHandler(createURLRequest: { endpoint, authorization in
                 var requestBuilder = try self.requestFactory.createRequestBuilder(
@@ -100,12 +100,12 @@ extension REST {
                 return requestBuilder.getRequest()
             }, authorizationProvider: createAuthorizationProvider(accountNumber: accountNumber))
 
-            let responseHandler = AnyResponseHandler { response, data -> ResponseHandlerResult<Bool> in
+            let responseHandler = AnyResponseHandler { response, data -> ResponseHandlerResult<Void> in
                 let statusCode = HTTPStatus(rawValue: response.statusCode)
 
                 switch statusCode {
                 case let statusCode where statusCode.isSuccess:
-                    return .success(true)
+                    return .success(())
                 default:
                     return .unhandledResponse(
                         try? self.responseDecoder.decode(
diff --git a/ios/MullvadVPN/TunnelManager/DeleteAccountOperation.swift b/ios/MullvadVPN/TunnelManager/DeleteAccountOperation.swift
index 906228c91..3230e9fde 100644
--- a/ios/MullvadVPN/TunnelManager/DeleteAccountOperation.swift
+++ b/ios/MullvadVPN/TunnelManager/DeleteAccountOperation.swift
@@ -12,7 +12,7 @@ import MullvadREST
 import MullvadTypes
 import Operations
 
-class DeleteAccountOperation: ResultOperation<Bool> {
+class DeleteAccountOperation: ResultOperation<Void> {
     private let logger = Logger(label: "\(DeleteAccountOperation.self)")
 
     private let accountNumber: String
@@ -36,8 +36,8 @@ class DeleteAccountOperation: ResultOperation<Bool> {
             completion: { [weak self] result in
                 self?.dispatchQueue.async {
                     switch result {
-                    case let .success(isDeleted):
-                        self?.finish(result: .success(isDeleted))
+                    case .success:
+                        self?.finish(result: .success(()))
                     case let .failure(error):
                         self?.logger.error(
                             error: error,
diff --git a/ios/MullvadVPN/TunnelManager/TunnelManager.swift b/ios/MullvadVPN/TunnelManager/TunnelManager.swift
index f45cb121f..6ace5425a 100644
--- a/ios/MullvadVPN/TunnelManager/TunnelManager.swift
+++ b/ios/MullvadVPN/TunnelManager/TunnelManager.swift
@@ -433,7 +433,7 @@ final class TunnelManager: StorePaymentObserver {
 
     func deleteAccount(
         accountNumber: String,
-        completion: ((Result<Bool, Error>) -> Void)? = nil
+        completion: ((Result<Void, Error>) -> Void)? = nil
     ) -> Cancellable {
         let operation = DeleteAccountOperation(
             dispatchQueue: internalQueue,
@@ -449,7 +449,7 @@ final class TunnelManager: StorePaymentObserver {
                     self?.operationQueue.cancelAllOperations()
                     self?.wipeAllUserData()
                     self?.setDeviceState(.loggedOut, persist: true)
-                    completion?(.success(true))
+                    completion?(.success(()))
                 }
             case let .failure(error):
                 completion?(.failure(error))
diff --git a/ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift b/ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift
index f818054d3..d0dfd7458 100644
--- a/ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift	
+++ b/ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift	
@@ -48,7 +48,7 @@ class AccountDeletionInteractor {
         }
     }
 
-    func delete(accountNumber: String, completionHandler: @escaping (Result<Bool, Error>) -> Void) -> Cancellable {
+    func delete(accountNumber: String, completionHandler: @escaping (Result<Void, Error>) -> Void) -> Cancellable {
         return tunnelManager.deleteAccount(accountNumber: accountNumber, completion: completionHandler)
     }
 }

```
</blockquote></details>

You are right.first I consider that server returns different `Status code` for the account has been removed already.
Done!
___
*[`ios/MullvadREST/RESTRequestFactory.swift` line 122 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_rWP6yE0WyU8r-4Q-y:-N_sAKwv9wBJEjXIClpW:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8ddb37d1bea7b56998d0fe6018f832d1915ee6a/ios/MullvadREST/RESTRequestFactory.swift#L122)):*

If we are copying Apple's API names, then this should be named
`addValue(_ value: String, forHTTPHeaderField field: String)`

Here it's confusing which one is the HTTP header field and which one is the value otherwise.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 65 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_rn4Px6s1S4iqrsY4u:-N_sAVUW9IqOM5KWjVWm:b-896fja) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f8ddb37d1bea7b56998d0fe6018f832d1915ee6a/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L65)):*

nit: Do we need the `:` at the end ?
</blockquote></details>

Done!

buggmagnet (changes requested):
> Reviewed 5 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 17 of 18 files reviewed, 4 unresolved discussions (waiting on @mojganii and @pinkisemils)

___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 322 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_t-Ci51ScRCCrI42OD:-N_t-Ci51ScRCCrI42OE:b-3o1xuy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f580f7968c8465c2d1abd0be659931a3d8e95033/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L322)):*
> ```Swift
>         switch Action(rawValue: button.accessibilityIdentifier ?? "") {
>         case .ok:
>             delegate?.didTapDeleteButtonButton(contentView: self, button: button)
> ```

This enum case should really be `delete` instead of `ok`, because otherwise this line is very confusing to read.
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift` line 51 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_sxrrV5ILrmhkIpVjQ:-N_sxrrV5ILrmhkIpVjR:b-hg1tks) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f580f7968c8465c2d1abd0be659931a3d8e95033/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionInteractor.swift#L51)):*
> ```Swift
>     }
> 
>     func delete(accountNumber: String, completionHandler: @escaping (Result<Bool, Error>) -> Void) -> Cancellable {
> ```

This should also be `Result<Void, Error>`

Sorry if I missed it the first time
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionViewController.swift` line 13 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_syM8A7hof2fz6EF5A:-N_syM8A7hof2fz6EF5B:b-kief4f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f580f7968c8465c2d1abd0be659931a3d8e95033/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionViewController.swift#L13)):*
> ```Swift
> 
> protocol AccountDeletionViewControllerDelegate: AnyObject {
>     func deleteAccountDidSucceeded(controller: AccountDeletionViewController)
> ```

This should be named `deleteAccountDidSucceed` since we already have `did` we don't need the `ed` suffix.
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionViewController.swift` line 83 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_szGEy39DKhidXatH_:-N_szGEy39DKhidXatHa:b-77jeex) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f580f7968c8465c2d1abd0be659931a3d8e95033/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionViewController.swift#L83)):*
> ```Swift
>     }
> 
>     func didTapDeleteButtonButton(contentView: AccountDeletionContentView, button: AppButton) {
> ```

I believe you meant `didTapDeleteButton`

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 17 of 18 files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 322 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_t-Ci51ScRCCrI42OD:-Na7c5fQETWFzew3BhFX:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f580f7968c8465c2d1abd0be659931a3d8e95033/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L322)):*

This enum case should really be `delete` instead of `ok`, because otherwise this line is very confusing to read.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionInteractor.swift` line 51 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_sxrrV5ILrmhkIpVjQ:-Na7dAm99HCUCf7bjOOE:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f580f7968c8465c2d1abd0be659931a3d8e95033/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionInteractor.swift#L51)):*

This should also be `Result<Void, Error>`

Sorry if I missed it the first time
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionViewController.swift` line 13 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_syM8A7hof2fz6EF5A:-Na7dOS-B7cwARUdzKBs:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f580f7968c8465c2d1abd0be659931a3d8e95033/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionViewController.swift#L13)):*

This should be named `deleteAccountDidSucceed` since we already have `did` we don't need the `ed` suffix.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionViewController.swift` line 83 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-N_szGEy39DKhidXatH_:-Na7dZnX40PhsPYnkG6H:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f580f7968c8465c2d1abd0be659931a3d8e95033/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionViewController.swift#L83)):*

I believe you meant `didTapDeleteButton` 
</blockquote></details>

Done.

buggmagnet (changes requested):
> Reviewed 2 of 17 files at r1, 1 of 6 files at r5, 8 of 11 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 17 of 19 files reviewed, 4 unresolved discussions (waiting on @mojganii and @pinkisemils)

___
*[`ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift` line 23 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NaB3maG8DD1U5YQ6v7l:-NaB3maG8DD1U5YQ6v7m:bjaxu4q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/88936eed4abdea7f9b3b19cdec4a0462f3fb32d2/ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift#L23)):*
This is not needed anymore since iOS 9.0

> If your app targets iOS 9.0 and later or macOS 10.11 and later, and you used [addObserver(\_:selector:name:object:)](doc://com.apple.documentation/documentation/foundation/notificationcenter/1415360-addobserver?language=swift), you do not need to unregister the observer. If you forget or are unable to remove the observer, the system cleans up the next time it would have posted to it.
___
*[`ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift` line 162 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NaB4FEB0DbRKYXIygi2:-NaB4FEB0DbRKYXIygi3:b-mni1w9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/88936eed4abdea7f9b3b19cdec4a0462f3fb32d2/ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift#L162)):*
> ```Swift
>         guard let targetView, let superview = targetView.superview else { return }
> 
>         // layoutIfNeeded is a method that forces the view to immediately update its layout and the layout of its subviews.
> ```

I don't think we should be adding this comment, there already is documentation for the `layoutIfneeded` method
___
*[`ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift` line 163 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NaB5M40CmJt7p1M17xZ:-NaB5M40CmJt7p1M17x_:b-vphg49) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/88936eed4abdea7f9b3b19cdec4a0462f3fb32d2/ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift#L163)):*
> ```Swift
> 
>         // layoutIfNeeded is a method that forces the view to immediately update its layout and the layout of its subviews.
>         targetView.layoutIfNeeded()
> ```

You want to call `layoutIfNeeded` *after* doing the frame changes, not before, otherwise the method has no effect.
Also I don't think it has any effect here, even if I remove the call to it, the buttons still jump around.

I suggest we remove these 2 `layoutIfNeeded` 
___
*[`ios/MullvadVPN/UI appearance/UIMetrics.swift` line 39 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NaBcqcf-ZD52XMX3DLY:-NaBcqcgE4_r9Dt-uGqM:b-j7jgoe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/04f4f88018d7b8ec3c2114ace46df9f265ef88c3/ios/MullvadVPN/UI%20appearance/UIMetrics.swift#L39)):*
> ```Swift
>     enum RedeemVoucher {
>         static let cornerRadius = 8.0
>         static let preferredContentSize = UIDevice.current.userInterfaceIdiom == .pad
> ```

This has been discussed this offline.
The conclusion is that we don't want to use static sizes because it doesn't work with dynamic layouts.

Instead we should let Autolayout work its magic.

Here is a patch that should address the resizing problem, we can expand on that and see if nothing else breaks by removing code.

```patch
diff --git a/ios/MullvadVPN/Coordinators/Base/ModalPresentationConfiguration.swift b/ios/MullvadVPN/Coordinators/Base/ModalPresentationConfiguration.swift
index e2a5271d3..7fbb6a548 100644
--- a/ios/MullvadVPN/Coordinators/Base/ModalPresentationConfiguration.swift
+++ b/ios/MullvadVPN/Coordinators/Base/ModalPresentationConfiguration.swift
@@ -25,9 +25,9 @@ struct ModalPresentationConfiguration {
             vc.modalPresentationStyle = modalPresentationStyle
         }
 
-        if let preferredContentSize {
-            vc.preferredContentSize = preferredContentSize
-        }
+//        if let preferredContentSize {
+//            vc.preferredContentSize = preferredContentSize
+//        }
 
         if let isModalInPresentation {
             vc.isModalInPresentation = isModalInPresentation
diff --git a/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift b/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift
index e472aa19e..e5c2f347f 100644
--- a/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift     
+++ b/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift     
@@ -53,27 +53,27 @@ class FormSheetPresentationController: UIPresentationController {
             traitCollection.horizontalSizeClass == .compact
     }
 
-    override var frameOfPresentedViewInContainerView: CGRect {
-        guard let containerView else {
-            return super.frameOfPresentedViewInContainerView
-        }
-
-        if isInFullScreenPresentation {
-            return containerView.bounds
-        }
-
-        let preferredContentSize = presentedViewController.preferredContentSize
-
-        assert(preferredContentSize.width > 0 && preferredContentSize.height > 0)
-
-        return CGRect(
-            origin: CGPoint(
-                x: containerView.bounds.midX - preferredContentSize.width * 0.5,
-                y: containerView.bounds.midY - preferredContentSize.height * 0.5
-            ),
-            size: preferredContentSize
-        )
-    }
+//    override var frameOfPresentedViewInContainerView: CGRect {
+//        guard let containerView else {
+//            return super.frameOfPresentedViewInContainerView
+//        }
+//
+//        if isInFullScreenPresentation {
+//            return containerView.bounds
+//        }
+//
+//        let preferredContentSize = presentedViewController.preferredContentSize
+//
+//        assert(preferredContentSize.width > 0 && preferredContentSize.height > 0)
+//
+//        return CGRect(
+//            origin: CGPoint(
+//                x: containerView.bounds.midX - preferredContentSize.width * 0.5,
+//                y: containerView.bounds.midY - preferredContentSize.height * 0.5
+//            ),
+//            size: preferredContentSize
+//        )
+//    }
 
     override func containerViewWillLayoutSubviews() {
         dimmingView.frame = containerView?.bounds ?? .zero

```

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 15 of 19 files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift` line 23 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NaB3maG8DD1U5YQ6v7l:-NaBfRob1qP9czfJqUta:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/88936eed4abdea7f9b3b19cdec4a0462f3fb32d2/ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift#L23)):*

This is not needed anymore since iOS 9.0

> If your app targets iOS 9.0 and later or macOS 10.11 and later, and you used [addObserver(\_:selector:name:object:)](doc://com.apple.documentation/documentation/foundation/notificationcenter/1415360-addobserver?language=swift), you do not need to unregister the observer. If you forget or are unable to remove the observer, the system cleans up the next time it would have posted to it.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift` line 162 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NaB4FEB0DbRKYXIygi2:-NaBfX1g1Nb3C4S8jyuU:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/88936eed4abdea7f9b3b19cdec4a0462f3fb32d2/ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift#L162)):*

I don't think we should be adding this comment, there already is documentation for the `layoutIfneeded` method
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift` line 163 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NaB5M40CmJt7p1M17xZ:-NaBgNqPB2VhORQ7XWGE:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/88936eed4abdea7f9b3b19cdec4a0462f3fb32d2/ios/MullvadVPN/Classes/AutomaticKeyboardResponder.swift#L163)):*

You want to call `layoutIfNeeded` *after* doing the frame changes, not before, otherwise the method has no effect.
Also I don't think it has any effect here, even if I remove the call to it, the buttons still jump around.

I suggest we remove these 2 `layoutIfNeeded` 
</blockquote></details>

Done

buggmagnet (commented):
> Reviewed 1 of 11 files at r6, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: all files reviewed, 1 unresolved discussion (waiting on @mojganii)

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/UI appearance/UIMetrics.swift` line 39 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NaBcqcf-ZD52XMX3DLY:-NaClPqwDYxCh1jUFzks:b-886gsx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/04f4f88018d7b8ec3c2114ace46df9f265ef88c3/ios/MullvadVPN/UI%20appearance/UIMetrics.swift#L39)):*

This has been discussed this offline.
The conclusion is that we don't want to use static sizes because it doesn't work with dynamic layouts.

Instead we should let Autolayout work its magic.

Here is a patch that should address the resizing problem, we can expand on that and see if nothing else breaks by removing code.

```patch
diff --git a/ios/MullvadVPN/Coordinators/Base/ModalPresentationConfiguration.swift b/ios/MullvadVPN/Coordinators/Base/ModalPresentationConfiguration.swift
index e2a5271d3..7fbb6a548 100644
--- a/ios/MullvadVPN/Coordinators/Base/ModalPresentationConfiguration.swift
+++ b/ios/MullvadVPN/Coordinators/Base/ModalPresentationConfiguration.swift
@@ -25,9 +25,9 @@ struct ModalPresentationConfiguration {
             vc.modalPresentationStyle = modalPresentationStyle
         }
 
-        if let preferredContentSize {
-            vc.preferredContentSize = preferredContentSize
-        }
+//        if let preferredContentSize {
+//            vc.preferredContentSize = preferredContentSize
+//        }
 
         if let isModalInPresentation {
             vc.isModalInPresentation = isModalInPresentation
diff --git a/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift b/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift
index e472aa19e..e5c2f347f 100644
--- a/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift     
+++ b/ios/MullvadVPN/Presentation controllers/FormsheetPresentationController.swift     
@@ -53,27 +53,27 @@ class FormSheetPresentationController: UIPresentationController {
             traitCollection.horizontalSizeClass == .compact
     }
 
-    override var frameOfPresentedViewInContainerView: CGRect {
-        guard let containerView else {
-            return super.frameOfPresentedViewInContainerView
-        }
-
-        if isInFullScreenPresentation {
-            return containerView.bounds
-        }
-
-        let preferredContentSize = presentedViewController.preferredContentSize
-
-        assert(preferredContentSize.width > 0 && preferredContentSize.height > 0)
-
-        return CGRect(
-            origin: CGPoint(
-                x: containerView.bounds.midX - preferredContentSize.width * 0.5,
-                y: containerView.bounds.midY - preferredContentSize.height * 0.5
-            ),
-            size: preferredContentSize
-        )
-    }
+//    override var frameOfPresentedViewInContainerView: CGRect {
+//        guard let containerView else {
+//            return super.frameOfPresentedViewInContainerView
+//        }
+//
+//        if isInFullScreenPresentation {
+//            return containerView.bounds
+//        }
+//
+//        let preferredContentSize = presentedViewController.preferredContentSize
+//
+//        assert(preferredContentSize.width > 0 && preferredContentSize.height > 0)
+//
+//        return CGRect(
+//            origin: CGPoint(
+//                x: containerView.bounds.midX - preferredContentSize.width * 0.5,
+//                y: containerView.bounds.midY - preferredContentSize.height * 0.5
+//            ),
+//            size: preferredContentSize
+//        )
+//    }
 
     override func containerViewWillLayoutSubviews() {
         dimmingView.frame = containerView?.bounds ?? .zero

```
</blockquote></details>

Fixed

pronebird (commented):
> Reviewed 4 of 8 files at r10.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 19 of 20 files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/UI appearance/UIMetrics.swift` line 39 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NaBcqcf-ZD52XMX3DLY:-Nal6NaZ3MPpAI0XdmlA:b-7o7d4u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/04f4f88018d7b8ec3c2114ace46df9f265ef88c3/ios/MullvadVPN/UI%20appearance/UIMetrics.swift#L39)):*

Fixed
</blockquote></details>

The default implementation of `var frameOfPresentedViewInContainerView: CGRect` returns the frame of container view which is used during rotation and presentation to set the presented view size. So commenting it out is probably not going to work.

I think traditionally `preferredContentSize` when set to `.zero` or lower is treated as intrinsic. But that has to be handled in the presentation controller. 

If we were to handle that case then we could calculate the frame of the presented view using AutoLayout engine, i.e:

```swift
guard let presentedView else { return super.frameOfPresentedViewInContainerView }

var preferred = preferredContentSize
var calculated: CGSize

if preferred.width <= 0 || preferred.height <= 0 {
    let layoutSize = CGSize(
        width: preferred.width <= 0 ? UIView.layoutFittingCompressedSize.width : preferred.width,
        height: preferred.height <= 0 ? UIView.layoutFittingCompressedSize.height : preferred.height
    )
    calculated = presentedView.systemLayoutSizeFitting(layoutSize)
} else {
    calculated = preferred
}

return CGRect(...)
```

This article could shed some light on how to compute view size: https://useyourloaf.com/blog/self-sizing-popovers/

I do however hope that such changes could be tested in isolation and not merged as a part of larger PR if we opt-in to work on that.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 18 of 20 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 192 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NalJR3n-U_3JpJAz-pa:-NalJR3o801wlYJPoJ0u:b-jnmbo0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L192)):*
> ```Swift
>     var isEditing = false {
>         didSet {
>             _ = accountTextField.isFirstResponder
> ```

`didSet` acts as a toggle. Shouldn't it respect the value of `isEditing`?

```
if isEditing {
    accountTextField.becomeFirstResponder()
} else {
    accountTextField.resignFirstResponder()
}
```

pronebird (commented):
> Reviewed 1 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 18 of 20 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 356 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-Nal8MYZ7I5nRxIxJDBO:-NalMHNe4l2cS0TO52R-:b-uud26o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L356)):*

Let's add a `self.scrollView.flashScrollIndicators()` before the `layoutIfNeeded`. It adds for a nicer UX.
</blockquote></details>

This closure will be called in response to keyboard events so flashing scrollbars is better be done from `viewDidAppear` of the view controller hosting this view.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 18 of 20 files reviewed, 5 unresolved discussions (waiting on @mojganii and @pronebird)

___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 356 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-Nal8MYZ7I5nRxIxJDBO:-Nap05Eo0K04EVlmdCSJ:b-fbcx67) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L356)):*

This closure will be called in response to keyboard events so flashing scrollbars is better be done from `viewDidAppear` of the view controller hosting this view.
</blockquote></details>

The view controller doesn't (and shouldn't) know that the content view has a scroll view. Exposing the scroll view would break encapsulation. 

Also, I don't think the scroll indicator will show if there's nothing to scroll, so calling it before the keyboard appears on screen wouldn't work

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 18 of 20 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadVPN/UI appearance/UIMetrics.swift` line 39 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NaBcqcf-ZD52XMX3DLY:-NalCKOZ67Zeh-gn542h:b-q8hfcy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/04f4f88018d7b8ec3c2114ace46df9f265ef88c3/ios/MullvadVPN/UI%20appearance/UIMetrics.swift#L39)):*

The default implementation of `var frameOfPresentedViewInContainerView: CGRect` returns the frame of container view which is used during rotation and presentation to set the presented view size. So commenting it out is probably not going to work.

I think traditionally `preferredContentSize` when set to `.zero` or lower is treated as intrinsic. But that has to be handled in the presentation controller. 

If we were to handle that case then we could calculate the frame of the presented view using AutoLayout engine, i.e:

```swift
guard let presentedView else { return super.frameOfPresentedViewInContainerView }

var preferred = preferredContentSize
var calculated: CGSize

if preferred.width <= 0 || preferred.height <= 0 {
    let layoutSize = CGSize(
        width: preferred.width <= 0 ? UIView.layoutFittingCompressedSize.width : preferred.width,
        height: preferred.height <= 0 ? UIView.layoutFittingCompressedSize.height : preferred.height
    )
    calculated = presentedView.systemLayoutSizeFitting(layoutSize)
} else {
    calculated = preferred
}

return CGRect(...)
```

This article could shed some light on how to compute view size: https://useyourloaf.com/blog/self-sizing-popovers/

I do however hope that such changes could be tested in isolation and not merged as a part of larger PR if we opt-in to work on that.
</blockquote></details>

I didn't touch this part.but we can spend time on it for the future to make that adaptable for different size classes 
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 192 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NalJR3n-U_3JpJAz-pa:-NalQ3dnDx-82CkBeyFI:bh2uy5d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L192)):*

`didSet` acts as a toggle. Shouldn't it respect the value of `isEditing`?

```
if isEditing {
    accountTextField.becomeFirstResponder()
} else {
    accountTextField.resignFirstResponder()
}
```
</blockquote></details>

Nope this property is for toggling value for out of the class and I'm changing that at `viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator)`.
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 356 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-Nal8MYZ7I5nRxIxJDBO:-NalClfo4TDQdjddsFlB:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L356)):*

Let's add a `self.scrollView.flashScrollIndicators()` before the `layoutIfNeeded`. It adds for a nicer UX.
</blockquote></details>

Done.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 18 of 20 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 192 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NalJR3n-U_3JpJAz-pa:-NapFmMW9UQ7OZCOo-Sf:b-7rl0kd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L192)):*

Nope this property is for toggling value for out of the class and I'm changing that at `viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator)`.
</blockquote></details>

done

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 15 of 20 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionViewController.swift` line 65 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-Nal6H6yBvA4gX0pahH7:-Napf4tvEQR-OcU3ChdB:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionViewController.swift#L65)):*

No need for `self` here
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherViewController.swift` line 60 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-Nal9gVuErMb6EMt3RPl:-Napf3wIAx55DVkpncUc:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherViewController.swift#L60)):*

We can remove `self` here and in `enableEditing` and `disableEditing` 
</blockquote></details>

Done.

pronebird (commented):
> Reviewed 2 of 17 files at r1, 1 of 1 files at r4, 2 of 6 files at r5, 1 of 11 files at r6, 1 of 8 files at r10, 5 of 5 files at r12.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/UI appearance/UIMetrics.swift` line 39 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NaBcqcf-ZD52XMX3DLY:-NaqGN7SAvHT1-tB7ZSi:b-hmp9vu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/04f4f88018d7b8ec3c2114ace46df9f265ef88c3/ios/MullvadVPN/UI%20appearance/UIMetrics.swift#L39)):*

I didn't touch this part.but we can spend time on it for the future to make that adaptable for different size classes 
</blockquote></details>

Cool. Agreed.
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 356 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-Nal8MYZ7I5nRxIxJDBO:-NaqE7L06KWa6MaGwTHT:b38navi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L356)):*
@buggmagnet that's how it's been with standard `UITableViewController`, see documentation states that explicitly:

> -   Implements the [`viewDidAppear(_:)`](doc://com.apple.documentation/documentation/uikit/uiviewcontroller/1621423-viewdidappear?language=swift) method and automatically flashes the table view’s scroll indicators when it first appears

https://developer.apple.com/documentation/uikit/uitableviewcontroller
___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 276 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-Naq7bsL9ya23wrSCVr8:-Naq7bsL9ya23wrSCVr9:bqb2utd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75c397b7c9f918da514079e4bdf43e1a46fe1c9a/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L276)):*
> ```Swift
>     override func didMoveToWindow() {
>         if self.window != nil {
>             NotificationCenter.default.addObserver(
> ```

Nit: technically you could add it from `init()` and never unregister it. Notifications for selector based handlers are automatically unregistered around time when observer (`self`) deinits and they don't retain the observer.

buggmagnet (commented):
> Reviewed 1 of 8 files at r10, 2 of 5 files at r12.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: all files reviewed, 3 unresolved discussions (waiting on @mojganii and @pronebird)

___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 356 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-Nal8MYZ7I5nRxIxJDBO:-Naqh5-p5phQGKsdLrkF:b4pnnix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L356)):*

@buggmagnet that's how it's been with standard `UITableViewController`, see documentation states that explicitly:

> -   Implements the [`viewDidAppear(_:)`](doc://com.apple.documentation/documentation/uikit/uiviewcontroller/1621423-viewdidappear?language=swift) method and automatically flashes the table view’s scroll indicators when it first appears

https://developer.apple.com/documentation/uikit/uitableviewcontroller
</blockquote></details>

Right I understand that, but this is not a `UITableViewController` that we're dealing with here, and it's a separate problem.
The UX we want to have is that a view that's not scrollable becomes scrollable in certain cases (in our case, the content doesn't fit the screen size anymore because the keyboard eats the screen estate) . 
When it becomes scrollable, we want to flash the scroll indicator. 

It's not because `UITableViewController` flashes it in `viewDidAppear` that it means it's the only place where it should be called.
If you want to test is yourself, look in the iOS settings.

If you go to the `VPN` section, you won't see the flash indicator scrolling, because there is not enough content to scroll through.
However, if you open the `General` section, you will see the flash indicator scrolling about after a second.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 356 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-Nal8MYZ7I5nRxIxJDBO:-NauPS8H6rnuunCiy9ba:bgfn7qg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L356)):*

Right I understand that, but this is not a `UITableViewController` that we're dealing with here, and it's a separate problem.
The UX we want to have is that a view that's not scrollable becomes scrollable in certain cases (in our case, the content doesn't fit the screen size anymore because the keyboard eats the screen estate) . 
When it becomes scrollable, we want to flash the scroll indicator. 

It's not because `UITableViewController` flashes it in `viewDidAppear` that it means it's the only place where it should be called.
If you want to test is yourself, look in the iOS settings.

If you go to the `VPN` section, you won't see the flash indicator scrolling, because there is not enough content to scroll through.
However, if you open the `General` section, you will see the flash indicator scrolling about after a second.
</blockquote></details>

Alright. I don't recall stock apps flashing scroll indicators in response to keyboard but if you think it improves UX then it's fine by me. 

My only suggestion would be, could we please use `[weak self]` here because the closure is retained by `AutomaticKeyboardResponder` and `keyboardResponder` is retained by `self`, so that would ultimately leak memory unless we break it with `[weak self]`.

pronebird (commented):
> Reviewed 2 of 11 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 436 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-NauTUmz5o0n2Fhoel-G:-NauTUmz5o0n2Fhoel-H:b-buzsc2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75c397b7c9f918da514079e4bdf43e1a46fe1c9a/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L436)):*
> ```Swift
>     func deleteAccount(
>         accountNumber: String,
>         completion: ((Result<Void, Error>) -> Void)? = nil
> ```

Nit: if no value is expected to be returned then you can instead of `Result<Void, Error>` you can use `Error?` instead. where `nil` is success, and `.some(Error)` means error.
___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 448 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-Naub3s8BSDltZrk6A9i:-Naub3s8BSDltZrk6A9j:b-ii0sma) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75c397b7c9f918da514079e4bdf43e1a46fe1c9a/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L448)):*
> ```Swift
>             switch result {
>             case .success:
>                 self?.unsetTunnelConfiguration {
> ```

Nit: I'd rather see it being a part of `DeleteAccountOperation` and then you could cancel all operations from completion handler.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914)* status: 18 of 20 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift` line 356 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4914#-Nal8MYZ7I5nRxIxJDBO:-NauaTBf31nFZwdpVaG5:b-s8gcnm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/16036b2e6548964d7527f8d192bfb2a65835a288/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L356)):*

Alright. I don't recall stock apps flashing scroll indicators in response to keyboard but if you think it improves UX then it's fine by me. 

My only suggestion would be, could we please use `[weak self]` here because the closure is retained by `AutomaticKeyboardResponder` and `keyboardResponder` is retained by `self`, so that would ultimately leak memory unless we break it with `[weak self]`.
</blockquote></details>

good catch.Done!

PR #4921, opened 2023-07-24
Cache the TransportStrategy counter and reuse it between processes

buggmagnet:
This PR implements caching the `connectionAttempts` from the `TransportStrategy` in order to provide a better UX.

Suppose only network connections via shadowsocks would work.
The first request made by the application at launch time would always fail.

In this PR, we start recording the number of failures so that the next time the application (or network extension) is launched, we will directly select the last successful connection mode (i.e. one that did not fail)

The connection attempt cached value is shared by both processes via `UserDefaults` shared by the same application group.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)

### DIFF
--- a/ios/MullvadREST/TransportStrategy.swift
+++ b/ios/MullvadREST/TransportStrategy.swift
@@ -1,5 +1,5 @@
 //
-//  RESTTransportStrategy.swift
+//  TransportStrategy.swift
 //  MullvadREST
 //
 //  Created by Marco Nikic on 2023-04-27.
@@ -7,8 +7,9 @@
 //
 
 import Foundation
+import MullvadTypes
 
-public struct TransportStrategy: Codable, Equatable {
+public struct TransportStrategy: Equatable {
     /// The different transports suggested by the strategy
     public enum Transport {
         /// Suggests using a direct connection
@@ -23,10 +24,17 @@ public struct TransportStrategy: Codable, Equatable {
     /// suggestion.
     ///
     /// `internal` instead of `private` for testing purposes.
-    internal var connectionAttempts: UInt
+    internal var connectionAttempts: Int
 
-    public init() {
-        connectionAttempts = 0
+    /// Enables recording of failed connection attempts.
+    private let userDefaults: UserDefaults
+
+    /// `UserDefaults` key shared by both processes. Used to cache and synchronize connection attempts between them.
+    internal static let connectionAttemptsSharedCacheKey = "ConnectionAttemptsSharedCacheKey"
+
+    public init(_ userDefaults: UserDefaults) {
+        self.connectionAttempts = userDefaults.integer(forKey: Self.connectionAttemptsSharedCacheKey)
+        self.userDefaults = userDefaults
     }
 
     /// Instructs the strategy that a network connection failed
@@ -34,8 +42,10 @@ public struct TransportStrategy: Codable, Equatable {
     /// Every third failure results in a direct transport suggestion.
     public mutating func didFail() {
         let (partial, isOverflow) = connectionAttempts.addingReportingOverflow(1)
-        // UInt.max is a multiple of 3, go directly to 1 when overflowing
-        connectionAttempts = isOverflow ? 1 : partial
+        // (Int.max - 1) is a multiple of 3, go directly to 2 when overflowing
+        // to keep the "every third failure" algorithm correct
+        connectionAttempts = isOverflow ? 2 : partial
+        userDefaults.set(connectionAttempts, forKey: Self.connectionAttemptsSharedCacheKey)
     }
 
     /// The suggested connection transport
@@ -43,5 +53,9 @@ public struct TransportStrategy: Codable, Equatable {
     /// - Returns: `.useURLSession` for every 3rd failed attempt, `.useShadowsocks` otherwise
     public func connectionTransport() -> Transport {
         connectionAttempts.isMultiple(of: 3) ? .useURLSession : .useShadowsocks
+    }
+
+    public static func == (lhs: TransportStrategy, rhs: TransportStrategy) -> Bool {
+        lhs.connectionAttempts == rhs.connectionAttempts
     }
 }
--- a/ios/MullvadRESTTests/TransportStrategyTests.swift
+++ b/ios/MullvadRESTTests/TransportStrategyTests.swift
@@ -7,41 +7,53 @@
 //
 
 @testable import MullvadREST
+@testable import MullvadTypes
 import XCTest
 
 final class TransportStrategyTests: XCTestCase {
+    var userDefaults: UserDefaults!
+    static var suiteName: String!
+
+    override class func setUp() {
+        super.setUp()
+        suiteName = UUID().uuidString
+    }
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+        userDefaults = UserDefaults(suiteName: Self.suiteName)
+    }
+
+    override func tearDownWithError() throws {
+        userDefaults.removePersistentDomain(forName: Self.suiteName)
+        try super.tearDownWithError()
+    }
+
     func testEveryThirdConnectionAttemptsIsDirect() {
-        loopStrategyTest(with: TransportStrategy())
+        loopStrategyTest(with: TransportStrategy(userDefaults), in: 0 ... 12)
     }
 
     func testOverflowingConnectionAttempts() {
-        var strategy = TransportStrategy()
-        strategy.connectionAttempts = UInt.max
+        userDefaults.set(Int.max, forKey: TransportStrategy.connectionAttemptsSharedCacheKey)
+        let strategy = TransportStrategy(userDefaults)
 
-        loopStrategyTest(with: strategy)
+        // (Int.max - 1) is a multiple of 3, so skip the first iteration
+        loopStrategyTest(with: strategy, in: 1 ... 12)
     }
 
-    func testLoadingFromCacheDoesNotImpactStrategy() throws {
-        var strategy = TransportStrategy()
+    func testConnectionAttemptsAreRecordedAfterFailure() {
+        var strategy = TransportStrategy(userDefaults)
 
-        // Fail twice, the next suggested transport mode should be via Shadowsocks proxy
         strategy.didFail()
-        strategy.didFail()
-        XCTAssertEqual(strategy.connectionTransport(), .useShadowsocks)
 
-        // Serialize the strategy and reload it from memory to simulate an application restart
-        let encodedRawStrategy = try JSONEncoder().encode(strategy)
-        var reloadedStrategy = try JSONDecoder().decode(TransportStrategy.self, from: encodedRawStrategy)
-
-        // This should be the third failure, the next suggested transport will be a direct one
-        reloadedStrategy.didFail()
-        XCTAssertEqual(reloadedStrategy.connectionTransport(), .useURLSession)
+        let recordedValue = userDefaults.integer(forKey: TransportStrategy.connectionAttemptsSharedCacheKey)
+        XCTAssertEqual(1, recordedValue)
     }
 
-    private func loopStrategyTest(with strategy: TransportStrategy) {
+    private func loopStrategyTest(with strategy: TransportStrategy, in range: ClosedRange<Int>) {
         var strategy = strategy
 
-        for index in 0 ... 12 {
+        for index in range {
             let expectedResult: TransportStrategy.Transport
             expectedResult = index.isMultiple(of: 3) ? .useURLSession : .useShadowsocks
             XCTAssertEqual(strategy.connectionTransport(), expectedResult)
--- a/ios/MullvadTransport/TransportProvider.swift
+++ b/ios/MullvadTransport/TransportProvider.swift
@@ -31,7 +31,7 @@ public final class TransportProvider: RESTTransport {
         relayCache: RelayCache,
         addressCache: REST.AddressCache,
         shadowsocksCache: ShadowsocksConfigurationCache,
-        transportStrategy: TransportStrategy = .init(),
+        transportStrategy: TransportStrategy,
         constraintsUpdater: RelayConstraintsUpdater
     ) {
         self.urlSessionTransport = urlSessionTransport
--- a/ios/MullvadTypes/AttemptsRecording.swift
+++ b/ios/MullvadTypes/AttemptsRecording.swift
@@ -1,0 +1,16 @@
+//
+//  AttemptsRecording.swift
+//  MullvadTypes
+//
+//  Created by Marco Nikic on 2023-07-24.
+//  Copyright © 2023 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+
+/// Defines a generic way to record an attempt.
+///
+/// Used by `TransportStrategy` to record failed connection attempts in cache.
+public protocol AttemptsRecording {
+    func record(_ attempts: Int)
+}
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -399,7 +399,8 @@
 		7ABE318D2A1CDD4500DF4963 /* UIFont+Weight.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7ABE318C2A1CDD4500DF4963 /* UIFont+Weight.swift */; };
 		7AE47E522A17972A000418DA /* CustomAlertViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AE47E512A17972A000418DA /* CustomAlertViewController.swift */; };
 		7AF0419E29E957EB00D492DD /* AccountCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AF0419D29E957EB00D492DD /* AccountCoordinator.swift */; };
-		A917351F29FAA9C400D5DCFD /* RESTTransportStrategy.swift in Sources */ = {isa = PBXBuildFile; fileRef = A917351E29FAA9C400D5DCFD /* RESTTransportStrategy.swift */; };
+		A90F9DEC2A6E91A5009DC8B3 /* AttemptsRecording.swift in Sources */ = {isa = PBXBuildFile; fileRef = A90F9DEB2A6E91A5009DC8B3 /* AttemptsRecording.swift */; };
+		A917351F29FAA9C400D5DCFD /* TransportStrategy.swift in Sources */ = {isa = PBXBuildFile; fileRef = A917351E29FAA9C400D5DCFD /* TransportStrategy.swift */; };
 		A917352129FAAA5200D5DCFD /* TransportStrategyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A917352029FAAA5200D5DCFD /* TransportStrategyTests.swift */; };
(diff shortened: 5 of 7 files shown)


mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: 0 of 9 files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/RESTTransportStrategy.swift` line 2 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-NafsfUfDmwSXbirmUhC:-NafsfUfDmwSXbirmUhD:b-ie1va5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadREST/RESTTransportStrategy.swift#L2)):*
> ```Swift
> //
> //  RESTTransportStrategy.swift
> ```

I believe that it slipped the name of the class is not equal to this
___
*[`ios/MullvadREST/RESTTransportStrategy.swift` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-NafqUSYESZuaPPITo1w:-NafqUSYESZuaPPITo1x:b-y1o9ui) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadREST/RESTTransportStrategy.swift#L32)):*
> ```Swift
>     private let attemptsRecorder: AttemptsRecording
> 
>     public init(connectionAttempts: Int = 0, attemptsRecorder: AttemptsRecording) {
> ```

when we have cached value in `UserDefaults`,there is no need to take connection attempt as a parameter. `AttemptsRecording` can have method for reading value and set it internally in the class.
if there is for the sake of test,let's go that with mocking `AttemptsRecording` as I saw it's already implemented 
___
*[`ios/MullvadREST/RESTTransportStrategy.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Naftrpe5BgZ-68aQIOo:-Naftrpe5BgZ-68aQIOp:b-3yvsom) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadREST/RESTTransportStrategy.swift#L41)):*
> ```Swift
>     /// Every third failure results in a direct transport suggestion.
>     public mutating func didFail() {
>         let (partial, isOverflow) = connectionAttempts.addingReportingOverflow(1)
> ```

as I understood,this line increases 1 to the value of `connectionAttempts` with caring about overflow flag,how it understands about whether the result is a multiple of 3 or not,while we have another `isMultiple`  for that.It looked liked to me when overflow happens let's go 2 otherwise add the counter.maybe it's better to change the comment for this part
___
*[`ios/MullvadVPN/AppDelegate.swift` line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Nafwc8o9uSQL1ydIxvN:-Nafwc8o9uSQL1ydIxvO:bqyxzm3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadVPN/AppDelegate.swift#L106)):*
> ```Swift
>         // This init cannot fail as long as the security group identifier is valid
>         let sharedUserDefaults = UserDefaults(suiteName: ApplicationConfiguration.securityGroupIdentifier)!
>         let connectionAttempts = sharedUserDefaults
> ```

it should be retrieved from the you storage protocol
___
*[`ios/Shared/ApplicationConfiguration.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Nafx7jd8wMHx0_WCwa_:-Nafx7jd8wMHx0_WCwaa:b-q2800b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/Shared/ApplicationConfiguration.swift#L41)):*
> ```Swift
> 
>     /// `UserDefaults` key shared by both processes. Used to cache and synchronize connection attempts between them.
>     static let connectionAttemptsSharedCacheKey = "ConnectionAttemptsSharedCacheKey"
> ```

it's better to consider that in your protocol storage.there is no need to be available in the whole of the application

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: 0 of 9 files reviewed, 5 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadREST/RESTTransportStrategy.swift` line 2 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-NafsfUfDmwSXbirmUhC:-Nag3nx_3jmqd66-KHjv:buhy53) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadREST/RESTTransportStrategy.swift#L2)):*

I believe that it slipped the name of the class is not equal to this
</blockquote></details>

Indeed, nice catch !
___
*[`ios/MullvadREST/RESTTransportStrategy.swift` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-NafqUSYESZuaPPITo1w:-Nag48Lz77zgSKSOaD4W:b-1hl9jd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadREST/RESTTransportStrategy.swift#L32)):*
> AttemptsRecording can have method for reading value and set it internally in the class.
I disagree, the point is that it's a write only protocol, is doesn't need to read values. That would break separation of concerns.
___
*[`ios/MullvadREST/RESTTransportStrategy.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Naftrpe5BgZ-68aQIOo:-Nag4qVC--4CVhbraTsQ:bcnyzwt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadREST/RESTTransportStrategy.swift#L41)):*

as I understood,this line increases 1 to the value of `connectionAttempts` with caring about overflow flag,how it understands about whether the result is a multiple of 3 or not,while we have another `isMultiple`  for that.It looked liked to me when overflow happens let's go 2 otherwise add the counter.maybe it's better to change the comment for this part
</blockquote></details>

I'm really sorry but I don't understand your comment. What are you suggesting to change, and to what ?
___
*[`ios/Shared/ApplicationConfiguration.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Nafx7jd8wMHx0_WCwa_:-Nag5VdDDxmBPzULHmg5:bdk28up) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/Shared/ApplicationConfiguration.swift#L41)):*

it's better to consider that in your protocol storage.there is no need to be available in the whole of the application
</blockquote></details>

Do you mean moving this value to the `AttemptsRecording` file ?

mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: 0 of 9 files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/Shared/ApplicationConfiguration.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Nafx7jd8wMHx0_WCwa_:-Nag737B4ZUYGe6yLBwl:b2nign3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/Shared/ApplicationConfiguration.swift#L41)):*

Do you mean moving this value to the `AttemptsRecording` file ?
</blockquote></details>

yep
___
*[`ios/MullvadREST/RESTTransportStrategy.swift` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-NafqUSYESZuaPPITo1w:-Nag6KmN6psGnlHqPIZy:b-4us8ig) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadREST/RESTTransportStrategy.swift#L32)):*

> AttemptsRecording can have method for reading value and set it internally in the class.
I disagree, the point is that it's a write only protocol, is doesn't need to read values. That would break separation of concerns.
</blockquote></details>

it has nothing to do separation of concerns when you are feeding that from your storage.I believe that is related to single source of truth.

mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: 0 of 9 files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/RESTTransportStrategy.swift` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-NafqUSYESZuaPPITo1w:-NahQm0e1opBGSMbYJKO:b-6jbzwz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadREST/RESTTransportStrategy.swift#L32)):*

If we read from `UserDefaults` every time instead of keeping a local counter, it would break the `Equatable` implementation, and in turn the way `TransportProvider` guarantees that multiple requests failing at the same time do not change the `TransportStrategy` several times in a row.
</blockquote></details>

there is no need to read that everytime, when we check if there is a value for that on memory or not.if there is not then load from `UserDefaults`.It doesn't make sense to me to store the value from `AttemptsRecording` and read that out of the abstraction.
maybe it's better to know other.
___
*[`ios/MullvadREST/RESTTransportStrategy.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Naftrpe5BgZ-68aQIOo:-NahQUg_Dd34RJ88hmaI:batsiq6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadREST/RESTTransportStrategy.swift#L41)):*

I'm really sorry but I don't understand your comment. What are you suggesting to change, and to what ?
</blockquote></details>

the comment is not the thing is doing.it means to me, if there is overflow after adding 1 to the value of `connectionAttempts`,reset the value to 2( `Shadowsocks`)  otherwise keep the counter. the way we are choosing the transport type is in `connectionTransport()`.there is no need to repeat the here.
```
      **let** (partial, isOverflow) = connectionAttempts.addingReportingOverflow(1)

        // reset the value to 2 when overflowing, otherwise keep the counter

        connectionAttempts = isOverflow ? 2 : partial
```

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: 0 of 9 files reviewed, 5 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadREST/RESTTransportStrategy.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Naftrpe5BgZ-68aQIOo:-NajzHAn81Y3qCtAgy-E:bfym3in) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadREST/RESTTransportStrategy.swift#L41)):*

the comment is not the thing is doing.it means to me, if there is overflow after adding 1 to the value of `connectionAttempts`,reset the value to 2( `Shadowsocks`)  otherwise keep the counter. the way we are choosing the transport type is in `connectionTransport()`.there is no need to repeat the here.
```
      **let** (partial, isOverflow) = connectionAttempts.addingReportingOverflow(1)

        // reset the value to 2 when overflowing, otherwise keep the counter

        connectionAttempts = isOverflow ? 2 : partial
```
</blockquote></details>

We are not repeating anything here. 
This is handling the edge case scenario where we reach `Int.max` and we need to loop over. 
If we were reverting to `1` instead of `2` it would break the "every third failure loops back to trying to directly connect" algorithm.

The comment is explaining why we go directly to 2 instead of going back to 1. I will add a small addendum to make it extra clear.

mojganii (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @buggmagnet)

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: 0 of 9 files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 106 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Nafwc8o9uSQL1ydIxvN:-NakN2D75cgNbM04cAGg:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/MullvadVPN/AppDelegate.swift#L106)):*

it should be retrieved from the you storage protocol
</blockquote></details>

Done.

pronebird (changes requested):
> Reviewed 1 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: 1 of 9 files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 105 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Nakc9dL9isZkVTLNYMf:-Nakc9dL9isZkVTLNYMg:b-7wtu49) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b494a57e6adf3e654c0824639e8abd67b319e13/ios/MullvadVPN/AppDelegate.swift#L105)):*
> ```Swift
> 
>         // This init cannot fail as long as the security group identifier is valid
>         let sharedUserDefaults = UserDefaults(suiteName: ApplicationConfiguration.securityGroupIdentifier)!
> ```

I think that `TransportStrategy` could accept `UserDefaults` and use it directly to read the initial value from preferences and then update it as it goes.

Current implementation is somewhat fragmented and spread across many places in the app:

1. The user defaults key resides in `ApplicationConfiguration` and global for the app.
2. `AppDelegate` knows that it needs to read `Int` value from a specific key in order to initialize `TransportStrategy`.
3. `UserDefaults` extension that implements `AttemptsRecording` knows that it needs to store `Int` value using the specific key from `ApplicationConfiguration`.
4. `UserDefaults+Extensions` resides in `TunnelManager` which is a bit odd as it's not really used in there.

It would be much simpler if all this was isolated within `TransportStrategy` and `AppDelegate` would only care about creating `UserDefaults` instance and passing it over to `TransportStrategy`, i.e:

```swift
private static let userDefaultsKey = "<user default key>"

let userDefaults: UserDefaults
var connectionAttempts: Int

public init(userDefaults: UserDefaults) {
    connectionAttempts = userDefaults.integer(forKey: Self.userDefaultsKey)
    self.userDefaults = userDefaults
}

private func persist() {
    userDefaults.set(connectionAttempts, forKey: Self.userDefaultsKey)
}
```

For mocking, instead of `UserDefaults` you could pass some protocol which `UserDefaults` either conform to or some interstitial struct that implements it.

This way the implementation detail in regards of where and how the connection attempts are stored remains hidden within transport strategy.

pronebird (commented):
> Reviewed 6 of 9 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @mojganii)

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: 3 of 9 files reviewed, 4 unresolved discussions (waiting on @mojganii and @pronebird)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 105 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Nakc9dL9isZkVTLNYMf:-NakzZ6e2DoSvhv93GYE:buyec6o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b494a57e6adf3e654c0824639e8abd67b319e13/ios/MullvadVPN/AppDelegate.swift#L105)):*
I wanted to avoid keeping a reference to `UserDefaults` because it opens a can of worms, which is also why I created the `AttemptsRecording` abstraction. This class shouldn't know that it's backing data to `UserDefaults`. It should just "write" somewhere IMHO.

But since both you and @mojganii  suggested that we should be more straightforward in `UserDefaults` usage here, I will go with this suggestion.

For unit tests I just remembered we can still use it, and use the `removePersistentDomain(forName:)`  API in order to reset `UserDefaults` and avoid affecting other tests.

> `UserDefaults+Extensions` resides in `TunnelManager` which is a bit odd as it's not really used in there.

I wasn't entirely sure where to put it initially. With this suggestion we don't need the extension anymore.

pronebird (commented):
> Reviewed 6 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: all files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/Shared/ApplicationConfiguration.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Nafx7jd8wMHx0_WCwa_:-NalUWaT120BNpK_esUd:b-vw73ub) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/Shared/ApplicationConfiguration.swift#L41)):*

It has to be available from both the Packet Tunnel and the main application, so regardless how we cut it, it will be available to the whole app anyway, I didn't feel like creating an enum for storing a single value. Maybe in the future if we have more related settings in that way, we can regroup them together.
</blockquote></details>

With the most recent change, this key is now a part of `TransportStrategy` so we could perhaps remove it here.

mojganii (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-:-NalUkZf0Qisw5c3CO78:bnfp4nl)* status: all files reviewed, 3 unresolved discussions

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @mojganii and @pronebird)

___
*[`ios/Shared/ApplicationConfiguration.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-Nafx7jd8wMHx0_WCwa_:-Nal_gQD2LHblpJkg9eI:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/62f24f910996c8402398eac6ffd6f47b070b2982/ios/Shared/ApplicationConfiguration.swift#L41)):*

With the most recent change, this key is now a part of `TransportStrategy` so we could perhaps remove it here.
</blockquote></details>

Done.

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4921#-:-NalcaNr63k_S60hdxoW:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @mojganii)

PR #4925, opened 2023-07-25
Add an hourly device state read to update the header bar view

buggmagnet:
This PR introduces a new method in the `TunnelObserver` protocol.
`func tunnelManager(_ manager: TunnelManager, didPeriodicallyRead deviceState: DeviceState)` 

As the name implies, periodic reads from the `DeviceState` are now scheduled when the app becomes active via
`func didBecomeActive(_ notification: Notification)` 

Right now, the timer will fire once, and then every hour. It is only active when the app is running, otherwise the timer is cancelled. This is to avoid impact on the battery.

Whenever users now resume the app, the `HeaderBarView` should display the most recent information it has available.

### DIFF
--- a/ios/MullvadVPN/AppDelegate.swift
+++ b/ios/MullvadVPN/AppDelegate.swift
@@ -157,12 +157,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
     @objc private func didBecomeActive(_ notification: Notification) {
         tunnelManager.startPeriodicPrivateKeyRotation()
+        tunnelManager.startPeriodicStateRefresh()
         relayCacheTracker.startPeriodicUpdates()
         addressCacheTracker.startPeriodicUpdates()
     }
 
     @objc private func willResignActive(_ notification: Notification) {
         tunnelManager.stopPeriodicPrivateKeyRotation()
+        tunnelManager.stopPeriodicStateRefresh()
         relayCacheTracker.stopPeriodicUpdates()
         addressCacheTracker.stopPeriodicUpdates()
     }
--- a/ios/MullvadVPN/Coordinators/App/ApplicationCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/App/ApplicationCoordinator.swift
@@ -72,7 +72,7 @@ final class ApplicationCoordinator: Coordinator, Presenting, RootContainerViewCo
 
     private let apiProxy: REST.APIProxy
     private let devicesProxy: REST.DevicesProxy
-    private var tunnelObserver: TunnelObserver?
+    private var deviceStateChangeObserver: TunnelObserver?
 
     private var outOfTimeTimer: Timer?
 
@@ -756,14 +756,19 @@ final class ApplicationCoordinator: Coordinator, Presenting, RootContainerViewCo
     }
 
     private func addTunnelObserver() {
-        let tunnelObserver =
-            TunnelBlockObserver(didUpdateDeviceState: { [weak self] manager, deviceState, previousDeviceState in
-                self?.deviceStateDidChange(deviceState, previousDeviceState: previousDeviceState)
-            })
+        let deviceStateChangeObserver =
+            TunnelBlockObserver(
+                didUpdateDeviceState: { [weak self] manager, deviceState, previousDeviceState in
+                    self?.deviceStateDidChange(deviceState, previousDeviceState: previousDeviceState)
+                },
+                didPeriodicallyReadDeviceState: { [weak self] _, deviceState in
+                    self?.updateDeviceInfo(deviceState: deviceState)
+                }
+            )
 
-        tunnelManager.addObserver(tunnelObserver)
+        tunnelManager.addObserver(deviceStateChangeObserver)
 
-        self.tunnelObserver = tunnelObserver
+        self.deviceStateChangeObserver = deviceStateChangeObserver
 
         updateDeviceInfo(deviceState: tunnelManager.deviceState)
 
--- a/ios/MullvadVPN/TunnelManager/TunnelBlockObserver.swift
+++ b/ios/MullvadVPN/TunnelManager/TunnelBlockObserver.swift
@@ -18,25 +18,29 @@ final class TunnelBlockObserver: TunnelObserver {
     ) -> Void
     typealias DidUpdateTunnelSettingsHandler = (TunnelManager, TunnelSettingsV2) -> Void
     typealias DidFailWithErrorHandler = (TunnelManager, Error) -> Void
+    typealias DidPeriodicallyReadDeviceStateHandler = (TunnelManager, DeviceState) -> Void
 
     private let didLoadConfiguration: DidLoadConfigurationHandler?
     private let didUpdateTunnelStatus: DidUpdateTunnelStatusHandler?
     private let didUpdateDeviceState: DidUpdateDeviceStateHandler?
     private let didUpdateTunnelSettings: DidUpdateTunnelSettingsHandler?
     private let didFailWithError: DidFailWithErrorHandler?
+    private let didPeriodicallyReadDeviceState: DidPeriodicallyReadDeviceStateHandler?
 
     init(
         didLoadConfiguration: DidLoadConfigurationHandler? = nil,
         didUpdateTunnelStatus: DidUpdateTunnelStatusHandler? = nil,
         didUpdateDeviceState: DidUpdateDeviceStateHandler? = nil,
         didUpdateTunnelSettings: DidUpdateTunnelSettingsHandler? = nil,
-        didFailWithError: DidFailWithErrorHandler? = nil
+        didFailWithError: DidFailWithErrorHandler? = nil,
+        didPeriodicallyReadDeviceState: DidPeriodicallyReadDeviceStateHandler? = nil
     ) {
         self.didLoadConfiguration = didLoadConfiguration
         self.didUpdateTunnelStatus = didUpdateTunnelStatus
         self.didUpdateDeviceState = didUpdateDeviceState
         self.didUpdateTunnelSettings = didUpdateTunnelSettings
         self.didFailWithError = didFailWithError
+        self.didPeriodicallyReadDeviceState = didPeriodicallyReadDeviceState
     }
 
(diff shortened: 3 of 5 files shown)


mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 68 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925#-NaGFclN91-tlNKVTv0B:-NaGFclN91-tlNKVTv0C:bl9cp76) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/de6114747b6f242e95f84227de1fd4dc0d79388d/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L68)):*
> ```Swift
>     private var deviceStateRefreshTimer: DispatchSourceTimer?
>     private var isRunningPeriodicStateRefresh = false
>     private let deviceStateRefreshPeriod: TimeInterval = 60 * 60
> ```

Did you get confirmation about the interval? it seems to me it's a little bit long and the result of that doesn't effects on having better user experience 
___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 136 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925#-NaGGOpA5yrq9l48AoiP:-NaGGOpA5yrq9l48AoiQ:bm5gk6u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/de6114747b6f242e95f84227de1fd4dc0d79388d/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L136)):*
> ```Swift
>     }
> 
>     func startPeriodicStateRefresh() {
> ```

nit:   StartPeriodicDeviceStateRefresh(). when it comes to `state` it's not well defined what it's talked about.
___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 154 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925#-NaGHhB83zbGWBmgTUXn:-NaGHhB83zbGWBmgTUXo:b-v2j4vc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/de6114747b6f242e95f84227de1fd4dc0d79388d/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L154)):*
> ```Swift
>     }
> 
>     func stopPeriodicStateRefresh() {
> ```

nit:   StopPeriodicDeviceStateRefresh()
___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 941 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925#-NaGDhaqDmcypTiPETlK:-NaGDhaqDmcypTiPETlL:bwmh50g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/de6114747b6f242e95f84227de1fd4dc0d79388d/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L941)):*
> ```Swift
>             guard let deviceState = try? SettingsManager.readDeviceState(),
>                   let self = self else { return }
>             nslock.withLock {
> ```

do we need to do this part of the code with locking while this method is not accessible from others ?

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 68 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925#-NaGFclN91-tlNKVTv0B:-NaGNN594nlRa_CFIXLo:b-wprsng) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/de6114747b6f242e95f84227de1fd4dc0d79388d/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L68)):*

Did you get confirmation about the interval? it seems to me it's a little bit long and the result of that doesn't effects on having better user experience 
</blockquote></details>

It's by design. The timer will fire immediately when the application re enters foreground which is the desired effect -- refreshing the header bar view to display accurate information.

I don't think people spend hours showing the UI of the application, so this is high enough as to not impact battery life by firing too often. 
___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 136 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925#-NaGGOpA5yrq9l48AoiP:-NaGPiiz8bjQFrwip4Ju:b5uc3r8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/de6114747b6f242e95f84227de1fd4dc0d79388d/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L136)):*

nit:   StartPeriodicDeviceStateRefresh(). when it comes to `state` it's not well defined what it's talked about.
</blockquote></details>

Fair enough, I will also remove the `refresh` work as we already use it for a different purposes.
Since we only read the device state, I will rename this "Readings" 
___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 941 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925#-NaGDhaqDmcypTiPETlK:-NaGPcyP3ORap_gevU4E:bqn9kc2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/de6114747b6f242e95f84227de1fd4dc0d79388d/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L941)):*

do we need to do this part of the code with locking while this method is not accessible from others ?
</blockquote></details>

You're right, the lock is not needed here, thanks for catching that I will remove it.

mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 68 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925#-NaGFclN91-tlNKVTv0B:-NaGTJfJ9i33fAdKZZ7N:b8zl5jx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/de6114747b6f242e95f84227de1fd4dc0d79388d/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L68)):*

It's by design. The timer will fire immediately when the application re enters foreground which is the desired effect -- refreshing the header bar view to display accurate information.

I don't think people spend hours showing the UI of the application, so this is high enough as to not impact battery life by firing too often. 
</blockquote></details>

I am wondering if user buys more credit via other platforms while the app is active,our app will be notified quickly or user has to be wait for fire time?

mojganii (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925)* status: 0 of 5 files reviewed, all discussions resolved

pronebird (changes requested):
> Reviewed 3 of 5 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925#-NakPUP_1DiOSyQl6GPc:-NakPUPa5KeLZ8DkgrQG:b-czaecg) (no related file):*
I think we already handle the device state refresh. This is the logic that's currently in place:

- When the app becomes active we call `TunnelManager.refreshDeviceState()` in tunnel manager, that makes sure to read new state from settings and propagate it to observers. (`TunnelManager.setDeviceState()` does the caching, persistence and propagation part).
- We poll tunnel status periodically to obtain the connection related info, but also last device check (see `TunnelManager.setTunnelStatus()`). Based on the last and new device check received from packet tunnel, tunnel manager infers whether it needs to refresh device state from settings or switch device to revoked state or any other.


Since we already implement the refresh of device state, I think the only missing piece here is to implement a timer in `HeaderBarView` that would re-format the label every once in a while. The date assigned to `HeaderBar.timeLeft` is probably up to date, but the `timeLeftLabel.text` is only assigned on `didSet`.

___

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/Containers/Root/HeaderBarView.swift` line 250 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925#-NalNdUCAEWJMJeLcJ6T:-NalNdUCAEWJMJeLcJ6U:b-9uzx37) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c534dd2e8e95d64152ccd9cebb7a4cff92a5c94d/ios/MullvadVPN/Containers/Root/HeaderBarView.swift#L250)):*
> ```Swift
>         timeLeftTimer = Timer.scheduledTimer(withTimeInterval: timeLeftUpdateInterval, repeats: true) { _ in
>             DispatchQueue.main.async { [weak self] in
>                 self?.timeLeft = self?.timeLeft
> ```

Couple of things:

1. I know it works but it would be cleaner to extract the `didSet` from `var timeLeft: Date?` into a separate function and call it from here and from there instead to keep it straightforward what we're trying to achieve.
2. Block timer will retain `self` strongly before retaining it weak before passing it into `DispatchQueue.main.async {}`. That's because strong is default I think.
3. Repeating `Timer` still needs to be invalidated from `deinit` because `RunLoop` holds a strong reference to it once the timer is scheduled.
4. Perhaps you don't need to wrap the timer handler into `DispatchQueue.main` because the timer is scheduled on current runloop which is `RunLoop.main` so the timer handler block should execute on main thread.

pronebird (approved):
:lgtm:

> Reviewed 6 of 6 files at r3, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4925#-:-NalV7WmE1vLGD7_r7rU:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

PR #4937, opened 2023-08-01
Reduce duplicate windows-sys versions

faern:
We have loads of dependencies where we pull in multiple versions. It's kind of impossible to not have that in a mid-sized Rust project. But at least some work can be put into somewhat reducing the numbers. This PR upgrades some of our dependencies.

The result is that we at least get rid of one (out of four!) versions of `windows-sys` from our dependency tree. We also gain and lose some other dependencies. The net change is fewer dependencies at least.

I also created a PR on our only dependency pulling in `windows-sys 0.36`. When that is merged and released we might get rid of another instance. https://github.com/zonyitoo/tokio-tfo/pull/6

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4937)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1091,7 +1091,7 @@ checksum = "5cbc844cecaee9d4443931972e1289c8ff485cb4cc2767cb03ca139ed6885153"
 dependencies = [
  "cfg-if",
  "libc",
- "redox_syscall",
+ "redox_syscall 0.2.10",
  "windows-sys 0.48.0",
 ]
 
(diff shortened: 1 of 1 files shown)


MarkusPettersson98 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4937#-:-Nakr2nlBYEn6-7s3qmE:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4938, opened 2023-08-01
Remove dependency on `time` (disable `chrono` features)

faern:
When we decided to ignore `RUSTSEC-2020-0071` we apparently draw the conclusion that it was not possible to get rid of `time`. I'm not going to investigate very deeply what has changed or why we reached that conclusion. But it seems like we can without trouble disable the `oldtime` feature of `chrono` and as a result get rid of `time 0.1` from our dependency tree :partying_face: 

It's sad that this is a default feature, requiring us to add `default-features = false` *everywhere* (default features was a mistake, change my mind)


MarkusPettersson98 (approved):
:lgtm:

> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4938#-:-NakvN7i-1iU92lQkz8j:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4943, opened 2023-08-01
Move windows-sys to a workspace dependency

faern:
Sort of a continuation of #4902. `windows-sys` is one of those dependencies that we depend in in a massive amount of crates, but we really only ever want to pull in a single version, and there are semver incompatible releases fairly often. So it makes a lot of sense to make this a workspace dependency. Will making upgrading to version `0.48` and later versions way easier later.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4943)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,9 @@ tower = "0.4"
 prost = "0.11"
 prost-types = "0.11"
 
+windows-sys = "0.45.0"
+
+
 [profile.release]
 opt-level = 3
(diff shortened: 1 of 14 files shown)


MarkusPettersson98 (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4943)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4954, opened 2023-08-02
Migrate translation-converter to err_derive

faern:
On my dependency excursion lately I found out that we use `derive_more` in `translations-converter` to derive the standard `Error` macro for us. We do the same but with the `err_derive` crate everywhere else. So I figured if I migrate this tool to also use that things would be more uniform and we would drop a few dependencies along the way.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4954)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -566,12 +566,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "13418e745008f7349ec7e449155f419a61b92b58a99cc3616942b926825ec76b"
 
 [[package]]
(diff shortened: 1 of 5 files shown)


MarkusPettersson98 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4954#-:-NaqL6N15OfiZD1xBNTq:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4947, opened 2023-08-01
Remove unused talpid core dependencies

faern:
I went through all dependencies of `talpid-core` and checked each of them if they were actually used in the code or not. Found *a lot* of unused dependencies. I suspect this is mostly from #4041 and was forgotten in that PR? (ping @pinkisemils).

This does not remove a single dependency from our dependency tree, but it sure does simplify `taplid-core` by a significant amount.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4947)

PR #4956, opened 2023-08-02
Use std::ffi::c_void in Rust<->iOS code instead of libc::c_void

faern:
On my recent excursion hunting down unused Rust dependencies (https://github.com/mullvad/mullvadvpn-app/pull/4953 & https://github.com/mullvad/mullvadvpn-app/pull/4947) I realized that both these iOS Rust packages depend on `libc` solely for the `c_void` definition. This type is available in the standard library since a long time now. So I figured we could use those and make things just a tad bit simpler/cleaner.

I have not tested this change. But these types should be equivalent. On Rust >=1.30 libc just re-exports the standard library version of it, so they are the same. https://github.com/rust-lang/libc/blob/main/build.rs#L109-L114 + https://github.com/rust-lang/libc/blob/main/src/unix/mod.rs#L1603-L1604

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3366,7 +3366,6 @@ name = "shadowsocks-proxy"
 version = "0.0.0"
 dependencies = [
  "cbindgen",
- "libc",
  "log",
  "oslog",
  "shadowsocks-service",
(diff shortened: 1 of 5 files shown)


pronebird (approved):
:lgtm: please sign your commit.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4956#-:-NauoZmJAuJcP_YQQzqh:b-ibpf4q)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4957, opened 2023-08-02
Use C FFI types from `std::ffi` instead of `libc`

faern:
This PR is related to #4956. In that PR I managed to actually get rid of a dependency because of the migration. That's not the case in this PR. But it might be nice to be consistent with what FFI types we use. Since Rust 1.64 the `std::ffi` module was expanded with char and int type definitions.

The motivation for this PR is just a tiny cleanup. But it has no known downsides either afaict, so why not.

If you approve of this, I might add to our coding guidelines to prefer `std::ffi` where possible. Most packages will depend on `libc` in one way or another anyway. But why pollute `Cargo.toml` for no reason.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4957)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3772,7 +3772,6 @@ version = "0.0.0"
 dependencies = [
  "err-derive",
  "futures",
- "libc",
  "socket2 0.4.9",
  "talpid-types",
  "winapi",
--- a/mullvad-daemon/src/exception_logging/unix.rs
+++ b/mullvad-daemon/src/exception_logging/unix.rs
@@ -1,8 +1,9 @@
 //! Installs signal handlers to catch critical program faults and logs them.
 
-use libc::{c_int, c_void, siginfo_t};
+use libc::{c_int, siginfo_t};
 use nix::sys::signal::{sigaction, SaFlags, SigAction, SigHandler, SigSet, Signal};
 
+use std::ffi::c_void;
 use std::{convert::TryFrom, sync::Once};
 
 static INIT_ONCE: Once = Once::new();
--- a/mullvad-daemon/src/exception_logging/unix.rs
+++ b/mullvad-daemon/src/exception_logging/unix.rs
@@ -1,9 +1,9 @@
 //! Installs signal handlers to catch critical program faults and logs them.
 
-use libc::{c_int, siginfo_t};
+use libc::siginfo_t;
 use nix::sys::signal::{sigaction, SaFlags, SigAction, SigHandler, SigSet, Signal};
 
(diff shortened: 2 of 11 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4957)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`mullvad-daemon/src/macos_launch_daemon.rs` line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4957#-NbEkJlc9-GmNCF-1EH-:-NbFmvWg4kx3Z0fb4pX-:b-i9h4w5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cfc2b387c5ca1a4ad36b18f6733f2eb48834f22a/mullvad-daemon/src/macos_launch_daemon.rs#L83)):*

OOS, but this looks a bit suspicious:

> `*const c_void` is equivalent to C’s `const void*` and `*mut c_void` is equivalent to C’s `void*`. That said, this is _not_ the same as C’s `void` return type, which is Rust’s `()` type.

I wonder if the type should be `()` rather than `c_void`.
</blockquote></details>

At worst, it's probably just copying some garbage from the return-value register? I don't think it could mess up the stack.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4957)* status: 5 of 11 files reviewed, all discussions resolved (waiting on @dlon)

___
*[`mullvad-daemon/src/macos_launch_daemon.rs` line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4957#-NbEkJlc9-GmNCF-1EH-:-NbIze3B5S1ldPU2aI-k:bcf5u3i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cfc2b387c5ca1a4ad36b18f6733f2eb48834f22a/mullvad-daemon/src/macos_launch_daemon.rs#L83)):*
"At worst just read and copy around random memory" -  sounds like a pretty terrible place to be.

The docs for `objc` uses `()` for void return types:

> // Even void methods must have their return type annotated
> let \_: () \= msg\_send!\[obj, release\];

So yeah. Sounds like we should change that. Fixed.

dlon (approved):
> Reviewed 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4957)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4966, opened 2023-08-03
Upgrade `windows-sys` to 0.48

faern:
The `windows-sys` ecosystem moves along. A lot of our dependencies already use version 0.48. It was about time to upgrade. I'm not the most familiar with the win32 APIs, but I did my best.

This is mostly constants changing which module they are defined in.

But then it's also the fact that a lot of fallible functions now return `WIN32_ERROR` (== `u32`) instead of the previous `NTSTATUS` (== `i32`). We used this return code to instantiate `io::Error`s. But the constructor `io::Error::from_raw_os_error` takes a `i32` and we now have a `u32`. The simplest thing would just be to cast it. But I don't know if any of these Windows functions could return an error code `> i32::MAX`. If they ever do we will cast it to an invalid value and get a bogus error. That's fine maybe. But I thought it was also easily fixable.

First I just implemented an extension trait on `io::Error` with a constructor method `from_raw_win32_error(u32) -> Self`. But I then also thought we did quite a lot of boilerplate of the sort:
```rust
let result = unsafe {SomeWindowsFunctionW(...)};
if result != NO_ERROR {
    return Err(...);
}
```
Since all of them just want to early return on the same constant I thought I could remove a lot of code by wrapping this up in a handy macro (`win32_err!`). Tell me what you think about this. I personally think this provides more ergonomics than it introduces magic.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1968,7 +1968,7 @@ dependencies = [
  "natord",
  "talpid-types",
  "tokio",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
  "winres",
 ]
 
@@ -2012,7 +2012,7 @@ dependencies = [
  "tokio-stream",
  "winapi",
  "windows-service",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
  "winres",
 ]
 
@@ -2093,7 +2093,7 @@ dependencies = [
  "log",
  "once_cell",
  "widestring",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -2115,7 +2115,7 @@ dependencies = [
  "talpid-types",
  "tokio",
  "uuid",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
  "winres",
 ]
 
@@ -3606,7 +3606,7 @@ dependencies = [
  "which",
  "widestring",
  "windows-service",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
  "winreg 0.7.0",
 ]
 
@@ -3650,7 +3650,7 @@ dependencies = [
  "triggered",
  "uuid",
  "widestring",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
  "winreg 0.7.0",
 ]
 
@@ -3670,7 +3670,7 @@ dependencies = [
  "tonic",
  "tonic-build",
  "tower",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
  "winres",
 ]
 
@@ -3680,7 +3680,7 @@ version = "0.0.0"
 dependencies = [
  "rs-release",
  "talpid-dbus",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -3703,7 +3703,7 @@ dependencies = [
  "talpid-windows-net",
  "tokio",
  "widestring",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -3731,7 +3731,7 @@ dependencies = [
  "talpid-windows-net",
  "tokio",
  "tun",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -3749,7 +3749,7 @@ dependencies = [
  "tonic",
  "tonic-build",
  "tower",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
  "zeroize",
 ]
 
@@ -3775,7 +3775,7 @@ dependencies = [
  "libc",
  "socket2 0.4.9",
  "winapi",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -3813,7 +3813,7 @@ dependencies = [
  "tokio-stream",
  "tunnel-obfuscation",
  "widestring",
- "windows-sys 0.45.0",
+ "windows-sys 0.48.0",
  "zeroize",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3774,6 +3774,7 @@ dependencies = [
  "futures",
  "libc",
  "socket2 0.4.9",
+ "talpid-types",
  "winapi",
  "windows-sys 0.48.0",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ tower = "0.4"
 prost = "0.11"
 prost-types = "0.11"
 
-windows-sys = "0.45.0"
+windows-sys = "0.48.0"
 
 chrono = { version = "0.4.26", default-features = false}
 clap = { version = "4.2.7", features = ["cargo", "derive"] }
--- a/mullvad-paths/src/windows.rs
+++ b/mullvad-paths/src/windows.rs
@@ -5,8 +5,8 @@ use windows_sys::{
     core::{GUID, PWSTR},
     Win32::{
         Foundation::{
-            CloseHandle, ERROR_INSUFFICIENT_BUFFER, ERROR_SUCCESS, HANDLE, INVALID_HANDLE_VALUE,
-            LUID, S_OK,
+            CloseHandle, ERROR_INSUFFICIENT_BUFFER, ERROR_SUCCESS, GENERIC_READ, HANDLE,
+            INVALID_HANDLE_VALUE, LUID, S_OK,
         },
         Security::{
             AdjustTokenPrivileges, CreateWellKnownSid, EqualSid, GetTokenInformation,
@@ -17,8 +17,7 @@ use windows_sys::{
         Storage::FileSystem::MAX_SID_SIZE,
         System::{
             Com::CoTaskMemFree,
-            ProcessStatus::K32EnumProcesses,
-            SystemServices::GENERIC_READ,
+            ProcessStatus::EnumProcesses,
             Threading::{
                 GetCurrentThread, OpenProcess, OpenProcessToken, OpenThreadToken,
                 PROCESS_QUERY_INFORMATION,
@@ -209,9 +208,7 @@ fn find_process<T>(handle_process: impl Fn(HANDLE) -> Option<T>) -> io::Result<T
 
     let bytes_available = num_procs * (mem::size_of::<u32>() as u32);
     let mut bytes_written = 0;
-    if unsafe { K32EnumProcesses(pid_buffer.as_mut_ptr(), bytes_available, &mut bytes_written) }
-        == 0
-    {
+    if unsafe { EnumProcesses(pid_buffer.as_mut_ptr(), bytes_available, &mut bytes_written) } == 0 {
         return Err(io::Error::last_os_error());
     }
 
--- a/talpid-core/src/dns/windows/auto.rs
+++ b/talpid-core/src/dns/windows/auto.rs
@@ -82,7 +82,7 @@ impl DnsMonitor {
     fn fallback_due_to_dnscache(&mut self, result: &Result<(), super::Error>) -> bool {
         let is_dnscache_error = match result {
             Err(super::Error::Iphlpapi(iphlpapi::Error::SetInterfaceDnsSettings(error))) => {
-                *error == RPC_S_SERVER_UNAVAILABLE
+                error.raw_os_error() == Some(RPC_S_SERVER_UNAVAILABLE)
             }
             Err(super::Error::Netsh(netsh::Error::Netsh(Some(1)))) => true,
             _ => false,
--- a/talpid-core/src/dns/windows/iphlpapi.rs
+++ b/talpid-core/src/dns/windows/iphlpapi.rs
@@ -7,12 +7,13 @@ use std::{
     os::windows::ffi::OsStrExt,
     ptr,
 };
+use talpid_types::win32_err;
 use talpid_windows_net::{guid_from_luid, luid_from_alias};
 use windows_sys::{
     core::GUID,
     s, w,
     Win32::{
-        Foundation::{ERROR_PROC_NOT_FOUND, NO_ERROR, NTSTATUS},
+        Foundation::{ERROR_PROC_NOT_FOUND, WIN32_ERROR},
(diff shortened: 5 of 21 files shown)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966)* status: 0 of 18 files reviewed, 4 unresolved discussions

___
*[`talpid-routing/src/windows/default_route_monitor.rs` line 148 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-Nav4iCSEcAtd2e61H40:-Nav4iCTC9aTsmtM04OH:bstbh9o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96a3af542ffd61acd67507c44e5b71a54a2d9693/talpid-routing/src/windows/default_route_monitor.rs#L148)):*
> ```XML
>             // If this callback is called after we free the context that could result in UB, in
>             // order to avoid that we panic.
>             panic!("Could not cancel change notification callback: {}", e)
> ```

Here we get the benefit of actually including the human facing error message in the panic.
___
*[`talpid-routing/src/windows/get_best_default_route.rs` line 48 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-Nav4q2S495ggzctnBcy:-Nav4q2T1XrSLwE9J_YR:b-sskxa2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96a3af542ffd61acd67507c44e5b71a54a2d9693/talpid-routing/src/windows/get_best_default_route.rs#L48)):*
> ```XML
>     // contain pointers to the table so they will not be dangling after this free.
>     if let Err(e) = win32_err!(unsafe { FreeMibTable(table_ptr as *const _) }) {
>         log::error!("FreeMibTable should not return a value, but did: {e}");
> ```

This is a weird one. `FreeMibTable` does not have a return value according to the Microsoft docs. But the `windows-sys` bindings went from `-> ()` to `-> WIN32_ERROR`. I guess this can't really happen? But it did not hurt to add some logging, so I did.
___
*[`talpid-routing/src/windows/route_manager.rs` line 270 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-Nav5UJF0R_4zDNnRkiD:-Nav5UJGD45xgMcKzOIi:bqauii3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96a3af542ffd61acd67507c44e5b71a54a2d9693/talpid-routing/src/windows/route_manager.rs#L270)):*
> ```XML
>                             }) {
>                                 log::error!("Unable to get interface LUID for interface \"{device_name:?}\": {e}");
>                                 return Err(Error::DeviceNameNotFound);
> ```

Both logging and returning the same error is usually not ideal. Depending on how the caller handles the error, it quite often just ends up in the logs twice. But here I opted to keep it anyway since I'm not fully familiar with the code.
___
*[`talpid-routing/src/windows/route_manager.rs` line 394 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-Nav5hRf7fiVEvOvHO48:-Nav5hRf7fiVEvOvHO49:bgd0n5s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96a3af542ffd61acd67507c44e5b71a54a2d9693/talpid-routing/src/windows/route_manager.rs#L394)):*
> ```XML
>         win32_err!(unsafe { CreateIpForwardEntry2(&spec) }).map_err(|e| {
>             log::error!("Could not register route in routing table. Route: {route}");
>             Error::AddToRouteTable(e)
> ```

Here I opted to stop including the status in the logging. For the same reason as above (don't both log and return the same error). I see that the caller logs it anyway. And now they get a nice error message instead of just the error code.
___
*[`talpid-types/src/error.rs` line 61 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-Nav5rZV9-9MgpYXMVWU:-Nav5rZWBlj7HrlH8f6B:b6frkfi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96a3af542ffd61acd67507c44e5b71a54a2d9693/talpid-types/src/error.rs#L61)):*
> ```XML
> /// return `ErrorKind::Other` instead.
> ///
> /// The caller of this macro must have `windows_sys` as a dependency.
> ```

I could also depend on `windows-sys` here in `talpid-types` and use `$crate::windows_sys::..::NO_ERROR`. But I'm not sure if that helps in any way. Anyone who calls this macro must depend on `windows-sys` anyway, because the `$expr` is expected to be a Windows FFI function.

With this implementation we are now using the `NO_ERROR` definition of the `windows-sys` version in the *caller*. If, for some reason there are multiple callers to this macro and they depend on different versions of `windows-sys`, they could theoretically have different values in `NO_ERROR`, and then it makes sense to check **their** `$expr` code against **their** `NO_ERROR` constant. This will never happen, more of a hypothetical benefit.
___
*[`talpid-wireguard/Cargo.toml` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-Nav6bWH3YWMAKmkysHK:-Nav6bWI35XWpjtaf9vf:blmn2ua) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96a3af542ffd61acd67507c44e5b71a54a2d9693/talpid-wireguard/Cargo.toml#L26)):*
> ```TOML
> talpid-tunnel = { path = "../talpid-tunnel" }
> zeroize = "1"
> chrono = { workspace = true, features = ["clock"] }
> ```

Ooops. Not related to this PR, but a miss I made earlier. Might put this in a separate tiny PR and get it merged asap.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966)* status: 0 of 19 files reviewed, 5 unresolved discussions

___
*[`mullvad-paths/src/windows.rs` line 212 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-NavI4Kw-3s7wx2q6r6u:-NavI4Kw-3s7wx2q6r6v:bncvubk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b938a36a2f21473245c963cbb3c6c6be719945/mullvad-paths/src/windows.rs#L212)):*
> ```XML
>     let bytes_available = num_procs * (mem::size_of::<u32>() as u32);
>     let mut bytes_written = 0;
>     if unsafe { K32EnumProcesses(pid_buffer.as_mut_ptr(), bytes_available, &mut bytes_written) }
> ```

This change apparently causes issues:
```
mullvad\_nsis.lib(mullvad\_nsis-ebc56842bce0b2e4.mullvad\_nsis.50497f40892b68ca-cgu.3.rcgu.o) : error LNK2001: unresolved external symbol \_\_imp\_\_EnumProcesses@12 \[D:\\a\\mullvadvpn-app\\mullvadvpn-app\\windows\\nsis-plugins\\src\\cleanup\\cleanup.vcxproj\]
```
source: https://github.com/mullvad/mullvadvpn-app/actions/runs/5749967632/job/15585800262?pr=4966#step:10:651

But `windows-sys` no longer exports anything with the name `K32EnumProcesses`.

dlon (commented):
> If they ever do we will cast it to an invalid value and get a bogus error.

Are you sure about this? Casting `-1i32` to a `u32` and back to `i32` preserves the bits, so it should be fine. I'm not sure if this is defined behavior, though.

> Reviewed 17 of 18 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-:-NbE7Xgo7ICwQYO0hzZu:b-mt07d8)* status: 18 of 19 files reviewed, 4 unresolved discussions (waiting on @faern)

___
*[`talpid-routing/src/windows/get_best_default_route.rs` line 48 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-Nav4q2S495ggzctnBcy:-NbEAyo-564uvXbQTygN:bx5n0n0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96a3af542ffd61acd67507c44e5b71a54a2d9693/talpid-routing/src/windows/get_best_default_route.rs#L48)):*

This is a weird one. `FreeMibTable` does not have a return value according to the Microsoft docs. But the `windows-sys` bindings went from `-> ()` to `-> WIN32_ERROR`. I guess this can't really happen? But it did not hurt to add some logging, so I did.
</blockquote></details>

This does look suspect. If the signature is simply wrong, can we trust the return value? I think we should perhaps just ignore it.
___
*[`talpid-types/src/error.rs` line 61 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-Nav5rZV9-9MgpYXMVWU:-NbEE1nmEcJex3pOtmGV:b-kzv3x9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96a3af542ffd61acd67507c44e5b71a54a2d9693/talpid-types/src/error.rs#L61)):*

I could also depend on `windows-sys` here in `talpid-types` and use `$crate::windows_sys::..::NO_ERROR`. But I'm not sure if that helps in any way. Anyone who calls this macro must depend on `windows-sys` anyway, because the `$expr` is expected to be a Windows FFI function.

With this implementation we are now using the `NO_ERROR` definition of the `windows-sys` version in the *caller*. If, for some reason there are multiple callers to this macro and they depend on different versions of `windows-sys`, they could theoretically have different values in `NO_ERROR`, and then it makes sense to check **their** `$expr` code against **their** `NO_ERROR` constant. This will never happen, more of a hypothetical benefit.
</blockquote></details>

I think that this is fine. Beats depending on `windows-sys` unnecessarily.
___
*[`talpid-types/src/error.rs` line 64 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-NbE3D-J-Zm-8YI3QHul:-NbE3D-J-Zm-8YI3QHum:bxfzgrm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cc16d9923f4212ab623acb4089746360843c4e33/talpid-types/src/error.rs#L64)):*
> ```XML
> #[cfg(windows)]
> #[macro_export]
> macro_rules! win32_err {
> ```

Great cleanup.

faern (commented):
You are probably right. Upon digging into `from_raw_os_error` a bit it seems they cast this status code between a myriad of integer types. They probably only care about the bits in it, not the actual integer value. The docs for the function even mentions using `as i32`: https://doc.rust-lang.org/stable/std/io/type.RawOsError.html

I simplified the `win32_err` macro to simply cast to `i32`.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-:-NbEJurJ6QLJM-gQBpkl:bxbpnb1)* status: 12 of 19 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[`talpid-routing/src/windows/get_best_default_route.rs` line 48 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-Nav4q2S495ggzctnBcy:-NbEJL-jBCH_8UiiM0qA:b-ufw9z8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96a3af542ffd61acd67507c44e5b71a54a2d9693/talpid-routing/src/windows/get_best_default_route.rs#L48)):*

This does look suspect. If the signature is simply wrong, can we trust the return value? I think we should perhaps just ignore it.
</blockquote></details>

The `windows-sys` crate is generated from Microsofts own definitions. So either the MSDN docs is wrong, or the `windows-sys` definition mapper has a bug. We could also just ignore it if you think that is better.

dlon (approved):
:lgtm:

> Reviewed 1 of 18 files at r1, 5 of 5 files at r3, 5 of 5 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-:-NbF9CzH0ggEgpNAADyb:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`mullvad-paths/src/windows.rs` line 212 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-NavI4Kw-3s7wx2q6r6u:-NbF8ygkC_yntqKRdJ2U:b-wrqbf6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b938a36a2f21473245c963cbb3c6c6be719945/mullvad-paths/src/windows.rs#L212)):*

This change apparently causes issues:
```
mullvad\_nsis.lib(mullvad\_nsis-ebc56842bce0b2e4.mullvad\_nsis.50497f40892b68ca-cgu.3.rcgu.o) : error LNK2001: unresolved external symbol \_\_imp\_\_EnumProcesses@12 \[D:\\a\\mullvadvpn-app\\mullvadvpn-app\\windows\\nsis-plugins\\src\\cleanup\\cleanup.vcxproj\]
```
source: https://github.com/mullvad/mullvadvpn-app/actions/runs/5749967632/job/15585800262?pr=4966#step:10:651

But `windows-sys` no longer exports anything with the name `K32EnumProcesses`.
</blockquote></details>

Fixed the linker issue.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`mullvad-paths/src/windows.rs` line 212 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966#-NavI4Kw-3s7wx2q6r6u:-NbFNUei1XBZfQDBpbYV:beivuqi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b938a36a2f21473245c963cbb3c6c6be719945/mullvad-paths/src/windows.rs#L212)):*

Fixed the linker issue.
</blockquote></details>

Awesome! :100:

dlon (approved):
> Reviewed 1 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966)* status: :shipit: complete! all files reviewed, all discussions resolved

dlon (approved):
> Reviewed 1 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4966)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4976, opened 2023-08-04
Make every crate compile on its own

faern:
With this handly little "oneliner" I was able to automatically check every crate independently if it would build or not:
```
time for package in $(cargo metadata --format-version 1 | jq '.workspace_members[]' | grep -P -o '(?<=path\+file://).*(?=\)\")'); do (cd $package; echo "Building $package"; cargo build); done
```
This allowed me to find three crates that did not currently build. All of them because of missing features in dependency specifications.

Building every crate independently takes much longer than just `cargo build` directly in the workspace. So I'm not going to make the CI do this on every PR. This is mostly good hygiene, and not strictly necessary.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,7 @@ prost-types = "0.11"
 windows-sys = "0.45.0"
 
 chrono = { version = "0.4.26", default-features = false}
+clap = { version = "4.2.7", features = ["cargo", "derive"] }
 
 
 [profile.release]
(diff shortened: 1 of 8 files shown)


MarkusPettersson98 (approved):
Very nice!
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4976#-:-NbDg80E4h1AhDn0rMBA:b-lr7p9l)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4985, opened 2023-08-07
Simplify dnsapi module

dlon:
Using run-time linking is overly defensive here. It seems very unlikely that this function will be removed, even though it's not documented.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4985)


faern (approved):
If you think this function will always be available, I'm fine with this simplification :+1:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4985#-:-NbFTPY52W7NfycygKK4:bc7gx4)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #4995, opened 2023-08-08
Remove wireguard-go on Windows

dlon:
It could still be manually enabled from the CLI, but the fallback was already gone. This PR removes the remaining parts.

As of this PR, go and mingw are no longer required to build the app on Windows.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995)


faern (commented):
> Reviewed 22 of 31 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995)* status: 22 of 31 files reviewed, 5 unresolved discussions (waiting on @dlon)

___
*[`talpid-wireguard/src/lib.rs` line 775 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbPz8pN9m5It9Ydm_Ec:-NbPz8pN9m5It9Ydm_Ed:b8xc97q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0be47e9efe223a0d15707856015db980c0491d76/talpid-wireguard/src/lib.rs#L775)):*
> ```XML
>         }
> 
>         #[cfg(unix)]
> ```

A bit of a nitpick. But I actually think `#[cfg(not(windows))]` is better than `#[cfg(unix)]`. The very best, but most verbose and maybe overkill might even be to always list the full list of exact platforms where we want this code. That's more explicitly telling the reader where we want this code to run. But `not(windows)` is at least clear that it's "all platforms we support except Windows". Whereas `unix` to me only reads as "all Unix platforms" which is more vague. Is Android a unix platform? Not 100% clear IMO. Any other non-unix platforms where this is not supposed to run?

Since the cfg just above this is `cfg(target_os = "windows")` I think this is not symmetrical. Since it's basically an if-else situation I think they should have the same cfg-logic but inverted. Or, consider using `cfg_if!` to make that if-else relationship clearer.
___
*[`talpid-wireguard/src/logging.rs` line 49 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbQ-b0YEw0fOj3PoNwg:-NbQ-b0YEw0fOj3PoNwh:be5uknn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0be47e9efe223a0d15707856015db980c0491d76/talpid-wireguard/src/logging.rs#L49)):*
> ```XML
> }
> 
> #[allow(dead_code)]
> ```

Is this still needed? I don't know where/why this is dead code. But if it's still needed on some platform, can that be highlighted with `cfg_attr(..., allow(dead_code))`?
___
*[`talpid-wireguard/src/logging.rs` line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbQ-0B5An10PZwBoxTr:-NbQ-0B5An10PZwBoxTs:bwbdfq5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0be47e9efe223a0d15707856015db980c0491d76/talpid-wireguard/src/logging.rs#L94)):*
> ```XML
> 
> // Callback that receives messages from WireGuard
> #[cfg(unix)]
> ```

Same as the other comment I left. I honestly think `#[cfg(unix)]` is too vague here/not ideal. What do you think about `cfg(not(windows))`? Do you consider it worse for some reason?
___
*[`talpid-wireguard/src/stats.rs` line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbQ0DTq35CQfkNQPluO:-NbQ0DTq35CQfkNQPluP:bkzitcp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0be47e9efe223a0d15707856015db980c0491d76/talpid-wireguard/src/stats.rs#L30)):*
> ```XML
> 
> impl Stats {
>     #[cfg(unix)]
> ```

Is this module even used by non-unix? All tests and most code is cfgd away on non-unix. Can maybe the entire module be cfgd to reduce the number of cfgs *in* the module?
___
*[`talpid-wireguard/src/wireguard_go.rs` line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbQ0YHEDLVrMwskeWrJ:-NbQ0YHEDLVrMwskeWrK:b-9nbdlt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0be47e9efe223a0d15707856015db980c0491d76/talpid-wireguard/src/wireguard_go.rs#L1)):*
> ```XML
> #![cfg(unix)]
> ```

I think this belongs above the `mod wireguard_go` statement in `lib.rs`. That's usually how we do it. And then it does not being an existent but empty module on non-unix.

faern (commented):
> Reviewed 9 of 31 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon)

___
*[`talpid-wireguard/src/stats.rs` line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbQ0DTq35CQfkNQPluO:-NbTY0RH9BFeJPgDXmVh:b-6tscrk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0be47e9efe223a0d15707856015db980c0491d76/talpid-wireguard/src/stats.rs#L30)):*

We do need `Stats`/`StatsMap` in wireguard-nt, but they're constructed directly. It doesn't make sense to put implementation details for wireguard-nt in this module, in my opinion. Even though that is something we do for wireguard-go and the Linux kernel impl.

Maybe we should instead move that code out of this module?
</blockquote></details>

I see. So the code that is cfgd is actually just `wireguard-go` parsing code? Maybe that code should live in the wireguard-go module then?

It's also not very clear that `cfg(unix)` means "code for wireguard-go". I wonder if we should use feature flags instead and explicitly do `--features wireguard-go` on the platforms where we want it :shrug: (this is not a decision, but just brainstorming) We can also avoid explicitly stating the features. We can go lower level and do something like:

```rust
// In build.rs
#[cfg(any(linux, target_os = "macos", target_os = "android"))]
println!("cargo:rustc-cfg=wireguard_go");

// In the crate code
#[cfg(wireguard_go)]
...
```
That would make things WAY easier to understand IMO

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995)* status: 26 of 34 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[`talpid-wireguard/src/logging.rs` line 49 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbQ-b0YEw0fOj3PoNwg:-NbU4FFH4P7p_fNlzTGn:b-52qxyt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0be47e9efe223a0d15707856015db980c0491d76/talpid-wireguard/src/logging.rs#L49)):*

Is this still needed? I don't know where/why this is dead code. But if it's still needed on some platform, can that be highlighted with `cfg_attr(..., allow(dead_code))`?
</blockquote></details>

Done. In line with your other suggestion, `cfg_attr(wireguard_go, allow(dead_code))` might be more readable, though.
___
*[`talpid-wireguard/src/logging.rs` line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbQ-0B5An10PZwBoxTr:-NbU4QWu6moilX0cJkw8:bbed46z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0be47e9efe223a0d15707856015db980c0491d76/talpid-wireguard/src/logging.rs#L94)):*

Same as the other comment I left. I honestly think `#[cfg(unix)]` is too vague here/not ideal. What do you think about `cfg(not(windows))`? Do you consider it worse for some reason?
</blockquote></details>

I moved this to `wireguard_go` where it belongs.
___
*[`talpid-wireguard/src/stats.rs` line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbQ0DTq35CQfkNQPluO:-NbU52zQ6NMZJiFqKnbe:bsy6d23) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0be47e9efe223a0d15707856015db980c0491d76/talpid-wireguard/src/stats.rs#L30)):*

I see. So the code that is cfgd is actually just `wireguard-go` parsing code? Maybe that code should live in the wireguard-go module then?

It's also not very clear that `cfg(unix)` means "code for wireguard-go". I wonder if we should use feature flags instead and explicitly do `--features wireguard-go` on the platforms where we want it :shrug: (this is not a decision, but just brainstorming) We can also avoid explicitly stating the features. We can go lower level and do something like:

```rust
// In build.rs
#[cfg(any(linux, target_os = "macos", target_os = "android"))]
println!("cargo:rustc-cfg=wireguard_go");

// In the crate code
#[cfg(wireguard_go)]
...
```
That would make things WAY easier to understand IMO
</blockquote></details>

I've moved the implementation-specific code into `wireguard_nt`, `wireguard_kernel`, and `wireguard_go`.
___
*[`talpid-wireguard/src/wireguard_go.rs` line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbQ0YHEDLVrMwskeWrJ:-NbU5qMe-F_tktTC-fW1:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0be47e9efe223a0d15707856015db980c0491d76/talpid-wireguard/src/wireguard_go.rs#L1)):*

I think this belongs above the `mod wireguard_go` statement in `lib.rs`. That's usually how we do it. And then it does not being an existent but empty module on non-unix.
</blockquote></details>

Done.

faern (commented):
> Reviewed 1 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995)* status: 27 of 34 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[`talpid-core/src/dns/windows/iphlpapi.rs` line 85 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbULrTBCL3qVyRcix6B:-NbULrTBCL3qVyRcix6C:b6jjdez) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85e7d6a3e55ec706ffb47be27428fa3bd1b34234/talpid-core/src/dns/windows/iphlpapi.rs#L85)):*
> ```XML
>         // This function is loaded at runtime since it may be unavailable. See the module-level
>         // docs. TODO: `windows_sys` can be used directly when support for Windows 10, 2004,
>         // is dropped.
> ```

This comment is actually wrong. The TODO applies when we drop support for anything *older* than 2004. However, this is completely unrelated to this PR and I'm not sure why this PR also does a bunch of formatting.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995)* status: 27 of 34 files reviewed, 4 unresolved discussions (waiting on @faern)

___
*[`talpid-wireguard/src/wireguard_kernel/stats.rs` line 4 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbUMeWn7eLb5L4BOf5v:-NbUPhU23tWUsJqNJ1iM:bxjjqy8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85e7d6a3e55ec706ffb47be27428fa3bd1b34234/talpid-wireguard/src/wireguard_kernel/stats.rs#L4)):*

I'm not sure I'm a fan of how this structure ended up. Now we have a bunch of `stats` modules in this crate and the `impl Foo` blocks are scattered all over.

Why is `parse_device_message` even an associated function on `stats`. It neither takes `self` nor returns `Self`. It does not look like it belongs on this type really. Looks more like it should be a method on `DeviceMessage` maybe (`parse_stats(&self) -> StatsMap`)? Or freestanding function in `wireguard_kernel`.

It's just a smell to me to have a dedicated module for *a single function* that is defined *somewhere else than the type it operates on* and also *does not really operate on the type it's accociated with*.
</blockquote></details>

I think of it as a constructor for `Stats` (or `StatsMap`, in reality). This particular constructor is only visible to `wireguard_kernel`. Does that make sense?

As for why it's in a module, that's just because I put everything in `impl::stats` for all implementations, and it made more sense in the other cases.

I can move it to a function if you still disagree, though.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995)* status: 27 of 34 files reviewed, 4 unresolved discussions (waiting on @dlon)

___
*[`talpid-wireguard/src/wireguard_kernel/stats.rs` line 4 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbUMeWn7eLb5L4BOf5v:-NbYyIKM1dusQIobTVSD:bfwm8it) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85e7d6a3e55ec706ffb47be27428fa3bd1b34234/talpid-wireguard/src/wireguard_kernel/stats.rs#L4)):*

I think of it as a constructor for `Stats` (or `StatsMap`, in reality). This particular constructor is only visible to `wireguard_kernel`. Does that make sense?

As for why it's in a module, that's just because I put everything in `impl::stats` for all implementations, and it made more sense in the other cases.

I can move it to a function if you still disagree, though.
</blockquote></details>

I actually don't know exactly what I prefer. But I think this current solution maybe has one slight code smell too many :thinking:

faern (changes requested):
> Reviewed 4 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995)* status: 31 of 34 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[`talpid-wireguard/src/wireguard_kernel/stats.rs` line 4 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbUMeWn7eLb5L4BOf5v:-Nd9hPJN9a2RkQgLIa2Z:b-4gaac4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85e7d6a3e55ec706ffb47be27428fa3bd1b34234/talpid-wireguard/src/wireguard_kernel/stats.rs#L4)):*

I actually don't know exactly what I prefer. But I think this current solution maybe has one slight code smell too many :thinking: 
</blockquote></details>

Can we make this a free standing function? Then at least we get rid of having `impl Stats` blocks scattered in so many places. Since there is no benefit of having this associated with `Stats` from what I can see. Then I guess we can live with the rest.

faern (commented):
> Reviewed 1 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995)* status: 32 of 34 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`talpid-wireguard/src/wireguard_kernel/stats.rs` line 4 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbUMeWn7eLb5L4BOf5v:-Nd9jVqs83ZPMqzWFPXI:b-9rhp16) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85e7d6a3e55ec706ffb47be27428fa3bd1b34234/talpid-wireguard/src/wireguard_kernel/stats.rs#L4)):*

Can we make this a free standing function? Then at least we get rid of having `impl Stats` blocks scattered in so many places. Since there is no benefit of having this associated with `Stats` from what I can see. Then I guess we can live with the rest.
</blockquote></details>

Well, let's go then. We do this in the other stats modules as well. I'm not particularly fond of it. But let's move ahead instead of changing all of them. Better to be consistent at least. So we should have it like this either in all stats modules or none.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995)* status: 20 of 35 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[`talpid-wireguard/src/logging.rs` line 49 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbQ-b0YEw0fOj3PoNwg:-Nd9vJaX53jkZcw75s2W:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0be47e9efe223a0d15707856015db980c0491d76/talpid-wireguard/src/logging.rs#L49)):*

Yes, if we are going with the `wireguard_go` cfg this should definitely have it as well.
</blockquote></details>

Done.
___
*[`talpid-wireguard/src/wireguard_go.rs` line 325 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbUN7FU6p5iAtdYMIP3:-Nd9ttt99qoGRAuvQXzp:b2nu6e1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85e7d6a3e55ec706ffb47be27428fa3bd1b34234/talpid-wireguard/src/wireguard_go.rs#L325)):*

This will just print `"_0"` not the actual content of `self.0`. It's missing `{}`. Same on the error below
</blockquote></details>

True. This was actually there before, but fixed.
___
*[`talpid-core/src/dns/windows/iphlpapi.rs` line 85 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NbULrTBCL3qVyRcix6B:-Nd9v9Wp93r11bgmGC6P:b-g9n02w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85e7d6a3e55ec706ffb47be27428fa3bd1b34234/talpid-core/src/dns/windows/iphlpapi.rs#L85)):*

This comment is actually wrong. The TODO applies when we drop support for anything *older* than 2004. However, this is completely unrelated to this PR and I'm not sure why this PR also does a bunch of formatting.
</blockquote></details>

You're correct, but I've reverted this diff since it's irrelevant to this PR.

faern (approved):
> Reviewed 1 of 8 files at r2, 10 of 12 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`CHANGELOG.md` line 105 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4995#-NdAEDMr60_Tiq5AUZEk:-NdAEDMr60_Tiq5AUZEl:b-efcc3p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d4096b8a6f8fd06dcde069a9de2527408af742a3/CHANGELOG.md#L105)):*
> ```Markdown
> 
> #### Windows
> - Removed wireguard-go (userspace WireGuard) support.
> ```

Nit: `Remove` for correct form

PR #4998, opened 2023-08-09
Ensure the app fetches account data whenever it enters the new account flow view

rablador:
Since it's possible to pay for account time in many ways in the app, and it's possible to do it on a completely different device, when in the new account flow, the app should try to pro-actively deduce if time has been added. So, whenever it enters a new account flow view, it should fetch new account data from the API.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998)

### DIFF
--- a/ios/MullvadVPN/Coordinators/App/WelcomeCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/App/WelcomeCoordinator.swift
@@ -38,7 +38,8 @@ final class WelcomeCoordinator: Coordinator, Presentable {
         }
         let interactor = WelcomeInteractor(
             deviceData: storedDeviceData,
-            accountData: storedAccountData
+            accountData: storedAccountData,
+            tunnelManager: tunnelManager
         )
 
         let controller = WelcomeViewController(interactor: interactor)
@@ -123,5 +124,11 @@ extension WelcomeCoordinator: WelcomeViewControllerDelegate {
         addChild(coordinator)
 
         coordinator.start()
+    }
+
+    func didUpdateDeviceState(deviceState: DeviceState) {
+        if deviceState.accountData?.isExpired == false {
+            didFinishPayment?(self)
+        }
     }
 }
--- a/ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeInteractor.swift
+++ b/ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeInteractor.swift
@@ -7,9 +7,21 @@
 //
 
 import Foundation
+import MullvadLogging
+
 class WelcomeInteractor {
+    private let logger = Logger(label: "WelcomeInteractor")
+
+    private let accountUpdateTimerInterval: TimeInterval = 60
+    private var accountUpdateTimer: DispatchSourceTimer?
+
     private let deviceData: StoredDeviceData
     private let accountData: StoredAccountData
+
+    private let tunnelManager: TunnelManager
+    private var tunnelObserver: TunnelObserver?
+
+    var didUpdateDeviceState: ((DeviceState) -> Void)?
 
     var viewModel: WelcomeViewModel {
         WelcomeViewModel(
@@ -18,8 +30,43 @@ class WelcomeInteractor {
         )
     }
 
-    init(deviceData: StoredDeviceData, accountData: StoredAccountData) {
+    init(deviceData: StoredDeviceData, accountData: StoredAccountData, tunnelManager: TunnelManager) {
         self.deviceData = deviceData
         self.accountData = accountData
+        self.tunnelManager = tunnelManager
+
+        let tunnelObserver =
+            TunnelBlockObserver(didUpdateDeviceState: { [weak self] tunnelManager, deviceState, previousDeviceState in
+                self?.didUpdateDeviceState?(deviceState)
+            })
+
+        tunnelManager.addObserver(tunnelObserver)
+        self.tunnelObserver = tunnelObserver
+    }
+
+    func startAccountUpdateTimer() {
+        logger.debug(
+            "Start polling account updates every \(accountUpdateTimerInterval) second(s)."
+        )
+
+        let timer = DispatchSource.makeTimerSource(queue: .main)
+        timer.setEventHandler { [weak self] in
+            self?.tunnelManager.updateAccountData()
+        }
+
+        accountUpdateTimer?.cancel()
+        accountUpdateTimer = timer
+
+        timer.schedule(wallDeadline: .now() + accountUpdateTimerInterval, repeating: accountUpdateTimerInterval)
+        timer.activate()
+    }
+
+    func stopAccountUpdateTimer() {
+        logger.debug(
+            "Stop polling account updates."
+        )
+
+        accountUpdateTimer?.cancel()
+        accountUpdateTimer = nil
     }
 }
--- a/ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeViewController.swift
+++ b/ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeViewController.swift
@@ -11,6 +11,7 @@ protocol WelcomeViewControllerDelegate: AnyObject {
     func didRequestToPurchaseCredit(controller: WelcomeViewController)
     func didRequestToRedeemVoucher(controller: WelcomeViewController)
     func didRequestToShowInfo(controller: WelcomeViewController)
+    func didUpdateDeviceState(deviceState: DeviceState)
 }
 
 class WelcomeViewController: UIViewController, RootContainment {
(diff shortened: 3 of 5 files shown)


mojganii (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998#-:-NbPhzH7-69vFYOdzN0l:bnfp4nl)* status: 0 of 3 files reviewed, all discussions resolved

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeInteractor.swift` line 48 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998#-NbTM0oG5G9wmb0oVA0o:-NbTM0oG5G9wmb0oVA0p:b-wa4s1t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78f683ea586cb3415973faf8cc9bbbe856c05915/ios/MullvadVPN/View%20controllers/CreationAccount/Welcome/WelcomeInteractor.swift#L48)):*
> ```Swift
>     }
> 
>     func startAccountUpdateTimer() {
> ```

We also need a `stopAccountUpdateTimer` method here otherwise this timer will keep running forever.

Since we tie it to `viewDidLoad` but there's no more `viewDidUnload` I would suggest moving both `startAccountUpdateTimer` (and its stop equivalent) to `viewWillAppear` and `viewWillDisappear` respectively for example

mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeInteractor.swift` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998#-NbUX2Gk5ZjUMIO7ZEt1:-NbUX2Gk5ZjUMIO7ZEt2:bqko35q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78f683ea586cb3415973faf8cc9bbbe856c05915/ios/MullvadVPN/View%20controllers/CreationAccount/Welcome/WelcomeInteractor.swift#L13)):*
> ```Swift
> 
> /// Interval used for periodic polling account updates.
> private let accountUpdateTimerInterval: TimeInterval = 60
> ```

I think it belongs the scope of interactor.
___
*[`ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeInteractor.swift` line 48 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998#-NbTM0oG5G9wmb0oVA0o:-NbUWHuR3dlr7i3d8oY2:bil8g0o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78f683ea586cb3415973faf8cc9bbbe856c05915/ios/MullvadVPN/View%20controllers/CreationAccount/Welcome/WelcomeInteractor.swift#L48)):*

We also need a `stopAccountUpdateTimer` method here otherwise this timer will keep running forever.

Since we tie it to `viewDidLoad` but there's no more `viewDidUnload` I would suggest moving both `startAccountUpdateTimer` (and its stop equivalent) to `viewWillAppear` and `viewWillDisappear` respectively for example
</blockquote></details>

can't he stop the timer on the deinit of interactor instead of `viewWillDisappear` ? if it is going to be placed on `viewWillAppear` please be sure it's not revoked multiple times on view life cycle

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeInteractor.swift` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998#-NbUX2Gk5ZjUMIO7ZEt1:-NbUji5kA7T87c27jXRh:b-79jju8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78f683ea586cb3415973faf8cc9bbbe856c05915/ios/MullvadVPN/View%20controllers/CreationAccount/Welcome/WelcomeInteractor.swift#L13)):*

I think it belongs the scope of interactor.
</blockquote></details>

Good catch ! Yes please don't make it a global
___
*[`ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeInteractor.swift` line 48 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998#-NbTM0oG5G9wmb0oVA0o:-NbUjpQ8CzzI8RDl2w3y:b-yolp90) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78f683ea586cb3415973faf8cc9bbbe856c05915/ios/MullvadVPN/View%20controllers/CreationAccount/Welcome/WelcomeInteractor.swift#L48)):*

can't he stop the timer on the deinit of interactor instead of `viewWillDisappear` ? if it is going to be placed on `viewWillAppear` please be sure it's not revoked multiple times on view life cycle
</blockquote></details>

Good point, that would be equally good. Please do test that the `WelcomeInteractor` gets properly deallocated if we go that way.

As a side note, I usually don't like to put code in `deinit` because it's hard to write unit tests for.
So I would still prefer if we avoid that, but if we go the `deinit` way, let's not make it a habit.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeInteractor.swift` line 48 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998#-NbTM0oG5G9wmb0oVA0o:-NbYlALo5Zf0XpKaQ8y-:bctwve1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78f683ea586cb3415973faf8cc9bbbe856c05915/ios/MullvadVPN/View%20controllers/CreationAccount/Welcome/WelcomeInteractor.swift#L48)):*

Good point, that would be equally good. Please do test that the `WelcomeInteractor` gets properly deallocated if we go that way.

As a side note, I usually don't like to put code in `deinit` because it's hard to write unit tests for.
So I would still prefer if we avoid that, but if we go the `deinit` way, let's not make it a habit. 
</blockquote></details>

I somewhat agree that it would be better to use view lifecycle for that so that even if we have a bug or leak, that we don't bombard our server with requests

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeInteractor.swift` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998#-NbUX2Gk5ZjUMIO7ZEt1:-NbYeYdQ9y7YvuxF643U:b2ij78o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78f683ea586cb3415973faf8cc9bbbe856c05915/ios/MullvadVPN/View%20controllers/CreationAccount/Welcome/WelcomeInteractor.swift#L13)):*

Good catch ! Yes please don't make it a global
</blockquote></details>

Just following the same pattern from similar use case elsewhere. I can scope it though.
___
*[`ios/MullvadVPN/View controllers/CreationAccount/Welcome/WelcomeInteractor.swift` line 48 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998#-NbTM0oG5G9wmb0oVA0o:-NbYmXowBDPzkezOabz-:bvb18t5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78f683ea586cb3415973faf8cc9bbbe856c05915/ios/MullvadVPN/View%20controllers/CreationAccount/Welcome/WelcomeInteractor.swift#L48)):*

I somewhat agree that it would be better to use view lifecycle for that so that even if we have a bug or leak, that we don't bombard our server with requests
</blockquote></details>

Fixed! Also updated OutOfTimeInteractor (which has a similar timer) with a stop function as well.

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4998#-:-NbxYcv1ESfmyrNhnWKM:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @mojganii)

PR #5023, opened 2023-08-21
Add billing client library

Pururun:
After we updated the build variants we should only depend on this in the Google Play version.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5023)


Pururun:
Closing this since it will be part of a larger PR in the future.

PR #5041, opened 2023-08-24
Change address cache schema IOS-282

pronebird:
Recently we received a number of reports in regards of persistent communication errors with the API, namely SSL errors and timeouts. 

In <= 2023.2 we used to shuffle a list of IPs and manage them as a stack, first IP in array of endpoints was used to talk to the API. If it didn't work, we picked next by moving the first IP to the end of the list. From >= 2023.3 we always use the first IP in the list.

It's quite likely that the app that ended up using the broken API IP would remain stuck using it after update 2023.3 which always relies on the first IP in the list. I also confirmed that one of API IPs hasn't been working for a while.

I think the most logical and simplest would be to modify the cache schema to store a single endpoint IP instead of array of them, which should then fail to decode after update from 2023.2 to 2023.3 and invalidate cache as address cache would fallback to default hardcoded IP in that case.

Additionally this PR adds a mutex guard to `AddressCache.loadFromFile()` since this call mutates internal cache data.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5041)

### DIFF
--- a/ios/CHANGELOG.md
+++ b/ios/CHANGELOG.md
@@ -27,6 +27,11 @@ Line wrap the file at 100 chars.                                              Th
 - Allow deleting account in account view.
 - Add new account flow
 
+### Fixed
+- Invalidate API IP address cache to fix connectivity issues for some of devices updating from 
+  2023.2 or earlier.
+
+
 ## [2023.3 - 2023-07-15]
 ### Added
 - Add search functionality to location selection view.
--- a/ios/MullvadREST/AddressCache.swift
+++ b/ios/MullvadREST/AddressCache.swift
@@ -113,17 +113,24 @@ extension REST {
             return cache.updatedAt
         }
 
-        /// Initializes the cache by reading the a cached file on disk.
+        /// Initializes cache by reading it from file on disk.
         ///
         /// If no cache file is present, a default API endpoint will be selected instead.
         public func loadFromFile() {
-            // The first time the application is ran, this statement will fail as there is no cache. This is fine.
-            // The cache will be filled when either `getCurrentEndpoint` or `setEndpoints()` are called.
-            do {
-                cache = try fileCache.read()
-            } catch {
-                logger.debug("Initialized cache with default API endpoint.")
-                cache = Self.defaultCachedAddresses
+            cacheLock.withLock {
+                // The first time the application is ran, this statement will fail as there is no cache. This is fine.
+                // The cache will be filled when either `getCurrentEndpoint()` or `setEndpoints()` are called.
+                do {
+                    cache = try fileCache.read()
+                } catch {
+                    // Log all errors except when file is not on disk.
+                    if (error as? CocoaError)?.code != .fileReadNoSuchFile {
+                        logger.error(error: error, message: "Failed to load address cache from disk.")
+                    }
+
+                    logger.debug("Initialized cache with default API endpoint.")
+                    cache = Self.defaultCachedAddresses
+                }
             }
         }
     }
--- a/ios/MullvadREST/AddressCache.swift
+++ b/ios/MullvadREST/AddressCache.swift
@@ -16,10 +16,10 @@ extension REST {
         private let logger = Logger(label: "AddressCache")
 
         /// Disk cache.
-        private let fileCache: any FileCacheProtocol<CachedAddresses>
+        private let fileCache: any FileCacheProtocol<StoredAddressCache>
 
         /// Memory cache.
-        private var cache: CachedAddresses = defaultCachedAddresses
+        private var cache: StoredAddressCache = defaultStoredCache
 
         /// Lock used for synchronizing access to instance members.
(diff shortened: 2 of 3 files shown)


mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5041)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`ios/MullvadREST/AddressCache.swift` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5041#-NcbaMLP-xBJnKH9ymhv:-NcbaMLP-xBJnKH9ymhw:b7nclbf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1174ede27b2d156b81ccd116b944c290ec5112ed/ios/MullvadREST/AddressCache.swift#L30)):*
> ```Swift
>         private let canWriteToCache: Bool
> 
>         /// The default set of endpoints to use as a fallback mechanism
> ```

now we are keeping one endpoint,I think we should change the comment

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5041)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @mojganii)

___
*[`ios/MullvadREST/AddressCache.swift` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5041#-NcbaMLP-xBJnKH9ymhv:-NcbphCR8QSmthc66V2j:budoq2d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1174ede27b2d156b81ccd116b944c290ec5112ed/ios/MullvadREST/AddressCache.swift#L30)):*

now we are keeping one endpoint,I think we should change the comment
</blockquote></details>

Now when you said that, I renamed `CachedAddresses` to `StoredAddressCache`. Please have another look.

rablador (changes requested):
> Reviewed 1 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5041)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii and @pronebird)

___
*[`ios/MullvadREST/AddressCache.swift` line 60 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5041#-NcfajhdDuYDyt7TtLGp:-NcfajheBpaJmHFOm0No:b-h1k6tx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d0a6181d0ca65afd0a2a36f1c29468de2000201/ios/MullvadREST/AddressCache.swift#L60)):*
> ```Swift
>                 // Reload from disk cache when in the Network Extension as there is no `AddressCacheTracker` running
>                 // there
>                 if canWriteToCache == false {
> ```

Not changed in this PR, but why do we do this check? There's another check further down that makes sense, ie checking before writing, but here we want to check before read...?

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5041)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii and @pronebird)

___
*[`ios/MullvadREST/AddressCache.swift` line 60 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5041#-NcfajhdDuYDyt7TtLGp:-Ncg06In5GN5aVxbeFlh:b-tmuqmr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d0a6181d0ca65afd0a2a36f1c29468de2000201/ios/MullvadREST/AddressCache.swift#L60)):*

We basically always read cache from disk when in read-only mode in network extension, because main app does not notify network extension when cache file is being modified. Maybe this could be done differently?
</blockquote></details>

Aha, so in practice this is actually a check for network extension? I think it's fine functionality-wise, but perhaps a little confusing when reading the code.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5041)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadREST/AddressCache.swift` line 60 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5041#-NcfajhdDuYDyt7TtLGp:-Ncg7zK7E1egco6TQERX:b-6bs12w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1d0a6181d0ca65afd0a2a36f1c29468de2000201/ios/MullvadREST/AddressCache.swift#L60)):*

Aha, so in practice this is actually a check for network extension? I think it's fine functionality-wise, but perhaps a little confusing when reading the code.
</blockquote></details>

Technically yes. Maybe we should make it clearer.

PR #5045, opened 2023-08-24
Show time added view when adding time in other view than out of time view

raksooo:
This PR fixes a bug where the "Out of time" or new-account views remained after adding time when in other views. To solve this a global `expired` state has been added and the history is modified when time is added to replace the "Out of time" view with the "Time added" view.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5045)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5045)* status: all files reviewed, 4 unresolved discussions (waiting on @raksooo)

___
*[`gui/src/renderer/app.tsx` line 222 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5045#-NchKS-N0DhaP486NXsY:-NchKS-OD8nuLxF4keo2:b-wnajww) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9fc57298575b8d824e0ac13981a9512c42e6df46/gui/src/renderer/app.tsx#L222)):*
> ```XML
>       );
>     }
>     this.setAccountExpiry(initialState.accountData?.expiry);
> ```

Is there any significance to this call being moved down? As far as I can tell, there is not. If there is, this seems brittle.
___
*[`gui/src/renderer/lib/history.tsx` line 124 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5045#-NchL4uj5KUZxj2acc_F:-NchL4uj5KUZxj2acc_G:b-uulwfp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9fc57298575b8d824e0ac13981a9512c42e6df46/gui/src/renderer/lib/history.tsx#L124)):*
> ```XML
>   ) => {
>     if (this.index === 0) {
>       this.reset(replacementLocation, replacementState);
> ```

Seems non-obvious that this clears `this.entries`, while the other case does not. Is that intentional?
___
*[`gui/src/renderer/redux/account/reducers.ts` line 5 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5045#-NchIWQ03be69QJQYqdp:-NchIWQ03be69QJQYqdq:b-t1g04z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9fc57298575b8d824e0ac13981a9512c42e6df46/gui/src/renderer/redux/account/reducers.ts#L5)):*
> ```XML
> 
> type LoginMethod = 'existing_account' | 'new_account';
> type ExpiredState = 'expired' | 'time-added';
> ```

Should we not be using underlines here instead of dashes?
___
*[`gui/src/renderer/redux/account/reducers.ts` line 151 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5045#-NchJzJ479Jxq-gGNJYl:-NchJzJ479Jxq-gGNJYm:b5xdqr5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9fc57298575b8d824e0ac13981a9512c42e6df46/gui/src/renderer/redux/account/reducers.ts#L151)):*
> ```XML
>         expiry: action.expiry,
>         status:
>           state.status.type === 'ok'
> ```

Is it not possible to assign `state.status` to a variable `status` and replace `status.expiredState` in the if statement above? Then you could simply return `status`.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5045)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5047, opened 2023-08-25
Update CI

pronebird:
1. Set working directory for both `swiftformat` and `swiftlint` to prevent scanning the whole file tree from the root.
2. Use `env` to pass `NSUnbufferedIO` to `xcodebuild` because the way it was wired up was wrong.
3. Print SwiftLint version before linting so that we know which version is installed on CI. Should help with troubleshooting if something breaks.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5047)

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -25,6 +25,7 @@ jobs:
         run: |
           swiftformat --version
           swiftformat --lint .
+        working-directory: ios
 
   swiftlint:
     name: Run swiftlint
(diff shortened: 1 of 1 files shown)


rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5047)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5053, opened 2023-08-25
Use appTarget in address cache

pronebird:
As @rablador pointed out it's probably better to use app target where we want to achieve a specific behavior such as in AddressCache.

This is a draft PR to gather feedback.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5053)

### DIFF
--- a/ios/MullvadREST/AddressCache.swift
+++ b/ios/MullvadREST/AddressCache.swift
@@ -24,8 +24,8 @@ extension REST {
         /// Lock used for synchronizing access to instance members.
         private let cacheLock = NSLock()
 
-        /// Whether address cache can be written to.
-        private let canWriteToCache: Bool
+        /// Indicates which target requested access address cache. Main bundle has read/write access to cache, while packet tunnel can only read.
+        private let appTarget: ApplicationTarget
 
         /// The default endpoint to use as a fallback mechanism.
         private static let defaultStoredCache = StoredAddressCache(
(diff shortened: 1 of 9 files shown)


rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5053)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5053#-Ncuwvey6IsJJS6TPBXa:-NcuwvezD-g9I464NIZi:bpck6cy) (no related file):*
If the intent is indeed to handle things differently according to app target, then I think this is a lot easer to read and follow.

However, if app target just so happens to coincide with read/write access (and mapping [main -> read/write] and [tunnel -> read only] isn't always in line with the logic we want to achieve, or loses important granularity), then I think we should keep the write check.

___

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5053)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5053#-Ncuwvey6IsJJS6TPBXa:-NcvbEfd5M5dPNXEYXPa:bqb8paq) (no related file):*

If the intent is indeed to handle things differently according to app target, then I think this is a lot easer to read and follow.

However, if app target just so happens to coincide with read/write access (and mapping [main -> read/write] and [tunnel -> read only] isn't always in line with the logic we want to achieve, or loses important granularity), then I think we should keep the write check.
</blockquote></details>

For the most part that's exactly the intent but I left this PR as a draft for others to share their thoughts.

___

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5053)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`ios/MullvadVPNTests/AddressCacheTests.swift` line 72 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5053#-Nd59aPK92ZhpbKvvdvs:-Nd59aPK92ZhpbKvvdvt:b-21u2lr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/eec9b7dc80fa2124a8e78f038566a08dfa7eef3f/ios/MullvadVPNTests/AddressCacheTests.swift#L72)):*
> ```Swift
> 
>         let fileCache = MockFileCache<REST.StoredAddressCache>()
>         let addressCache = REST.AddressCache(appTarget: .mainApp, fileCache: fileCache)
> ```

If you look just at the test, and replace `mainApp` with`packetTunnel` the test fails, but it doesn't really make sense why.
If you switch `canWriteToCache` to `false` the test fails, but it makes sense.

This tells me that `AddressCache` shouldn't (and doesn't) really care about the concept of `AppTarget`.

It's a much stronger intent to say that "Packet tunnel cannot write to the cache" rather than encoding that fact in `appTarget`, and then look at the code to figure that out.

I don't think we should make this change.

PR #5055, opened 2023-08-28
Update shadowsocks to 1.16

pinkisemils:
Since `shadowsocks` has a release that includes all the changes we need, I've gone and updated our dependencies to use the _crates.io_ release.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5055)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -382,7 +382,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a6358dedf60f4d9b8db43ad187391afe959746101346fe51bb978126bec61dfb"
 dependencies = [
  "heck",
- "indexmap",
+ "indexmap 1.7.0",
  "log",
  "proc-macro2",
  "quote",
@@ -390,7 +390,7 @@ dependencies = [
  "serde_json",
  "syn 1.0.100",
  "tempfile",
- "toml",
+ "toml 0.5.8",
 ]
 
 [[package]]
@@ -879,6 +879,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "drain"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2f1a0abf3fcefad9b4dd0e414207a7408e12b68414a01e6bb19b897d5bd7632d"
+dependencies = [
+ "tokio",
+]
+
+[[package]]
 name = "duct"
 version = "0.13.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5055)* status: all files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[`Cargo.lock` line 4263 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5055#-NcwB7IBEh1nvez2QS0z:-NcwB7IBEh1nvez2QS1-:b5gwnye) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e24abb9960a004a5d3cf0c6ca1a696313472c7c0/Cargo.lock#L4263)):*
> ```TOML
> [[package]]
> name = "trust-dns-resolver"
> version = "0.23.0"
> ```

Can we not avoid having both 0.22 and 0.23? That would be nice.
___
*[`ios/MullvadTransport/shadowsocks-proxy/Cargo.toml` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5055#-NcwA7iN1ZBKZwP-RnsC:-NcwA7iN1ZBKZwP-RnsD:bxp254x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e24abb9960a004a5d3cf0c6ca1a696313472c7c0/ios/MullvadTransport/shadowsocks-proxy/Cargo.toml#L16)):*
> ```TOML
> [dependencies]
> shadowsocks-service.version = "1.16"
> shadowsocks-service.features = [ "local", "stream-cipher", "local-http", "local-tunnel" ]
> ```

A bit nitpicky. But this is to my knowledge the only place we use this bespoke format to define a dependency. Can we not just do it the same way we do for other dependencies to make it consistent?

Slightly related. Not going to block the PR on this, but I suggest uplifting `shadowsocks` and `shadowsocks-service` to workspace dependencies. Since we always want to upgrade them in lockstep anyway.

faern (approved):
> Reviewed 2 of 3 files at r2, 5 of 5 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5055)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`ios/MullvadTransport/shadowsocks-proxy/Cargo.toml` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5055#-NcwA7iN1ZBKZwP-RnsC:-NcwLOM62IhBfFm34NMj:bqrc3i2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e24abb9960a004a5d3cf0c6ca1a696313472c7c0/ios/MullvadTransport/shadowsocks-proxy/Cargo.toml#L16)):*

A bit nitpicky. But this is to my knowledge the only place we use this bespoke format to define a dependency. Can we not just do it the same way we do for other dependencies to make it consistent?

Slightly related. Not going to block the PR on this, but I suggest uplifting `shadowsocks` and `shadowsocks-service` to workspace dependencies. Since we always want to upgrade them in lockstep anyway.
</blockquote></details>

Are the features no longer needed? You still use this special syntax we don't use anywhere else. How about `shadowsocks-service = { workspace = true }`?

dlon (approved):
> Reviewed 2 of 3 files at r2, 3 of 5 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5055)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

PR #5059, opened 2023-08-28
Prevent OpenVPN servers from showing in relay list when protocol is automatic

raksooo:
This PR prevents OpenVPN relays from showing in the select location view when protocol is set to automatic. This is the last part of the process to make WireGuard the default protocol.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5059)

### DIFF
--- a/gui/src/renderer/lib/filter-locations.ts
+++ b/gui/src/renderer/lib/filter-locations.ts
@@ -44,9 +44,6 @@ function getTunnelProtocolFilter(
     endpointTypes.push('bridge');
   } else if (tunnelProtocol === 'any') {
     endpointTypes.push('wireguard');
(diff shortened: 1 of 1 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5059)* status: :shipit: complete! all files reviewed, all discussions resolved

raksooo:
We've realized that we want the behavior we currently want on main. If `Automatic` is selected then we want to show both OpenVPN and WireGuard servers unless multihop is enabled. The reason to why we don't want to show OpenVPN servers when multihop is enabled is because it would make it very easy for users to choose an invalid combination of setting resulting in the error state.

PR #5062, opened 2023-08-29
Only set TCP flags condition for PF rules if the protocol is TCP

dlon:
`get_allow_relay_rule` in particular produces the following rule, which is considered invalid when added using `pfctl`:

```
pass out log (all) quick inet proto udp from any to IP port = PORT user = 0 flags S/SA keep state
```

Incidentally, this works around everything being blocked on macOS 14 Beta 6.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5062)

### DIFF
--- a/talpid-core/src/firewall/macos.rs
+++ b/talpid-core/src/firewall/macos.rs
@@ -6,7 +6,7 @@ use std::{
     net::{IpAddr, Ipv4Addr},
 };
 use subslice::SubsliceExt;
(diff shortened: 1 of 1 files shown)


faern:
Our current rules are correct. It's PF that is broken in macOS 14 betas. We have reported this to Apple and we just hope they will fix PF before macOS 14 goes stable. Otherwise things will be really bad.

PR #5063, opened 2023-08-29
Drop socket before the address is reused

pinkisemils:
The socket used to get a free local address has to be dropped sooner since the newer version of  shadowsocks binds the sockets it needs before entering the server's main loop. This fixes the issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5063)

### DIFF
--- a/talpid-openvpn/src/proxy/shadowsocks.rs
+++ b/talpid-openvpn/src/proxy/shadowsocks.rs
@@ -82,10 +82,10 @@ impl ShadowsocksProxyMonitor {
             config.outbound_fwmark = settings.fwmark;
         }
 
+        std::mem::drop(sock);
         let srv = local::Server::new(config).await?;
 
         let (fut, server_abort_handle) = abortable(async move {
-            let _ = sock;
             let result = srv.run().await;
             if let Err(error) = &result {
                 log::error!(
--- a/talpid-openvpn/src/proxy/shadowsocks.rs
+++ b/talpid-openvpn/src/proxy/shadowsocks.rs
@@ -35,28 +35,16 @@ impl ShadowsocksProxyMonitor {
     }
 
     async fn start_inner(settings: &ShadowsocksProxySettings) -> io::Result<Self> {
-        // TODO: Patch shadowsocks so the bound address can be obtained afterwards.
-        let addr = SocketAddrV4::new(Ipv4Addr::LOCALHOST, 0);
-        let sock = Socket::new(
-            Domain::IPV4,
-            socket2::Type::STREAM,
-            Some(socket2::Protocol::TCP),
-        )?;
-        sock.set_reuse_address(true)?;
-        sock.bind(&SockAddr::from(addr))?;
-
-        let bound_addr = sock
-            .local_addr()?
-            .as_socket_ipv4()
-            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "missing IPv4 address"))?;
-
         let mut config = Config::new(ConfigType::Local);
 
         config.fast_open = true;
 
         let mut local = LocalConfig::new(ProtocolType::Socks);
         local.mode = Mode::TcpOnly;
-        local.addr = Some(ServerAddr::SocketAddr(SocketAddr::from(bound_addr)));
+        local.addr = Some(ServerAddr::SocketAddr(SocketAddr::from((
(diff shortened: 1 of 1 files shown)


dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5063)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`talpid-openvpn/src/proxy/shadowsocks.rs` line 85 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5063#-Nd1N89N9Q-070xm7vA9:-Nd1N89OAl0aOV5M5sHC:b-g89rua) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9b61d6ccf7e18bc4c95473447950c6a7ac5c88a2/talpid-openvpn/src/proxy/shadowsocks.rs#L85)):*
> ```XML
>         }
> 
>         std::mem::drop(sock);
> ```

I think we should just move this whole thing into a function that returns a `SocketAddrV4` instead of dropping above:

```rust
  let addr = SocketAddrV4::new(Ipv4Addr::LOCALHOST, 0);
  let sock = Socket::new(
      Domain::IPV4,
      socket2::Type::STREAM,
      Some(socket2::Protocol::TCP),
  )?;
  sock.bind(&SockAddr::from(addr))?;
  sock
      .local_addr()?
      .as_socket_ipv4()
      .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "missing IPv4 address"))
```

Also, get rid of `SO_REUSEADDR`.

In principle, it's possible for some other thread/process to steal the port before we can reuse it, but that's already the case now with the updated code. So we might as well clean it up a bit?

dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5063)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`talpid-openvpn/src/proxy/shadowsocks.rs` line 85 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5063#-Nd1N89N9Q-070xm7vA9:-Nd1OpQh0p8uz9WM_DP6:btxtjlx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9b61d6ccf7e18bc4c95473447950c6a7ac5c88a2/talpid-openvpn/src/proxy/shadowsocks.rs#L85)):*

I think we just move this whole thing into a function that returns a `SocketAddrV4` instead of dropping above:

```rust
  let addr = SocketAddrV4::new(Ipv4Addr::LOCALHOST, 0);
  let sock = Socket::new(
      Domain::IPV4,
      socket2::Type::STREAM,
      Some(socket2::Protocol::TCP),
  )?;
  sock.bind(&SockAddr::from(addr))?;
  sock
      .local_addr()?
      .as_socket_ipv4()
      .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "missing IPv4 address"))
```

Also, get rid of `SO_REUSEADDR`.

In principle, it's possible for some other thread/process to steal the port before we can reuse it, but that's already the case now with the updated code. So we might as well clean it up a bit?
</blockquote></details>

We could also just use `TcpListener::bind` (or `TcpSocket`) and `TcpListener::local_addr` from `tokio`.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5063)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`talpid-openvpn/src/proxy/shadowsocks.rs` line 85 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5063#-Nd1N89N9Q-070xm7vA9:-Nd5c1sc2TqKlgL45QKT:b1xl80q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9b61d6ccf7e18bc4c95473447950c6a7ac5c88a2/talpid-openvpn/src/proxy/shadowsocks.rs#L85)):*

We could also just use `TcpListener::bind` and `TcpListener::local_addr` from `tokio`.
</blockquote></details>

Or we can not bother at all since now the server will bind it's sockets before entering it's main loop. :tada:

Issue #4952, opened 2023-08-02
Ubuntu 23.04 - Graphics Corruption in 2023.5-beta1


tommodev:
## Issue report

Howdy, sorry couldn't get to the report issue button due to graphics bork but captured a log.

### Operating system

Neofetch Dump -

```
OS: Ubuntu 23.04 x86_64
Kernel: 6.2.0-26-generic
Resolution: 3840x2160
DE: GNOME 44.3
WM: Mutter
WM Theme: Adwaita
Theme: Yaru-red [GTK2/3]
Icons: Yaru-red [GTK2/3]
CPU: AMD Ryzen 5 2600 (12) @ 3.400GHz
GPU: AMD ATI Radeon RX 470/480/570/570X/580/580X/590
Memory: 5993MiB / 31993MiB
```

### App version

2023.5-beta1 installed via Deb, sourced from GH

### Issue description

Mullvad Desktop is in tray, connection OK, but on opening Gui from systray all graphics within the window (excluding system border) were corrupted. Apologies didn't get screenshot.

Logfile from CLI launch of '/opt/Mullvad VPN/mullvad-gui' attached to issue.
[mullvad-exec-log.txt](https://github.com/mullvad/mullvadvpn-app/files/12239397/mullvad-exec-log.txt)

Log seems pretty upset about shader compilation:

```
...
[13829:0802/185437.071501:ERROR:shared_context_state.cc(81)] Skia shader compilation error
------------------------
// Vertex SKSL
#extension GL_NV_shader_noperspective_interpolation: require
...
Errors:
link failed but did not provide an info log
[13829:0802/185437.075354:ERROR:shared_context_state.cc(81)] Skia shader compilation error
...
```

tommodev:
Can repo. 2 screenshots attached. Downgrading to 2023.4 and visuals are OK, upgrading to 2023.5-beta1 introduces the corruption again.

Otherwise app is "ok" and mullvad connects to server as expected.

![Screenshot from 2023-08-03 07-23-59](https://github.com/mullvad/mullvadvpn-app/assets/79486309/d08fe7c1-1172-4142-827e-16d021299af2)

![Screenshot from 2023-08-03 07-24-54](https://github.com/mullvad/mullvadvpn-app/assets/79486309/6ad88a7a-ed6f-4cca-a96d-b161157cb0b2)

tommodev:
In case it's relevant, running using Ubuntu's default of Wayland, not X11.

[...]

PR #5068, opened 2023-08-30
Clear GPU cache on upgrade on Linux

raksooo:
Reported in https://github.com/mullvad/mullvadvpn-app/issues/4952 the GPU cache can be incompatible with future Electron versions. This PR removes the cache directory on upgrade by removing it after the app has been installed. The reason why we don't do it in `after-remove.sh` (where we already have code for removing app data for all users) is that it would take one more release before the change would kick into action. When upgrading `*-remove.sh` scripts are run from the old versions package while the `*-install.sh` scripts are run from the new package.

### DIFF
--- a/dist-assets/linux/after-remove.sh
+++ b/dist-assets/linux/after-remove.sh
@@ -8,32 +8,36 @@ function remove_logs_and_cache {
     echo "Failed to remove mullvad-vpn cache"
 }
 
+function get_home_dirs {
+  if [[ -f "/etc/passwd" ]] && command -v cut > /dev/null; then
+      cut -d: -f6 /etc/passwd
+  fi
+}
+
 function remove_config {
   rm -r --interactive=never /etc/mullvad-vpn || \
     echo "Failed to remove mullvad-vpn config"
 
   # Remove app settings and auto-launcher for all users. This doesn't respect XDG_CONFIG_HOME due
   # to the complexity required.
-  if [[ -f "/etc/passwd" ]] && command -v cut > /dev/null; then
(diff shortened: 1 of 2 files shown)


raksooo (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5068)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`dist-assets/linux/before-install.sh` line 39 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5068#-Nd5UqCg-lJsAi_DY3mI:-Nd5sseN5Qv7v3njb6k8:bbr3dvq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ece0b4ad1e0d04efc3b46bc3bc913220cb4068af/dist-assets/linux/before-install.sh#L39)):*

This is a bit anecdotal and workaround focused description of this functionality. Is it not true that Electron upgrades in general could have incompatible caches, and this being a potential issue regardless of any reports? If so, I think a more direct description such as the following would be more appropriate:
> Different electron versions can have incompatible GPU caches. Clearing it on upgrades makes sure the same cache is not used across versions.
</blockquote></details>

Sounds good. Done.

faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5068#-:-Nd5yZ6wEmXXKsrEgoiz:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5069, opened 2023-08-30
On delete account screen, add account number validation during input

rablador:
When entering the last 4 digits of the account number on the account deletion screen, the OK button lights up even if the number is incorrect. When pressing the OK button an error message is displayed, saying that the number is indeed incorrect. The process has now been simplified by having the OK button itself function as validation instead and only enabling it if input is valid.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5069)

### DIFF
--- a/ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift
+++ b/ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift
@@ -253,9 +253,18 @@ class AccountDeletionContentView: UIView {
         }
     }
 
-    private var isAccountNumberLengthSatisfied: Bool {
-        let length = accountTextField.text?.count ?? 0
-        return length == 4
+    private var isInputValid: Bool {
+        guard let input = accountTextField.text,
+              let accountNumber = viewModel?.accountNumber,
+              !accountNumber.isEmpty
+        else {
+            return false
+        }
+
+        let inputLengthIsValid = input.count == 4
+        let inputMatchesAccountNumber = accountNumber.suffix(4) == input
+
+        return inputLengthIsValid && inputMatchesAccountNumber
     }
 
     weak var delegate: AccountDeletionContentViewDelegate?
@@ -334,7 +343,7 @@ class AccountDeletionContentView: UIView {
         } else {
             activityIndicator.stopAnimating()
         }
-        deleteButton.isEnabled = isDeleteButtonEnabled && isAccountNumberLengthSatisfied
+        deleteButton.isEnabled = isDeleteButtonEnabled && isInputValid
         statusLabel.text = text
         statusLabel.textColor = textColor
     }


ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift
```diff
@@ -245,9 +245,18 @@ class AccountDeletionContentView: UIView {
...
+              let accountNumber = viewModel?.accountNumber,
+              !accountNumber.isEmpty
+        else {
+            return false
+        }
+
+        let inputLengthIsValid = input.count == 4
+        let inputMatchesAccountNumber = accountNumber.suffix(4) == input
+
+        return inputLengthIsValid && inputMatchesAccountNumber
```
MrChocolatine, line 259: Is `inputLengthIsValid` really needed?

`accountNumber.suffix(4) == input` seems to imply that `input` is indeed 4 characters long when it matches `accountNumber.suffix(4)`... ?

ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift
```diff
@@ -326,7 +335,7 @@ class AccountDeletionContentView: UIView {
         } else {
             activityIndicator.stopAnimating()
         }
-        deleteButton.isEnabled = isDeleteButtonEnabled && isAccountNumberLengthSatisfied
+        deleteButton.isEnabled = isDeleteButtonEnabled && isInputValid
```
MrChocolatine, line 338: Should `isDeleteButtonEnabled` be updated too?

Currently it still says the button should be enabled upon failure:
https://github.com/mullvad/mullvadvpn-app/blob/9cc7a43a5dd76ccae541a43a7b95ffe61e627898/ios/MullvadVPN/View%20controllers/AccountDeletion/AccountDeletionContentView.swift#L221-L228

rablador: We want to disable the button when loading (to avoid interaction while trying to delete the account), otherwise enabled.

ios/MullvadVPN/View controllers/AccountDeletion/AccountDeletionContentView.swift
rablador: The param in .suffix() specifies a max length, meaning the result could be less than that. In practice the account number - if present - should never be less than 4 digits, so for most (or perhaps all realistic) cases it should work as intended.
Theoretically though, accountNumber.suffix(4) could for some reason return something less than 4 digits, and if so the condition would be incorrect (eg. 3 digit account number matching 3 digit input). It's a safeguard, but perhaps it's not really necessary...

MrChocolatine: > In practice the account number - if present - should never be less than 4 digits

Yep this is what I assumed, hence my message. But okay 👍  sorry for the bother!

mojganii (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5069#-:-Nd_YCd3BRkfcnxaFb-2:bnfp4nl)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @MrChocolatine)

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5069)* status: all files reviewed, 2 unresolved discussions (waiting on @MrChocolatine)

PR #5077, opened 2023-08-31
Use the block based KVO API in AsyncOperation

buggmagnet:


ios/Operations/AsyncOperation.swift
```diff
@@ -238,45 +238,22 @@ open class AsyncOperation: Operation {
...
-            context: &Self.observerContext
-        )
-    }
-
-    deinit {
-        removeObserver(self, forKeyPath: #keyPath(isReady), context: &Self.observerContext)
+        isReadyObserver = observe(\.isReady, options: []) { operation, _ in
+            operation.checkReadiness()
+            // Clear the observer when the operation is finished to avoid leaking memory
+            if operation.state == .finished {
```
pronebird, line 249: It looks like observation API does not maintain a strong reference to target anymore. Try this in playground:

```swift
class MyOperation: Operation {
    private var observer: NSKeyValueObservation?

    override init() {
        super.init()

        observer = observe(\.isReady, changeHandler: { op, _ in
            print(op)
        })
    }

    deinit {
        print("deinit!")
    }
}

var op: MyOperation? = MyOperation()

op = nil

print("done")
```

Don't forget to remove `observerContext` on the class if we don't need it anymore.

pronebird (changes requested):

ios/Operations/AsyncOperation.swift
```diff
@@ -117,7 +117,7 @@ open class AsyncOperation: Operation {
         _error
     }
 
-    override public final var isReady: Bool {
+    dynamic override public final var isReady: Bool {
```
pronebird, line 120: Any particular reason why that has to be marked `dynamic`?

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5077)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[`ios/Operations/AsyncOperation.swift` line 120 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5077#gh-1311460035:-NdAr3oa3jGdoWmvgL8M:b-4wywvt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ebd6a0b2b42c740b6f4daf7a65a31570e32d756f/ios/Operations/AsyncOperation.swift#L120)):*

Any particular reason why that has to be marked `dynamic`?
</blockquote></details>

It's to force dynamic dispatching via the ObjC runtime which is required for KVO to work
Technically it has to be marked both `@objc` and `dynamic` but we inherit from `NSObject` so all our properties are already inherently `@objc` 

[This article](https://developer.apple.com/documentation/swift/using-key-value-observing-in-swift) shows an example on how to do it
___
*[`ios/Operations/AsyncOperation.swift` line 249 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5077#gh-1311452841:-NdB44Fp4hTGc4iWPIfv:boe7mp6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ebd6a0b2b42c740b6f4daf7a65a31570e32d756f/ios/Operations/AsyncOperation.swift#L249)):*

It looks like observation API does not maintain a strong reference to target anymore. Try this in playground:

```swift
class MyOperation: Operation {
    private var observer: NSKeyValueObservation?

    override init() {
        super.init()

        observer = observe(\.isReady, changeHandler: { op, _ in
            print(op)
        })
    }

    deinit {
        print("deinit!")
    }
}

var op: MyOperation? = MyOperation()

op = nil

print("done")
```

</blockquote></details>

I was under that impression too, but I used Instruments to make sure we were not leaking memory, and without that it would leak memory for each operation created.

You can run the check yourself if you want, remove the `operation.isReadyObserver = nil` line and run under instruments to see the result.

pronebird (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5077#-:-NdUVhtR57vD0Z9vmuwy:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5089, opened 2023-09-04
Detect whether OS is Windows Server

dlon:
We do not officially support Windows Server, but we could still distinguish between it and the desktop OSes. Currently, we incorrectly assume that the user is necessarily on the desktop version.

This code simply returns "Windows Server" and does not bother to detect the specific version/year. This is fine because we log the build number as well, and the main point is to distinguish between desktop and server.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5089)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,10 @@ Line wrap the file at 100 chars.                                              Th
 #### Windows
 - Remove wireguard-go (userspace WireGuard) support.
 
+### Fixed
+#### Windows
+- Correctly detect whether OS is Windows Server (primarily for logging in daemon.log).
+
 
 ## [android/2023.6-beta1] - 2023-08-29
 
--- a/talpid-platform-metadata/src/lib.rs
+++ b/talpid-platform-metadata/src/lib.rs
@@ -15,6 +15,3 @@ mod imp;
 mod imp;
 
 pub use self::imp::{extra_metadata, short_version, version};
-
-#[cfg(target_os = "windows")]
-pub use self::imp::WindowsVersion;
--- a/talpid-platform-metadata/src/windows.rs
+++ b/talpid-platform-metadata/src/windows.rs
@@ -37,7 +37,7 @@ pub fn extra_metadata() -> impl Iterator<Item = (String, String)> {
     std::iter::empty()
 }
 
-pub struct WindowsVersion {
+struct WindowsVersion {
     inner: RTL_OSVERSIONINFOEXW,
 }
 
--- a/talpid-platform-metadata/src/windows.rs
+++ b/talpid-platform-metadata/src/windows.rs
@@ -6,11 +6,12 @@ use std::{
 };
 use windows_sys::Win32::System::{
     LibraryLoader::{GetModuleHandleW, GetProcAddress},
-    SystemInformation::OSVERSIONINFOW,
+    SystemInformation::OSVERSIONINFOEXW,
+    SystemServices::VER_NT_WORKSTATION,
 };
 
 #[allow(non_camel_case_types)]
-type RTL_OSVERSIONINFOW = OSVERSIONINFOW;
+type RTL_OSVERSIONINFOEXW = OSVERSIONINFOEXW;
 
 pub fn version() -> String {
     let (major, build) = WindowsVersion::new()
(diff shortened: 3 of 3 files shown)


faern (changes requested):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5089)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`talpid-platform-metadata/src/windows.rs` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5089#-NdUPl3vCBH9IIJV86rn:-NdUPl3vCBH9IIJV86ro:b-5zta2o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ee87483abba9a1ce8060c9c79d54a21be01579f/talpid-platform-metadata/src/windows.rs#L32)):*
> ```XML
>         .map(|version| version.windows_version_string())
>         .unwrap_or("N/A".into());
>     format!("Windows {}", version_string)
> ```

This is IMHO a bit harder to understand than the old version. Because now the reader has to check in more places to know if the prefix "Windows" is part of the string or not. This PR currently goes from hardcoding "Windows" in two places to having it in three places.

Personally I think I would have preferred to keep "Windows" out of `windows_version_string` and just have it return "Server" when it detects server. Not going to block the PR on this, but it's my suggestion for an improvement.
___
*[`talpid-platform-metadata/src/windows.rs` line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5089#-NdUOXmgAudC9QcnYfJu:-NdUOXmh3RUMRaW80fn2:b1v6l16) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4bebd7087a5829a1ed7d339c079b2ac37d8a8d02/talpid-platform-metadata/src/windows.rs#L42)):*
> ```XML
> }
> 
> pub struct WindowsVersion {
> ```

Is there any reason this struct is public? I can't find any usage of it outside of this module. I'm asking because the currenly public facing method `WindowsVersion::windows_version_string` now changed format (The word "Windows" was added to the start) and I was unsure what parts of the app this would affect.

If this struct is indeed not used outside of this module, let's make it private to better show that.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5089)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5089)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`talpid-platform-metadata/src/windows.rs` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5089#-NdUPl3vCBH9IIJV86rn:-NdUjJk8EDtKaD6D8Yqg:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ee87483abba9a1ce8060c9c79d54a21be01579f/talpid-platform-metadata/src/windows.rs#L32)):*

This is IMHO a bit harder to understand than the old version. Because now the reader has to check in more places to know if the prefix "Windows" is part of the string or not. This PR currently goes from hardcoding "Windows" in two places to having it in three places.

Personally I think I would have preferred to keep "Windows" out of `windows_version_string` and just have it return "Server" when it detects server. Not going to block the PR on this, but it's my suggestion for an improvement.
</blockquote></details>

Done.
___
*[`talpid-platform-metadata/src/windows.rs` line 42 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5089#-NdUOXmgAudC9QcnYfJu:-NdUj15zC3C5VM9ANJSM:b-hm92d3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4bebd7087a5829a1ed7d339c079b2ac37d8a8d02/talpid-platform-metadata/src/windows.rs#L42)):*

Is there any reason this struct is public? I can't find any usage of it outside of this module. I'm asking because the currenly public facing method `WindowsVersion::windows_version_string` now changed format (The word "Windows" was added to the start) and I was unsure what parts of the app this would affect.

If this struct is indeed not used outside of this module, let's make it private to better show that.
</blockquote></details>

Nope, we do not use it elsewhere. Done.

faern (approved):
Sweet! Overall way fewer changes and IMHO nicer code. This is perfect :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5089#-:-NdV3f6X5bYtBBLFinMX:b3surz1)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5091, opened 2023-09-04
Add 100 64 range

tristandruyen:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5091)


tristandruyen:
My bad, i miss-clicked the button in my fork, this should stay a fork for now :)

faern:
Unless some evidence as to why this would be safe is provided, I don't think we are going to merge this upstream anytime soon. `100.64.0.0/10` is not a *local* range. It's a carrier grade range and intended to be used, among other things, between ISPs and their customers. This is by definition not local.

PR #5093, opened 2023-09-04
Add http client module to use for prototyping

Pururun:
Also add the two google play billing end points.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5093)


Pururun:
Closing this as with the daemon implementation this is no longer needed.

PR #5100, opened 2023-09-04
Fix or report all current swiftlint warnings in xcode

rablador:
The goal is to go through the existing lint warnings and either fix them or disable + report them (in Linear).

### DIFF
--- a/ios/.swiftlint.yml
+++ b/ios/.swiftlint.yml
@@ -35,7 +35,8 @@ line_length:
 type_name:
   min_length: 4
   max_length:
-    error: 50
+    warning: 50
+    error: 60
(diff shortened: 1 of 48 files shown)


rablador (commented):
For the [weak self] caption warnings, see discussion in Linear: https://mullvad.slack.com/archives/C052D9X882U/p1693834413327759

I opted to go for `guard let self = self` and no explicit self in closure. Can easily be changed if other way is preferred.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5100#-:-NdXDI9A8ql_aWduJKw6:buekypk)* status: 0 of 48 files reviewed, all discussions resolved

PR #5149, opened 2023-09-12
remove user token after account is deleted

mojganii:
This PR introduces a crucial improvement to our system. After an account is deleted, it is essential to remove the associated user token from memory. This enhancement ensures that user data remains secure and reduces the risk of unauthorized access.

### DIFF
--- a/ios/MullvadREST/RESTAuthorization.swift
+++ b/ios/MullvadREST/RESTAuthorization.swift
@@ -13,6 +13,7 @@ import Operations
 protocol RESTAuthorizationProvider {
     func getAuthorization(completion: @escaping (Result<REST.Authorization, Swift.Error>) -> Void)
         -> Cancellable
+    func invalidateToken(for accountNumber: String)
 }
 
 extension REST {
@@ -36,6 +37,10 @@ extension REST {
                     tokenData.accessToken
                 })
             }
+        }
+
+        func invalidateToken(for accountNumber: String) {
+            accessTokenManager.invalidateToken(for: accountNumber)
         }
     }
 }
--- a/ios/MullvadREST/RESTAccessTokenManager.swift
+++ b/ios/MullvadREST/RESTAccessTokenManager.swift
@@ -62,5 +62,14 @@ extension REST {
 
             return operation
         }
+
+        func invalidateToken(for accountNumber: String) {
+            operationQueue.addOperation(AsyncBlockOperation(dispatchQueue: dispatchQueue) { [weak self] in
+                guard let self else {
+                    return
+                }
+                self.tokens.removeValue(forKey: accountNumber)
+            })
+        }
     }
 }
--- a/ios/MullvadREST/RESTAccountsProxy.swift
+++ b/ios/MullvadREST/RESTAccountsProxy.swift
@@ -94,6 +94,7 @@ extension REST {
             retryStrategy: RetryStrategy,
             completion: @escaping CompletionHandler<Void>
         ) -> Cancellable {
+            let accessTokenProvider = createAuthorizationProvider(accountNumber: accountNumber)
             let requestHandler = AnyRequestHandler(createURLRequest: { endpoint, authorization in
                 var requestBuilder = try self.requestFactory.createRequestBuilder(
                     endpoint: endpoint,
@@ -104,13 +105,14 @@ extension REST {
                 requestBuilder.addValue(accountNumber, forHTTPHeaderField: "Mullvad-Account-Number")
 
                 return requestBuilder.getRequest()
-            }, authorizationProvider: createAuthorizationProvider(accountNumber: accountNumber))
+            }, authorizationProvider: accessTokenProvider)
 
             let responseHandler = AnyResponseHandler { response, data -> ResponseHandlerResult<Void> in
                 let statusCode = HTTPStatus(rawValue: response.statusCode)
 
                 switch statusCode {
                 case let statusCode where statusCode.isSuccess:
+                    accessTokenProvider.invalidateToken(for: accountNumber)
                     return .success(())
                 default:
                     return .unhandledResponse(


pinkisemils (commented):
As per the updated issue description, we should do this when the user logs out too. It's slightly less relevant there, as logging in will reset the token anyway, but there's no reason to keep the old auth token around if the user has logged out.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149#-:-Ne8Lqjx9BO27Vkt15g-:b-c056h4)* status: 0 of 3 files reviewed, all discussions resolved

ios/MullvadREST/RESTAuthorization.swift
```diff
@@ -37,6 +38,10 @@ extension REST {
                 })
             }
         }
+
+        func invalidateToken(for accountNumber: String) {
```
pronebird, line 42: AccessTokenProvider is bound by accountNumber so technically you don't have to accept `accountNumber` as input, you could just use `self.accountNumber`

pronebird (changes requested):
Looks good but `REST.AccessTokenProvider` can be improved a bit as per comment.

pronebird (approved):
:lgtm:

> Reviewed 1 of 3 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149#-:-Ne8ZE47-sa-xvBEm9Es:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @mojganii)

___
*[`ios/MullvadREST/RESTDevicesProxy.swift` line 163 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149#-Ne9LSw119X9a4hB3C2F:-Ne9LSw21eiseD8revNh:b-e9b3u6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68bedf00b2d1614c488a97329b859695d2806a12/ios/MullvadREST/RESTDevicesProxy.swift#L163)):*
> ```Swift
>         /// The completion handler will receive `true` if device is successfully removed,
>         /// otherwise `false` if device is not found or already removed.
>         public func deleteDevice(
> ```

I am afraid that removing device does not qualify as "log out", because we do that from other places too. It would be better to release all access tokens from `SetAccountOperation` when `.unset` action is given and perhaps at the very end after we delete a device.

In essence `REST.ProxyFactory` is the one creating `REST.AccessTokenManager` and then it holds it in `REST.ProxyFactory.configuration.accessTokenManager` property.

We could perhaps pass `REST.AccessTokenManager` all the way from `AppDelegate` into `TunnelManager` and then to `SetAccountOperation` so that we could tell it directly to remove all tokens right after logging out.

Note that the task card specifies that we should delete _all_ tokens on logout.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149)* status: all files reviewed, 1 unresolved discussion (waiting on @mojganii)

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii and @pronebird)

___
*[`ios/MullvadREST/RESTDevicesProxy.swift` line 163 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149#-Ne9LSw119X9a4hB3C2F:-NeCRPyy2yolhokwaHkq:b-jm49x3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68bedf00b2d1614c488a97329b859695d2806a12/ios/MullvadREST/RESTDevicesProxy.swift#L163)):*

I am afraid that removing device does not qualify as "log out", because we do that from other places too. It would be better to release all access tokens from `SetAccountOperation` when `.unset` action is given and perhaps at the very end after we delete a device.

In essence `REST.ProxyFactory` is the one creating `REST.AccessTokenManager` and then it holds it in `REST.ProxyFactory.configuration.accessTokenManager` property.

We could perhaps pass `REST.AccessTokenManager` all the way from `AppDelegate` into `TunnelManager` and then to `SetAccountOperation` so that we could tell it directly to remove all tokens right after logging out.

Note that the task card specifies that we should delete _all_ tokens on logout.
</blockquote></details>

I agree. 
I would suggest to revert the current PR, and go with @pronebird 's suggestion
___
*[`ios/MullvadREST/RESTDevicesProxy.swift` line 169 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149#-NeCR851BH2UQS5w_vor:-NeCR851BH2UQS5w_vos:bzgtkud) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68bedf00b2d1614c488a97329b859695d2806a12/ios/MullvadREST/RESTDevicesProxy.swift#L169)):*
> ```Swift
>             completion: @escaping CompletionHandler<Bool>
>         ) -> Cancellable {
>             let accessTokenProvider = createAuthorizationProvider(accountNumber: accountNumber)
> ```

This shouldn't be needed here.

rablador (commented):
> Reviewed 1 of 3 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii and @pronebird)

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/RESTDevicesProxy.swift` line 169 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149#-NeCR851BH2UQS5w_vor:-NeCWgt93tWB2LllHv94:b-69tv86) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68bedf00b2d1614c488a97329b859695d2806a12/ios/MullvadREST/RESTDevicesProxy.swift#L169)):*

This shouldn't be needed here.
</blockquote></details>

It's a matter of preference and technically REST can invalidate tokens for accounts that it removes, because it knows when it does that. 

It's just that there is no separate call for logout so in that regard certainly tunnel manager is the best equipped to clean _all_ tokens after the fact.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadREST/RESTDevicesProxy.swift` line 163 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149#-Ne9LSw119X9a4hB3C2F:-NeDzMSXCgVG6MzhXjwU:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68bedf00b2d1614c488a97329b859695d2806a12/ios/MullvadREST/RESTDevicesProxy.swift#L163)):*

I agree. 
I would suggest to revert the current PR, and go with @pronebird 's suggestion
</blockquote></details>

Done.
___
*[`ios/MullvadREST/RESTDevicesProxy.swift` line 169 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149#-NeCR851BH2UQS5w_vor:-NeDzNJEEL-NmMHZfxT4:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68bedf00b2d1614c488a97329b859695d2806a12/ios/MullvadREST/RESTDevicesProxy.swift#L169)):*

It's a matter of preference and technically REST can invalidate tokens for accounts that it removes, because it knows when it does that. 

It's just that there is no separate call for logout so in that regard certainly tunnel manager is the best equipped to clean _all_ tokens after the fact.
</blockquote></details>

Done.

pronebird (approved):
> Reviewed 8 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5149)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

PR #5160, opened 2023-09-18
Fix reliability and time consumption of testInitialConnectionTimings test

rablador:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160)

### DIFF
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -393,16 +393,17 @@
 		7A11DD0B2A9495D400098CD8 /* AppRoutes.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5802EBC42A8E44AC00E5CE4C /* AppRoutes.swift */; };
 		7A1A26432A2612AE00B978AA /* PaymentAlertPresenter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A1A26422A2612AE00B978AA /* PaymentAlertPresenter.swift */; };
 		7A21DACF2A30AA3700A787A9 /* UITextField+Appearance.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A21DACE2A30AA3700A787A9 /* UITextField+Appearance.swift */; };
+		7A2960F62A963F7500389B82 /* AlertCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A2960F52A963F7500389B82 /* AlertCoordinator.swift */; };
+		7A2960FD2A964BB700389B82 /* AlertPresentation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A2960FC2A964BB700389B82 /* AlertPresentation.swift */; };
 		7A307AD92A8CD8DA0017618B /* Duration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A307AD82A8CD8DA0017618B /* Duration.swift */; };
 		7A307ADB2A8F56DF0017618B /* Duration+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A307ADA2A8F56DF0017618B /* Duration+Extensions.swift */; };
 		7A33538F2AA9FF1600F0A71C /* SimulatorTunnelProviderManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A33538E2AA9FF1600F0A71C /* SimulatorTunnelProviderManager.swift */; };
 		7A3353912AAA014400F0A71C /* SimulatorVPNConnection.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A3353902AAA014400F0A71C /* SimulatorVPNConnection.swift */; };
 		7A3353932AAA089000F0A71C /* SimulatorTunnelInfo.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A3353922AAA089000F0A71C /* SimulatorTunnelInfo.swift */; };
 		7A3353972AAA0F8600F0A71C /* OperationBlockObserverSupport.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A3353962AAA0F8600F0A71C /* OperationBlockObserverSupport.swift */; };
-		7A2960F62A963F7500389B82 /* AlertCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A2960F52A963F7500389B82 /* AlertCoordinator.swift */; };
-		7A2960FD2A964BB700389B82 /* AlertPresentation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A2960FC2A964BB700389B82 /* AlertPresentation.swift */; };
 		7A42DEC92A05164100B209BE /* SettingsInputCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A42DEC82A05164100B209BE /* SettingsInputCell.swift */; };
 		7A42DECD2A09064C00B209BE /* SelectableSettingsCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A42DECC2A09064C00B209BE /* SelectableSettingsCell.swift */; };
+		7A6B4F592AB8412E00123853 /* TunnelMonitorTimings.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A6B4F582AB8412E00123853 /* TunnelMonitorTimings.swift */; };
 		7A7AD28D29DC677800480EF1 /* FirstTimeLaunch.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A7AD28C29DC677800480EF1 /* FirstTimeLaunch.swift */; };
 		7A7AD28F29DEDB1C00480EF1 /* SettingsHeaderView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A7AD28E29DEDB1C00480EF1 /* SettingsHeaderView.swift */; };
 		7A818F1F29F0305800C7F0F4 /* RootConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A818F1E29F0305800C7F0F4 /* RootConfiguration.swift */; };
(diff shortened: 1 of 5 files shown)


buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160)* status: 4 of 5 files reviewed, 5 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 406 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-NecaqhV-Fz9ibrmOU-v:-NecaqhV-Fz9ibrmOU-w:b-jiyt9m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b08ef2713a781dd78628eee2becec7a3b85476/ios/MullvadVPN.xcodeproj/project.pbxproj#L406)):*
> ```
> 		7A42DEC92A05164100B209BE /* SettingsInputCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A42DEC82A05164100B209BE /* SettingsInputCell.swift */; };
> 		7A42DECD2A09064C00B209BE /* SelectableSettingsCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A42DECC2A09064C00B209BE /* SelectableSettingsCell.swift */; };
> 		7A6B4F592AB8412E00123853 /* TunnelMonitorTimings.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A6B4F582AB8412E00123853 /* TunnelMonitorTimings.swift */; };
> ```

Please move the file after `TunnelMonitorProtocol.swift`
___
*[`ios/PacketTunnelCoreTests/TunnelMonitorTests.swift` line 62 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-NeccNoxCvdDmj6RJeJH:-NeccNoxCvdDmj6RJeJI:bnf869m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b08ef2713a781dd78628eee2becec7a3b85476/ios/PacketTunnelCoreTests/TunnelMonitorTests.swift#L62)):*
nit
I would say that the documentation bit is redundant here, we already established the timeouts when creating the `timings` variable. 

The part that says
> default first connection attempt starts at 4 second timeout, then doubles with each subsequent attempt, while being capped at 15s max.

Is probably more relevant in the `TunnelMonitor` file, or maybe the `TunnelMonitorTimings` file itself. 

But it's not a big deal either.
___
*[`ios/PacketTunnelCoreTests/TunnelMonitorTests.swift` line 73 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-NecdgdREGH0leJ7xDdp:-NecdgdSDCPwKlWT5rwG:bg69hgf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b08ef2713a781dd78628eee2becec7a3b85476/ios/PacketTunnelCoreTests/TunnelMonitorTests.swift#L73)):*
> ```Swift
> 
>         // Calculate the amount of time necessary to perform the test adding some leeway.
>         let timeout = expectedTimings.reduce(1000, +)
> ```

On my machine, this runs roughly in 900ms, so that leeway could work. But it can easily run slower in a VM / on the CI environment. 

Just today I had a test where it ran in 43.017 seconds instead of the original 42 seconds. 

So it would also fail here for the wrong reasons. 

I suggest being conserative and tripling that value for now, so 3 seconds expected at most. 
This should be low enough that we never wait too long, and high enough that if there is a deadlock, it doesn't make the test runner stall for too long before moving on.
___
*[`ios/PacketTunnelCoreTests/TunnelMonitorTests.swift` line 98 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-NecfPEN2bU-ARUmX-pP:-NecfPEN2bU-ARUmX-pQ:b-36uwj9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b08ef2713a781dd78628eee2becec7a3b85476/ios/PacketTunnelCoreTests/TunnelMonitorTests.swift#L98)):*
Currently it's printing 
> [1/3] connectionLost, time elapsed: 100s

You can use the following trick to avoid computing time manually
`let measuredTimeElapsed = Measurement(value: Double(timeElapsed), unit: UnitDuration.milliseconds)`

Then if you want to convert you can simply do `measuredTimeElapsed.converted(to: UnitDuration.seconds)` 
___
*[`ios/PacketTunnelCoreTests/TunnelMonitorTests.swift` line 148 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-NeciJTz8ShJcl7yAxk_:-NeciJTz8ShJcl7yAxka:b544t80) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b08ef2713a781dd78628eee2becec7a3b85476/ios/PacketTunnelCoreTests/TunnelMonitorTests.swift#L148)):*
> ```Swift
> 
>     private func roundToHundreds(_ value: Int) -> Int {
>         return (value / 100 * 100) + ((value % 100) / 50 * 100)
> ```

Couple of things come to mind here 

- This could probably be an extension on `Int` instead
- The first expression `(value / 100 * 100)` is redundant... Unless we're dealing with `Double` 

Which we are, because we're doing maths on `TimeInterval` 
Also I think the whole computation is unnecessary. 
Instead of rounding our values, we should just use the `accuracy` option of `XCTestEqual` 

If I do the following instead 
`let elapsed = Int(Date().timeIntervalSince(startDate) * 1000)
Then I can simply do 
`XCTAssertEqual(elapsed, expectedDuration, accuracy: 10, "Some fancy error message here")`


Without doing any complicated maths.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160)* status: 4 of 5 files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 406 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-NecaqhV-Fz9ibrmOU-v:-NegSLNVDS3p7-zUKWbD:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b08ef2713a781dd78628eee2becec7a3b85476/ios/MullvadVPN.xcodeproj/project.pbxproj#L406)):*

Please move the file after `TunnelMonitorProtocol.swift`
</blockquote></details>

Done.
___
*[`ios/PacketTunnelCoreTests/TunnelMonitorTests.swift` line 62 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-NeccNoxCvdDmj6RJeJH:-NegS0zJ3xD1xl7bY0ej:bss4a2b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b08ef2713a781dd78628eee2becec7a3b85476/ios/PacketTunnelCoreTests/TunnelMonitorTests.swift#L62)):*

nit
I would say that the documentation bit is redundant here, we already established the timeouts when creating the `timings` variable. 

The part that says
> default first connection attempt starts at 4 second timeout, then doubles with each subsequent attempt, while being capped at 15s max.

Is probably more relevant in the `TunnelMonitor` file, or maybe the `TunnelMonitorTimings` file itself. 

But it's not a big deal either.
</blockquote></details>

Removed.
___
*[`ios/PacketTunnelCoreTests/TunnelMonitorTests.swift` line 73 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-NecdgdREGH0leJ7xDdp:-NegQL5T0tOrIJlUdPXy:bkctd81) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b08ef2713a781dd78628eee2becec7a3b85476/ios/PacketTunnelCoreTests/TunnelMonitorTests.swift#L73)):*

On my machine, this runs roughly in 900ms, so that leeway could work. But it can easily run slower in a VM / on the CI environment. 

Just today I had a test where it ran in 43.017 seconds instead of the original 42 seconds. 

So it would also fail here for the wrong reasons. 

I suggest being conserative and tripling that value for now, so 3 seconds expected at most. 
This should be low enough that we never wait too long, and high enough that if there is a deadlock, it doesn't make the test runner stall for too long before moving on.
</blockquote></details>

Not that much, though? Perhaps upping to 1500? 
If each individual measure is about 100-300 millis, adding 3000 as leeway is a little overkill (not that it really matters). What I mean is that if we're going with your proposal below to use "accuracy" when asserting the time elapsed, I think 50ms would be good for that. Everything outside that will fail. So if all four asserts fail there's going to be 200ms added in total. Ie, if the tests takes more that 1100 millis it should fail anyway.
___
*[`ios/PacketTunnelCoreTests/TunnelMonitorTests.swift` line 98 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-NecfPEN2bU-ARUmX-pP:-NegPMj4EaAMzlAdeTAZ:b3ymb4w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b08ef2713a781dd78628eee2becec7a3b85476/ios/PacketTunnelCoreTests/TunnelMonitorTests.swift#L98)):*

Currently it's printing 
> [1/3] connectionLost, time elapsed: 100s

You can use the following trick to avoid computing time manually
`let measuredTimeElapsed = Measurement(value: Double(timeElapsed), unit: UnitDuration.milliseconds)`

Then if you want to convert you can simply do `measuredTimeElapsed.converted(to: UnitDuration.seconds)`
</blockquote></details>

Oh, simply forgot to add an "m". I like your suggestion, but perhaps a little overkill?
___
*[`ios/PacketTunnelCoreTests/TunnelMonitorTests.swift` line 148 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-NeciJTz8ShJcl7yAxk_:-NegJ20l342tBytIEoeT:b-qn5eat) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/85b08ef2713a781dd78628eee2becec7a3b85476/ios/PacketTunnelCoreTests/TunnelMonitorTests.swift#L148)):*
> The first expression `(value / 100 * 100)` is redundant...

Since we're dealing with integers there's some rounding going on here. Eg. 123 / 100 * 100 = 100.

The reason for normalising the values is that we - apart from the assertion - also print out what happens. So in `Expected to report connection loss after \(expectedDuration)s, instead reported it after \(timeElapsed)s` we probably want numbers that match.

I do agree that accuracy in XCTAssertEqual is generally a better choice (didn't know about it until now), but I'm not sure what to do about those logs. Maybe `Expected to report connection loss after *roughly* \(expectedDuration)s [...]` or something?

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160)* status: 4 of 5 files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/PacketTunnelCoreTests/TunnelMonitorTests.swift` line 98 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-NecfPEN2bU-ARUmX-pP:-Neh9Mvg33wp7sp0vLCr:b-tqkdqa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e8de84cee7c8329733d47ed36adc831586838cdc/ios/PacketTunnelCoreTests/TunnelMonitorTests.swift#L98)):*

Oh, simply forgot to add an "m". I like your suggestion, but perhaps a little overkill?
</blockquote></details>

Discussed offline and the decision was to leave it as-is

buggmagnet (approved):
nice job ! 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5160#-:-NehAXUhAmrvX8VH3xkv:b-v5yun4)* status: 4 of 5 files reviewed, all discussions resolved

PR #5187, opened 2023-09-22
Fix black screen bug if redeem voucher view if canceled to early

rablador:
If redeem voucher view is canceled before its presentation transition is done, the view will be removed prematurely and we end up with a black screen.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5187)

### DIFF
--- a/ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherViewController.swift
+++ b/ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherViewController.swift
@@ -30,6 +30,8 @@ class RedeemVoucherViewController: UIViewController, UINavigationControllerDeleg
     ) {
         self.contentView = RedeemVoucherContentView(configuration: configuration)
         self.interactor = interactor
+        self.contentView.isUserInteractionEnabled = false
+
         super.init(nibName: nil, bundle: nil)
     }
 
@@ -67,6 +69,8 @@ class RedeemVoucherViewController: UIViewController, UINavigationControllerDeleg
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
+
+        contentView.isUserInteractionEnabled = true
         contentView.isEditing = true
     }
 


buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5187)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherViewController.swift` line 93 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5187#-NfAMTgj7ziFiIQtd8fx:-NfAMTgj7ziFiIQtd8fy:b-8eyt4k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/668941893de3cf280ec592536d4e719a6ce78cde/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherViewController.swift#L93)):*
> ```Swift
> 
>             if contentView.isEditing {
> ```

If the intent is to forbid the user to ever exit the redeem voucher screen if they don't have a valid code, then I guess that works :upside_down_face: 

Jokes aside, this is making the `cancel` button do nothing when the text field in the `RedeemVoucherContentView` is not active. 

It seems that the root problem is not here, but in `RootContainerViewController` in how we handle view controllers, I still need to figure out what's exactly happening.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5187)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/View controllers/RedeemVoucher/RedeemVoucherViewController.swift` line 93 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5187#-NfAMTgj7ziFiIQtd8fx:-NfAS8Iq0Aol-5FmWgKY:b-uyza1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/668941893de3cf280ec592536d4e719a6ce78cde/ios/MullvadVPN/View%20controllers/RedeemVoucher/RedeemVoucherViewController.swift#L93)):*

If the intent is to forbid the user to ever exit the redeem voucher screen if they don't have a valid code, then I guess that works :upside_down_face: 

Jokes aside, this is making the `cancel` button do nothing when the text field in the `RedeemVoucherContentView` is not active. 

It seems that the root problem is not here, but in `RootContainerViewController` in how we handle view controllers, I still need to figure out what's exactly happening.
</blockquote></details>

It's perfect I think. 🫠
The problem is not in here as such, this was just an attempt to prevent a bad state. I'll keep digging a little.

buggmagnet (approved):
nice job ! 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5187#-:-NfAjn3a2I3ZT3HU8uD7:b-v5yun4)* status: 0 of 1 files reviewed, 1 unresolved discussion

PR #5188, opened 2023-09-22
Update tunnel manager to not return error for configuration loading

rablador:
It seems we have sufficient handling of cases where we cannot load the config we need. The current path does not allow the load config operation to return a failure/error, so we cannot end up with a thrown fatalError unless there's an implementation error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5188)

### DIFF
--- a/ios/MullvadVPN/AppDelegate.swift
+++ b/ios/MullvadVPN/AppDelegate.swift
@@ -434,12 +434,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     }
 
     private func getInitTunnelManagerOperation() -> AsyncBlockOperation {
+        // This operation is always treated as successful no matter what the configuration load yields.
+        // If the tunnel settings or device state can't be read, we simply pretend they are not there
+        // and leave user in logged out state. VPN config will be removed as well.
         AsyncBlockOperation(dispatchQueue: .main) { finish in
-            self.tunnelManager.loadConfiguration { error in
-                if let error {
-                    fatalError(error.localizedDescription)
-                }
-
+            self.tunnelManager.loadConfiguration {
                 self.logger.debug("Finished initialization.")
 
                 NotificationManager.shared.updateNotifications()
--- a/ios/MullvadVPN/TunnelManager/TunnelManager.swift
+++ b/ios/MullvadVPN/TunnelManager/TunnelManager.swift
@@ -168,7 +168,7 @@ final class TunnelManager: StorePaymentObserver {
 
     // MARK: - Public methods
 
-    func loadConfiguration(completionHandler: @escaping (Error?) -> Void) {
+    func loadConfiguration(completionHandler: @escaping () -> Void) {
         let loadTunnelOperation = LoadTunnelConfigurationOperation(
             dispatchQueue: internalQueue,
             interactor: TunnelInteractorProxy(self)
@@ -187,7 +187,7 @@ final class TunnelManager: StorePaymentObserver {
             self.updatePrivateKeyRotationTimer()
             self.startNetworkMonitor()
 
-            completionHandler(completion.error)
+            completionHandler()
         }
 
         loadTunnelOperation.addObserver(


pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5188#-:-NexrNsS4cOWrWVIQ2Ua:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5188)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5188#-NfAIGUACD9lRsQyVqTx:-NfAIGUACD9lRsQyVqTy:b3seu24) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8018102629433117d281c16a307a86f2b4cd480b/ios/MullvadVPN/AppDelegate.swift#L439)):*
> ```Swift
>         AsyncBlockOperation(dispatchQueue: .main) { finish in
>             self.tunnelManager.loadConfiguration { error in
>                 if let error {
> ```

I don't think we should be dismissing the error and completely ignore it. This means we will continue operation as if nothing happened, but we will most likely be running in an invalid state, which probably means nothing will work as expected. 

I suggest to log the error at least so future ourselves can have a clue about what's happening. 

The best would be to trigger a way to have `loadConfiguration` return an error in a real use scenario so we can understand which cases error out, and act accordingly.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5188)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5188#-NfAIGUACD9lRsQyVqTx:-NfAhJlk5bqPwEdRKBT3:b-eamcxg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8018102629433117d281c16a307a86f2b4cd480b/ios/MullvadVPN/AppDelegate.swift#L439)):*

I don't think we should be dismissing the error and completely ignore it. This means we will continue operation as if nothing happened, but we will most likely be running in an invalid state, which probably means nothing will work as expected. 

I suggest to log the error at least so future ourselves can have a clue about what's happening. 

The best would be to trigger a way to have `loadConfiguration` return an error in a real use scenario so we can understand which cases error out, and act accordingly.
</blockquote></details>

The thing is that this error is misleading to being with. We always treat the operation as successful here and handle state changes elsewhere. According to @pronebird, "worst case if we can’t read settings or device state, we simply pretend they are not there and leave user in logged out state". Also, VPN config will be removed, meaning you'll get another try and system settings prompt when trying to connect later again.

buggmagnet (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5188)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5188#-NfAIGUACD9lRsQyVqTx:-NfAkLP6EIkzu-YxS9e6:bizlhe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8018102629433117d281c16a307a86f2b4cd480b/ios/MullvadVPN/AppDelegate.swift#L439)):*

The thing is that this error is misleading to being with. We always treat the operation as successful here and handle state changes elsewhere. According to @pronebird, "worst case if we can’t read settings or device state, we simply pretend they are not there and leave user in logged out state". Also, VPN config will be removed, meaning you'll get another try and system settings prompt when trying to connect later again.
</blockquote></details>

All right fair enough, would you mind adding a summary of this as a comment in here and we can merge this ?

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5188)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 439 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5188#-NfAIGUACD9lRsQyVqTx:-NfAkccK9VTJv_o3Q7N5:b-x2qcd2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8018102629433117d281c16a307a86f2b4cd480b/ios/MullvadVPN/AppDelegate.swift#L439)):*

All right fair enough, would you mind adding a summary of this as a comment in here and we can merge this ?
</blockquote></details>

Sure thing. 👍

PR #5228, opened 2023-10-04
Log if Same IP is used when starting a tunnel

buggmagnet:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5228)

### DIFF
--- a/ios/PacketTunnel/PacketTunnelProvider.swift
+++ b/ios/PacketTunnel/PacketTunnelProvider.swift
@@ -208,6 +208,7 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
             let selectorResult = tunnelConfiguration.selectorResult
             self.selectorResult = selectorResult
             self.providerLogger.debug("Set tunnel relay to \(selectorResult.relay.hostname).")
+            self.logIfDeviceHasSameIP(than: tunnelConfiguration.wgTunnelConfig.interface.addresses)
 
             // Start tunnel.
             self.adapter.start(tunnelConfiguration: tunnelConfiguration.wgTunnelConfig) { error in
@@ -236,6 +237,16 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
                 }
             }
         }
+    }
+
+    private func logIfDeviceHasSameIP(than addresses: [IPAddressRange]) {
+        let hasIPv4SameAddress = addresses.compactMap { $0.address as? IPv4Address }
+            .contains { $0 == ApplicationConfiguration.sameIPv4 }
+        let hasIPv6SameAddress = addresses.compactMap { $0.address as? IPv6Address }
+            .contains { $0 == ApplicationConfiguration.sameIPv6 }
+
+        let isUsingSameIP = (hasIPv4SameAddress || hasIPv6SameAddress) ? "" : "NOT "
+        providerLogger.debug("Same IP is \(isUsingSameIP)being used")
     }
 
     override func stopTunnel(with reason: NEProviderStopReason, completionHandler: @escaping () -> Void) {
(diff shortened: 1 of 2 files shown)


ios/PacketTunnel/PacketTunnelProvider.swift
```diff
@@ -238,6 +239,16 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
...
     }
 
+    private func logIfDeviceHasSameIP(than addresses: [IPAddressRange]) {
+        let hasIPv4SameAddress = addresses.compactMap { $0.address as? IPv4Address }
+            .contains { $0 == ApplicationConfiguration.sameIPv4 }
+        let hasIPv6SameAddress = addresses.compactMap { $0.address as? IPv6Address }
+            .contains { $0 == ApplicationConfiguration.sameIPv6 }
+
+        let isNotUsingSameIP = (hasIPv4SameAddress || hasIPv6SameAddress) == false ? " NOT " : " "
+        providerLogger.debug("Same IP is\(isNotUsingSameIP)being used")
```
faern, line 249: On first sight this really looked like it would miss a space between "is" and "being" when Same IP is being used. I'm not going to nitpick about your codebase, but if I wrote it I would have `IsNotUsingSameIp` be empty when it is being used and otherwise "NOT " and adjust formatting string accordingly. :)

ios/Shared/ApplicationConfiguration.swift
```diff
@@ -37,4 +37,9 @@ enum ApplicationConfiguration {
 
     /// Maximum number of devices per account.
     static let maxAllowedDevices = 5
+
+    // swiftlint disable:force_cast
+    static let sameIPv4 = IPv4Address("10.127.255.254")!
+    static let sameIPv6 = IPv6Address("fc00:bbbb:bbbb:bb01:ffff:ffff:ffff:ffff")!
```
faern, line 43: Since the app in an ideal world should not need to know these IPs, I would add a comment, maybe a `// TODO` style comment, explaining that this is just temporary for aiding with debugging during migration to the Same IP system. And that it can be removed once the migration is done.

To help avoid it becoming arcane code living forever for no reason :)

faern (approved):

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5228)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[`ios/PacketTunnel/PacketTunnelProvider.swift` line 249 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5228#gh-1346427897:-Nfyh4DM8MAUGzPJ1yMu:bbrd6vf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8d710c06cb8896b8cf197dca602b56f805c3cee7/ios/PacketTunnel/PacketTunnelProvider.swift#L249)):*

On first sight this really looked like it would miss a space between "is" and "being" when Same IP is being used. I'm not going to nitpick about your codebase, but if I wrote it I would have `IsNotUsingSameIp` be empty when it is being used and otherwise "NOT " and adjust formatting string accordingly. :)
</blockquote></details>

You're right, I've reversed the check to make it (slightly) easier to read.
___
*[`ios/Shared/ApplicationConfiguration.swift` line 43 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5228#gh-1346429437:-Nfyh2bW8gGq_IzGfjCx:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8d710c06cb8896b8cf197dca602b56f805c3cee7/ios/Shared/ApplicationConfiguration.swift#L43)):*

Since the app in an ideal world should not need to know these IPs, I would add a comment, maybe a `// TODO` style comment, explaining that this is just temporary for aiding with debugging during migration to the Same IP system. And that it can be removed once the migration is done.

To help avoid it becoming arcane code living forever for no reason :)
</blockquote></details>

Done.

faern (approved):

PR #5236, opened 2023-10-05
Log if Same IP is being used or not

faern:
Desktop+Android equivalent of #5228. Logs when the known Same IP IPs are being used for a WireGuard tunnel. Should aid in debugging any issues that might arise during the rollout of Same IP.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5236)

### DIFF
--- a/mullvad-daemon/src/tunnel.rs
+++ b/mullvad-daemon/src/tunnel.rs
@@ -1,4 +1,10 @@
-use std::{future::Future, pin::Pin, sync::Arc};
+use std::{
+    future::Future,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr},
+    pin::Pin,
+    str::FromStr,
+    sync::Arc,
+};
 
 use tokio::sync::Mutex;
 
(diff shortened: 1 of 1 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5236)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`mullvad-daemon/src/tunnel.rs` line 33 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5236#-Nfz-WWT66Ocf0_oBQwq:-Nfz-WWT66Ocf0_oBQwr:b-hvf9p0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2d7ff6c16e3022d125f7dec77e915dde30d47350/mullvad-daemon/src/tunnel.rs#L33)):*
> ```XML
> /// peer.
> static SAME_IP_V4: Lazy<IpAddr> =
>     Lazy::new(|| Ipv4Addr::from_str("10.127.255.254").unwrap().into());
> ```

Nit: Might be a bit nicer to just use `Ipv{4,6}Addr::new` here since they're const.

PR #5254, opened 2023-10-09
Detect changes to interfaces in the dynamic store

dlon:
This PR makes a few changes:
* Use the primary interface as the "best (scoped) default route", or else the first active interface in the network service order. (Same as before, but the flow is much easier to understand.)
* Check whether the "best default route" has changed in response to any changes to network services in the dynamic store.
* Check whether the "best default route" has changed when the user tries to connect in the offline state.
* Log all the things.

Fix DES-398
Fix DES-396

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5254)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,9 @@ Line wrap the file at 100 chars.                                              Th
 #### Windows
 - Correctly detect whether OS is Windows Server (primarily for logging in daemon.log).
 
+#### macOS
+- Fix connectivity issues when switching between networks or disconnecting.
+
 
 ## [android/2023.6] - 2023-09-25
 ### Fixed
--- a/talpid-core/src/offline/macos.rs
+++ b/talpid-core/src/offline/macos.rs
@@ -3,9 +3,23 @@
 //! user from connecting to a relay.
 //!
 //! See [RouteManagerHandle::default_route_listener].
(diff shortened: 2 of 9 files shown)


pinkisemils (approved):
> Reviewed 2 of 7 files at r1, 4 of 6 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5254)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`talpid-core/src/offline/macos.rs` line 112 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5254#-NgNnpB69XPgSGOaw78Y:-NgNnpB69XPgSGOaw78Z:bvzsl38) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/00fa03a4ebcf942e3106f670bd32cf965f733048/talpid-core/src/offline/macos.rs#L112)):*
> ```XML
>                 route_event = route_listener.next() => {
>                     let Some(event) = route_event else {
>                         break;
> ```

Should it `break` or `continue` in this case? Feels like this could be coalesced with `let Some(state) = weak_state.ugprade()...` here.

PR #5259, opened 2023-10-10
Tunnel async actor tests

buggmagnet:
This PR enables testing for the following classes:
- `PacketTunnelActor`
- `TunnelManager`
- `AppMessageHandler`

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)


rablador (changes requested):
> Reviewed 13 of 13 files at r1, 57 of 57 files at r2, 13 of 13 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 175 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgO_nxl8XCLJS6Geba3:-NgO_nxl8XCLJS6Geba4:b-9gl8o8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6d2ed0d1ec8543b4944b96b0eb350d8a41abe4b8/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L175)):*
> ```Swift
>         // Wait for the connected state to happen so it doesn't get coalesced immediately after the call to `actor.stop`
>         actor.start(options: StartOptions(launchSource: .app))
>         await fulfillment(of: [connectedStateExpectation], timeout: 1)
> ```

We should probably use the actor.waitUntilConnected instead. Same for similar other places in the file.

pronebird (changes requested):
> Reviewed 12 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/PacketTunnelCore/Actor/State.swift` line 59 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgOh6B86MM6vcmWe44s:-NgOh6B86MM6vcmWe44t:bm6ucmy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6d2ed0d1ec8543b4944b96b0eb350d8a41abe4b8/ios/PacketTunnelCore/Actor/State.swift#L59)):*
> ```Swift
>  */
> public enum State: Equatable {
>     public static func == (lhs: State, rhs: State) -> Bool {
> ```

This is a very specialized equality check. How about we move that extension into tests and not implement it in the actor state itself? I'd also use a method instead because the `name` itself only takes into account the enum variant, so it's rather a similarity check based on variant ignoring all associated data which can be misleading if we were to use it in other contexts.
___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 175 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgO_nxl8XCLJS6Geba3:-NgOjCvIDmTEkyYqjOMU:bdiast5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6d2ed0d1ec8543b4944b96b0eb350d8a41abe4b8/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L175)):*

We should probably use the actor.waitUntilConnected instead. Same for similar other places in the file.
</blockquote></details>

`waitUntilConnected()` does not accept timeout, so it's perfectly normal to observe state directly.
___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 191 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgOYuzd1CM5lPdXnLas:-NgOYuzd1CM5lPdXnLat:bfeqt98) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6d2ed0d1ec8543b4944b96b0eb350d8a41abe4b8/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L191)):*
> ```Swift
>         actor.stop()
> 
>         switch await actor.state {
> ```

Since internally commands are handled in a separate detached task, picking `actor.state` right away could yield the initial value. It would be a bit more reliable to wait for commands in the channel to be consumed. 

Maybe we could add some internal method to wait for commands to be processed either inside of actor or directly on command channel. Also I don't think we need 3 calls in a row to prove the test case, one should be enough :-)
___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 237 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgOVj8975u-6vUYevCI:-NgOVj8975u-6vUYevCJ:b-lfej8m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6d2ed0d1ec8543b4944b96b0eb350d8a41abe4b8/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L237)):*
> ```Swift
>                         actor.setErrorState(reason: .readSettings)
>                     }
>                     task.cancel()
> ```

What's the purpose of this detached task, given that it is cancelled immediately and therefore will not advance past `Task.sleep()`?
___
*[`ios/PacketTunnelCoreTests/Mocks/DefaultPathObserverFake.swift` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgOeeT6BdvEutoxSF-z:-NgOeeT89i8hCCRwJ5aQ:b-n9sxx8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3bed86d9cccbf280c47a1b85fd9276981ddefec2/ios/PacketTunnelCoreTests/Mocks/DefaultPathObserverFake.swift#L26)):*
> ```Swift
>     private var stateLock = NSLock()
> 
>     private var defaultPathHandler: ((NetworkPath) -> Void)?
> ```

It's unsafe to access internal properties normally guarded behind the mutex lock. If you have a race within a test case, then tests could prove to be faulty.

I think it would be better to implement a stub accepting expectation and then fulfill it on `stop()` as I understand this is the requirement of one of tests.

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/PacketTunnelCoreTests/AppMessageHandlerTests.swift` line 16 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgOvJl24VAC59SkQeF2:-NgOvJl24VAC59SkQeF3:b-bx1a81) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b8b9da8eab8bee4996528d3e1ed9686f3c2ebede/ios/PacketTunnelCoreTests/AppMessageHandlerTests.swift#L16)):*
> ```Swift
> import XCTest
> 
> final class AppMessageHandlerTests: XCTestCase {
> ```

Looks like integration test. Maybe we should focus on testing inputs/outputs and signals as opposed to testing the reaction of underlying dependencies. I left a rather long comment on Slack to explain my point.

pronebird (changes requested):
> Reviewed 2 of 13 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 10 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadREST/RESTProxy.swift` line 13 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgSsh4t6S2j2NeVsW0y:-NgSsh4t6S2j2NeVsW0z:b3f36o6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b8b9da8eab8bee4996528d3e1ed9686f3c2ebede/ios/MullvadREST/RESTProxy.swift#L13)):*
> ```Swift
> import Operations
> 
> public typealias ProxyCompletionHandler<Success> = (Result<Success, Swift.Error>) -> Void
> ```

Why not define it under REST namespace where everything else is currently defined to keep it contained?
___
*[`ios/PacketTunnelCore/Actor/State.swift` line 196 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgSt-lHApX1fHJy5JeO:-NgSt-lHApX1fHJy5JeP:b-bnrwla) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b8b9da8eab8bee4996528d3e1ed9686f3c2ebede/ios/PacketTunnelCore/Actor/State.swift#L196)):*
> ```Swift
> 
>     /// Invalid public key.
>     case invalidPublicKey
> ```

Nit: Perhaps name this case so that it better reflects that the invalid public key is in the context of relay.
___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 260 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgTIkEk3P6-eVPJdZ4Y:-NgTIkEk3P6-eVPJdZ4Z:b-y1argg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8cf10809fb14fe526153c9d689c9fc14a9b5c1fd/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L260)):*
> ```Swift
>         let initialStateExpectation = expectation(description: "Expect initial state")
> 
>         stateSink = await actor.$state.receive(on: DispatchQueue.main).sink { newState in
> ```

`$state` always emits initial value first, if I remember that right. So you'd have to wait a bit to observe if there are any changes occur to state after `actor.reconnect()` is called. 

Maybe "inverse" expectation with a reasonable timeout can be used instead of `XCTFail()`.
___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 287 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgTT0nj5D6PxlofEMlM:-NgTT0nk9uuRb1naXQJh:b-jm355g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8cf10809fb14fe526153c9d689c9fc14a9b5c1fd/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L287)):*
> ```Swift
>         await fulfillment(of: [disconnectedStateExpectation], timeout: 1)
> 
>         await expect(.initial, on: actor) {
> ```

1. You probably mean to expect `.disconnected` state and not `.initial` in here. 
2. Since you assign a new sink within `expect()`,  the first value it emits should be current value (`.disconnected`) and this test will finish before `actor.reconnect()` executes. So again I can only think of using a short timeout and inverse expectation to ensure that actor remains in the same state.
___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 311 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgTUqFx7ReIDz4zBC79:-NgTUqFx7ReIDz4zBC7A:b9eznji) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8cf10809fb14fe526153c9d689c9fc14a9b5c1fd/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L311)):*
> ```Swift
>             connectedExpectation.fulfill()
>         }
>         actor.start(options: StartOptions(launchSource: .app))
> ```

Nit: Maybe we should provide a default value for `launchSource = .app` to avoid repeating it in every test
___
*[`ios/PacketTunnelCoreTests/Mocks/DefaultPathObserverFake.swift` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgOeeT6BdvEutoxSF-z:-NgTVAzmA-UU4Ux93oJB:b-a9vry9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3bed86d9cccbf280c47a1b85fd9276981ddefec2/ios/PacketTunnelCoreTests/Mocks/DefaultPathObserverFake.swift#L26)):*

Fair enough, we can do that.
</blockquote></details>

What if we make `onStart` and `onStop` immutable and fill them over `init()` to guarantee that they can be safely accessed without lock within the class? Or do we need to mutate them dynamically? 

It feels odd to have object half of which is protected by mutex lock and the other half isn't. If the object is expected to be used from multiple threads then we should protect all properties that are accessed concurrently. 

I know it's for tests only but as you said, we should give some love to tests, so that they serve us well for the years to come.

rablador (approved):
> Reviewed 2 of 2 files at r4, 1 of 85 files at r5, 7 of 76 files at r6, 6 of 19 files at r8, 53 of 66 files at r11, 19 of 28 files at r12, 11 of 11 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 8 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/PacketTunnelCore/Actor/State.swift` line 196 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgSt-lHApX1fHJy5JeO:-NgTz1ITAaxkwW5Koj7h:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b8b9da8eab8bee4996528d3e1ed9686f3c2ebede/ios/PacketTunnelCore/Actor/State.swift#L196)):*

Nit: Perhaps name this case so that it better reflects that the invalid public key is in the context of relay.
</blockquote></details>

Done.
___
*[`ios/PacketTunnelCoreTests/AppMessageHandlerTests.swift` line 16 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgOvJl24VAC59SkQeF2:-NgTyqyJ3Yo26q_mZbY2:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b8b9da8eab8bee4996528d3e1ed9686f3c2ebede/ios/PacketTunnelCoreTests/AppMessageHandlerTests.swift#L16)):*

Fair enough, I'll look at it.
</blockquote></details>

Done.

rablador (commented):
> Reviewed 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 8 unresolved discussions (waiting on @buggmagnet and @pronebird)

pronebird (changes requested):
> Reviewed 5 of 76 files at r6, 52 of 66 files at r11, 6 of 28 files at r12, 9 of 11 files at r13, 1 of 2 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 15 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/RESTAccessTokenManager.swift` line 14 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYOa34BiOb7WwYJW9a:-NgYOa34BiOb7WwYJW9b:bdizuhq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadREST/RESTAccessTokenManager.swift#L14)):*
> ```Swift
> import Operations
> 
> public protocol AccessTokenManagement {
> ```

Nit: it would be in spirit of how other protocols are named within the framework, i.e with "REST" prefix.
___
*[`ios/MullvadREST/RESTAccountsProxy.swift` line 12 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYOsFi0zSeP3gAg3fz:-NgYOsFi0zSeP3gAg3g-:b-85hti0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadREST/RESTAccountsProxy.swift#L12)):*
> ```Swift
> import MullvadTypes
> 
> public protocol AccountHandling {
> ```

Nit: same here `RESTAccountHandling` would be a bit less generic.
___
*[`ios/MullvadREST/RESTAuthenticationProxy.swift` line 60 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYPnt40W9bWDPEfDE4:-NgYPnt5ElDoptEo0KkU:bfv9zpk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadREST/RESTAuthenticationProxy.swift#L60)):*
> ```Swift
>     }
> 
>     public struct AccessTokenData: Decodable {
> ```

Nit: note that `AuthenticationProxy.getAccessToken()` is `internal`.
___
*[`ios/MullvadVPN/Classes/ObserverList.swift` line 26 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYQhuP8LnOGJobjcHI:-NgYQhuP8LnOGJobjcHJ:b-d7vlha) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/Classes/ObserverList.swift#L26)):*
> ```Swift
>     }
> 
>     static func == (lhs: WeakBox<T>, rhs: WeakBox<T>) -> Bool {
> ```

Nit: I don't mind this change but it is completely out of scope of this PR and could have been done in isolation.
___
*[`ios/MullvadVPN/TunnelManager/Tunnel.swift` line 24 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYSjP-8Mu7elL6q-he:-NgYSjP-8Mu7elL6q-hf:b-4e2eud) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/TunnelManager/Tunnel.swift#L24)):*
> ```Swift
> 
> protocol TunnelProtocol: AnyObject, Equatable {
>     func addObserver(_ observer: TunnelStatusObserver)
> ```

Do you mind to group observation methods a bit below around `addBlockObserver()`?
___
*[`ios/MullvadVPN/TunnelManager/TunnelStore.swift` line 98 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYU39u9j5_MAl0PelU:-NgYU39u9j5_MAl0PelV:b-eyfibh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/TunnelManager/TunnelStore.swift#L98)):*
> ```Swift
> 
>     private func handleTunnelStatus(tunnel: any TunnelProtocol, status: NEVPNStatus) {
>         guard let tunnel = tunnel as? Tunnel else { return }
> ```

This does not look right. This is defeats the purpose of having `TunnelProtocol`. Why do we need to cast?
___
*[`ios/MullvadVPN/TunnelManager/UIApplication+Extensions.swift` line 14 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYULCi5P7FP_gCiTKt:-NgYULCi5P7FP_gCiTKu:b-j49131) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/TunnelManager/UIApplication%2BExtensions.swift#L14)):*
> ```Swift
> import UIKit
> 
> public protocol UIApplicationProtocol {
> ```

Nit: you could name it something like `BackgroundTaskProviding` to better describe why we need this object in tunnel manager.
___
*[`ios/MullvadVPNTests/AccessTokenManager+Stubs.swift` line 18 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYVB790nE8VYs-dWb6:-NgYVB790nE8VYs-dWb7:b-537bcs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPNTests/AccessTokenManager%2BStubs.swift#L18)):*
> ```Swift
>         completionHandler: @escaping ProxyCompletionHandler<REST.AccessTokenData>
>     ) -> Cancellable {
>         AnyCancellable()
> ```

FYI I am not sure in which context this stub is used but absence of call to `completionHandler` will paralyze.

pronebird (changes requested):
> Reviewed 2 of 76 files at r6, 7 of 28 files at r12, 1 of 11 files at r13, 1 of 1 files at r14.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 17 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/PacketTunnelCore/Actor/State+Extensions.swift` line 40 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYaHHJ309ETZQO79CQ:-NgYaHHJ309ETZQO79CR:b-vu6fzu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/PacketTunnelCore/Actor/State%2BExtensions.swift#L40)):*
> ```Swift
>     }
> 
>     var packetTunnelStatus: PacketTunnelStatus {
> ```

Where was this copied from?

pronebird (commented):
> Reviewed 1 of 28 files at r12.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 17 unresolved discussions (waiting on @buggmagnet)

pronebird (commented):
> Reviewed 1 of 66 files at r11, 3 of 28 files at r12, 1 of 11 files at r13.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 17 unresolved discussions (waiting on @buggmagnet)

pronebird (commented):
> Reviewed 1 of 2 files at r15.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 18 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/PacketTunnelCore/Actor/State+Extensions.swift` line 2 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYfp9w9K40qzsilK_I:-NgYfp9w9K40qzsilK_J:b-qcatih) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/PacketTunnelCore/Actor/State%2BExtensions.swift#L2)):*
> ```Swift
> //
> //  State+.swift
> ```

Nit: does not reflect the name

buggmagnet (commented):
> Reviewed 2 of 76 files at r6, 52 of 66 files at r11, 13 of 28 files at r12, 1 of 1 files at r14, 2 of 2 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 13 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/MullvadREST/RESTAccessTokenManager.swift` line 14 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYOa34BiOb7WwYJW9a:-NgYe69W30uSe33Kfen-:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadREST/RESTAccessTokenManager.swift#L14)):*

Nit: it would be in spirit of how other protocols are named within the framework, i.e with "REST" prefix.
</blockquote></details>

Done
___
*[`ios/MullvadREST/RESTAuthenticationProxy.swift` line 60 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYPnt40W9bWDPEfDE4:-NgYgVgo31gprEmV933r:brxzhfk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadREST/RESTAuthenticationProxy.swift#L60)):*

Nit: note that `AuthenticationProxy.getAccessToken()` is `internal`.
</blockquote></details>

Good point, I've made it public
___
*[`ios/MullvadREST/RESTProxy.swift` line 13 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgSsh4t6S2j2NeVsW0y:-NgYkvpC0CERyPt88vTF:b-ve82h1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b8b9da8eab8bee4996528d3e1ed9686f3c2ebede/ios/MullvadREST/RESTProxy.swift#L13)):*

Why not define it under REST namespace where everything else is currently defined to keep it contained?
</blockquote></details>

Because it makes it really cumbersome to have to write 
`REST.ProxyCompletionHandler<REST.NewAccountData>`
___
*[`ios/MullvadVPN/Classes/ObserverList.swift` line 26 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYQhuP8LnOGJobjcHI:-NgYbpyw3Mhd-DtktNWH:bkdm38w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/Classes/ObserverList.swift#L26)):*

Nit: I don't mind this change but it is completely out of scope of this PR and could have been done in isolation.
</blockquote></details>

The reason I made this change is because I was hitting a compiler bug where it would choose this overload for an unrelated comparison and I couldn't get around it in `TunnelStore.handleTunnelStatus` 
___
*[`ios/MullvadVPN/TunnelManager/Tunnel.swift` line 24 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYSjP-8Mu7elL6q-he:-NgYdrz391Om5iFdxRHH:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/TunnelManager/Tunnel.swift#L24)):*

Do you mind to group observation methods a bit below around `addBlockObserver()`?
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/TunnelManager/TunnelStore.swift` line 98 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYU39u9j5_MAl0PelU:-NgYlqyS-Kam3LUzBUen:byg2pm8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/TunnelManager/TunnelStore.swift#L98)):*

This does not look right. This is defeats the purpose of having `TunnelProtocol`. Why do we need to cast?
</blockquote></details>

I replied to this in the comment about the `WeakBox` compiler issue I had to deal with.

Actually I realize that I had left an `Equatable` in the `TunnelProtocol` definition (which doesn't make sense)
and I didn't want to introduce a type eraser just for this case.

[This blog post](https://khawerkhaliq.com/blog/swift-protocols-equatable-part-two/) explains why better than I can
___
*[`ios/MullvadVPN/TunnelManager/UIApplication+Extensions.swift` line 14 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYULCi5P7FP_gCiTKt:-NgYp0Yl9TutZPaMynfU:b-w9ay0l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/TunnelManager/UIApplication%2BExtensions.swift#L14)):*

Nit: you could name it something like `BackgroundTaskProviding` to better describe why we need this object in tunnel manager.
</blockquote></details>

As you can see, I was not very inspired when naming things 😂🫠

I like your suggestion, I think however `BackgroundTaskProvider` has a better ring to it, thanks !
___
*[`ios/MullvadVPNTests/AccessTokenManager+Stubs.swift` line 18 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYVB790nE8VYs-dWb6:-NgYpdFdCufOnWbx77Km:b68zwtq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPNTests/AccessTokenManager%2BStubs.swift#L18)):*

FYI I am not sure in which context this stub is used but absence of call to `completionHandler` will paralyze.
</blockquote></details>

Yes, as discussed offline, this PR makes `TunnelManager` testable, but doesn't implement any test. 
We can make further modifications as needed when we will write tests for `TunnelManager` 
___
*[`ios/PacketTunnelCore/Actor/State+Extensions.swift` line 2 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYfp9w9K40qzsilK_I:-NgYsEF62pXhtMLxZpJs:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/PacketTunnelCore/Actor/State%2BExtensions.swift#L2)):*

Nit: does not reflect the name
</blockquote></details>

Done
___
*[`ios/PacketTunnelCore/Actor/State+Extensions.swift` line 40 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYaHHJ309ETZQO79CQ:-NgYslyP3_tvYSZsDKUT:b4mn35f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/PacketTunnelCore/Actor/State%2BExtensions.swift#L40)):*

Where was this copied from?
</blockquote></details>

I don't remember introducing this. Maybe @rablador knows ?
___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 260 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgTIkEk3P6-eVPJdZ4Y:-NgYswiC9k5ZYB0UqkD2:b-jygm85) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8cf10809fb14fe526153c9d689c9fc14a9b5c1fd/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L260)):*

`$state` always emits initial value first, if I remember that right. So you'd have to wait a bit to observe if there are any changes occur to state after `actor.reconnect()` is called. 

Maybe "inverse" expectation with a reasonable timeout can be used instead of `XCTFail()`.
</blockquote></details>

Hmmmm that's true, I hadn't realized, thanks for catching this. 
It looks like I need to rework quite a lot of assumptions I made in the `PacketTunnelActorTests` 
I will come back to this soon
___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 311 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgTUqFx7ReIDz4zBC79:-NgYtFf_4nJ_Ui_D-e41:b-896fjx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8cf10809fb14fe526153c9d689c9fc14a9b5c1fd/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L311)):*

Nit: Maybe we should provide a default value for `launchSource = .app` to avoid repeating it in every test
</blockquote></details>

Done
___
*[`ios/PacketTunnelCoreTests/Mocks/DefaultPathObserverFake.swift` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgOeeT6BdvEutoxSF-z:-NgU8Wik6kECjiJe2UCU:b-xmpabk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3bed86d9cccbf280c47a1b85fd9276981ddefec2/ios/PacketTunnelCoreTests/Mocks/DefaultPathObserverFake.swift#L26)):*

What if we make `onStart` and `onStop` immutable and fill them over `init()` to guarantee that they can be safely accessed without lock within the class? Or do we need to mutate them dynamically? 

It feels odd to have object half of which is protected by mutex lock and the other half isn't. If the object is expected to be used from multiple threads then we should protect all properties that are accessed concurrently. 

I know it's for tests only but as you said, we should give some love to tests, so that they serve us well for the years to come.
</blockquote></details>

Ah yes I forgot to put the `onStart` and `onStop` in the locking statement.
It's not horribly bad because it's unlikely that we'll use those in a multithreaded context, even in tests, but it's still bad practice. 

I will put those in the lock.

rablador (commented):
> Reviewed 25 of 25 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 13 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/PacketTunnelCore/Actor/State+Extensions.swift` line 40 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYaHHJ309ETZQO79CQ:-NgZ-9QU2vfGEgXZLWEX:b-g0xbfb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/PacketTunnelCore/Actor/State%2BExtensions.swift#L40)):*

I don't remember introducing this. Maybe @rablador knows ?
</blockquote></details>

There was a `State+Extensions.swift`in `PacketTunnel` module as well. This code bit was copied from there and that file was then deleted.

buggmagnet (commented):
> Reviewed 5 of 76 files at r6, 1 of 66 files at r11, 3 of 28 files at r12, 9 of 11 files at r13, 24 of 25 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 10 unresolved discussions (waiting on @pronebird)

pronebird (changes requested):
> Reviewed 5 of 25 files at r16.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/Classes/ObserverList.swift` line 26 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYQhuP8LnOGJobjcHI:-NgcRotw8kQ7n4_G8VbP:b-xchuhy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/Classes/ObserverList.swift#L26)):*

The reason I made this change is because I was hitting a compiler bug where it would choose this overload for an unrelated comparison and I couldn't get around it in `TunnelStore.handleTunnelStatus`
</blockquote></details>

I see. Thanks for explanation.
___
*[`ios/MullvadVPN/TunnelManager/TunnelStore.swift` line 98 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYU39u9j5_MAl0PelU:-NgcUR4_8TQIUuv1KwMU:b-1thqlg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/TunnelManager/TunnelStore.swift#L98)):*

I replied to this in the comment about the `WeakBox` compiler issue I had to deal with.

Actually I realize that I had left an `Equatable` in the `TunnelProtocol` definition (which doesn't make sense)
and I didn't want to introduce a type eraser just for this case.

[This blog post](https://khawerkhaliq.com/blog/swift-protocols-equatable-part-two/) explains why better than I can
</blockquote></details>

I still don't quite grasp the benefit of `TunnelProtocol` because:

- The internal implementation always downcasts to `Tunnel` otherwise it can't work with objects implementing `TunnelProtocol`.
- Internal factory method creates new `Tunnel` but then it erases the type by casting to `any TunnelProtocol`

Maybe what you meant to do was:

- Introduce associated type in `TunnelStoreProtocol`, i.e:

```
protocol TunnelStoreProtocol {
    associatedtype TunnelType: TunnelProtocol
    func createNewTunnel() -> TunnelType
    func getPersistentTunnels() -> [TunnelType]
}
```

- Declare `TunnelStore` with generic constraint, i.e:

```
class TunnelStore<TunnelType: TunnelProtocol>
```

Use `TunnelType` internally?
___
*[`ios/MullvadVPN/TunnelManager/UIApplication+Extensions.swift` line 14 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYULCi5P7FP_gCiTKt:-NgcS6R6DVzTRqeP_Y3F:bspaj2o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/TunnelManager/UIApplication%2BExtensions.swift#L14)):*

As you can see, I was not very inspired when naming things 😂🫠

I like your suggestion, I think however `BackgroundTaskProvider` has a better ring to it, thanks !
</blockquote></details>

Ha! Ok that looks better now. :+1: 
___
*[`ios/MullvadVPNTests/AccessTokenManager+Stubs.swift` line 18 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYVB790nE8VYs-dWb6:-NgcSEIJ14N7u_d6wX_5:b-88bvjv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPNTests/AccessTokenManager%2BStubs.swift#L18)):*

Yes, as discussed offline, this PR makes `TunnelManager` testable, but doesn't implement any test. 
We can make further modifications as needed when we will write tests for `TunnelManager`
</blockquote></details>

Fair.
___
*[`ios/PacketTunnelCore/Actor/State+Extensions.swift` line 2 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYfp9w9K40qzsilK_I:-NgcSKBo0K0_pZ-wDQEL:bk0oy87) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/PacketTunnelCore/Actor/State%2BExtensions.swift#L2)):*

Done
</blockquote></details>

Mhm probably should be `State+Extensions.swift`
___
*[`ios/PacketTunnelCore/Actor/State+Extensions.swift` line 40 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYaHHJ309ETZQO79CQ:-NgcSOm_1d5X2V_fbNrl:b-hnf61g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/PacketTunnelCore/Actor/State%2BExtensions.swift#L40)):*

There was a `State+Extensions.swift`in `PacketTunnel` module as well. This code bit was copied from there and that file was then deleted.
</blockquote></details>

Thanks. I guess since app message handler is a part of framework, the extensions should be available for it to function too.

pronebird (commented):
> Reviewed 20 of 25 files at r16.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 235 at r16](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgcqNQY1WpBPVsQy0Cz:-NgcqNQY1WpBPVsQy0D-:bks0liz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bbdd8438de1b3eaa559eff64158df48f66e42180/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L235)):*
> ```Swift
>                 switch newState {
>                 case .connecting:
>                     actor.setErrorState(reason: .readSettings)
> ```

Given that we use a channel behind:

- actor.start() and actor.stop() should schedule two commands right away.
- error state is scheduled after "stop" because `actor.setErrorState()` is called from the sink, so to verify that it doesn't do anything, you could introduce an inverse expectation:

```
let ex = expectation(description: "Should not enter blocked state")
ex.isInverted = true
```

then plug it in `error` case:

```
case .error:
  ex.fulfill()
```

Finally wait for all of them:

```
 await fulfillment(of: [connectingStateExpectation, disconnectedStateExpectation, ex], timeout: 1)
```

I believe that `fulfillment(of:)` would wait for 1 second before concluding that "ex" hasn't been fulfilled which would be correct given that we don't want it to be fulfilled, hence it's inverted expectation.

buggmagnet (commented):
> Reviewed 1 of 25 files at r16.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: all files reviewed, 6 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN/TunnelManager/TunnelStore.swift` line 98 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYU39u9j5_MAl0PelU:-NgdMioK7AM0XVqB6FV_:bsxi58r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/MullvadVPN/TunnelManager/TunnelStore.swift#L98)):*

I still don't quite grasp the benefit of `TunnelProtocol` because:

- The internal implementation always downcasts to `Tunnel` otherwise it can't work with objects implementing `TunnelProtocol`.
- Internal factory method creates new `Tunnel` but then it erases the type by casting to `any TunnelProtocol`

Maybe what you meant to do was:

- Introduce associated type in `TunnelStoreProtocol`, i.e:

```
protocol TunnelStoreProtocol {
    associatedtype TunnelType: TunnelProtocol
    func createNewTunnel() -> TunnelType
    func getPersistentTunnels() -> [TunnelType]
}
```

- Declare `TunnelStore` with generic constraint, i.e:

```
class TunnelStore<TunnelType: TunnelProtocol>
```

Use `TunnelType` internally?
</blockquote></details>

You're right that we want to use an erased type here. I didn't feel like doing it until I would have to write actual tests. I have fixed that now. 
___
*[`ios/PacketTunnelCore/Actor/State+Extensions.swift` line 2 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgYfp9w9K40qzsilK_I:-NgdS3CL-8ZHZ53IEnMS:b-3sqn5r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c64ef21a59f96e60f7f77d09355e0f8fbc739d3/ios/PacketTunnelCore/Actor/State%2BExtensions.swift#L2)):*

Mhm probably should be `State+Extensions.swift`
</blockquote></details>

Awkward... Done !
___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 287 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgTT0nj5D6PxlofEMlM:-NgdUbkM8L7XiJjYoIII:bk0jity) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8cf10809fb14fe526153c9d689c9fc14a9b5c1fd/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L287)):*

1. You probably mean to expect `.disconnected` state and not `.initial` in here. 
2. Since you assign a new sink within `expect()`,  the first value it emits should be current value (`.disconnected`) and this test will finish before `actor.reconnect()` executes. So again I can only think of using a short timeout and inverse expectation to ensure that actor remains in the same state.
</blockquote></details>

Done.
Thanks for catching that !
___
*[`ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift` line 235 at r16](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgcqNQY1WpBPVsQy0Cz:-NgdV6nVEkZNagcNnG9J:b-rdziv8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bbdd8438de1b3eaa559eff64158df48f66e42180/ios/PacketTunnelCoreTests/PacketTunnelActorTests.swift#L235)):*

Given that we use a channel behind:

- actor.start() and actor.stop() should schedule two commands right away.
- error state is scheduled after "stop" because `actor.setErrorState()` is called from the sink, so to verify that it doesn't do anything, you could introduce an inverse expectation:

```
let ex = expectation(description: "Should not enter blocked state")
ex.isInverted = true
```

then plug it in `error` case:

```
case .error:
  ex.fulfill()
```

Finally wait for all of them:

```
 await fulfillment(of: [connectingStateExpectation, disconnectedStateExpectation, ex], timeout: 1)
```

I believe that `fulfillment(of:)` would wait for 1 second before concluding that "ex" hasn't been fulfilled which would be correct given that we don't want it to be fulfilled, hence it's inverted expectation.
</blockquote></details>

We can also wait for 100ms _after_ we got the stopped expectation. Since the inverted expectation will blow up if fulfilled, once we have reached the `stopped` state, we know we don't need to wait any longer.

buggmagnet (commented):
> Reviewed 1 of 76 files at r18, 30 of 58 files at r19, 13 of 20 files at r20, 33 of 33 files at r21, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: 14 of 91 files reviewed, 2 unresolved discussions

___
*[`ios/MullvadVPNTests/AddressCacheTests.swift` line 50 at r21](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgrkoyS41BM3xWYcr6c:-NgrkoyS41BM3xWYcr6d:b4j48gs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68b289326859efe95bfe183d517900c47dac4385/ios/MullvadVPNTests/AddressCacheTests.swift#L50)):*
> ```Swift
>         let timeDifference = dateAfterUpdate.timeIntervalSince(dateBeforeUpdate)
> 
>         XCTAssertNotEqual(0.0, timeDifference)
> ```

I got fed up of this test failing frequently on my machine (the test was incorrect to begin with) 
This should fix it for good. (famous last words)
___
*[`ios/PacketTunnelCoreTests/ActorTests.swift` line 19 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259#-NgrlUfK15L2Oc3q7cxR:-NgrlUfK15L2Oc3q7cxS:b-crplcm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e739e2df248bd27b26e266169f38576c2879ceb1/ios/PacketTunnelCoreTests/ActorTests.swift#L19)):*
> ```Swift
> import XCTest
> 
> final class ActorTests: XCTestCase {
> ```

This file got back here by mistake, I will delete it.

pronebird (commented):
> Reviewed 1 of 76 files at r18, 3 of 20 files at r20.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5259)* status: 18 of 91 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

PR #5273, opened 2023-10-11
Ensure atomicity between (re)connection attempts

pronebird:
Currently reconnection is a two fold operation:
- First we increment attempt counter
- Then schedule command for reconnect to leverage coalescing system.

This approach produces two individual changes to `state` which can be misleading when observed from the outside. For instance:

-  If the tunnel is currently in `.connected` state, the first mutation increments attempt counter updating it to `1`.
- The observer then receives a `.connected` state with attempt counter updated to `1`. 
- The second mutation changes `state` to `.reconnecting` with attempt counter set to `1`.
- The observer then receives a `.reconnecting` state with attempt counter set to `1`.

Because `connected`, `reconnecting`, `connecting` were handled the same way, the tunnel losing connectivity also starts connection attempt counter at `1` which is suboptimal.

This PR aims to fix the aforementioned issues by introducing `ReconnectReason` which is passed along with `reconnect()` command in place of `shouldStopTunnelMonitor` parameter. `tryStart()` then is responsible for incrementing the attempt counter following these rules:

- The counter is not incremented If current state is `.connected`. That's because the tunnel is going to transition to `.reconnecting` phase and attempt counter should start at zero during the first attempt, then increment with each subsequent attempt.
- The counter is incremented if current state is either `.connecting` or `.reconnecting` as normal.

Two tests added to ensure that state transitions for (re)connection attempts happen as expected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5273)

### DIFF
--- a/ios/PacketTunnelCore/Actor/Actor+ConnectionMonitoring.swift
+++ b/ios/PacketTunnelCore/Actor/Actor+ConnectionMonitoring.swift
@@ -47,28 +47,14 @@ extension PacketTunnelActor {
         }
     }
 
-    /// Increment connection attempt counter and reconnect the tunnel.
+    /// Tell the tunnel to reconnect providing the correct reason to ensure that the attempt counter is incremented before reconnect.
     private func onHandleConnectionRecovery() async {
         switch state {
-        case var .connecting(connState):
-            connState.incrementAttemptCount()
-            state = .connecting(connState)
-
-        case var .reconnecting(connState):
-            connState.incrementAttemptCount()
-            state = .reconnecting(connState)
-
-        case var .connected(connState):
-            connState.incrementAttemptCount()
-            state = .connected(connState)
+        case .connecting, .reconnecting, .connected:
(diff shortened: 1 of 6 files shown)


rablador (approved):
> Reviewed 3 of 6 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5273)* status: :shipit: complete! all files reviewed, all discussions resolved

buggmagnet (approved):
:lgtm:

> Reviewed 6 of 6 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5273#-:-Ngd4HNm87R-kUimqYro:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5302, opened 2023-10-16
Introduce observed state IOS-343

pronebird:
As we discussed the other day, actor `State` contains class references and some internal data which should not be retained outside of actor nor of any interest to client code.

This PR adds `ObservedState` which is a value type with all private data stripped out. It's also implements `Equatable, Codable`. In essence it mirrors `State` stripping metadata that should not be shared with the client code.

Worth to note: this work is a stepping stone to sharing actor state over IPC because since iOS 16.5 Apple introduced timeout for `NE.startTunnel()` and therefore we will have to switch the tunnel into `connected` state immediately upon start to avoid being forcefully stopped by the system while in reconnection loop. 

On the tunnel manager side we'd have to work with actor state directly and therefore I think that `ObservedState` shared over IPC will supersede `PacketTunnelStatus` entirely, `SelectedRelay` will supersede `PacketTunnelRelay`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302)

### DIFF
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -161,6 +161,7 @@
 		5878F50029CDA742003D4BE2 /* UIView+AutoLayoutBuilder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5878F4FF29CDA742003D4BE2 /* UIView+AutoLayoutBuilder.swift */; };
 		587988C728A2A01F00E3DF54 /* AccountDataThrottling.swift in Sources */ = {isa = PBXBuildFile; fileRef = 587988C628A2A01F00E3DF54 /* AccountDataThrottling.swift */; };
 		587A01FC23F1F0BE00B68763 /* SimulatorTunnelProviderHost.swift in Sources */ = {isa = PBXBuildFile; fileRef = 587A01FB23F1F0BE00B68763 /* SimulatorTunnelProviderHost.swift */; };
+		587A5E522ADD7569003A70F1 /* ObservedState+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 587A5E512ADD7569003A70F1 /* ObservedState+Extensions.swift */; };
 		587B7536266528A200DEF7E9 /* NotificationManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 587B7535266528A200DEF7E9 /* NotificationManager.swift */; };
 		587B753B2666467500DEF7E9 /* NotificationBannerView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 587B753A2666467500DEF7E9 /* NotificationBannerView.swift */; };
 		587B753D2666468F00DEF7E9 /* NotificationController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 587B753C2666468F00DEF7E9 /* NotificationController.swift */; };
@@ -292,6 +293,7 @@
 		58CE5E66224146200008646E /* LoginViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58CE5E65224146200008646E /* LoginViewController.swift */; };
 		58CE5E6B224146210008646E /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 58CE5E6A224146210008646E /* Assets.xcassets */; };
 		58CE5E81224146470008646E /* PacketTunnel.appex in Embed Foundation Extensions */ = {isa = PBXBuildFile; fileRef = 58CE5E79224146470008646E /* PacketTunnel.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
+		58CF95A22AD6F35800B59F5D /* ObservedState.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58CF95A12AD6F35800B59F5D /* ObservedState.swift */; };
 		58D0C79E23F1CEBA00FE9BA7 /* SnapshotHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58D0C79D23F1CEBA00FE9BA7 /* SnapshotHelper.swift */; };
 		58D0C7A223F1CECF00FE9BA7 /* MullvadVPNScreenshots.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58D0C7A023F1CECF00FE9BA7 /* MullvadVPNScreenshots.swift */; };
 		58D223A8294C8A490029F5F8 /* Operations.h in Headers */ = {isa = PBXBuildFile; fileRef = 58D223A7294C8A490029F5F8 /* Operations.h */; settings = {ATTRIBUTES = (Public, ); }; };
@@ -1315,6 +1317,7 @@
 		587988C628A2A01F00E3DF54 /* AccountDataThrottling.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountDataThrottling.swift; sourceTree = "<group>"; };
 		58799A352A84FC9F007BE51F /* PingerProtocol.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PingerProtocol.swift; sourceTree = "<group>"; };
 		587A01FB23F1F0BE00B68763 /* SimulatorTunnelProviderHost.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SimulatorTunnelProviderHost.swift; sourceTree = "<group>"; };
+		587A5E512ADD7569003A70F1 /* ObservedState+Extensions.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "ObservedState+Extensions.swift"; sourceTree = "<group>"; };
 		587AD7C523421D7000E93A53 /* TunnelSettingsV1.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = TunnelSettingsV1.swift; sourceTree = "<group>"; };
 		587B7535266528A200DEF7E9 /* NotificationManager.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NotificationManager.swift; sourceTree = "<group>"; };
 		587B753A2666467500DEF7E9 /* NotificationBannerView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NotificationBannerView.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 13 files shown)


rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302)* status: all files reviewed, 1 unresolved discussion (waiting on @pronebird)

___
*[`ios/PacketTunnelCore/Actor/ObservedState+Extensions.swift` line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-NgsrFnVBJwHiJkW6Jeq:-NgsrFnVBJwHiJkW6Jer:b49acod) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/eeacdb412072deaa056674068adbb67d8b2cccb9/ios/PacketTunnelCore/Actor/ObservedState%2BExtensions.swift#L27)):*
> ```Swift
>                 status.isNetworkReachable = false
>             case .undetermined:
>                 // TODO: fix me
> ```

This should either be fixed or removed.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/PacketTunnelCore/Actor/ObservedState+Extensions.swift` line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-NgsrFnVBJwHiJkW6Jeq:-NgwU7MX0XfB1AejUcT7:bnxkswf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/eeacdb412072deaa056674068adbb67d8b2cccb9/ios/PacketTunnelCore/Actor/ObservedState%2BExtensions.swift#L27)):*

This should either be fixed or removed.
</blockquote></details>

If `PacketTunnelStatus` is indeed going away, then we don't have to take action immediately. Let's iterate.

buggmagnet (changes requested):
> Reviewed 12 of 13 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302)* status: all files reviewed, 4 unresolved discussions (waiting on @pronebird)

___
*[`ios/PacketTunnelCore/Actor/ObservedState+Extensions.swift` line 27 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh0nPO9CfaQD4uyc5we:-Nh0nPOA7fHPRiV2tZHJ:b-8d6g7g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fa30bbc04a6d105babdf144c9fa4dba6fd4ab15e/ios/PacketTunnelCore/Actor/ObservedState%2BExtensions.swift#L27)):*
> ```Swift
>                 status.isNetworkReachable = false
>             case .undetermined:
>                 // TODO: fix me
> ```

Should this TODO still be here ?
___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActor+Extensions.swift` line 15 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh0nmOd-oCu-f5f6Y6t:-Nh0nmOd-oCu-f5f6Y6u:buy08w8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fa30bbc04a6d105babdf144c9fa4dba6fd4ab15e/ios/PacketTunnelCore/Actor/PacketTunnelActor%2BExtensions.swift#L15)):*
> ```Swift
>     /// Note that the stream yields current value when created.
>     public var observedStates: AsyncStream<ObservedState> {
>         return AsyncStream { continuation in
> ```

`return` can be omitted here
___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActor+Extensions.swift` line 15 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh11__66fBCC0AbglHG:-Nh11__66fBCC0AbglHH:b9595j5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fa30bbc04a6d105babdf144c9fa4dba6fd4ab15e/ios/PacketTunnelCore/Actor/PacketTunnelActor%2BExtensions.swift#L15)):*
> ```Swift
>     /// Note that the stream yields current value when created.
>     public var observedStates: AsyncStream<ObservedState> {
>         return AsyncStream { continuation in
> ```

I don't know how I feel about this.
Technically there is a chance that `state` and `observedState` mismatch since the observed values are produced using an `AsyncStream` which means we cannot guarantee that `state` and `Observed` state are always matching.

Can we perhaps put an assert here to verify that they are matching ? Or maybe a warning log message ?
Maybe we could write a test that guarantees that we cannot have mismatching states, or that if we do, it's okay ?

I'm more erring on the side of caution here.
___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActor+Extensions.swift` line 16 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh0nxUGENQVN3sAc-pa:-Nh0nxUGENQVN3sAc-pb:bsy33tj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fa30bbc04a6d105babdf144c9fa4dba6fd4ab15e/ios/PacketTunnelCore/Actor/PacketTunnelActor%2BExtensions.swift#L16)):*
> ```Swift
>     public var observedStates: AsyncStream<ObservedState> {
>         return AsyncStream { continuation in
>             let cancellable = self.$observedState.sink { newState in
> ```

`self` should be removed

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/PacketTunnelCore/Actor/ObservedState+Extensions.swift` line 27 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh0nPO9CfaQD4uyc5we:-Nh1H5N28Cc4K5idRd5X:b9pgb2p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fa30bbc04a6d105babdf144c9fa4dba6fd4ab15e/ios/PacketTunnelCore/Actor/ObservedState%2BExtensions.swift#L27)):*

Should this TODO still be here ?
</blockquote></details>

Yes, what changed?
___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActor+Extensions.swift` line 15 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh11__66fBCC0AbglHG:-Nh1HiAs1Vc6VAvN30AW:b9awgpb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fa30bbc04a6d105babdf144c9fa4dba6fd4ab15e/ios/PacketTunnelCore/Actor/PacketTunnelActor%2BExtensions.swift#L15)):*

I don't know how I feel about this.
Technically there is a chance that `state` and `observedState` mismatch since the observed values are produced using an `AsyncStream` which means we cannot guarantee that `state` and `Observed` state are always matching.

Can we perhaps put an assert here to verify that they are matching ? Or maybe a warning log message ?
Maybe we could write a test that guarantees that we cannot have mismatching states, or that if we do, it's okay ?

I'm more erring on the side of caution here.
</blockquote></details>

How come? `ObservedState` is produced from `didSet` in `var state: State` and `@Published` property should immediately emit new value.
___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActor+Extensions.swift` line 15 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh0nmOd-oCu-f5f6Y6t:-Nh1Lp6Q5k5e5NQChAHp:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fa30bbc04a6d105babdf144c9fa4dba6fd4ab15e/ios/PacketTunnelCore/Actor/PacketTunnelActor%2BExtensions.swift#L15)):*

`return` can be omitted here
</blockquote></details>

Done.
___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActor+Extensions.swift` line 16 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh0nxUGENQVN3sAc-pa:-Nh1Loc47NdCfm95_ES8:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fa30bbc04a6d105babdf144c9fa4dba6fd4ab15e/ios/PacketTunnelCore/Actor/PacketTunnelActor%2BExtensions.swift#L16)):*

`self` should be removed
</blockquote></details>

Done.

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302)* status: 12 of 13 files reviewed, 2 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/PacketTunnelCore/Actor/ObservedState+Extensions.swift` line 27 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh0nPO9CfaQD4uyc5we:-Nh1OZtVDlrx6VArSmkS:brlqaj7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c396de9356e4c8bfbd635daf8359b7e163c8bd48/ios/PacketTunnelCore/Actor/ObservedState%2BExtensions.swift#L27)):*

Yes, what changed?
</blockquote></details>

Just wanted to check, that's all 
___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActor+Extensions.swift` line 15 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh11__66fBCC0AbglHG:-Nh1TgW7E7jmEx9o6LxS:b-opk145) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c396de9356e4c8bfbd635daf8359b7e163c8bd48/ios/PacketTunnelCore/Actor/PacketTunnelActor%2BExtensions.swift#L15)):*

How come? `ObservedState` is produced from `didSet` in `var state: State` and `@Published` property should immediately emit new value.
</blockquote></details>

Imagine the following scenario:
```swift
for await newState in await actor.observedStates {
    // Do expensive and long computation here
}
```
While the long and expensive computation runs here, the `state` can change any number of times. 
Those changes won't be perceived before the long computation happened.

Since we don't coalesce the states in any way, that code depending on `observedStates` cannot assume safely that the observed state will be the actor's real state.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActor+Extensions.swift` line 15 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh11__66fBCC0AbglHG:-Nh1ehKpDAkQHM413nwO:b-rfmfpf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c396de9356e4c8bfbd635daf8359b7e163c8bd48/ios/PacketTunnelCore/Actor/PacketTunnelActor%2BExtensions.swift#L15)):*

Imagine the following scenario:
```swift
for await newState in await actor.observedStates {
    // Do expensive and long computation here
}
```
While the long and expensive computation runs here, the `state` can change any number of times. 
Those changes won't be perceived before the long computation happened.

Since we don't coalesce the states in any way, that code depending on `observedStates` cannot assume safely that the observed state will be the actor's real state.
</blockquote></details>

This behavior matches what we had when accessing `var states: AsyncStream<State>` directly. 

In practice packet tunnel does not suspend within the iteration and the `state` does not change that frequently. 

If packet tunnel ever needs to suspend within iteration then I think it would be appropriate to consider one of the following: 

- Move long-running tasks from within the body of the loop into background task.
- Filter intermediate values by using`.bufferingNewest(1)` buffering strategy with `AsyncStream`.

The latest state can still be accessed directly by calling into `await actor.observedState` which I think `AppMessageHandler` will be returning in request to get tunnel status in the work that @rablador  is doing today.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302)* status: all files reviewed, 2 unresolved discussions (waiting on @rablador)

___
*[`ios/PacketTunnelCore/Actor/ObservedState+Extensions.swift` line 27 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh0nPO9CfaQD4uyc5we:-Nh1q3246Tq_GsPRIaEM:bj2t74m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c396de9356e4c8bfbd635daf8359b7e163c8bd48/ios/PacketTunnelCore/Actor/ObservedState%2BExtensions.swift#L27)):*

Just wanted to check, that's all
</blockquote></details>

I see. @rablador will be removing this in his work. I am keeping it as scoped as possible.

buggmagnet (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302)* status: all files reviewed, 2 unresolved discussions (waiting on @rablador)

___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActor+Extensions.swift` line 15 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5302#-Nh11__66fBCC0AbglHG:-Nh2DSzY7U2rvxFyl9ul:bqy67xj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c396de9356e4c8bfbd635daf8359b7e163c8bd48/ios/PacketTunnelCore/Actor/PacketTunnelActor%2BExtensions.swift#L15)):*

This behavior matches what we had when accessing `var states: AsyncStream<State>` directly. 

In practice packet tunnel does not suspend within the iteration and the `state` does not change that frequently. 

If packet tunnel ever needs to suspend within iteration then I think it would be appropriate to consider one of the following: 

- Move long-running tasks from within the body of the loop into background task.
- Filter intermediate values by using`.bufferingNewest(1)` buffering strategy with `AsyncStream`.

The latest state can still be accessed directly by calling into `await actor.observedState` which I think `AppMessageHandler` will be returning in request to get tunnel status in the work that @rablador  is doing today.
</blockquote></details>

Fair enough

PR #5324, opened 2023-10-18
Fix DNS content blockers list order for consistency across platform

wavygecko:
This seems to be the only place where gambling comes before adult content, so this PR fixes that to be consistent with the rest of the Mullvad platform.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5324)

### DIFF
--- a/gui/src/renderer/components/VpnSettings.tsx
+++ b/gui/src/renderer/components/VpnSettings.tsx
@@ -375,31 +375,6 @@ function BlockMalware() {
   );
 }
 
-function BlockGambling() {
-  const [dns, setBlockGambling] = useDns('blockGambling');
-
-  return (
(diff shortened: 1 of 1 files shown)


raksooo:
The current design is in line with the Android app and the design specifications for the app. I'll raise the question with the iOS and web developers to sync this.

wavygecko:
Ah, okay. I (and maybe iOS team, too?) based the order off of the [DNS server IP addresses](https://github.com/mullvad/dns-blocklists#custom-dns-entries-for-use-with-our-vpn-service) order.

faern:
The order of the gambling blocker seems to be consistent on all platforms now. So I'm closing this.

PR #5329, opened 2023-10-19
Prevent iOS from stopping the tunnel if it remains in connecting state for too long

rablador:
Since iOS 16.5 NE.startTunnel() has a 60s timeout to return from that method. To handle this the idea is to never keep the VPN extension in connecting state and instead return immediately from NE.startTunnel() to make the system think that the tunnel is established to prevent being shutdown by a system timer.

Internally the actor will use its own state and will remain in the actual state that we expect the tunnel to be in. Only NEVPNStatus will be a lie.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)


rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: 0 of 14 files reviewed, 1 unresolved discussion

___
*[`ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift` line 63 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh6ixyBAIkio4gmrooW:-Nh6ixyBAIkio4gmrooX:b-rwvg7n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b7d99c6aa6b2b3458c46890d5dad98e384cb0186/ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift#L63)):*
> ```Swift
>                     return .error(blockedState.reason)
>                 default:
>                     return .none
> ```

What to return here?

buggmagnet (approved):
> Reviewed 13 of 14 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 6 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift` line 94 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh7F2R-3UNAq27RaQLg:-Nh7F2R-3UNAq27RaQLh:b-3fez1q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/791ce28d9d71780b16dbae1335a42d660889e5bd/ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift#L94)):*
> ```Swift
>         default:
>             interactor.updateTunnelStatus { tunnelStatus in
>                 let isNetworkReachable = tunnelStatus.observedState.connectionState?.isNetworkReachable ?? false
> ```

nit
Should we care about reachability when we are disconnecting, since we will soon be disconnected and shut down the VPN ?
Maybe we should just force `tunnelStatus.state = .disconnecting(nothing)` here.
What do you think @pronebird  ?
___
*[`ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift` line 135 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh7CaFECRVBZ8RYFpd2:-Nh7CaFECRVBZ8RYFpd3:bn60ho2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/791ce28d9d71780b16dbae1335a42d660889e5bd/ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift#L135)):*
> ```Swift
>         mapToState: @escaping (ObservedState) -> TunnelState?
>     ) {
>         request = tunnel.getTunnelStatus { [weak self] completion in
> ```

nit
let's rename this `result`
___
*[`ios/MullvadVPN/View controllers/Tunnel/TunnelControlView.swift` line 226 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh7G6BJCl4ZzONycIj_:-Nh7G6BJCl4ZzONycIja:b-yzqu1r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/791ce28d9d71780b16dbae1335a42d660889e5bd/ios/MullvadVPN/View%20controllers/Tunnel/TunnelControlView.swift#L226)):*
> ```Swift
> 
>             connectionPanel.dataSource = ConnectionPanelData(
>                 inAddress: "\(tunnelRelay.endpoint.ipv4Relay) UDP",
> ```

Nit
Note for future @buggmagnet , should we change this dynamically based on whether we're obfuscating the connection via TCP ?

Yes we should. That's what the desktop does.
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 105 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh7OYM5CoEuUaCZK33_:-Nh7OYM5CoEuUaCZK33a:bz2aj2p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/791ce28d9d71780b16dbae1335a42d660889e5bd/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L105)):*
> ```Swift
>         for await state in await actor.observedStates {
>             switch state {
>             case .connected:
> ```

As discussed with @rablador, we established that we should probably respect the previous behaviour where we were also returning normally in case we were already disconnected here.
___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 59 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh7PETd3QUqwMONk68i:-Nh7PETd3QUqwMONk68j:bgl66nh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/791ce28d9d71780b16dbae1335a42d660889e5bd/ios/PacketTunnelCore/Actor/ObservedState.swift#L59)):*
> ```Swift
>     }
> 
>     public init(
> ```

This is only used for the simulator, we should probably hide this behind an `#if targetEnvironment(simulator)` statement

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift` line 135 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh7CaFECRVBZ8RYFpd2:-Nh7m7mT-1WqolWa20g0:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/791ce28d9d71780b16dbae1335a42d660889e5bd/ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift#L135)):*

nit
let's rename this `result`
</blockquote></details>

Done.
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 105 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh7OYM5CoEuUaCZK33_:-Nh7mLlQ2Sg_Gaqk8ZJR:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/791ce28d9d71780b16dbae1335a42d660889e5bd/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L105)):*

As discussed with @rablador, we established that we should probably respect the previous behaviour where we were also returning normally in case we were already disconnected here.
</blockquote></details>

Done.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 3 unresolved discussions (waiting on @rablador)

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 59 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh7PETd3QUqwMONk68i:-NhB8qic1IOQrCrDJvUK:b-9zv1cv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/791ce28d9d71780b16dbae1335a42d660889e5bd/ios/PacketTunnelCore/Actor/ObservedState.swift#L59)):*

This is only used for the simulator, we should probably hide this behind an `#if targetEnvironment(simulator)` statement
</blockquote></details>

Would have liked making a local extension in `SimulatorTunnelProviderHost` with the init, but that's not really doable with a struct. Should I just wrap it in `targetEnv` here and call it a day?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift` line 94 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh7F2R-3UNAq27RaQLg:-NhBjdot3xq7UHZAqY2f:b9vcwxs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/791ce28d9d71780b16dbae1335a42d660889e5bd/ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift#L94)):*

nit
Should we care about reachability when we are disconnecting, since we will soon be disconnected and shut down the VPN ?
Maybe we should just force `tunnelStatus.state = .disconnecting(nothing)` here.
What do you think @pronebird  ?
</blockquote></details>

I think it's fine. Path monitor will take care of providing the next reachability status once packet tunnel is fully disconnected.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift` line 63 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh6ixyBAIkio4gmrooW:-NhBlX4w8A6SbqjkgAjs:bhk30l7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b7d99c6aa6b2b3458c46890d5dad98e384cb0186/ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift#L63)):*

What to return here?
</blockquote></details>

Which states did you omit with `default`?

pronebird (changes requested):
> Reviewed 6 of 14 files at r1, 1 of 2 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 10 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift` line 44 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBmoJt4XLexjOIUlz7:-NhBmoJt4XLexjOIUlz8:b-s3c4am) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift#L44)):*
> ```Swift
>     ) {
>         dispatchQueue.async { [weak self] in
> ```

It would be simpler to use a strong reference here. Functions that accept completion handler and never bother to call it on a whim are typically the source of sleepless nights.
___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift` line 76 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBm5C57_z-3cLJbFjl:-NhBm5C57_z-3cLJbFjm:b-p6agx0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift#L76)):*
> ```Swift
> 
>     override func stopTunnel(with reason: NEProviderStopReason, completionHandler: @escaping () -> Void) {
>         dispatchQueue.async { [weak self] in
> ```

It would be simpler to use strong reference here.
___
*[`ios/MullvadVPN/TunnelManager/TunnelState.swift` line 24 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBu79_ACxR9UTqCqhN:-NhBu79_ACxR9UTqCqhO:b-wn5xhr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/MullvadVPN/TunnelManager/TunnelState.swift#L24)):*
> ```Swift
>         var s = "\(state), network "
> 
>         if observedState.connectionState?.isNetworkReachable == true {
> ```

This will print that network is unreachable in `.disconnected` state which would be a lie. So either fix this message to reflect the actual status or remove it altogether.
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 103 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBtlOfDOptWF8QXvmL:-NhBtlOfDOptWF8QXvmM:bh7kzs6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8fc622a104481731cf56aa6a18666eea51da636d/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L103)):*
> ```Swift
>         actor.start(options: startOptions)
> 
>         await actor.waitUntilConnected()
> ```

Nit: If we don't use `waitUntilConnected()` anywhere, then feel free to remove the function altogether.

pronebird (commented):
> Reviewed 5 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 10 unresolved discussions (waiting on @buggmagnet and @rablador)

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 8 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift` line 44 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBmoJt4XLexjOIUlz7:-NhRYD4p4fYBU8zkZWt0:b9faque) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift#L44)):*

It would be simpler to use a strong reference here. Functions that accept completion handler and never bother to call it on a whim are typically the source of sleepless nights.
</blockquote></details>

It's probably fine with strong ref here (especially in simulator), but I still think we should avoid it unless necessary. Added completion handler to early return.
___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift` line 76 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBm5C57_z-3cLJbFjl:-NhRYlc0BeA2D1EnIQ0U:b-q64pkt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift#L76)):*

It would be simpler to use strong reference here.
</blockquote></details>

See above.
___
*[`ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift` line 63 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh6ixyBAIkio4gmrooW:-NhRZxMD71LC19YXqYCn:b2av6uk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b7d99c6aa6b2b3458c46890d5dad98e384cb0186/ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift#L63)):*

Which states did you omit with `default`?
</blockquote></details>

Looking at the old code we're supposedly not interested in anything but .connected, .connecting and .error. I guess getting any other state here means we've messed up pretty badly somewhere, so perhaps .error here would be ok too?
___
*[`ios/MullvadVPN/TunnelManager/TunnelState.swift` line 24 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBu79_ACxR9UTqCqhN:-NhRa6WT8lOoqndWXqTa:b-kwoj2c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/MullvadVPN/TunnelManager/TunnelState.swift#L24)):*

This will print that network is unreachable in `.disconnected` state which would be a lie. So either fix this message to reflect the actual status or remove it altogether.
</blockquote></details>

Why is it incorrect? It only looks at whether network is reachable and prints accordingly.
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 103 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBtlOfDOptWF8QXvmL:-NhRaMv_7fB1IQsZo1XY:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8fc622a104481731cf56aa6a18666eea51da636d/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L103)):*

Nit: If we don't use `waitUntilConnected()` anywhere, then feel free to remove the function altogether.
</blockquote></details>

Done.
___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 59 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh7PETd3QUqwMONk68i:-NhRf-hC70rnatJvKLC8:b-tlg7cw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/791ce28d9d71780b16dbae1335a42d660889e5bd/ios/PacketTunnelCore/Actor/ObservedState.swift#L59)):*

It's unusual that the framework has to adapt to the way it's being used in the application target. I'd say we should keep `init()` public and not over-engineer this.
</blockquote></details>

Change reverted.
___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 24 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBgvvW8ziMHaM-uZYC:-NhRepfp5hoM9WhqBepZ:b-fxzzkh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/PacketTunnelCore/Actor/ObservedState.swift#L24)):*

Ideally we want to encourage consumer to do the proper matching. So these helpers that extract the associated values should probably be defined as extensions in the target doing this.
</blockquote></details>

Putting them in the existing ObservedState+Extensions file.
___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 37 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBiIVjCbfDnQgfYD35:-NhReu_3CNZXgyOpqtij:b-fxzzkh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/PacketTunnelCore/Actor/ObservedState.swift#L37)):*

Same goes here.
</blockquote></details>

Putting them in the existing ObservedState+Extensions file.
___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 56 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBj8WuC-8tARaEONf8:-NhRfUPpCMeaRg1gBhRQ:b-lp4iuc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/PacketTunnelCore/Actor/ObservedState.swift#L56)):*

The whole point of introducing `NetworkReachability` was to avoid making assumption that the undetermined state means we have internet.  I understand that we may have to bridge some gaps before we start using it as it was intended, but them maybe the framework itself does not have to serve these shortcuts for the main target and instead we could define extension where it's used this way.
</blockquote></details>

Would it be bad to switch it around and rely on .reachable instead? If .undetermined is sketchy, then perhaps we should only claim to have network when we're certain.

pronebird (changes requested):
> Reviewed 4 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift` line 63 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh6ixyBAIkio4gmrooW:-NhRzvTy3p9Wg-IGATSu:b-r8yxkz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b7d99c6aa6b2b3458c46890d5dad98e384cb0186/ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift#L63)):*

Looking at the old code we're supposedly not interested in anything but .connected, .connecting and .error. I guess getting any other state here means we've messed up pretty badly somewhere, so perhaps .error here would be ok too?
</blockquote></details>

You  should handle `ObservedState.reconnecting` phase which we still have because we raise the `reasserting = true` flag inside of packet tunnel provider (this is what makes `NEVPNStatus` switch into `reasserting` state; see:`startObservingActorState()`). 

Return `.none` for `disconnecting` and `disconnected` states as those will be mapped separately.

Remember:

1. In the initial phase before the tunnel is connected in the first time, we remain in the `Actor.State.connecting` phase. Upon success we move to the `Actor.State.connected` phase.

    ```
    connecting -> connected
    ```

2. The tunnel once connected can only enter `Actor.State.reconnecting` phase on reconnect and never returns to `Actor.State.connecting` phase as it's only used during the initial connection sequence.

    ```
    connected -> reconnecting -> connected
    ```

3. `Actor.State.error` state knows whether the tunnel had previously connected and would toggle between `error -> connecting -> (connected | error)` or `error -> reconnecting -> (connected | error)` states based on that knowledge.

I suggest you to expand the `default` case and generally avoid it because if we add a new state tomorrow, the `default` case will gladly swallow it, while the compiler would warn you if each state is handled explicitly.
___
*[`ios/MullvadVPN/TunnelManager/TunnelState.swift` line 24 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBu79_ACxR9UTqCqhN:-NhS3KcnCmgpp-Q-2Vuu:bslwphk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/MullvadVPN/TunnelManager/TunnelState.swift#L24)):*

Why is it incorrect? It only looks at whether network is reachable and prints accordingly.
</blockquote></details>

Because it would print `unreachable` in the `.disconnected` state.
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 104 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhS3kmDEHqxxWYIi8KT:-NhS3kmDEHqxxWYIi8KU:bb7felk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ae8cee3398bc0f4d8a0c720d7d457dc80d5ce9ed/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L104)):*
> ```Swift
> 
>         for await state in await actor.observedStates {
>             switch state {
> ```

We probably need to handle `error` state in here too, and return immediately.
___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 24 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBgvvW8ziMHaM-uZYC:-NhRz8CT8GoPxDNPrjpI:b-e9ld4z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/PacketTunnelCore/Actor/ObservedState.swift#L24)):*

Putting them in the existing ObservedState+Extensions file.
</blockquote></details>

Alright keep them in the framework.
___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 56 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBj8WuC-8tARaEONf8:-NhRxjoc6n2-YDSChL_W:b-sabgfm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/PacketTunnelCore/Actor/ObservedState.swift#L56)):*

Would it be bad to switch it around and rely on .reachable instead? If .undetermined is sketchy, then perhaps we should only claim to have network when we're certain.
</blockquote></details>

I think the check you have here should maintain the existing behavior, so don't invert it. I'd much rather iterate over the codebase, than spend too much time in this PR trying to handle all of the rough edges, so I am unblocking this issue. 

However note that `.undetermined` means that we don't know the reachability status yet. When we don't know, we cannot really make assumption whether we have network or not, it would be biased.

We used to make that assumption but it's not pretty. What should we do when the network status is unknown is a good question but perhaps not claim that we know what's going on, or maybe indicate that the network status is being determined. This seems like something for a broader discussion in the team.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift` line 63 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-Nh6ixyBAIkio4gmrooW:-NhW0aOW9T5souJ_Zydv:b-emhpkz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b7d99c6aa6b2b3458c46890d5dad98e384cb0186/ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift#L63)):*

You  should handle `ObservedState.reconnecting` phase which we still have because we raise the `reasserting = true` flag inside of packet tunnel provider (this is what makes `NEVPNStatus` switch into `reasserting` state; see:`startObservingActorState()`). 

Return `.none` for `disconnecting` and `disconnected` states as those will be mapped separately.

Remember:

1. In the initial phase before the tunnel is connected in the first time, we remain in the `Actor.State.connecting` phase. Upon success we move to the `Actor.State.connected` phase.

    ```
    connecting -> connected
    ```

2. The tunnel once connected can only enter `Actor.State.reconnecting` phase on reconnect and never returns to `Actor.State.connecting` phase as it's only used during the initial connection sequence.

    ```
    connected -> reconnecting -> connected
    ```

3. `Actor.State.error` state knows whether the tunnel had previously connected and would toggle between `error -> connecting -> (connected | error)` or `error -> reconnecting -> (connected | error)` states based on that knowledge.

I suggest you to expand the `default` case and generally avoid it because if we add a new state tomorrow, the `default` case will gladly swallow it, while the compiler would warn you if each state is handled explicitly.
</blockquote></details>

Right, the .reconnecting state as well. I mapped it like the others, while mapping .initial to .none with the other two remaining states.
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 104 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhS3kmDEHqxxWYIi8KT:-NhW6RPCE0AxThPrSS3T:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ae8cee3398bc0f4d8a0c720d7d457dc80d5ce9ed/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L104)):*

We probably need to handle `error` state in here too, and return immediately.
</blockquote></details>

Done.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: 14 of 16 files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadVPN/TunnelManager/TunnelState.swift` line 24 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhBu79_ACxR9UTqCqhN:-NhW7zLVBW3e-jVlpU1L:b-9uou8e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1ccf5474e31b0b2994d70fa08090efc23f428bc8/ios/MullvadVPN/TunnelManager/TunnelState.swift#L24)):*

Because it would print `unreachable` in the `.disconnected` state.
</blockquote></details>

Ok, got your meaning by digging a little. Updated the description.

pronebird (changes requested):
> Reviewed 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift` line 66 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhaLfue-0NsHxc-j8kA:-NhaLfuf5D_CPqtGX7nv:bqumip1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/92f64a65693e3710c2ef60f63568bbd9ef4bd8a7/ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift#L66)):*
> ```Swift
>                 }
>             }
>             return
> ```

Removing that `return` statement would mark operation as finished right away, instead of after fetching the tunnel status! See the last line in this function.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift` line 66 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-NhaLfue-0NsHxc-j8kA:-Nhaa65N31DcmEdnxgMX:b9brp5h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/92f64a65693e3710c2ef60f63568bbd9ef4bd8a7/ios/MullvadVPN/TunnelManager/MapConnectionStatusOperation.swift#L66)):*

Removing that `return` statement would mark operation as finished right away, instead of after fetching the tunnel status! See the last line in this function.
</blockquote></details>

Aha, reverted!

pronebird (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329#-:-NhaajvQ2eUFJoMC-eO2:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

buggmagnet (approved):
> Reviewed 3 of 4 files at r5, 3 of 4 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5329)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5340, opened 2023-10-21
Refactor mullvad-api

dlon:
This PR gets rid of a lot of repetition surrounding authentication, error handling, deserialization, and other things, in `mullvad-api`.

Fix DES-416.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5340)

### DIFF
--- a/mullvad-api/src/access.rs
+++ b/mullvad-api/src/access.rs
@@ -179,6 +179,5 @@ async fn fetch_access_token(
     let rest_request = factory
         .post_json(&format!("{AUTH_URL_PREFIX}/token"), &request)?
         .expected_status(&[StatusCode::OK]);
-    let response = service.request(rest_request).await?;
-    rest::deserialize_body(response).await
+    service.request(rest_request).await?.deserialize().await
 }
--- a/mullvad-api/src/access.rs
+++ b/mullvad-api/src/access.rs
@@ -38,7 +38,7 @@ struct AccountState {
 
 impl AccessTokenStore {
     pub(crate) fn new(service: RequestServiceHandle) -> Self {
-        let factory = rest::RequestFactory::new(API.host.clone(), None);
+        let factory = rest::RequestFactory::new(&API.host, None);
         let (tx, rx) = mpsc::unbounded();
         tokio::spawn(Self::service_requests(rx, service, factory));
         Self { tx }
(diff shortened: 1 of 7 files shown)


MarkusPettersson98 (changes requested):
> Reviewed 6 of 7 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5340)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-api/src/rest.rs` line 332 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5340#-NhVdN2783ihtKBIbupY:-NhVdN2783ihtKBIbupZ:b-3utuff) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2985492ebe6924fbba2feb3c20e8e65987f01536/mullvad-api/src/rest.rs#L332)):*
> ```XML
>     /// Set the account token to obtain authentication for.
>     /// This fails if no store is set.
>     pub fn account(mut self, auth: Option<AccountToken>) -> Result<Self> {
> ```

Is it necessary for the `auth` parameter to be optional? Could this function signature be rewritten as:
```rust
    pub fn account(mut self, auth: AccountToken) -> Result<Self>
```
I didn't find any call site where `fn account` was called with `auth: None`, and if this is needed I think having a separate, explicit function for setting the account token to `None` is preferable



_Code quote:_
```XML
pub fn account(mut self, auth: Option<AccountToken>) -> Result<Self> {
```

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5340)* status: 3 of 7 files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`mullvad-api/src/rest.rs` line 332 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5340#-NhVdN2783ihtKBIbupY:-NhYC6NnEIg2nUUVK0-T:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2985492ebe6924fbba2feb3c20e8e65987f01536/mullvad-api/src/rest.rs#L332)):*

Is it necessary for the `auth` parameter to be optional? Could this function signature be rewritten as:
```rust
    pub fn account(mut self, auth: AccountToken) -> Result<Self>
```
I didn't find any call site where `fn account` was called with `auth: None`, and if this is needed I think having a separate, explicit function for setting the account token to `None` is preferable
</blockquote></details>

Done.

PR #5348, opened 2023-10-23
Add socks5 proxy [POC]

pronebird:
This PR adds a socks5 forwarding proxy and socks5 transport to communicate with our API. Similar to shadowsocks implementation, the socks5 forwarding proxy runs on a local TCP port and transparently handles all traffic, making it possible to direct URLSession at it in order to communicate with the API over the remote socks proxy.

[SOCKS5](https://datatracker.ietf.org/doc/html/rfc1928) is a simple protocol and in order to establish the connection over socks proxy, the client has to perform the following steps: 

#### 1. Initial handshake 

During that phase the client sends the following information over the wire:

---------------------------------
|VER | NMETHODS | METHODS  |
| ----|--------------|------------|
|   1   |    1                  |    1 to 255   |
---------------------------------

- `VER` is the socks version, always `0x05`. 1 byte.
- `NMETHODS` - number of authentication methods supported by the client. 1 byte. 
- `METHODS` - variable length byte stream, depending on `NMETHODS`. Each byte corresponds to `UInt8` constant that correlates with an authentication method defined in the protocol. See `Socks5AuthenticationMethod`.

The handshake response contains two bytes:

------------------
|VER | METHOD  |
| ----|-----------|
|   1   |    1             |
-----------------

- `VER` is the socks version, always `0x05`. 1 byte.
- `METHOD` is the authentication method accepted by the server. When this value is set to `0xFF`, it means that the client doesn't support any of the authentication methods accepted by the server. The client should terminate connection in such case.

#### 2. Authentication

Current implementation only supports anonymous access to socks server and therefore when authentication is not required the client can move on to the next phase.

#### 3. Connect command

The client requests the socks proxy to establish a connection to the remote server, to which the client wants to talk to, over the proxy.

A `CONNECT` command should be sent over the wire to initiate the connection and the layout looks as such:

---------------------------------------------------
|VER | CMD |  RSV  | ATYP | DST.ADDR | DST.PORT |
|----|-------|------|-------|------------|-----------|
| 1    |        1  | X'00' |  1        | Variable     |    2             |
---------------------------------------------------

- `VER` - socks version. 1 byte, always `0x05`.
- `CMD` - CONNECT command, which is represented by a byte code: `0x01`. 1 byte.
- `RSV` - reserved field. Always zero. 1 byte.
- `ATYP` - address type. Socks support IPv4, IPv6 and DNS domain name (See `Socks5AddressType`). 1 byte.
- `DST.ADDR` - the address of the remote server to which the clients wants to connect.
    - IPv4: 4 octets
    - IPv6: 16 octets
    - Domain name: the first byte represents the length of domain name, the reminder of it is the domain name itself. Note that since only one byte is used to represent the length of domain name, it means that the domain name cannot be longer than 255 bytes (`UInt8.max`)
    - `DST.PORT` - the port of the remote server to which the client wants to connect. Typically our API runs on port `443` (SSL). This field is `UInt16` in _network byte order_, 2 bytes long.

The response to this request is nearly identical:

---------------------------------------------------
| VER | REP |  RSV  | ATYP | BND.ADDR | BND.PORT |
|-----|-----|-------|------|-------------|------------|
| 1      |  1     | X'00'  |  1      | Variable       |    2               |
---------------------------------------------------

- `VER` - socks version. 1 byte, always `0x05`.
- `REP` - response code. 1 byte. See `Socks5StatusCode`. Zero byte indicates success.
- `RSV` - reserved byte, always zero. 1 byte.
- `ATYP` - address type. Socks support IPv4, IPv6 and DNS domain name (See `Socks5AddressType`). 1 byte.
- `BND.ADDR` and `BND.PORT` - typically seem to coincide with the socks proxy address and port.

if `REP` contains zero, the client can begin the data exchange with the remote server (our API).

#### 4. Begin data exchange with the remote server

At this point the socks5 proxy turns into the tunnel to the remote server, so the client can begin sending and receiving HTTP/S or any other traffic over the wire. This is the simplest part where the client passes bytes from local to remote connection and back.

### Implementation structure

- `Socks5Connection` - type implementing a bidirectional data connection between a local and remote endpoint over the socks proxy. it's responsible for handling the socks5 flow, local and remote connections. It delegates each individual phase to the following objects:
    - `Socks5HandshakeNegotiation` - a type responsible for performing a handshake negotiation.
    - `Socks5ConnectNegotiation` - a type responsible for performing a CONNECT command negotiation.
    - `Socks5DataStreamHandler` - a type responsible for bidirectional data exchange.
- `Socks5ForwardingProxy` - the forwarding proxy that opens a local TCP port and starts listening on it. It creates a new instance of `Socks5Connection` for each new connection, then it initiates the socks5 exchange by calling `Socks5Connection.start()`.

### Threading considerations

In network framework, the objects implementing networking are initialized with a dispatch queue which is used for dispatching connection-related events. 

Since the socks5 flow is driven by connection related events, the execution context is guaranteed to remain the same and therefore the socks5 related types do not do any synchronization with exception of public methods, which are typically wrapped to run on the same dispatch queue.

### Testing

Still looking into unit tests. 

In order to active the socks5 transport you need to:

- Return `.useSocks5` from `TransportStrategy.connectionTransport()`. 
- Tweak the socks5 proxy address in `TransportProvider.socks5()`. When testing on device, replace `.loopback` with the IP of your machine in the local network or wherever you run the socks5 proxy, `.loopback` should work on simulator if you run socks5 proxy locally.

If you need socks5 server software, I recommend using _Charles - web debugging proxy_. Enable SOCKS proxy under Proxy > Proxy settings > Enable SOCKS then run the app. You don't need to configure SSL certificates or anything. You can also try one of public & free SOCKS proxies available online.


rablador (changes requested):
> Reviewed 24 of 24 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 4 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 128 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NhaxrNC5AzwtRTodWTs:-NhaxrNC5AzwtRTodWTt:bg8yw98) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5Connection.swift#L128)):*
> ```Swift
>     }
> 
>     private func onLocalConnectiondState(_ connectionState: NWConnection.State) {
> ```

Nit: ConnectionD
___
*[`ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Nhb3Vw5AlcVscqf1DQL:-Nhb3Vw5AlcVscqf1DQM:b-dadeoj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift#L78)):*
> ```Swift
>             addressType: addressType,
>             onComplete: { [self] endpoint in
>                 let reply = Socks5ConnectReply(status: statusCode, serverBoundEndpoint: endpoint)
> ```

Since we guard for a specific code above (.succeeded) and then pass the var here anyway, if the guard is ever removed we might end up with unexpected behavior. If we intead pass .succeeded we will never end up with anything but that no matter what.
___
*[`ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift` line 130 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NhbF6-v7hlFO_rU5mRM:-NhbF6-v7hlFO_rU5mRN:b-85aypi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift#L130)):*
> ```Swift
> 
>         case let .starting(listener, previousCompletion):
>             // Accumulate completion handlers when requested to start multiple times in a row.
> ```

What's the reason to accumulate completions? Can't we just drop multiple calls here since we're waiting to start anyway?
___
*[`ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift` line 257 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Nhfhgo0Cfx8wmkz_7IE:-Nhfhgo0Cfx8wmkz_7IF:b-isvkzn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift#L257)):*
> ```Swift
>             )
>             socks5Connection.setStateHandler { [weak self] socks5Connection, state in
>                 if case let .stopped(error) = state {
> ```

Only handle one state here?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 4 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 128 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NhaxrNC5AzwtRTodWTs:-Nhl6Elm3Oqsx0RXN0Pk:becat05) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5Connection.swift#L128)):*

Nit: ConnectionD
</blockquote></details>

Really need to fix my keyboard. Done.
___
*[`ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Nhb3Vw5AlcVscqf1DQL:-Nhl6fL08r59sebjUr2l:b-7evwaj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift#L78)):*

Since we guard for a specific code above (.succeeded) and then pass the var here anyway, if the guard is ever removed we might end up with unexpected behavior. If we intead pass .succeeded we will never end up with anything but that no matter what.
</blockquote></details>

What "var" and why would you remove a `statusCode` check above?
___
*[`ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift` line 130 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NhbF6-v7hlFO_rU5mRM:-Nhl7GFe1UkKTjr_VyhC:b-gg0jd2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift#L130)):*

What's the reason to accumulate completions? Can't we just drop multiple calls here since we're waiting to start anyway?
</blockquote></details>

To keep this call idempotent.
___
*[`ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift` line 257 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Nhfhgo0Cfx8wmkz_7IE:-Nhl7ftdBvu7Of34TRBU:b-ga6px6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift#L257)):*

Only handle one state here?
</blockquote></details>

Yeah? As far as proxy concerned, it's only interested to remove the connection from the list of connections when it stopped.

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Nhb3Vw5AlcVscqf1DQL:-Ni-s9_B0QYuqnu3Q0rs:b-imvyri) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift#L78)):*

What "var" and why would you remove a `statusCode` check above?
</blockquote></details>

Sorry, I meant the `statusCode` variable.
No, the guard should stay, but I think `.succeeded` could be passed instead of `statusCode`. We know that - in the current impl. - `statusCode` will always be `.succeeded` if we get to this point in the code, which implies that we should always pass `.succeeded` to the `Socks5ConnectReply` init. But what if the code above (incl where the guard is) changes? Then we might pass a status code that's not successful and thus get possibly incorrect behavior.
___
*[`ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift` line 130 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NhbF6-v7hlFO_rU5mRM:-Ni-uiAc39wQmrIBxC0G:bzd6jas) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift#L130)):*

To keep this call idempotent.
</blockquote></details>

Alright, to keep track of potential errors?

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 2 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Nhb3Vw5AlcVscqf1DQL:-Ni3vsp17YcCxq4LpX63:bm9v2to) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift#L78)):*

Sorry, I meant the `statusCode` variable.
No, the guard should stay, but I think `.succeeded` could be passed instead of `statusCode`. We know that - in the current impl. - `statusCode` will always be `.succeeded` if we get to this point in the code, which implies that we should always pass `.succeeded` to the `Socks5ConnectReply` init. But what if the code above (incl where the guard is) changes? Then we might pass a status code that's not successful and thus get possibly incorrect behavior.
</blockquote></details>

I don't understand the premise of your suggestion. 

So you're saying that we can infer that the `statusCode == .succeeded` after the guard check and that we should hardcode the status code instead.

Then you say that someone could remove that guard by mistake, which could lead to logical error and that it's ok if we keep reporting that everything is hunky-dory to the call site instead of reporting the actual status that we received from server?
___
*[`ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift` line 130 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NhbF6-v7hlFO_rU5mRM:-Ni41wrD-zgqs8lPdADD:b-3s8zl3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift#L130)):*
Not exactly, but to simplify the code. I think it's worth to remind ourselves what idempotence means:

> Idempotence, in programming and mathematics, is **a property of some operations such that no matter how many times you execute them, you achieve the same result**.

For example, when you say that we should "drop" multiple calls, what do you really mean by that? Correct me if I am wrong, but to me it sounds like you suggest to simply return and never call the completion handler given by the caller when the object is already in the `starting` state. 

Apologize for the use of that term, but "ghosting" the caller typically leads to hard to debug problems, outside flows hit the road block and never continue, swift continuations tend to break, . Workarounds include the usage of timers, we had exactly that problem in the packet tunnel when using network extension APIs. 

Sometimes it's ok this way, but not always. When designing independent components it's better to keep their behavior as straightforward as possible, to lessen the impact on the caller.

Yes we could return (aka drop?), we could even call the completion handler right away and give it the error indicating that the object is already starting but why complicate and introduce additional code paths when we can simply accumulate the completion handler and carry on?

In this particular scenario, isn't it easier when the object behaves identical and accumulate the blocks? For instance:

```swift
proxy.start { error in
  // called first
}
proxy.start { error in
  // called second
}
```

and not like that:

```swift
proxy.start { error in
  // called first
}
proxy.start { error in
  // never called??
}
```

nor

```swift
proxy.start { error in
  // called first
}
proxy.start { error in
  // called with error telling the object is already starting
}
```

Consumer code is a "storm", it's often complex and perhaps a mess and small components like this can be an "island" of sanity, that can keep the "storm" from the outside from ripping the "island" and everything outside apart. 

I can imagine that the consumer code be sharing the same proxy, and keeping the call to `start()` idempotent like this means that the behavior of object is consistent regardless if there are two or more external components that call `proxy.start()` at the same time. 

It's just practical in that case and it avoids handling additional edge cases. You can argue we know exactly how we gonna use it but things change and when someone decides to use in the other way, we have edge cases, this approach eliminates all that.

buggmagnet (changes requested):
> Reviewed 12 of 24 files at r1, 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 12 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/MullvadTransport/AnyIPEndpoint+Socks5.swift` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni-v2U-76Vqx2O2T1bo:-Ni-v2U-76Vqx2O2T1bp:bw0tvl4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/AnyIPEndpoint%2BSocks5.swift#L13)):*
> ```Swift
> import Network
> 
> extension AnyIPEndpoint {
> ```

What's the difference between `AnyIPEndpoint` and `AnyIPAddress` ? 
And by the same logic, the question extends to the types `IPv[4|6]Address` and `IPv[4|6]Endpoint`
___
*[`ios/MullvadTransport/DeferredCancellable.swift` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni-w1Sy1Wa4_MCuZ-ke:-Ni-w1Sy1Wa4_MCuZ-kf:bzgkbue) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/DeferredCancellable.swift#L13)):*
> ```Swift
> 
> /// Cancellable object that defers cancellation until the other token is connected to it.
> final class DeferredCancellable: Cancellable {
> ```

This doesn't do any deferred actions. 
I would suggest to rename it to `TokenChain` or something similar since it links the lifetime of several `Cancellable` tokens.
___
*[`ios/MullvadTransport/DeferredCancellable.swift` line 23 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni-wMme0dMHBz3LNe3V:-Ni-wMme0dMHBz3LNe3W:b43ezkk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/DeferredCancellable.swift#L23)):*
> ```Swift
>     ///
>     /// The token is cancelled immediately, if the deferred object is already cancelled.
>     func connect(_ token: Cancellable) {
> ```

I see why you named it `connect` but I think it's a misleading name. Considering we do a lot of network related calls, we usually think of `connect` as "connect to a webserver".
Here rather, we `link` tokens together. 
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 11 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4b7j73c3STrMuTHiR:-Ni4b7j73c3STrMuTHiS:bbvml6h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L11)):*
> ```Swift
> import Network
> 
> /// A bidirection data connection between a local endpoint and remote endpoint over socks proxy.
> ```

typo
`bidirection` -> `bidirectional` 
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 17 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4bM38-4Zq369A8TND:-Ni4bM398U7zgbXGWPOI:bc5ltin) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L17)):*
> ```Swift
> 
>     /**
>      Initilizes a new connection passing data between local and remote TCP connection over the socks proxy.
> ```

typo
`Initilizes`
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 72 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4avX6EvBAEtmQjPkt:-Ni4avX6EvBAEtmQjPku:bu4ycu7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L72)):*
> ```Swift
>      Set a handler that receives connection state events.
> 
>      It's advised to set the state handler before starting the conncetion to avoid missing updates to the connection state.
> ```

typo
`conncetion`
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 128 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4aUXfCXlm_BFWlsxM:-Ni4aUXfCXlm_BFWlsxN:bcy9x6f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L128)):*
> ```Swift
>     }
> 
>     private func onLocalConnectionState(_ connectionState: NWConnection.State) {
> ```

nit
What do you think of renaming this`onConnectionStateChanged` ?
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4ELG-0NM0TMfZVFcE:-Ni4ELG-0NM0TMfZVFcF:bo9iz52) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L136)):*
> ```Swift
>             initiateConnection()
> 
>         case let .waiting(error), let .failed(error):
> ```

The `.waiting` state is not fatal, we shouldn't error out until we reach the `.failed` state

The same comment applies for `onRemoteConnectionState` 
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 144 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4_6Ga8Z4skhjKd16Q:-Ni4_6Gb9Y9J2BYZK90d:b1e8gse) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L144)):*
> ```Swift
>     }
> 
>     private func onRemoteConnectionState(_ connectionState: NWConnection.State) {
> ```

Right now, both `onLocalConnectionState` and `onRemoteConnectionState` have the same implementation.
We can use a single function to do both if we don't intend to separate the behavior based on whether the connection is remote or local
___
*[`ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift` line 98 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4dpxS-yKvvQSszYb1:-Ni4dpxT1cKBL-HVpneD:b-er6r6g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift#L98)):*
> ```Swift
>         // Read status code, reserved field and address type from reply.
>         guard let rawStatusCode = iterator.next(),
>               let _ = iterator.next(), // skip reserved field
> ```

:warning: :police_officer: 
`Unused Optional Binding Violation: Prefer `!= nil` over `let _ =` (unused_optional_binding)`

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 9 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadTransport/AnyIPEndpoint+Socks5.swift` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni-v2U-76Vqx2O2T1bo:-Ni9ljMJDCob7Y9fOSW0:b-uymzcn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/AnyIPEndpoint%2BSocks5.swift#L13)):*

What's the difference between `AnyIPEndpoint` and `AnyIPAddress` ? 
And by the same logic, the question extends to the types `IPv[4|6]Address` and `IPv[4|6]Endpoint`
</blockquote></details>

The endpoint is typically a pair of address and port or hostname and port. An IP address is just that - an address without the port component. I hope that answers your question. 

Note that `AnyIPAddress` hasn't been used anywhere but in `PreferencesViewModel`.
___
*[`ios/MullvadTransport/DeferredCancellable.swift` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni-w1Sy1Wa4_MCuZ-ke:-Ni9g4GPAu7cgSrh4Aw2:b9aedkv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/DeferredCancellable.swift#L13)):*
The deferred cancellable _postpones_, or in other words _defers_ the cancellation in the event when cancellation request arrives prior to the actual cancellable being connected to the deferred one. 

This is the dictionary definition of the word _defer_ which precisely describes what's happening here:

> Defer
> put off (an action or event) to a later time; postpone. 
> "they deferred the decision until February"
___
*[`ios/MullvadTransport/DeferredCancellable.swift` line 23 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni-wMme0dMHBz3LNe3V:-Ni9gVuQ7ogbN5nAPlvS:b-g6ytsf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/DeferredCancellable.swift#L23)):*

I see why you named it `connect` but I think it's a misleading name. Considering we do a lot of network related calls, we usually think of `connect` as "connect to a webserver".
Here rather, we `link` tokens together.
</blockquote></details>

These are synonyms. In general the word "connect" does not necessarily have anything to do with network.

In fact network framework does not even use _connect_ anywhere and mostly sticks to _start_ or _cancel_ instead of _connect_ or _disconnect_. I don't think it's even the case anywhere with `URLSession` whos data tasks are being started with a call to _activate_ or _resume_.

On the other side "connect" is commonly used in Combine when connecting to the publisher that yields elements. In this particular case the deferred cancellable object is a publisher and we connect the other cancellable token to it.
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 11 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4b7j73c3STrMuTHiR:-Ni9hSyO5hN4W6wHVRXm:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L11)):*

typo
`bidirection` -> `bidirectional`
</blockquote></details>

Done
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 17 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4bM38-4Zq369A8TND:-Ni9hTiD96bFh4A06sE_:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L17)):*

typo
`Initilizes`
</blockquote></details>

Done
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 72 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4avX6EvBAEtmQjPkt:-Ni9hZDU-O3vj_vnUKSR:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L72)):*

typo
`conncetion`
</blockquote></details>

Done.
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 128 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4aUXfCXlm_BFWlsxM:-Ni9jHuj4LXGMW7w17Am:b-s2eguz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L128)):*

nit
What do you think of renaming this`onConnectionStateChanged` ?
</blockquote></details>

Since we have two different connections, having a generic "on_Connection_..." used for state observing function is misleading imo.
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4ELG-0NM0TMfZVFcE:-Ni9iJtED41uc6RAIZpP:b-7kqyd2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L136)):*

The `.waiting` state is not fatal, we shouldn't error out until we reach the `.failed` state

The same comment applies for `onRemoteConnectionState`
</blockquote></details>

It's not fatal if you intend to wait for connectivity. The idea was to error right away and let the REST framework handle the retry instead. This aligns with `URLSession` configuration we use, where we do not wait for network connectivity (`waitsForConnectivity` in `URLSessionConfiguration`).
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 144 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4_6Ga8Z4skhjKd16Q:-Ni9hf5lBMcMY-ZccvjT:bout519) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L144)):*

Right now, both `onLocalConnectionState` and `onRemoteConnectionState` have the same implementation.
We can use a single function to do both if we don't intend to separate the behavior based on whether the connection is remote or local
</blockquote></details>

The implementation is similar but not entirely the same. Different error is passed to `handleError()`. I thought about adding a factory to produce a closure mapping to the right error, but then opted-in for simplicity as it's easier to debug it this way since this way we know which connection state is being received.
___
*[`ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift` line 98 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4dpxS-yKvvQSszYb1:-Ni9kXM9Ard1sN-1EWud:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift#L98)):*

:warning: :police_officer: 
`Unused Optional Binding Violation: Prefer `!= nil` over `let _ =` (unused_optional_binding)`
</blockquote></details>

Done.

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: 21 of 24 files reviewed, 9 unresolved discussions (waiting on @buggmagnet and @pronebird)

___
*[`ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift` line 130 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NhbF6-v7hlFO_rU5mRM:-Ni9vt9U4JZXetd5ZgHq:b2loodt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift#L130)):*

Not exactly, but to simplify the code. I think it's worth to remind ourselves what idempotence means:

> Idempotence, in programming and mathematics, is **a property of some operations such that no matter how many times you execute them, you achieve the same result**.

For example, when you say that we should "drop" multiple calls, what do you really mean by that? Correct me if I am wrong, but to me it sounds like you suggest to simply return and never call the completion handler given by the caller when the object is already in the `starting` state. 

Apologize for the use of that term, but "ghosting" the caller typically leads to hard to debug problems, outside flows hit the road block and never continue, swift continuations tend to break, . Workarounds include the usage of timers, we had exactly that problem in the packet tunnel when using network extension APIs. 

Sometimes it's ok this way, but not always. When designing independent components it's better to keep their behavior as straightforward as possible, to lessen the impact on the caller.

Yes we could return (aka drop?), we could even call the completion handler right away and give it the error indicating that the object is already starting but why complicate and introduce additional code paths when we can simply accumulate the completion handler and carry on?

In this particular scenario, isn't it easier when the object behaves identical and accumulate the blocks? For instance:

```swift
proxy.start { error in
  // called first
}
proxy.start { error in
  // called second
}
```

and not like that:

```swift
proxy.start { error in
  // called first
}
proxy.start { error in
  // never called??
}
```

nor

```swift
proxy.start { error in
  // called first
}
proxy.start { error in
  // called with error telling the object is already starting
}
```

Consumer code is a "storm", it's often complex and perhaps a mess and small components like this can be an "island" of sanity, that can keep the "storm" from the outside from ripping the "island" and everything outside apart. 

I can imagine that the consumer code be sharing the same proxy, and keeping the call to `start()` idempotent like this means that the behavior of object is consistent regardless if there are two or more external components that call `proxy.start()` at the same time. 

It's just practical in that case and it avoids handling additional edge cases. You can argue we know exactly how we gonna use it but things change and when someone decides to use in the other way, we have edge cases, this approach eliminates all that.
</blockquote></details>

Yes, by dropping I meant that perhaps we wouldn't need completions for a state that's not changing anyway, but you're right that it wouldn't be idempotent that way. It's not a foreign pattern, eg. "calling twice method multiple times does nothing", but I think I prefer your suggestion anyway. Let's keep it as it is.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 7 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Nhb3Vw5AlcVscqf1DQL:-Ni9xpjn7RjET3Mt6fAZ:b7smg72) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift#L78)):*

I don't understand the premise of your suggestion. 

So you're saying that we can infer that the `statusCode == .succeeded` after the guard check and that we should hardcode the status code instead.

Then you say that someone could remove that guard by mistake, which could lead to logical error and that it's ok if we keep reporting that everything is hunky-dory to the call site instead of reporting the actual status that we received from server?
</blockquote></details>

Sure, that could happen and wouldn't be good, but the idea behind my claim still holds up I think. The way I read this function is that basically the entirety of it should be run only if status is `.succeeded` (the comment even says "Parse server bound endpoint when partial reply indicates success."), so in that sense the guard might as well be moved to the call site and the function be renamed "handleSuccessfulReply" or something. That way no errors would be made. It IS nice to handle all this in the same function though, and I'm not particularly adamant about it. Just pointing it out.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: 19 of 24 files reviewed, 7 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift` line 78 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Nhb3Vw5AlcVscqf1DQL:-NiFJ6-YAYYU3ngB3KR6:b-vupjlm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72acf38846489f1edd42bbdef4ff64a7598e8a05/ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift#L78)):*

Sure, that could happen and wouldn't be good, but the idea behind my claim still holds up I think. The way I read this function is that basically the entirety of it should be run only if status is `.succeeded` (the comment even says "Parse server bound endpoint when partial reply indicates success."), so in that sense the guard might as well be moved to the call site and the function be renamed "handleSuccessfulReply" or something. That way no errors would be made. It IS nice to handle all this in the same function though, and I'm not particularly adamant about it. Just pointing it out.
</blockquote></details>

Moved reply handling to the call site.

rablador (approved):
> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 7 unresolved discussions (waiting on @buggmagnet)

buggmagnet (changes requested):
> Reviewed 5 of 24 files at r1, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 8 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadTransport/AnyIPEndpoint+Socks5.swift` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni-v2U-76Vqx2O2T1bo:-NiEMP7d9YZW9ayuQ4Ci:bcpg9hd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/AnyIPEndpoint%2BSocks5.swift#L13)):*

The endpoint is typically a pair of address and port or hostname and port. An IP address is just that - an address without the port component. I hope that answers your question. 

Note that `AnyIPAddress` hasn't been used anywhere but in `PreferencesViewModel`.
</blockquote></details>

Thanks for the clarification. I'm wondering then if we shouldn't deprecate `AnyIPAddress` and replace them with an `AnyIPEndpoint` with port 0 instead.
___
*[`ios/MullvadTransport/DeferredCancellable.swift` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni-w1Sy1Wa4_MCuZ-ke:-NiEHfyj33N6Akzz7R7_:byunfwg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/DeferredCancellable.swift#L13)):*

The deferred cancellable _postpones_, or in other words _defers_ the cancellation in the event when cancellation request arrives prior to the actual cancellable being connected to the deferred one. 

This is the dictionary definition of the word _defer_ which precisely describes what's happening here:

> Defer
> put off (an action or event) to a later time; postpone. 
> "they deferred the decision until February"
</blockquote></details>

Right, it still doesn't do any deferred action. It doesn't postpone cancellation. It will cancel _right away_ all the tokens if it is cancelled. That is not a deferred mechanism.

If we look at the doc for `Task`, the term mentioned is "cooperative".
So this is really a `CooperativeCancellable`.

If you cancel one token, it cancels all tokens linked to it.
___
*[`ios/MullvadTransport/DeferredCancellable.swift` line 23 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni-wMme0dMHBz3LNe3V:-NiEJG9oC5YL54qyqrHL:bsp9dzu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/DeferredCancellable.swift#L23)):*
> In general the word "connect" does not necessarily have anything to do with network.
Yes, but we're writing a VPN application. So there's a really strong contextual connotation with the word "connect".

> On the other side "connect" is commonly used in Combine when connecting to the publisher that yields elements.
> the deferred cancellable object is a publisher

It doesn't have any subscribers, and doesn't implement the `Publisher` protocol. 
I don't understand why you're trying to tie `Combine` related terms or concept for something that has nothing to do with it.
___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4ELG-0NM0TMfZVFcE:-NiYp_5F5EkctfhKKJFf:bfjc2ap) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L136)):*
> This aligns with `URLSession` configuration we use, where we do not wait for network connectivity.

We should revisit this then, right now we have an inconsistent behaviour. Every operation that uses `BackgroundObserver` will always wait for connectivity, even if the flag is set to ignore it.

From the docs of `waitsForConnectivity`
> This property is ignored by background sessions, which always wait for connectivity.

That aside, I still think we shouldn't error in the waiting state.
> error right away and let the REST framework handle the retry instead

Right now, in case of error, we just call the completion handler with `URLError(.cancelled)` 
which means a device with a bad connectivity will likely consume all the retry attempts before it can get good connectivity.

I don't see any reason to not wait it out first, and let the system decide whether the call is failed or not.
We lose nothing by waiting a bit instead of erroring out immediately.
___
*[`ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift` line 53 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NiZKbp5ER4Kf8DukPQh:-NiZKbp5ER4Kf8DukPQi:br5rng5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e689f81b2faa2128d95e469120d01d3a69ca5cd8/ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift#L53)):*
> ```Swift
>                 }
>             } else {
>                 onFailure(Socks5Error.unexpectedEndOfStream)
> ```

nit
Technically this would be only true if we checked the `contentContext` `isFinal` message and it were true.
I would suggest either checking that, and logging an error if there is neither error, nor data to handle.

IIRC, TCP messages cannot be sent with 0 data inside, and the `NWConnection` API has tons of edge cases since the same API is used regardless of the protocol.
I would suggest erring on the side of caution for now.
___
*[`ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift` line 187 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NiZmi_M4Ta6ZF47Xtwk:-NiZmi_M4Ta6ZF47Xtwl:b84pne9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e689f81b2faa2128d95e469120d01d3a69ca5cd8/ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift#L187)):*
> ```Swift
>             state = .stopped
>             cancelListener(listener)
>             openConnections.forEach { $0.cancel() }
> ```

Shouldn't we remove all the connections after cancelling them here ?

in `onEndConnection` we only remove the connection if `self.state` == `.started` because otherwise we cannot access the `openConnections`.

However, when we call `stopInner`, we immediately enter the `.stopped` state, which means the completion handler that calls `onEndConnection` 
when a connection is stopped  won't trigger the path to remove a single connection.
___
*[`ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift` line 275 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NiZkq3LCp0aYSK569RQ:-NiZkq3LCp0aYSK569RR:b-oy59ga) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e689f81b2faa2128d95e469120d01d3a69ca5cd8/ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift#L275)):*
> ```Swift
> 
>         case .started(let listener, var openConnections):
>             guard let index = openConnections.firstIndex(where: { $0 === connection }) else { return }
> ```

Nice use of the identity operator !

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: 21 of 24 files reviewed, 8 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadTransport/Socks5/Socks5Connection.swift` line 136 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni4ELG-0NM0TMfZVFcE:-NidPzm55vJ65BarraD7:bvx2fkb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/Socks5/Socks5Connection.swift#L136)):*
> We should revisit this then, right now we have an inconsistent behaviour. Every operation that uses `BackgroundObserver` will always wait for connectivity, even if the flag is set to ignore it.

Background observer only starts a background task which is not the same thing as a background URL session. We configure URLSession with `ephemeral` configuration. 

URL sessions configured with background configuration are quite different, they use a delegate and a unique identifier. Their primary purpose is to offload upload and download transfers to the system daemon which enable transfers to happen even when the app is terminated or suspended.

> Right now, in case of error, we just call the completion handler with `URLError(.cancelled)`  
which means a device with a bad connectivity will likely consume all the retry attempts before it can get good connectivity.

If you refer to `URLSessionSocks5Transport`, then typically we return the underlying error in any case. `URLError(.cancelled)` is only returned when there is no error and there is no local port opened by forwarding proxy. This scenario is very unlikely to happen.
___
*[`ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift` line 53 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NiZKbp5ER4Kf8DukPQh:-NidaQwIEH9pTDH-i9Hq:b-3zov75) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e689f81b2faa2128d95e469120d01d3a69ca5cd8/ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift#L53)):*

nit
Technically this would be only true if we checked the `contentContext` `isFinal` message and it were true.
I would suggest either checking that, and logging an error if there is neither error, nor data to handle.

IIRC, TCP messages cannot be sent with 0 data inside, and the `NWConnection` API has tons of edge cases since the same API is used regardless of the protocol.
I would suggest erring on the side of caution for now.
</blockquote></details>

By the API contract, the call to `receive(minimumIncompleteLength:maximumLength:completion:)` is guaranteed to return an error or data. I don't think this code should be more complex than what it is.
___
*[`ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift` line 187 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NiZmi_M4Ta6ZF47Xtwk:-NidPiwr3fK5QHtEIDDO:b-waiqc2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e689f81b2faa2128d95e469120d01d3a69ca5cd8/ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift#L187)):*

Shouldn't we remove all the connections after cancelling them here ?

in `onEndConnection` we only remove the connection if `self.state` == `.started` because otherwise we cannot access the `openConnections`.

However, when we call `stopInner`, we immediately enter the `.stopped` state, which means the completion handler that calls `onEndConnection` 
when a connection is stopped  won't trigger the path to remove a single connection.
</blockquote></details>

I think it's fine. All connections are cancelled right away, the state is moved to `.stopped` and `openConnections` array with all connections is released upon leaving the scope.
___
*[`ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift` line 275 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NiZkq3LCp0aYSK569RQ:-NidTEr4CYzc4HrhDjIX:bh3bnx3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e689f81b2faa2128d95e469120d01d3a69ca5cd8/ios/MullvadTransport/Socks5/Socks5ForwardingProxy.swift#L275)):*

Nice use of the identity operator !
</blockquote></details>

Thanks. I think it's appropriate since each connection is unique and so there is no need to use identifiers otherwise to support equality.
___
*[`ios/MullvadTransport/DeferredCancellable.swift` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni-w1Sy1Wa4_MCuZ-ke:-NidgBJGCpx9CmEjP9fL:b-nsfjep) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/DeferredCancellable.swift#L13)):*

Right, it still doesn't do any deferred action. It doesn't postpone cancellation. It will cancel _right away_ all the tokens if it is cancelled. That is not a deferred mechanism.

If we look at the doc for `Task`, the term mentioned is "cooperative".
So this is really a `CooperativeCancellable`.

If you cancel one token, it cancels all tokens linked to it.
</blockquote></details>

I have no energy to keep this exchange going. I have renamed the type to `CancellableChain` and I hope it's satisfactory for your vision.
___
*[`ios/MullvadTransport/DeferredCancellable.swift` line 23 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Ni-wMme0dMHBz3LNe3V:-Nidjo3wDYTHVOTyfgTp:bg2v6lc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fcbfef96264a94bda73426435193bd2f07a479/ios/MullvadTransport/DeferredCancellable.swift#L23)):*

> In general the word "connect" does not necessarily have anything to do with network.
Yes, but we're writing a VPN application. So there's a really strong contextual connotation with the word "connect".

> On the other side "connect" is commonly used in Combine when connecting to the publisher that yields elements.
> the deferred cancellable object is a publisher

It doesn't have any subscribers, and doesn't implement the `Publisher` protocol. 
I don't understand why you're trying to tie `Combine` related terms or concept for something that has nothing to do with it.
</blockquote></details>

Let it be your way. Renamed to link()

mojganii (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-:-NieGG0P6Oo_JSZ6GJoQ:bnfp4nl)* status: 21 of 24 files reviewed, 9 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadTransport/URLSessionSocks5Transport.swift` line 20 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Nids4qdCbBbQ5yOIe4k:-Nids4qdCbBbQ5yOIe4l:b-sy9g1r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7adc6daed7eea61466c40ed00581c3af308e8a77/ios/MullvadTransport/URLSessionSocks5Transport.swift#L20)):*
> ```Swift
> 
>     /// The IPv4 representation of the loopback address used by `socksProxy`.
>     private let localhost = "127.0.0.1"
> ```

nit : I would suggest to use `IPv4Address.loopback` and assign that to the host.

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 9 unresolved discussions (waiting on @buggmagnet and @pronebird)

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 9 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadTransport/URLSessionSocks5Transport.swift` line 20 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-Nids4qdCbBbQ5yOIe4k:-Nii1ISX3V5HYiyOeSIx:bg4m8az) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7adc6daed7eea61466c40ed00581c3af308e8a77/ios/MullvadTransport/URLSessionSocks5Transport.swift#L20)):*

nit : I would suggest to use `IPv4Address.loopback` and assign that to the host.
</blockquote></details>

I'd too, but I spoke about that with @buggmagnet a while ago and he raised his concerns that a conversion from `IPv*Address`  to `String` implemented via `CustomDebugStringConvertible` by Apple could change in the future and break the app. Shadowsocks transport uses the same approach, perhaps we could move that string into extension since we use it that often?

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-:-NitZxr12SaFB21Ami0o:bg7igk9)* status: all files reviewed, 3 unresolved discussions (waiting on @mojganii and @pronebird)

___
*[`ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift` line 53 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348#-NiZKbp5ER4Kf8DukPQh:-NitZdda8Zph7j6Tu8iX:bqy67xj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e689f81b2faa2128d95e469120d01d3a69ca5cd8/ios/MullvadTransport/Socks5/Socks5ConnectNegotiation.swift#L53)):*

By the API contract, the call to `receive(minimumIncompleteLength:maximumLength:completion:)` is guaranteed to return an error or data. I don't think this code should be more complex than what it is.
</blockquote></details>

Fair enough

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 3 unresolved discussions (waiting on @mojganii and @pronebird)

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5348)* status: all files reviewed, 1 unresolved discussion (waiting on @mojganii)

PR #5368, opened 2023-10-26
Clarify naming of variables in `Socks5Local` datastructures

MarkusPettersson98:
This change aims to make the intent of the different fields of the `Socks5Local` datastructures more clear

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5368)

### DIFF
--- a/mullvad-management-interface/src/types/conversions/access_method.rs
+++ b/mullvad-management-interface/src/types/conversions/access_method.rs
@@ -42,6 +42,8 @@ mod settings {
 /// [`crate::types::proto::ApiAccessMethod`] type to the internal
 /// [`mullvad_types::access_method::AccessMethodSetting`] data type.
 mod data {
+    use std::net::Ipv4Addr;
+
     use crate::types::{proto, FromProtobufTypeError};
     use mullvad_types::access_method::{
         AccessMethod, AccessMethodSetting, BuiltInAccessMethod, CustomAccessMethod, Id,
@@ -142,15 +144,15 @@ mod data {
         type Error = FromProtobufTypeError;
 
         fn try_from(value: proto::access_method::Socks5Local) -> Result<Self, Self::Error> {
-            Socks5Local::from_args(
-                value.remote_ip,
-                value.remote_port as u16,
+            let remote_ip = value.remote_ip.parse::<Ipv4Addr>().map_err(|_| {
+                FromProtobufTypeError::InvalidArgument(
+                    "Could not parse Socks5 (local) message from protobuf",
+                )
+            })?;
+            Ok(AccessMethod::from(Socks5Local::new(
+                (remote_ip, value.remote_port as u16),
                 value.local_port as u16,
-            )
-            .ok_or(FromProtobufTypeError::InvalidArgument(
-                "Could not parse Socks5 (local) message from protobuf",
-            ))
-            .map(AccessMethod::from)
+            )))
         }
     }
 
@@ -163,19 +165,19 @@ mod data {
                 port,
                 authentication,
             } = value;
-            let port = port as u16;
-            match authentication.map(SocksAuth::from) {
-                Some(SocksAuth { username, password }) => {
-                    Socks5Remote::from_args_with_password(ip, port, username, password)
-                }
-                None => Socks5Remote::from_args(ip, port),
-            }
-            .ok_or({
+            let ip = ip.parse::<Ipv4Addr>().map_err(|_| {
                 FromProtobufTypeError::InvalidArgument(
                     "Could not parse Socks5 (remote) message from protobuf",
                 )
-            })
-            .map(AccessMethod::from)
+            })?;
+            let port = port as u16;
+
+            Ok(AccessMethod::from(
+                match authentication.map(SocksAuth::from) {
+                    Some(auth) => Socks5Remote::new_with_authentication((ip, port), auth),
+                    None => Socks5Remote::new((ip, port)),
+                },
+            ))
         }
     }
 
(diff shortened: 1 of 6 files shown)


mullvad-management-interface/src/types/conversions/access_method.rs
```diff
@@ -219,8 +223,8 @@ mod data {
                 CustomAccessMethod::Socks5(Socks5::Local(Socks5Local { peer, port })) => {
```
faern, line 223: I think `peer` here should also be renamed `remote_peer`. It's otherwise extremely easy to think that's where you connect to talk to the socks5, which is not actually the case. And `port` -> `local_port`, since we use that in other places. As soon as you have two of anything, you need to describe which one is for what. If `Socks5Local` only described how to connect to the socks5 server it would be fine to call the field `port`. But since it also has to contain more data, also containing ports, we should try to be more explicit IMO

faern (changes requested):

dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5368)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5368)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

___
*[`mullvad-management-interface/src/types/conversions/access_method.rs` line 223 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5368#gh-1372970962:-NhgPhGI04_cocVg1EcX:b-7rxyvo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/06899ec2321d65e16ae4a5c2f6f160469bbde7fc/mullvad-management-interface/src/types/conversions/access_method.rs#L223)):*

I think `peer` here should also be renamed `remote_peer`. It's otherwise extremely easy to think that's where you connect to talk to the socks5, which is not actually the case. And `port` -> `local_port`, since we use that in other places. As soon as you have two of anything, you need to describe which one is for what. If `Socks5Local` only described how to connect to the socks5 server it would be fine to call the field `port`. But since it also has to contain more data, also containing ports, we should try to be more explicit IMO
</blockquote></details>

I've attempted an exhaustive search+replace where `Socks5Local` pop up in order to apply the `<remote|local>_`-prefix naming scheme consistently. Please have a look :blush:

dlon (commented):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5368)* status: all files reviewed, 2 unresolved discussions (waiting on @faern and @MarkusPettersson98)

___
*[`mullvad-types/src/access_method.rs` line 271 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5368#-NhgUMEJ21Lfw31KByN_:-NhgUMEJ21Lfw31KByNa:b-5edrbg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c24dfcce4accbd88e26fc952ac14a2dfa41bca4/mullvad-types/src/access_method.rs#L271)):*
> ```XML
>     /// [`std::net::SocketAddr`] for you. If `remote_ip` or `remote_port` are
>     /// valid [`Some(Socks5Local)`] is returned, otherwise [`None`].
>     pub fn from_args(remote_ip: String, remote_port: u16, local_port: u16) -> Option<Self> {
> ```

Nit: I actually think this parsing doesn't belong here. This probably belongs in `mullvad-management-interface`

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5368)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[`mullvad-types/src/access_method.rs` line 271 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5368#-NhgUMEJ21Lfw31KByN_:-NhgWhiS9i5rum-YCKT8:bxkesi5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7c24dfcce4accbd88e26fc952ac14a2dfa41bca4/mullvad-types/src/access_method.rs#L271)):*

Nit: I actually think this parsing doesn't belong here. This probably belongs in `mullvad-management-interface`
</blockquote></details>

I'll challenge this Nit by saying that this pattern shows up elsewhere, for example in `talpid-core`

https://github.com/mullvad/mullvadvpn-app/blob/1de3f4efb1c6fd5d506353811f6a74b09c5bc172/talpid-core/src/firewall/macos.rs#L19-L47

Except that in the example, the semantics of `new` and `from_args` are swapped. In general, I think it is natural to keep constructors of types close to the type definition itself :blush:

In the case of `Socks5Local::from_args` we are constructing the type `Socks5Local` from its parts, whereas I see it as conversions between equivalent (but different) types belong in `mullvad-management-interface`.

dlon (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5368#-:-NhlF_scA8lhmwutSCRG:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5368)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

PR #5377, opened 2023-10-27
Review top bar

Rawa:
Adjust the top bar colors according to design. Before merging we need final answer from design whether or not the Mullvad text should be tinted or not. Design currently does not have it but both Android and Desktop do.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5377)


Rawa:
On hold, while waiting for a clear answer from design. Current design has no alpha on the Mullvad VPN text (like this commit), however both desktop and android has currently. It was changed from 60% to 80% in 2019. Still unclear whether we should have it or not.

Rawa:
Decision now made from design, we should have 80% opacity on the logo until further notice. I'll revert some of the PR.

PR #5396, opened 2023-11-03
show outgoing connection on map

mojganii:
this PR introduces a valuable feature that enhances the user experience by showing outgoing addresses in the connection view. This information is especially useful when access to exit IPs is available, and the implementation ensures the accuracy and completeness of the displayed data by making two network requests and waiting for their successful completion.these actions are taken to achieve the result :

1.**Network Requests for IP Addresses**: To gather the necessary information, two network requests are initiated, one for IPv4 and the other for IPv6 addresses. These requests are essential for determining the exit IP address associated with the established tunnel.
2. **Wait for Completion:** The system intelligently waits for both of these requests to be completed successfully, ensuring that we have comprehensive data for both IPv4 and IPv6 addresses if they are available .
3. **Exit Addresses Display**: Once both IPv4 and IPv6 address data are done, the connection view is updated to prominently display the exit addresses.

<table style="height: 568px; width: 600px; border-collapse: collapse; margin-left: auto; margin-right: auto;" cellspacing="20" cellpadding="28">
<tbody>
<tr>
    <th>Before</th>
    <th colspan="5">After</th>
</tr>
<tr>
<td style="width: 50%;"></td>
<td style="width: 50%;"></td>
</tr>
</tbody>
</table>

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)

### DIFF
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -650,10 +650,21 @@
 		F09A297C2A9F8A9B00EA3B6F /* VoucherTextField.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09A29792A9F8A9B00EA3B6F /* VoucherTextField.swift */; };
 		F09A297D2A9F8A9B00EA3B6F /* RedeemVoucherContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09A297A2A9F8A9B00EA3B6F /* RedeemVoucherContentView.swift */; };
 		F09A29822A9F8AD200EA3B6F /* RedeemVoucherInteractor.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09A297F2A9F8AD200EA3B6F /* RedeemVoucherInteractor.swift */; };
+		F09D04B32AE919AC003D4F89 /* OutgoingConnectionProxy.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09D04AF2AE7F83D003D4F89 /* OutgoingConnectionProxy.swift */; };
+		F09D04B52AE93CB6003D4F89 /* OutgoingConnectionProxy+Stub.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09D04B42AE93CB6003D4F89 /* OutgoingConnectionProxy+Stub.swift */; };
+		F09D04B72AE941DA003D4F89 /* OutgoingConnectionProxyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09D04B62AE941DA003D4F89 /* OutgoingConnectionProxyTests.swift */; };
+		F09D04B92AE95111003D4F89 /* OutgoingConnectionProxy.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09D04AF2AE7F83D003D4F89 /* OutgoingConnectionProxy.swift */; };
+		F09D04BB2AE95396003D4F89 /* MockURLProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09D04BA2AE95396003D4F89 /* MockURLProtocol.swift */; };
+		F09D04BD2AEBB7C5003D4F89 /* OutgoingConnectionService.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09D04BC2AEBB7C5003D4F89 /* OutgoingConnectionService.swift */; };
+		F09D04C02AF39D63003D4F89 /* OutgoingConnectionServiceTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09D04BF2AF39D63003D4F89 /* OutgoingConnectionServiceTests.swift */; };
+		F09D04C12AF39EA2003D4F89 /* OutgoingConnectionService.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09D04BC2AEBB7C5003D4F89 /* OutgoingConnectionService.swift */; };
+		F0B0E6972AFE6E7E001DC66B /* XCTest+Async.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0B0E6962AFE6E7E001DC66B /* XCTest+Async.swift */; };
 		F0C2AEFD2A0BB5CC00986207 /* NotificationProviderIdentifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0C2AEFC2A0BB5CC00986207 /* NotificationProviderIdentifier.swift */; };
 		F0C6A8432AB08E54000777A8 /* RedeemVoucherViewConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0C6A8422AB08E54000777A8 /* RedeemVoucherViewConfiguration.swift */; };
 		F0C6FA812A66E23300F521F0 /* DeleteAccountOperation.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0C6FA802A66E23300F521F0 /* DeleteAccountOperation.swift */; };
 		F0C6FA852A6A733700F521F0 /* InAppPurchaseInteractor.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0C6FA842A6A733700F521F0 /* InAppPurchaseInteractor.swift */; };
+		F0D8825B2B04F53600D3EF9A /* OutgoingConnectionData.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0D8825A2B04F53600D3EF9A /* OutgoingConnectionData.swift */; };
+		F0D8825C2B04F70E00D3EF9A /* OutgoingConnectionData.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0D8825A2B04F53600D3EF9A /* OutgoingConnectionData.swift */; };
 		F0DA87472A9CB9A2006044F1 /* AccountExpiryRow.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0DA87462A9CB9A2006044F1 /* AccountExpiryRow.swift */; };
 		F0DA87492A9CBA9F006044F1 /* AccountDeviceRow.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0DA87482A9CBA9F006044F1 /* AccountDeviceRow.swift */; };
 		F0DA874B2A9CBACB006044F1 /* AccountNumberRow.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0DA874A2A9CBACB006044F1 /* AccountNumberRow.swift */; };
@@ -1636,10 +1647,18 @@
 		F09A29792A9F8A9B00EA3B6F /* VoucherTextField.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = VoucherTextField.swift; sourceTree = "<group>"; };
 		F09A297A2A9F8A9B00EA3B6F /* RedeemVoucherContentView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RedeemVoucherContentView.swift; sourceTree = "<group>"; };
 		F09A297F2A9F8AD200EA3B6F /* RedeemVoucherInteractor.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RedeemVoucherInteractor.swift; sourceTree = "<group>"; };
+		F09D04AF2AE7F83D003D4F89 /* OutgoingConnectionProxy.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OutgoingConnectionProxy.swift; sourceTree = "<group>"; };
+		F09D04B42AE93CB6003D4F89 /* OutgoingConnectionProxy+Stub.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "OutgoingConnectionProxy+Stub.swift"; sourceTree = "<group>"; };
+		F09D04B62AE941DA003D4F89 /* OutgoingConnectionProxyTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OutgoingConnectionProxyTests.swift; sourceTree = "<group>"; };
+		F09D04BA2AE95396003D4F89 /* MockURLProtocol.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MockURLProtocol.swift; sourceTree = "<group>"; };
+		F09D04BC2AEBB7C5003D4F89 /* OutgoingConnectionService.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OutgoingConnectionService.swift; sourceTree = "<group>"; };
+		F09D04BF2AF39D63003D4F89 /* OutgoingConnectionServiceTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OutgoingConnectionServiceTests.swift; sourceTree = "<group>"; };
+		F0B0E6962AFE6E7E001DC66B /* XCTest+Async.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "XCTest+Async.swift"; sourceTree = "<group>"; };
 		F0C2AEFC2A0BB5CC00986207 /* NotificationProviderIdentifier.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NotificationProviderIdentifier.swift; sourceTree = "<group>"; };
 		F0C6A8422AB08E54000777A8 /* RedeemVoucherViewConfiguration.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RedeemVoucherViewConfiguration.swift; sourceTree = "<group>"; };
 		F0C6FA802A66E23300F521F0 /* DeleteAccountOperation.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DeleteAccountOperation.swift; sourceTree = "<group>"; };
 		F0C6FA842A6A733700F521F0 /* InAppPurchaseInteractor.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = InAppPurchaseInteractor.swift; sourceTree = "<group>"; };
+		F0D8825A2B04F53600D3EF9A /* OutgoingConnectionData.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OutgoingConnectionData.swift; sourceTree = "<group>"; };
 		F0DA87462A9CB9A2006044F1 /* AccountExpiryRow.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountExpiryRow.swift; sourceTree = "<group>"; };
 		F0DA87482A9CBA9F006044F1 /* AccountDeviceRow.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountDeviceRow.swift; sourceTree = "<group>"; };
 		F0DA874A2A9CBACB006044F1 /* AccountNumberRow.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountNumberRow.swift; sourceTree = "<group>"; };
@@ -2128,14 +2147,15 @@
 		583FE01E29C197D5006E85F9 /* Tunnel */ = {
 			isa = PBXGroup;
 			children = (
-				58B43C1825F77DB60002C8C3 /* TunnelControlView.swift */,
-				58C3F4F82964B08300D72515 /* MapViewController.swift */,
 				58A1AA8B23F5584B009F7EA6 /* ConnectionPanelView.swift */,
-				58CCA00F224249A1004F3011 /* TunnelViewController.swift */,
 				5878A27A2909649A0096FC88 /* CustomOverlayRenderer.swift */,
-				5878A278290954790096FC88 /* TunnelViewControllerInteractor.swift */,
 				58907D9424D17B4E00CFC3F5 /* DisconnectSplitButton.swift */,
+				58C3F4F82964B08300D72515 /* MapViewController.swift */,
+				F09D04BC2AEBB7C5003D4F89 /* OutgoingConnectionService.swift */,
 				5862805322428EF100F5A6E1 /* TranslucentButtonBlurView.swift */,
+				58B43C1825F77DB60002C8C3 /* TunnelControlView.swift */,
+				58CCA00F224249A1004F3011 /* TunnelViewController.swift */,
+				5878A278290954790096FC88 /* TunnelViewControllerInteractor.swift */,
 			);
 			path = Tunnel;
 			sourceTree = "<group>";
@@ -2493,6 +2513,10 @@
 				F07BF2572A26112D00042943 /* InputTextFormatterTests.swift */,
 				A9B6AC172ADE8F4300F7802A /* MigrationManagerTests.swift */,
 				58C3FA652A38549D006A450A /* MockFileCache.swift */,
+				F09D04BA2AE95396003D4F89 /* MockURLProtocol.swift */,
+				F09D04B42AE93CB6003D4F89 /* OutgoingConnectionProxy+Stub.swift */,
+				F09D04B62AE941DA003D4F89 /* OutgoingConnectionProxyTests.swift */,
+				F09D04BF2AF39D63003D4F89 /* OutgoingConnectionServiceTests.swift */,
 				A9467E7E2A29DEFE000DC21F /* RelayCacheTests.swift */,
 				A9C342C22ACC3EE90045F00E /* RelayCacheTracker+Stubs.swift */,
 				584B26F3237434D00073B10E /* RelaySelectorTests.swift */,
@@ -2503,6 +2527,7 @@
 				A9E0317B2ACBFC7E0095D843 /* TunnelStore+Stubs.swift */,
 				A9E031792ACB0AE70095D843 /* UIApplication+Stubs.swift */,
 				58165EBD2A262CBB00688EAD /* WgKeyRotationTests.swift */,
+				F0B0E6962AFE6E7E001DC66B /* XCTest+Async.swift */,
 			);
 			path = MullvadVPNTests;
 			sourceTree = "<group>";
@@ -2737,6 +2762,7 @@
 				58C774C929AB543C003A1A56 /* Containers */,
 				58CAF9F22983D32200BE19F7 /* Coordinators */,
 				583FE02329C1AC9F006E85F9 /* Extensions */,
+				F09D04B82AE94F27003D4F89 /* GeneralAPIs */,
 				58B26E1F2943516500D5980C /* Notifications */,
 				586A950B2901250A007BAF2B /* Operations */,
 				5864859729A0D012006C5743 /* Presentation controllers */,
@@ -3016,6 +3042,15 @@
 				F09A29792A9F8A9B00EA3B6F /* VoucherTextField.swift */,
 			);
 			path = RedeemVoucher;
+			sourceTree = "<group>";
+		};
+		F09D04B82AE94F27003D4F89 /* GeneralAPIs */ = {
+			isa = PBXGroup;
+			children = (
+				F0D8825A2B04F53600D3EF9A /* OutgoingConnectionData.swift */,
+				F09D04AF2AE7F83D003D4F89 /* OutgoingConnectionProxy.swift */,
+			);
+			path = GeneralAPIs;
 			sourceTree = "<group>";
 		};
 		F0E361892A4ADCF500AEEF2B /* Welcome */ = {
@@ -4095,15 +4130,19 @@
 				A9B6AC1B2ADEA3AD00F7802A /* MemoryCache.swift in Sources */,
 				A9A5F9EA2ACB05160083449F /* Bundle+ProductVersion.swift in Sources */,
 				A9A5F9EB2ACB05160083449F /* CharacterSet+IPAddress.swift in Sources */,
+				F0D8825C2B04F70E00D3EF9A /* OutgoingConnectionData.swift in Sources */,
 				A9A5F9EC2ACB05160083449F /* CodingErrors+CustomErrorDescription.swift in Sources */,
 				A9A5F9ED2ACB05160083449F /* NSRegularExpression+IPAddress.swift in Sources */,
 				A9A5F9EE2ACB05160083449F /* RESTCreateApplePaymentResponse+Localization.swift in Sources */,
 				A9A5F9EF2ACB05160083449F /* String+AccountFormatting.swift in Sources */,
 				A9A5F9F02ACB05160083449F /* String+FuzzyMatch.swift in Sources */,
+				F09D04C12AF39EA2003D4F89 /* OutgoingConnectionService.swift in Sources */,
 				A9A5F9F12ACB05160083449F /* String+Split.swift in Sources */,
 				A9A5F9F22ACB05160083449F /* NotificationConfiguration.swift in Sources */,
 				A9A5F9F32ACB05160083449F /* AccountExpirySystemNotificationProvider.swift in Sources */,
 				A9A5F9F52ACB05160083449F /* RegisteredDeviceInAppNotificationProvider.swift in Sources */,
+				F09D04B72AE941DA003D4F89 /* OutgoingConnectionProxyTests.swift in Sources */,
+				F09D04B92AE95111003D4F89 /* OutgoingConnectionProxy.swift in Sources */,
 				A9A5F9F62ACB05160083449F /* TunnelStatusNotificationProvider.swift in Sources */,
 				A9A5F9F72ACB05160083449F /* NotificationProviderProtocol.swift in Sources */,
 				A9A5F9F82ACB05160083449F /* NotificationProviderIdentifier.swift in Sources */,
@@ -4144,12 +4183,14 @@
 				A9A5FA142ACB05160083449F /* MapConnectionStatusOperation.swift in Sources */,
 				A9A5FA152ACB05160083449F /* RedeemVoucherOperation.swift in Sources */,
 				A9A5FA162ACB05160083449F /* RotateKeyOperation.swift in Sources */,
+				F09D04B52AE93CB6003D4F89 /* OutgoingConnectionProxy+Stub.swift in Sources */,
 				A9A5FA172ACB05160083449F /* SendTunnelProviderMessageOperation.swift in Sources */,
 				A9A5FA182ACB05160083449F /* SetAccountOperation.swift in Sources */,
 				A9A5FA192ACB05160083449F /* StartTunnelOperation.swift in Sources */,
 				A9A5FA1A2ACB05160083449F /* StopTunnelOperation.swift in Sources */,
 				A9A5FA1B2ACB05160083449F /* Tunnel.swift in Sources */,
 				A9A5FA1C2ACB05160083449F /* Tunnel+Messaging.swift in Sources */,
+				F09D04BB2AE95396003D4F89 /* MockURLProtocol.swift in Sources */,
 				A9A5FA1D2ACB05160083449F /* TunnelBlockObserver.swift in Sources */,
 				A9A5FA1E2ACB05160083449F /* TunnelConfiguration.swift in Sources */,
 				A9A5FA1F2ACB05160083449F /* TunnelInteractor.swift in Sources */,
@@ -4159,6 +4200,7 @@
 				A9A5FA222ACB05160083449F /* TunnelObserver.swift in Sources */,
 				A988A3E22AFE54AC0008D2C7 /* AccountExpiry.swift in Sources */,
 				A9E0317F2ACC331C0095D843 /* TunnelStatusBlockObserver.swift in Sources */,
+				F09D04C02AF39D63003D4F89 /* OutgoingConnectionServiceTests.swift in Sources */,
 				A9A5FA232ACB05160083449F /* TunnelState.swift in Sources */,
 				A9A5FA242ACB05160083449F /* TunnelStore.swift in Sources */,
 				A9A5FA252ACB05160083449F /* UpdateAccountDataOperation.swift in Sources */,
@@ -4174,6 +4216,7 @@
 				A9A5FA2E2ACB05160083449F /* FileCacheTests.swift in Sources */,
 				A9A5FA2F2ACB05160083449F /* FixedWidthIntegerArithmeticsTests.swift in Sources */,
 				A9A5FA302ACB05160083449F /* InputTextFormatterTests.swift in Sources */,
+				F0B0E6972AFE6E7E001DC66B /* XCTest+Async.swift in Sources */,
 				A9A5FA312ACB05160083449F /* MockFileCache.swift in Sources */,
 				A9A5FA322ACB05160083449F /* RelayCacheTests.swift in Sources */,
 				A9A5FA332ACB05160083449F /* RelaySelectorTests.swift in Sources */,
@@ -4485,6 +4528,7 @@
 				F0C6A8432AB08E54000777A8 /* RedeemVoucherViewConfiguration.swift in Sources */,
 				7AF10EB42ADE85BC00C090B9 /* RelayFilterCoordinator.swift in Sources */,
 				58FB865526E8BF3100F188BC /* StorePaymentManagerError.swift in Sources */,
+				F09D04B32AE919AC003D4F89 /* OutgoingConnectionProxy.swift in Sources */,
 				58FD5BF42428C67600112C88 /* InAppPurchaseButton.swift in Sources */,
 				7AF10EB22ADE859200C090B9 /* AlertViewController.swift in Sources */,
 				587D9676288989DB00CD8F1C /* NSLayoutConstraint+Helpers.swift in Sources */,
@@ -4502,12 +4546,14 @@
 				063687BA28EB234F00BE7161 /* PacketTunnelTransport.swift in Sources */,
 				A9C342C12ACC37E30045F00E /* TunnelStatusBlockObserver.swift in Sources */,
 				587425C12299833500CA2045 /* RootContainerViewController.swift in Sources */,
+				F09D04BD2AEBB7C5003D4F89 /* OutgoingConnectionService.swift in Sources */,
 				5896AE84246D5889005B36CB /* CustomDateComponentsFormatting.swift in Sources */,
 				5871167F2910035700D41AAC /* PreferencesInteractor.swift in Sources */,
 				7A9CCCC22A96302800DD6A34 /* SafariCoordinator.swift in Sources */,
 				58E20771274672CA00DE5D77 /* LaunchViewController.swift in Sources */,
 				F0E8CC032A4C753B007ED3B4 /* WelcomeViewController.swift in Sources */,
 				584D26C4270C855B004EA533 /* PreferencesDataSource.swift in Sources */,
+				F0D8825B2B04F53600D3EF9A /* OutgoingConnectionData.swift in Sources */,
 				7A9CCCB62A96302800DD6A34 /* OutOfTimeCoordinator.swift in Sources */,
 				58FD5BF024238EB300112C88 /* SKProduct+Formatting.swift in Sources */,
 				58B43C1925F77DB60002C8C3 /* TunnelControlView.swift in Sources */,
--- a/ios/MullvadVPN/Coordinators/ApplicationCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/ApplicationCoordinator.swift
@@ -77,6 +77,7 @@ final class ApplicationCoordinator: Coordinator, Presenting, RootContainerViewCo
     private let accountsProxy: RESTAccountHandling
     private var tunnelObserver: TunnelObserver?
     private var appPreferences: AppPreferencesDataSource
+    private var outgoingConnectionService: OutgoingConnectionServiceHandling
 
     private var outOfTimeTimer: Timer?
 
@@ -91,6 +92,7 @@ final class ApplicationCoordinator: Coordinator, Presenting, RootContainerViewCo
         apiProxy: APIQuerying,
         devicesProxy: DeviceHandling,
         accountsProxy: RESTAccountHandling,
+        outgoingConnectionService: OutgoingConnectionServiceHandling,
         appPreferences: AppPreferencesDataSource
     ) {
         self.tunnelManager = tunnelManager
@@ -100,6 +102,7 @@ final class ApplicationCoordinator: Coordinator, Presenting, RootContainerViewCo
         self.devicesProxy = devicesProxy
         self.accountsProxy = accountsProxy
         self.appPreferences = appPreferences
+        self.outgoingConnectionService = outgoingConnectionService
 
         super.init()
 
@@ -676,7 +679,10 @@ final class ApplicationCoordinator: Coordinator, Presenting, RootContainerViewCo
     }
 
     private func makeTunnelCoordinator() -> TunnelCoordinator {
-        let tunnelCoordinator = TunnelCoordinator(tunnelManager: tunnelManager)
+        let tunnelCoordinator = TunnelCoordinator(
+            tunnelManager: tunnelManager,
+            outgoingConnectionService: outgoingConnectionService
+        )
 
         tunnelCoordinator.showSelectLocationPicker = { [weak self] in
             self?.router.present(.selectLocation, animated: true)
--- a/ios/MullvadVPN/Coordinators/TunnelCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/TunnelCoordinator.swift
@@ -25,10 +25,16 @@ class TunnelCoordinator: Coordinator, Presenting {
 
     var showSelectLocationPicker: (() -> Void)?
 
-    init(tunnelManager: TunnelManager) {
+    init(
+        tunnelManager: TunnelManager,
+        outgoingConnectionService: OutgoingConnectionServiceHandling
+    ) {
         self.tunnelManager = tunnelManager
 
-        let interactor = TunnelViewControllerInteractor(tunnelManager: tunnelManager)
+        let interactor = TunnelViewControllerInteractor(
+            tunnelManager: tunnelManager,
+            outgoingConnectionService: outgoingConnectionService
+        )
         controller = TunnelViewController(interactor: interactor)
 
         super.init()
--- a/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionData.swift
+++ b/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionData.swift
@@ -1,0 +1,28 @@
+//
+//  OutgoingConnectionData.swift
+//  MullvadVPN
+//
+//  Created by Mojgan on 2023-11-15.
+//  Copyright © 2023 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+import Network
+
+typealias IPV4ConnectionData = OutgoingConnectionData<IPv4Address>
+typealias IPV6ConnectionData = OutgoingConnectionData<IPv6Address>
+
+// MARK: - OutgoingConnectionData
+
+struct OutgoingConnectionData<T: Codable & IPAddress>: Codable, Equatable {
+    let ip: T
+    let exitIP: Bool
+
+    enum CodingKeys: String, CodingKey {
+        case ip, exitIP = "mullvad_exit_ip"
+    }
+
+    static func == (lhs: Self, rhs: Self) -> Bool {
+        lhs.ip.rawValue == rhs.ip.rawValue && lhs.exitIP == rhs.exitIP
+    }
+}
--- a/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift
+++ b/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift
@@ -1,0 +1,94 @@
+//
+//  OutgoingConnectionProxy.swift
+//  MullvadREST
+//
+//  Created by Mojgan on 2023-10-24.
+//  Copyright © 2023 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+import MullvadREST
+import MullvadTypes
+import Network
+
+protocol OutgoingConnectionHandling {
+    func getIPV6(retryStrategy: REST.RetryStrategy) async throws -> IPV6ConnectionData
+    func getIPV4(retryStrategy: REST.RetryStrategy) async throws -> IPV4ConnectionData
+}
+
+final class OutgoingConnectionProxy: OutgoingConnectionHandling {
+    private enum ExitIPVersion: String {
+        case v4 = "ipv4", v6 = "ipv6"
+
+        var host: String {
+            "\(rawValue).am.i.mullvad.net"
+        }
+    }
+
+    let urlSession: URLSession
+
+    init(urlSession: URLSession) {
+        self.urlSession = urlSession
+    }
+
+    func getIPV6(retryStrategy: REST.RetryStrategy) async throws -> IPV6ConnectionData {
+        try await perform(retryStrategy: retryStrategy, version: .v6)
+    }
+
+    func getIPV4(retryStrategy: REST.RetryStrategy) async throws -> IPV4ConnectionData {
+        try await perform(retryStrategy: retryStrategy, version: .v4)
+    }
+
+    private func perform<T: Decodable>(retryStrategy: REST.RetryStrategy, version: ExitIPVersion) async throws -> T {
+        let delayIterator = retryStrategy.makeDelayIterator()
+        for _ in 0 ..< retryStrategy.maxRetryCount {
+            do {
+                return try await perform(host: version.host)
+            } catch {
+                // ignore if request is cancelled
+                if case URLError.cancelled = error {
+                    throw error
+                } else {
+                    // retry with the delay
+                    guard let delay = delayIterator.next() else { throw error }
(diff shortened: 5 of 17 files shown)


rablador (commented):
> Reviewed 2 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 2 of 15 files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NieTUfe8IOj8Y0TUUBm:-NieTUfe8IOj8Y0TUUBn:b-h365xv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L94)):*
> ```Swift
>         let exitIP: Bool
> 
>         enum CodingKeys: String, CodingKey {
> ```

This struct already conforms to Codable, so no need to encode or decode manually. This enum, along with `init(from decoder: Decoder)` and `encode(to encoder: Encoder)`, can be removed.
___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 98 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NieTw7y0P4-ZgwCAgY4:-NieTw7y0P4-ZgwCAgY5:b-udpr7g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L98)):*
> ```Swift
>         }
> 
>         init(ip: T, exitIP: Bool) {
> ```

This init is never used and should be ok to remove.
___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 116 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NieU1bnE4Hkh5F2IhZl:-NieU1bnE4Hkh5F2IhZm:bnovo2s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L116)):*
> ```Swift
> 
>         static func == (
>             lhs: OutgoingConnectionProxy.OutgoingConnectionData<T>,
> ```

Nit: `OutgoingConnectionProxy.OutgoingConnectionData<T>` can be substituted with `Self` instead to increase readability.

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 2 of 15 files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NieTUfe8IOj8Y0TUUBm:-NiiSXqd4VGCF5k2gvK8:b3h194p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L94)):*

This struct already conforms to Codable, so no need to encode or decode manually. This enum, along with `init(from decoder: Decoder)` and `encode(to encoder: Encoder)`, can be removed.
</blockquote></details>

Update: We actually need the mapping enum here, because of `exitIP = "mullvad_exit_ip"`, but the rest is ok to remove.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 1 of 15 files reviewed, 3 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NieTUfe8IOj8Y0TUUBm:-NiiryMJBwOH8RwNjh4R:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L94)):*

Update: We actually need the mapping enum here, because of `exitIP = "mullvad_exit_ip"`, but the rest is ok to remove.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 98 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NieTw7y0P4-ZgwCAgY4:-Niis-Jx6DCzqYiprKWt:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L98)):*

This init is never used and should be ok to remove.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 116 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NieU1bnE4Hkh5F2IhZl:-NiitLkEEM9ryCHWAeuH:b-3cslex) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L116)):*

Nit: `OutgoingConnectionProxy.OutgoingConnectionData<T>` can be substituted with `Self` instead to increase readability.
</blockquote></details>

indeed!

rablador (changes requested):
> Reviewed 13 of 15 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/View controllers/Tunnel/OutgoingConnectionService.swift` line 67 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NiirGc11IO9aEUDdluH:-NiirGc11IO9aEUDdluI:bhxnflg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPN/View%20controllers/Tunnel/OutgoingConnectionService.swift#L67)):*
> ```Swift
> 
>     var outAddress: String? {
>         let v4 = ipv4.exitIP ? ipv4.ip : nil
> ```

This can be compressed a bit, also only adding the line break if there are two items to separate them:



_Code snippet:_
```Swift
        let v4: IPAddress? = ipv4.exitIP ? ipv4.ip : nil
        let v6: IPAddress? = ipv6.flatMap { $0.exitIP ? $0.ip : nil }

        let string = [v4, v6].compactMap { $0 }.map { $0.debugDescription }.joined(separator: "\n")
        return string.isEmpty ? nil : string
```
___
*[`ios/MullvadVPNTests/OutgoingConnectionServiceTests.swift` line 35 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NiivQ-M7nqrC3_r8ri-:-NiivQ-M7nqrC3_r8ri0:bbm2x8p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPNTests/OutgoingConnectionServiceTests.swift#L35)):*
> ```Swift
>         let result = try await outgoingConnectionService.getOutgoingConnectionInfo()
>         if result.ipv4 == .mock,
>            let ipv6 = result.ipv6,
> ```

Just `result.ipv6 == .mock` works too.

pronebird (changes requested):
> Reviewed 14 of 15 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 3 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 91 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NinXOYJ6eMy2Q1RiEvy:-NinXOYJ6eMy2Q1RiEvz:budjsqy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L91)):*
> ```Swift
> 
>     struct OutgoingConnectionData<T: IPAddressType>: Codable, Equatable {
>         let ip: T?
> ```

Can `ip` really be `nil`?
___
*[`ios/MullvadVPN/View controllers/Tunnel/OutgoingConnectionService.swift` line 67 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NiirGc11IO9aEUDdluH:-NinXegfCMdJOyddVzBL:b-bcdght) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPN/View%20controllers/Tunnel/OutgoingConnectionService.swift#L67)):*

This can be compressed a bit, also only adding the line break if there are two items to separate them:
</blockquote></details>

@rablador transforming IP to string can be done using `"\(ipAddress)"`, no need to call `debugDescription` directly.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 11 of 15 files reviewed, 3 unresolved discussions (waiting on @pronebird and @rablador)

___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 91 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NinXOYJ6eMy2Q1RiEvy:-NinfvhOA0JBZ0fi7reo:bgwgy4w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L91)):*

Can `ip` really be `nil`?
</blockquote></details>

After checking the server response I figured out it cannot be optional. 
___
*[`ios/MullvadVPN/View controllers/Tunnel/OutgoingConnectionService.swift` line 67 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NiirGc11IO9aEUDdluH:-NinDOlAAVVWXPhwidSv:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPN/View%20controllers/Tunnel/OutgoingConnectionService.swift#L67)):*

This can be compressed a bit, also only adding the line break if there are two items to separate them:
</blockquote></details>

Done.
___
*[`ios/MullvadVPNTests/OutgoingConnectionServiceTests.swift` line 35 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NiivQ-M7nqrC3_r8ri-:-NioMspz9z3SXpDGG0D9:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7461215221267bf8ea9f085339f0c385dc8d2397/ios/MullvadVPNTests/OutgoingConnectionServiceTests.swift#L35)):*

Just `result.ipv6 == .mock` works too.
</blockquote></details>

Done.

buggmagnet (changes requested):
That's really good work ! 

> Reviewed 7 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-:-NioggXB9lCiD93-qYMZ:bx6zp3s)* status: 11 of 15 files reviewed, 13 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/SceneDelegate.swift` line 73 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nio9ENq8Zte_3owxAxa:-Nio9ENq8Zte_3owxAxb:b-wwlewi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/SceneDelegate.swift#L73)):*
> ```Swift
>             accountsProxy: appDelegate.accountsProxy,
>             outgoingConnectionService: OutgoingConnectionService(
>                 outgoingConnectionProxy: OutgoingConnectionProxy(urlSession: .shared)
> ```

We don't want to cache the result of calls `to am.i.mullvad.net`, therefore we don't want to use the `.shared` sessions.

Please use `URLSessionConfiguration.ephemeral` instead for getting a `URLSession` object
___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 83 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioVJIl8TGpSXbp8nQC:-NioVJIl8TGpSXbp8nQD:bu6u8i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L83)):*
> ```Swift
> }
> 
> extension OutgoingConnectionProxy {
> ```

IMHO `OutgoingConnectionData` should be in its own file and not part of `OutgoingConnectionProxy` 

The protocol signature would look simpler as well
```swift
protocol OutgoingConnectionHandling {
    func getIPV6(retryStrategy: REST.RetryStrategy) async throws -> IPV6ConnectionData
    func getIPV4(retryStrategy: REST.RetryStrategy) async throws -> IPV4ConnectionData
}

```
___
*[`ios/MullvadVPN/View controllers/Tunnel/OutgoingConnectionService.swift` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioYboUEavqbb6AbCTx:-NioYboUEavqbb6AbCTy:b-yetoj7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/View%20controllers/Tunnel/OutgoingConnectionService.swift#L30)):*
> ```Swift
> 
>     func getOutgoingConnectionInfo() async throws -> OutgoingConnectionInfo {
>         try await withThrowingTaskGroup(of: OutgoingConnectionResult.self) { taskGroup -> OutgoingConnectionInfo in
> ```


We can simplify this.
However, technically we are not running the calls in parallel but that shouldn't matter in practice. Also we're shaving 40 lines. IMHO it's worth it.

```swift
func getOutgoingConnectionInfo() async throws -> OutgoingConnectionInfo {
    let ipv4 = try await outgoingConnectionProxy.getIPV4(retryStrategy: .default)
    let ipv6 = try await outgoingConnectionProxy.getIPV6(retryStrategy: .noRetry)

    return OutgoingConnectionInfo(ipv4: ipv4, ipv6: ipv6)
}

```

I don't think we should be touching the task priority until we have good reason to believe that this would be significantly better for the UX.
If we absolutely need to, then we can change the call site instead (which I still think we shouldn't)

```swift
outgoingConnectionTask = Task(priority: .userInitiated) { [weak self] in
    guard let outgoingConnectionInfo = try await self?.outgoingConnectionService
        .getOutgoingConnectionInfo() else {
        return
    }
    await self?.didGetOutGoingAddress?(outgoingConnectionInfo)
}

```
___
*[`ios/MullvadVPN/View controllers/Tunnel/TunnelControlView.swift` line 153 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NiofFGc81OSXatmaapG:-NiofFGc81OSXatmaapH:b-6bn07n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/View%20controllers/Tunnel/TunnelControlView.swift#L153)):*
> ```Swift
>         if let tunnelRelay = tunnelState.relay {
>             connectionPanel.dataSource = ConnectionPanelData(
>                 inAddress: "\(tunnelRelay.endpoint.ipv4Relay) UDP",
> ```

We shouldn't be hardcoding `UDP` here since it's depending on whether we're using the UDP-over-TCP obfuscation.

Maybe let's add a TODO comment for now that we will be solving this soon.
___
*[`ios/MullvadVPN/View controllers/Tunnel/TunnelViewControllerInteractor.swift` line 42 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NiogVqZ1t8J9v4ZNLYs:-NiogVqZ1t8J9v4ZNLYt:b-a3l107) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/View%20controllers/Tunnel/TunnelViewControllerInteractor.swift#L42)):*
> ```Swift
>         self.outgoingConnectionService = outgoingConnectionService
> 
>         let tunnelObserver = TunnelBlockObserver(
> ```

Is there anyway we can do this when we start the tunnel instead of the init ?
___
*[`ios/MullvadVPNTests/MockURLProtocol.swift` line 12 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioRlCQ7FuU27IrJ2ds:-NioRlCQ7FuU27IrJ2dt:bl7piul) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/MockURLProtocol.swift#L12)):*
> ```Swift
> 
> class MockURLProtocol: URLProtocol {
>     static var error: Error?
> ```

Same remark as in `OutgoingConnectionProxyStub`, this should not be `static`
___
*[`ios/MullvadVPNTests/OutgoingConnectionProxy+Stub.swift` line 12 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioR7CjDSDU0zAYNljO:-NioR7CjDSDU0zAYNljP:b-2to0sj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/OutgoingConnectionProxy%2BStub.swift#L12)):*
> ```Swift
> 
> struct OutgoingConnectionProxyStub: OutgoingConnectionHandling {
>     static var hasError = false
> ```

Please let's not use `static` unless we absolutely have to. It always leads to hard to test designs and unwanted shared states.

Here in particular, it makes it so that all the `OutgoingConnectionProxyStub` have the same error state which is definitely not what we want when running multiple tests in parallel.
___
*[`ios/MullvadVPNTests/OutgoingConnectionProxyTests.swift` line 18 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioARt2EugA95O2tPty:-NioARt2EugA95O2tPtz:b8f1zsb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/OutgoingConnectionProxyTests.swift#L18)):*
> ```Swift
>     private let encoder: JSONEncoder = REST.Coding.makeJSONEncoder()
> 
>     override func setUp() {
> ```

You can use `setUpWithError() throws` and remove the `force try` as well as the `swiftlint:disable` commands instead 

Remember to also use `tearDownWithError() throws` as well in that case
___
*[`ios/MullvadVPNTests/OutgoingConnectionProxyTests.swift` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioB08k95LulEe3axoj:-NioB08k95LulEe3axok:b4ugqmh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/OutgoingConnectionProxyTests.swift#L34)):*
> ```Swift
>     func testNoInternetConnection() async throws {
>         let failureExpectation = expectation(description: "should fail")
>         let successExpectation = expectation(description: "should not succeed")
> ```

There are a couple of things we can improve in this test.

- Expectations :
Because we are using inverted expectations here, when this fails, the message produced will be 
`Fulfilled inverted expectation "should not succeed".` 
Which is a double negation. 

In essence, we should keep the same kind messaging so that the error is easy to read and understand.
In this case, it would make sense to read `Fulfilled inverted expectation "Received IPv4".`

- Expectation descriptions :
A message such as "should fail" is not very descriptive. What are we expecting here ? We shouldn't have to read what the code does when a test fail to know _what_ failed.
A much better message would be "Did not receive IPv4". 
That way, if the test fails, this is what we would see "Asynchronous wait failed [...] with unfulfilled expectations: "Did not receive IPv4".
Now we know that we expected to not receive IPV4, and that the expectation was not fullfilled, therefore, we did receive an IPv4. We probably know how to fix this.

- Expectation names :
As important as the message, usually, we try to name the expectation according to what's being expected.
In this case, a good name would be `let noIPv4Expectation = expectation(description: "Did not receive IPv4")` for example.

- Catching the error :

Unfortunately, the `XCTest` API is not well suited for `async` results, otherwise, I would have said to use `XCTAssertThrowsError` which is suited for that.

So I wrote a helper method to help with that
```swift
extension XCTest {
    func XCTAssertThrowsErrorAsync<T: Sendable>(
        _ expression: @autoclosure () async throws -> T,
        _ message: @autoclosure () -> String = "",
        file: StaticString = #filePath,
        line: UInt = #line,
        _ errorHandler: (_ error: Error) -> Void = { _ in }
    ) async {
        do {
            _ = try await expression()
            XCTFail(message(), file: file, line: line)
        } catch {
            errorHandler(error)
        }
    }
}
```

Here's what the rewritten test looks like now 

```swift
func testGetIPv4FailsWhenNotConnected() async throws {
    let noIPv4Expectation = expectation(description: "Did not receive IPv4")
    MockURLProtocol.error = URLError(URLError.notConnectedToInternet)
    MockURLProtocol.requestHandler = nil

    await XCTAssertThrowsErrorAsync(try await outgoingConnectionProxy.getIPV4(retryStrategy: .noRetry)) { error in
        noIPv4Expectation.fulfill()
        XCTAssertEqual((error as? URLError)?.code, .notConnectedToInternet)
    }
    await fulfillment(of: [noIPv4Expectation], timeout: 1)
}

```

I suggest rewriting the other tests in this file with these improvements in mind
___
*[`ios/MullvadVPNTests/OutgoingConnectionServiceTests.swift` line 54 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioSOWwEP4eytJE3UFL:-NioSOWxCzW85e0O4sl-:b-l6d6lg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/OutgoingConnectionServiceTests.swift#L54)):*
> ```Swift
> }
> 
> enum MockNetworkError: Error {
> ```

nit
Technically, this is a [stub, not a mock](https://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs)

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 11 of 15 files reviewed, 12 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/SceneDelegate.swift` line 73 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nio9ENq8Zte_3owxAxa:-NiohmTj0WyVRezg-qUf:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/SceneDelegate.swift#L73)):*

We don't want to cache the result of calls `to am.i.mullvad.net`, therefore we don't want to use the `.shared` sessions.

Please use `URLSessionConfiguration.ephemeral` instead for getting a `URLSession` object
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/View controllers/Tunnel/OutgoingConnectionService.swift` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioYboUEavqbb6AbCTx:-NiohzRT5RDj9457OE7t:b-wu3ntu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/View%20controllers/Tunnel/OutgoingConnectionService.swift#L30)):*

We can simplify this.
However, technically we are not running the calls in parallel but that shouldn't matter in practice. Also we're shaving 40 lines. IMHO it's worth it.

```swift
func getOutgoingConnectionInfo() async throws -> OutgoingConnectionInfo {
    let ipv4 = try await outgoingConnectionProxy.getIPV4(retryStrategy: .default)
    let ipv6 = try await outgoingConnectionProxy.getIPV6(retryStrategy: .noRetry)

    return OutgoingConnectionInfo(ipv4: ipv4, ipv6: ipv6)
}

```

I don't think we should be touching the task priority until we have good reason to believe that this would be significantly better for the UX.
If we absolutely need to, then we can change the call site instead (which I still think we shouldn't)

```swift
outgoingConnectionTask = Task(priority: .userInitiated) { [weak self] in
    guard let outgoingConnectionInfo = try await self?.outgoingConnectionService
        .getOutgoingConnectionInfo() else {
        return
    }
    await self?.didGetOutGoingAddress?(outgoingConnectionInfo)
}

```
</blockquote></details>

but task group does it parallely.regarding the task priority I think we need to change it for better UX
___
*[`ios/MullvadVPN/View controllers/Tunnel/TunnelControlView.swift` line 153 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NiofFGc81OSXatmaapG:-NioiP5r4Dcg6Rb4g8D0:bltvybu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/View%20controllers/Tunnel/TunnelControlView.swift#L153)):*

We shouldn't be hardcoding `UDP` here since it's depending on whether we're using the UDP-over-TCP obfuscation.

Maybe let's add a TODO comment for now that we will be solving this soon.
</blockquote></details>

I'll put todo for that
___
*[`ios/MullvadVPN/View controllers/Tunnel/TunnelViewControllerInteractor.swift` line 42 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NiogVqZ1t8J9v4ZNLYs:-Nioji3J7iFlu4KHRMAp:b-3meg1x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/View%20controllers/Tunnel/TunnelViewControllerInteractor.swift#L42)):*

Is there anyway we can do this when we start the tunnel instead of the init ?
</blockquote></details>

is it the responsibility of tunnel to fetch exit ip.I don't think so.
___
*[`ios/MullvadVPNTests/MockURLProtocol.swift` line 12 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioRlCQ7FuU27IrJ2ds:-NitGb3g6fBlEY1rgdSd:b-a0mh7e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/MockURLProtocol.swift#L12)):*

Same remark as in `OutgoingConnectionProxyStub`, this should not be `static`
</blockquote></details>

if I'm gonna eliminate static variable in this particular part I need to go protocol-oriented for url-session and data task and so on that takes much efforts and bring more complexity.do you have any solution for this part?
___
*[`ios/MullvadVPNTests/OutgoingConnectionProxy+Stub.swift` line 12 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioR7CjDSDU0zAYNljO:-NiomBiS5ojjiXmuM9-f:b-jnts5j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/OutgoingConnectionProxy%2BStub.swift#L12)):*

Please let's not use `static` unless we absolutely have to. It always leads to hard to test designs and unwanted shared states.

Here in particular, it makes it so that all the `OutgoingConnectionProxyStub` have the same error state which is definitely not what we want when running multiple tests in parallel.
</blockquote></details>

fair enough 
___
*[`ios/MullvadVPNTests/OutgoingConnectionProxyTests.swift` line 18 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioARt2EugA95O2tPty:-Nion-k80fCO74I4m7Ex:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/OutgoingConnectionProxyTests.swift#L18)):*

You can use `setUpWithError() throws` and remove the `force try` as well as the `swiftlint:disable` commands instead 

Remember to also use `tearDownWithError() throws` as well in that case
</blockquote></details>

Done.
___
*[`ios/MullvadVPNTests/OutgoingConnectionServiceTests.swift` line 54 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioSOWwEP4eytJE3UFL:-NionKvO7Z9eLbVOWP6M:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/OutgoingConnectionServiceTests.swift#L54)):*

nit
Technically, this is a [stub, not a mock](https://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs)
</blockquote></details>

Done

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 5 of 16 files reviewed, 8 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/View controllers/Tunnel/OutgoingConnectionService.swift` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioYboUEavqbb6AbCTx:-Nj6MpDd5rL9C8Z0_KRV:b-e0o3oj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/View%20controllers/Tunnel/OutgoingConnectionService.swift#L30)):*
> but task group does it parallely.
Yes but there are a couple of considerations at play here:
- We are only running 2 tasks
- We are doing network calls
- 3 lines of code is vastly easier to understand than 40 with a custom built enum

There's very likely more overhead to try to run 2 tasks in parallel rather than running them one after the other given how simple the calls are (and how fast iPhone processors are nowadays)
And since it's networking code, on a bad connection, it won't make any difference whether the tasks are running in parallel rather than in sequence.

> I think we need to change it for better UX
Doing this will in the best case, not do anything, and in the worst case, make the whole system perform worse.
This is a good example of the [Priority Inversion problem](https://en.wikipedia.org/wiki/Priority_inversion).
However, the swift async runtime should help us in those situations, hopefully. Happy to talk more about why that's a problematic design, but not in this comment.
___
*[`ios/MullvadVPNTests/MockURLProtocol.swift` line 12 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioRlCQ7FuU27IrJ2ds:-Nj6djR8E134O52LOUkU:bxua40e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/MockURLProtocol.swift#L12)):*

if I'm gonna eliminate static variable in this particular part I need to go protocol-oriented for url-session and data task and so on that takes much efforts and bring more complexity.do you have any solution for this part?
</blockquote></details>

Technically, the only part where we use that `URLSession` object is in `OutgoingConnectionProxy` where we do 
`let (data, response) = try await urlSession.data(from: url)` 

Ideally, we should have extracted that call to a single interface, then we wouldn't need all this mocking of `URLProtocol` and the complicated setup of tests.
That's a bit unfortunate, but it's not the end of the world.

I will create a cleanup janitor task instead, it's not critical.

pronebird (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 6 of 16 files reviewed, 7 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPN/View controllers/Tunnel/OutgoingConnectionService.swift` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioYboUEavqbb6AbCTx:-Nj6v_FY16o58UByxegb:b-9g9fcm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/View%20controllers/Tunnel/OutgoingConnectionService.swift#L30)):*

> but task group does it parallely.
Yes but there are a couple of considerations at play here:
- We are only running 2 tasks
- We are doing network calls
- 3 lines of code is vastly easier to understand than 40 with a custom built enum

There's very likely more overhead to try to run 2 tasks in parallel rather than running them one after the other given how simple the calls are (and how fast iPhone processors are nowadays)
And since it's networking code, on a bad connection, it won't make any difference whether the tasks are running in parallel rather than in sequence.

> I think we need to change it for better UX
Doing this will in the best case, not do anything, and in the worst case, make the whole system perform worse.
This is a good example of the [Priority Inversion problem](https://en.wikipedia.org/wiki/Priority_inversion).
However, the swift async runtime should help us in those situations, hopefully. Happy to talk more about why that's a problematic design, but not in this comment.
</blockquote></details>

Let's not get into micro optimizations.

Parallelization for IP lookup requests makes sense. This part looks right to me. 

However I think that using `async let` binding to run two tasks in parallel and then collect the result would be simpler way to approach this.

```swift
async let ipv4Result = getIPv4(..)
async let ipv6Result = getIPv6(..)

OutgoingConnectionInfo(
  ipv4: try await ipv4Result, 
  ipv6: try? await ipv6Result
)
```

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 6 of 16 files reviewed, 7 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/View controllers/Tunnel/OutgoingConnectionService.swift` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioYboUEavqbb6AbCTx:-Nj7-T7l59YDc5DvlYsn:boyt2k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/View%20controllers/Tunnel/OutgoingConnectionService.swift#L30)):*

> but task group does it parallely.
Yes but there are a couple of considerations at play here:
- We are only running 2 tasks
- We are doing network calls
- 3 lines of code is vastly easier to understand than 40 with a custom built enum

There's very likely more overhead to try to run 2 tasks in parallel rather than running them one after the other given how simple the calls are (and how fast iPhone processors are nowadays)
And since it's networking code, on a bad connection, it won't make any difference whether the tasks are running in parallel rather than in sequence.

> I think we need to change it for better UX
Doing this will in the best case, not do anything, and in the worst case, make the whole system perform worse.
This is a good example of the [Priority Inversion problem](https://en.wikipedia.org/wiki/Priority_inversion).
However, the swift async runtime should help us in those situations, hopefully. Happy to talk more about why that's a problematic design, but not in this comment.
</blockquote></details>

accepted
___
*[`ios/MullvadVPNTests/OutgoingConnectionProxyTests.swift` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioB08k95LulEe3axoj:-Nj7CBM_0W-2X0HptxYX:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/OutgoingConnectionProxyTests.swift#L34)):*

There are a couple of things we can improve in this test.

- Expectations :
Because we are using inverted expectations here, when this fails, the message produced will be 
`Fulfilled inverted expectation "should not succeed".` 
Which is a double negation. 

In essence, we should keep the same kind messaging so that the error is easy to read and understand.
In this case, it would make sense to read `Fulfilled inverted expectation "Received IPv4".`

- Expectation descriptions :
A message such as "should fail" is not very descriptive. What are we expecting here ? We shouldn't have to read what the code does when a test fail to know _what_ failed.
A much better message would be "Did not receive IPv4". 
That way, if the test fails, this is what we would see "Asynchronous wait failed [...] with unfulfilled expectations: "Did not receive IPv4".
Now we know that we expected to not receive IPV4, and that the expectation was not fullfilled, therefore, we did receive an IPv4. We probably know how to fix this.

- Expectation names :
As important as the message, usually, we try to name the expectation according to what's being expected.
In this case, a good name would be `let noIPv4Expectation = expectation(description: "Did not receive IPv4")` for example.

- Catching the error :

Unfortunately, the `XCTest` API is not well suited for `async` results, otherwise, I would have said to use `XCTAssertThrowsError` which is suited for that.

So I wrote a helper method to help with that
```swift
extension XCTest {
    func XCTAssertThrowsErrorAsync<T: Sendable>(
        _ expression: @autoclosure () async throws -> T,
        _ message: @autoclosure () -> String = "",
        file: StaticString = #filePath,
        line: UInt = #line,
        _ errorHandler: (_ error: Error) -> Void = { _ in }
    ) async {
        do {
            _ = try await expression()
            XCTFail(message(), file: file, line: line)
        } catch {
            errorHandler(error)
        }
    }
}
```

Here's what the rewritten test looks like now 

```swift
func testGetIPv4FailsWhenNotConnected() async throws {
    let noIPv4Expectation = expectation(description: "Did not receive IPv4")
    MockURLProtocol.error = URLError(URLError.notConnectedToInternet)
    MockURLProtocol.requestHandler = nil

    await XCTAssertThrowsErrorAsync(try await outgoingConnectionProxy.getIPV4(retryStrategy: .noRetry)) { error in
        noIPv4Expectation.fulfill()
        XCTAssertEqual((error as? URLError)?.code, .notConnectedToInternet)
    }
    await fulfillment(of: [noIPv4Expectation], timeout: 1)
}

```

I suggest rewriting the other tests in this file with these improvements in mind
</blockquote></details>

Done.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 6 of 16 files reviewed, 7 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 83 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioVJIl8TGpSXbp8nQC:-Nj7HqaO0hhz-6coQuug:b4em6ar) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L83)):*

IMHO `OutgoingConnectionData` should be in its own file and not part of `OutgoingConnectionProxy` 

The protocol signature would look simpler as well
```swift
protocol OutgoingConnectionHandling {
    func getIPV6(retryStrategy: REST.RetryStrategy) async throws -> IPV6ConnectionData
    func getIPV4(retryStrategy: REST.RetryStrategy) async throws -> IPV4ConnectionData
}

```
</blockquote></details>

It's a personal preferences.I'm fan of keeping protocols in individual files.let's know other's opinion.
@pronebird @rablador

buggmagnet (approved):
> Reviewed 1 of 15 files at r1, 1 of 4 files at r3, 7 of 9 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 14 of 16 files reviewed, 5 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

___
*[`ios/MullvadVPNTests/MockURLProtocol.swift` line 12 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioRlCQ7FuU27IrJ2ds:-Nj7U9QC0o6RDNdCFqey:bt45rrh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPNTests/MockURLProtocol.swift#L12)):*

Technically, the only part where we use that `URLSession` object is in `OutgoingConnectionProxy` where we do 
`let (data, response) = try await urlSession.data(from: url)` 

Ideally, we should have extracted that call to a single interface, then we wouldn't need all this mocking of `URLProtocol` and the complicated setup of tests.
That's a bit unfortunate, but it's not the end of the world.

I will create a cleanup janitor task instead, it's not critical.
</blockquote></details>

IOS-383 has been created for that
___
*[`ios/MullvadVPNTests/OutgoingConnectionProxy+Stub.swift` line 12 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nj7UNsN3u8AP96WPpp6:-Nj7UNsN3u8AP96WPpp7:b-oi2ysu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aab41f1b519852fc8a8d9eccb351aec6fe3ea888/ios/MullvadVPNTests/OutgoingConnectionProxy%2BStub.swift#L12)):*
> ```Swift
> 
> struct OutgoingConnectionProxyStub: OutgoingConnectionHandling {
>     var hasError = false
> ```

nit
Technically we don't need this `hasError`
We could have just made error an Optional, and then in the get function do
```swift
guard let error else { return ipVXXX }
throw error
```

But not a big deal

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 14 of 16 files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 15 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nj7fgUjAMBknnrgmP3H:-Nj7fgUkCY7mk5ts0SGc:b-qt10zn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aab41f1b519852fc8a8d9eccb351aec6fe3ea888/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L15)):*
> ```Swift
> 
> protocol OutgoingConnectionHandling {
>     func getIPV6(retryStrategy: REST.RetryStrategy) async throws -> OutgoingConnectionProxy.IPV6ConnectionData
> ```

The protocol should not reference a type from a concrete implementation. 

Consider moving the `IPV*ConnectionData` outside of concrete implementation, perhaps somewhere next to protocol. 

Alternatively use a protocol to describe the return type and conform the concrete implementation to it, i.e:

```swift
protocol OutgoingConnectionnDataProtocol<IPAddressType> {
   associated type IPAddressType: IPAddress

   var ip: IPAddressType { get }
   var isMullvadExit: Bool { get }
}
```

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 15 of 16 files reviewed, 7 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadVPN/View controllers/Tunnel/TunnelViewControllerInteractor.swift` line 46 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nj7hMfT6h2usdirLGpS:-Nj7hMfT6h2usdirLGpT:badwyhk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5f544b7129410edfc84e3c3a18eacfaafececfdb/ios/MullvadVPN/View%20controllers/Tunnel/TunnelViewControllerInteractor.swift#L46)):*
> ```Swift
>                 guard let self else { return }
>                 outgoingConnectionTask?.cancel()
>                 switch tunnelStatus.state {
> ```

Fallthrough and default case are somewhat considered enemies unless used with care, akin to `goto` in the good old days. Since you're only interested in `connected` state, you could rewrite it as a simple `if case .connected = ..` and then you eliminate both `fallthrough` and `default` case this way.

pronebird (commented):
> Reviewed 5 of 9 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 7 unresolved discussions (waiting on @mojganii and @rablador)

rablador (approved):
> Reviewed 2 of 4 files at r3, 8 of 9 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @mojganii, and @pronebird)

___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 83 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NioVJIl8TGpSXbp8nQC:-Nj7qIq51xwwpRHyTOnc:b-o6k8fj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/95e79f8bf177e42bd93a1c9b135baf9287500210/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L83)):*

I don't mind it either way (in separate file or in here). You could however prefer a separate file if you want to mock the actual implementation, for example when you only need to include the protocol definition in the tests bundle.
</blockquote></details>

I think it's fine to combine them if they're tightly coupled/associated with each other.
___
*[`ios/MullvadVPN/View controllers/Tunnel/TunnelViewControllerInteractor.swift` line 46 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nj7hMfT6h2usdirLGpS:-Nj7s7Vz5lTy1lY0xB2H:b32v57) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5f544b7129410edfc84e3c3a18eacfaafececfdb/ios/MullvadVPN/View%20controllers/Tunnel/TunnelViewControllerInteractor.swift#L46)):*

Fallthrough and default case are somewhat considered enemies unless used with care, akin to `goto` in the good old days. Since you're only interested in `connected` state, you could rewrite it as a simple `if case .connected = ..` and then you eliminate both `fallthrough` and `default` case this way.
</blockquote></details>

+1

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 13 of 19 files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadVPN/View controllers/Tunnel/TunnelViewControllerInteractor.swift` line 46 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nj7hMfT6h2usdirLGpS:-Nj97f1XBZQmPywwJxGk:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5f544b7129410edfc84e3c3a18eacfaafececfdb/ios/MullvadVPN/View%20controllers/Tunnel/TunnelViewControllerInteractor.swift#L46)):*

+1
</blockquote></details>

Done.

rablador (approved):
> Reviewed 6 of 6 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @pronebird)

buggmagnet (approved):
> Reviewed 1 of 4 files at r3, 3 of 6 files at r8, 2 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 18 of 19 files reviewed, 5 unresolved discussions (waiting on @mojganii, @pronebird, and @rablador)

pronebird (commented):
> Reviewed 1 of 1 files at r5, 1 of 6 files at r8, 1 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 18 of 19 files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadREST/RESTDefaults.swift` line 17 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NjCTy8X1dkKLgUF7QIN:-NjCTy8Y20TgShOcvDuw:b-2y8ioh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b90ae3cea1892a711912a5ef63cec34c54292d44/ios/MullvadREST/RESTDefaults.swift#L17)):*
> ```Swift
> 
>     /// Exit ipV4  API hostname.
>     public static let ipV4APIHostname = "ipv4.am.i.mullvad.net"
> ```

Nit: Maybe we should call it `exitIPv4Hostname` and `exitIPv6Hostname` instead because IPv4/6 API hostname sounds like the way we reach to `api.mullvad.net`. These are two different services however.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 18 of 19 files reviewed, 4 unresolved discussions (waiting on @buggmagnet, @pronebird, and @rablador)

___
*[`ios/MullvadREST/RESTDefaults.swift` line 17 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NjCTy8X1dkKLgUF7QIN:-NjCcB2O93lZJRVIjQhT:b-896fjx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b90ae3cea1892a711912a5ef63cec34c54292d44/ios/MullvadREST/RESTDefaults.swift#L17)):*

Nit: Maybe we should call it `exitIPv4Hostname` and `exitIPv6Hostname` instead because IPv4/6 API hostname sounds like the way we reach to `api.mullvad.net`. These are two different services however.
</blockquote></details>

Done
___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 15 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nj7fgUjAMBknnrgmP3H:-NjCcaLPDT6z6GxAueek:b-b6ub6z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aab41f1b519852fc8a8d9eccb351aec6fe3ea888/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L15)):*

The protocol should not reference a type from a concrete implementation. 

Consider moving the `IPV*ConnectionData` outside of concrete implementation, perhaps somewhere next to protocol. 

Alternatively use a protocol to describe the return type and conform the concrete implementation to it, i.e:

```swift
protocol OutgoingConnectionnDataProtocol<IPAddressType> {
   associated type IPAddressType: IPAddress

   var ip: IPAddressType { get }
   var isMullvadExit: Bool { get }
}
```
</blockquote></details>

this approach adds more complexity. Apple is using the same format for `Codable`



_Code snippet:_
```Swift
public typealias Codable = Decodable & Encodable
```

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: 18 of 19 files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 15 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nj7fgUjAMBknnrgmP3H:-NjCr4Md8OyfH-jVPzsP:b7uyam6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aab41f1b519852fc8a8d9eccb351aec6fe3ea888/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L15)):*

this approach adds more complexity. Apple is using the same format for `Codable`
</blockquote></details>

Extract the concrete value types 
___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 88 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NjCs_PIEUE8jq2U-aWh:-NjCs_PIEUE8jq2U-aWi:b-bjkod1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b90ae3cea1892a711912a5ef63cec34c54292d44/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L88)):*
> ```Swift
> 
>         } catch {
>             throw error
> ```

do-catch is superficial as it doesn't do anything but rethrow an error. You can eliminate it to streamline the code.

rablador (approved):
> Reviewed 3 of 3 files at r9, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @mojganii)

buggmagnet (approved):
> Reviewed 1 of 9 files at r4, 1 of 6 files at r8, 1 of 3 files at r9, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii)

pronebird (commented):
> Reviewed 1 of 9 files at r4, 1 of 2 files at r10.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii)

pronebird (commented):
> Reviewed 2 of 9 files at r4, 2 of 3 files at r9, 1 of 2 files at r10.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii)

pronebird (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 15 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nj7fgUjAMBknnrgmP3H:-NjHyWIh3hu8hCzZ3khB:b-gs1285) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aab41f1b519852fc8a8d9eccb351aec6fe3ea888/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L15)):*

Extract the concrete value types
</blockquote></details>

To follow up on that. If the `IPv{4|6}ConnectionData` is a requirement of the `OutgoingConnectionHandling` protocol, then it's better to keep the connection data type around the protocol and not embed it into the concrete implementation to avoid cross-dependencies.

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii)

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: all files reviewed, 2 unresolved discussions (waiting on @pronebird)

___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 15 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nj7fgUjAMBknnrgmP3H:-NjHvB85-kYemjCInR6p:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aab41f1b519852fc8a8d9eccb351aec6fe3ea888/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L15)):*

Extract the concrete value types
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift` line 88 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-NjCs_PIEUE8jq2U-aWh:-NjHvdRtE05Mdn8TUmGk:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b90ae3cea1892a711912a5ef63cec34c54292d44/ios/MullvadVPN/GeneralAPIs/OutgoingConnectionProxy.swift#L88)):*

do-catch is superficial as it doesn't do anything but rethrow an error. You can eliminate it to streamline the code.
</blockquote></details>

Done.
___
*[`ios/MullvadVPNTests/OutgoingConnectionProxy+Stub.swift` line 12 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-Nj7UNsN3u8AP96WPpp6:-NjI1lSd1Q15cnSxY8gZ:b-7rl0kd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aab41f1b519852fc8a8d9eccb351aec6fe3ea888/ios/MullvadVPNTests/OutgoingConnectionProxy%2BStub.swift#L12)):*

nit
Technically we don't need this `hasError`
We could have just made error an Optional, and then in the get function do
```swift
guard let error else { return ipVXXX }
throw error
```

But not a big deal
</blockquote></details>

done

pronebird (approved):
:lgtm:

> Reviewed 7 of 7 files at r11.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396#-:-NjIHRDO3-kUKsrli_uT:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

buggmagnet (commented):
> Reviewed 6 of 7 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5396)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5406, opened 2023-11-08
Add support for links in markdown

pronebird:
This PR revamps the markdown parser which previously was limited to supporting bold text and paragraphs, but with this PR it gets extended to support links too. Our previous implementation was very simple and to support links I had to make a more formal parser.

The syntax for links looks as following:

```
[Mullvad VPN](https://mullvad.net/)
```

Typically `NSAttributedString` uses `NSLinkAttributeName` attribute to highlight links within the attributed string, however when such attribute is used with `UILabel`, the color of the link is impossible to change as it's somehow fixed to default blue. `MarkdownStylingOptions` offers a way to customize the attribute that's being used, a custom `.hyperlink` attribute is offered as alternative.

Standard link attribute with custom `UILabel` supporting touch on links:

Screenshot 2023-11-08 at 10 45 39

Custom hyperlink attribute with custom `UILabel` supporting touch on links:

Screenshot 2023-11-08 at 10 09 57


rablador (commented):
I haven't properly looked at the code yet, but I'm wondering if we really need a more complex parser for this. Having a more full-fledged markdown parser could come in handy, sure, but if the only use case right now is the one above can't we just use a UITextView with attributed text and call it a day?

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5406#-:-Nij9_lR-VatzNXKqtNk:b-qptksp)* status: 0 of 24 files reviewed, all discussions resolved

rablador (commented):
Sorry, I meant that by setting a link attribute (`.setAttributes([.link: <some url>], range: <some range>)`) on an attributed string for the link portion, and then assigning it to a `UITextView`, will make the link clickable by default.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5406#-:-NiohmNH3VV6XiL_HMs3:b-tq7tws)* status: 0 of 24 files reviewed, all discussions resolved

PR #5425, opened 2023-11-13
Fix race condition when updating provider and ownership at the same time

Pururun:


Rawa (commented):
Have we checked with daemon to ensure this is not an issue on their end? I mean it feels weird that it does not handle two updates, this becomes more of a workaround than an actual fix.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5425#-:-NjH5AHR6nUMi93tmqep:b-bz0dec)* status: 0 of 4 files reviewed, all discussions resolved

albin-mullvad (commented):
Yes, would be good to double check this :thumbsup: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5425#-:-NjHFiDsB7hV5GS__f6k:b-p567xd)* status: :shipit: complete! all files reviewed, all discussions resolved

Pururun (commented):
I have checked and it is an issue on our end. Since we are getting the relay settings from the daemon and update one value and then send the whole class back to the daemon if you do it in parallell you will get two updates that will overwrite each other, which kind of creates something similar to a race condition, but I should have used a better word in this case. Another way to solve this would be to add some kind of write lock, but then you would make two separate updates that might trigger two separate reconnects.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5425#-:-NjHKuRi0SGNAEKDG6uN:b-vnan0q)* status: :shipit: complete! all files reviewed, all discussions resolved

albin-mullvad (commented):
Is there any benefit of separating the ownership and provider requests? I believe it would be best to skip the separate requests and just use this new combined one since that's more similar to the settings we hand over to the daemon.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5425#-:-NjHd99910laxkhS_-7s:bhtpspy)* status: :shipit: complete! all files reviewed, all discussions resolved

Rawa (commented):
Good to know that it is only on our end, then I'm fine with merging this, it can be something we can look into reworking the daemon interaction. I believe I hit this before when working on the settings so it is good to be aware of, I've also created an issue for it:
https://linear.app/mullvad/issue/DROID-492/settings-race-condition

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5425#-:-NjI-CcO4CzH5wbBS9tj:b-54923i)* status: :shipit: complete! all files reviewed, all discussions resolved

sabercodic (approved):
> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5425)* status: :shipit: complete! all files reviewed, all discussions resolved

Pururun (commented):
@albin-mullvad I guess it is possible to have them together all the time. I can change it to that.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5425#-:-NjIBlju5dWScFVPMuFH:b-c2e2al)* status: :shipit: complete! all files reviewed, all discussions resolved

albin-mullvad (commented):
Sounds good! :thumbsup: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5425#-:-NjIDjHw1aM-9WVJWNs2:bht10at)* status: :shipit: complete! all files reviewed, all discussions resolved

Rawa (approved):
I think that long-term we should be able to support individual settings. It is cumbersome to have to prove (and also fetch) the settings before setting them. It also opens up for potential errors if we change the 2 variables at the same time in separate calls.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5425#-:-NjIFiG07KvzTDaNSPHk:b-xlz0o6)* status: :shipit: complete! all files reviewed, all discussions resolved

albin-mullvad (approved):
:lgtm:

> Reviewed 1 of 3 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5425#-:-NjNravcASd0iznkvUuQ:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5434, opened 2023-11-14
Change - to ~ before beta in rpm version to make dnf treat it as upgrade

faern:
Same fix as #2266 but for RPM instead of DEB.

Basically the Linux package managers compares versions in such a way that `YYYY.NN-betaX` is regarded as a later version than `YYYY.NN`. But in our case it's the other way around. A beta is a pre-release to the corresponding stable. And for both deb and rpm this is denoted with a tilde:

> The tilde symbol ('\~') is used before a version component which must sort earlier than any non-tilde component. It is used for any pre-release versions which wouldn't otherwise sort appropriately.
For example, with upstream releases 0.4.0, 0.4.1, 0.5.0-rc1, 0.5.0-rc2, 0.5.0, the two "release candidates" should use 0.5.0\~rc1 and 0.5.0\~rc2 in the Version: field.

https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_complex_versioning

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5434)

### DIFF
--- a/gui/tasks/distribution.js
+++ b/gui/tasks/distribution.js
@@ -179,7 +179,7 @@ const config = {
     fpm: [
       '--no-depends',
       '--version',
-      getDebVersion(),
+      getLinuxVersion(),
       '--before-install',
       distAssets('linux/before-install.sh'),
       '--before-remove',
@@ -202,6 +202,8 @@ const config = {
 
   rpm: {
     fpm: [
+      '--version',
+      getLinuxVersion(),
       // Prevents RPM from packaging build-id metadata, some of which is the
       // same across all electron-builder applications, which causes package
       // conflicts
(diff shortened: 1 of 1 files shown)


dlon (approved):
Is `$1 == upgrade` true for downgrades in `before-remove.sh`? If not, maybe we should mention in a changelog entry that we temporarily leaked when upgrading from beta versions.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5434#-:-NjD9DYG61m6OvcfI7V3:b-xtmv4m)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (commented):
On Fedora upgrades are caught by the `if [[ "$1" =~ $is_number_re ]] && [ $1 -gt 0 ]; then` step

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5434#-:-NjDCtlfBRdmlu_tqL-4:bah683k)* status: :shipit: complete! all files reviewed, all discussions resolved

faern (commented):
I have now verified that going from "2023.5-beta1" to "2023.5" (considered a downgrade) behaves correctly. `before-remove.sh` still aborts as if it's an upgrade properly.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5434#-:-NjDMqCYBoBJW6-4y0HO:b-fzmfb0)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5438, opened 2023-11-15
Publish RPM repository on build servers

faern:
With this PR the build server now generates an RPM repository and publish it to the development repository servers (and staging if this is a release build, just like we do for deb).

Just like for deb, production repos are manually published later by running:
```
./publish-linux-repositories.sh --production $app_version deb/$app_version rpm/$app_version
```

This also fixes a bug where we previously called `publish_linux_repositories` on all platforms. That's now guarded behind a `if [[ "$(uname -s)" == "Linux" ]]; then`. This was not caught earlier because these build scripts have not been updated on the non-linux machines since it was added a few weeks ago.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5438)

### DIFF
--- a/ci/buildserver-build.sh
+++ b/ci/buildserver-build.sh
@@ -46,9 +46,14 @@ esac
 function publish_linux_repositories {
     local artifact_dir=$1
     local version=$2
-    local deb_repo_dir="$SCRIPT_DIR/deb/$version"
 
+    local deb_repo_dir="$SCRIPT_DIR/deb/$version"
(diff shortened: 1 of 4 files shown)


raksooo (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5438)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5440, opened 2023-11-15
Fix missing account history on new start

Rawa:
Fixes a regression from previous bugfix where we set flowables to be eagerly started, however, this specific flow uses a `onStart` that would execute before the Daemon connection is setup and ready thus not getting a reply. I've now changed it to be lazily loaded.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5440)

### DIFF
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/repository/AccountRepository.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/repository/AccountRepository.kt
@@ -52,7 +52,7 @@ class AccountRepository(
             .events<Event.AccountHistoryEvent>()
             .map { it.history }
             .onStart { fetchAccountHistory() }
-            .stateIn(CoroutineScope(dispatcher), SharingStarted.Eagerly, AccountHistory.Missing)
+            .stateIn(CoroutineScope(dispatcher), SharingStarted.Lazily, AccountHistory.Missing)
 
     private val loginEvents: SharedFlow<LoginResult> =
         messageHandler


albin-mullvad (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5440)* status: all files reviewed, 1 unresolved discussion (waiting on @Rawa)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/repository/AccountRepository.kt` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5440#-NjI9XZ2-CqIzH8PjmNx:-NjI9XZ2-CqIzH8PjmNy:b-7igqu4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/25b1d5b3343d5a4217f306e7fb022de0df6e8fc8/android/app/src/main/kotlin/net/mullvad/mullvadvpn/repository/AccountRepository.kt#L55)):*
> ```Kotlin
>             .map { it.history }
>             .onStart { fetchAccountHistory() }
>             .stateIn(CoroutineScope(dispatcher), SharingStarted.Lazily, AccountHistory.Missing)
> ```

Wasn't there a reason for changing from `Lazily` -> `Eagerly` recently? Also, can we cover this by a test to avoid regression?



_Code quote:_
```Kotlin
Lazily
```

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5440)* status: all files reviewed, 1 unresolved discussion (waiting on @albin-mullvad)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/repository/AccountRepository.kt` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5440#-NjI9XZ2-CqIzH8PjmNx:-NjIAuCn-6WyV2Q6mveF:b-cchb2m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/25b1d5b3343d5a4217f306e7fb022de0df6e8fc8/android/app/src/main/kotlin/net/mullvad/mullvadvpn/repository/AccountRepository.kt#L55)):*

Wasn't there a reason for changing from `Lazily` -> `Eagerly` recently? Also, can we cover this by a test to avoid regression?
</blockquote></details>

There was a change from `WhileSubscribed` -> `Eagerly` before, also on the flow above (`accountExpiryState`). This was done to fix an issue where we would miss a event on accountExpiryState when logging out since no subscriptions was active when logging out and thus we would keep the old state. This fix is correct for `accountExpiryState` but since the `accountHistory` uses` onStart` to communicate with the daemon, and `Eagerly` start would call the daemon before everything has been setup it does not work in this case.

I think it would be rather hard to test this, since it is related to our start up process and combined with the daemon, lets talk about it offline if it is worth testing, and if, how we would approach testing it.

albin-mullvad (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5440#-:-NjIEoHh03c-eL6xkefp:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/repository/AccountRepository.kt` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5440#-NjI9XZ2-CqIzH8PjmNx:-NjIEc014lJ9rwwa9Arn:byj54y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/25b1d5b3343d5a4217f306e7fb022de0df6e8fc8/android/app/src/main/kotlin/net/mullvad/mullvadvpn/repository/AccountRepository.kt#L55)):*

There was a change from `WhileSubscribed` -> `Eagerly` before, also on the flow above (`accountExpiryState`). This was done to fix an issue where we would miss a event on accountExpiryState when logging out since no subscriptions was active when logging out and thus we would keep the old state. This fix is correct for `accountExpiryState` but since the `accountHistory` uses` onStart` to communicate with the daemon, and `Eagerly` start would call the daemon before everything has been setup it does not work in this case.

I think it would be rather hard to test this, since it is related to our start up process and combined with the daemon, lets talk about it offline if it is worth testing, and if, how we would approach testing it.
</blockquote></details>

Sounds good! Let's merge this fix as-is now

PR #5448, opened 2023-11-16
Do not use global dispatch queue in tests

buggmagnet:
This test is flaky when the system is under a lot of pressure due to the usage of the global dispatch queue.
Use a local queue instead to avoid that problem.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5448)


rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5448)* status: :shipit: complete! all files reviewed, all discussions resolved

chremb (commented):
nthc

PR #5486, opened 2023-11-22
Retry fetching out IP DES-118

Serock3:
Add periodic retry for `GetCurrentLocation`.

~~I can not reproduce the original error so I haven't been able to make the new code actually run. We need to see if this fixes the problem, and decide what an appropriate retry interval and max attempts are.~~

I tested by blocking traffic to am.i.mullvad.net, and it seems to work as expected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486)

### DIFF
--- a/mullvad-api/src/lib.rs
+++ b/mullvad-api/src/lib.rs
@@ -352,7 +352,7 @@ impl Runtime {
     }
 
     /// Returns a new request service handle
-    pub async fn rest_handle(&mut self) -> rest::RequestServiceHandle {
+    pub async fn rest_handle(&self) -> rest::RequestServiceHandle {
         self.new_request_service(
             None,
             ApiConnectionMode::Direct.into_repeat(),
--- a/mullvad-daemon/src/geoip.rs
+++ b/mullvad-daemon/src/geoip.rs
@@ -1,11 +1,18 @@
+use std::time::Duration;
+
 use futures::join;
 use mullvad_api::{
     self,
+    availability::ApiAvailabilityHandle,
     rest::{Error, RequestServiceHandle},
 };
 use mullvad_types::location::{AmIMullvad, GeoIpLocation};
 use once_cell::sync::Lazy;
+use talpid_core::future_retry::{retry_future, ConstantInterval};
 use talpid_types::ErrorExt;
+
+/// Retry interval for fetching location
+const RETRY_LOCATION_STRATEGY: ConstantInterval = ConstantInterval::new(Duration::ZERO, Some(3));
 
 // Define the Mullvad connection checking api endpoint.
 //
@@ -36,7 +43,32 @@ static MULLVAD_CONNCHECK_HOST: Lazy<String> = Lazy::new(|| {
     host.to_string()
 });
 
-pub async fn send_location_request(
+/// Fetch the current `GeoIpLocation` with retrys
+pub async fn get_geo_location(
+    rest_service: RequestServiceHandle,
+    use_ipv6: bool,
+    api_handle: ApiAvailabilityHandle,
+) -> Option<GeoIpLocation> {
+    log::debug!("Fetching GeoIpLocation");
+    match retry_future(
+        move || send_location_request(rest_service.clone(), use_ipv6),
+        move |result| match result {
+            Err(error) if error.is_network_error() => !api_handle.get_state().is_offline(),
+            _ => false,
+        },
+        RETRY_LOCATION_STRATEGY,
+    )
+    .await
+    {
+        Ok(loc) => Some(loc),
+        Err(e) => {
+            log::warn!("Unable to fetch GeoIP location: {}", e.display_chain());
+            None
+        }
+    }
+}
+
+async fn send_location_request(
     request_sender: RequestServiceHandle,
     use_ipv6: bool,
 ) -> Result<GeoIpLocation, Error> {
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -27,7 +27,7 @@ mod tunnel;
 pub mod version;
 mod version_check;
 
-use crate::target_state::PersistentTargetState;
+use crate::{geoip::get_geo_location, target_state::PersistentTargetState};
 use device::{AccountEvent, PrivateAccountAndDevice, PrivateDeviceEvent};
 use futures::{
     channel::{mpsc, oneshot},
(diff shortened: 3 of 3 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486)* status: all files reviewed, 2 unresolved discussions (waiting on @Serock3)

___
*[`mullvad-daemon/src/lib.rs` line 97 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486#-NkFgpVt5ghY7Bj6ZMXT:-NkFgpVt5ghY7Bj6ZMXU:bz85efm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/645f870cedcf0ae79ec0314d5f67abbb121adce0/mullvad-daemon/src/lib.rs#L97)):*
> ```XML
> /// Retry interval for fetching location
> const RETRY_LOCATION_STRATEGY: ConstantInterval =
>     ConstantInterval::new(Duration::from_secs(2), Some(3));
> ```

The reason for setting the interval to `0` in other cases is that there's also a timeout interval on the REST requests themselves. It's a bit confusing, but that means that it's already only retrying once every few seconds, if the request fails due to a timeout/network error.

So it should be fine do so the same thing here.
___
*[`mullvad-daemon/src/lib.rs` line 1394 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486#-NkFgu-31u9CXZgWTqku:-NkFgu-31u9CXZgWTqkv:bzdklp1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/645f870cedcf0ae79ec0314d5f67abbb121adce0/mullvad-daemon/src/lib.rs#L1394)):*
> ```XML
>         let use_ipv6 = self.settings.tunnel_options.generic.enable_ipv6;
>         let api_handle = self.api_handle.availability.clone();
>         retry_future(
> ```

We should avoid blocking the daemon from handling further commands here. So it would be better to spawn a new tokio task here and handle the requests there.

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[`mullvad-api/src/lib.rs` line 355 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486#-NkGHDPT07Bv1ZR-uifM:-NkGHDPT07Bv1ZR-uifN:bl0gmjg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/57a1c4ce5535299e3ca659f69d1aa0b6f6adcfd0/mullvad-api/src/lib.rs#L355)):*
> ```XML
> 
>     /// Returns a new request service handle
>     pub async fn rest_handle(&mut self) -> rest::RequestServiceHandle {
> ```

This change is really unrelated, but I snuck it in.
___
*[`mullvad-daemon/src/lib.rs` line 97 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486#-NkFgpVt5ghY7Bj6ZMXT:-NkFybsm2bW09LPRHs7G:b-2mcmwx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/645f870cedcf0ae79ec0314d5f67abbb121adce0/mullvad-daemon/src/lib.rs#L97)):*

The reason for setting the interval to `0` in other cases is that there's also a timeout interval on the REST requests themselves. It's a bit confusing, but that means that it's already only retrying once every few seconds, if the request fails due to a timeout/network error.

So it should be fine do so the same thing here.
</blockquote></details>

Alright. After some testing it feels like the REST API adds ~5 secs on its own, which should do.

This may be another argument for writing how retrys work (in addition to handling the `Reply-after` header`). The delay timer you give should be respected.
___
*[`mullvad-daemon/src/lib.rs` line 1394 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486#-NkFgu-31u9CXZgWTqku:-NkFyHaZ8jKMcb4pzDgX:b-btx1a8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/645f870cedcf0ae79ec0314d5f67abbb121adce0/mullvad-daemon/src/lib.rs#L1394)):*

We should avoid blocking the daemon from handling further commands here. So it would be better to spawn a new tokio task here and handle the requests there.
</blockquote></details>

I knew the nested future that I removed had some purpose! Good catch. I have made the RPC non-blocking again, but hopefully with a slightly more readable implementation. I also tested that the daemon remains responsive (through the CLI) while it is performing the retrys.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486)* status: all files reviewed, 1 unresolved discussion (waiting on @Serock3)

___
*[`mullvad-daemon/src/lib.rs` line 1394 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486#-NkFgu-31u9CXZgWTqku:-NkGh8ZpAPGlcZT2n-MM:bg7igkv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/645f870cedcf0ae79ec0314d5f67abbb121adce0/mullvad-daemon/src/lib.rs#L1394)):*

I knew the nested future that I removed had some purpose! Good catch. I have made the RPC non-blocking again, but hopefully with a slightly more readable implementation. I also tested that the daemon remains responsive (through the CLI) while it is performing the retrys.
</blockquote></details>

:lgtm: 
___
*[`mullvad-daemon/src/lib.rs` line 1407 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486#-NkGhJKz1qxgT_QuB5Le:-NkGhJKz1qxgT_QuB5Lf:bje0qp2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c035bd5b19726e7d08d005fa74b2a2d597c81b63/mullvad-daemon/src/lib.rs#L1407)):*
> ```XML
>         log::debug!("Fetching GeoIpLocation");
>         match retry_future(
>             move || geoip::send_location_request(rest_service.clone(), use_ipv6),
> ```

I wonder if we shouldn't move the retry logic into `geoip`. Just a thought.

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-daemon/src/lib.rs` line 1407 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486#-NkGhJKz1qxgT_QuB5Le:-NkKGLOeEZ2aP4TH9rwM:bzicnis) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c035bd5b19726e7d08d005fa74b2a2d597c81b63/mullvad-daemon/src/lib.rs#L1407)):*

I wonder if we shouldn't move the retry logic into `geoip`. Just a thought.
</blockquote></details>

I agree. I moved the retry logic  to `geoip` and refactored `mullvad-daemon/lib` a bit to make it a bit more concise. The behavior should be the same, but you may have to look at it again.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5486)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5501, opened 2023-11-25
Set default route MTU when using multihop on Linux

dlon:
This prevents fragmentation when multihop is enabled.

Fix DES-483.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5501)


pinkisemils (approved):
> Reviewed 3 of 5 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5501)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

PR #5529, opened 2023-11-30
Replace all instances of Calendar.current with Calendar.autoupdatingCurrent

rablador:
`Calendar.current` does not observe any user made changes to the calendar.

This will not affect caching problems, or calendar values that were cached, but if the user is frequently travelling time zones, this should make a nicer experience for them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5529)

### DIFF
--- a/ios/MullvadVPN/Classes/AccountDataThrottling.swift
+++ b/ios/MullvadVPN/Classes/AccountDataThrottling.swift
@@ -47,7 +47,7 @@ struct AccountDataThrottling {
             break
 
         case .whenCloseToExpiryAndBeyond:
-            guard let closeToExpiry = Calendar.current.date(
+            guard let closeToExpiry = Calendar.autoupdatingCurrent.date(
                 byAdding: .day,
                 value: Self.closeToExpiryDays * -1,
                 to: accountData.expiry
--- a/ios/MullvadVPN/Classes/CustomDateComponentsFormatting.swift
+++ b/ios/MullvadVPN/Classes/CustomDateComponentsFormatting.swift
@@ -22,7 +22,7 @@ extension CustomDateComponentsFormatting {
     static func localizedString(
         from start: Date,
         to end: Date,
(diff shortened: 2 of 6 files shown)


pronebird (commented):
I think the difference between `.current` and `.autoupdatingCurrent` is that the latter listens to `currentLocaleDidChangeNotification` and updates the internal reference to the current calendar. Since we never hold onto the instance of a calendar, the use of `.current` is appropriate and is probably lighter alternative to `autoupdatingCurrent`.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5529#-:-NkUUVi99KLYxI5mwnQa:bdxyfo8)* status: 0 of 6 files reviewed, all discussions resolved

rablador (commented):
You're right, we don't hold any references, so I agree that this change is probably unnecessary. @buggmagnet, as the auther of the issue, any thoughts on this?

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5529#-:-NkUtQQPC81Zcq8cAbP7:b-d0sud4)* status: 0 of 6 files reviewed, all discussions resolved

mojganii (commented):
I don't perceive any advantages in implementing this change.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5529#-:-NkoEUeg4hnPIGwG30Dg:bp8sm0b)* status: 0 of 6 files reviewed, all discussions resolved

rablador (commented):
Testing both with `.autoupdatingCurrent` and `.current` I can see no changes in behaviour. Which means it also doesn't fix the original problem. Closing this PR.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5529#-:-NktIrqE4m6G0Pudcenu:b-f2rvfc)* status: 0 of 6 files reviewed, all discussions resolved

PR #5543, opened 2023-12-01
Update settings schema to V4 to fix a migration bug in obfuscation settings

rablador:
We've introduced breaking changes in the current main that necessitate a new settings schema version. We should add a migration and bump the schema version for this. This can currently be verified by installing a main version over 2023.7.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5543)

### DIFF
--- a/ios/MullvadSettings/WireGuardObfuscationSettings.swift
+++ b/ios/MullvadSettings/WireGuardObfuscationSettings.swift
@@ -20,7 +20,7 @@ public enum WireGuardObfuscationState: Codable {
 /// The port to select when using UDP-over-TCP obfuscation
 ///
 /// `.automatic` means an algorith will decide between using `port80` or `port5001`
-public enum WireGuardObfuscationPort: UInt16, Codable {
+public enum WireGuardObfuscationPort: UInt16, Codable, CaseIterable {
     case automatic = 0
     case port80 = 80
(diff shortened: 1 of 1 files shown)


mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5543)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadSettings/TunnelSettings.swift` line 31 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5543#-Nko93Gk-QB-H3zOWKwW:-Nko93Gk-QB-H3zOWKwX:b9htw81) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71332a465492a19685a647877870b99aaae8d4a7/ios/MullvadSettings/TunnelSettings.swift#L31)):*
> ```Swift
>     case v3 = 3
> 
>     /// V3 format with support for WireGuard obfuscation options with UInt16 port backing values,
> ```

I would suggest to consider more descriptive comments for these sort of  schemas .I'm confused when I read them...
`V4 supports V3 configuration plus back port abilities for WireGuard obfuscation `

Issue #3049, opened 2021-10-19
Mullvad Repository for apt install mullvad-vpn


Danrancan:
## Issue report

Operating system: <ubuntu 20.04>

App version: <Latest -->

### Issue description

I cannot find a known apt repository to add mullvad to my app repos. I would like to automatically install and update mullvad with the command `apt install mullvadvpn` on ubuntu 20.4, and include that package in my updates when using `apt update`. Is there a repository for mullvad that I can add to automatically download and install the package instead of using gdebi for a manual installation?
![Screenshot_20211019_183318](https://user-images.githubusercontent.com/30222551/138004660-25f9fc61-4ddd-4e27-8943-d1960b76b77b.png)

faern:
We want to provide more methods of installation. We want to eventually add this type of repository. But I have no ETA for this. Let's leave the issue open for tracking this "feature".

faern:
It is not a priority for us right now. So in the long run pretty good chances, in the short term: Very low chances sadly.

Danrancan:
> It is not a priority for us right now. So in the long run pretty good chances, in the short term: Very low chances sadly.

How hard is it to actually set up a mullvad.net ppa and just run it off of the current mullvad.net servers? Wouldn't this be a relatively easy task?

Danrancan:
> We want to provide more methods of installation. We want to eventually add this type of repository. But I have no ETA for this. Let's leave the issue open for tracking this "feature".

Glad it is in the long-term plan. This would make things much easier for linux users.

sunjam:
fyi #524 is also an open issue for more general Flatpak support across all distributions.  There is even a free Mullvad key offered to anyone who can make it happen.  Cheers.

Danrancan:
> > How hard is it to actually set up a mullvad.net ppa and just run it off of the current mullvad.net servers? Wouldn't this be a relatively easy task?
> 
> No PPA, just a regular repo. Not everyone uses Ubuntu or derivatives. In my opinion, only a regular repo makes sense. In the case of packages for Debian, you will cover more distributions (including Ubuntu)

Gotcha.

d4f5409d:
> It is not a priority for us right now. So in the long run pretty good chances, in the short term: Very low chances sadly.

I installed the app via a .deb file and it would be useful to me to update Mullvad together with my system instead of every time redownloading the .deb package. Vivaldi does somehow let you update it with the system even if it is installed as a .deb.

faern:
[...]

PR #5550, opened 2023-12-05
Add Ubuntu 23.10 (Mantic) to DEB repository

faern:
Mantic has been out since ~2 months. We really should have it in our DEB repository. Requested here: https://github.com/mullvad/mullvadvpn-app/issues/3049#issuecomment-1840264691

We officially only support the latest non-LTS release, not further back. But I feel like it would cause a lot of issues for users if we immediately on an Ubuntu release unsupport one version when we switch to the new one. So I propose we have a ~6 month overlap. Meaning we technically support the two latest non-LTS releases in the DEB repository.

We could remove "lunar" (release N-1) earlier than when the next version comes out. But then we would need to change this file more often. Easier to just change it once on every Ubuntu release.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5550)

### DIFF
--- a/ci/buildserver-config.sh
+++ b/ci/buildserver-config.sh
@@ -8,8 +8,12 @@ export CODE_SIGNING_KEY_FINGERPRINT="A1198702FC3E0A09A9AE5B75D5A1D4F266DE8DDF"
 
 # Debian codenames we support.
 SUPPORTED_DEB_CODENAMES=("sid" "testing" "bookworm" "bullseye")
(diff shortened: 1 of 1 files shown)


raksooo (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5550)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #5572, opened 2023-12-08
Querying the outgoing connection IP should not affect the in IP

rablador:
If the connection is slow, the In IP address won't be updated until we are done querying the Out addresses. We should:

- Be able to immediately show the In address.
- Not wait for the Out address to be present to show the In address.
- Be able to show IPv4 without IPv6 being present (i.e. `getOutgoingConnectionInfo` should not fail if IPv4 succeeded but IPv6 threw).


buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5572)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadVPN/View controllers/Tunnel/TunnelControlViewModel.swift` line 62 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5572#-Nl8NGwY04gYYafulMON:-Nl8NGwY04gYYafulMOO:b-a0jkiu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fc6bc43879e9bef70473bb1930b3cf13f7d02671/ios/MullvadVPN/View%20controllers/Tunnel/TunnelControlViewModel.swift#L62)):*
> ```Swift
> }
> 
> extension TunnelControlViewModel {
> ```

We don't need to add an extension here, it's in the same file.

buggmagnet (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5572)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5576, opened 2023-12-08
Add missing private IP range

dlon:
Add missing local network sharing range `198.18.0.0/15`. This is used in benchmark tests of network interconnect devices. See [RFC2544](https://datatracker.ietf.org/doc/html/rfc5735).


faern:
Just leaving a comment here on why we closed this: We can't quickly really prove this is a LAN only range, and it's use overall seems very edge-casey. And very low demand for this to be unlocked.

PR #5569, opened 2023-12-08
Use app build container for building test runner

MarkusPettersson98:
Since 96d2e3a9c709b3757a62e8bfe0b559413b125c94 was merged a few days ago, the end-to-end test runner has failed to compile on Linux in Github Actions due to the container image having a too old version of `rustc` to make use of the `is_some_and` feature stabilized in Rust `1.70`.

This PR updates the container image for building the test runner to use the container image for building the Mullvad App.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5569)

PR #5585, opened 2023-12-12
Fix missing test runner binary in CI

MarkusPettersson98:
The new image used for building the test framework (since #5569) overrode `CARGO_TARGET_DIR`, which lead to the `test-runner` artifact being output in an unexpected target directory. This PR sets the expected `CARGO_TARGET_DIR` explicitly.

This PR also drive-by fixes an issue with the macOS test-runner not being able to start the mullvad-daemon after it has been stopped by a test case, causing the test suite to halt.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5585)


MarkusPettersson98:
@dlon OK'd this PR AFK.

PR #5596, opened 2023-12-15
Try to mitigate missing relay list errors

Pururun:
This PR tries to make situations where a user would see an empty relay list rarer by trying to mitigate some potential sources for this error in the app layer.
The two changes are:
1. Change from `SharingStarted.Eagerly` to `SharingStarted.WhileSubscribed()`. This, according to my tests, delays the call to fetch the relay list by around 400 ms. This in my opinion could make things less likely to break since the call is made later and thus the call (or the answer) is less likely be lost.
2. Make a call to get the relay list again when entering the `SelectLocationScreen`. This would mean that if the user still encounters this error it _can_ be resolved when they opening the actual screen. It still would cause them to see an empty list for a second.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5596)

### DIFF
--- a/android/app/src/androidTest/kotlin/net/mullvad/mullvadvpn/compose/screen/SelectLocationScreenTest.kt
+++ b/android/app/src/androidTest/kotlin/net/mullvad/mullvadvpn/compose/screen/SelectLocationScreenTest.kt
@@ -9,6 +9,7 @@ import io.mockk.MockKAnnotations
 import io.mockk.mockk
 import io.mockk.verify
 import net.mullvad.mullvadvpn.compose.setContentWithTheme
+import net.mullvad.mullvadvpn.compose.state.RelayListState
 import net.mullvad.mullvadvpn.compose.state.SelectLocationUiState
 import net.mullvad.mullvadvpn.compose.test.CIRCULAR_PROGRESS_INDICATOR
 import net.mullvad.mullvadvpn.model.Constraint
@@ -54,9 +55,12 @@ class SelectLocationScreenTest {
             setContentWithTheme {
                 SelectLocationScreen(
                     uiState =
-                        SelectLocationUiState.ShowData(
-                            countries = DUMMY_RELAY_COUNTRIES,
-                            selectedRelay = null,
+                        SelectLocationUiState.Data(
+                            relayListState =
+                                RelayListState.RelayList(
+                                    countries = DUMMY_RELAY_COUNTRIES,
+                                    selectedRelay = null
+                                ),
                             selectedOwnership = null,
                             selectedProvidersCount = 0,
                             searchTerm = ""
@@ -91,9 +95,12 @@ class SelectLocationScreenTest {
             setContentWithTheme {
                 SelectLocationScreen(
                     uiState =
-                        SelectLocationUiState.ShowData(
-                            countries = updatedDummyList,
-                            selectedRelay = updatedDummyList[0].cities[0].relays[0],
+                        SelectLocationUiState.Data(
+                            relayListState =
+                                RelayListState.RelayList(
+                                    countries = updatedDummyList,
+                                    selectedRelay = updatedDummyList[0].cities[0].relays[0]
+                                ),
                             selectedOwnership = null,
                             selectedProvidersCount = 0,
                             searchTerm = ""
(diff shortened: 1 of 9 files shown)


Rawa (changes requested):
> ies to make situations where a user would see an empty relay list rarer by trying to mitigate some potential sources for this error in the app layer.  
> The two changes are:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5596#-:-Nlvx8lk2BW_ZuwSCXFv:b-aiib5g)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @Pururun)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/SelectLocationScreen.kt` line 183 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5596#-NlvxlsWA2_Q9WpGBByX:-NlvxlsWA2_Q9WpGBByY:b-3pv5kp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f41a2f26ae6385bc2658acc1d949ac6c383cfa2a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/SelectLocationScreen.kt#L183)):*
> ```Kotlin
>                     }
>                     is SelectLocationUiState.ShowData -> {
>                         if (uiState.countries.isEmpty() && uiState.searchTerm.isNotEmpty()) {
> ```

I'm a bit confused about this one. Is it necessary? I mean if the countries list is Empty we should show the message no matter what right? Is there a case where the search term is empty and we don't have any countries? If so shouldn't it have a completely separate state?

Rawa (commented):
Not sure how this first message got added :point_up:  and now I can't remove it :upside_down_face: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5596#-:-Nlvz8IJDtFpHFTO4vu0:bpgkjrx)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @Pururun)

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5596)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @Rawa)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/SelectLocationScreen.kt` line 183 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5596#-NlvxlsWA2_Q9WpGBByX:-Nlw2iLp7IqZTqPBGkw7:b-h40hrc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f41a2f26ae6385bc2658acc1d949ac6c383cfa2a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/SelectLocationScreen.kt#L183)):*

I'm a bit confused about this one. Is it necessary? I mean if the countries list is Empty we should show the message no matter what right? Is there a case where the search term is empty and we don't have any countries? If so shouldn't it have a completely separate state?
</blockquote></details>

This happens when the list we got from the daemon is empty. This currently shows a text that there is no search results for an empty term. But will move this check to the view model.

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5596)* status: 2 of 9 files reviewed, 1 unresolved discussion (waiting on @Rawa)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/SelectLocationScreen.kt` line 183 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5596#-NlvxlsWA2_Q9WpGBByX:-NlwWJJc2ujA5uj1lz3y:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f41a2f26ae6385bc2658acc1d949ac6c383cfa2a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/SelectLocationScreen.kt#L183)):*

This happens when the list we got from the daemon is empty. This currently shows a text that there is no search results for an empty term. But will move this check to the view model.
</blockquote></details>

Done

albin-mullvad (approved):
:lgtm:

> Reviewed 2 of 5 files at r1, 4 of 7 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5596#-:-NoWF2ZM-VBJ4u3kvooc:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Rawa)

albin-mullvad (approved):
> Dismissed @Rawa from a discussion.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5596)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5603, opened 2023-12-19
Add username password command

buggmagnet:
### PR Description
This PR implements username password authentication for our `Socks5Connection` object.

This authentication method is following
- [RFC-1928](datatracker.ietf.org/doc/html/rfc1928) for the socks protocol
- [RFC-1929](https://datatracker.ietf.org/doc/html/rfc1929) for the username / password authentication system.

In order to use username / password authentication, we need to declare that our client supports that method.
We do so here.

```swift
    /// Start handshake with the socks proxy.
    private func sendHandshake() {
        var handshake = Socks5Handshake()
        if configuration.username != nil && configuration.password != nil {
            handshake.methods.append(.usernamePassword)
        }
```

Once we have exchanged the handshake, and the server has decided to use that handshake, we send a `Socks5Authentication` message.
**This sends the username / password in plain text on the wire**

If the server authenticates us, then we proceed normally by sending a `Socks5ConnectNegotiation`.
The connection is closed in case of error, as per protocol requirements.

### How to test ?
- Install shadowsocks via brew
- `brew install shadowsocks-rust` 
- Follow [the guide](github.com/shadowsocks/shadowsocks-rust#socks5-authentication-configuration) to create a server configuration 
- Run a server locally 
- `ssserver -c path/to/server/conf.json`
- Run a socks5 client locally
- `sslocal -c path/to/server/conf.json # It's the same configuration file as the server uses`
- Configure the iOS client to connect to the socks5 client you have running locally

### What is left to do
We will need to define a way to pass the configuration, username and password to the `URLSessionSocks5Transport` when using a socks5 access method. A TODO comment was left for that to be solved later.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5603)

### DIFF
--- a/ios/MullvadREST/Transport/Socks5/AnyIPEndpoint+Socks5.swift
+++ b/ios/MullvadREST/Transport/Socks5/AnyIPEndpoint+Socks5.swift
@@ -20,4 +20,14 @@ extension AnyIPEndpoint {
             .ipv6(endpoint)
         }
     }
+
+    /// Convert `AnyIPEndpoint` to `NWEndpoint`.
+    var nwEndpoint: NWEndpoint {
+        switch self {
+        case let .ipv4(endpoint):
+            .hostPort(host: .ipv4(endpoint.ip), port: NWEndpoint.Port(integerLiteral: endpoint.port))
+        case let .ipv6(endpoint):
+            .hostPort(host: .ipv6(endpoint.ip), port: NWEndpoint.Port(integerLiteral: endpoint.port))
+        }
+    }
 }
--- a/ios/MullvadREST/Transport/Socks5/Socks5Authentication.swift
+++ b/ios/MullvadREST/Transport/Socks5/Socks5Authentication.swift
@@ -6,9 +6,66 @@
 //
 
 import Foundation
+import Network
 
 /// Authentication methods supported by socks protocol.
 enum Socks5AuthenticationMethod: UInt8 {
     case notRequired = 0x00
     case usernamePassword = 0x02
+}
+
+struct Socks5Authentication {
+    let connection: NWConnection
+    let endpoint: Socks5Endpoint
+    let configuration: Socks5Configuration
+
+    typealias AuthenticationComplete = () -> Void
+    typealias AuthenticationFailure = (Error) -> Void
+
+    func authenticate(onComplete: @escaping AuthenticationComplete, onFailure: @escaping AuthenticationFailure) {
+        guard let username = configuration.username, let password = configuration.password else {
+            onFailure(Socks5Error.invalidUsernameOrPassword)
+            return
+        }
+        let authenticateCommand = Socks5UsernamePasswordCommand(username: username, password: password)
+
+        connection.send(content: authenticateCommand.rawData, completion: .contentProcessed { error in
+            if let error {
+                onFailure(error)
+            } else {
+                readNegotiationReply(onComplete: onComplete, onFailure: onFailure)
+            }
+        })
+    }
+
+    func readNegotiationReply(
+        onComplete: @escaping AuthenticationComplete,
+        onFailure: @escaping AuthenticationFailure
+    ) {
+        let replySize = MemoryLayout<Socks5UsernamePasswordReply>.size
+
+        // Read in one shot, the payload is very small to not care about a reading loop.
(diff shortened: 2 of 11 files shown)


buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5603)* status: all files reviewed, 4 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadREST/Transport/Socks5/Socks5Connection.swift` line 179 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5603#-Nm6XG9698uMD3i4XfkK:-NnIbij24SFdyuOSYzS7:bbfe95k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aacc869fa7da97a969e27f7414109571af5c917c/ios/MullvadREST/Transport/Socks5/Socks5Connection.swift#L179)):*

Had a chat with Matilda regarding the UI and according to spec we should allow authentication even for cases with empty user name and password.
</blockquote></details>

This doesn't prevent authentication without password. As part of the protocol described in the RFC, if we want to use authentication, we need to advertise that we support the scheme.
This is what this bit of code is doing.
___
*[`ios/MullvadREST/Transport/Socks5/Socks5Negotiation.swift` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5603#-NmBBdFOADFTcoAqDhpa:-NnIdMtLBiZUcJzyObzJ:b-rupp2a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aacc869fa7da97a969e27f7414109571af5c917c/ios/MullvadREST/Transport/Socks5/Socks5Negotiation.swift#L17)):*

Nit: Can't we use a `Result` instead?
</blockquote></details>

We could, and we probably should have.
But ironically, it would barely change the readability in this specific case, and I would have to refactor the rest of the code for consistency's sake.
___
*[`ios/MullvadREST/Transport/Socks5/Socks5Negotiation.swift` line 45 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5603#-NmBCQg_-eMXfDBcwiMG:-NnIlPL29XJjGVHOKElq:b-qdei1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aacc869fa7da97a969e27f7414109571af5c917c/ios/MullvadREST/Transport/Socks5/Socks5Negotiation.swift#L45)):*

Nit: Inconsistent use of `maybeError` in `authenticate(onComplete:, onFailure:)` above and `error` here.
</blockquote></details>

Indeed, before we could do `if let error { ...` I would use `maybeError` to indicate that it's an optional, it would be nice to read
`if let error = maybeError { ...` 

Now that's not the case anymore, I'm wondering if I want to keep doing it that way.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5603)* status: all files reviewed, 3 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadREST/Transport/Socks5/Socks5UsernamePasswordCommand.swift` line 79 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5603#-NmBF-eH4-Ix88p5il38:-NnIm24k6GzGs2VE4bP4:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/aacc869fa7da97a969e27f7414109571af5c917c/ios/MullvadREST/Transport/Socks5/Socks5UsernamePasswordCommand.swift#L79)):*

Nit: Inconsistent use of guard. Here in one statement with the status code, but above on its own.
</blockquote></details>

Done

mojganii (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5603#-:-NnJWBZU5XvVNuJ45sVB:bnfp4nl)* status: 7 of 11 files reviewed, 3 unresolved discussions (waiting on @rablador)

rablador (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5603#-:-NnOu-UEA65ri2uS0KJN:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5648, opened 2024-01-03
using access methods in transport layer

mojganii:
This PR enables the app to utilize different API access methods in response to network errors. Each method is associated with a boolean flag indicating whether it's enabled or not. Following a failed network call, the app switches to the next enabled API access method.

For the subsequent method, the app selects the first enabled method from the list, following the one that was just used (and failed).

Upon reaching the end of the list, the app cycles back to the beginning for the next method.

- If no method is enabled, the app defaults to using the 'direct' method.

- In the case where only one method is enabled, the app persists with that method, even if it experienced a recent failure.

Note: Network errors, such as timeouts or dropped connections, trigger this mechanism. It excludes errors that indicate successful communication with the remote machine, such as an HTTP 500 response.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)


buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: 0 of 17 files reviewed, 4 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadREST/Transport/TransportProvider.swift` line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnIz40Q4kAgRep-kYIP:-NnIz40Q4kAgRep-kYIQ:bwk3ubq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadREST/Transport/TransportProvider.swift#L94)):*
> ```Swift
> 
>     /// Returns a randomly selected shadowsocks configuration.
>     private func makeBridgeConfiguration() throws -> ShadowsocksConfiguration {
> ```

We shouldn't be renaming this. 
The term `bridge` has some implication in networks, and a shadowsocks connection is not necessarily used to create bridged networks. 
Please revert this to the previous name.
___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 19 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJ8Q5a0sZQTcb3t_xx:-NnJ8Q5a0sZQTcb3t_xy:b-7u4z38) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadREST/Transport/TransportStrategy.swift#L19)):*
> ```Swift
> 
>         /// Connecting via local Shadowsocks proxy
>         case bridge
> ```

We don't need this case, since we're using shadowsocks to connect via bridges.
We can just reuse `shadowsocks` with a local configuration instead of a user defined one.
___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 45 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJ9HlbB-TrqQ2H4mUE:-NnJ9HlbB-TrqQ2H4mUF:brxyhtf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadREST/Transport/TransportStrategy.swift#L45)):*
> ```Swift
>         self.userDefaults = userDefaults
>         self.dataSource = datasource
>         self.lastReachableConfigurationId = UUID(
> ```

There are 2 problems with this
- ` UUID(uuidString: "")` always returns `nil` 
So if there is no value for `LastReachableConfigurationCacheKey` it will fail anyway, and we don't need to construct an invalid `UUID` for this
- This class shouldn't select itself how to get a `lastReachableConfigurationId`, it should be dependency injected to it.

That guarantees that the class is using the parameters provided to it, and will make it easier to debug if something goes wrong.
___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 86 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJElxkBKof-SE9pdVU:-NnJElxl5g3-nSh0YFGu:biymyqr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadREST/Transport/TransportStrategy.swift#L86)):*
> ```Swift
>     /// Returning `Direct` if there is no enabled configuration
>     private func next() -> PersistentAccessMethod {
>         let direct = dataSource.accessMethods.first(where: { $0.kind == .direct })!
> ```

We do this a lot, it would probably be useful to add that as an extension in `AccessMethodRepositoryDataSource` to query it directly. 

```swift
public extension AccessMethodRepositoryDataSource {
    var directAccess: PersistentAccessMethod {
        accessMethods.first(where: {$0.kind == .direct})!
    }
}

```



_Code quote:_
```Swift
dataSource.accessMethods.first(where: { $0.kind == .direct })!
```

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: 0 of 17 files reviewed, 10 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJuGQvD2XmBeOy5EvF:-NnJuGQvD2XmBeOy5EvG:br575ds) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L10)):*
> ```Swift
> 
> @testable import MullvadREST
> import MullvadSettings
> ```

`import @testable MullvadSettings`
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 33 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJcYDA7ObuVNPGMu8v:-NnJcYDA7ObuVNPGMu8w:blb4hor) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L33)):*
> ```Swift
>     }
> 
>     func testContinuesToUseDirectWhenNoOneIsEnabled() {
> ```

How about `func testDefaultStrategyIsDirectWhenAllMethodsAreDisabled` ?



_Code quote:_
```Swift
testContinuesToUseDirectWhenNoOneIsEnabled
```
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 58 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJmtzU2sPI7XB_6QD1:-NnJmtzU2sPI7XB_6QD2:b-2uhzh2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L58)):*
> ```Swift
>     }
> 
>     func testContinuesToUseBridgeWhenJustOneIsEnabled() {
> ```

This test name is misleading IMHO
What we are testing here is that if one method is enabled, it's always selected. 
The fact that it's bridges or something else doesn't change that. 

I would probably rename this `testSameMethodIsReusedIfEverythingElseIsDisabled`



_Code quote:_
```Swift
testContinuesToUseBridgeWhenJustOneIsEnabled
```
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJp1X95oAtV5vrmGwb:-NnJp1X95oAtV5vrmGwc:b-l57bvv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L83)):*
> ```Swift
>     }
> 
>     func testContinuesToUseDirectWhenItReachesEnd() {
> ```

Same as above, this is misleading. `direct` can be disabled, so what we are really testing here is that we loop over all the enabled methods infinitely when one fails.

To this effect, I propopse that we do the following for this test
- Disable `direct`
- Add 2 more methods , one shadowSocks, one socks5
- We loop over all the strategies available twice (to make sure we did loop) and we assert the current strategy after each failure is different than the previous one. (We don't need to check the strategy itself, since the next test does that)

It's a big test, but it's important that we get this right. The current version of the test doesn't really test much, and wouldn't detect if the selection was broken to only ever return `direct` 

Also we could probably rename this `func testLoopsFromTheStartAfterTryingAllEnabledStrategies`
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 152 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJth5pEvd3AVqT4rfj:-NnJth5pEvd3AVqT4rfk:b-wpi4bo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L152)):*
> ```Swift
>         XCTAssertEqual(
>             transportStrategy.connectionTransport(),
>             .shadowsocks(configuration: ShadowsocksConfiguration(
> ```

We can (and should, for readability) reuse that object across the test by declaring it as class member
Same for the direct and bridges methods, it would make the tests easier to read.
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 171 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJugg64_AuBc7LR13V:-NnJugg64_AuBc7LR13W:b-w6gys3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L171)):*
> ```Swift
>                 .cipher == config2.cipher && config1.password == config2.password
>         case let (.socks5(config1), .socks5(config2)):
>             return config1.address.rawValue == config2.address.rawValue && config1.port == config2.port
> ```

`AnyIPAddress` implements `Equatable` 
No need for the `rawValue` checks here

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: 0 of 17 files reviewed, 10 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadREST/Transport/TransportProvider.swift` line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnIz40Q4kAgRep-kYIP:-NnO6JLn3nj3lAhXk4HI:bim0t0q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadREST/Transport/TransportProvider.swift#L94)):*

We shouldn't be renaming this. 
The term `bridge` has some implication in networks, and a shadowsocks connection is not necessarily used to create bridged networks. 
Please revert this to the previous name.
</blockquote></details>

I took the part are related to loading/creating local shadow-socks proxy  into separate object.while we added custom shadow-socks and socks5, it doesn't seem loading local shadow-socks configuration shall be the responsibility of `TransportProvider`
___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 19 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJ8Q5a0sZQTcb3t_xx:-NnJMijq5KsAW-OPYOEB:b-eyct0s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadREST/Transport/TransportStrategy.swift#L19)):*

We don't need this case, since we're using shadowsocks to connect via bridges.
We can just reuse `shadowsocks` with a local configuration instead of a user defined one.
</blockquote></details>

I put `optional` for the  associated value in case `.shadowsocks` at `TransportStrategy` that `nil` means load shadow-socks proxy with local configuration.
___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 45 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJ9HlbB-TrqQ2H4mUE:-NnOQHPh7hXJOMiAvvE5:b7rvxhg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadREST/Transport/TransportStrategy.swift#L45)):*

There are 2 problems with this
- ` UUID(uuidString: "")` always returns `nil` 
So if there is no value for `LastReachableConfigurationCacheKey` it will fail anyway, and we don't need to construct an invalid `UUID` for this
- This class shouldn't select itself how to get a `lastReachableConfigurationId`, it should be dependency injected to it.

That guarantees that the class is using the parameters provided to it, and will make it easier to debug if something goes wrong.
</blockquote></details>

I couldn't follow the second bullet.do you mean we need to save `PersistentAccessMethod` data type into `UserDefaults` instead of casting the string value of`UUID` ? 
___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 86 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJElxkBKof-SE9pdVU:-NnO7o6-AM9YAnIaWe5d:b-e5s1hv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadREST/Transport/TransportStrategy.swift#L86)):*

We do this a lot, it would probably be useful to add that as an extension in `AccessMethodRepositoryDataSource` to query it directly. 

```swift
public extension AccessMethodRepositoryDataSource {
    var directAccess: PersistentAccessMethod {
        accessMethods.first(where: {$0.kind == .direct})!
    }
}

```
</blockquote></details>

fair enough.I doubt to get it done based on the discussion we had before on how to use extension
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJuGQvD2XmBeOy5EvF:-NnO960C7drrzRQ2eFAU:bb7f2mf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L10)):*

`import @testable MullvadSettings`
</blockquote></details>

done
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 33 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJcYDA7ObuVNPGMu8v:-NnO9NL87fi2wH1QlCY_:blj9ktd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L33)):*

How about `func testDefaultStrategyIsDirectWhenAllMethodsAreDisabled` ?
</blockquote></details>

is not it a bit long? how about this :



_Code snippet:_
```Swift
testUseDefaultStrategyWhenAllIsDisabled
```
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 58 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJmtzU2sPI7XB_6QD1:-NnOC7vK-dR3gW7-QvEv:bt4l175) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L58)):*

This test name is misleading IMHO
What we are testing here is that if one method is enabled, it's always selected. 
The fact that it's bridges or something else doesn't change that. 

I would probably rename this `testSameMethodIsReusedIfEverythingElseIsDisabled`
</blockquote></details>

how about `testReuseSameStrategyWhenEverythingElseIsDisabled`
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJp1X95oAtV5vrmGwb:-NnOCWtg6rmSa7cEJVo2:b32wtd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L83)):*

Same as above, this is misleading. `direct` can be disabled, so what we are really testing here is that we loop over all the enabled methods infinitely when one fails.

To this effect, I propopse that we do the following for this test
- Disable `direct`
- Add 2 more methods , one shadowSocks, one socks5
- We loop over all the strategies available twice (to make sure we did loop) and we assert the current strategy after each failure is different than the previous one. (We don't need to check the strategy itself, since the next test does that)

It's a big test, but it's important that we get this right. The current version of the test doesn't really test much, and wouldn't detect if the selection was broken to only ever return `direct` 

Also we could probably rename this `func testLoopsFromTheStartAfterTryingAllEnabledStrategies`
</blockquote></details>

ok
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 152 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJth5pEvd3AVqT4rfj:-NnOPFLI9EOzBcapE8bN:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L152)):*

We can (and should, for readability) reuse that object across the test by declaring it as class member
Same for the direct and bridges methods, it would make the tests easier to read.
</blockquote></details>

Done.
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 171 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJugg64_AuBc7LR13V:-NnOLazrBssNGl-6Fo1O:buketii) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L171)):*

`AnyIPAddress` implements `Equatable` 
No need for the `rawValue` checks here
</blockquote></details>

Indeed

rablador (changes requested):
> Reviewed 2 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: 2 of 19 files reviewed, 13 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadREST/Transport/TransportProvider.swift` line 21 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nnc7oBY-eTu3RfiphCl:-Nnc7oBY-eTu3RfiphCm:b-9lg6yf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadREST/Transport/TransportProvider.swift#L21)):*
> ```Swift
>     private let parallelRequestsMutex = NSLock()
>     private var relayConstraints = RelayConstraints()
> ```

`relayConstraints` are not used anymore and can be removed.
___
*[`ios/MullvadREST/Transport/TransportProvider.swift` line 90 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NncHP4c15FJxzQp65DD:-NncHP4c15FJxzQp65DE:b-sg7js3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadREST/Transport/TransportProvider.swift#L90)):*
> ```Swift
>             do {
>                 let localConfiguration = try localShadowsocksLoader.configuration
>                 let config = configuration ?? localConfiguration
> ```

Nit: We don't need to try to load a local shadowsocks config unless incoming config is nil.
___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 86 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJElxkBKof-SE9pdVU:-NncNe9TErjv0He4cEZS:bkurv0i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadREST/Transport/TransportStrategy.swift#L86)):*

fair enough.I doubt to get it done based on the discussion we had before on how to use extension
</blockquote></details>

Doesn't really need to be an extension, just a public function.
___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 131 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NncOkVf0o47-AmD7PR1:-NncOkVf0o47-AmD7PR2:b-fajoql) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadREST/Transport/TransportStrategy.swift#L131)):*
> ```Swift
>         let fromData: (Data) -> UUID = { data in
>             let str = String(data: data, encoding: .ascii)
>             return UUID(uuidString: str!)!
> ```

Nit: These forced unwraps feel unsafe since incoming values are not guaranteed to be compatible. Sure, it's a private function, but I'm a little on the fence about it.

rablador (changes requested):
> Reviewed 13 of 17 files at r1, 4 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: all files reviewed, 16 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadRESTTests/AccessMethodRepositoryStub.swift` line 12 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nnd9cww39pwk1Q9kYYB:-Nnd9cww39pwk1Q9kYYC:bg820jc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadRESTTests/AccessMethodRepositoryStub.swift#L12)):*
> ```Swift
> 
> typealias PersistentAccessMethod = MullvadSettings.PersistentAccessMethod
> class AccessMethodRepositoryStub: AccessMethodRepositoryDataSource {
> ```

Nit: Could this be a struct with a single `let accessMethods: [[PersistentAccessMethod]` instead?
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 107 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NndCS878lxQKDZKJRGD:-NndCS878lxQKDZKJRGE:b-n310tc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadRESTTests/TransportStrategyTests.swift#L107)):*
> ```Swift
>             transportStrategy.didFail()
>         }
>         XCTAssertEqual(transportStrategy.connectionTransport(), .direct)
> ```

This should be moved into the loop instead, comparing the last and current values to make sure they're not the same.
___
*[`ios/MullvadSettings/AppStorage.swift` line 38 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NndDvj-5RpK29cgjta4:-NndDvj-5RpK29cgjta5:b-fclvw5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadSettings/AppStorage.swift#L38)):*
> ```Swift
> }
> 
> protocol AnyOptional {
> ```

Nit: Do we need an extension for this when AppStorage is the only place it's used?

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: 8 of 24 files reviewed, 15 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadREST/Transport/TransportProvider.swift` line 21 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nnc7oBY-eTu3RfiphCl:-Nnd4cJSDP0VXnE3pNyX:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadREST/Transport/TransportProvider.swift#L21)):*

`relayConstraints` are not used anymore and can be removed.
</blockquote></details>

Done.
___
*[`ios/MullvadREST/Transport/TransportProvider.swift` line 90 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NncHP4c15FJxzQp65DD:-Nny0YlHBWIKQpD8RN9j:b-onza6g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadREST/Transport/TransportProvider.swift#L90)):*

Nit: We don't need to try to load a local shadowsocks config unless incoming config is nil.
</blockquote></details>

when`TransportStrategy` fails to retrieve the configuration for `bridges` case,It iterates to the next one and preload new configuration for the next try attempt on `bridges` 
___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 86 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJElxkBKof-SE9pdVU:-NnxYmdV3g3XquWdpl_3:b2nizhc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadREST/Transport/TransportStrategy.swift#L86)):*

Doesn't really need to be an extension, just a public function.
</blockquote></details>

the extension has been deleted 
___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 131 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NncOkVf0o47-AmD7PR1:-Nnd5m3926Hns8LbLbrv:bhsi0yh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadREST/Transport/TransportStrategy.swift#L131)):*

Nit: These forced unwraps feel unsafe since incoming values are not guaranteed to be compatible. Sure, it's a private function, but I'm a little on the fence about it.
</blockquote></details>

When we have default value for that and we are sure we're saving UUID it doesn't seem to me unsafe. after discussion Maroc, we came up to save string value od UUID like as before. 
___
*[`ios/MullvadRESTTests/AccessMethodRepositoryStub.swift` line 12 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nnd9cww39pwk1Q9kYYB:-NnxR-7UEGPsWNRzohoO:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadRESTTests/AccessMethodRepositoryStub.swift#L12)):*

Nit: Could this be a struct with a single `let accessMethods: [[PersistentAccessMethod]` instead?
</blockquote></details>

Done
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 107 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NndCS878lxQKDZKJRGD:-NnxVREZCZxtPancBLsk:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadRESTTests/TransportStrategyTests.swift#L107)):*

This should be moved into the loop instead, comparing the last and current values to make sure they're not the same.
</blockquote></details>

Done.
___
*[`ios/MullvadSettings/AppStorage.swift` line 38 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NndDvj-5RpK29cgjta4:-NnxZ0YUBq-yFnVxPvC4:b-pncp5y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f9ff1c9275d413ff218a21b6314031b8eb8bac1/ios/MullvadSettings/AppStorage.swift#L38)):*

Nit: Do we need an extension for this when AppStorage is the only place it's used?
</blockquote></details>

it's merged before,I've just moved it around.we can keep discussing out of the PR

rablador (commented):
> Reviewed 2 of 12 files at r3, 7 of 13 files at r4, 7 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: all files reviewed, 13 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadREST/Transport/AccessMethodIterator.swift` line 19 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nnx_GKR4A6_jfZ6_WrC:-Nnx_GKR4A6_jfZ6_WrD:bj4lfe1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadREST/Transport/AccessMethodIterator.swift#L19)):*
> ```Swift
>     private var cancellables = Set<Combine.AnyCancellable>()
> 
>     /// `UserDefaults` key shared by both processes. Used to cache and synchronize last reachable api access method between them.
> ```

Nit: "Both processes", meaning packet tunnel and main app?
___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 43 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nny2FGs71kotSwH_uFL:-Nny2FGs71kotSwH_uFM:bz06784) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L43)):*
> ```Swift
> 
>     public func preloadNewConfiguration() throws {
>         // because the previous shadowsocks configuration was invalid, therefore generate a new one.
> ```

Nit: Why was it invalid?
___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 55 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nny2UziCqSkW6VrsHqm:-Nny2UziCqSkW6VrsHqn:b9oc83n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L55)):*
> ```Swift
>         } catch {
>             // There is no previous configuration either if this is the first time this code ran
>             let newConfiguration = try create()
> ```

Nit: Reuse `preloadNewConfiguration()` instead?

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: 11 of 25 files reviewed, 15 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadREST/Transport/AccessMethodIterator.swift` line 19 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nnx_GKR4A6_jfZ6_WrC:-NoCbB3Y-QV96CO90sRH:b2nign3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadREST/Transport/AccessMethodIterator.swift#L19)):*

Yes
</blockquote></details>

yep
___
*[`ios/MullvadREST/Transport/AccessMethodIterator.swift` line 61 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nnx_vvy0esKJQx42Bb6:-NnyRI579phg3psWk4dv:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadREST/Transport/AccessMethodIterator.swift#L61)):*

We can afford not having to use abbreviations, let's rename this `firstIndex`
</blockquote></details>

Done.
___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 95 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nnx--m44m1TpT2OQakW:-NnySOkZ8Jme25TzvLEG:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadREST/Transport/TransportStrategy.swift#L95)):*

This should be defined in `SocksConfiguration` instead.
</blockquote></details>

Done.
___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 42 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnxtR6fAXPnSg-awVPT:-NnySlBS4e8lg90XAQzB:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L42)):*

We should rename this `reloadConfiguration`, and we can get rid of the comment, it doens't add information.
</blockquote></details>

Done.
___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 43 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nny2FGs71kotSwH_uFL:-NoCbI7T0fgBoEErUwDs:brt3b8q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L43)):*

Nit: Why was it invalid?
</blockquote></details>

when the relay is not available/reachable it should try with new configuration 
___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 54 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnxyCEQ3jahzYBnL4Bh:-NnyT05lDZ69xQKb79QS:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L54)):*

Half of the comment is missin here, the original comment was
```
// There is no previous configuration either if this is the first time this code ran
// Or because the previous shadowsocks configuration was invalid, therefore generate a new one.
```
</blockquote></details>

Done.
___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 55 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nny2UziCqSkW6VrsHqm:-NnyT0mj-yAeaJfOlTsu:b-hfr3z8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L55)):*

Nit: Reuse `preloadNewConfiguration()` instead?
</blockquote></details>

reload just fetch and updates the cache
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 33 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJcYDA7ObuVNPGMu8v:-NoBR_iP5mC7idRUySPQ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L33)):*

It's better to be explicit about test names since usually they test should be testing one thing only
</blockquote></details>

Done.
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 58 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnJmtzU2sPI7XB_6QD1:-NoCbVL_2bcuksYHMm0a:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c57fca136d63e374251dbf20421bba36a7aab43/ios/MullvadRESTTests/TransportStrategyTests.swift#L58)):*

Sounds good
</blockquote></details>

Done.
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 159 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-Nny-oZG10RYU6L_zwBW:-NoC1c0O9woh67TeGU29:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadRESTTests/TransportStrategyTests.swift#L159)):*

We are missing a couple of tests here
- A test to make sure that when the `bridges` configuration is suggested, but fails to load a shadowsocks configuration, the next available method will be suggested
- A test that verifies we do not enter an endless loop when only `bridges` API is selected, but we cannot load a shadowsocks configuration
- A test for a socks5 configuration (that the configuration returns the correct username / password combo)
</blockquote></details>

Done.
___
*[`ios/MullvadVPNTests/APIAccessMethodsTests.swift` line 29 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnyImRv3x7NtcLcszqe:-NnyX60UEBvWyq8KhcXp:b-xoi83u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadVPNTests/APIAccessMethodsTests.swift#L29)):*

Given what we found, let's add a test to make sure that when settings are wiped, the 2 default access methods are still available
</blockquote></details>

I'll do that but it's out of this PR
___
*[`ios/MullvadRESTTests/AccessMethodIteratorTests.swift` line 14 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NnxySGJ4bBy9tMFksx4:-NoBRBy66653ojMrPzO-:b-hfpv0l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3ae100e968d55781e7a35aea822cc4ad4223a81f/ios/MullvadRESTTests/AccessMethodIteratorTests.swift#L14)):*

It looks like this class is missing tests ?
</blockquote></details>

removed

buggmagnet (changes requested):
> Reviewed 1 of 17 files at r1, 13 of 14 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: 24 of 25 files reviewed, 7 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 82 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NoGUQbx0jb45EiH8GYT:-NoGUQbx0jb45EiH8GYU:b-cjrcy2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b20832046345448864a3ffa4c9b2f621af2b240/ios/MullvadREST/Transport/TransportStrategy.swift#L82)):*
> ```Swift
>         case let .socks5(configuration):
>             switch configuration.authentication {
>             case .noAuthentication:
> ```

nit
It's not a big deal, but ideally the `TransportStrategy` shouldn't care about how the SOCKS5 protocol works. (abstraction leaking)
The `PersistentProxyConfiguration.usernamePassword` should accept optional `username` and `password` instead, since the actual protocol negotiation happens in the SOCKS5 layer itself.

I will create a cleanup task to fix this. 
___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 37 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NoGWqoNBNgyBD1Ilutm:-NoGWqoNBNgyBD1Ilutn:b-pbcolq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b20832046345448864a3ffa4c9b2f621af2b240/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L37)):*
> ```Swift
> 
>     public func reloadConfiguration() throws {
>         // because the previous shadowsocks configuration was invalid, therefore generate a new one.
> ```

This comment doesn't really make sense here anymore, we should remove it.
___
*[`ios/MullvadRESTTests/ShadowsocksLoaderStub.swift` line 31 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NoGXDZBEtySGLks9dAR:-NoGXDZBEtySGLks9dAS:b6gi54b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b20832046345448864a3ffa4c9b2f621af2b240/ios/MullvadRESTTests/ShadowsocksLoaderStub.swift#L31)):*
> ```Swift
>     @discardableResult
>     func load() throws -> ShadowsocksConfiguration {
>         if hasError {
> ```

This implementation has a couple of drawbacks:
- You have to create the stub _with_ an error by default, which doesn't make sense (a stub should not be in an error state by default, it should work immediately)
- You are forced to pass an error in the constructor even if you don't need it

instead of having a variable named `hasError` you can make the `error` optional and do it this way instead (not that we use the compiler to automatically generate the constructor for us here)

```swift

struct ShadowsocksLoaderStub: ShadowsocksLoaderProtocol {
    let configuration: ShadowsocksConfiguration
    let error: Error?

    func reloadConfiguration() throws {
        try load()
    }

    @discardableResult
    func load() throws -> ShadowsocksConfiguration {
        if let error { throw error }
        return configuration
    }
}
```

Also this is the pattern we've used in the tests so far (see `OutgoingConnectionProxyStub`)
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 258 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NoGbuynBJjBPYcwdBNh:-NoGbuynBJjBPYcwdBNi:b-boyzpv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b20832046345448864a3ffa4c9b2f621af2b240/ios/MullvadRESTTests/TransportStrategyTests.swift#L258)):*
> ```Swift
> }
> 
> private enum IOError: Error {
> ```

nit
If we just need an error without any cases, we can use an empty struct instead like so
```swift
private struct IOError: Error { }
```
___
*[`ios/MullvadSettings/AccessMethodRepository.swift` line 96 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NoGhOo24vsJJXY4gmJ3:-NoGhOo380BpqmqM5yE2:bsrk8dd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b20832046345448864a3ffa4c9b2f621af2b240/ios/MullvadSettings/AccessMethodRepository.swift#L96)):*
> ```Swift
>     }
> 
>     public func reloadWithDefaultsAfterDataRemoval() {
> ```

nit
To be honest, I'm not feeling very happy with how we are doing this, but it works for now so we can leave it.

However, I want us to change this in the future. (I don't like the precedent it sets of re-adding things after we wiped all the settings)

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: 24 of 25 files reviewed, 7 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadREST/Transport/TransportStrategy.swift` line 82 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NoGUQbx0jb45EiH8GYT:-NoGi-TtDpH3j8Y0_8Hr:bika6pw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b20832046345448864a3ffa4c9b2f621af2b240/ios/MullvadREST/Transport/TransportStrategy.swift#L82)):*

nit
It's not a big deal, but ideally the `TransportStrategy` shouldn't care about how the SOCKS5 protocol works. (abstraction leaking)
The `PersistentProxyConfiguration.usernamePassword` should accept optional `username` and `password` instead, since the actual protocol negotiation happens in the SOCKS5 layer itself.

I will create a cleanup task to fix this.
</blockquote></details>

The cleanup task is in `IOS-455`

rablador (commented):
> Reviewed 14 of 14 files at r6, 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadRESTTests/ShadowsocksLoaderStub.swift` line 31 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NoGXDZBEtySGLks9dAR:-NoGnn0b6DeFbAeaQGFx:b32v57) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b20832046345448864a3ffa4c9b2f621af2b240/ios/MullvadRESTTests/ShadowsocksLoaderStub.swift#L31)):*

This implementation has a couple of drawbacks:
- You have to create the stub _with_ an error by default, which doesn't make sense (a stub should not be in an error state by default, it should work immediately)
- You are forced to pass an error in the constructor even if you don't need it

instead of having a variable named `hasError` you can make the `error` optional and do it this way instead (not that we use the compiler to automatically generate the constructor for us here)

```swift

struct ShadowsocksLoaderStub: ShadowsocksLoaderProtocol {
    let configuration: ShadowsocksConfiguration
    let error: Error?

    func reloadConfiguration() throws {
        try load()
    }

    @discardableResult
    func load() throws -> ShadowsocksConfiguration {
        if let error { throw error }
        return configuration
    }
}
```

Also this is the pattern we've used in the tests so far (see `OutgoingConnectionProxyStub`)
</blockquote></details>

+1
___
*[`ios/MullvadRESTTests/TransportStrategyTests.swift` line 196 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NoGq9gi77dNl0oUsbGV:-NoGq9gi77dNl0oUsbGW:b-m6v19f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b20832046345448864a3ffa4c9b2f621af2b240/ios/MullvadRESTTests/TransportStrategyTests.swift#L196)):*
> ```Swift
>     }
> 
>     func testUsesSocks5WithAuthenticationWhenItReaches() throws {
> ```

Nit: Reaches, as in makes a successful connection?
___
*[`ios/MullvadSettings/AccessMethodRepository.swift` line 96 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NoGhOo24vsJJXY4gmJ3:-NoGqbiJDZjSpTiznqF_:bjtyrnp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b20832046345448864a3ffa4c9b2f621af2b240/ios/MullvadSettings/AccessMethodRepository.swift#L96)):*

nit
To be honest, I'm not feeling very happy with how we are doing this, but it works for now so we can leave it.

However, I want us to change this in the future. (I don't like the precedent it sets of re-adding things after we wiped all the settings)
</blockquote></details>

If we're simply adding the defaults again I think we can use `add()` instead of adding another function for it.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648)* status: all files reviewed, 6 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadSettings/AccessMethodRepository.swift` line 96 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-NoGhOo24vsJJXY4gmJ3:-NoGsGPaE_trlqZu2iC9:b-gegcr3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b20832046345448864a3ffa4c9b2f621af2b240/ios/MullvadSettings/AccessMethodRepository.swift#L96)):*

If we're simply adding the defaults again I think we can use `add()` instead of adding another function for it.
</blockquote></details>

I think  what Mojgan did here is clearer from the API name instead of just calling `add`, the intent is stronger IMHO.

buggmagnet (approved):
LGTM !

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5648#-:-NoHC0jjCIV05FaCq3nr:bvdsy9m)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii and @rablador)

PR #5668, opened 2024-01-09
Add patch export function to the management interface and CLI

dlon:
Add `mullvad export-settings <FILE>`. As with `import-settings`, `FILE` can be either a path or `-`. If it is `-`, the output is written to standard output.

Closes DES-476.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5668)


mullvad-cli/src/cmds/patch.rs
```diff
@@ -26,10 +26,28 @@ pub async fn handle(source: String) -> Result<()> {
...
+        .export_json_settings()
+        .await
+        .context("Error exporting patch")?;
+
+    tokio::task::spawn_blocking(|| put_blob(dest, json_blob))
+        .await
+        .unwrap()?;
+
+    Ok(())
+}
```
Serock3, line 43: What does `tokio::task::spawn_blocking` do here? Why not just block?

mullvad-cli/src/main.rs
```diff
@@ -138,11 +138,19 @@ enum Cli {
...
         file: String,
     },
+
+    /// Export a JSON patch based on the current settings
+    #[clap(arg_required_else_help = true)]
+    ExportSettings {
+        /// File to write to. If this is "-", write to standard output
+        file: String,
+    },
 }
```
Serock3, line 154: I know this is what the linear issue specified, but why write to a file as default? Is it not easy enough to pipe the output to a file?

mullvad-cli/src/cmds/patch.rs
```diff
@@ -26,10 +26,28 @@ pub async fn handle(source: String) -> Result<()> {
...
+
 fn get_blob(source: String) -> Result<String> {
     match source.as_str() {
         "-" => read_settings_from_stdin().context("Failed to read from stdin"),
-        _ => read_settings_from_file(source).context("Failed to read from path: {source}"),
+        _ => {
+            read_settings_from_file(&source).context(format!("Failed to read from path: {source}"))
```
Serock3, line 49: Good catch!

mullvad-cli/src/cmds/patch.rs
```diff
@@ -99,3 +117,30 @@ fn read_settings_from_reader(mut reader: impl BufRead) -> Result<String> {
...
+        blob,
+    )
+}
+
+fn write_settings_using_writer(mut writer: impl Write, blob: String) -> Result<()> {
+    writer
+        .write_all(blob.as_bytes())
+        .context("Failed to write blob to destination")?;
+    writer.write_all(b"\n").context("Failed to write newline")
+}
```
Serock3, line 146: Why do we need to use a `BufWriter` here? Look at the documentation it says "BufWriter<W> can improve the speed of programs that make small and repeated write calls to the same file or network socket. It does not help when writing very large amounts at once, or writing just one or a few times. It also provides no advantage when writing to a destination that is in memory, like a Vec<u8>".

Writing the settings a single time doesn't seem like the right use case to me.

mullvad-daemon/src/settings/patch.rs
```diff
@@ -417,6 +459,25 @@ fn test_valid_patch_files() {
...
+    relay_override.ipv6_addr_in = Some("::1".parse().unwrap());
+    settings.relay_overrides.push(relay_override);
+
+    let exported = export_settings_inner(&settings).expect("patch export failed");
+
+    let expected = r#"{ "relay_overrides": [ { "hostname": "test", "ipv4_addr_in": "1.2.3.4", "ipv6_addr_in": "::1" } ] }"#;
+    let expected: serde_json::Value = serde_json::from_str(expected).unwrap();
+
+    assert_eq!(exported, expected);
+}
```
Serock3, line 479: This is a bit nitpicky, but if `Settings::default()` ever returns a non-empty setting then this test will fail.

Serock3 (commented):
Here's a thought. Why is the serialization and deserialization of the settings done in the daemon and sent through the gRPC? Why not have a `rcp SetSettings(Setting) returns (Empty)` and do all the (de)serialization in the CLI? Seems unecessary to send unstructured data over gRPC when we can encode the data with the types defined in the same file.

mullvad-cli/src/cmds/patch.rs
dlon: It makes no practical difference here because there are no concurrent tasks, but using blocking I/O on tokio worker threads prevents the runtime from making progress on/completing any other futures. So I avoid it in general.

As for why I'm using `std::io` instead of `tokio::io`, that's just because I read that the latter is not recommended for "interactive uses": https://docs.rs/tokio/latest/tokio/io/struct.Stdin.html

mullvad-cli/src/cmds/patch.rs
dlon: Nope, I don't disagree. :) I guess one `write` syscall is slightly better than two (due to the newline)? But it really does not matter, so I'll remove it.

dlon:
> Here's a thought. Why is the serialization and deserialization of the settings done in the daemon and sent through the gRPC? Why not have a `rcp SetSettings(Setting) returns (Empty)` and do all the (de)serialization in the CLI? Seems unecessary to send unstructured data over gRPC when we can encode the data with the types defined in the same file.

It's in the spec, but one reason is that we don't have to serialize or deserialize data in the various implementations (android, desktop frontend, CLI). That makes it both simpler and guarantees that input is handled identically by all of them.

mullvad-cli/src/cmds/patch.rs
Serock3: Can there really be other concurrent futures in the CLI crate? The text we export is also very small, I doubt blocking for that short duration could cause issues.

dlon: > Can there really be other concurrent futures in the CLI crate?

There are none, but there _could_ be. There probably never will be any.

> The text we export is also very small, I doubt blocking for that short duration could cause issues.

It's not an issue in practice. I just consider it bad in principle to perform (long-running) blocking work on tokio workers.

What are the downsides of doing it this way?

Serock3 (changes requested):
Sure, that might be a good reason for keeping it this way. Do we (or do we plan to) actually export/import settings as json in any other frontend? 

> Reviewed 5 of 9 files at r1, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5668#-:-NnmUg3l5n7BM2DEXkhF:bl6njcf)* status: 8 of 9 files reviewed, 5 unresolved discussions (waiting on @dlon)

___
*[`mullvad-daemon/src/settings/patch.rs` line 479 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5668#gh-1446300253:-NnmXCBp6G1p955EFdtq:b-id8cld) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bf63834c026aeb8160245f884868c36414083ecd/mullvad-daemon/src/settings/patch.rs#L479)):*

True, but is that a problem? CI will catch the error if we decide to add overrides by default. (Also, it probably won't ever happen.)

I guess it wouldn't hurt to clear the vec.
</blockquote></details>

It's probably not a problem, and you can leave it as is if you prefer, but perhaps a slightly more robust test would be to just compare



_Code snippet:_
```XML
assert_eq!(exported.relay_overrides, expected.relay_overrides)
```
___
*[`mullvad-cli/src/cmds/patch.rs` line 146 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5668#gh-1446285697:-NnmVPTt2CmhoGWqQBbt:b593w57) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bf63834c026aeb8160245f884868c36414083ecd/mullvad-cli/src/cmds/patch.rs#L146)):*

Nope, I don't disagree. :) I guess one `write` syscall is slightly better than two (due to the newline)? But it really does not matter, so I'll remove it.
</blockquote></details>

:lgtm:

Serock3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5668)* status: 8 of 9 files reviewed, 4 unresolved discussions (waiting on @dlon)

___
*[`mullvad-cli/src/cmds/patch.rs` line 43 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5668#gh-1446275591:-NnmZKjGAkuvOnsPnmZV:b-69pzm9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bf63834c026aeb8160245f884868c36414083ecd/mullvad-cli/src/cmds/patch.rs#L43)):*

> Can there really be other concurrent futures in the CLI crate?

There are none, but there _could_ be. There probably never will be any.

> The text we export is also very small, I doubt blocking for that short duration could cause issues.

It's not an issue in practice. I just consider it bad in principle to perform (long-running) blocking work on tokio workers.

What are the downsides of doing it this way?
</blockquote></details>

I can see your point, but I think that the right implementation is simply `print!(json_blob)`. We don't have a good reason for making it more complex than that.

mullvad-daemon/src/settings/patch.rs
dlon: Discussed AFK.

mullvad-cli/src/cmds/patch.rs
dlon: Done.

mullvad-cli/src/main.rs
dlon: Discussed AFK.

mullvad-cli/src/cmds/patch.rs
dlon: Discussed AFK.

Serock3 (approved):
> Reviewed 1 of 9 files at r1, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5668)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Serock3 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5668#-:-NnmpubrEgEstpbZp0Bj:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (commented):
> Sure, that might be a good reason for keeping it this way. Do we (or do we plan to) actually export/import settings as json in any other frontend?

Yes! All of them. Desktop, Android and iOS will all soon need to be able to allow the user to paste/load a JSON patch blob. And we don't want them to re-implement the same parsing logic.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5668#-:-NnnLbRL59l5XSudMxmX:b-ln7pez)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5670, opened 2024-01-09
Fix binaries paths in mac build

raksooo:
This PR changes `distribution.js` to look for the Rust binaries in the correct location.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5670)

### DIFF
--- a/gui/tasks/distribution.js
+++ b/gui/tasks/distribution.js
@@ -11,11 +11,14 @@ const shouldNotarize = process.argv.includes('--notarize');
 const universal = process.argv.includes('--universal');
 const release = process.argv.includes('--release');
 
-const targetsIndex = process.argv.indexOf('--targets');
-let targets = null;
+const targets = getOptionValue('--targets');
+const hostTargetTriple = getOptionValue('--host-target-triple');
 
-if (targetsIndex !== -1) {
-  targets = process.argv[targetsIndex + 1];
+function getOptionValue(option) {
+  const optionIndex = process.argv.indexOf(option);
+  if (optionIndex !== -1) {
+    return process.argv[optionIndex + 1];
+  }
 }
 
 const config = {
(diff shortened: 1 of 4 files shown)


gui/tasks/distribution.js
```diff
@@ -284,6 +287,10 @@ function packMac() {
             break;
         }
 
+        process.env.BINARIES_PATH = hostTargetTriple !== process.env.TARGET_TRIPLE
```
Serock3, line 290: Nit: Perhaps a parentheses could make this more readable
```suggestion
        process.env.BINARIES_PATH = (hostTargetTriple !== process.env.TARGET_TRIPLE)
```

gui/tasks/distribution.js
```diff
@@ -11,11 +11,14 @@ const shouldNotarize = process.argv.includes('--notarize');
 const universal = process.argv.includes('--universal');
 const release = process.argv.includes('--release');
 
-const targetsIndex = process.argv.indexOf('--targets');
-let targets = null;
+const targets = getOptionValue('--targets');
+const hostTargetTriple = getOptionValue('--host-target-triple');
```
Serock3, line 15: Maybe "host target" is a bit of a contradiction? My interpretation is that the target is the machine you're building for, so the variable should just be called `hostTriple`.

gui/tasks/distribution.js
Serock3: It may just be me who's unused to ternary operators. If you disagree then just leave it as it was.

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5670)* status: all files reviewed, 2 unresolved discussions (waiting on @raksooo)

gui/tasks/distribution.js
raksooo: Without parentheses is how the linter wants it :)

gui/tasks/distribution.js
raksooo: I think "target triple" is the name of [the format](https://wiki.osdev.org/Target_Triplet). And it is the target triple for the host running the script, even though it might not be a target.

Serock3 (approved):
Built and tested on windows.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5670#-:-NoCnKdNEFOzi3X8Gwr-:bdepu3x)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5678, opened 2024-01-11
Add login UI tests

niklasberglund:
The first basic login tests. Will implement more login-related tests later on to cover more of the login flow(more tests documented in https://linear.app/mullvad/document/tests-09cbefd4a4a6). Using Page Object Pattern(AKA Page Object Model) where code looking for elements, tapping them etc is abstracted away into page classes. So the test cases are very high level and easy to read, making it easy to verify that the tests are testing what they should and it's easier to maintain and create new tests reusing existing page classes.

A new `.xcconfig` file template has been added under `Configurations` so you need to copy the template for example by `cd Configurations ; cp UITests.xcconfig.template UITests.xcconfig` or Xcode might complain. To run the tests the variables on the configuration need to be populated with valid account numbers.

The tests are in the `MullvadVPNUITests` test plan and they can be launched for example from the test navigator in Xcode. For now there's an assumption that the app has been uninstalled before running tests. There will be a solution for this in upcoming PR.
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678)

### DIFF
--- a/ios/Configurations/UITests.xcconfig.template
+++ b/ios/Configurations/UITests.xcconfig.template
@@ -1,0 +1,13 @@
+//
+//  MullvadVPNUITests.xcconfig
+//  MullvadVPN
+//
+//  Created by Niklas Berglund on 2024-01-10.
+//  Copyright © 2024 Mullvad VPN AB. All rights reserved.
+//
+
+// Mullvad accounts used by UI tests
+MULLVAD_NO_TIME_ACCOUNT_NUMBER = 
+MULLVAD_HAS_TIME_ACCOUNT_NUMBER = 
+MULLVAD_FIVE_WIREGUARD_KEYS_ACCOUNT_NUMBER =
+ 
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -566,6 +566,13 @@
 		7AF9BE902A39F26000DBFEDB /* Collection+Sorting.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AF9BE8F2A39F26000DBFEDB /* Collection+Sorting.swift */; };
 		7AF9BE952A40461100DBFEDB /* RelayFilterView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AF9BE942A40461100DBFEDB /* RelayFilterView.swift */; };
 		7AF9BE972A41C71F00DBFEDB /* RelayFilterChipView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AF9BE962A41C71F00DBFEDB /* RelayFilterChipView.swift */; };
+		850201E12B51389500EF8C96 /* BaseUITestCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 850201E02B51389500EF8C96 /* BaseUITestCase.swift */; };
+		852969282B4D9C1F007EAD4C /* AccountTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 852969272B4D9C1F007EAD4C /* AccountTests.swift */; };
+		852969332B4E9232007EAD4C /* Page.swift in Sources */ = {isa = PBXBuildFile; fileRef = 852969322B4E9232007EAD4C /* Page.swift */; };
+		852969352B4E9270007EAD4C /* LoginPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 852969342B4E9270007EAD4C /* LoginPage.swift */; };
+		852969362B4E9724007EAD4C /* AccessbilityIdentifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A0B311D2B303A0D004B12E0 /* AccessbilityIdentifier.swift */; };
+		8529693A2B4F0238007EAD4C /* TermsOfServicePage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 852969392B4F0238007EAD4C /* TermsOfServicePage.swift */; };
+		8529693C2B4F0257007EAD4C /* Alert.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8529693B2B4F0257007EAD4C /* Alert.swift */; };
 		A900E9B82ACC5C2B00C95F67 /* AccountsProxy+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A900E9B72ACC5C2B00C95F67 /* AccountsProxy+Stubs.swift */; };
 		A900E9BA2ACC5D0600C95F67 /* RESTRequestExecutor+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A900E9B92ACC5D0600C95F67 /* RESTRequestExecutor+Stubs.swift */; };
 		A900E9BC2ACC609200C95F67 /* DevicesProxy+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A900E9BB2ACC609200C95F67 /* DevicesProxy+Stubs.swift */; };
@@ -1030,6 +1037,13 @@
 			proxyType = 1;
 			remoteGlobalIDString = 7A88DCCD2A8FABBE00D2FF0E;
 			remoteInfo = Routing;
+		};
+		8529692B2B4D9C1F007EAD4C /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 58CE5E58224146200008646E /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 58CE5E5F224146200008646E;
+			remoteInfo = MullvadVPN;
 		};
 		A91614D22B108F4D00F416EB /* PBXContainerItemProxy */ = {
 			isa = PBXContainerItemProxy;
@@ -1697,6 +1711,16 @@
 		7AF9BE8F2A39F26000DBFEDB /* Collection+Sorting.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "Collection+Sorting.swift"; sourceTree = "<group>"; };
 		7AF9BE942A40461100DBFEDB /* RelayFilterView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RelayFilterView.swift; sourceTree = "<group>"; };
 		7AF9BE962A41C71F00DBFEDB /* RelayFilterChipView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RelayFilterChipView.swift; sourceTree = "<group>"; };
+		850201E02B51389500EF8C96 /* BaseUITestCase.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BaseUITestCase.swift; sourceTree = "<group>"; };
+		852969252B4D9C1F007EAD4C /* MullvadVPNUITests.xctest */ = {isa = PBXFileReference; explicitFileType = wrapper.cfbundle; includeInIndex = 0; path = MullvadVPNUITests.xctest; sourceTree = BUILT_PRODUCTS_DIR; };
+		852969272B4D9C1F007EAD4C /* AccountTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountTests.swift; sourceTree = "<group>"; };
+		852969302B4D9E70007EAD4C /* MullvadVPNUITests.xctestplan */ = {isa = PBXFileReference; lastKnownFileType = text; path = MullvadVPNUITests.xctestplan; sourceTree = "<group>"; };
+		852969322B4E9232007EAD4C /* Page.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Page.swift; sourceTree = "<group>"; };
+		852969342B4E9270007EAD4C /* LoginPage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LoginPage.swift; sourceTree = "<group>"; };
+		852969372B4ED20E007EAD4C /* Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist; path = Info.plist; sourceTree = "<group>"; };
+		852969382B4ED818007EAD4C /* UITests.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; path = UITests.xcconfig; sourceTree = "<group>"; };
+		852969392B4F0238007EAD4C /* TermsOfServicePage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TermsOfServicePage.swift; sourceTree = "<group>"; };
+		8529693B2B4F0257007EAD4C /* Alert.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Alert.swift; sourceTree = "<group>"; };
 		A900E9B72ACC5C2B00C95F67 /* AccountsProxy+Stubs.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "AccountsProxy+Stubs.swift"; sourceTree = "<group>"; };
 		A900E9B92ACC5D0600C95F67 /* RESTRequestExecutor+Stubs.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "RESTRequestExecutor+Stubs.swift"; sourceTree = "<group>"; };
 		A900E9BB2ACC609200C95F67 /* DevicesProxy+Stubs.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "DevicesProxy+Stubs.swift"; sourceTree = "<group>"; };
@@ -1972,6 +1996,13 @@
 			buildActionMask = 2147483647;
 			files = (
 				7A88DCD82A8FABBE00D2FF0E /* Routing.framework in Frameworks */,
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+		};
+		852969222B4D9C1F007EAD4C /* Frameworks */ = {
+			isa = PBXFrameworksBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
@@ -2905,6 +2936,7 @@
 				58C7A4432A863F490060C66F /* PacketTunnelCoreTests */,
 				7A88DCCF2A8FABBE00D2FF0E /* Routing */,
 				7A88DCDD2A8FABBE00D2FF0E /* RoutingTests */,
+				852969262B4D9C1F007EAD4C /* MullvadVPNUITests */,
 				58CE5E61224146200008646E /* Products */,
 				584F991F2902CBDD001F858D /* Frameworks */,
 			);
@@ -2930,6 +2962,7 @@
 				7A88DCCE2A8FABBE00D2FF0E /* Routing.framework */,
 				7A88DCD72A8FABBE00D2FF0E /* RoutingTests.xctest */,
 				58B2FDD32AA71D2A003EB5C6 /* MullvadSettings.framework */,
+				852969252B4D9C1F007EAD4C /* MullvadVPNUITests.xctest */,
 			);
 			name = Products;
 			sourceTree = "<group>";
@@ -3143,6 +3176,7 @@
 				5808273B284888BC006B77A4 /* App.xcconfig */,
 				5808273C284888E5006B77A4 /* PacketTunnel.xcconfig */,
 				58ECD29123F178FD004298B6 /* Screenshots.xcconfig */,
+				852969382B4ED818007EAD4C /* UITests.xcconfig */,
 			);
 			path = Configurations;
 			sourceTree = "<group>";
@@ -3221,6 +3255,7 @@
 		7A83C3FC2A55B39500DFB83A /* TestPlans */ = {
 			isa = PBXGroup;
 			children = (
+				852969302B4D9E70007EAD4C /* MullvadVPNUITests.xctestplan */,
 				7A83C3FE2A55B72E00DFB83A /* MullvadVPNApp.xctestplan */,
 				7A83C4002A55B81A00DFB83A /* MullvadVPNCI.xctestplan */,
 				7A02D4EA2A9CEC7A00C19E31 /* MullvadVPNScreenshots.xctestplan */,
@@ -3260,6 +3295,28 @@
 				7AF9BE8D2A331C7B00DBFEDB /* RelayFilterViewModel.swift */,
 			);
 			path = RelayFilter;
+			sourceTree = "<group>";
+		};
+		852969262B4D9C1F007EAD4C /* MullvadVPNUITests */ = {
+			isa = PBXGroup;
+			children = (
+				852969372B4ED20E007EAD4C /* Info.plist */,
+				852969312B4E9220007EAD4C /* Pages */,
+				852969272B4D9C1F007EAD4C /* AccountTests.swift */,
+				850201E02B51389500EF8C96 /* BaseUITestCase.swift */,
+			);
+			path = MullvadVPNUITests;
+			sourceTree = "<group>";
+		};
+		852969312B4E9220007EAD4C /* Pages */ = {
+			isa = PBXGroup;
+			children = (
+				852969322B4E9232007EAD4C /* Page.swift */,
+				852969342B4E9270007EAD4C /* LoginPage.swift */,
+				852969392B4F0238007EAD4C /* TermsOfServicePage.swift */,
+				8529693B2B4F0257007EAD4C /* Alert.swift */,
+			);
+			path = Pages;
 			sourceTree = "<group>";
 		};
 		A907639F2B2857D50045ADF0 /* Socks5 */ = {
@@ -3913,6 +3970,24 @@
 			productReference = 7A88DCD72A8FABBE00D2FF0E /* RoutingTests.xctest */;
 			productType = "com.apple.product-type.bundle.unit-test";
 		};
+		852969242B4D9C1F007EAD4C /* MullvadVPNUITests */ = {
+			isa = PBXNativeTarget;
+			buildConfigurationList = 8529692F2B4D9C1F007EAD4C /* Build configuration list for PBXNativeTarget "MullvadVPNUITests" */;
+			buildPhases = (
+				852969212B4D9C1F007EAD4C /* Sources */,
+				852969222B4D9C1F007EAD4C /* Frameworks */,
+				852969232B4D9C1F007EAD4C /* Resources */,
+			);
+			buildRules = (
+			);
+			dependencies = (
+				8529692C2B4D9C1F007EAD4C /* PBXTargetDependency */,
+			);
+			name = MullvadVPNUITests;
+			productName = MullvadVPNUITests;
+			productReference = 852969252B4D9C1F007EAD4C /* MullvadVPNUITests.xctest */;
+			productType = "com.apple.product-type.bundle.ui-testing";
+		};
 /* End PBXNativeTarget section */
 
 /* Begin PBXProject section */
@@ -3920,7 +3995,7 @@
 			isa = PBXProject;
 			attributes = {
 				BuildIndependentTargetsInParallel = YES;
-				LastSwiftUpdateCheck = 1430;
+				LastSwiftUpdateCheck = 1510;
 				LastUpgradeCheck = 1430;
 				ORGANIZATIONNAME = "Mullvad VPN AB";
 				TargetAttributes = {
@@ -3995,6 +4070,10 @@
 					7A88DCD62A8FABBE00D2FF0E = {
 						CreatedOnToolsVersion = 14.3.1;
 					};
+					852969242B4D9C1F007EAD4C = {
+						CreatedOnToolsVersion = 15.1;
+						TestTargetID = 58CE5E5F224146200008646E;
+					};
 				};
 			};
 			buildConfigurationList = 58CE5E5B224146200008646E /* Build configuration list for PBXProject "MullvadVPN" */;
@@ -4019,6 +4098,7 @@
 				58CE5E78224146470008646E /* PacketTunnel */,
 				58B0A29F238EE67E00BC001D /* MullvadVPNTests */,
 				58D0C79223F1CE7000FE9BA7 /* MullvadVPNScreenshots */,
+				852969242B4D9C1F007EAD4C /* MullvadVPNUITests */,
 				58D223A4294C8A480029F5F8 /* Operations */,
 				589A455128E094B300565204 /* OperationsTests */,
 				06799ABB28F98E1D00ACD94E /* MullvadREST */,
@@ -4145,6 +4225,13 @@
 			runOnlyForDeploymentPostprocessing = 0;
 		};
 		7A88DCD52A8FABBE00D2FF0E /* Resources */ = {
+			isa = PBXResourcesBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+		};
+		852969232B4D9C1F007EAD4C /* Resources */ = {
 			isa = PBXResourcesBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
@@ -5070,6 +5157,20 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
+		852969212B4D9C1F007EAD4C /* Sources */ = {
+			isa = PBXSourcesBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
+				8529693C2B4F0257007EAD4C /* Alert.swift in Sources */,
+				852969362B4E9724007EAD4C /* AccessbilityIdentifier.swift in Sources */,
+				852969352B4E9270007EAD4C /* LoginPage.swift in Sources */,
+				850201E12B51389500EF8C96 /* BaseUITestCase.swift in Sources */,
+				852969282B4D9C1F007EAD4C /* AccountTests.swift in Sources */,
+				8529693A2B4F0238007EAD4C /* TermsOfServicePage.swift in Sources */,
+				852969332B4E9232007EAD4C /* Page.swift in Sources */,
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+		};
 /* End PBXSourcesBuildPhase section */
 
 /* Begin PBXTargetDependency section */
@@ -5254,6 +5355,11 @@
 			isa = PBXTargetDependency;
 			target = 7A88DCCD2A8FABBE00D2FF0E /* Routing */;
 			targetProxy = 7ABCA5B52A9349F20044A708 /* PBXContainerItemProxy */;
+		};
+		8529692C2B4D9C1F007EAD4C /* PBXTargetDependency */ = {
+			isa = PBXTargetDependency;
+			target = 58CE5E5F224146200008646E /* MullvadVPN */;
+			targetProxy = 8529692B2B4D9C1F007EAD4C /* PBXContainerItemProxy */;
 		};
 		A91614D32B108F4D00F416EB /* PBXTargetDependency */ = {
 			isa = PBXTargetDependency;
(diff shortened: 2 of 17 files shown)


pinkisemils (approved):
> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678)* status: all files reviewed, 10 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 1713 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt5i8a82JOH-MVBmXw:-Nnt5i8a82JOH-MVBmXx:b-o7ls7q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPN.xcodeproj/project.pbxproj#L1713)):*
> ```
> 		7AF9BE942A40461100DBFEDB /* RelayFilterView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RelayFilterView.swift; sourceTree = "<group>"; };
> 		7AF9BE962A41C71F00DBFEDB /* RelayFilterChipView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RelayFilterChipView.swift; sourceTree = "<group>"; };
> 		852969252B4D9C1F007EAD4C /* MullvadVPNUITests.xctest */ = {isa = PBXFileReference; explicitFileType = wrapper.cfbundle; includeInIndex = 0; path = MullvadVPNUITests.xctest; sourceTree = BUILT_PRODUCTS_DIR; };
> ```

We usually sort every file by name alphabetically
If you right click on the `Pages` group, you can select "Sort by name" and it should do it automatically for you
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnsu0IX6mGH-ldEPUIp:-Nnsu0IX6mGH-ldEPUIq:b-nul2ub) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L11)):*
> ```Swift
> import XCTest
> 
> final class AccountTests: XCTestCase {
> ```

We don't need to declare classes as `final`, the compiler can infer it for us, and will automatically do the right thing (and write it on our behalf) most of the time.

It's the same line of reasoning as to why we want to avoid writing `self` most of the time if we can avoid it. 
So that when we see it, we can think "This is intentional, and I need to think carefully about why it was put there in the first place"
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-NnsslfoAg3hWVAkpxkm:-NnsslfoAg3hWVAkpxkn:brkj7sh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L12)):*
> ```Swift
> 
> final class AccountTests: XCTestCase {
>     let noTimeAccountNumber = Bundle(for: AccountTests.self).infoDictionary?["MullvadNoTimeAccountNumber"] as! String
> ```

Given that a lot (if not all ?) of the UI tests will likely start from a clean slate, I think we can generalize a bit how to access those properties. 

Something as simple as the below example should suffice for now 
```swift
class BaseUITest: XCTestCase {
    // swiftlint:disable force_cast
    let noTimeAccountNumber = Bundle(for: AccountTests.self).infoDictionary?["MullvadNoTimeAccountNumber"] as! String
    let hasTimeAccountNumber = Bundle(for: AccountTests.self).infoDictionary?["MullvadHasTimeAccountNumber"] as! String
    let fiveWireGuardKeysAccountNumber = Bundle(for: AccountTests.self)
        .infoDictionary?["MullvadFiveWireGuardKeysAccountNumber"] as! String
    // swiftlint:enable force_cast
}

class AccountTests: BaseUITest {
```
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-NnsvUCGB6kTzVJkmtCe:-NnsvUCGB6kTzVJkmtCf:beah9b8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L25)):*
> ```Swift
>     func testLogin() throws {
>         let app = XCUIApplication()
>         app.launch()
> ```

If we ever want to run the UI automation suite manually, this will most likely fail if we had a previous build running, or a previous UI test running.
It's not super important for now, but for the future, we should uninstall the app in each class of UITests we run before launching the app.
I'm thinking something along the lines of (Or this can probably also be in the `BaseUITest` class)

```swift
class SomeUITest: BaseUITest {

    class override func setUp() {
        uninstallApplication()
        let app = XCUIApplication()
        app.launch()
    }
}

```

Alternatively, we can force a logout at the end of each test, and clear the cached account.
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnsw3lH77UOGjeGuQNC:-Nnsw3lH77UOGjeGuQND:b3mp2xh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L28)):*
> ```Swift
> 
>         TermsOfServicePage(app)
>             .tapAgree()
> ```

This is good, but I think we can do even better.

Here we don't really know what "agree" is, what are we agreeing to ?
Here's a suggestion:

```swift
TermsOfServicePage(app)
    .AgreeToTermsAndConditions()
```
IMHO UI Tests should read like a good book, it's easier to understand the intent if the test tells me what is supposed to be achieved rather than what it is trying to do.
The distinction is subtle, but important.

Here it's just a button to tap, so that's easy. 
But let's imagine that it would have been a flow like "enter password, wait 2 seconds, go accept 2FA token, turn around 3 times in a counter clockwise rotation on your chair"
`tapAgree` wouldn't quite cut it.

This way, the underlying implementation can change, but the intent stays the same. 
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnsz1FJ7nJPSeuMNd3t:-Nnsz1FJ7nJPSeuMNd3u:b-nhpf6x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L53)):*
> ```Swift
>         LoginPage(app)
>             .tapAccountNumberTextField()
>             .enterText("1234123412341234")
> ```

We try to avoid magic numbers as much as possible, also technically, "1234123412341234" _is_ a valid account number.

suggestion
```swift
let incorrectAccountNumber = "0000000000000000"
LoginPage(app)
    .tapAccountNumberTextField()
    .enterText(incorrectAccountNumber)
```
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt3I0XCmmjuk1pDWOW:-Nnt3I0XCmmjuk1pDWOX:b44w6j2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L55)):*
> ```Swift
>             .enterText("1234123412341234")
>             .tapAccountNumberSubmitButton()
>             .verifyFailIconShown()
> ```

nit
I don't think we should verify that an image exists instead to verify that.
Instead we should verify that the text `Login Failed` is shown. 
We should also be mindful that this will need eventually to read localized text.
There is a subtext that is showing why the login failed, which we will want to verify in the future as well.

I'm on the fence on this, on one hand, this is good enough for now.
On the other hand, we will have to change that in the future when we want to validate localization too.
___
*[`ios/MullvadVPNUITests/Pages/LoginPage.swift` line 13 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt5Cmf2BabTMjNbU1H:-Nnt5Cmf2BabTMjNbU1I:b-yn8var) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/Pages/LoginPage.swift#L13)):*
> ```Swift
> 
> class LoginPage: Page {
>     @discardableResult override init(_ app: XCUIApplication) {
> ```

nit
We could probably use the new macro features introduced in swift 5.9 to automatically mark all functions / variables `@discardableResult` in this class :thinking: 
___
*[`ios/MullvadVPNUITests/Pages/Page.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt094I35eAWR81WPNp:-Nnt094I35eAWR81WPNq:b-vkprwo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/Pages/Page.swift#L25)):*
> ```Swift
>     }
> 
>     public func waitForPageToBeShown() {
> ```

It would be nicer to be able to override the timeouts instead of using hardcoded values.
suggestion

```swift
public func waitForPageToBeShown(_ timeout: TimeInterval = 10) {
    if let pageAccessibilityIdentifier = self.pageAccessibilityIdentifier {
        XCTAssertTrue(
            self.app.otherElements[pageAccessibilityIdentifier.rawValue]
                .waitForExistence(timeout: timeout)
        )
    }
}

```
___
*[`ios/MullvadVPNUITests/Pages/Page.swift` line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt058d211isOB1BB-2:-Nnt058d211isOB1BB-3:b-hz3l9m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/Pages/Page.swift#L27)):*
> ```Swift
>     public func waitForPageToBeShown() {
>         if let pageAccessibilityIdentifier = self.pageAccessibilityIdentifier {
>             XCTAssert(self.app.otherElements[pageAccessibilityIdentifier.rawValue].waitForExistence(timeout: 10))
> ```

Let's be explicit in the intent and use `XCTAssertTrue` here

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678)* status: 13 of 17 files reviewed, 10 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 1713 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt5i8a82JOH-MVBmXw:-NntFDqvAFTGQmMKMM-I:b-2129zc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPN.xcodeproj/project.pbxproj#L1713)):*

We usually sort every file by name alphabetically
If you right click on the `Pages` group, you can select "Sort by name" and it should do it automatically for you
</blockquote></details>

Good, done 👍
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnsu0IX6mGH-ldEPUIp:-NnwuFyw0NqMhj6dWrAS:b-kbtfbq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L11)):*

We don't need to declare classes as `final`, the compiler can infer it for us, and will automatically do the right thing (and write it on our behalf) most of the time.

It's the same line of reasoning as to why we want to avoid writing `self` most of the time if we can avoid it. 
So that when we see it, we can think "This is intentional, and I need to think carefully about why it was put there in the first place"
</blockquote></details>

Makes sense. Removed `final`
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-NnsslfoAg3hWVAkpxkm:-NntKyWzCnipXHBvXo-U:bl7j4r8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L12)):*

Given that a lot (if not all ?) of the UI tests will likely start from a clean slate, I think we can generalize a bit how to access those properties. 

Something as simple as the below example should suffice for now 
```swift
class BaseUITest: XCTestCase {
    // swiftlint:disable force_cast
    let noTimeAccountNumber = Bundle(for: AccountTests.self).infoDictionary?["MullvadNoTimeAccountNumber"] as! String
    let hasTimeAccountNumber = Bundle(for: AccountTests.self).infoDictionary?["MullvadHasTimeAccountNumber"] as! String
    let fiveWireGuardKeysAccountNumber = Bundle(for: AccountTests.self)
        .infoDictionary?["MullvadFiveWireGuardKeysAccountNumber"] as! String
    // swiftlint:enable force_cast
}

class AccountTests: BaseUITest {
```
</blockquote></details>

Yes, good idea 👍 Created a base class
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-NnsvUCGB6kTzVJkmtCe:-NntAanUCTQMKwysDL-1:b15ckt8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L25)):*

If we ever want to run the UI automation suite manually, this will most likely fail if we had a previous build running, or a previous UI test running.
It's not super important for now, but for the future, we should uninstall the app in each class of UITests we run before launching the app.
I'm thinking something along the lines of (Or this can probably also be in the `BaseUITest` class)

```swift
class SomeUITest: BaseUITest {

    class override func setUp() {
        uninstallApplication()
        let app = XCUIApplication()
        app.launch()
    }
}

```

Alternatively, we can force a logout at the end of each test, and clear the cached account.
</blockquote></details>

Yes! This is something I haven't implemented a solution for yet and been thinking of as outside of the scope of this first PR, but maybe it shouldn't be. The tests in this PR assume the app isn't installed on the phone. I should've added that to the description. Added it now but let me know if it's something that should be part of this PR even.

Been looking into how to start XCUITest tests with a clear state but it doesn't seem trivial or like there's a very good programmatic solution from what I've found so far. Pondering creating a separate ticket for this. How does that sound?
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnsw3lH77UOGjeGuQNC:-NntLTwE7jycfu8v7Rfv:br7x7l8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L28)):*

This is good, but I think we can do even better.

Here we don't really know what "agree" is, what are we agreeing to ?
Here's a suggestion:

```swift
TermsOfServicePage(app)
    .AgreeToTermsAndConditions()
```
IMHO UI Tests should read like a good book, it's easier to understand the intent if the test tells me what is supposed to be achieved rather than what it is trying to do.
The distinction is subtle, but important.

Here it's just a button to tap, so that's easy. 
But let's imagine that it would have been a flow like "enter password, wait 2 seconds, go accept 2FA token, turn around 3 times in a counter clockwise rotation on your chair"
`tapAgree` wouldn't quite cut it.

This way, the underlying implementation can change, but the intent stays the same.
</blockquote></details>

The `tapAgree` method isn't following the convention other methods follow and should be named `tapAgreeButton`. Does that make it more clear? 

All those two lines do is verify `TermsOfServiceContentView` shows and then taps the accept button to accept the terms.
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnsz1FJ7nJPSeuMNd3t:-NntBBzC112-Hy14MKzA:bnpo2c3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L53)):*

We try to avoid magic numbers as much as possible, also technically, "1234123412341234" _is_ a valid account number.

suggestion
```swift
let incorrectAccountNumber = "0000000000000000"
LoginPage(app)
    .tapAccountNumberTextField()
    .enterText(incorrectAccountNumber)
```
</blockquote></details>

True. Is account number`0000000000000000` guaranteed to be invalid so using that instead solves the issue or should it also be broken out of the code to the configuration file? 
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt3I0XCmmjuk1pDWOW:-NntBjBl5T00le0ATxa5:b-zi2f0j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L55)):*

nit
I don't think we should verify that an image exists instead to verify that.
Instead we should verify that the text `Login Failed` is shown. 
We should also be mindful that this will need eventually to read localized text.
There is a subtext that is showing why the login failed, which we will want to verify in the future as well.

I'm on the fence on this, on one hand, this is good enough for now.
On the other hand, we will have to change that in the future when we want to validate localization too.
</blockquote></details>

At first I also wanted to look for the label, but why I finally went with checking the image name instead of checking the label is because it is a status label that is reused so it's the same label object for all login states. And trying to avoid checking for specific text strings because it's so easy to break checks against specific strings. There is also an additional check on the line underneath, `waitForPageToBeShown()` which verifies that we're still on the login page so the user haven't been sent to the tunnel control page you get to after being logged in.



_Code snippet:_
```Swift
private extension LoginState {
    var localizedTitle: String {
        switch self {
        case .default:
            return NSLocalizedString(
                "HEADING_TITLE_DEFAULT",
                tableName: "Login",
                value: "Login",
                comment: ""
            )

        case .authenticating:
            return NSLocalizedString(
                "HEADING_TITLE_AUTHENTICATING",
                tableName: "Login",
                value: "Logging in...",
                comment: ""
            )

        case .failure:
            return NSLocalizedString(
                "HEADING_TITLE_FAILURE",
                tableName: "Login",
                value: "Login failed",
                comment: ""
            )

        case .success:
            return NSLocalizedString(
                "HEADING_TITLE_SUCCESS",
                tableName: "Login",
                value: "Logged in",
                comment: ""
            )
        }
    }
    ...
```
___
*[`ios/MullvadVPNUITests/Pages/LoginPage.swift` line 13 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt5Cmf2BabTMjNbU1H:-NntDw4CAEF--TMNBJHU:b-w9m38s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/Pages/LoginPage.swift#L13)):*

nit
We could probably use the new macro features introduced in swift 5.9 to automatically mark all functions / variables `@discardableResult` in this class :thinking:
</blockquote></details>

That would be neat! The app target is using Swift 5.0. I'm not sure how much 5.0 and 5.9 differs, would it be okay to use different Swift versions for these targets?
___
*[`ios/MullvadVPNUITests/Pages/Page.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt094I35eAWR81WPNp:-NntE3uP8VyEwNoC73FZ:b-yd213c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/Pages/Page.swift#L25)):*

It would be nicer to be able to override the timeouts instead of using hardcoded values.
suggestion

```swift
public func waitForPageToBeShown(_ timeout: TimeInterval = 10) {
    if let pageAccessibilityIdentifier = self.pageAccessibilityIdentifier {
        XCTAssertTrue(
            self.app.otherElements[pageAccessibilityIdentifier.rawValue]
                .waitForExistence(timeout: timeout)
        )
    }
}

```
</blockquote></details>

Should the time it takes to navigate to a page differ though? For example 10 seconds is a super high value actually not meant to catch performance issues really but a timeout that is reached when there's an issue with the test usually.
___
*[`ios/MullvadVPNUITests/Pages/Page.swift` line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt058d211isOB1BB-2:-NntF0uh2Oq2CX9R-ZT7:b7weumd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/Pages/Page.swift#L27)):*

Let's be explicit in the intent and use `XCTAssertTrue` here
</blockquote></details>

Good idea, changed 👍

mojganii (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678)* status: 13 of 17 files reviewed, 10 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-NnsvUCGB6kTzVJkmtCe:-NoBMd0U4Z_eA02bbvna:b-y6e2er) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L25)):*

Yes! This is something I haven't implemented a solution for yet and been thinking of as outside of the scope of this first PR, but maybe it shouldn't be. The tests in this PR assume the app isn't installed on the phone. I should've added that to the description. Added it now but let me know if it's something that should be part of this PR even.

Been looking into how to start XCUITest tests with a clear state but it doesn't seem trivial or like there's a very good programmatic solution from what I've found so far. Pondering creating a separate ticket for this. How does that sound?
</blockquote></details>

I agree with @buggmagnet

buggmagnet (changes requested):
> Reviewed 12 of 16 files at r1, 3 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678)* status: all files reviewed, 7 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-NnsvUCGB6kTzVJkmtCe:-NoBHyx64DhuMBGDC8vW:b-cd0ilm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L25)):*

Yes! This is something I haven't implemented a solution for yet and been thinking of as outside of the scope of this first PR, but maybe it shouldn't be. The tests in this PR assume the app isn't installed on the phone. I should've added that to the description. Added it now but let me know if it's something that should be part of this PR even.

Been looking into how to start XCUITest tests with a clear state but it doesn't seem trivial or like there's a very good programmatic solution from what I've found so far. Pondering creating a separate ticket for this. How does that sound?
</blockquote></details>

Sounds good, I'll create a separate ticket for that :thumbsup: 
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnsw3lH77UOGjeGuQNC:-NoBKzPR9pzc7ZmYO0Op:b-swy38d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L28)):*

The `tapAgree` method isn't following the convention other methods follow and should be named `tapAgreeButton`. Does that make it more clear? 

All those two lines do is verify `TermsOfServiceContentView` shows and then taps the accept button to accept the terms.
</blockquote></details>

nit:
Turning this into a nitpick cause it's not super important at this stage.

Yes it would make it more clear, but this isn't exactly what I was trying to say.
Please take with a grain of salt what I'm proposing here because it might be early optimization, but what I'm suggesting is that the test looks like this 

```swift
/// Validates that the user can login with a valid account number
func testLoginWithValidAccountNumber() throws {
    let app = XCUIApplication()
    app.launch()

    TermsOfServicePage(app)
        .agreeToTermsAndServiceConditions()

    LoginPage(app)
        .verifyLoginWith(
            app,
            validAccountNumber: self.noTimeAccountNumber
        )
}

```

The idea is that the test doesn't require looking at the functions to see what it does, and what it tries to validate, it's very self explanatory.
There's a TOS page, which will need to agree to conditions, then there's a login page where you enter an account number.

IMHO we shouldn't need to read more than that to see what the test is about.

That being said, it's okay if you prefer to just rename `tapAgree` to `tapAgreeButton` and move on.
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnsz1FJ7nJPSeuMNd3t:-NoBNU5QExZoe-8Rtw5M:b-wuyy4t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L53)):*

True. Is account number`0000000000000000` guaranteed to be invalid so using that instead solves the issue or should it also be broken out of the code to the configuration file?
</blockquote></details>

Let's assume it is for now, I will check around and come back to you with a better answer than "I'm almost sure it is, but not quite"
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt3I0XCmmjuk1pDWOW:-NoBNeL11a7zfG987Lx5:ba9xot9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L55)):*

At first I also wanted to look for the label, but why I finally went with checking the image name instead of checking the label is because it is a status label that is reused so it's the same label object for all login states. And trying to avoid checking for specific text strings because it's so easy to break checks against specific strings. There is also an additional check on the line underneath, `waitForPageToBeShown()` which verifies that we're still on the login page so the user haven't been sent to the tunnel control page you get to after being logged in.
</blockquote></details>

Fair enough, let's keep it like this for now then :thumbsup: 
___
*[`ios/MullvadVPNUITests/BaseTestCase.swift` line 14 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-NoBNkz94pV1HHX_h55L:-NoBNkz94pV1HHX_h55M:b-40wxb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f952d65d99522f8e378da9c324fc5488356fca43/ios/MullvadVPNUITests/BaseTestCase.swift#L14)):*
> ```Swift
> class BaseTestCase: XCTestCase {
>     // swiftlint:disable force_cast
>     let noTimeAccountNumber = Bundle(for: AccountTests.self).infoDictionary?["MullvadNoTimeAccountNumber"] as! String
> ```

I know I made the original suggestion _but_ I forgot one important detail.  We should rename this `BaseUITestCase` to avoid clashing with Unit Tests (if we ever get to have a `BaseUnitTestCase`), also the bundle class refered to should be this one, it's likely to out-live `AccountTests` if we ever get rid of them.

```swift
class BaseUITestCase: XCTestCase {
    // swiftlint:disable force_cast
    let noTimeAccountNumber = Bundle(for: BaseUITestCase.self).infoDictionary?["MullvadNoTimeAccountNumber"] as! String
    let hasTimeAccountNumber = Bundle(for: BaseUITestCase.self).infoDictionary?["MullvadHasTimeAccountNumber"] as! String
    let fiveWireGuardKeysAccountNumber = Bundle(for: BaseUITestCase.self)
        .infoDictionary?["MullvadFiveWireGuardKeysAccountNumber"] as! String
    // swiftlint:enable force_cast
}
```
___
*[`ios/MullvadVPNUITests/Pages/LoginPage.swift` line 13 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt5Cmf2BabTMjNbU1H:-NoBOWjV68G60YttPlyC:bpglu89) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/Pages/LoginPage.swift#L13)):*

That would be neat! The app target is using Swift 5.0. I'm not sure how much 5.0 and 5.9 differs, would it be okay to use different Swift versions for these targets?
</blockquote></details>

Xcode 15 ships with the swift 5.9 compiler. Which means we can already use the features from that version.
If you run the following command, you will likely get the same input (or at least, the same version)
```
[~/Code/mullvadvpn-app/ios (test-app-login-ios-428)]% xcrun swift --version
swift-driver version: 1.87.3 Apple Swift version 5.9.2 (swiftlang-5.9.2.2.56 clang-1500.1.0.2.5)
Target: arm64-apple-macosx13.0
```

What you see (probably) in "Swift Language Version" means little nowadays since Swift reached ABI stability in version 5.0. (It was more important when swift didn't have ABI stability, and when the syntax was still evolving a lot)

What really matters is which tool we use, and that means we choose to tie which swift version we can use, with which Xcode version we use (I'm taking a shortcut here).

Let's not bother too much with this for now, we can take time to do this later if it proves to be too much boilerplate
___
*[`ios/MullvadVPNUITests/Pages/Page.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt094I35eAWR81WPNp:-NoBSslwD0mVXbDLBngc:b-9xi5se) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/Pages/Page.swift#L25)):*
> Should the time it takes to navigate to a page differ though?
Most likely not. But since we're trying to avoid magic values in general, it's better to have this at least as a constant defined somewhere.
What do you think ?

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678)* status: 0 of 73 files reviewed, 7 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 28 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnsw3lH77UOGjeGuQNC:-NoBZhv162dOCWVGwUGq:b65egsq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L28)):*
Ah, I see your point. In the past I've made the same suggestion as you but I was convinced to not go with this higher abstraction. There's some big benefits of automated test cases reading the same as manual test cases which list all the steps, such as making it easier to maintain test cases and be sure that all functionality is covered. With a higher abstraction level the steps are less visible, you'd have to jump around in the code to figure out exactly what user actions the tests are doing. And some flows such as login will be slightly different in different test cases also - for instance there are two submit buttons. 

> That being said, it's okay if you prefer to just rename `tapAgree` to `tapAgreeButton` and move on.

Renamed it to `tapAgreeButton`. If I haven't convinced you yet I'll try to do so with time! 😊
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnsz1FJ7nJPSeuMNd3t:-NoBaQYY9EggV109ad9I:b-ka6gyi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L53)):*

Let's assume it is for now, I will check around and come back to you with a better answer than "I'm almost sure it is, but not quite"
</blockquote></details>

Maybe it doesn't answer our question but for the Android end to end tests there's a configurable `invalid_test_account_token` https://github.com/mullvad/mullvadvpn-app/tree/main/android/test/e2e

Have changed the invalid account number to `number`0000000000000000``
___
*[`ios/MullvadVPNUITests/BaseTestCase.swift` line 14 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-NoBNkz94pV1HHX_h55L:-NoBgWOmCOOe1-livacT:bxpv4l0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f952d65d99522f8e378da9c324fc5488356fca43/ios/MullvadVPNUITests/BaseTestCase.swift#L14)):*

I know I made the original suggestion _but_ I forgot one important detail.  We should rename this `BaseUITestCase` to avoid clashing with Unit Tests (if we ever get to have a `BaseUnitTestCase`), also the bundle class refered to should be this one, it's likely to out-live `AccountTests` if we ever get rid of them.

```swift
class BaseUITestCase: XCTestCase {
    // swiftlint:disable force_cast
    let noTimeAccountNumber = Bundle(for: BaseUITestCase.self).infoDictionary?["MullvadNoTimeAccountNumber"] as! String
    let hasTimeAccountNumber = Bundle(for: BaseUITestCase.self).infoDictionary?["MullvadHasTimeAccountNumber"] as! String
    let fiveWireGuardKeysAccountNumber = Bundle(for: BaseUITestCase.self)
        .infoDictionary?["MullvadFiveWireGuardKeysAccountNumber"] as! String
    // swiftlint:enable force_cast
}
```
</blockquote></details>

Good point! Renamed it to `BaseUITestCase`

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678)* status: 0 of 73 files reviewed, 7 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadVPNUITests/Pages/LoginPage.swift` line 13 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt5Cmf2BabTMjNbU1H:-NoCvzbMAmxxOxLFRUil:b-5d9yw1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/Pages/LoginPage.swift#L13)):*

Xcode 15 ships with the swift 5.9 compiler. Which means we can already use the features from that version.
If you run the following command, you will likely get the same input (or at least, the same version)
```
[~/Code/mullvadvpn-app/ios (test-app-login-ios-428)]% xcrun swift --version
swift-driver version: 1.87.3 Apple Swift version 5.9.2 (swiftlang-5.9.2.2.56 clang-1500.1.0.2.5)
Target: arm64-apple-macosx13.0
```

What you see (probably) in "Swift Language Version" means little nowadays since Swift reached ABI stability in version 5.0. (It was more important when swift didn't have ABI stability, and when the syntax was still evolving a lot)

What really matters is which tool we use, and that means we choose to tie which swift version we can use, with which Xcode version we use (I'm taking a shortcut here).

Let's not bother too much with this for now, we can take time to do this later if it proves to be too much boilerplate
</blockquote></details>

Sounds good, let's see later then
___
*[`ios/MullvadVPNUITests/Pages/Page.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnt094I35eAWR81WPNp:-NoCw3sZ34Y6CvxDAuou:b-ai4pe1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/Pages/Page.swift#L25)):*

> Should the time it takes to navigate to a page differ though?
Most likely not. But since we're trying to avoid magic values in general, it's better to have this at least as a constant defined somewhere.
What do you think ?
</blockquote></details>

Agree. I created a constant for the timeout value.

buggmagnet (approved):
> Reviewed 3 of 17 files at r11, 2 of 59 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678)* status: 5 of 73 files reviewed, 2 unresolved discussions (waiting on @mojganii, @niklasberglund, and @pinkisemils)

___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-NnsvUCGB6kTzVJkmtCe:-NoGtnJo-gve7tjh1trh:bogqn4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L25)):*

Sounds good, I'll create a separate ticket for that :thumbsup:
</blockquote></details>

@niklasberglund I created `IOS-456` in regards to this
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5678#-Nnsz1FJ7nJPSeuMNd3t:-NoGu3919A3z5EatOliG:b-n0svj1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2ade8a1fa60248f7d6aa3a33915153faeafc978/ios/MullvadVPNUITests/AccountTests.swift#L53)):*

Maybe it doesn't answer our question but for the Android end to end tests there's a configurable `invalid_test_account_token` https://github.com/mullvad/mullvadvpn-app/tree/main/android/test/e2e

Have changed the invalid account number to `number`0000000000000000``
</blockquote></details>

nice !

PR #5687, opened 2024-01-15
Fix app getting stuck in error state on macOS

dlon:
Previously, the app would get stuck in the error state when a default route was removed while connecting. Specifically, this occurred if the default was missing while adding tunnel routes, but before the offline monitor had had time to catch it. That caused adding routes to fail due to `ENETUNREACH`.

This PR fixes it by simply retrying/reconnecting when adding routes fails due to an unreachable destination.

Closes DES-496.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5687)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ Line wrap the file at 100 chars.                                              Th
 #### macOS
 - Fix default route not being restored when disconnecting when the gateway was a link-local IPv6
   address.
+- Fix app sometimes getting stuck in error state when the connection is unstable. This occurred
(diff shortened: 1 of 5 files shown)


MarkusPettersson98 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5687#-:-NoCMjKmD6qRBKWrClxc:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5691, opened 2024-01-16
Do not try to resolve API address if `MULLVAD_API_HOST` is not set

MarkusPettersson98:
If the `api-override` feature is enabled, we allow overriding the ip address to the Mullvad API by settings the `MULLVAD_API_ADDRESS` env variable or have it be resolved automatically if `MULLVAD_API_HOST` is overriden.

This commit fixes a bug where we would try to resolve the API address when `MULLVAD_API_ADDRESS` was not set even if `MULLVAD_API_ADDRESS` was not set either. This would cause the daemon to panic when starting, potentially being unable to start up on consecutive retry attempts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5691)

### DIFF
--- a/mullvad-api/src/lib.rs
+++ b/mullvad-api/src/lib.rs
@@ -148,35 +148,42 @@ impl ApiEndpoint {
         let disable_tls_var = Self::read_var(ApiEndpoint::DISABLE_TLS_VAR);
 
         let mut api = ApiEndpoint {
-            host: host_var.clone(),
+            host: None,
             address: None,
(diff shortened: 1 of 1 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5691)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5697, opened 2024-01-17
Bind to the service in a coroutine to avoid a crash

Pururun:
We have a crash on Google Play that throws android.app.BackgroundServiceStartNotAllowedException.
Now this exception is thrown when a background service is started by an app in the background.
Since we are attempting to start our service in `onStart` the app should be in the foreground when the service is starting so it is a bit unclear under which conditions this exception is thrown. Even more so since there's a grace period of about 30 seconds after the app is moved to the background that the app is still allowed to start a background service.
I have thought about two possible scenarios:
1. The app is started up and for some reason the app is for some reason not yet considered to be in the foreground when start service is called. This is mitigated by only running the coroutine when the lifecycle state changes to OnStart.
2. The app is backgrounded but the call to start services is delayed/blocked for some reason and then it is continued after the app has been in the background for quite some time. This partially fixed by cancelling the coroutine in onStop, but there are scenarios that I have tested that even though the coroutine is cancelled that start service is called anyway.

So my hope is that this will decrease the amount of crashes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697)

### DIFF
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt
@@ -9,8 +9,12 @@ import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.core.view.WindowCompat
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
 import kotlinx.coroutines.flow.filterIsInstance
 import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.launch
 import net.mullvad.mullvadvpn.compose.screen.MullvadApp
 import net.mullvad.mullvadvpn.di.paymentModule
 import net.mullvad.mullvadvpn.di.uiModule
@@ -60,46 +64,40 @@ class MainActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
 
         setContent { AppTheme { MullvadApp() } }
+
+        // We use lifecycleScope here to get less start service in background exceptions
+        // Se this article for more information:
+        // https://medium.com/@lepicekmichal/android-background-service-without-hiccup-501e4479110f
+        lifecycleScope.launch {
+            repeatOnLifecycle(Lifecycle.State.STARTED) {
+                if (privacyDisclaimerRepository.hasAcceptedPrivacyDisclosure()) {
+                    startServiceSuspend(waitForConnectedReady = false)
+                }
+            }
+        }
     }
 
-    fun initializeStateHandlerAndServiceConnection() {
+    suspend fun startServiceSuspend(waitForConnectedReady: Boolean = true) {
         requestNotificationPermissionIfMissing(requestNotificationPermissionLauncher)
         serviceConnectionManager.bind(
             vpnPermissionRequestHandler = ::requestVpnPermission,
             apiEndpointConfiguration = intent?.getApiEndpointConfigurationExtras()
         )
-    }
-
-    suspend fun startServiceSuspend() {
-        requestNotificationPermissionIfMissing(requestNotificationPermissionLauncher)
-        serviceConnectionManager.bind(
-            vpnPermissionRequestHandler = ::requestVpnPermission,
-            apiEndpointConfiguration = intent?.getApiEndpointConfigurationExtras()
-        )
-        // Ensure we wait until the service is ready
-        serviceConnectionManager.connectionState
-            .filterIsInstance<ServiceConnectionState.ConnectedReady>()
-            .first()
+        if (waitForConnectedReady) {
+            // Ensure we wait until the service is ready
+            serviceConnectionManager.connectionState
+                .filterIsInstance<ServiceConnectionState.ConnectedReady>()
+                .first()
+        }
     }
 
     override fun onActivityResult(requestCode: Int, resultCode: Int, resultData: Intent?) {
         serviceConnectionManager.onVpnPermissionResult(resultCode == Activity.RESULT_OK)
     }
 
-    override fun onStart() {
-        super.onStart()
-
-        if (privacyDisclaimerRepository.hasAcceptedPrivacyDisclosure()) {
-            initializeStateHandlerAndServiceConnection()
-        }
-    }
-
     override fun onStop() {
         Log.d("mullvad", "Stopping main activity")
         super.onStop()
-
-        // NOTE: `super.onStop()` must be called before unbinding due to the fragment state handling
-        // otherwise the fragments will believe there was an unexpected disconnect.
         serviceConnectionManager.unbind()
     }
 


albin-mullvad (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697)* status: all files reviewed, 2 unresolved discussions (waiting on @Pururun)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697#-NoQb4UnBtoO-ggeJ4JN:-NoQb4UnBtoO-ggeJ4JO:b-jpigoq) (no related file):*
Can you elaborate a bit on why/how this PR is reducing the risk of crashing? Would be nice to add that additional information to the PR description.

___
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt` line 102 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697#-NoQaUljBx7UTI_jC2QV:-NoQaUljBx7UTI_jC2QW:b-a4b1z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9b18dbd638b06d0aa6d1eb063ebe9b81294a597/android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt#L102)):*
> ```Kotlin
> 
>         // NOTE: `super.onStop()` must be called before unbinding due to the fragment state handling
>         // otherwise the fragments will believe there was an unexpected disconnect.
> ```

Would be nice to clean up this comment since it relates to fragments that we no longer use.



_Code quote:_
```Kotlin
        // NOTE: `super.onStop()` must be called before unbinding due to the fragment state handling
        // otherwise the fragments will believe there was an unexpected disconnect.
```

Rawa (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @Pururun)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697#-NoQb4UnBtoO-ggeJ4JN:-Np-WXaD5VAU3SP0yhTQ:b-tfep2d) (no related file):*

Can you elaborate a bit on why/how this PR is reducing the risk of crashing? Would be nice to add that additional information to the PR description.
</blockquote></details>

1. Is it really that the app is not considered to be in the foreground that is causing this?  `initializeStateHandlerAndServiceConnection`is only called from onStart right so our app should be alive? Or could it be something along the lines that we call `onStart`, app crashes for other reason and since our service is separate process it keeps going? I'm not exactly sure how the logic for `android.app.BackgroundServiceStartNotAllowedException` so this might not make sense.

2. If the app is put to background wouldn't onStop be called and the service unbound? Since we use `synchronized` and flags to keep track of `bound` I can totally see how this could also cause us no not correctly unbind our service given if a flag was called or not?
 

___
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt` line 77 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697#-Np-UWdX4sUFygLXBsWD:-Np-UWdX4sUFygLXBsWE:b-105pmx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d32c5c5d495ef34b2a9ffac720affca135cd384f/android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt#L77)):*
> ```Kotlin
>     }
> 
>     suspend fun startServiceSuspend(waitForConnectedReady: Boolean = true) {
> ```

Why do we need this flag if we are only settings it to `true`?

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @Rawa)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697#-NoQb4UnBtoO-ggeJ4JN:-Np0HMH6ETNdNh6eEvxT:byvjzj7) (no related file):*

1. Is it really that the app is not considered to be in the foreground that is causing this?  `initializeStateHandlerAndServiceConnection`is only called from onStart right so our app should be alive? Or could it be something along the lines that we call `onStart`, app crashes for other reason and since our service is separate process it keeps going? I'm not exactly sure how the logic for `android.app.BackgroundServiceStartNotAllowedException` so this might not make sense.

2. If the app is put to background wouldn't onStop be called and the service unbound? Since we use `synchronized` and flags to keep track of `bound` I can totally see how this could also cause us no not correctly unbind our service given if a flag was called or not?
</blockquote></details>

1. It is exactly the call to `Context.startService` that causes the crash it is in the crash log:
```
at android.app.ContextImpl.startServiceCommon (ContextImpl.java:1940)
at android.app.ContextImpl.startService (ContextImpl.java:1886)
at android.content.ContextWrapper.startService (ContextWrapper.java:791)
at net.mullvad.mullvadvpn.ui.serviceconnection.ServiceConnectionManager.bind (ServiceConnectionManager.java)
```
And this is only called from `onStart` in MainActivity. So for me there can only be 2 causes for this:
1. OnStart is called even though the lifecycle state is not the correct one
2. The function is called and then blocked somewhere before it reaches startService and then continues after a while when the lifecycle state of the app has been updated.
Not sure how the service continue running in the background would cause this?

2. Not sure what you mean, do you mean that my change would cause an issue or that flag would cause the crash?

___
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt` line 77 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697#-Np-UWdX4sUFygLXBsWD:-Np0LMsI6VS0wuB9bnmc:b8rgny6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d32c5c5d495ef34b2a9ffac720affca135cd384f/android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt#L77)):*

Why do we need this flag if we are only settings it to `true`?
</blockquote></details>

My mistake I lost the false somewhere along the way, fixed.

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @Pururun)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697#-NoQb4UnBtoO-ggeJ4JN:-NpOYQ5n6h7gLxQ1fsoc:bys03tp) (no related file):*

1. It is exactly the call to `Context.startService` that causes the crash it is in the crash log:
```
at android.app.ContextImpl.startServiceCommon (ContextImpl.java:1940)
at android.app.ContextImpl.startService (ContextImpl.java:1886)
at android.content.ContextWrapper.startService (ContextWrapper.java:791)
at net.mullvad.mullvadvpn.ui.serviceconnection.ServiceConnectionManager.bind (ServiceConnectionManager.java)
```
And this is only called from `onStart` in MainActivity. So for me there can only be 2 causes for this:
1. OnStart is called even though the lifecycle state is not the correct one
2. The function is called and then blocked somewhere before it reaches startService and then continues after a while when the lifecycle state of the app has been updated.
Not sure how the service continue running in the background would cause this?

2. Not sure what you mean, do you mean that my change would cause an issue or that flag would cause the crash?
</blockquote></details>

1. I'm fine with this, I also read the linear comment and related post. We should probably add that as a comment to  change.
2. & 3. Let's discuss when you are in the office to avoid any confusion :)

___

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @Rawa)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697#-NoQb4UnBtoO-ggeJ4JN:-NpUR3GZEb1IJgEPsIY2:b-cqtw9d) (no related file):*

1. I'm fine with this, I also read the linear comment and related post. We should probably add that as a comment to  change.
2. & 3. Let's discuss when you are in the office to avoid any confusion :)
</blockquote></details>

Added a comment linking to the article.

___
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt` line 102 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697#-NoQaUljBx7UTI_jC2QV:-NpUR2LG5Ra5GRr6GdFk:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7e3fc6f7b76411e2519bf289b9ba3be20338801/android/app/src/main/kotlin/net/mullvad/mullvadvpn/ui/MainActivity.kt#L102)):*

Would be nice to clean up this comment since it relates to fragments that we no longer use.
</blockquote></details>

Done.

albin-mullvad (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697#-:-NpxiL2PDND4QkYQQcHT:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Rawa)

Rawa (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Rawa (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5697#-:-Nq1hifUBFdO0Eq7q-in:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5701, opened 2024-01-17
Add ad blocking test for iOS

niklasberglund:
Implemented test covering the ad blocking functionality. This PR introduces the attributes `MULLVAD_IOS_DEVICE_PIN_CODE` and `MULLVAD_AD_SERVING_DOMAIN` which need to be added to your `Configurations/UITests.xcconfig` file.

The test assumes a blank state, so the app should be uninstalled before running the test. It can be tried out by running the test `testAdBlockingViaDNS`.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701)

### DIFF
--- a/ios/Configurations/UITests.xcconfig.template
+++ b/ios/Configurations/UITests.xcconfig.template
@@ -6,8 +6,14 @@
 //  Copyright © 2024 Mullvad VPN AB. All rights reserved.
 //
 
+// Pin code of the iOS device under test
+MULLVAD_IOS_DEVICE_PIN_CODE = 
+
+// Ad serving domain used when testing ad blocking. Not that we are assuming there's an HTTP server running on the host.
+MULLVAD_AD_SERVING_DOMAIN = 
+
 // Mullvad accounts used by UI tests
 MULLVAD_NO_TIME_ACCOUNT_NUMBER = 
 MULLVAD_HAS_TIME_ACCOUNT_NUMBER = 
-MULLVAD_FIVE_WIREGUARD_KEYS_ACCOUNT_NUMBER =
+MULLVAD_FIVE_WIREGUARD_KEYS_ACCOUNT_NUMBER = 
  
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -568,13 +568,18 @@
 		7AF9BE902A39F26000DBFEDB /* Collection+Sorting.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AF9BE8F2A39F26000DBFEDB /* Collection+Sorting.swift */; };
 		7AF9BE952A40461100DBFEDB /* RelayFilterView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AF9BE942A40461100DBFEDB /* RelayFilterView.swift */; };
 		7AF9BE972A41C71F00DBFEDB /* RelayFilterChipView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AF9BE962A41C71F00DBFEDB /* RelayFilterChipView.swift */; };
+		850201DB2B503D7700EF8C96 /* RelayTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 850201DA2B503D7700EF8C96 /* RelayTests.swift */; };
+		850201DD2B503D8C00EF8C96 /* SelectLocationPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 850201DC2B503D8C00EF8C96 /* SelectLocationPage.swift */; };
+		850201DF2B5040A500EF8C96 /* TunnelControlPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 850201DE2B5040A500EF8C96 /* TunnelControlPage.swift */; };
 		850201E12B51389500EF8C96 /* BaseUITestCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 850201E02B51389500EF8C96 /* BaseUITestCase.swift */; };
+		850201E32B51A93C00EF8C96 /* SettingsPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 850201E22B51A93C00EF8C96 /* SettingsPage.swift */; };
 		852969282B4D9C1F007EAD4C /* AccountTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 852969272B4D9C1F007EAD4C /* AccountTests.swift */; };
 		852969332B4E9232007EAD4C /* Page.swift in Sources */ = {isa = PBXBuildFile; fileRef = 852969322B4E9232007EAD4C /* Page.swift */; };
 		852969352B4E9270007EAD4C /* LoginPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 852969342B4E9270007EAD4C /* LoginPage.swift */; };
 		852969362B4E9724007EAD4C /* AccessbilityIdentifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A0B311D2B303A0D004B12E0 /* AccessbilityIdentifier.swift */; };
 		8529693A2B4F0238007EAD4C /* TermsOfServicePage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 852969392B4F0238007EAD4C /* TermsOfServicePage.swift */; };
 		8529693C2B4F0257007EAD4C /* Alert.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8529693B2B4F0257007EAD4C /* Alert.swift */; };
+		85557B162B5ABBBE00795FE1 /* XCUIElementQuery+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85557B152B5ABBBE00795FE1 /* XCUIElementQuery+Extensions.swift */; };
 		A900E9B82ACC5C2B00C95F67 /* AccountsProxy+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A900E9B72ACC5C2B00C95F67 /* AccountsProxy+Stubs.swift */; };
 		A900E9BA2ACC5D0600C95F67 /* RESTRequestExecutor+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A900E9B92ACC5D0600C95F67 /* RESTRequestExecutor+Stubs.swift */; };
 		A900E9BC2ACC609200C95F67 /* DevicesProxy+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A900E9BB2ACC609200C95F67 /* DevicesProxy+Stubs.swift */; };
@@ -1720,7 +1725,11 @@
 		7AF9BE8F2A39F26000DBFEDB /* Collection+Sorting.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "Collection+Sorting.swift"; sourceTree = "<group>"; };
 		7AF9BE942A40461100DBFEDB /* RelayFilterView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RelayFilterView.swift; sourceTree = "<group>"; };
 		7AF9BE962A41C71F00DBFEDB /* RelayFilterChipView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RelayFilterChipView.swift; sourceTree = "<group>"; };
+		850201DA2B503D7700EF8C96 /* RelayTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RelayTests.swift; sourceTree = "<group>"; };
+		850201DC2B503D8C00EF8C96 /* SelectLocationPage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SelectLocationPage.swift; sourceTree = "<group>"; };
+		850201DE2B5040A500EF8C96 /* TunnelControlPage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TunnelControlPage.swift; sourceTree = "<group>"; };
 		850201E02B51389500EF8C96 /* BaseUITestCase.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BaseUITestCase.swift; sourceTree = "<group>"; };
+		850201E22B51A93C00EF8C96 /* SettingsPage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SettingsPage.swift; sourceTree = "<group>"; };
 		852969252B4D9C1F007EAD4C /* MullvadVPNUITests.xctest */ = {isa = PBXFileReference; explicitFileType = wrapper.cfbundle; includeInIndex = 0; path = MullvadVPNUITests.xctest; sourceTree = BUILT_PRODUCTS_DIR; };
 		852969272B4D9C1F007EAD4C /* AccountTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountTests.swift; sourceTree = "<group>"; };
 		852969302B4D9E70007EAD4C /* MullvadVPNUITests.xctestplan */ = {isa = PBXFileReference; lastKnownFileType = text; path = MullvadVPNUITests.xctestplan; sourceTree = "<group>"; };
@@ -1730,6 +1739,7 @@
 		852969382B4ED818007EAD4C /* UITests.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; path = UITests.xcconfig; sourceTree = "<group>"; };
 		852969392B4F0238007EAD4C /* TermsOfServicePage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TermsOfServicePage.swift; sourceTree = "<group>"; };
 		8529693B2B4F0257007EAD4C /* Alert.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Alert.swift; sourceTree = "<group>"; };
+		85557B152B5ABBBE00795FE1 /* XCUIElementQuery+Extensions.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "XCUIElementQuery+Extensions.swift"; sourceTree = "<group>"; };
 		A900E9B72ACC5C2B00C95F67 /* AccountsProxy+Stubs.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "AccountsProxy+Stubs.swift"; sourceTree = "<group>"; };
 		A900E9B92ACC5D0600C95F67 /* RESTRequestExecutor+Stubs.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "RESTRequestExecutor+Stubs.swift"; sourceTree = "<group>"; };
 		A900E9BB2ACC609200C95F67 /* DevicesProxy+Stubs.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "DevicesProxy+Stubs.swift"; sourceTree = "<group>"; };
@@ -2937,6 +2947,7 @@
 				58CE5E62224146200008646E /* MullvadVPN */,
 				58D0C79423F1CE7000FE9BA7 /* MullvadVPNScreenshots */,
 				58B0A2A1238EE67E00BC001D /* MullvadVPNTests */,
+				852969262B4D9C1F007EAD4C /* MullvadVPNUITests */,
 				58D223F4294C8FF00029F5F8 /* MullvadLogging */,
 				581943F228F8014500B0CB5E /* MullvadTypes */,
 				06799ABD28F98E1D00ACD94E /* MullvadREST */,
@@ -2952,7 +2963,6 @@
 				58C7A4432A863F490060C66F /* PacketTunnelCoreTests */,
 				7A88DCCF2A8FABBE00D2FF0E /* Routing */,
 				7A88DCDD2A8FABBE00D2FF0E /* RoutingTests */,
-				852969262B4D9C1F007EAD4C /* MullvadVPNUITests */,
 				58CE5E61224146200008646E /* Products */,
 				584F991F2902CBDD001F858D /* Frameworks */,
 			);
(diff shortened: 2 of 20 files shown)


niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701)* status: 0 of 18 files reviewed, 1 unresolved discussion

___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 84 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoNGFKb-LP18_HA9wJL:-NoNGFKb-LP18_HA9wJM:b2ozytn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/RelayTests.swift#L84)):*
> ```Swift
> 
>         if let urlError = requestError as? URLError {
>             if urlError.code == .cannotFindHost {
> ```

I'm not entirely sure whether this is a good way to verify that the domain can or cannot be resolved. Does DNS lookup do a better job at verifying that the domain is resolvable? There's also a limitation that this will only work when the domain points to a host running a HTTP server.

buggmagnet (changes requested):
> Reviewed 18 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701)* status: all files reviewed, 6 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPN.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` line 0 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoVvS306Pm5m9O67Q7v:-NoVvS306Pm5m9O67Q7w:bos3m7r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPN.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved#L0)):*
That usually happens when you switch branches from the CLI, sometimes Ccode gets conflicted and automatically deletes that file.
We actually need that file, please re-checkout that file from main like so
`git checkout main -- MullvadVPN.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved`
___
*[`ios/MullvadVPNUITests/BaseUITestCase.swift` line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoVxoKH7JXzXNz2m6Tj:-NoVxoKH7JXzXNz2m6Tk:bmxwpzt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/BaseUITestCase.swift#L15)):*
> ```Swift
>     public static let defaultTimeout = 10.0
> 
>     // swiftlint:disable force_cast line_length
> ```

Superfluous Disable Command Violation: SwiftLint rule 'line_length' did not trigger a violation in the disabled region; remove the disable command (superfluous_disable_command)
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 56 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoVx0ZI1_pt6cTrC9L-:-NoVx0ZI1_pt6cTrC9L0:b-59m9ef) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/RelayTests.swift#L56)):*
> ```Swift
>     }
> 
>     /// Verify that  ad serving domain is reachable by making sure the host can be found when sending HTTP request to it
> ```

Verify that `an` ad serving...
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 81 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoW1dk523W6AuQlkPwb:-NoW1dk523W6AuQlkPwc:b-uey0qt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/RelayTests.swift#L81)):*
> ```Swift
>         task.resume()
> 
>         _ = semaphore.wait(timeout: .distantFuture)
> ```

This is a surefire way to deadlock the test when things go wrong. Since we are running with `XCTest`, we can take advantage of its asynchronous waiting APIs instead

```swift
    private func canReachAdServingDomain() -> Bool {
        guard let url = URL(string: "http://\(adServingDomain)") else { return false }
        var requestError: Error?

        let urlErrorExpectation = expectation(description: "URLSession encountered an error")
        let task = URLSession.shared.dataTask(with: url) { _, _, error in
            requestError = error
            urlErrorExpectation.fulfill()
        }

        task.resume()

        wait(for: [urlErrorExpectation], timeout: 30)

        if let urlError = requestError as? URLError {
            if urlError.code == .cannotFindHost {
                return false
            }
        }

        return true
    }
```
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 84 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoNGFKb-LP18_HA9wJL:-NoWBlqu9t0lolqsJX_P:b-vqo7d1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/RelayTests.swift#L84)):*

I'm not entirely sure whether this is a good way to verify that the domain can or cannot be resolved. Does DNS lookup do a better job at verifying that the domain is resolvable? There's also a limitation that this will only work when the domain points to a host running a HTTP server.
</blockquote></details>

I dug a bit around, I think this error code is the correct one to use.
If we want to be extra sure, when that happens, we can assert that the `response` parameter in the callback (the 2nd, before the error) is `nil`, which will be the case when the DNS is blocked. 
___
*[`ios/MullvadVPNUITests/Pages/SettingsPage.swift` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoWErKOBFM6iijFHYno:-NoWErKPAUQna774mZH0:b8tpj7b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/Pages/SettingsPage.swift#L20)):*
> ```Swift
> 
>     @discardableResult func tapVPNSettingsCell() -> Self {
>         app.tables[AccessibilityIdentifier.settingsTableView.rawValue]
> ```

To avoid calling `rawValue` everywhere, we could use write extension that does it on our behalf like so

```swift
extension XCUIElementQuery {
    subscript(key: any RawRepresentable<String>) -> XCUIElement {
        self[key.rawValue]
    }
}
```

It makes reading a nicer experience 

```swift
    @discardableResult func tapVPNSettingsCell() -> Self {
        app.tables[AccessibilityIdentifier.settingsTableView]
            .cells[AccessibilityIdentifier.preferencesCell]
            .tap()

        return self
    }
````

**Review by rablador — CHANGES_REQUESTED** [review]
> Reviewed 18 of 18 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @niklasberglund)

___
*[`ios/MullvadVPNUITests/BaseUITestCase.swift` line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoVxoKH7JXzXNz2m6Tj:-NoXPIXCE3NwwMUeAba_:bj8a292) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/BaseUITestCase.swift#L15)):*

Superfluous Disable Command Violation: SwiftLint rule 'line_length' did not trigger a violation in the disabled region; remove the disable command (superfluous_disable_command)
</blockquote></details>

Also, please line break it somehow. Eg. 



_Code snippet:_
```Swift
Bundle(for: BaseUITestCase.self)
    .infoDictionary?["MullvadNoTimeAccountNumber"] as! String
```
___
*[`ios/MullvadVPNUITests/Pages/SettingsPage.swift` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoWErKOBFM6iijFHYno:-NoXQpo89maz_pc6txAe:b32v57) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/Pages/SettingsPage.swift#L20)):*

To avoid calling `rawValue` everywhere, we could use write extension that does it on our behalf like so

```swift
extension XCUIElementQuery {
    subscript(key: any RawRepresentable<String>) -> XCUIElement {
        self[key.rawValue]
    }
}
```

It makes reading a nicer experience 

```swift
    @discardableResult func tapVPNSettingsCell() -> Self {
        app.tables[AccessibilityIdentifier.settingsTableView]
            .cells[AccessibilityIdentifier.preferencesCell]
            .tap()

        return self
    }
````
</blockquote></details>

+1

**Review by niklasberglund — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701)* status: 9 of 21 files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPNUITests/BaseUITestCase.swift` line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoVxoKH7JXzXNz2m6Tj:-NoWej4G6GqW7Pcwk-W7:byscv74) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/BaseUITestCase.swift#L15)):*
Ah yes 👍 There was a line exceeding the limit before but it has been moved.

> Also, please line break it somehow. Eg.

Oops, I see now that some were separated to multiple lines but some not. Totally agree that they should all follow the same format. Not sure all separated into multiple lines is clearer, see first code snippet below where the lines have been separated into two, and second code snippet which disables `line_length` and let these lines be long. How do you think?



_Code snippet (i):_
```Swift
    // swiftlint:disable force_cast
    let noTimeAccountNumber = Bundle(for: BaseUITestCase.self)
        .infoDictionary?["MullvadNoTimeAccountNumber"] as! String
    let hasTimeAccountNumber = Bundle(for: BaseUITestCase.self)
        .infoDictionary?["MullvadHasTimeAccountNumber"] as! String
    let fiveWireGuardKeysAccountNumber = Bundle(for: BaseUITestCase.self)
        .infoDictionary?["MullvadFiveWireGuardKeysAccountNumber"] as! String
    let iOSDevicePinCode = Bundle(for: BaseUITestCase.self)
        .infoDictionary?["MullvadIOSDevicePinCode"] as! String
    let adServingDomain = Bundle(for: BaseUITestCase.self)
        .infoDictionary?["MullvadAdServingDomain"] as! String
    // swiftlint:enable force_cast
```



_Code snippet (ii):_
```Swift
    // swiftlint:disable force_cast line_length
    let noTimeAccountNumber = Bundle(for: BaseUITestCase.self).infoDictionary?["MullvadNoTimeAccountNumber"] as! String
    let hasTimeAccountNumber = Bundle(for: BaseUITestCase.self).infoDictionary?["MullvadHasTimeAccountNumber"] as! String
    let fiveWireGuardKeysAccountNumber = Bundle(for: BaseUITestCase.self).infoDictionary?["MullvadFiveWireGuardKeysAccountNumber"] as! String
    let iOSDevicePinCode = Bundle(for: BaseUITestCase.self).infoDictionary?["MullvadIOSDevicePinCode"] as! String
    let adServingDomain = Bundle(for: BaseUITestCase.self).infoDictionary?["MullvadAdServingDomain"] as! String
    // swiftlint:enable force_cast line_length
```
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 56 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoVx0ZI1_pt6cTrC9L-:-NoWjAxD3s50xekV5dCt:b-886gsx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/RelayTests.swift#L56)):*

Verify that `an` ad serving...
</blockquote></details>

Fixed
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 81 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoW1dk523W6AuQlkPwb:-NoWkeJX4jLOWTuR-wMx:b-fkywll) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/RelayTests.swift#L81)):*

This is a surefire way to deadlock the test when things go wrong. Since we are running with `XCTest`, we can take advantage of its asynchronous waiting APIs instead

```swift
    private func canReachAdServingDomain() -> Bool {
        guard let url = URL(string: "http://\(adServingDomain)") else { return false }
        var requestError: Error?

        let urlErrorExpectation = expectation(description: "URLSession encountered an error")
        let task = URLSession.shared.dataTask(with: url) { _, _, error in
            requestError = error
            urlErrorExpectation.fulfill()
        }

        task.resume()

        wait(for: [urlErrorExpectation], timeout: 30)

        if let urlError = requestError as? URLError {
            if urlError.code == .cannotFindHost {
                return false
            }
        }

        return true
    }
```
</blockquote></details>

Ah yes, `.distantFuture` is not a good choice for timeout. Is wait preferred over semaphore? It's in a layer under the testing code, and using semaphore feels more natural than `expectation` to me here because the expectation is not on some specific data but used to wait for response.
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 84 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoNGFKb-LP18_HA9wJL:-NoXE_3cAYJmIwusDUM1:bird6xi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/RelayTests.swift#L84)):*

I dug a bit around, I think this error code is the correct one to use.
If we want to be extra sure, when that happens, we can assert that the `response` parameter in the callback (the 2nd, before the error) is `nil`, which will be the case when the DNS is blocked.
</blockquote></details>

Thanks for verifying! See the updated code where both `.cannotFindHost` error and `nil` value for `response` is taken into account.
___
*[`ios/MullvadVPNUITests/Pages/SettingsPage.swift` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoWErKOBFM6iijFHYno:-NoXENu-1c0MtvtsR7wj:bri491m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/Pages/SettingsPage.swift#L20)):*

To avoid calling `rawValue` everywhere, we could use write extension that does it on our behalf like so

```swift
extension XCUIElementQuery {
    subscript(key: any RawRepresentable<String>) -> XCUIElement {
        self[key.rawValue]
    }
}
```

It makes reading a nicer experience 

```swift
    @discardableResult func tapVPNSettingsCell() -> Self {
        app.tables[AccessibilityIdentifier.settingsTableView]
            .cells[AccessibilityIdentifier.preferencesCell]
            .tap()

        return self
    }
````
</blockquote></details>

Reads much better 👍 Have added the extension and removed all `.rawValue`.
___
*[`ios/MullvadVPN.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` line  at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoVvS306Pm5m9O67Q7v:-NoWeTOk9RD7LHbjUbAD:b22psr6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPN.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved#L)):*

That usually happens when you switch branches from the CLI, sometimes Ccode gets conflicted and automatically deletes that file.
We actually need that file, please re-checkout that file from main like so
`git checkout main -- MullvadVPN.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved`
</blockquote></details>

Oops! Added it again

rablador (approved):
> Reviewed 12 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPNUITests/BaseUITestCase.swift` line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoVxoKH7JXzXNz2m6Tj:-NokRJQn4D6E3GVWQLqN:b-2ob90t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/BaseUITestCase.swift#L15)):*

Ah yes 👍 There was a line exceeding the limit before but it has been moved.

> Also, please line break it somehow. Eg.

Oops, I see now that some were separated to multiple lines but some not. Totally agree that they should all follow the same format. Not sure all separated into multiple lines is clearer, see first code snippet below where the lines have been separated into two, and second code snippet which disables `line_length` and let these lines be long. How do you think?
</blockquote></details>

I think the first snippet is fine. We usually follow the swiftlint rather than disable it, with some exceptions where needed of course.

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701)* status: 20 of 21 files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPNUITests/BaseUITestCase.swift` line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoVxoKH7JXzXNz2m6Tj:-Nom6cV80HP-rWxQbHRb:bld9chs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/BaseUITestCase.swift#L15)):*

I think the first snippet is fine. We usually follow the swiftlint rather than disable it, with some exceptions where needed of course.
</blockquote></details>

Sounds good. Have changed it to follow the format of the first snippet.

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet)

buggmagnet (changes requested):
> Reviewed 11 of 12 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701)* status: all files reviewed, 2 unresolved discussions (waiting on @niklasberglund and @rablador)

___
*[`ios/MullvadVPNUITests/BaseUITestCase.swift` line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoVxoKH7JXzXNz2m6Tj:-NoqfGti4lfEdTXwoW22:bg1kpza) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/BaseUITestCase.swift#L15)):*

Sounds good. Have changed it to follow the format of the first snippet.
</blockquote></details>

Yeah I would say usually as long as swiftlint + swiftformat don't complain about it, then I won't either, we've put those tools in place because we used to have endless arguments about how code should look like.
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 81 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoW1dk523W6AuQlkPwb:-NoqfwzP9KeyBTurcXbU:b-rvbi2y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/RelayTests.swift#L81)):*
There are a couple of problems with that piece of code
- The return value of `semaphore.wait(timeout: .now() + 10)`, is not checked, so we don't know whether the code timed out, or did run to completion.
By using an expectation, we get the desired outcome, if that network call didn't run as expected, we fail the test (which would / should fail anyway)

- If the test were using the swift `async` runtime, it would be a violation of the contract that "a thread must always be allowed to progress" which semaphore prevents, which would lead to either undefined behaviour, runtime crashes, or deadlocks.

> the expectation is not on some specific data but used to wait for response.

The expectation is asserting essentially that the completion was calledback, right now, if it doesn't happen, this function is relying on the call site to indicate failure.

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 81 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-NoW1dk523W6AuQlkPwb:-NouiIyfC38deBwquZQb:bet187u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfc8813c24faa437eed7a66608c1d408d9696ef8/ios/MullvadVPNUITests/RelayTests.swift#L81)):*

There are a couple of problems with that piece of code
- The return value of `semaphore.wait(timeout: .now() + 10)`, is not checked, so we don't know whether the code timed out, or did run to completion.
By using an expectation, we get the desired outcome, if that network call didn't run as expected, we fail the test (which would / should fail anyway)

- If the test were using the swift `async` runtime, it would be a violation of the contract that "a thread must always be allowed to progress" which semaphore prevents, which would lead to either undefined behaviour, runtime crashes, or deadlocks.

> the expectation is not on some specific data but used to wait for response.

The expectation is asserting essentially that the completion was calledback, right now, if it doesn't happen, this function is relying on the call site to indicate failure.
</blockquote></details>

Ah yes, if the timeout is reached `resultIndicatesDNSBlock` will be `false` since that's the initial value, which will be incorrect in that case. Could rewrite a bit to fix but it's easier to understand the code if checking whether timeout was reached 👍 

I'm not opposed to changing the syntax. Isn't the difference between `wait` and `DispatchSemaphore` semantical, or do they also work in different ways? For an application it would be bad to write synchronous code but for tests I'd say it's preferred, want to avoid asynchronous code in test case abstraction layer.

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5701#-:-NovcLKT4JtYpFJ8Va27:bnfp4nl)* status: 19 of 21 files reviewed, all discussions resolved (waiting on @rablador)

PR #5709, opened 2024-01-19
Fix App Icon crashes in Split Tunneling

Pururun:
This PR aims to handle two exceptions when getting an app icon in split tunneling.
First is `IllegalArgumentException` which can happen if the icon has width or height set to 0 for some reason. This has been reported by users.
The second is `NameNotFoundException` which can happen if the app is not installed, but shows up in the app list anyway (for example if the app has been installed on an external drive that is currently not available). This was handled before we removed the old implementation so this is a regression fix.

Both cases are handled by just returning a null bitmap which is a valid.

Also since we are no longer using `ApplicationsIconManager` this class was removed and some code related to it was also removed in the tests and ui module.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5709)


albin-mullvad (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5709#-:-Nol9SAu6QWpDYFjc1l4:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5709)* status: all files reviewed, 1 unresolved discussion (waiting on @Pururun)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/SplitTunnelingScreen.kt` line 98 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5709#-NolPxiL5y4F-B7EqHR3:-NolPxiL5y4F-B7EqHR4:bvl19nc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/743b0353e8c003107f205bdcea3cac5860d82c68/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/SplitTunnelingScreen.kt#L98)):*
> ```Kotlin
>         onResolveIcon = { packageName ->
>             try {
>                 packageManager.getApplicationIcon(packageName).toBitmapOrNull()
> ```

For readability it would be nice to break this out to a extension function or similar, what do you think?

Rawa (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5709#-:-NolQ6Z82GNAZK8s5vpb:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Pururun)

PR #5715, opened 2024-01-22
Use infallible HeaderValue::from() when converting from usize

pronebird:
`HeaderValue` in hyper support a multitude of `From` traits, which are infallible for primitives such as `usize` but fallible for strings. This minor refactoring PR removes the need for going via string when setting the `Content-Length` header.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5715)


dlon (approved):
Nice :)

faern:
Thanks @pronebird :) nice little cleanup

PR #5720, opened 2024-01-23
Add account expiry mockapi test

Pururun:
This has essentially two potential success states since I was not able to figure out a way to make sure either of them happen.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5720)

### DIFF
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/OutOfTimeScreen.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/OutOfTimeScreen.kt
@@ -18,6 +18,7 @@ import androidx.compose.runtime.collectAsState
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalUriHandler
+import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
@@ -42,6 +43,7 @@ import net.mullvad.mullvadvpn.compose.destinations.SettingsDestination
 import net.mullvad.mullvadvpn.compose.destinations.VerificationPendingDialogDestination
 import net.mullvad.mullvadvpn.compose.extensions.createOpenAccountPageHook
 import net.mullvad.mullvadvpn.compose.state.OutOfTimeUiState
+import net.mullvad.mullvadvpn.compose.test.OUT_OF_TIME_SCREEN_TITLE_TEST_TAG
 import net.mullvad.mullvadvpn.compose.transitions.HomeTransition
 import net.mullvad.mullvadvpn.lib.payment.model.ProductId
 import net.mullvad.mullvadvpn.lib.theme.AppTheme
@@ -220,7 +222,9 @@ fun OutOfTimeScreen(
                 text = stringResource(id = R.string.out_of_time),
                 style = MaterialTheme.typography.headlineLarge,
                 color = MaterialTheme.colorScheme.onPrimary,
-                modifier = Modifier.padding(horizontal = Dimens.sideMargin)
+                modifier =
+                    Modifier.padding(horizontal = Dimens.sideMargin)
+                        .testTag(OUT_OF_TIME_SCREEN_TITLE_TEST_TAG)
             )
             Text(
                 text =
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/test/ComposeTestTagConstants.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/test/ComposeTestTagConstants.kt
@@ -39,3 +39,6 @@ const val LOGIN_INPUT_TEST_TAG = "login_input_test_tag"
 
 // VoucherDialog
 const val VOUCHER_INPUT_TEST_TAG = "voucher_input_test_tag"
+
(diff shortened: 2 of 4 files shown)


albin-mullvad (commented):
I'm not sure it's a good idea to proceed with this PR unless we can come up with a fully deterministic flow. It might be a good idea to check how or if the desktop team is testing this.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5720#-:-Nov6FmK1H-D5urklmOX:bd0cb6u)* status: 0 of 3 files reviewed, all discussions resolved

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5720)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @Pururun)

___
*[`android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/interactor/AppInteractor.kt` line 92 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5720#-Nozvf_CDlSBbO2lkTVo:-Nozvf_D8hjBxa1dfaqu:b-jk49rw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9b16e81c8789253287c866f21b4f54f943f9fae4/android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/interactor/AppInteractor.kt#L92)):*
> ```Kotlin
> 
>     fun ensureOutOfTime() {
>         device.findObjectWithTimeout(By.text("Out of time"))
> ```

`By.res` and test tag imo

Rawa (commented):
We had some discussions around this the name of `testOpenAccountPageOfAccountThatJustExpired` leaves a me a bit confused. 

Somethings we should take in regard:
- We should see if this behavior is not caused by a bug, e.g navigating twice.
- - We should also assert that the strategy how we update the expiry date actually works as expected, e.g if the daemon in the background does polling or if we are accidentally doing it somewhere else. 
- Possible split the test two different test, to have suffix "ThatJustExpired" leaves me do think we might can split this test multiple

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5720#-:-NozwgK77u3DDHq8mJak:b-87cjzi)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @Pururun)

Pururun (commented):
So I updated this PR to more closely follow our discussion. It has now two tests, one is just to make sure that when we open the account screen it should fetch the latest account expiry. I now changed it so that the account expiry is moved to the future to make it more deterministic. The second test is more like the original one, but I am now sure that the reason for why it is inconsistent is that the `ConnectScreen` will try to navigate to the `OutOfTimeScreen` despite it is has almost a minute before started navigating to the `AccountScreen`. I made a `try-catch` solution to handle both scenarios, which I am not really that happy with. So for me there are 3 solutions:
1. Keep the test as is in the PR and fix the issue in some other PR.
2. Write the test as it should be and just put it as ignore until the navigation behaviour in the app is fixed.
3. Delete the second test and return to this test in some future in which the app works as expected.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5720#-:-Np0M0OMErwSPSLTQQgd:b-hx5ue1)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @Rawa)

Rawa (commented):
> So I updated this PR to more closely follow our discussion. It has now two tests, one is just to make sure that when we open the account screen it should fetch the latest account expiry. I now changed it so that the account expiry is moved to the future to make it more deterministic.

Nice, this is a much healthier approach I feel like.

> The second test is more like the original one, but I am now sure that the reason for why it is inconsistent is that the `ConnectScreen` will try to navigate to the `OutOfTimeScreen` despite it is has almost a minute before started navigating to the `AccountScreen`. I made a `try-catch` solution to handle both scenarios, which I am not really that happy with. So for me there are 3 solutions:

Sweet, this sounds like something we should fix. If we are on account screen I feel we should wait with that navigation until it has shown the ConnectScreen again. Incorporating these screens as one would be one way to fix it. I think this is something we should solve with the ConnectScreen and not something that this test is at fault for.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5720#-:-NpOZA1D1eswmDGHh0CE:b-5btma9)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @Pururun)

Pururun (commented):
I made the second test as it should be and disabled it. The fix for the navigation issue is tracked here:
https://linear.app/mullvad/issue/DROID-658/outoftimescreen-is-navigated-to-despite-the-connectscreen-not-showing

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5720#-:-NpPW8aS65v83S4h8lSE:b-k66gsm)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @Rawa)

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5720)* status: 2 of 3 files reviewed, 1 unresolved discussion

Rawa (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5720#-:-NpUMJ9dEVORaOlO5rDN:bnfp4nl)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @Pururun)

PR #5723, opened 2024-01-24
Strip binaries with cargo instead of build.sh

faern:
I found this neat little blog post: https://kobzol.github.io/rust/cargo/2024/01/23/making-rust-binaries-smaller-by-default.html

On nightly Rust cargo will now by default strip debuginfo (coming to stable later). However symbols are still there. We have a more complex solution where we manually invoke `strip` on the produced binaries to yield the same result as we can get by setting `strip = true` in `Cargo.toml`.

This project fixes that, and simplifies build.sh. We also get stripping for anyone manually invoking `cargo build --release`. I see no downsides.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5723)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,7 @@ once_cell = "1.13"
 [profile.release]
 opt-level = 3
 lto = true
+strip = true
 
 # Key generation may take over one minute without optimizations
 # enabled.
--- a/build.sh
+++ b/build.sh
@@ -200,17 +200,12 @@ function sign_win {
 }
 
 # Build the daemon and other Rust/C++ binaries, optionally
-# sign them, strip them of debug symbols and copy to `dist-assets/`.
+# sign them, and copy to `dist-assets/`.
 function build {
     local current_target=${1:-""}
     local for_target_string=" for local target $HOST"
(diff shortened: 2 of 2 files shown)


dlon (approved):
:lgtm:

This means we're no longer stripping binaries without `--optimize`, right? That seems like a good change, though.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5723#-:-NovzSKq77ndUdNGrCUp:b6zqzgm)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (commented):
You are correct. I did not think about that. I assumed we kept debug symbols for non-optimized builds, but no. The non-optimized builds are *huge* anyway, debug symbols or not. And the build server never builds non-optimized builds. So yeah, I think this is fine.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5723#-:-NozVk6-2rUTiMP71F-c:b9uttbq)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5731, opened 2024-01-25
Modified ad blocking test flow

niklasberglund:
In the ad blocking test there's a race condition when enabling ad blocking and then verifying that an ad blocking domain cannot be reached. The problem is that changing the setting triggers a reconnect and there is no visual update in the app to wait for in order to make sure we're connected again before verifying that the ad blocking domain no longer can be reached. 

We decided to modify the test flow to:
1. Log into an account
2. Enable the ad blocking relay
3. Connect
4. Verify if the ad-serving domain is not reachable

See issue in Linear https://linear.app/mullvad/issue/IOS-466/race-condition-in-ad-blocking-test

The changes in this PR can be tested by running the `testAdBlockingViaDNS` test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5731)

### DIFF
--- a/ios/MullvadVPNUITests/Pages/Page.swift
+++ b/ios/MullvadVPNUITests/Pages/Page.swift
@@ -30,4 +30,10 @@ class Page {
         app.typeText(text)
         return self
     }
+
+    /// Fast swipe down action to dismiss a modal view. Will swipe on the middle of the screen.
+    @discardableResult func swipeDownToDismissModal() -> Self {
+        app.swipeDown(velocity: .fast)
+        return self
+    }
 }
--- a/ios/MullvadVPNUITests/RelayTests.swift
+++ b/ios/MullvadVPNUITests/RelayTests.swift
@@ -11,6 +11,16 @@ import XCTest
 
 class RelayTests: LoggedInWithTimeUITestCase {
     func testAdBlockingViaDNS() throws {
+        HeaderBar(app)
+            .tapSettingsButton()
+
+        SettingsPage(app)
+            .tapVPNSettingsCell()
+            .tapDNSSettingsCell()
+            .tapDNSContentBlockingHeaderExpandButton()
+            .tapBlockAdsSwitch()
+            .swipeDownToDismissModal()
+
         TunnelControlPage(app)
             .tapSelectLocationButton()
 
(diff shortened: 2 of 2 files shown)


buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5731)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 14 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5731#-NpYwsNt6wVjKip6EDV4:-NpYwsNt6wVjKip6EDV5:b-44zd0e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9bd0d77321200664e5a01fe5571aab9b02e4eba7/ios/MullvadVPNUITests/RelayTests.swift#L14)):*
> ```Swift
> class RelayTests: LoggedInWithTimeUITestCase {
>     func testAdBlockingViaDNS() throws {
>         TunnelControlPage(app)
> ```

Don't forget to `verifyCanReachAdServingDomain` before we run the test, otherwise we could get false positives if a previous run left ad blocking turned on for whatever reason
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5731#-NpYxFYU0yMCTIUkX8FO:-NpYxFYU0yMCTIUkX8FP:b-uzxgio) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9bd0d77321200664e5a01fe5571aab9b02e4eba7/ios/MullvadVPNUITests/RelayTests.swift#L15)):*
> ```Swift
>     func testAdBlockingViaDNS() throws {
>         TunnelControlPage(app)
>             .tapSettingsButton()
> ```

:red_circle:  Value of type 'TunnelControlPage' has no member 'tapSettingsButton'

buggmagnet (approved):
> Reviewed 1 of 3 files at r2, 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5731)* status: all files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 14 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5731#-NpYwsNt6wVjKip6EDV4:-Npd8fm10cub1uaYDqC6:b6p9k4q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9bd0d77321200664e5a01fe5571aab9b02e4eba7/ios/MullvadVPNUITests/RelayTests.swift#L14)):*

Brought this up with @pinkisemils before implementing this, we talked about not including `verifyCanReachAdServingDomain` since with this test flow the app is never in a state of being both connected and having ad blocking turned on. So even if ad blocking was left on from a previous run it wouldn't catch this, or would it even though we're not connected to a relay?
</blockquote></details>

Right, we uninstall the app before starting the test here, good point then !

PR #5732, opened 2024-01-26
Replace secondary container with solid color

Pururun:
Part of material 3 hack day thingy.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5732)

### DIFF
--- a/android/lib/theme/src/main/kotlin/net/mullvad/mullvadvpn/lib/theme/Theme.kt
+++ b/android/lib/theme/src/main/kotlin/net/mullvad/mullvadvpn/lib/theme/Theme.kt
@@ -15,7 +15,6 @@ import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.TextUnit
 import androidx.compose.ui.unit.dp
-import net.mullvad.mullvadvpn.lib.theme.color.MullvadBlue20
 import net.mullvad.mullvadvpn.lib.theme.color.MullvadDarkBlue
 import net.mullvad.mullvadvpn.lib.theme.color.MullvadGreen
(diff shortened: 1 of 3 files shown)


Rawa (commented):
Would it not just be possible to update `secondaryContainer` to be a proper color? Or is it used for something else in other places? Generally I would try to avoid using `alpha`, but if needed I think this solution is okay, at least we use the `compositeOver` so we don't get weird behaviour with other parts.

> Reviewed 7 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5732#-:-NpO_KHW41Iya4PYPegu:b3imlr3)* status: 7 of 8 files reviewed, all discussions resolved

Pururun (commented):
The reason was to free up `secondaryContainer` for the color we want to use for selections since that is more in line what they use as standard colors in various material components, which is a bit weird I agree. I do think we should use a solid color, probably better to change secondary container to that color and then switch colors around later.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5732#-:-NpQ6Fn0EMUn72OKotSb:bgfike)* status: 7 of 8 files reviewed, all discussions resolved

Rawa (commented):
Discussed offline to use fixed color

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5732#-:-NpUPQ6z9vQ9DD6tDGSp:bsyxl4s)* status: 7 of 8 files reviewed, all discussions resolved

Rawa (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5732#-:-NpUPUW7-GzWnw0xjEvW:bnfp4nl)* status: 7 of 8 files reviewed, all discussions resolved

PR #5748, opened 2024-02-01
Refactor TunnelManager account setting methods replacing callbacks with async

acb-mv:
This PR replaces public methods in TunnelManager, and their downstream consumers, which have used callbacks with versions using async instead, in the interest of modernising the codebase to use modern Swift concurrency.

The affected code deals with logging in, creating new accounts, deleting accounts, logging out and being bumped out due to device deletion.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5748)

### DIFF
--- a/ios/MullvadVPN/Coordinators/AccountCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/AccountCoordinator.swift
@@ -141,7 +141,7 @@ final class AccountCoordinator: Coordinator, Presentable, Presenting {
         )
 
         let alertPresenter = AlertPresenter(context: self)
-        
+
         Task {
             await interactor.logout()
             DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(1)) { [weak self] in
--- a/ios/MullvadVPN/Coordinators/AccountCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/AccountCoordinator.swift
@@ -141,8 +141,9 @@ final class AccountCoordinator: Coordinator, Presentable, Presenting {
         )
 
         let alertPresenter = AlertPresenter(context: self)
-
-        interactor.logout {
+        
+        Task {
+            await interactor.logout()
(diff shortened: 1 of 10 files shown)


buggmagnet (approved):
> Reviewed 10 of 10 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5748)* status: all files reviewed, 2 unresolved discussions (waiting on @acb-mv)

___
*[`ios/MullvadVPN/Coordinators/AccountCoordinator.swift` line 144 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5748#-Npcr0Na5lRzMhakF6p_:-Npcr0Na5lRzMhakF6pa:bjs8zpp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3d4b2ffdc661f79b3312a7c5c03762e4cf2eb491/ios/MullvadVPN/Coordinators/AccountCoordinator.swift#L144)):*
> ```Swift
> 
>         let alertPresenter = AlertPresenter(context: self)
>         
> ```

Please make sure to run `swiftformat .` in the `ios` folder before pushing changes
___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 374 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5748#-Npd-x7o1e7vwiENhc44:-Npd-x7o1e7vwiENhc45:betw4mg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3d4b2ffdc661f79b3312a7c5c03762e4cf2eb491/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L374)):*
> ```Swift
>         try await withCheckedThrowingContinuation { continuation in
>             setAccount(action: action) { result in
>                 continuation.resume(with: result)
> ```

It always irked me that we have a `Result` type with an Optional type defined as its success. 
I wonder if we could make the success type non optional here

Or at least we could make the callsite nicer by unwrapping the optional here

What do you think ?

PR #5761, opened 2024-02-05
Detect available IP versions

MarkusPettersson98:
Try to detect available IP versions by looking at the available routes on the host.

* On Linux, we check if there exists IPv4 and/or IPv6 routes to some public IP addresses. 
* On macOS and Windows, we check if there exists default routes for IPv4 and/or IPv6.

The intention is to be able to generate better default constraints for tunnel endpoints. If we can detect "working" IPv4 and/or IPv6 and forward this information to a `TunnelParametersGenerator`, we may choose to connect to a Wireguard relay using IPv6 as part of our retry-strategy.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761)

### DIFF
--- a/mullvad-daemon/src/api.rs
+++ b/mullvad-daemon/src/api.rs
@@ -21,7 +21,9 @@ use mullvad_types::access_method::{
 };
 use std::{net::SocketAddr, path::PathBuf};
 use talpid_core::mpsc::Sender;
-use talpid_types::net::{AllowedClients, AllowedEndpoint, Endpoint, TransportProtocol};
+use talpid_types::net::{
+    AllowedClients, AllowedEndpoint, Connectivity, Endpoint, TransportProtocol,
+};
 
 pub enum Message {
     Get(ResponseTx<ResolvedConnectionMode>),
@@ -527,18 +529,26 @@ pub fn allowed_clients(connection_mode: &ApiConnectionMode) -> AllowedClients {
     }
 }
 
+/// Forwards the received values from `offline_state_rx` to the [`ApiAvailabilityHandle`].
 pub(crate) fn forward_offline_state(
     api_availability: ApiAvailabilityHandle,
-    mut offline_state_rx: mpsc::UnboundedReceiver<bool>,
+    mut offline_state_rx: mpsc::UnboundedReceiver<Connectivity>,
 ) {
     tokio::spawn(async move {
-        let initial_state = offline_state_rx
+        let is_offline = offline_state_rx
             .next()
             .await
-            .expect("missing initial offline state");
-        api_availability.set_offline(initial_state);
-        while let Some(is_offline) = offline_state_rx.next().await {
-            api_availability.set_offline(is_offline);
+            .expect("missing initial offline state")
+            .is_offline();
+        log::info!(
+            "Initial offline state - {state}",
+            state = if is_offline { "offline" } else { "online" },
+        );
+        api_availability.set_offline(is_offline);
+
+        while let Some(state) = offline_state_rx.next().await {
+            log::info!("Detecting changes to offline state - {state:?}");
+            api_availability.set_offline(state.is_offline());
         }
     });
 }
--- a/mullvad-types/src/access_method.rs
+++ b/mullvad-types/src/access_method.rs
@@ -77,18 +77,9 @@ impl Settings {
     /// Check that `self` contains atleast one enabled access methods. If not,
     /// the `Direct` access method is re-enabled.
     fn ensure_consistent_state(&mut self) {
-        if self.collect_enabled().is_empty() {
+        if self.iter().all(|access_method| access_method.disabled()) {
             self.direct.enable();
         }
-    }
-
-    // TODO(markus): This can surely be removed.
-    /// Retrieve all [`AccessMethodSetting`]s which are enabled.
-    pub fn collect_enabled(&self) -> Vec<AccessMethodSetting> {
-        self.iter()
-            .filter(|access_method| access_method.enabled)
-            .cloned()
-            .collect()
     }
 
     /// Iterate over references of built-in & custom access methods.
@@ -128,8 +119,7 @@ impl Settings {
         &self.mullvad_bridges
     }
 
-    // TODO(markus): This can probably be made private
-    pub fn create_direct() -> AccessMethodSetting {
+    fn create_direct() -> AccessMethodSetting {
         let method = BuiltInAccessMethod::Direct;
         AccessMethodSetting::new(method.canonical_name(), true, AccessMethod::from(method))
     }
--- a/talpid-core/src/offline/android.rs
+++ b/talpid-core/src/offline/android.rs
@@ -4,13 +4,13 @@ use jnix::{
         self,
         objects::{GlobalRef, JObject, JValue},
         signature::{JavaType, Primitive},
-        sys::{jboolean, jlong, JNI_FALSE},
+        sys::{jboolean, jlong, JNI_TRUE},
         JNIEnv, JavaVM,
     },
     JnixEnv,
 };
 use std::sync::{Arc, Weak};
-use talpid_types::{android::AndroidContext, ErrorExt};
+use talpid_types::{android::AndroidContext, net::Connectivity, ErrorExt};
 
 #[derive(err_derive::Error, Debug)]
 #[error(no_from)]
@@ -43,13 +43,13 @@ pub struct MonitorHandle {
     jvm: Arc<JavaVM>,
     class: GlobalRef,
     object: GlobalRef,
-    _sender: Arc<UnboundedSender<bool>>,
+    _sender: Arc<UnboundedSender<Connectivity>>,
 }
 
 impl MonitorHandle {
     pub fn new(
         android_context: AndroidContext,
-        sender: Arc<UnboundedSender<bool>>,
+        sender: Arc<UnboundedSender<Connectivity>>,
     ) -> Result<Self, Error> {
         let env = JnixEnv::from(
             android_context
@@ -101,30 +101,29 @@ impl MonitorHandle {
     }
 
     #[allow(clippy::unused_async)]
-    pub async fn host_is_offline(&self) -> bool {
-        match self.get_is_connected() {
-            Ok(is_connected) => !is_connected,
-            Err(error) => {
+    pub async fn connectivity(&self) -> Connectivity {
+        self.get_is_connected()
+            .map(|connected| Connectivity::Status { connected })
+            .unwrap_or_else(|error| {
                 log::error!(
                     "{}",
                     error.display_chain_with_msg("Failed to check connectivity status")
                 );
-                false
-            }
-        }
+                Connectivity::PresumeOnline
+            })
     }
 
     fn get_is_connected(&self) -> Result<bool, Error> {
-        let result = self.call_method(
+        let is_connected = self.call_method(
             "isConnected",
             "()Z",
             &[],
             JavaType::Primitive(Primitive::Boolean),
         )?;
 
-        match result {
-            JValue::Bool(JNI_FALSE) => Ok(false),
-            JValue::Bool(_) => Ok(true),
+        match is_connected {
+            JValue::Bool(JNI_TRUE) => Ok(true),
+            JValue::Bool(_) => Ok(false),
             value => Err(Error::InvalidMethodResult(
                 "ConnectivityListener",
                 "isConnected",
@@ -133,7 +132,7 @@ impl MonitorHandle {
         }
     }
 
-    fn set_sender(&self, sender: Weak<UnboundedSender<bool>>) -> Result<(), Error> {
+    fn set_sender(&self, sender: Weak<UnboundedSender<Connectivity>>) -> Result<(), Error> {
         let sender_ptr = Box::new(sender);
         let sender_address = Box::into_raw(sender_ptr) as jlong;
 
@@ -182,14 +181,16 @@ impl MonitorHandle {
 pub extern "system" fn Java_net_mullvad_talpid_ConnectivityListener_notifyConnectivityChange(
     _: JNIEnv<'_>,
     _: JObject<'_>,
-    is_connected: jboolean,
+    connected: jboolean,
     sender_address: jlong,
 ) {
+    let connected = JNI_TRUE == connected;
     let sender_ref = Box::leak(unsafe { get_sender_from_address(sender_address) });
-    let is_offline = is_connected == JNI_FALSE;
-
     if let Some(sender) = sender_ref.upgrade() {
-        if sender.unbounded_send(is_offline).is_err() {
+        if sender
+            .unbounded_send(Connectivity::Status { connected })
+            .is_err()
+        {
             log::warn!("Failed to send offline change event");
         }
     }
@@ -206,13 +207,13 @@ pub extern "system" fn Java_net_mullvad_talpid_ConnectivityListener_destroySende
     let _ = unsafe { get_sender_from_address(sender_address) };
 }
 
-unsafe fn get_sender_from_address(address: jlong) -> Box<Weak<UnboundedSender<bool>>> {
-    Box::from_raw(address as *mut Weak<UnboundedSender<bool>>)
+unsafe fn get_sender_from_address(address: jlong) -> Box<Weak<UnboundedSender<Connectivity>>> {
+    Box::from_raw(address as *mut Weak<UnboundedSender<Connectivity>>)
 }
 
 #[allow(clippy::unused_async)]
 pub async fn spawn_monitor(
-    sender: UnboundedSender<bool>,
+    sender: UnboundedSender<Connectivity>,
     android_context: AndroidContext,
 ) -> Result<MonitorHandle, Error> {
     let sender = Arc::new(sender);
--- a/talpid-core/src/offline/linux.rs
+++ b/talpid-core/src/offline/linux.rs
@@ -4,7 +4,7 @@ use std::{
     sync::Arc,
 };
 use talpid_routing::{self, RouteManagerHandle};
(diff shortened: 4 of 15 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761)* status: all files reviewed, 11 unresolved discussions (waiting on @MarkusPettersson98)

___
*[`talpid-core/src/offline/android.rs` line 107 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt-hnJA_zguZljxHph:-Npt-hnJA_zguZljxHpi:b-4osfpu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/offline/android.rs#L107)):*
> ```XML
>         match self.get_is_connected() {
>             Ok(is_connected) => Connectivity::Status {
>                 ipv6: !is_connected,
> ```

I suspect that this should not be negated?
___
*[`talpid-core/src/offline/linux.rs` line 98 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt3uIaBXTKZ9PzW95l:-Npt3uIaBXTKZ9PzW95m:b3ry0dx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/offline/linux.rs#L98)):*
> ```XML
>             .get_destination_route(destination, fwmark)
>             .await
>             .map_err(Error::RouteManagerError)
> ```

Seems like we previously didn't fail if `get_destination_route` failed for IPv6. We treated it as there being no IPv6 connectivity. It is probably fine to change this behavior, but I'm not sure. I'm somewhat afraid that it will fail in some case and cause the daemon to think that there is IPv6 connectivity even though there is not.
___
*[`talpid-core/src/offline/macos.rs` line 86 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt4sNz20WV3WfcsXAn:-Npt4sNz20WV3WfcsXAo:bnnbzyo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/offline/macos.rs#L86)):*
> ```XML
> 
>                     let mut state = state.lock().unwrap();
>                     if *state != real_state && !real_state.is_offline() {
> ```

Should we keep the existing logic here? In particular, checking if `state != real_state` seems unnecessary, and I feel like it can make you think that they *might* be equal here, even though that would be a bug. Maybe `assert_ne!` make sense here?
___
*[`talpid-core/src/offline/macos.rs` line 86 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt5iJ3DA_t51gIJDnQ:-Npt5iJ45qyCypOJD07x:b9bs79b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/offline/macos.rs#L86)):*
> !real_state.is_offline(...)

Maybe we should add an `is_online` or `has_connectivity` method to avoid all of these negations.
___
*[`talpid-core/src/offline/mod.rs` line 36 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt0DoyBZomQihMwfhB:-Npt0DoyBZomQihMwfhC:b-fy4cup) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b6c9aba1c1b41b0cf994365159b27d3cf7b42470/talpid-core/src/offline/mod.rs#L36)):*
> ```XML
> 
> impl MonitorHandle {
>     pub async fn host_is_offline(&self) -> bool {
> ```

Method name is somewhat misleading now. Maybe this should be renamed to `connectivity()` or something like that?
___
*[`talpid-core/src/offline/mod.rs` line 76 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt12Mw3AGK-YIBRdjf:-Npt12Mw3AGK-YIBRdjg:b2jcfzc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/offline/mod.rs#L76)):*
> ```XML
>         None
>     } else {
>         let tx = forward_map(sender, |b| Connectivity::Status { ipv6: b });
> ```

It should be safe to change the type of the channel used by the monitor. We could skip this mapping then.
___
*[`talpid-core/src/tunnel_state_machine/disconnecting_state.rs` line 108 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npsy1MG1eS1nSUR7IEX:-Npsy1MG1eS1nSUR7IEY:b7dbebk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/tunnel_state_machine/disconnecting_state.rs#L108)):*
> ```XML
>                     AfterDisconnect::Block(reason)
>                 }
>                 Some(TunnelCommand::IsOffline(status)) => {
> ```

Nit: `IsOffline` is a bit wrong now.
___
*[`talpid-core/src/tunnel_state_machine/mod.rs` line 444 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt-KQX3xLL4Klx83FD:-Npt-KQX3xLL4Klx83FE:b-pspktf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b6c9aba1c1b41b0cf994365159b27d3cf7b42470/talpid-core/src/tunnel_state_machine/mod.rs#L444)):*
> ```XML
>     block_when_disconnected: bool,
>     /// True when the computer is known to be offline.
>     is_offline: bool,
> ```

What do you think about renaming this to `connectivity`, `connectivity_status`, or something like that?
___
*[`talpid-types/src/net/mod.rs` line 532 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt1FTHC1FrG7Svy7Ji:-Npt1FTHC1FrG7Svy7Jj:bj4z895) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-types/src/net/mod.rs#L532)):*
> ```XML
>     Status {
>         /// Whether IPv6 connectivity seems to be available on the host.
>         ipv6: bool,
> ```

Why is it that we have no status for IPv4 here?

In my opinion, if the connectivity monitor impl can't distinguish between IPv4 and IPv6, it should just set both to `true`. Since we want to err on the safe side and not get stuck in the offline state. Does that make sense?
___
*[`talpid-types/src/net/mod.rs` line 538 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt3H0uDhBC_fe8-ogD:-Npt3H0uDhBC_fe8-ogE:b-nyiqb6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-types/src/net/mod.rs#L538)):*
> ```XML
>     PresumeOnline,
>     /// Could not verify connectivity status of the host.
>     Unknown,
> ```

Seems to me like `PresumeOnline` and `Unknown` serve the same purpose. Maybe we could keep only `PresumeOnline`?
___
*[`talpid-types/src/net/mod.rs` line 544 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt2MZW3eoA0DgJCCTn:-Npt2MZW3eoA0DgJCCTo:b1d3k5n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-types/src/net/mod.rs#L544)):*
> ```XML
>     /// consider the host to be offline.
>     #[cfg(windows)]
>     Suspended,
> ```

We don't necessarily need to leak this information to the public type. Could we perhaps just "synthesize" `Status { ipv4: false, ipv6: false }` in the case where `suspended` is true?

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761)* status: 1 of 14 files reviewed, 11 unresolved discussions (waiting on @dlon)

___
*[`talpid-core/src/offline/android.rs` line 107 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt-hnJA_zguZljxHph:-NpxQbmm-oQL93V8f74B:b47nrny) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/offline/android.rs#L107)):*

I suspect that this should not be negated?
</blockquote></details>

I think it should, since the function we are returning from is `host_is_offline` :smile:
___
*[`talpid-core/src/offline/macos.rs` line 86 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt4sNz20WV3WfcsXAn:-NpxabOv3NrLXWs8Ych-:bqjcuek) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/offline/macos.rs#L86)):*

Should we keep the existing logic here? In particular, checking if `state != real_state` seems unnecessary, and I feel like it can make you think that they *might* be equal here, even though that would be a bug. Maybe `assert_ne!` make sense here?
</blockquote></details>

Valid point, but as you suggest we should probably add an `assert_ne!` here instead :blush: 
___
*[`talpid-core/src/offline/macos.rs` line 86 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt5iJ3DA_t51gIJDnQ:-Nq1lesU9nrTc5oGFrH_:bn4mlya) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/offline/macos.rs#L86)):*

> !real_state.is_offline(...)

Maybe we should add an `is_online` or `has_connectivity` method to avoid all of these negations.
</blockquote></details>

Sure, that's trivial :blush: 
___
*[`talpid-core/src/offline/mod.rs` line 36 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt0DoyBZomQihMwfhB:-Nq1pmbo92IQzbNgQ1_r:b31vmiu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b6c9aba1c1b41b0cf994365159b27d3cf7b42470/talpid-core/src/offline/mod.rs#L36)):*

Method name is somewhat misleading now. Maybe this should be renamed to `connectivity()` or something like that?
</blockquote></details>

Yes, that is probably better. The proper way to check for offline status is to use `Connectivity::is_offline` :blush: 
___
*[`talpid-core/src/tunnel_state_machine/disconnecting_state.rs` line 108 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npsy1MG1eS1nSUR7IEX:-Nq1scj2-SqFoCquu8BI:b-bb7621) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/tunnel_state_machine/disconnecting_state.rs#L108)):*

Nit: `IsOffline` is a bit wrong now.
</blockquote></details>

Yup, updating to `Connectivity`.
___
*[`talpid-core/src/tunnel_state_machine/mod.rs` line 444 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt-KQX3xLL4Klx83FD:-Nq1siPkDHIZVR-MW20s:ba0n9r3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b6c9aba1c1b41b0cf994365159b27d3cf7b42470/talpid-core/src/tunnel_state_machine/mod.rs#L444)):*

What do you think about renaming this to `connectivity`, `connectivity_status`, or something like that?
</blockquote></details>

Sure, updating to `connectivity`!
___
*[`talpid-types/src/net/mod.rs` line 538 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt3H0uDhBC_fe8-ogD:-Nq1uYhr3CgnEsA7_cHt:b-4wrp7g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-types/src/net/mod.rs#L538)):*

Seems to me like `PresumeOnline` and `Unknown` serve the same purpose. Maybe we could keep only `PresumeOnline`?
</blockquote></details>

We could definitely merge them into one. Keeping `PresumeOnline` :blush:
___
*[`talpid-types/src/net/mod.rs` line 544 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt2MZW3eoA0DgJCCTn:-Nq1wJNw2kvO6eCNAXR9:b-4y3zef) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-types/src/net/mod.rs#L544)):*

We don't necessarily need to leak this information to the public type. Could we perhaps just "synthesize" `Status { ipv4: false, ipv6: false }` in the case where `suspended` is true?
</blockquote></details>

I hid the `suspended` status in a private type in the `talpid_core::offline::windows` module, which maps to `Status { ipv4: false, ipv6: false }` if `suspended` is `true`.

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761)* status: 1 of 15 files reviewed, 11 unresolved discussions (waiting on @dlon)

___
*[`talpid-core/src/offline/mod.rs` line 76 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt12Mw3AGK-YIBRdjf:-Nq2a1091hFPYcMpDAqY:b4f45pe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/offline/mod.rs#L76)):*

It should be safe to change the type of the channel used by the monitor. We could skip this mapping then.
</blockquote></details>

Nice, that's cleaner!
___
*[`talpid-types/src/net/mod.rs` line 532 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt1FTHC1FrG7Svy7Ji:-Nq2_QtgDjR9a4n88qjs:b-xsi0z8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-types/src/net/mod.rs#L532)):*

Why is it that we have no status for IPv4 here?

In my opinion, if the connectivity monitor impl can't distinguish between IPv4 and IPv6, it should just set both to `true`. Since we want to err on the safe side and not get stuck in the offline state. Does that make sense?
</blockquote></details>

Updated Android-specific `Status` variant to have one field: `connected: bool`. Hopefully this should clear up some confusion.

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761)* status: 1 of 15 files reviewed, 11 unresolved discussions (waiting on @dlon)

___
*[`talpid-core/src/offline/linux.rs` line 98 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761#-Npt3uIaBXTKZ9PzW95l:-Nq2nBiV1w0Fu4wSkID1:b5o98nr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f5bf90f63d2b660f33d0085a122cc620d50ab44/talpid-core/src/offline/linux.rs#L98)):*

Seems like we previously didn't fail if `get_destination_route` failed for IPv6. We treated it as there being no IPv6 connectivity. It is probably fine to change this behavior, but I'm not sure. I'm somewhat afraid that it will fail in some case and cause the daemon to think that there is IPv6 connectivity even though there is not.
</blockquote></details>

If `get_destination_route` fails, the daemon will never interpret that as if IPv6 connectivity exists. What will happen in that case is that the offline monitor can't be sure if *any* connectivity exists, which means that it will likely propagate a `Connectivity::PresumeOnline` value to the daemon. If it were to propagate a `Connectivity::Status` where `ipv6: true` instead, that's a bug on its own :blush: 

What we could do here is to differentiate between `RouteManager` actor errors and the more low-level, Linux-specific routing integration errors. But to be honest I see no downsides to simply propagating an `Error` here, since we are not creating any false-positives.

dlon (approved):
> Reviewed 5 of 13 files at r3, 7 of 12 files at r4, 3 of 4 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5761)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5762, opened 2024-02-05
Fix location pin shown on map when disconnected

rablador:
Location pin is shown on map when connecting and then quickly disconnecting while map is animating. It should not be visible unless actually connected.


acb-mv (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5762)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5784, opened 2024-02-07
Add iOS test for connecting to API via bridges

niklasberglund:
This PR introduces the test `testAPIConnectionViaBridges` which blocks API access and makes sure the app can still successfully function. It also contains some more or less related improvements:
* The connectivity check for ad blocking test now checks whether host is connectable on port 80 instead of attempting to send HTTP requests
* Waiting for successful connection to relay in `testAdBlockingViaDNS` - sometimes the connectivity check happened before connected to relay
* Dropped `MULLVAD_` and `Mullvad` prefix before configuration and plist entries.

To test running `testAPIConnectionViaBridges` introduced in this PR you have to be connected to the `app-team-ios-tests` wifi at the office. It's a wifi running on the NUC where the firewall API is also running. The tests use the wifi and firewall API to block traffic.

You also need to configure `UITests.xcconfig`. Note that some variable names have changed(removed `MULLVAD_` prefix). `TEST_DEVICE_IDENTIFIER_UUID´ needs to have a value set. Can be generated with `uuidgen`. `FIREWALL_API_BASE_URL` is also introduced, but it's value should not be changed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784)

### DIFF
--- a/ios/Configurations/UITests.xcconfig.template
+++ b/ios/Configurations/UITests.xcconfig.template
@@ -1,19 +1,18 @@
-//
-//  MullvadVPNUITests.xcconfig
-//  MullvadVPN
-//
-//  Created by Niklas Berglund on 2024-01-10.
-//  Copyright © 2024 Mullvad VPN AB. All rights reserved.
-//
+#include "Base.xcconfig"
 
 // Pin code of the iOS device under test
-MULLVAD_IOS_DEVICE_PIN_CODE = 
+IOS_DEVICE_PIN_CODE = 
 
-// Ad serving domain used when testing ad blocking. Not that we are assuming there's an HTTP server running on the host.
-MULLVAD_AD_SERVING_DOMAIN = 
+// UUID to identify test runs. Should be unique per test device. Generate with for example uuidgen on macOS.
+TEST_DEVICE_IDENTIFIER_UUID = 
 
 // Mullvad accounts used by UI tests
-MULLVAD_NO_TIME_ACCOUNT_NUMBER = 
-MULLVAD_HAS_TIME_ACCOUNT_NUMBER = 
-MULLVAD_FIVE_WIREGUARD_KEYS_ACCOUNT_NUMBER = 
+NO_TIME_ACCOUNT_NUMBER = 
+HAS_TIME_ACCOUNT_NUMBER = 
+FIVE_WIREGUARD_KEYS_ACCOUNT_NUMBER = 
+
+// Ad serving domain used when testing ad blocking. Note that we are assuming there's an HTTP server running on the host.
+AD_SERVING_DOMAIN = vpnlist.to
  
+// Base URL for the firewall API, Note that // will be treated as a comment, therefor you need to insert a ${} between the slashes for example http:/${}/8.8.8.8
+FIREWALL_API_BASE_URL = http:/${}/8.8.8.8
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -38,10 +38,10 @@
 		06799AFC28F98EE300ACD94E /* AddressCache.swift in Sources */ = {isa = PBXBuildFile; fileRef = 06AC114128F8413A0037AF9A /* AddressCache.swift */; };
 		0697D6E728F01513007A9E99 /* TransportMonitor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0697D6E628F01513007A9E99 /* TransportMonitor.swift */; };
 		06AC116228F94C450037AF9A /* ApplicationConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58BFA5CB22A7CE1F00A6173D /* ApplicationConfiguration.swift */; };
-		44DD7D2D2B74E44A0005F67F /* QuantumResistanceSettings.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44DD7D2C2B74E44A0005F67F /* QuantumResistanceSettings.swift */; };
 		44DD7D242B6CFFD70005F67F /* StartTunnelOperationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44DD7D232B6CFFD70005F67F /* StartTunnelOperationTests.swift */; };
 		44DD7D272B6D18FB0005F67F /* MockTunnelInteractor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44DD7D262B6D18FB0005F67F /* MockTunnelInteractor.swift */; };
 		44DD7D292B7113CA0005F67F /* MockTunnel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44DD7D282B7113CA0005F67F /* MockTunnel.swift */; };
+		44DD7D2D2B74E44A0005F67F /* QuantumResistanceSettings.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44DD7D2C2B74E44A0005F67F /* QuantumResistanceSettings.swift */; };
 		5803B4B02940A47300C23744 /* TunnelConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5803B4AF2940A47300C23744 /* TunnelConfiguration.swift */; };
 		5803B4B22940A48700C23744 /* TunnelStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5803B4B12940A48700C23744 /* TunnelStore.swift */; };
 		5807E2C02432038B00F5FF30 /* String+Split.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5807E2BF2432038B00F5FF30 /* String+Split.swift */; };
@@ -590,13 +590,18 @@
 		852969362B4E9724007EAD4C /* AccessbilityIdentifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A0B311D2B303A0D004B12E0 /* AccessbilityIdentifier.swift */; };
 		8529693A2B4F0238007EAD4C /* TermsOfServicePage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 852969392B4F0238007EAD4C /* TermsOfServicePage.swift */; };
 		8529693C2B4F0257007EAD4C /* Alert.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8529693B2B4F0257007EAD4C /* Alert.swift */; };
+		85557B0E2B591B2600795FE1 /* FirewallAPIClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85557B0D2B591B2600795FE1 /* FirewallAPIClient.swift */; };
+		85557B102B59215F00795FE1 /* FirewallRule.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85557B0F2B59215F00795FE1 /* FirewallRule.swift */; };
+		85557B122B594FC900795FE1 /* ConnectivityTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85557B112B594FC900795FE1 /* ConnectivityTests.swift */; };
+		85557B142B5983CF00795FE1 /* MullvadAPIWrapper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85557B132B5983CF00795FE1 /* MullvadAPIWrapper.swift */; };
 		85557B162B5ABBBE00795FE1 /* XCUIElementQuery+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85557B152B5ABBBE00795FE1 /* XCUIElementQuery+Extensions.swift */; };
 		85557B1E2B5FB8C700795FE1 /* HeaderBar.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85557B1D2B5FB8C700795FE1 /* HeaderBar.swift */; };
 		85557B202B5FBBD700795FE1 /* AccountPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85557B1F2B5FBBD700795FE1 /* AccountPage.swift */; };
+		855D9F5B2B63E56B00D7C64D /* ProblemReportPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 855D9F5A2B63E56B00D7C64D /* ProblemReportPage.swift */; };
 		8590896C2B61763B003AF5F5 /* LoggedInWithoutTimeUITestCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 859089682B61763B003AF5F5 /* LoggedInWithoutTimeUITestCase.swift */; };
-		8590896D2B61763B003AF5F5 /* LoggedInWithTimeUITestCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 859089692B61763B003AF5F5 /* LoggedInWithTimeUITestCase.swift */; };
-		8590896E2B61763B003AF5F5 /* BaseUITestCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8590896A2B61763B003AF5F5 /* BaseUITestCase.swift */; };
 		8590896F2B61763B003AF5F5 /* LoggedOutUITestCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8590896B2B61763B003AF5F5 /* LoggedOutUITestCase.swift */; };
+		85D2B0B12B6BD32400DF9DA7 /* BaseUITestCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8590896A2B61763B003AF5F5 /* BaseUITestCase.swift */; };
+		85E3BDE52B70E18C00FA71FD /* Networking.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85E3BDE42B70E18C00FA71FD /* Networking.swift */; };
 		A900E9B82ACC5C2B00C95F67 /* AccountsProxy+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A900E9B72ACC5C2B00C95F67 /* AccountsProxy+Stubs.swift */; };
 		A900E9BA2ACC5D0600C95F67 /* RESTRequestExecutor+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A900E9B92ACC5D0600C95F67 /* RESTRequestExecutor+Stubs.swift */; };
 		A900E9BC2ACC609200C95F67 /* DevicesProxy+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A900E9BB2ACC609200C95F67 /* DevicesProxy+Stubs.swift */; };
@@ -1245,10 +1250,10 @@
 		06FAE67A28F83CA50033DD93 /* RESTDevicesProxy.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RESTDevicesProxy.swift; sourceTree = "<group>"; };
 		06FAE67B28F83CA50033DD93 /* REST.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = REST.swift; sourceTree = "<group>"; };
 		06FAE67D28F83CA50033DD93 /* RESTTransport.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = RESTTransport.swift; sourceTree = "<group>"; };
-		44DD7D2C2B74E44A0005F67F /* QuantumResistanceSettings.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = QuantumResistanceSettings.swift; sourceTree = "<group>"; };
 		44DD7D232B6CFFD70005F67F /* StartTunnelOperationTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StartTunnelOperationTests.swift; sourceTree = "<group>"; };
 		44DD7D262B6D18FB0005F67F /* MockTunnelInteractor.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MockTunnelInteractor.swift; sourceTree = "<group>"; };
 		44DD7D282B7113CA0005F67F /* MockTunnel.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MockTunnel.swift; sourceTree = "<group>"; };
+		44DD7D2C2B74E44A0005F67F /* QuantumResistanceSettings.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = QuantumResistanceSettings.swift; sourceTree = "<group>"; };
 		5802EBC42A8E44AC00E5CE4C /* AppRoutes.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppRoutes.swift; sourceTree = "<group>"; };
 		5802EBC62A8E457A00E5CE4C /* AppRouteProtocol.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppRouteProtocol.swift; sourceTree = "<group>"; };
 		5802EBC82A8E45BA00E5CE4C /* ApplicationRouterDelegate.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ApplicationRouterDelegate.swift; sourceTree = "<group>"; };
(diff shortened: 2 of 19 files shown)


buggmagnet (changes requested):
> Reviewed 17 of 18 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784)* status: all files reviewed, 11 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/Configurations/UITests.xcconfig.template` line 19 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7LQh4C6r-mN9qMBGd:-Nq7LQh4C6r-mN9qMBGe:b-129bb5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/Configurations/UITests.xcconfig.template#L19)):*
> ```
> 
> // Mullvad accounts used by UI tests
> NO_TIME_ACCOUNT_NUMBER = 
> ```

nit
Let's make sure we don't rename those too often
___
*[`ios/Configurations/UITests.xcconfig.template` line 24 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7LWL39vnrzvU5DR7F:-Nq7LWL39vnrzvU5DR7G:b-oahem3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/Configurations/UITests.xcconfig.template#L24)):*
> ```
>  
> // Base URL for the firewall API, Note that // will be treated as a comment, therefor you need to insert a ${} between the slashes for example http:/${}/8.8.8.8
> FIREWALL_API_BASE_URL = http:/${}/8.8.8.8
> ```

Why not have it as part of the Info.plist file then ? 
It seems hacky to have to do this 
___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 3367 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7tY8DCAJgTnqWGU_E:-Nq7tY8DCAJgTnqWGU_F:bqgxza3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPN.xcodeproj/project.pbxproj#L3367)):*
> ```
> 			children = (
> 				8518F6392B601910009EB113 /* Test base classes */,
> 				85557B0C2B591B0F00795FE1 /* Networking */,
> ```

Don't forget to sort by name in the organiser 
___
*[`ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift` line 26 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7kGZm0Y3cmCddqWKF:-Nq7kGZm0Y3cmCddqWKG:b-phd4vx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift#L26)):*
> ```Swift
> 
>     /// Get IP address of the iOS device under test
>     static func getIPAddress() -> String? {
> ```

This doesn't have to be static
___
*[`ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7nHa-CTW1HylKybVK:-Nq7nHa-CTW1HylKybVL:b85xsym) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift#L34)):*
> ```Swift
> 
>         // For each interface
>         for ptr in sequence(first: firstAddr, next: { $0.pointee.ifa_next }) {
> ```

nit
Pretty cool use of `sequence` to deal with linked lists ! :thumbsup: 
___
*[`ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift` line 36 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7lpHPBpC54fAxqcwv:-Nq7lpHPBpC54fAxqcww:b651isz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift#L36)):*
> ```Swift
>         for ptr in sequence(first: firstAddr, next: { $0.pointee.ifa_next }) {
>             let flags = Int32(ptr.pointee.ifa_flags)
>             let addr = ptr.pointee.ifa_addr.pointee
> ```

Let's use reasonable names for readability's sake, also we try not to use abbreviations as per the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/#use-terminology-well)
___
*[`ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift` line 109 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7mssZ1OUifa52CRJB:-Nq7mssZ1OUifa52CRJC:b-lagc20) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift#L109)):*
> ```Swift
> 
>             if waitResult != .completed {
>                 fatalError("Failed to create firewall rule - timeout")
> ```

Please avoid using `fatalError` 
Even if the firewall is down, some tests can still run, it doesn't make sense to abort an entire test run because it was not accessible.
Instead, we could make this a throwing function and throw an error here. 

The same comment applies for other uses of `fatalError` 
___
*[`ios/MullvadVPNUITests/Networking/FirewallRule.swift` line 12 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7nWSj1Syx3GNMplJo:-Nq7nWSj1Syx3GNMplJp:b-m7hpnu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallRule.swift#L12)):*
> ```Swift
> import XCTest
> 
> class FirewallRule {
> ```

This can be a struct instead, it has no logic, therefore it doesn't need to be a class
___
*[`ios/MullvadVPNUITests/Networking/FirewallRule.swift` line 21 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7nbwBA9FoBJYy5Wek:-Nq7nbwBA9FoBJYy5Wel:b-raqa57) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallRule.swift#L21)):*
> ```Swift
>     ///     - toIPAddress: Block traffic to this destination IP address.
>     ///     - protocols: Protocols which should be blocked. Valid values: `tcp`, `udp`, `icmp`. If none is specified all will be blocked.
>     private init(fromIPAddress: String, toIPAddress: String, protocols: [String]) {
> ```

For a nicer API use, we could define an `enum` for the protocols backed by a String, also because strong typing FTW
___
*[`ios/MullvadVPNUITests/Networking/FirewallRule.swift` line 28 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7o9ZQ8xWGwTNmd3gm:-Nq7o9ZQ8xWGwTNmd3gn:btzk4ei) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallRule.swift#L28)):*
> ```Swift
> 
>     /// Make a firewall rule blocking API access for the current device under test
>     public static func makeBlockAPIAccessFirewallRule() -> FirewallRule {
> ```

This also doesn't need to be static 
___
*[`ios/MullvadVPNUITests/Networking/MullvadAPIWrapper.swift` line 11 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7oHvw70HcV602M554:-Nq7oHvw70HcV602M555:b-9xb8n0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/MullvadAPIWrapper.swift#L11)):*
> ```Swift
> import Foundation
> 
> class MullvadAPIWrapper {
> ```

This can be a struct

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784)* status: 12 of 18 files reviewed, 11 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/Configurations/UITests.xcconfig.template` line 19 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7LQh4C6r-mN9qMBGd:-Nq7vA_t7qLU_r1ZEaVI:bqg73p8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/Configurations/UITests.xcconfig.template#L19)):*

nit
Let's make sure we don't rename those too often
</blockquote></details>

Agree. At first for some reason I thought other attributes had a `MULLVAD_` prefix. But later realised they don't so I corrected my mistake and removed all `MULLVAD_` prefixes.
___
*[`ios/Configurations/UITests.xcconfig.template` line 24 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7LWL39vnrzvU5DR7F:-Nq7y84k85exKLXZC5xk:b90m1tp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/Configurations/UITests.xcconfig.template#L24)):*

Why not have it as part of the Info.plist file then ? 
It seems hacky to have to do this
</blockquote></details>

It does look hacky. But if some configuration values are in `.xcconfig` files and some in `Info.plist` it might be difficult to keep track of where to modify what configuration values? Another way to get rid of the hackiness could be to not include `http://` in the configuration value, so it'd be something like `FIREWALL_API_HOST` rather than `FIREWALL_API_BASE_URL`.
___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 3367 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7tY8DCAJgTnqWGU_E:-NqC48cX-weA2M6d_fGp:b8wr7c4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPN.xcodeproj/project.pbxproj#L3367)):*

Don't forget to sort by name in the organiser
</blockquote></details>

Sorted by name now 👍
___
*[`ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift` line 26 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7kGZm0Y3cmCddqWKF:-NqC4SiG1OhVJ92H6r5b:b-ovooeu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift#L26)):*

This doesn't have to be static
</blockquote></details>

Discussed on Slack but not sure if we reached any conclusion on this. `getIPAddress()` is very standalone and not tied to an instance of the firewall API client.
___
*[`ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7nHa-CTW1HylKybVK:-NqC4qjB1u1ivl7iUcPD:b-km0y8z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift#L34)):*

nit
Pretty cool use of `sequence` to deal with linked lists ! :thumbsup:
</blockquote></details>

Credit goes to ChatGPT-4
___
*[`ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift` line 36 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7lpHPBpC54fAxqcwv:-Nq7vZvT0h6c5jjKtXaI:bk5f5lr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift#L36)):*

Let's use reasonable names for readability's sake, also we try not to use abbreviations as per the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/#use-terminology-well)
</blockquote></details>

Ah yes agree. ChatGPT helped with this function, I just made some small modifications but totally fortgot to fix the variable names. Have updated to more descriptive names.
___
*[`ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift` line 109 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7mssZ1OUifa52CRJB:-Nq7xLGR2MZGc6IhvzbI:bronwp4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift#L109)):*

Please avoid using `fatalError` 
Even if the firewall is down, some tests can still run, it doesn't make sense to abort an entire test run because it was not accessible.
Instead, we could make this a throwing function and throw an error here. 

The same comment applies for other uses of `fatalError`
</blockquote></details>

Yes you're right. The current test should be aborted but not the entire test run. `fatalError` is way too harsh, have updated the approach - using `XCTFail()` for descriptive messages and force unwrapping which will fail if something crucial for the test is missing.
___
*[`ios/MullvadVPNUITests/Networking/FirewallRule.swift` line 12 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7nWSj1Syx3GNMplJo:-NqCp5A5E8PSpIF3gM0K:bgdmh2f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallRule.swift#L12)):*

This can be a struct instead, it has no logic, therefore it doesn't need to be a class
</blockquote></details>

Very true. Converted to struct.
___
*[`ios/MullvadVPNUITests/Networking/FirewallRule.swift` line 21 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7nbwBA9FoBJYy5Wek:-NqCp89X3yTJA_gXGzG1:bubrzm4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallRule.swift#L21)):*

For a nicer API use, we could define an `enum` for the protocols backed by a String, also because strong typing FTW
</blockquote></details>

`enum`ified it 👍
___
*[`ios/MullvadVPNUITests/Networking/FirewallRule.swift` line 28 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7o9ZQ8xWGwTNmd3gm:-Nq80DcV8H27VoVRDvWQ:b-6rybhn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallRule.swift#L28)):*

This also doesn't need to be static
</blockquote></details>

Discussed on Slack
___
*[`ios/MullvadVPNUITests/Networking/MullvadAPIWrapper.swift` line 11 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7oHvw70HcV602M554:-NqCm1GFAgajIf58zrxU:b-6rybhn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/MullvadAPIWrapper.swift#L11)):*

This can be a struct
</blockquote></details>

Discussed on Slack

buggmagnet (changes requested):
> Reviewed 5 of 6 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784)* status: all files reviewed, 4 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift` line 26 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-Nq7kGZm0Y3cmCddqWKF:-NqRSvAb1mY9rz4AYmYy:b-59i7vp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/59c9deff601b136da9fe5b89265283658e78807f/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift#L26)):*

Discussed on Slack but not sure if we reached any conclusion on this. `getIPAddress()` is very standalone and not tied to an instance of the firewall API client.
</blockquote></details>

Static functions tend to be vulnerable to thread safety if they have to store a buffer locally, so usually it's bad practice to use static functions (and it makes code harder to refactor)

This one should be fine however.
___
*[`ios/MullvadVPNUITests/Networking/MullvadAPIWrapper.swift` line 26 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqRYy_l2z9ZlmZfPt3b:-NqRYy_l2z9ZlmZfPt3c:bo44hzh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfcb555419873f76b86ec4e1e013972ba7e2c49e/ios/MullvadVPNUITests/Networking/MullvadAPIWrapper.swift#L26)):*
> ```Swift
>     }
> 
>     public static func getAPIIPAddress() -> String? {
> ```

Instead of returning an optional, it's a better developper experience if this `throws` instead

If the function is to fail, the encompassign tests should abort because the thrown error.

```swift
    enum APIWrapperError: Error {
        case invalidIPFormat
    }
    
    public static func getAPIIPAddress2() throws -> String {
        guard let ipAddress = endpoint.components(separatedBy: ":").first else {
            throw APIWrapperError.invalidIPFormat
        }
        return ipAddress
    }

...

    public static func verifyCanAccessAPI2() throws {
        let apiIPAddress = try MullvadAPIWrapper.getAPIIPAddress2()
        /// .. the rest of things
    }


// In the test

func testMyTest() throws {
    try NetworkTester.verifyCanAccessAPI2()
```

The call site doesn't have to manually to do if checks, and automatically fails if an error is thrown.

This would also avoid all the force unwraps we are trying to avoid in the first place

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784)* status: 11 of 20 files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN.xcodeproj/xcshareddata/xcschemes/MullvadVPNUITests.xcscheme` line 68 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqWzub79OujqNIE3a-p:-NqWzub79OujqNIE3a-q:botd547) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8a8da0ccf5fdd3093f70cd254f84cf142ef2e4d6/ios/MullvadVPN.xcodeproj/xcshareddata/xcschemes/MullvadVPNUITests.xcscheme#L68)):*
> ```
>             ReferencedContainer = "container:MullvadVPN.xcodeproj">
>          </BuildableReference>
>       </MacroExpansion>
> ```

Snuck this small scheme change in to make the `MullvadVPNUITests` target runnable, this way it can be built without running tests.
___
*[`ios/MullvadVPNUITests/Networking/MullvadAPIWrapper.swift` line 26 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqRYy_l2z9ZlmZfPt3b:-NqWz-dp2qhKj9hwSvsI:b-aghbad) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bfcb555419873f76b86ec4e1e013972ba7e2c49e/ios/MullvadVPNUITests/Networking/MullvadAPIWrapper.swift#L26)):*

Instead of returning an optional, it's a better developper experience if this `throws` instead

If the function is to fail, the encompassign tests should abort because the thrown error.

```swift
    enum APIWrapperError: Error {
        case invalidIPFormat
    }
    
    public static func getAPIIPAddress2() throws -> String {
        guard let ipAddress = endpoint.components(separatedBy: ":").first else {
            throw APIWrapperError.invalidIPFormat
        }
        return ipAddress
    }

...

    public static func verifyCanAccessAPI2() throws {
        let apiIPAddress = try MullvadAPIWrapper.getAPIIPAddress2()
        /// .. the rest of things
    }


// In the test

func testMyTest() throws {
    try NetworkTester.verifyCanAccessAPI2()
```

The call site doesn't have to manually to do if checks, and automatically fails if an error is thrown.

This would also avoid all the force unwraps we are trying to avoid in the first place
</blockquote></details>

Throwing and not handling looks a bit more intentional 👍 Have updated to throw errors instead. The error throwing also highlighted the awkwardness of `getIPAddress()` in `FirewallAPIClient` so I did a small refactor - moved it to a class `Networking` where now the methods from `NetworkTester` also reside.

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784)* status: 11 of 20 files reviewed, 6 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/Configurations/UITests.xcconfig.template` line 8 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqX_LUc2x0X-55ny9PG:-NqX_LUc2x0X-55ny9PH:b-qtciqc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8a8da0ccf5fdd3093f70cd254f84cf142ef2e4d6/ios/Configurations/UITests.xcconfig.template#L8)):*
> ```
> //  Copyright © 2024 Mullvad VPN AB. All rights reserved.
> //
> ```

This template is missing the includes for the base config
___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 60 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqXZuHs8ivbzzSaxZc8:-NqXZuHs8ivbzzSaxZc9:b-4h96n4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8a8da0ccf5fdd3093f70cd254f84cf142ef2e4d6/ios/MullvadVPNUITests/Networking/Networking.swift#L60)):*
> ```Swift
>                         NI_NUMERICHOST
>                     )
>                     ipAddress = String(cString: hostname)
> ```

As we have seen during debug, this will output an iPv6 address that include the interface the IP address is scoped to (i.e. `fe80::1%ne0`)

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784)* status: 10 of 20 files reviewed, 6 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/Configurations/UITests.xcconfig.template` line 8 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqX_LUc2x0X-55ny9PG:-NqXk34b3CWT5YSi5d3G:bwvptn5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8a8da0ccf5fdd3093f70cd254f84cf142ef2e4d6/ios/Configurations/UITests.xcconfig.template#L8)):*

This template is missing the includes for the base config
</blockquote></details>

Added include to the template file. Also removed the header because other xcconfig files don't have it, and reordered attributes a bit so the attributes that do not need to be changed are in the bottom of the file.
___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 60 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqXZuHs8ivbzzSaxZc8:-NqXjpQbAFVQ-Gj3S3OK:b-1fhu7m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8a8da0ccf5fdd3093f70cd254f84cf142ef2e4d6/ios/MullvadVPNUITests/Networking/Networking.swift#L60)):*

As we have seen during debug, this will output an iPv6 address that include the interface the IP address is scoped to (i.e. `fe80::1%ne0`)
</blockquote></details>

The issue was that both  `sa_famil` `AF_INET`(IPv4) and `AF_INET6`(IPv6) was accepted. Have pushed a fix.

buggmagnet (commented):
> Reviewed 8 of 9 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784)* status: all files reviewed, 4 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPN.xcodeproj/xcshareddata/xcschemes/MullvadVPNUITests.xcscheme` line 68 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqWzub79OujqNIE3a-p:-Nqaqj1M1dVlzPHD68Jw:b-n310r4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8a8da0ccf5fdd3093f70cd254f84cf142ef2e4d6/ios/MullvadVPN.xcodeproj/xcshareddata/xcschemes/MullvadVPNUITests.xcscheme#L68)):*

Snuck this small scheme change in to make the `MullvadVPNUITests` target runnable, this way it can be built without running tests.
</blockquote></details>

neat !
___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 89 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqatmFf3EEnqQ0TPYzP:-NqatmFf3EEnqQ0TPYzQ:bwsac7q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e4c7560647726fda04a270bae79ff7d3892c9f75/ios/MullvadVPNUITests/Networking/Networking.swift#L89)):*
> ```Swift
>     private static func canConnectSocket(host: String, port: String) -> Bool {
>         let socketHost = NWEndpoint.Host(host)
>         let socketPort = NWEndpoint.Port(port)!
> ```

Let's make this not crashing at runtime
```swift
guard let socketPort = NWEndpoint.Port(port) else { return false }
```
___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 142 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqauJlF2Y30siFRsXi0:-NqauJlF2Y30siFRsXi1:brmqpu6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e4c7560647726fda04a270bae79ff7d3892c9f75/ios/MullvadVPNUITests/Networking/Networking.swift#L142)):*
> ```Swift
>     }
> 
>     /// Verify that an ad serving domain is NOT reachable by making sure a connection can be established on port 80
> ```

nit
can **not** be established

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784)* status: 19 of 20 files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 89 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqatmFf3EEnqQ0TPYzP:-NqaugePEUmZu17UhdtO:b-oampgb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e4c7560647726fda04a270bae79ff7d3892c9f75/ios/MullvadVPNUITests/Networking/Networking.swift#L89)):*

Let's make this not crashing at runtime
```swift
guard let socketPort = NWEndpoint.Port(port) else { return false }
```
</blockquote></details>

A silent fail could lead to false test results. The crashing is intentional. How about make it more descriptive by using `XCTFail`?

```
guard let socketPort \= NWEndpoint.Port(port) else { XCTFail("Failed to set port") }
```
___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 142 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqauJlF2Y30siFRsXi0:-NqavcXx2SYhaxADvbve:bapx8q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e4c7560647726fda04a270bae79ff7d3892c9f75/ios/MullvadVPNUITests/Networking/Networking.swift#L142)):*

nit
can **not** be established
</blockquote></details>

Ohh I flipped it. Have flipped it back now.

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784)* status: 18 of 20 files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 89 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784#-NqatmFf3EEnqQ0TPYzP:-Nqb4n5qA5MdJ41kTdBV:b-22j3fl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e4c7560647726fda04a270bae79ff7d3892c9f75/ios/MullvadVPNUITests/Networking/Networking.swift#L89)):*

A silent fail could lead to false test results. The crashing is intentional. How about make it more descriptive by using `XCTFail`?

```
guard let socketPort \= NWEndpoint.Port(port) else { XCTFail("Failed to set port") }
```
</blockquote></details>

Updated code to use `XCTUnwrap`

buggmagnet (approved):
> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5784)* status: all files reviewed, 2 unresolved discussions (waiting on @niklasberglund)

PR #5805, opened 2024-02-12
Fix MTU detection below 1280 with IPv6 on windows

Serock3:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5805)

### DIFF
--- a/talpid-windows/src/net.rs
+++ b/talpid-windows/src/net.rs
@@ -333,25 +333,12 @@ pub fn add_ip_address_for_interface(luid: NET_LUID_LH, address: IpAddr) -> Resul
 }
 
 /// Sets MTU on the specified network interface identified by `luid`.
-pub fn set_mtu(luid: NET_LUID_LH, mtu: u32, use_ipv6: bool) -> io::Result<()> {
-    let ip_families: &[AddressFamily] = if use_ipv6 {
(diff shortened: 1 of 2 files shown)


dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5805)* status: all files reviewed, 1 unresolved discussion (waiting on @Serock3)

___
*[`talpid-windows/src/net.rs` line 339 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5805#-NqSnte92AzpPA-aATgM:-NqSnte92AzpPA-aATgN:b-2znnif) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ef407fb8572dfd4749b4ef461ac4fc3616a7e7b6/talpid-windows/src/net.rs#L339)):*
> ```XML
>     let mut row = match get_ip_interface_entry(ip_family, &luid) {
>         Ok(row) => row,
>         Err(error) if error.raw_os_error() == Some(ERROR_NOT_FOUND as i32) => return Ok(()),
> ```

I think we should return this error. It makes little sense to ignore the error given that the caller is specifically asking to set the MTU for the given family. Also, both interfaces/families should always exist when we call this.

PR #5811, opened 2024-02-14
Fix off-by-one error when choosing access method candidates

MarkusPettersson98:
As the code comment says: `get_next_inner` will skip the `start` position, which means that by starting from the back of the available access methods we effectively look for candidates starting from index 0, i.e. the position of the `direct` access method. It's probably a good idea to not default users to bridges/proxies to reach the API.

Edit: The solution was found elsewhere. As it turns out, `mullvad-api` always forced a rotation of access methods whenever `MullvadRestHandle` was initialized, which means that `get_next_inner`/`select_next_active` was called twice at startup. This was subsequently solved by passing along an initial access method to the constructor of `MullvadRestHandle`, as is done with the tunnel state machine.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5811)

### DIFF
--- a/mullvad-api/src/bin/relay_list.rs
+++ b/mullvad-api/src/bin/relay_list.rs
@@ -11,11 +11,10 @@ async fn main() {
     let runtime = mullvad_api::Runtime::new(tokio::runtime::Handle::current())
         .expect("Failed to load runtime");
 
-    let relay_list_request = RelayListProxy::new(
-        runtime
-            .mullvad_rest_handle(ApiConnectionMode::Direct.into_repeat())
-            .await,
-    )
+    let relay_list_request = RelayListProxy::new(runtime.mullvad_rest_handle(
+        ApiConnectionMode::Direct,
+        ApiConnectionMode::Direct.into_repeat(),
+    ))
     .relay_list(None)
     .await;
 
--- a/mullvad-api/src/lib.rs
+++ b/mullvad-api/src/lib.rs
@@ -382,9 +382,10 @@ impl Runtime {
     }
 
     /// Creates a new request service and returns a handle to it.
-    async fn new_request_service<T: Stream<Item = ApiConnectionMode> + Unpin + Send + 'static>(
+    fn new_request_service<T: Stream<Item = ApiConnectionMode> + Unpin + Send + 'static>(
         &self,
         sni_hostname: Option<String>,
+        initial_connection_mode: ApiConnectionMode,
         proxy_provider: T,
         #[cfg(target_os = "android")] socket_bypass_tx: Option<mpsc::Sender<SocketBypassRequest>>,
     ) -> rest::RequestServiceHandle {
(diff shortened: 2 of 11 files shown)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5811)* status: all files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`mullvad-daemon/src/api.rs` line 250 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5811#-NqbR2NG1LyJRUqiPvXV:-NqbR2NG1LyJRUqiPvXW:bxxzfpw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cc45d53aa9a87ff66ad8319dbbe06b7377c8c6b1/mullvad-daemon/src/api.rs#L250)):*
> ```XML
>             // by starting from the back of the available access methods we
>             // effectively look for candidates starting from index 0.
>             access_method_settings.cardinality() - 1,
> ```

Correct me if I'm wrong, but this seems to end up selecting the last method in the list if a method should be added before `get_next` is called.

I think a less clever solution might be better. Maybe by not setting an initial method at all, or not relying on `get_next` to select the "first" method.

But we could use this as a hotfix if we need one.

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5811)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-daemon/src/api.rs` line 250 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5811#-NqbR2NG1LyJRUqiPvXV:-NqcIpToDVkmBpPykNjB:bs9ikg6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cc45d53aa9a87ff66ad8319dbbe06b7377c8c6b1/mullvad-daemon/src/api.rs#L250)):*

Correct me if I'm wrong, but this seems to end up selecting the last method in the list if a method should be added before `get_next` is called.

I think a less clever solution might be better. Maybe by not setting an initial method at all, or not relying on `get_next` to select the "first" method.

But we could use this as a hotfix if we need one.
</blockquote></details>

Turns out that the solution in this module was very straightforward - the turned out to be somewhere else! See my updated PR comment for context :blush: 

Refactored `get_next_inner` (now `select_next_active`) to be more declarative and start looking for candidates starting from the `start` position, otherwise not much else has changed here.

dlon (approved):
> Reviewed 11 of 11 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5811)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

dlon (approved):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5811)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5852, opened 2024-02-23
Optimize map contour lines and fix long line bug

faern:
This PR implements the changes needed to match with https://github.com/mullvad/globe-mesh-builder/pull/1. The description for this change is basically the same, so I'll copy it over:

This PR does three things to the contour lines between geographical entities:

* Subdivide long contour edges just like we do with triangles. Makes long lines not take a shortcut through earth and end up not visible
* Removed duplicate contour line segments. Keep a `HashSet` with seen edges and skip already drawn edges
* Order the contour line index buffer so it can be drawn as a single continuous line with `GL_LINE_STRIP` instead of the current `GL_LINES`. This reduces the size of the `land_contour_indices.bin` file from `873 kb` bytes to `392 kb`.

If we had access to OpenGL 3.1 we could draw the line strips much nicer with a single draw call thanks to `glPrimitiveRestartIndex`. But sadly this is not available to us. Instead this uses a hack where we continue the continuous line by drawing underneath the earths crust to the new position and pop back up there.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5852)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5852)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[`android/lib/map/src/main/kotlin/net/mullvad/mullvadvpn/lib/map/internal/shapes/Globe.kt` line 156 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5852#-NrLQSHzAvoGrovh23bd:-NrLQSHzAvoGrovh23be:b-sdc9ct) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b67c62d17af383608524153d0a880dacf50e6a20/android/lib/map/src/main/kotlin/net/mullvad/mullvadvpn/lib/map/internal/shapes/Globe.kt#L156)):*
> ```Kotlin
> 
>     companion object {
>         private const val LAND_OCEAN_SCALE_FACTOR = 0.99999f
> ```

We can draw the land triangles closer to the contour lines now. This is because the contour lines are subdivided into shorter segments, staying closer to the sphere radius of 1.0. As such they don't go through the crust as easily and the crust can move closer!

Rawa (approved):
Nice work! :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5852#-:-NrZidE9-1Zy1IqsCgnR:bnahwna)* status: all files reviewed, 1 unresolved discussion

raksooo (approved):
:lgtm:

> Reviewed 1 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5852#-:-NrdZcq3-s57mUOpUVk_:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

PR #5863, opened 2024-02-26
Add unit tests to MTU detection

Serock3:
Split MTU detection into an inner pure function `max_ping_sized`
and an io-dependent outer function `detect_mtu`. Add unit and prop-testing to
`max_ping_sized`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863)

### DIFF
--- a/talpid-wireguard/Cargo.toml
+++ b/talpid-wireguard/Cargo.toml
@@ -82,3 +82,4 @@ features = [
 
 [dev-dependencies]
 proptest = "1.4"
+tokio = { workspace = true, features = ["time", "test-util"] }
--- a/talpid-wireguard/src/mtu_detection.rs
+++ b/talpid-wireguard/src/mtu_detection.rs
@@ -1,8 +1,9 @@
 use std::{io, net::IpAddr, time::Duration};
 
-use futures::{future, stream::FuturesUnordered, TryStreamExt};
+use futures::{future, stream::FuturesUnordered, Stream, TryStreamExt};
 use surge_ping::{Client, Config, PingIdentifier, PingSequence, SurgeError};
 use talpid_tunnel::{ICMP_HEADER_SIZE, IPV4_HEADER_SIZE, MIN_IPV4_MTU};
+use tokio::pin;
 use tokio_stream::StreamExt;
 
 #[derive(thiserror::Error, Debug)]
@@ -11,23 +12,31 @@ pub enum Error {
     #[error("Failed to set MTU on the active tunnel")]
     SetMtu(#[source] io::Error),
 
-    /// Failed to set MTU
-    #[error("Failed to detect MTU because every ping was dropped.")]
+    /// Failed to detect MTU because every ping was dropped
+    #[error("Failed to detect MTU because all pings timed out.")]
     MtuDetectionAllDropped,
 
-    /// Failed to set MTU
+    /// Failed to detect MTU because of unexpected ping error
     #[error("Failed to detect MTU because of unexpected ping error.")]
-    MtuDetectionPing(#[source] surge_ping::SurgeError),
+    MtuDetectionUnexpected(#[source] surge_ping::SurgeError),
 
-    /// Failed to set MTU
+    /// Failed to detect MTU because of an IO error when setting up the ping socket
     #[error("Failed to detect MTU because of an IO error when setting up the ping socket.")]
     MtuDetectionSetupSocket(#[source] io::Error),
 
-    /// Failed to set MTU
+    /// Failed to set buffer size
     #[cfg(target_os = "macos")]
     #[error("Failed to set buffer size")]
     MtuSetBufferSize(#[source] nix::Error),
 }
+
+/// Max time to wait for any ping, when this expires, we give up and throw an error.
+const PING_TIMEOUT: Duration = Duration::from_secs(10);
+/// Max time to wait after the first ping arrives. Every ping after this timeout is
+/// considered dropped, so we return the largest collected packet size.
+const PING_OFFSET_TIMEOUT: Duration = Duration::from_secs(2);
+const MTU_STEP_SIZE: u16 = 20;
+
 /// Verify that the current MTU doesn't cause dropped packets, otherwise lower it to the
 /// largest value which doesn't.
 ///
@@ -89,14 +98,7 @@ async fn detect_mtu(
     #[cfg(any(target_os = "macos", target_os = "linux"))] iface_name: String,
     current_mtu: u16,
 ) -> Result<u16, Error> {
-    /// Max time to wait for any ping, when this expires, we give up and throw an error.
-    const PING_TIMEOUT: Duration = Duration::from_secs(10);
-    /// Max time to wait after the first ping arrives. Every ping after this timeout is
-    /// considered dropped, so we return the largest collected packet size.
-    const PING_OFFSET_TIMEOUT: Duration = Duration::from_secs(2);
-
-    let step_size = 20;
-    let linspace = mtu_spacing(MIN_IPV4_MTU, current_mtu, step_size);
+    let linspace = mtu_spacing(MIN_IPV4_MTU, current_mtu, MTU_STEP_SIZE);
 
     let config_builder = Config::builder().kind(surge_ping::ICMP::V4);
     #[cfg(any(target_os = "macos", target_os = "linux"))]
@@ -107,7 +109,7 @@ async fn detect_mtu(
     // data we are sending here. The consequence will be dropped packets causing the MTU
     // detection to set a low value. Here we manually increase this value, which fixes
     // the problem.
-    // TODO: Make sure this fix is not needed for any other target OS
+    // NOTE: If pings drop on other unix platforms too, then enable this fix for them
     #[cfg(target_os = "macos")]
     {
         use nix::sys::socket::{setsockopt, sockopt};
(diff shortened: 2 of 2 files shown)


dlon (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-:-Nr_eOyt1Z1eiBuKt4fO:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @Serock3)

___
*[`talpid-wireguard/src/mtu_detection.rs` line 115 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_W4-u7TMTinJIN0Cq:-Nr_W4-u7TMTinJIN0Cr:b-n8cit4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L115)):*
> ```XML
>     // detection to set a low value. Here we manually increase this value, which fixes
>     // the problem.
>     // TODO: Make sure this fix is not needed for any other target OS
> ```

Should probably be moved to linear rather than be left as a comment here
___
*[`talpid-wireguard/src/mtu_detection.rs` line 148 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_VVrs5HHis6DD6vXs:-Nr_VVrs5HHis6DD6vXt:bhxu51p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L148)):*
> ```XML
> 
>                 // Validate the received ping response
>                 {
> ```

Nit: Is there a point to placing this inside a block?
___
*[`talpid-wireguard/src/mtu_detection.rs` line 170 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_a2Y5DswSZJRB7v6K:-Nr_a2Y5DswSZJRB7v6L:brbnvxp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L170)):*
> ```XML
> 
> /// Consumes a stream of pings, and returns the largest packet size within a given timeout from the
> /// first ping response. Short circuits on errors.
> ```

Might be worth clarifying that this only works as expected with `FuturesUnordered`? Or even have it only accept  `FutureUnordered`. Unless that complicates the tests.
___
*[`talpid-wireguard/src/mtu_detection.rs` line 240 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_WW2X-PtxxLgEV_dX:-Nr_WW2X-PtxxLgEV_dY:bfxc4cm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L240)):*
> ```XML
>     async fn all_pings_ok() {
>         let pings = stream::iter((0..=100).rev().map(Ok));
>         let max = max_ping_size(pings, Duration::from_millis(10))
> ```

Might it be possible to advance the timer here without actually waiting at all? E.g. using https://docs.rs/tokio/latest/tokio/time/fn.advance.html
___
*[`talpid-wireguard/src/mtu_detection.rs` line 260 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_YF_R3ryd7Q7eW1B2:-Nr_YF_R3ryd7Q7eW1B3:b-w76o2e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L260)):*
> ```XML
>     }
> 
>     /// The [`PING_OFFSET_TIMEOUT`] is counted from the return of the first ping, not from the
> ```

Is this actually what is being tested? Does it not need to check whether the timeout equals `offset_timeout` + timeout of the first ping to accomplish that?
___
*[`talpid-wireguard/src/mtu_detection.rs` line 310 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_cqClE0fJZAVbW8h-:-Nr_ax4B3OxKiyO-02kH:b-yajt0x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L310)):*
> ```XML
>     /// return an error instead of trusting result.
>     #[tokio::test]
>     async fn max_timeout_error() {
> ```

Should the test not arguably make sure that this cannot happen instead?

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863)* status: all files reviewed, 4 unresolved discussions (waiting on @dlon)

___
*[`talpid-wireguard/src/mtu_detection.rs` line 115 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_W4-u7TMTinJIN0Cq:-NraE0VI6-tyOPK3C-7A:btzo9t9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L115)):*

Should probably be moved to linear rather than be left as a comment here
</blockquote></details>

You're probably right. Though we have tested it on all platforms without finding any issues with this so far. I guess I left the note just to make it easier to debug in case similar issues turn up in the future.
___
*[`talpid-wireguard/src/mtu_detection.rs` line 148 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_VVrs5HHis6DD6vXs:-Nr_f6fbAr9xwkDPMvf7:b-8agbtq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L148)):*

Nit: Is there a point to placing this inside a block?
</blockquote></details>

No. Just thought it may be more readable that it doesn't affect the return value.
___
*[`talpid-wireguard/src/mtu_detection.rs` line 170 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_a2Y5DswSZJRB7v6K:-NraEW2Q8mXJKOl_KK3E:boya8mr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L170)):*

Might be worth clarifying that this only works as expected with `FuturesUnordered`? Or even have it only accept  `FutureUnordered`. Unless that complicates the tests.
</blockquote></details>

I see your point. The function has some intuitive behavior for arbitrary streams when it comes to the timeout. It may be possible to re-write the tests to also use `FutureUnordered` directly, but I'm not sure it's worth the effort.
___
*[`talpid-wireguard/src/mtu_detection.rs` line 240 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_WW2X-PtxxLgEV_dX:-NraFtLpDeX0untBV5nN:b-72ec40) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L240)):*

Might it be possible to advance the timer here without actually waiting at all? E.g. using https://docs.rs/tokio/latest/tokio/time/fn.advance.html
</blockquote></details>

Good find! I thought that there should be some way to mock the timeout but couldn't find anything. I will look into using that function.
___
*[`talpid-wireguard/src/mtu_detection.rs` line 260 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_YF_R3ryd7Q7eW1B2:-NraGh9oCamzos9XgNCa:b-1ykwz3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L260)):*

Is this actually what is being tested? Does it not need to check whether the timeout equals `offset_timeout` + timeout of the first ping to accomplish that?
</blockquote></details>

I would say so. It sets the `ping_offset_timeout` param to 5 ms and then sends all the pings in a single burst after 10 ms. If the timeout counter would have started from the fn call they would have timed out, but since they arrive close to each other in time they do not.
___
*[`talpid-wireguard/src/mtu_detection.rs` line 310 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_cqClE0fJZAVbW8h-:-NraHR4R4ZAKvHYNwXSG:b-52xwyc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L310)):*

Should the test not arguably make sure that this cannot happen instead?
</blockquote></details>

Not sure how I would prevent it from happening. This edge-case would trigger if e.g. half of the pings arrive after 9 s and the other half are delayed more than 10 s. In that case you should not pick the largest packet size of the first half of ping responses, but instead abort. Since I use build in timeout of `surge_ping` for the long `PING_TIMEOUT` I cannot ignore this timeout.

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @dlon)

___
*[`talpid-wireguard/src/mtu_detection.rs` line 170 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_a2Y5DswSZJRB7v6K:-NreWYEjBgebZa2ktOhS:bph37c5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L170)):*

I see your point. The function has some intuitive behavior for arbitrary streams when it comes to the timeout. It may be possible to re-write the tests to also use `FutureUnordered` directly, but I'm not sure it's worth the effort.
</blockquote></details>

I made it only accept `FutureUnordered`
___
*[`talpid-wireguard/src/mtu_detection.rs` line 240 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_WW2X-PtxxLgEV_dX:-NreWdQc37RgcRJki9JQ:b-9j72ww) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L240)):*

Good find! I thought that there should be some way to mock the timeout but couldn't find anything. I will look into using that function.
</blockquote></details>

Done. Turned out really well I think!
___
*[`talpid-wireguard/src/mtu_detection.rs` line 260 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5863#-Nr_YF_R3ryd7Q7eW1B2:-NreXCUUBORVzFEccNkK:b-mbb1ts) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d3169a6275a2a17d95535730f6aef0ced0d0b11/talpid-wireguard/src/mtu_detection.rs#L260)):*

I would say so. It sets the `ping_offset_timeout` param to 5 ms and then sends all the pings in a single burst after 10 ms. If the timeout counter would have started from the fn call they would have timed out, but since they arrive close to each other in time they do not.
</blockquote></details>

See the new implementation after the above changes

PR #5873, opened 2024-02-28
Rework side effects

Rawa:
Summary: This PR fixes some sideeffect issues we've seen. E.g when Navigation to the Account view, and OutOfTime view overriding and showing instead. It does this by reworking the OutOfTimeUseCase and rewriting how we handle side effects.

Details:
We have different ways side effects happens in our app here are some unique examples:
- User action caused effect (user does an action, which then may or may not result in side effect, e.g conditional navigation)
- Delayed effect (user does an action, but may result in a delayed effect, e.g in the Login screen, user logs in but then navigation happens after 1 seconds of us showing success screen)
- External effect (External component causes a side effect to happen, e.g user happens to become OutOfTime)

These types of side effects needs to be taken into consideration and discussions made on a case by case basis. This PR changes so we no longer buffer old side effects and then at a later time accidentally executes them. We then have to choose when executing a side effect if we want it to be blocking (`send`) or non-blocking as an attempt (`trySend`)

The biggest changes to this PR is:

`Connect Screen`, `Welcome`, `OutOfTime`, we now expect that, when subscribing to side effects, all side effects that can happen should emit at subscription. Previously the subscription of e.g OutOfTime happened on view model init, this is now moved to the subscription of `uiSideEffects`. This means if we stop subscribe (e.g because we are navigating away, the event will still be able to be consumed once we resume the view (e.g come back from account view)

`LoginScreen`: We no longer depend on the value being buffered but rather `send` blocking.

`Splash Screen`: We now re-evaluate on each Resume which destination we should go to. Edge case scenario, if user opens app, quickly closes it, come back it to a later stage, event though we never left the splash screen the destination was already evaluated and then invoked when user already left the app. Now we will re-evaluate when the splash screen is opened.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)

### DIFF
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/dialog/DnsDialog.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/dialog/DnsDialog.kt
@@ -8,7 +8,6 @@ import androidx.compose.material3.AlertDialog
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
@@ -22,6 +21,7 @@ import net.mullvad.mullvadvpn.R
 import net.mullvad.mullvadvpn.compose.button.NegativeButton
 import net.mullvad.mullvadvpn.compose.button.PrimaryButton
 import net.mullvad.mullvadvpn.compose.textfield.DnsTextField
+import net.mullvad.mullvadvpn.compose.util.LaunchedEffectCollect
 import net.mullvad.mullvadvpn.lib.theme.AppTheme
 import net.mullvad.mullvadvpn.lib.theme.Dimens
 import net.mullvad.mullvadvpn.viewmodel.DnsDialogSideEffect
@@ -100,11 +100,9 @@ fun DnsDialog(
     val viewModel =
         koinViewModel<DnsDialogViewModel>(parameters = { parametersOf(initialValue, index) })
 
-    LaunchedEffect(Unit) {
-        viewModel.uiSideEffect.collect {
-            when (it) {
-                DnsDialogSideEffect.Complete -> resultNavigator.navigateBack(result = true)
-            }
+    LaunchedEffectCollect(viewModel.uiSideEffect) {
+        when (it) {
+            DnsDialogSideEffect.Complete -> resultNavigator.navigateBack(result = true)
         }
     }
     val state by viewModel.uiState.collectAsStateWithLifecycle()
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/dialog/MtuDialog.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/dialog/MtuDialog.kt
@@ -8,7 +8,6 @@ import androidx.compose.material3.AlertDialog
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
@@ -21,6 +20,7 @@ import com.ramcosta.composedestinations.spec.DestinationStyle
 import net.mullvad.mullvadvpn.R
 import net.mullvad.mullvadvpn.compose.button.PrimaryButton
 import net.mullvad.mullvadvpn.compose.textfield.MtuTextField
+import net.mullvad.mullvadvpn.compose.util.LaunchedEffectCollect
 import net.mullvad.mullvadvpn.constant.MTU_MAX_VALUE
 import net.mullvad.mullvadvpn.constant.MTU_MIN_VALUE
 import net.mullvad.mullvadvpn.lib.theme.AppTheme
@@ -42,11 +42,9 @@ private fun PreviewMtuDialog() {
 fun MtuDialog(mtuInitial: Int?, navigator: DestinationsNavigator) {
     val viewModel = koinViewModel<MtuDialogViewModel>()
 
-    LaunchedEffect(Unit) {
-        viewModel.uiSideEffect.collect {
-            when (it) {
-                MtuDialogSideEffect.Complete -> navigator.navigateUp()
-            }
+    LaunchedEffectCollect(viewModel.uiSideEffect) {
+        when (it) {
+            MtuDialogSideEffect.Complete -> navigator.navigateUp()
         }
     }
     MtuDialog(
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/dialog/payment/PaymentDialog.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/dialog/payment/PaymentDialog.kt
@@ -20,6 +20,7 @@ import com.ramcosta.composedestinations.spec.DestinationStyle
 import net.mullvad.mullvadvpn.R
 import net.mullvad.mullvadvpn.compose.button.PrimaryButton
 import net.mullvad.mullvadvpn.compose.component.MullvadCircularProgressIndicatorMedium
+import net.mullvad.mullvadvpn.compose.util.LaunchedEffectCollect
 import net.mullvad.mullvadvpn.lib.payment.model.ProductId
 import net.mullvad.mullvadvpn.lib.theme.AppTheme
 import net.mullvad.mullvadvpn.lib.theme.color.AlphaDescription
@@ -125,12 +126,10 @@ fun Payment(productId: ProductId, resultBackNavigator: ResultBackNavigator<Boole
     val vm = koinViewModel<PaymentViewModel>()
     val state by vm.uiState.collectAsStateWithLifecycle()
 
-    LaunchedEffect(Unit) {
-        vm.uiSideEffect.collect {
-            when (it) {
-                is PaymentUiSideEffect.PaymentCancelled ->
-                    resultBackNavigator.navigateBack(result = false)
-            }
+    LaunchedEffectCollect(vm.uiSideEffect) {
+        when (it) {
+            is PaymentUiSideEffect.PaymentCancelled ->
+                resultBackNavigator.navigateBack(result = false)
         }
     }
 
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/AccountScreen.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/AccountScreen.kt
@@ -16,7 +16,6 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.SnackbarHostState
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
@@ -55,6 +54,7 @@ import net.mullvad.mullvadvpn.compose.destinations.RedeemVoucherDestination
 import net.mullvad.mullvadvpn.compose.destinations.VerificationPendingDialogDestination
 import net.mullvad.mullvadvpn.compose.state.PaymentState
 import net.mullvad.mullvadvpn.compose.transitions.SlideInFromBottomTransition
+import net.mullvad.mullvadvpn.compose.util.LaunchedEffectCollect
 import net.mullvad.mullvadvpn.compose.util.SecureScreenWhileInView
 import net.mullvad.mullvadvpn.lib.common.util.openAccountPageInBrowser
 import net.mullvad.mullvadvpn.lib.payment.model.PaymentProduct
@@ -171,22 +171,20 @@ fun AccountScreen(
     val clipboardManager = LocalClipboardManager.current
     val snackbarHostState = remember { SnackbarHostState() }
     val copyTextString = stringResource(id = R.string.copied_mullvad_account_number)
-    LaunchedEffect(Unit) {
-        uiSideEffect.collect { uiSideEffect ->
-            when (uiSideEffect) {
-                AccountViewModel.UiSideEffect.NavigateToLogin -> navigateToLogin()
-                is AccountViewModel.UiSideEffect.OpenAccountManagementPageInBrowser ->
-                    context.openAccountPageInBrowser(uiSideEffect.token)
-                is AccountViewModel.UiSideEffect.CopyAccountNumber ->
-                    launch {
-                        clipboardManager.setText(AnnotatedString(uiSideEffect.accountNumber))
+    LaunchedEffectCollect(uiSideEffect) { sideEffect ->
+        when (sideEffect) {
+            AccountViewModel.UiSideEffect.NavigateToLogin -> navigateToLogin()
+            is AccountViewModel.UiSideEffect.OpenAccountManagementPageInBrowser ->
+                context.openAccountPageInBrowser(sideEffect.token)
+            is AccountViewModel.UiSideEffect.CopyAccountNumber ->
+                launch {
+                    clipboardManager.setText(AnnotatedString(sideEffect.accountNumber))
 
-                        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {
-                            snackbarHostState.currentSnackbarData?.dismiss()
-                            snackbarHostState.showSnackbar(message = copyTextString)
-                        }
+                    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {
+                        snackbarHostState.currentSnackbarData?.dismiss()
+                        snackbarHostState.showSnackbar(message = copyTextString)
                     }
-            }
+                }
         }
     }
 
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt
@@ -38,8 +38,8 @@ import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
-import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import com.ramcosta.composedestinations.annotation.Destination
 import com.ramcosta.composedestinations.navigation.DestinationsNavigator
 import com.ramcosta.composedestinations.navigation.popUpTo
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt
@@ -66,6 +66,7 @@ import net.mullvad.mullvadvpn.compose.test.RECONNECT_BUTTON_TEST_TAG
 import net.mullvad.mullvadvpn.compose.test.SCROLLABLE_COLUMN_TEST_TAG
 import net.mullvad.mullvadvpn.compose.test.SELECT_LOCATION_BUTTON_TEST_TAG
 import net.mullvad.mullvadvpn.compose.transitions.HomeTransition
+import net.mullvad.mullvadvpn.compose.util.CollectSideEffectWithLifecycle
 import net.mullvad.mullvadvpn.constant.SECURE_ZOOM
 import net.mullvad.mullvadvpn.constant.SECURE_ZOOM_ANIMATION_MILLIS
 import net.mullvad.mullvadvpn.constant.UNSECURE_ZOOM
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt
@@ -21,7 +21,6 @@ import androidx.compose.foundation.verticalScroll
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableFloatStateOf
 import androidx.compose.runtime.mutableLongStateOf
@@ -40,6 +39,7 @@ import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.lifecycle.Lifecycle
 import com.ramcosta.composedestinations.annotation.Destination
 import com.ramcosta.composedestinations.navigation.DestinationsNavigator
 import com.ramcosta.composedestinations.navigation.popUpTo
@@ -112,27 +112,28 @@ fun Connect(navigator: DestinationsNavigator) {
     val state by connectViewModel.uiState.collectAsStateWithLifecycle()
 
     val context = LocalContext.current
-    LaunchedEffect(key1 = Unit) {
-        connectViewModel.uiSideEffect.collect { uiSideEffect ->
-            when (uiSideEffect) {
-                is ConnectViewModel.UiSideEffect.OpenAccountManagementPageInBrowser -> {
-                    context.openAccountPageInBrowser(uiSideEffect.token)
-                }
-                is ConnectViewModel.UiSideEffect.OutOfTime -> {
-                    navigator.navigate(OutOfTimeDestination) {
-                        launchSingleTop = true
-                        popUpTo(NavGraphs.root) { inclusive = true }
-                    }
-                }
-                ConnectViewModel.UiSideEffect.RevokedDevice -> {
-                    navigator.navigate(DeviceRevokedDestination) {
-                        launchSingleTop = true
-                        popUpTo(NavGraphs.root) { inclusive = true }
-                    }
-                }
+
+    CollectSideEffectWithLifecycle(
+        connectViewModel.uiSideEffect,
+        minActiveState = Lifecycle.State.RESUMED
+    ) { sideEffect ->
+        when (sideEffect) {
+            is ConnectViewModel.UiSideEffect.OpenAccountManagementPageInBrowser -> {
+                context.openAccountPageInBrowser(sideEffect.token)
             }
+            is ConnectViewModel.UiSideEffect.OutOfTime ->
+                navigator.navigate(OutOfTimeDestination, true) {
+                    launchSingleTop = true
+                    popUpTo(NavGraphs.root) { inclusive = true }
+                }
+            ConnectViewModel.UiSideEffect.RevokedDevice ->
+                navigator.navigate(DeviceRevokedDestination, true) {
+                    launchSingleTop = true
+                    popUpTo(NavGraphs.root) { inclusive = true }
+                }
         }
     }
+
     ConnectScreen(
         state = state,
         onDisconnectClick = connectViewModel::onDisconnectClick,
@@ -140,7 +141,7 @@ fun Connect(navigator: DestinationsNavigator) {
         onConnectClick = connectViewModel::onConnectClick,
         onCancelClick = connectViewModel::onCancelClick,
         onSwitchLocationClick = {
-            navigator.navigate(SelectLocationDestination) { launchSingleTop = true }
+            navigator.navigate(SelectLocationDestination, true) { launchSingleTop = true }
         },
         onUpdateVersionClick = {
             val intent =
@@ -156,8 +157,12 @@ fun Connect(navigator: DestinationsNavigator) {
             context.startActivity(intent)
         },
         onManageAccountClick = connectViewModel::onManageAccountClick,
-        onSettingsClick = { navigator.navigate(SettingsDestination) { launchSingleTop = true } },
-        onAccountClick = { navigator.navigate(AccountDestination) { launchSingleTop = true } },
+        onSettingsClick = {
+            navigator.navigate(SettingsDestination, true) { launchSingleTop = true }
+        },
+        onAccountClick = {
+            navigator.navigate(AccountDestination, true) { launchSingleTop = true }
+        },
         onDismissNewDeviceClick = connectViewModel::dismissNewDeviceNotification,
     )
 }
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/DeviceRevokedScreen.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/DeviceRevokedScreen.kt
@@ -9,7 +9,6 @@ import androidx.compose.foundation.layout.size
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
@@ -32,6 +31,7 @@ import net.mullvad.mullvadvpn.compose.component.ScaffoldWithTopBar
 import net.mullvad.mullvadvpn.compose.destinations.LoginDestination
 import net.mullvad.mullvadvpn.compose.destinations.SettingsDestination
 import net.mullvad.mullvadvpn.compose.state.DeviceRevokedUiState
+import net.mullvad.mullvadvpn.compose.util.LaunchedEffectCollect
 import net.mullvad.mullvadvpn.lib.theme.AppTheme
 import net.mullvad.mullvadvpn.lib.theme.Dimens
 import net.mullvad.mullvadvpn.viewmodel.DeviceRevokedSideEffect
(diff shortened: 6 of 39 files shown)


Pururun (changes requested):
> Reviewed 39 of 40 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: 39 of 40 files reviewed, 5 unresolved discussions (waiting on @Rawa)

___
*[`android/app/side_effect.md` line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-Nrjxh607Gwau1bb-BEu:-Nrjxh607Gwau1bb-BEv:b6q5iqc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/side_effect.md#L1)):*
> ```Markdown
> # Side Effects
> ```

This should be removed or rewritten
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt` line 161 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NrjyiyK7bJ6LkNTMyfK:-NrjyiyK7bJ6LkNTMyfL:bwu8p8u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt#L161)):*
> ```Kotlin
>         },
>         onAccountClick = {
>             navigator.navigate(AccountDestination, true) { launchSingleTop = true }
> ```

Since we probably should use navigate with single top and even only when resumed should we write a helper method for this so it is not forgotten?
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/di/UiModule.kt` line 111 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NrjzNvH--bgkzzTh4j-:-NrjzNvH--bgkzzTh4j0:bc7uxmh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/src/main/kotlin/net/mullvad/mullvadvpn/di/UiModule.kt#L111)):*
> ```Kotlin
>     single { PortRangeUseCase(get()) }
>     single { RelayListUseCase(get(), get()) }
>     single { OutOfTimeUseCase(get(), get(), MainScope()) }
> ```

Would this risk calls to the service/daemon being done on the main thread?
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/OutOfTimeViewModel.kt` line 132 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-Nrjzi6lAdMoLcapMNFv:-Nrjzi6mDF5aOoBb4Lv0:bjpn40d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/OutOfTimeViewModel.kt#L132)):*
> ```Kotlin
>     }
> 
>     private fun outOfTimeEffect() =
> ```

:star: 
___
*[`android/test/mockapi/src/main/kotlin/net/mullvad/mullvadvpn/test/mockapi/AccountExpiryMockApiTest.kt` line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-Nrk-6Lt4dMtZsaBqAX-:-Nrk-6Lt4dMtZsaBqAX0:bb76rno) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b74b1651a09e8bd38662427a91b27dc0fe8019d3/android/test/mockapi/src/main/kotlin/net/mullvad/mullvadvpn/test/mockapi/AccountExpiryMockApiTest.kt#L53)):*
> ```Kotlin
> 
>     @Test
>     @Disabled(
> ```

Woho

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: 39 of 40 files reviewed, 5 unresolved discussions (waiting on @albin-mullvad and @Pururun)

___
*[`android/app/side_effect.md` line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-Nrjxh607Gwau1bb-BEu:-Nrk0JQn9rp2eqS8Zp4I:b6av041) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/side_effect.md#L1)):*

This should be removed or rewritten
</blockquote></details>

Thanks, it should be removed.
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt` line 161 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NrjyiyK7bJ6LkNTMyfK:-Nrk32aA5o49-nsljGJH:bfihgd7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt#L161)):*

Since we probably should use navigate with single top and even only when resumed should we write a helper method for this so it is not forgotten?
</blockquote></details>

I'm open to it. Technically though I'm not sure we will always need launchSingleTop, and in many of these cases they are a bit doing "double" safety. E.g single top prevents us from having multiple AccountScreen in our backstack, but at the same time we only navigate there from Connect, and we protect ourself from duplicate navigation with the `onlyIfResumed` flag. So on side I agree, it makes it easier for us to hide this away, but on the otherside we make our code a bit more complicated by not showing explicitly what we are doing. We wouldn't stand to gain much in terms of lines of code since it is all on one line, so it would be for the fact that we might miss a launchSingleTop when we want it.

What do you think @albin-mullvad?
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/di/UiModule.kt` line 111 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NrjzNvH--bgkzzTh4j-:-Nrk22cg0PyRqewzESaW:bf9qq90) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/src/main/kotlin/net/mullvad/mullvadvpn/di/UiModule.kt#L111)):*

Would this risk calls to the service/daemon being done on the main thread?
</blockquote></details>

I don't believe so, since we use `AccountRepository` & `MessageHandler` in the OutOfTimeUseCase. It would indeed run on the main thread and I think later on it would be an improvement to looking into moving logic like this over to an IO dispatcher. From what I've read and my understanding (source: https://stackoverflow.com/a/63728529 & https://medium.com/androiddevelopers/coroutines-patterns-for-work-that-shouldnt-be-cancelled-e26c40f142ad ) this should be acceptable to do.
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/OutOfTimeViewModel.kt` line 132 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-Nrjzi6lAdMoLcapMNFv:-Nrk2z-g17B0uhHsXBZr:bgc7vqv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/OutOfTimeViewModel.kt#L132)):*

:star:
</blockquote></details>

:heart: 
___
*[`android/test/mockapi/src/main/kotlin/net/mullvad/mullvadvpn/test/mockapi/AccountExpiryMockApiTest.kt` line 53 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-Nrk-6Lt4dMtZsaBqAX-:-Nrk3uxp39GWADMEFGc-:bjwa7hh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b74b1651a09e8bd38662427a91b27dc0fe8019d3/android/test/mockapi/src/main/kotlin/net/mullvad/mullvadvpn/test/mockapi/AccountExpiryMockApiTest.kt#L53)):*

Woho
</blockquote></details>

:tada:

Pururun (approved):
> Reviewed 1 of 40 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: all files reviewed, 1 unresolved discussion (waiting on @albin-mullvad)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/di/UiModule.kt` line 111 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NrjzNvH--bgkzzTh4j-:-NroL4NX550jwK0Wj4qq:b3lbd3s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/src/main/kotlin/net/mullvad/mullvadvpn/di/UiModule.kt#L111)):*

I don't believe so, since we use `AccountRepository` & `MessageHandler` in the OutOfTimeUseCase. It would indeed run on the main thread and I think later on it would be an improvement to looking into moving logic like this over to an IO dispatcher. From what I've read and my understanding (source: https://stackoverflow.com/a/63728529 & https://medium.com/androiddevelopers/coroutines-patterns-for-work-that-shouldnt-be-cancelled-e26c40f142ad ) this should be acceptable to do.
</blockquote></details>

Right let's leave it like that then. :)

albin-mullvad (commented):
> Reviewed 7 of 40 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: all files reviewed, 2 unresolved discussions (waiting on @Rawa)

___
*[`android/app/side_effect.md` line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NrtYoYhBC5yavMZZRJ2:-NrtYoYhBC5yavMZZRJ3:b-kfc54w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/side_effect.md#L1)):*
> ```Markdown
> # Side Effects
> ```

We might want keep some of this documentation in a new document `android/app/docs/architecture.md`

How about starting such a document?
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/util/Effect.kt` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NrtbRKT7GMXnuvy1I09:-NrtbRKT7GMXnuvy1I0A:bt6cpmv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/util/Effect.kt#L12)):*
> ```Kotlin
> 
> @Composable
> inline fun <T> LaunchedEffectCollect(
> ```

I would like to discuss this function versus `CollectSideEffectWithLifecycle`. It seems to me that we might always want to use the lifecycle aware variant :thinking: 



_Code quote:_
```Kotlin
LaunchedEffectCollect
```

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: 38 of 40 files reviewed, 2 unresolved discussions (waiting on @albin-mullvad and @Pururun)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/util/Effect.kt` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NrtbRKT7GMXnuvy1I09:-Ns7QbHpD7n95EzazeHS:b8sy2an) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/util/Effect.kt#L12)):*

I would like to discuss this function versus `CollectSideEffectWithLifecycle`. It seems to me that we might always want to use the lifecycle aware variant :thinking:
</blockquote></details>

Discussed offline. 

Conclusion, most of the cases can probably be bound by the lifecycle. However, we should take each case into consideration individually since events triggered by UI does not get replayed and if we don't consume and miss them we might get unexpected behavior. So to limit the scope of this PR we use LaunchedEffectCollect until we've gone through each individual collection.
___
*[`android/app/side_effect.md` line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NrtYoYhBC5yavMZZRJ2:-Ns7RbRr2GOFOAlUwcsJ:bc6b5ux) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/side_effect.md#L1)):*

We might want keep some of this documentation in a new document `android/app/docs/architecture.md`

How about starting such a document?
</blockquote></details>

I agree, we should have such a document. I'd say we add it in a separate PR and after we've had internal meeting about plan for architecture in the feature. What do you think?

albin-mullvad (changes requested):
> Reviewed 31 of 40 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: all files reviewed, 9 unresolved discussions (waiting on @Pururun and @Rawa)

___
*[`-- commits` line 5 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-Ns8StQ5-K-y-546dKaH):*
Make shorter and skip initial space



_Code quote:_
```
  Update OutOfTimeUseCase to be reactive, and adjust related view models to handle side effect correctly
```
___
*[`-- commits` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-Ns8S_vX7WGdjZvALLpA):*
This message subject hurts by eyes :laughing: 

I suggest going with just: `Re-enable test`



_Code quote:_
```
Re-enable testAccountTimeExpiredWhileUsingTheAppShouldShowOutOfTimeScreen
```
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt` line 161 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NrjyiyK7bJ6LkNTMyfK:-Ns8rVLCDmBFHpzOVZJX:biihsgk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt#L161)):*

I'm open to it. Technically though I'm not sure we will always need launchSingleTop, and in many of these cases they are a bit doing "double" safety. E.g single top prevents us from having multiple AccountScreen in our backstack, but at the same time we only navigate there from Connect, and we protect ourself from duplicate navigation with the `onlyIfResumed` flag. So on side I agree, it makes it easier for us to hide this away, but on the otherside we make our code a bit more complicated by not showing explicitly what we are doing. We wouldn't stand to gain much in terms of lines of code since it is all on one line, so it would be for the fact that we might miss a launchSingleTop when we want it.

What do you think @albin-mullvad?
</blockquote></details>

Imo it would be nice if we could somehow transparantely use a decided default behavior for all `navigate` calls, however since we are still figuring out how handle it in terms of "double"/redundant safety and lifecycle (e.g. dialogs) I believe it probably makes sense to wait a bit with looking into if/how we can define a default behavior
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/usecase/OutOfTimeUseCase.kt` line 49 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsCRp1j617F3OZHnEsI:-NsCRp1j617F3OZHnEsJ:b-w34abo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75cdc742650ddcfc861828f5e6d37218877d1677/android/app/src/main/kotlin/net/mullvad/mullvadvpn/usecase/OutOfTimeUseCase.kt#L49)):*
> ```Kotlin
>         }
> 
>     // What if we already are out of time?
> ```

Remove or clarify
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/ConnectViewModel.kt` line 190 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsCS4QlBFBLph1E2FGV:-NsCS4QlBFBLph1E2FGW:b-ut332u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75cdc742650ddcfc861828f5e6d37218877d1677/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/ConnectViewModel.kt#L190)):*
> ```Kotlin
>     }
> 
>     // isOutOfTime=true and we should navigate to OutOfTime view.
> ```

nit: I'm not sure this comment adds much value compared to the code.
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/ConnectViewModel.kt` line 194 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsCSFnj7kW-x7rW_BcE:-NsCSFnj7kW-x7rW_BcF:b-ut332u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75cdc742650ddcfc861828f5e6d37218877d1677/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/ConnectViewModel.kt#L194)):*
> ```Kotlin
>         outOfTimeUseCase.isOutOfTime.filter { it == true }.map { UiSideEffect.OutOfTime }
> 
>     // If DeviceState is Revoked we should navigate to the RevokedDevice screen.
> ```

nit: I'm not sure this comment adds much value compared to the code.
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/LoginViewModel.kt` line 104 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsCSNs856gX5U8L1wjs:-NsCSNs856gX5U8L1wjt:b18nsdt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75cdc742650ddcfc861828f5e6d37218877d1677/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/LoginViewModel.kt#L104)):*
> ```Kotlin
>                 when (val result = loginDeferred.awaitWithTimeoutOrNull(LOGIN_TIMEOUT_MILLIS)) {
>                     LoginResult.Ok -> {
>                         newDeviceNotificationUseCase.newDeviceCreated()
> ```

Is this new order significant? In that case we should probably make sure to document that somehow.
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/OutOfTimeViewModel.kt` line 133 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsCTDpxABYkALOR3Lxu:-NsCTDpxABYkALOR3Lxv:b-38zwt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75cdc742650ddcfc861828f5e6d37218877d1677/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/OutOfTimeViewModel.kt#L133)):*
> ```Kotlin
> 
>     private fun outOfTimeEffect() =
>         // If the user is out of time, navigate to the out of time screen.
> ```

Incorrect comment. I suggest skippit it
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt` line 90 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsCScnx6BYgv0Zbyqy5:-NsCScnx6BYgv0Zbyqy6:b-cb4lhs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75cdc742650ddcfc861828f5e6d37218877d1677/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt#L90)):*
> ```Kotlin
> 
>     private fun outOfTimeEffect() =
>         // If the user is out of time, navigate to the out of time screen.
> ```

It looks like this comment is incorrect after being copied from somewhere else. Since it also doesn't add much value compared to the code I suggest skipping it
___
*[`android/app/side_effect.md` line 1 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NrtYoYhBC5yavMZZRJ2:-Ns8qqGYE3-yznl8Q8Yr:b-ltd9la) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cdbd9a547e667ce55dac3d93e050b64381d744ac/android/app/side_effect.md#L1)):*

I agree, we should have such a document. I'd say we add it in a separate PR and after we've had internal meeting about plan for architecture in the feature. What do you think?
</blockquote></details>

Sure :thumbsup:

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: 22 of 40 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @Pururun)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/usecase/OutOfTimeUseCase.kt` line 49 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsCRp1j617F3OZHnEsI:-NsIEfyp2VjzRCrxPjV8:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75cdc742650ddcfc861828f5e6d37218877d1677/android/app/src/main/kotlin/net/mullvad/mullvadvpn/usecase/OutOfTimeUseCase.kt#L49)):*

Remove or clarify
</blockquote></details>

Done.
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/LoginViewModel.kt` line 104 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsCSNs856gX5U8L1wjs:-NsIFAHI45Uy1w0jGZLa:baij20e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75cdc742650ddcfc861828f5e6d37218877d1677/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/LoginViewModel.kt#L104)):*

Is this new order significant? In that case we should probably make sure to document that somehow.
</blockquote></details>

Not really significant, it was just more clear. Technically, since a launch is happening below it could finish and launch a side effect and navigate away, however in practice it would never happen. This ensures it technically can't happen either, so it is just a small matter of correctness, and doesn't really need a comment I feel.
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/OutOfTimeViewModel.kt` line 133 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsCTDpxABYkALOR3Lxu:-NsIFwMO53Vf2La7Tmhl:b-e4xte1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75cdc742650ddcfc861828f5e6d37218877d1677/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/OutOfTimeViewModel.kt#L133)):*

Incorrect comment. I suggest skippit it
</blockquote></details>

Agree, I've removed comment an d changed the function name to more correctly represent what it does.
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt` line 90 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsCScnx6BYgv0Zbyqy5:-NsIG8Op6oUwi9H6urK3:b-e4xte1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75cdc742650ddcfc861828f5e6d37218877d1677/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt#L90)):*

It looks like this comment is incorrect after being copied from somewhere else. Since it also doesn't add much value compared to the code I suggest skipping it
</blockquote></details>

Agree, I've removed comment an d changed the function name to more correctly represent what it does.

albin-mullvad (approved):
> Reviewed 18 of 18 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: all files reviewed, 1 unresolved discussion (waiting on @Rawa)

___
*[`-- commits` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsSKKPa5AGhBtEpRvvq):*
Nit: typo "simply"?

Pururun (approved):
> Reviewed 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: all files reviewed, 1 unresolved discussion (waiting on @Rawa)

albin-mullvad (changes requested):
> Reviewed 1 of 40 files at r1, 7 of 7 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: all files reviewed, 8 unresolved discussions (waiting on @Rawa)

___
*[`-- commits` line 13 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshfjEy31yMqtiOVzNu):*
nit: this is a bit hard to read. I suggest going with just `Re-enable test` or `Re-enable expiry test`



_Code quote:_
```
Re-enable testAccountTimeExpiredWhileUsingTheAppShouldShowOutOfTimeScreen
```
___
*[`-- commits` line 18 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshgKs436EnmZUuyjBv):*
I suggest rephrasing this commit message subject if that wasn't already the plan. We have a plan/idea to improve this, but atm this commit rather reverts to something similar to the previous behavior. We should probably avoid the word "temp" unless we're sure when it will be addressed.
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt` line 88 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshnAWV5LonSFYjT8rr:-NshnAWV5LonSFYjT8rs:b-jj18fv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7d16e75a459825c3ac18d1542fcadd8aa28aca7/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt#L88)):*
> ```Kotlin
>     }
> 
>     private fun hasAddedTime() =
> ```

Can we clarify that this is a flow of potential updates? It's easy to mistake this for a regular boolean return function



_Code quote:_
```Kotlin
 hasAddedTime
```
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt` line 91 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-Nshn0JrAsodb7Bj2WBZ:-Nshn0JrAsodb7Bj2WB_:bndj2v0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7d16e75a459825c3ac18d1542fcadd8aa28aca7/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt#L91)):*
> ```Kotlin
>         accountRepository.accountExpiryState
>             .map { it.date() }
>             .filterNotNull()
> ```

I believe `mapNotNull` can be used here



_Code quote:_
```Kotlin
            .map { it.date() }
            .filterNotNull()
```
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt` line 94 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshlQtj2-akrwCLKIeg:-NshlQtj2-akrwCLKIeh:b-f8ex4i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7d16e75a459825c3ac18d1542fcadd8aa28aca7/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt#L94)):*
> ```Kotlin
>             .filter { it.minusHours(MIN_TIME_PAST_ACCOUNT_EXPIRY).isAfterNow }
>             .map {
>                 paymentUseCase.resetPurchaseResult()
> ```

This call isn't really part of mapping the data. so this map should probably be split into:

```
.onEach { paymentUseCase.resetPurchaseResult() }
.map {UiSideEffect.OpenConnectScreen }
```



_Code quote:_
```Kotlin
paymentUseCase.resetPurchaseResult()
```
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt` line 150 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshmhrO-ye5zFUGbYfI:-NshmhrO-ye5zFUGbYfJ:bw50wes) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7d16e75a459825c3ac18d1542fcadd8aa28aca7/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt#L150)):*
> ```Kotlin
> 
>     companion object {
>         private const val MIN_TIME_PAST_ACCOUNT_EXPIRY = 20
> ```

Clarify that it's time in hours



_Code quote:_
```Kotlin
MIN_TIME_PAST_ACCOUNT_EXPIRY
```
___
*[`android/app/src/test/kotlin/net/mullvad/mullvadvpn/usecase/OutOfTimeUseCaseTest.kt` line 174 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshocNjAhBJFxSdBmqV:-NshocNjAhBJFxSdBmqW:b-q5gwv1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7d16e75a459825c3ac18d1542fcadd8aa28aca7/android/app/src/test/kotlin/net/mullvad/mullvadvpn/usecase/OutOfTimeUseCaseTest.kt#L174)):*
> ```Kotlin
>             AccountExpiry.Available(initialAccountExpiry.expiryDateTime.plusDays(30))
>         // Act, Assert
>         outOfTimeUseCase.isOutOfTime.test {
> ```

Please add some comments to this test like you've done for the other tests. It really helped quickly understanding the intended test/behavior.

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: 38 of 40 files reviewed, 8 unresolved discussions (waiting on @albin-mullvad and @Pururun)

___
*[`-- commits` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NsSKKPa5AGhBtEpRvvq):*

Nit: typo "simply"?
</blockquote></details>

Thanks, should be simplied effects
___
*[`-- commits` line 13 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshfjEy31yMqtiOVzNu):*

nit: this is a bit hard to read. I suggest going with just `Re-enable test` or `Re-enable expiry test`
</blockquote></details>

Thought I fixed this but I believe I goofed my self with rebase and not using `reword`. Fixed now.
___
*[`-- commits` line 18 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshgKs436EnmZUuyjBv):*

I suggest rephrasing this commit message subject if that wasn't already the plan. We have a plan/idea to improve this, but atm this commit rather reverts to something similar to the previous behavior. We should probably avoid the word "temp" unless we're sure when it will be addressed.
</blockquote></details>

Done.
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt` line 88 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshnAWV5LonSFYjT8rr:-NshwFva7blmBSFwBuPv:b-dlyu75) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7d16e75a459825c3ac18d1542fcadd8aa28aca7/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt#L88)):*

Can we clarify that this is a flow of potential updates? It's easy to mistake this for a regular boolean return function
</blockquote></details>

I've renamed it to `hasAddedTimeEffect`, to be in line with other ViewModels. I think later once we reworked this logic and to look at expiry from account creation, it can be provided by a usecase as a boolean and then the viewmodel can filter and convert to side effect. 
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt` line 91 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-Nshn0JrAsodb7Bj2WBZ:-NshwhWy9XI4Ehz1QZxf:bv575wm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7d16e75a459825c3ac18d1542fcadd8aa28aca7/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt#L91)):*

I believe `mapNotNull` can be used here
</blockquote></details>

Thanks, good catch! :dart: 
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt` line 94 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshlQtj2-akrwCLKIeg:-NshwoFJ-HQd0TJa-uSQ:bsv6wca) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7d16e75a459825c3ac18d1542fcadd8aa28aca7/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt#L94)):*

This call isn't really part of mapping the data. so this map should probably be split into:

```
.onEach { paymentUseCase.resetPurchaseResult() }
.map {UiSideEffect.OpenConnectScreen }
```
</blockquote></details>

I agree, that is much more clear.
___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt` line 150 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshmhrO-ye5zFUGbYfI:-NshxhS65OSRWhqtPNFm:b-eowr4b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7d16e75a459825c3ac18d1542fcadd8aa28aca7/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/WelcomeViewModel.kt#L150)):*

Clarify that it's time in hours
</blockquote></details>

Clarified!
___
*[`android/app/src/test/kotlin/net/mullvad/mullvadvpn/usecase/OutOfTimeUseCaseTest.kt` line 174 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-NshxsBA3dKWK605YTj8:-NshxsBA3dKWK605YTj9:b1zau2f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c7d16e75a459825c3ac18d1542fcadd8aa28aca7/android/app/src/test/kotlin/net/mullvad/mullvadvpn/usecase/OutOfTimeUseCaseTest.kt#L174)):*
> ```Kotlin
>             AccountExpiry.Available(initialAccountExpiry.expiryDateTime.plusDays(30))
>         // Act, Assert
>         outOfTimeUseCase.isOutOfTime.test {
> ```

Added! :100:

albin-mullvad (approved):
:lgtm:

> Reviewed 1 of 1 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873#-:-Nshz2C26RBSmf48naPA:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

albin-mullvad (approved):
> Reviewed 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5873)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5903, opened 2024-03-05
Log when certain functions are called an excessive number of times

dlon:
Fix DES-683.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3744,6 +3744,7 @@ dependencies = [
  "base64 0.13.1",
  "ipnetwork",
  "jnix",
+ "log",
  "serde",
  "thiserror",
  "x25519-dalek",
--- a/talpid-core/src/dns/macos.rs
+++ b/talpid-core/src/dns/macos.rs
@@ -88,6 +88,8 @@ impl State {
         interface: &str,
         servers: &[IpAddr],
     ) -> Result<()> {
+        talpid_types::detect_flood!();
+
         let servers: Vec<DnsServer> = servers.iter().map(|ip| ip.to_string()).collect();
         let new_settings = DnsSettings::from_server_addresses(&servers, interface.to_string());
         match &self.dns_settings {
@@ -115,6 +117,8 @@ impl State {
     }
 
     fn on_changed_keys(&mut self, store: SCDynamicStore, changed_keys: CFArray<CFString>) {
+        talpid_types::detect_flood!();
+
         if let Some(expected_settings) = &self.dns_settings {
             for path in &changed_keys {
                 let should_set_dns = match DnsSettings::load(&store, path.clone()).ok() {
--- a/talpid-core/src/offline/macos.rs
+++ b/talpid-core/src/offline/macos.rs
@@ -96,6 +96,8 @@ pub async fn spawn_monitor(
         let mut route_listener = route_listener.fuse();
 
         loop {
+            talpid_types::detect_flood!();
+
             select! {
                 _ = timeout => {
                     // Update shared state
--- a/talpid-routing/src/debounce.rs
+++ b/talpid-routing/src/debounce.rs
@@ -99,6 +99,7 @@ impl BurstGuard {
 
     /// Asynchronously trigger burst
     pub fn trigger(&self) {
+        talpid_types::detect_flood!();
         self.trigger_with_period(self.buffer_period)
     }
 
--- a/talpid-routing/src/unix/macos/mod.rs
+++ b/talpid-routing/src/unix/macos/mod.rs
@@ -158,12 +158,16 @@ impl RouteManagerImpl {
         let mut completion_tx = None;
 
         loop {
+            talpid_types::detect_flood!();
+
             futures::select_biased! {
                 route_message = self.routing_table.next_message().fuse() => {
                     self.handle_route_message(route_message);
                 }
 
                 _ = self.check_default_routes_restored.next() => {
+                    talpid_types::detect_flood!();
+
                     if self.check_default_routes_restored.is_terminated() {
                         continue;
                     }
@@ -328,6 +332,8 @@ impl RouteManagerImpl {
         &mut self,
         message: std::result::Result<RouteSocketMessage, watch::Error>,
     ) {
+        talpid_types::detect_flood!();
+
         match message {
             Ok(RouteSocketMessage::DeleteRoute(route)) => {
                 // Forget about applied route, if relevant
@@ -380,6 +386,8 @@ impl RouteManagerImpl {
     ///   server. The gateway of the relay route is set to the first interface in the network
     ///   service order that has a working ifscoped default route.
     async fn refresh_routes(&mut self) -> Result<()> {
+        talpid_types::detect_flood!();
+
         self.update_best_default_route(interface::Family::V4)?;
         self.update_best_default_route(interface::Family::V6)?;
 
--- a/talpid-routing/src/unix/macos/routing_socket.rs
+++ b/talpid-routing/src/unix/macos/routing_socket.rs
@@ -87,6 +87,8 @@ impl RoutingSocket {
 
     pub async fn wait_for_response(&mut self, response_num: i32) -> Result<Vec<u8>> {
         loop {
+            talpid_types::detect_flood!();
+
             let mut buffer = vec![0u8; 2048];
             // do not truncate the buffer - trailing empty bytes won't be written but will be
             // assumed in the data format.
--- a/talpid-types/Cargo.toml
+++ b/talpid-types/Cargo.toml
@@ -17,6 +17,7 @@ base64 = "0.13"
 x25519-dalek = { version = "2.0.1", features = ["static_secrets", "zeroize", "getrandom"] }
 thiserror = { workspace = true }
 zeroize = "1.5.7"
+log = { workspace = true }
 
 [target.'cfg(target_os = "android")'.dependencies]
 jnix = { version = "0.5.1", features = ["derive"] }
--- a/talpid-types/src/error.rs
+++ b/talpid-types/src/error.rs
@@ -68,3 +68,59 @@ macro_rules! win32_err {
         }
     }};
 }
+
+pub mod flood {
+    #[doc(hidden)]
+    pub use log;
+
+    use std::time::{Duration, Instant};
+
+    const CALLS_INTERVAL: Duration = Duration::from_secs(5);
+    const CALLS_THRESHOLD: usize = 1000;
+
+    /// Log when a line is hit unusually frequently, that is, over `CALLS_THRESHOLD` times within a
+    /// period of `CALLS_INTERVAL`.
+    #[macro_export]
+    macro_rules! detect_flood {
+        () => {{
+            static FLOOD: ::std::sync::Mutex<$crate::flood::DetectFlood> =
+                ::std::sync::Mutex::new($crate::flood::DetectFlood::new());
+            if FLOOD.lock().unwrap().bump() {
+                $crate::flood::log::warn!(
+                    "Flood: {}, line {}, col {}",
+                    file!(),
+                    line!(),
+                    column!()
+                );
+            }
+        }};
+    }
+
+    /// Used to detect code that is running too frequently
+    pub struct DetectFlood {
+        last_clear: Option<Instant>,
+        counter: usize,
+    }
+
+    impl DetectFlood {
+        pub const fn new() -> Self {
+            DetectFlood {
+                last_clear: None,
+                counter: 0,
+            }
+        }
+
+        pub fn bump(&mut self) -> bool {
+            let now = Instant::now();
+            let last_clear = self.last_clear.get_or_insert(now);
+            if now.saturating_duration_since(*last_clear) >= CALLS_INTERVAL {
+                self.last_clear = Some(now);
+                self.counter = 0;
+                false
+            } else {
+                self.counter = self.counter.saturating_add(1);
+                self.counter == CALLS_THRESHOLD
+            }
+        }
+    }
+}


MarkusPettersson98 (changes requested):
> Reviewed 4 of 7 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`talpid-types/src/error.rs` line 79 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsMvg7IDJge6M0cVXDG:-NsMvg7IDJge6M0cVXDH:bwsj6oj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dc6bccc65cc5cf8363edb87d89a930143a52183d/talpid-types/src/error.rs#L79)):*
> ```XML
> 
>     #[macro_export]
>     macro_rules! detect_flood {
> ```

I think a doc-comment would be very nice to have, so that if I see `detect_flood!` being used somewhere I can quickly bring up the documentation at the call site instead of having to go to this source to figure out its purpose :blush:



_Code quote:_
```XML
    #[macro_export]
    macro_rules! detect_flood {
```

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`talpid-types/src/error.rs` line 79 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsMvg7IDJge6M0cVXDG:-NsNL2JqDdU62QURbpCw:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dc6bccc65cc5cf8363edb87d89a930143a52183d/talpid-types/src/error.rs#L79)):*

I think a doc-comment would be very nice to have, so that if I see `detect_flood!` being used somewhere I can quickly bring up the documentation at the call site instead of having to go to this source to figure out its purpose :blush:
</blockquote></details>

Done.

MarkusPettersson98 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-:-NsOageH5aAEdyi-WqqX:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (changes requested):
> Reviewed 4 of 7 files at r1, 2 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon)

___
*[`talpid-types/src/error.rs` line 72 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsTdLxl2JFUboVSGZVF:-NsTdLxl2JFUboVSGZVG:bzb9cop) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L72)):*
> ```XML
> }
> 
> pub mod flood {
> ```

I'm not a fan of more code ending up in this crate. But not going to block it this time since I don't have a concrete better suggestion right now.
___
*[`talpid-types/src/error.rs` line 81 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsT_a6z7t2BH7rRs4XJ:-NsT_a6z7t2BH7rRs4XK:b-dgfqm2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L81)):*
> ```XML
>     /// period of `CALLS_INTERVAL`.
>     ///
>     /// The caller of this macro must have `once_cell` and `log` as dependencies.
> ```

Instead of declaring that the caller crate must have `once_cel` and `log` I think it would be cleaner if the crate defining the macro re-exports these crates and uses `$crate::once_cell::...` etc instead.

Implicitly the macro also relies on special versions of these crates. If `log 0.5` is released one day, this macro might not be compatible.

The reexports can be tagged with `#[doc(hidden)]` since they are not supposed to be part of the "public" API of this crate.
___
*[`talpid-types/src/error.rs` line 86 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsTb-w94X8k24_lbOYC:-NsTb-w94X8k24_lbOYD:b-hucn88) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L86)):*
> ```XML
>         () => {{
>             static FLOOD: ::once_cell::sync::Lazy<
>                 ::std::sync::Mutex<talpid_types::flood::DetectFlood>,
> ```

This is fragile against the calling crate importing `talpid_types` under a different name. Can we use `$crate` here also instead?
___
*[`talpid-types/src/error.rs` line 91 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsTczO68u4NphcTLoQF:-NsTczO68u4NphcTLoQG:b99vh93) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L91)):*
> ```XML
>             });
>             if FLOOD.lock().unwrap().bump() {
>                 ::log::debug!("Flood: {}, line {}, col {}", file!(), line!(), column!());
> ```

Good log message! But should we maybe consider bumping this to at least a warning? I know we add it to debug something, but if this log line is ever executed it means the process is in a pretty bad state, so it's really a warning.
___
*[`talpid-types/src/error.rs` line 120 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsTcOCc5Zb3ogHabz--:-NsTcOCc5Zb3ogHabz-0:bnjtc0d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L120)):*
> ```XML
>                 let was_less = self.counter < CALLS_THRESHOLD;
>                 self.counter = self.counter.saturating_add(1);
>                 return was_less && self.counter >= CALLS_THRESHOLD;
> ```

A comment here would be justified IMHO. It took me a few iterations of thinking before I realized why we check the counter against the threshold twice.. Because we only want to return `true` *once*, right?

Can't this be simplified to just `self.counter == CALLS_THRESHOLD` to only return `true` exactly when we hit the threshold?
___
*[`talpid-types/src/error.rs` line 122 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsTcC0m82Ylc64hJCcc:-NsTcC0m82Ylc64hJCcd:b5aqdq5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L122)):*
> ```XML
>                 return was_less && self.counter >= CALLS_THRESHOLD;
>             }
>             false
> ```

IMHO it would be slightly more straight forward if this was moved into the if branch. That removes the early return. The `false` is only ever hit when we hit the if branch anyway.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903)* status: 6 of 9 files reviewed, 4 unresolved discussions (waiting on @faern and @MarkusPettersson98)

___
*[`talpid-types/src/error.rs` line 72 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsTdLxl2JFUboVSGZVF:-NsTt8DEERi9VkMkHBwQ:b-jnc2yh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L72)):*

I'm not a fan of more code ending up in this crate. But not going to block it this time since I don't have a concrete better suggestion right now.
</blockquote></details>

Same. If we don't mind crate inflation, I could create `talpid-log` or `talpid-debug`.
___
*[`talpid-types/src/error.rs` line 81 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsT_a6z7t2BH7rRs4XJ:-NsTtPMXDnbKKgKA5813:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L81)):*

Instead of declaring that the caller crate must have `once_cel` and `log` I think it would be cleaner if the crate defining the macro re-exports these crates and uses `$crate::once_cell::...` etc instead.

Implicitly the macro also relies on special versions of these crates. If `log 0.5` is released one day, this macro might not be compatible.

The reexports can be tagged with `#[doc(hidden)]` since they are not supposed to be part of the "public" API of this crate.
</blockquote></details>

Done.
___
*[`talpid-types/src/error.rs` line 86 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsTb-w94X8k24_lbOYC:-NsTuZAGABqlVMfzeEa5:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L86)):*

This is fragile against the calling crate importing `talpid_types` under a different name. Can we use `$crate` here also instead?
</blockquote></details>

Done.
___
*[`talpid-types/src/error.rs` line 120 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsTcOCc5Zb3ogHabz--:-NsTunJxA_i7FjUpHgPH:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L120)):*

A comment here would be justified IMHO. It took me a few iterations of thinking before I realized why we check the counter against the threshold twice.. Because we only want to return `true` *once*, right?

Can't this be simplified to just `self.counter == CALLS_THRESHOLD` to only return `true` exactly when we hit the threshold?
</blockquote></details>

Done.
___
*[`talpid-types/src/error.rs` line 122 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsTcC0m82Ylc64hJCcc:-NsTvDYw3mjW7X-5liV4:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L122)):*

IMHO it would be slightly more straight forward if this was moved into the if branch. That removes the early return. The `false` is only ever hit when we hit the if branch anyway.
</blockquote></details>

Done.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903)* status: 7 of 9 files reviewed, 4 unresolved discussions (waiting on @dlon and @MarkusPettersson98)

___
*[`talpid-types/Cargo.toml` line 20 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsmUm0O4qRoZTaG38YC:-NsmUm0O4qRoZTaG38YD:bt3q2b2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/57298ae14621aab283853278cfca7c0577517d41/talpid-types/Cargo.toml#L20)):*
> ```TOML
> thiserror = { workspace = true }
> zeroize = "1.5.7"
> log = "0.4"
> ```

`log` is also available as a workspace dependency.
___
*[`talpid-types/src/error.rs` line 81 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsT_a6z7t2BH7rRs4XJ:-NsmUyub3eEGBF6yXxCd:b6y7bo7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L81)):*

Done.
</blockquote></details>

Macro documentation not updated yet.
___
*[`talpid-types/src/error.rs` line 92 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsmWHqH6TEeRwONZy8r:-NsmWHqH6TEeRwONZy8s:b-tjouos) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/57298ae14621aab283853278cfca7c0577517d41/talpid-types/src/error.rs#L92)):*
> ```XML
>                 ::std::sync::Mutex<talpid_types::flood::DetectFlood>,
>             > = $crate::flood::once_cell::sync::Lazy::new(|| {
>                 ::std::sync::Mutex::new($crate::flood::DetectFlood::default())
> ```

`Mutex::new` is a `const fn`. If we change `last_clear: Instant` to `last_clear: Option<Instant>` we can have a `const fn `DetectFlood::new` and not need to deal with all the lazy stuff at all. It could just be a regular static. Would that not be simpler? At least it would get rid of needing `once_cell`. It would slightly complicate `bump`, but I think it can be done without much trouble.

What do you think? I think that would be somewhat simpler.
___
*[`talpid-types/src/error.rs` line 129 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsmVMsOAx_SH1UwKfXT:-NsmVMsOAx_SH1UwKfXU:b-jjsm1e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/57298ae14621aab283853278cfca7c0577517d41/talpid-types/src/error.rs#L129)):*
> ```XML
>             } else {
>                 self.counter = self.counter.saturating_add(1);
>                 self.counter == CALLS_THRESHOLD
> ```

If we move the `log::warn!` call into here, we do not need to re-export `log` to make it available to the caller, since it would then not be part of the macro.

Then the `DetectFlood` would be coupled to the logging, but that's what it will only ever be used for, so does it matter? :thinking:

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on @dlon and @MarkusPettersson98)

___
*[`talpid-types/src/error.rs` line 129 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsmVMsOAx_SH1UwKfXT:-NsmeDiKCUgOrlnxZcdB:b8ebpen) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/57298ae14621aab283853278cfca7c0577517d41/talpid-types/src/error.rs#L129)):*

If we move the `log::warn!` call into here, we do not need to re-export `log` to make it available to the caller, since it would then not be part of the macro.

Then the `DetectFlood` would be coupled to the logging, but that's what it will only ever be used for, so does it matter? :thinking:
</blockquote></details>

Aah, wait. I realize we depend on `file!`, `line!` etc to be inserted in the call position. Darn. They could theoretically be sent in as arguments, but it makes the solution dirtier. Let's ignore this, keeping the logging in the macro is fine!

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903)* status: 6 of 9 files reviewed, 2 unresolved discussions (waiting on @faern and @MarkusPettersson98)

___
*[`talpid-types/Cargo.toml` line 20 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsmUm0O4qRoZTaG38YC:-NssHpBL2mHdCHZ7jUGm:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/57298ae14621aab283853278cfca7c0577517d41/talpid-types/Cargo.toml#L20)):*

`log` is also available as a workspace dependency.
</blockquote></details>

Done.
___
*[`talpid-types/src/error.rs` line 81 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsT_a6z7t2BH7rRs4XJ:-NssHqHFDjvFNJHqP8lM:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bb994bc96dff8517c6dc6c7ddfe83306b49f9d27/talpid-types/src/error.rs#L81)):*

Macro documentation not updated yet.
</blockquote></details>

Done.
___
*[`talpid-types/src/error.rs` line 92 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsmWHqH6TEeRwONZy8r:-NssJPGWETWSFStelYDk:b-896fja) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/57298ae14621aab283853278cfca7c0577517d41/talpid-types/src/error.rs#L92)):*

`Mutex::new` is a `const fn`. If we change `last_clear: Instant` to `last_clear: Option<Instant>` we can have a `const fn `DetectFlood::new` and not need to deal with all the lazy stuff at all. It could just be a regular static. Would that not be simpler? At least it would get rid of needing `once_cell`. It would slightly complicate `bump`, but I think it can be done without much trouble.

What do you think? I think that would be somewhat simpler.
</blockquote></details>

Done!

faern (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-:-NssM-MDDcv2oYjxpPDJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`talpid-types/src/error.rs` line 92 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5903#-NsmWHqH6TEeRwONZy8r:-NssLmLC-eirv2HiEA7D:b8wvvf0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/57298ae14621aab283853278cfca7c0577517d41/talpid-types/src/error.rs#L92)):*

Done!
</blockquote></details>

NICE! The `get_or_insert` there really made in sparkling :heart:

PR #5912, opened 2024-03-07
Make the 'Clear all overrides' button only active when there are overrides

rablador:
Currently, the button is always active. According to the specs, we should make sure it is not active if there are no overrides being enabled.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5912)

### DIFF
--- a/ios/MullvadVPN/Coordinators/Settings/IPOverride/IPOverrideInteractor.swift
+++ b/ios/MullvadVPN/Coordinators/Settings/IPOverride/IPOverrideInteractor.swift
@@ -11,10 +11,11 @@ import MullvadLogging
 import MullvadSettings
 import MullvadTypes
 
-struct IPOverrideInteractor {
+final class IPOverrideInteractor {
     private let logger = Logger(label: "IPOverrideInteractor")
     private let repository: IPOverrideRepositoryProtocol
     private let tunnelManager: TunnelManager
+    private var statusWorkItem: DispatchWorkItem?
 
     private let statusSubject = CurrentValueSubject<IPOverrideStatus, Never>(.noImports)
     var statusPublisher: AnyPublisher<IPOverrideStatus, Never> {
@@ -87,8 +88,14 @@ struct IPOverrideInteractor {
     }
 
     private func resetToDefaultStatus(delay: Duration = .zero) {
-        DispatchQueue.main.asyncAfter(deadline: .now() + delay.timeInterval) {
-            statusSubject.send(defaultStatus)
+        statusWorkItem?.cancel()
+
+        let statusWorkItem = DispatchWorkItem { [weak self] in
+            guard let self, self.statusWorkItem?.isCancelled == false else { return }
+            self.statusSubject.send(self.defaultStatus)
         }
+        self.statusWorkItem = statusWorkItem
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + delay.timeInterval, execute: statusWorkItem)
     }
 }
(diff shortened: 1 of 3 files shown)


rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5912)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[`ios/MullvadVPN/Coordinators/Settings/IPOverride/IPOverrideInteractor.swift` line 90 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5912#-NsN7CRW54O2-69EiNm0:-NsN7CRW54O2-69EiNm1:b-43565v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ee5a8039463ff19f974fc079e61f5f64cd7fcc0d/ios/MullvadVPN/Coordinators/Settings/IPOverride/IPOverrideInteractor.swift#L90)):*
> ```Swift
>     }
> 
>     private func resetToDefaultStatus(delay: Duration = .zero) {
> ```

Added semi-related improvement here as well. Currently there can be weird timings for the status messages depending on user behavior (and patience).

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5912)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPN/Coordinators/Settings/IPOverride/IPOverrideInteractor.swift` line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5912#-NsO2Wl45IWadbhVDauI:-NsO2Wl45IWadbhVDauJ:b-xxoyjt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c2b4952160d74542cd829cae1f7032c7fa54b5d/ios/MullvadVPN/Coordinators/Settings/IPOverride/IPOverrideInteractor.swift#L94)):*
> ```Swift
> 
>         let statusWorkItem = DispatchWorkItem { [weak self] in
> ```

Calling `cancel` on dispatchWorkItems doesn't actually cancel the execution if the item has already scheduled.
We should check here as well  that the item itself is not cancelled
```swift
guard let self, self.statusWorkItem?.isCancelled == false else { return }
```

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5912)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/Coordinators/Settings/IPOverride/IPOverrideInteractor.swift` line 94 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5912#-NsO2Wl45IWadbhVDauI:-NsO5Ik67SH--WmoiEGs:b-l7rjqi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c2b4952160d74542cd829cae1f7032c7fa54b5d/ios/MullvadVPN/Coordinators/Settings/IPOverride/IPOverrideInteractor.swift#L94)):*

Calling `cancel` on dispatchWorkItems doesn't actually cancel the execution if the item has already scheduled.
We should check here as well  that the item itself is not cancelled
```swift
guard let self, self.statusWorkItem?.isCancelled == false else { return }
```
</blockquote></details>

Aha! Didn't know that. My intention was to cancel it altogether.

buggmagnet (approved):
:lgtm:

> Reviewed 1 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5912#-:-NsgVxLR2uEfWp8FH1dp:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

PR #5915, opened 2024-03-07
routing: use libc constants

pronebird:
This is a proposal PR to switch to using libc definitions on Mac because bitflags literally mirror what libc already has. I went through all of the values and they are identical with what was previously hardcoded in the source code.

On a side note, interesting approach that `nix` use is a special macro called `libc_bitflags!` which automatically generates a similar struct from the input of libc constants but I don't think it's public, so this PR continues the manual turk way for simplicity.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5915)


dlon (approved):
Much nicer. Thanks.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5915#-:-NsNvX5Y9zIfng94pRQZ:b-pbu1dg)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5917, opened 2024-03-07
routing: split data onto many files

pronebird:
`macos/data.rs` is over 1000 LOC long. This PR aims to split `data.rs` onto multiple smaller files under `macos/data` which should be easier to maintain in the long term.

Since this looks like huge change, I'd suggest you to pull the branch and run:

```
git -c color.diff.newMoved=black -c color.diff.oldMoved=black diff --color-moved=plain --unified=0 head~1
```

which should show you the same code in black and then what was added/replaced/removed in green and red.


pronebird:
On a side note wanted to ask about the use of futures crate in routing. Is it some ancient relic that perhaps should be replaced with tokio which seems to be uber popular and almost ubiquitous these days?

MarkusPettersson98:
Hi @pronebird, thanks for your contribution! Since we didn't act on this in a timely manner, and multiple merge conflicts have arisen at this point. We have decided to close your PR, but we greatly appreciate your effort. It can very well serve as inspiration for future refactors :blush:

PR #5918, opened 2024-03-07
Updated workflow to be compatible with xcconfig changes in main

niklasberglund:
Adapted to changes that have been merged to `main` which affect the workflow:
* `MULLVAD_IOS_DEVICE_PIN_CODE` has been renamed to `IOS_DEVICE_PIN_CODE `
* In `UITests.xcconfig.template` `HAS_TIME_ACCOUNT_NUMBER` and `NO_TIME_ACCOUNT_NUMBER` are specified per configuration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5918)

### DIFF
--- a/.github/workflows/ios-end-to-end-tests.yml
+++ b/.github/workflows/ios-end-to-end-tests.yml
@@ -35,18 +35,16 @@ jobs:
       - name: Configure Xcode project
         run: |
           for file in *.xcconfig.template ; do cp $file ${file//.template/} ; done
+          sed -i "" "/^HAS_TIME_ACCOUNT_NUMBER/d" UITests.xcconfig
+          sed -i "" "/^NO_TIME_ACCOUNT_NUMBER/d" UITests.xcconfig
           sed -i "" \
-            "/MULLVAD_IOS_DEVICE_PIN_CODE =/ s/= .*/= $IOS_DEVICE_PIN_CODE/" \
+            "/IOS_DEVICE_PIN_CODE =/ s/= .*/= $IOS_DEVICE_PIN_CODE/" \
             UITests.xcconfig
           sed -i "" \
-            "/MULLVAD_TEST_DEVICE_IDENTIFIER_UUID =/ s/= .*/= $TEST_DEVICE_IDENTIFIER_UUID/" \
+            "/TEST_DEVICE_IDENTIFIER_UUID =/ s/= .*/= $TEST_DEVICE_IDENTIFIER_UUID/" \
             UITests.xcconfig
-          sed -i "" \
-            "/MULLVAD_HAS_TIME_ACCOUNT_NUMBER =/ s/= .*/= $HAS_TIME_ACCOUNT_NUMBER/" \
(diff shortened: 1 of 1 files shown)


rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5918)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5918)* status: all files reviewed, 1 unresolved discussion (waiting on @niklasberglund)

___
*[`.github/workflows/ios-end-to-end-tests.yml` line 46 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5918#-NsgZUw4-CgwqcZ7pY8q:-NsgZUw4-CgwqcZ7pY8r:bn708bo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/288a7bf64adc5588216666ec289daf164c1da48f/.github/workflows/ios-end-to-end-tests.yml#L46)):*
> ```YAML
>             "/TEST_DEVICE_IDENTIFIER_UUID =/ s/= .*/= $TEST_DEVICE_IDENTIFIER_UUID/" \
>             UITests.xcconfig
>           printf "\n" >> UITests.xcconfig
> ```

Is there any reason we use `printf` instead of `echo` here ? We don't need to format any arguments, so I don't see why those wouldn't be `echo` calls instead

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5918)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @buggmagnet and @rablador)

___
*[`.github/workflows/ios-end-to-end-tests.yml` line 46 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5918#-NsgZUw4-CgwqcZ7pY8q:-Nsh-Kq49SI4GcoPN0le:b-5jmgdr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/288a7bf64adc5588216666ec289daf164c1da48f/.github/workflows/ios-end-to-end-tests.yml#L46)):*

Is there any reason we use `printf` instead of `echo` here ? We don't need to format any arguments, so I don't see why those wouldn't be `echo` calls instead
</blockquote></details>

Its common practice for shell scripts to use `printf` instead of `echo` because of portability issues with `echo`. In this case we don't need high portability anyways and it's not advanced usage that is likely to differ between different platforms which we're not going to be running this on either way. Have changed to using `echo`.

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5918#-:-Nsh9sbf4ccuISvNWbrR:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5913, opened 2024-03-07
routing: update nix

pronebird:
This PR bumps nix to the latest 0.28 and also fixes `OwnedFd` to `File` fiddling as Rust now supports the relevant `From` trait. Also replaced esoteric `std::os::unix::prelude` includes with `std::os::fd`, hope that doesn't break Linux builds in any way.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5913)

PR #5927, opened 2024-03-08
Deny nix git repo in deny.toml

faern:
Follow up to #5913. We stopped depending on the nix git repository. So we can finally stop allowing it as an external source. Bringing down the list of allowed git sources to the empty set, as it should be :sparkles: 

Also adding `deny.toml` to the list of important files that must be signed when changing. Because it controls what external sources we can pull in. So it's part of our supply chain security.


dlon (approved):
:+1: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5927#-:-NsSjfcs2WAPeu8o7oV9:b-8fx2u7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5931, opened 2024-03-09
routing: use block_in_place() in drop impl

pronebird:
As described in https://github.com/tokio-rs/tokio/issues/5843, current implementation of `Drop` in `RouteManager` would typically lead to panic. 

In a multi-threaded runtime the better approach would be to use `tokio::task::block_in_place`. In essence, even better approach would be to avoid doing async calls from `Drop` and call `stop()` somewhere in the clean up routine of a parent scope, but such change would encompass a larger scope and beyond what I am currently looking to fix.

I scanned the code for `current_thread_runtime` and found none in the way of that `Drop`, so I think it should work but don't take my word for it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5931)

### DIFF
--- a/talpid-routing/src/unix/mod.rs
+++ b/talpid-routing/src/unix/mod.rs
@@ -257,7 +257,6 @@ pub enum CallbackMessage {
 /// the route will be adjusted dynamically when the default route changes.
 pub struct RouteManager {
     manage_tx: Option<Arc<UnboundedSender<RouteManagerCommand>>>,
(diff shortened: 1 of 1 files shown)


dlon (approved):
Thanks. I followed this up with a non-blocking drop impl: https://github.com/mullvad/mullvadvpn-app/pull/5934

PR #5955, opened 2024-03-13
Add WireGuard tests for iOS app

niklasberglund:
In this PR the following tests have been implemented:
* `testWireGuardOverTCPManually` - https://linear.app/mullvad/issue/IOS-431/test-wireguard-over-tcp-manually
* `testWireGuardOverTCPAutomatically` - https://linear.app/mullvad/issue/IOS-429/test-wireguard-over-tcp-works-automatically
* `testWireGuardPortSettings` - https://linear.app/mullvad/issue/IOS-434/test-wireguard-port-settings

To test it out you can run the three tests, they are all under `RelayTests`. Another configuration property `SHOULD_BE_REACHABLE_DOMAIN = mullvad.net` has been introduced so make sure to set it in your `UITests.xcconfig` file. You must be connected to the `app-team-ios-tests` wifi to be able to run tests interacting with the firewall running on the NUC.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955)

### DIFF
--- a/ios/Configurations/UITests.xcconfig.template
+++ b/ios/Configurations/UITests.xcconfig.template
@@ -15,6 +15,9 @@ FIVE_WIREGUARD_KEYS_ACCOUNT_NUMBER =
 
 // Ad serving domain used when testing ad blocking. Note that we are assuming there's an HTTP server running on the host.
 AD_SERVING_DOMAIN = vpnlist.to
+
+// A domain which should be reachable. Used to verify Internet connectivity. Must be running a server on port 80.
+SHOULD_BE_REACHABLE_DOMAIN = mullvad.net
  
 // Base URL for the firewall API, Note that // will be treated as a comment, therefor you need to insert a ${} between the slashes for example http:/${}/8.8.8.8
 FIREWALL_API_BASE_URL = http:/${}/8.8.8.8
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -618,6 +618,7 @@
 		8529693A2B4F0238007EAD4C /* TermsOfServicePage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 852969392B4F0238007EAD4C /* TermsOfServicePage.swift */; };
 		8529693C2B4F0257007EAD4C /* Alert.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8529693B2B4F0257007EAD4C /* Alert.swift */; };
 		8532E6872B8CCED600ACECD1 /* ProblemReportSubmittedPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8532E6862B8CCED600ACECD1 /* ProblemReportSubmittedPage.swift */; };
+		8542CE242B95F7B9006FCA14 /* VPNSettingsPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8542CE232B95F7B9006FCA14 /* VPNSettingsPage.swift */; };
 		85557B0E2B591B2600795FE1 /* FirewallAPIClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85557B0D2B591B2600795FE1 /* FirewallAPIClient.swift */; };
 		85557B102B59215F00795FE1 /* FirewallRule.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85557B0F2B59215F00795FE1 /* FirewallRule.swift */; };
 		85557B122B594FC900795FE1 /* ConnectivityTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85557B112B594FC900795FE1 /* ConnectivityTests.swift */; };
@@ -1861,6 +1862,7 @@
 		852969392B4F0238007EAD4C /* TermsOfServicePage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TermsOfServicePage.swift; sourceTree = "<group>"; };
 		8529693B2B4F0257007EAD4C /* Alert.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Alert.swift; sourceTree = "<group>"; };
 		8532E6862B8CCED600ACECD1 /* ProblemReportSubmittedPage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ProblemReportSubmittedPage.swift; sourceTree = "<group>"; };
+		8542CE232B95F7B9006FCA14 /* VPNSettingsPage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = VPNSettingsPage.swift; sourceTree = "<group>"; };
 		85557B0D2B591B2600795FE1 /* FirewallAPIClient.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = FirewallAPIClient.swift; sourceTree = "<group>"; };
 		85557B0F2B59215F00795FE1 /* FirewallRule.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = FirewallRule.swift; sourceTree = "<group>"; };
 		85557B112B594FC900795FE1 /* ConnectivityTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ConnectivityTests.swift; sourceTree = "<group>"; };
@@ -3609,6 +3611,7 @@
 				852969392B4F0238007EAD4C /* TermsOfServicePage.swift */,
 				850201DE2B5040A500EF8C96 /* TunnelControlPage.swift */,
 				85FB5A0B2B6903990015DCED /* WelcomePage.swift */,
+				8542CE232B95F7B9006FCA14 /* VPNSettingsPage.swift */,
 			);
 			path = Pages;
 			sourceTree = "<group>";
@@ -5598,6 +5601,7 @@
 				8532E6872B8CCED600ACECD1 /* ProblemReportSubmittedPage.swift in Sources */,
 				85FB5A0C2B6903990015DCED /* WelcomePage.swift in Sources */,
 				850201DF2B5040A500EF8C96 /* TunnelControlPage.swift in Sources */,
+				8542CE242B95F7B9006FCA14 /* VPNSettingsPage.swift in Sources */,
 				85557B1E2B5FB8C700795FE1 /* HeaderBar.swift in Sources */,
 				85557B122B594FC900795FE1 /* ConnectivityTests.swift in Sources */,
 				852969332B4E9232007EAD4C /* Page.swift in Sources */,
--- a/ios/MullvadVPN/Classes/AccessbilityIdentifier.swift
+++ b/ios/MullvadVPN/Classes/AccessbilityIdentifier.swift
@@ -35,6 +35,8 @@ public enum AccessibilityIdentifier: String {
     case startUsingTheAppButton
     case problemReportAppLogsButton
     case problemReportSendButton
+    case relayStatusCollapseButton
+    case settingsDoneButton
 
     // Cells
     case vpnSettingsCell
@@ -44,11 +46,16 @@ public enum AccessibilityIdentifier: String {
     case apiAccessCell
     case relayFilterOwnershipCell
     case relayFilterProviderCell
+    case wireGuardPortsCell
+    case wireGuardObfuscationCell
+    case udpOverTCPPortCell
+    case quantumResistantTunnelCell
 
     // Labels
     case headerDeviceNameLabel
     case connectionStatusLabel
     case welcomeAccountNumberLabel
+    case connectionPanelDetailLabel
 
     // Views
     case accountView
@@ -62,17 +69,20 @@ public enum AccessibilityIdentifier: String {
     case selectLocationView
     case selectLocationTableView
     case settingsTableView
+    case vpnSettingsTableView
     case tunnelControlView
     case problemReportView
     case problemReportSubmittedView
     case revokedDeviceView
     case welcomeView
     case deleteAccountView
+    case settingsContainerView
 
     // Other UI elements
     case connectionPanelInAddressRow
     case connectionPanelOutAddressRow
     case customSwitch
+    case customWireGuardPortTextField
     case dnsContentBlockersHeaderView
     case loginTextField
     case selectLocationSearchTextField
--- a/ios/MullvadVPN/Coordinators/ApplicationCoordinator.swift
+++ b/ios/MullvadVPN/Coordinators/ApplicationCoordinator.swift
@@ -777,6 +777,7 @@ final class ApplicationCoordinator: Coordinator, Presenting, RootContainerViewCo
         )
 
         let navigationController = CustomNavigationController()
+        navigationController.view.accessibilityIdentifier = .settingsContainerView
 
         let configurationTester = ProxyConfigurationTester(transportProvider: configuredTransportProvider)
 
--- a/ios/MullvadVPN/View controllers/Settings/SettingsViewController.swift
+++ b/ios/MullvadVPN/View controllers/Settings/SettingsViewController.swift
@@ -45,7 +45,8 @@ class SettingsViewController: UITableViewController, SettingsDataSourceDelegate 
             value: "Settings",
             comment: ""
         )
-        navigationItem.rightBarButtonItem = UIBarButtonItem(
+
+        let doneButton = UIBarButtonItem(
             systemItem: .done,
             primaryAction: UIAction(handler: { [weak self] _ in
                 guard let self else { return }
@@ -53,6 +54,8 @@ class SettingsViewController: UITableViewController, SettingsDataSourceDelegate 
                 delegate?.settingsViewControllerDidFinish(self)
             })
         )
+        doneButton.accessibilityIdentifier = .settingsDoneButton
+        navigationItem.rightBarButtonItem = doneButton
 
         tableView.accessibilityIdentifier = .settingsTableView
         tableView.backgroundColor = .secondaryColor
--- a/ios/MullvadVPN/View controllers/Tunnel/ConnectionPanelView.swift
+++ b/ios/MullvadVPN/View controllers/Tunnel/ConnectionPanelView.swift
@@ -29,6 +29,7 @@ class ConnectionPanelView: UIView {
 
     var connectedRelayName = "" {
         didSet {
+            collapseButton.accessibilityIdentifier = .relayStatusCollapseButton
             collapseButton.setTitle(connectedRelayName, for: .normal)
             collapseButton.accessibilityLabel = NSLocalizedString(
                 "RELAY_ACCESSIBILITY_LABEL",
@@ -185,6 +186,7 @@ class ConnectionPanelAddressRow: UIView {
 
     private let detailTextLabel: UILabel = {
         let detailTextLabel = UILabel()
+        detailTextLabel.accessibilityIdentifier = .connectionPanelDetailLabel
         detailTextLabel.font = .systemFont(ofSize: 17)
         detailTextLabel.textColor = .white
         detailTextLabel.translatesAutoresizingMaskIntoConstraints = false
--- a/ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsCellFactory.swift
+++ b/ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsCellFactory.swift
@@ -97,6 +97,7 @@ final class VPNSettingsCellFactory: CellFactoryProtocol {
                 comment: ""
             )
 
+            cell.textField.accessibilityIdentifier = .customWireGuardPortTextField
             cell.accessibilityIdentifier = item.accessibilityIdentifier
             cell.applySubCellStyling()
 
--- a/ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsDataSource.swift
+++ b/ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsDataSource.swift
@@ -446,6 +446,7 @@ final class VPNSettingsDataSource: UITableViewDiffableDataSource<
             comment: ""
         )
 
+        header.accessibilityIdentifier = .wireGuardPortsCell
         header.titleLabel.text = title
         header.accessibilityCustomActionName = title
         header.infoButtonHandler = { [weak self] in
@@ -488,6 +489,7 @@ final class VPNSettingsDataSource: UITableViewDiffableDataSource<
             comment: ""
         )
 
+        header.accessibilityIdentifier = .wireGuardObfuscationCell
         header.titleLabel.text = title
         header.accessibilityCustomActionName = title
         header.didCollapseHandler = { [weak self] header in
@@ -516,6 +518,7 @@ final class VPNSettingsDataSource: UITableViewDiffableDataSource<
             comment: ""
         )
 
+        header.accessibilityIdentifier = .udpOverTCPPortCell
         header.titleLabel.text = title
         header.accessibilityCustomActionName = title
         header.didCollapseHandler = { [weak self] header in
@@ -545,6 +548,7 @@ final class VPNSettingsDataSource: UITableViewDiffableDataSource<
             comment: ""
         )
 
+        header.accessibilityIdentifier = .quantumResistantTunnelCell
         header.titleLabel.text = title
         header.accessibilityCustomActionName = title
         header.didCollapseHandler = { [weak self] header in
--- a/ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsViewController.swift
+++ b/ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsViewController.swift
@@ -38,6 +38,7 @@ class VPNSettingsViewController: UITableViewController, VPNSettingsDataSourceDel
     override func viewDidLoad() {
         super.viewDidLoad()
 
+        tableView.accessibilityIdentifier = .vpnSettingsTableView
         tableView.backgroundColor = .secondaryColor
         tableView.separatorColor = .secondaryColor
         tableView.rowHeight = UITableView.automaticDimension
--- a/ios/MullvadVPNUITests/Info.plist
+++ b/ios/MullvadVPNUITests/Info.plist
@@ -20,6 +20,8 @@
 	<string>$(IOS_DEVICE_PIN_CODE)</string>
 	<key>NoTimeAccountNumber</key>
 	<string>$(NO_TIME_ACCOUNT_NUMBER)</string>
+	<key>ShouldBeReachableDomain</key>
+	<string>$(SHOULD_BE_REACHABLE_DOMAIN)</string>
 	<key>TestDeviceIdentifier</key>
 	<string>$(TEST_DEVICE_IDENTIFIER_UUID</string>
 </dict>
--- a/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift
+++ b/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift
@@ -34,6 +34,7 @@ class FirewallAPIClient {
             "label": sessionIdentifier,
             "from": firewallRule.fromIPAddress,
             "to": firewallRule.toIPAddress,
+            "protocols": firewallRule.protocolsAsStringArray(),
         ]
 
         var requestError: Error?
@@ -80,7 +81,6 @@ class FirewallAPIClient {
 
         var request = URLRequest(url: removeRulesURL)
         request.httpMethod = "DELETE"
-        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
 
         var requestResponse: URLResponse?
         var requestError: Error?
--- a/ios/MullvadVPNUITests/Networking/FirewallRule.swift
+++ b/ios/MullvadVPNUITests/Networking/FirewallRule.swift
@@ -30,6 +30,10 @@ struct FirewallRule {
         self.protocols = protocols
     }
 
+    public func protocolsAsStringArray() -> [String] {
+        return protocols.map { $0.rawValue }
+    }
+
     /// Make a firewall rule blocking API access for the current device under test
     public static func makeBlockAPIAccessFirewallRule() throws -> FirewallRule {
         let deviceIPAddress = try Networking.getIPAddress()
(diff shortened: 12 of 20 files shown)


niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955)* status: 0 of 19 files reviewed, 1 unresolved discussion

___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nsx5S36-9sQnC1BQLnd:-Nsx5S36-9sQnC1BQLne:b-fg00hq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/52dc88cd1c3a58735dcb7963254536e1e6d933a2/ios/MullvadVPNUITests/RelayTests.swift#L16)):*
> ```Swift
>         super.setUp()
> 
>         FirewallAPIClient().removeRules()
> ```

Removing rules in both `setUp` and `tearDown` is a compromise. Because the firewall API cannot be reached when connected to VPN, and it's difficult and extends the time tests need to run if trying to make sure to always disconnect before invoking `removeRules()` from `tearDown`. In a perfect world `tearDown` would always make sure to disconnect from relay if connected.

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955)* status: 0 of 19 files reviewed, 1 unresolved discussion

___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nsx5S36-9sQnC1BQLnd:-Nsx8C8G-JFs8dRK4yFJ:b-l17hvc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/52dc88cd1c3a58735dcb7963254536e1e6d933a2/ios/MullvadVPNUITests/RelayTests.swift#L16)):*

Removing rules in both `setUp` and `tearDown` is a compromise. Because the firewall API cannot be reached when connected to VPN, and it's difficult and extends the time tests need to run if trying to make sure to always disconnect before invoking `removeRules()` from `tearDown`. In a perfect world `tearDown` would always make sure to disconnect from relay if connected.
</blockquote></details>

Also `testWireGuardOverTCPAutomatically` is the only test in this suite creating a firewall rule but `removeRules()` runs for all. So there could be logic for only running it when needed. I was thinking it doesn't hurt to run it even when it isn't needed and it doesn't add much time but would be great to hear your thoughts.

buggmagnet (changes requested):
> Reviewed 18 of 19 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955)* status: all files reviewed, 8 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nsx5S36-9sQnC1BQLnd:-Nt0Ge4w4K823z34G3Mg:bd32pyq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/52dc88cd1c3a58735dcb7963254536e1e6d933a2/ios/MullvadVPNUITests/RelayTests.swift#L16)):*

Also `testWireGuardOverTCPAutomatically` is the only test in this suite creating a firewall rule but `removeRules()` runs for all. So there could be logic for only running it when needed. I was thinking it doesn't hurt to run it even when it isn't needed and it doesn't add much time but would be great to hear your thoughts.
</blockquote></details>

I've been thinking about this since the discussion we had yesterday. 
If the firewall API is running on a local network, it should still be reachable when the VPN is turned on, since the `excludeLocalNetworks` on the VPN connection on iOS is on by default (i.e. local traffic won't be routed through the VPN)
We should double check this, and take a closer look why your original test attempts didn't work.
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 80 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0I9o45D1ncdtgWx6f:-Nt0I9o45D1ncdtgWx6g:bo50q92) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/RelayTests.swift#L80)):*
> ```Swift
> 
>     func testWireGuardOverTCPAutomatically() throws {
>         let wireGuardGot001RelayIPAddress = "185.213.154.66"
> ```

We can't really do this because the server IP might change in the future.
It's okay to leave it for now, but what we should probably be doing is either leveraging the JSON file that contains all the relays IPs  that we automatically download when we build the app, or do something similar to avoid having a static IP defined.
___
*[`ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift` line 83 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0J-tk2sNX1mw3jKML:-Nt0J-tk2sNX1mw3jKMM:bduxrm4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75c053739a33b0314b13951a9dce04cfd1fb52da/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift#L83)):*
> ```Swift
>         var request = URLRequest(url: removeRulesURL)
>         request.httpMethod = "DELETE"
>         request.setValue("application/json", forHTTPHeaderField: "Content-Type")
> ```

Why did we remove that line ? What was not working ?
___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 89 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0JhI14ge9AXFukO4N:-Nt0JhI14ge9AXFukO4O:b-pr9lpa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Networking/Networking.swift#L89)):*
> ```Swift
> 
>     /// Get configured domain to use for Internet connectivity checks
>     private static func getShouldBeReachableDomain() throws -> String {
> ```

nit
The name doesn't read super well, I think we can rename this `alwaysReachableDomain` and get the same intended idea.

With this addition, the checks become more readable too IMHO
```swift
XCTAssertTrue(try canConnectSocket(host: alwaysReachableDomain(), port: "80"))
```

What do you think ?
___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 154 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0KfoM2juaRflg2N-I:-Nt0KfoM2juaRflg2N-J:bscq3wm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Networking/Networking.swift#L154)):*
> ```Swift
>     }
> 
>     /// Verify that the device do not have Internet connectivity
> ```

nit 
`does not have`
___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 31 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0LihU44p2w4FkHgBb:-Nt0LihU44p2w4FkHgBc:b99fdls) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L31)):*
> ```Swift
>         var capturedTexts: [String] = []
> 
>         while -startTime.timeIntervalSinceNow < timeout {
> ```

I understand what you're trying to do, but it looks really convoluted.
It's not performance critical code, so we should do `Date().timeIntervalSince(startTime)` instead for readability's sake
___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 103 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0ZaMW-IkoURR7pRxq:-Nt0ZaMW-IkoURR7pRxr:b-z0xx1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L103)):*
> ```Swift
> 
>     /// Verify that the app attempts to connect over UDP before switching to TCP. For testing blocked UDP traffic.
>     @discardableResult func verifyConnectingOverTCPAfterUDPAttempts() -> Self {
> ```

nit
The logic in this function is a bit hard to follow, so the comments are very appreciated.
I'm wondering if we can make it easier to follow. I'll think about it.
___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 106 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0ZRUf19meNKTKx66D:-Nt0ZRUf19meNKTKx66E:b-iru6v8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L106)):*
> ```Swift
>         let connectionAttempts = waitForConnectionAttempts(3, timeout: 15)
> 
>         var i = 0
> ```

Can we rename this to something more meaningful ?

rablador (changes requested):
> Reviewed 18 of 19 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955)* status: all files reviewed, 15 unresolved discussions (waiting on @buggmagnet and @niklasberglund)

___
*[`ios/MullvadVPN/Coordinators/ApplicationCoordinator.swift` line 770 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0LInl4fbtsg2xXwuV:-Nt0LInl4fbtsg2xXwuW:b-9yx2yr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPN/Coordinators/ApplicationCoordinator.swift#L770)):*
> ```Swift
> 
>         let navigationController = CustomNavigationController()
>         navigationController.view.accessibilityIdentifier = .settingsContainerView
> ```

How come this isn't set on the view controller inside `SettingsCoordinator` instead?
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsViewController.swift` line 58 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0MdVf8F2UUTccUWAI:-Nt0MdVf8F2UUTccUWAJ:b-d5klvf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPN/View%20controllers/Settings/SettingsViewController.swift#L58)):*
> ```Swift
>         )
>         doneButton.accessibilityIdentifier = .settingsDoneButton
>         navigationItem.rightBarButtonItem = doneButton
> ```

I know we discussed the issues with navigation bar buttons before, but do we really need to go through these hoops to make it work?
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nsx5S36-9sQnC1BQLnd:-Nt0NBuV9UeXwr3-osDe:b-izop0e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/52dc88cd1c3a58735dcb7963254536e1e6d933a2/ios/MullvadVPNUITests/RelayTests.swift#L16)):*

Also `testWireGuardOverTCPAutomatically` is the only test in this suite creating a firewall rule but `removeRules()` runs for all. So there could be logic for only running it when needed. I was thinking it doesn't hurt to run it even when it isn't needed and it doesn't add much time but would be great to hear your thoughts.
</blockquote></details>

I'd say run `removeRules()` only when you needed, unless it's needed often enough to warrant a general inclusion. In general, I think we shouldn't proactively add stuff that we don't actually use.
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 59 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0Q2EO4YPTrUMQyY8x:-Nt0Q2EO4YPTrUMQyY8y:b-c73h6a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/RelayTests.swift#L59)):*
> ```Swift
>         VPNSettingsPage(app)
>             .tapWireGuardObfuscationExpandButton()
>             .tapWireGuardObfuscationValueOnCell()
> ```

Nit: I'd prefer omitting `Value` here, as I accidentally read it as "tap the wireguard value on the cell".
___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 67 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0R0c834WQXaeeL9Rr:-Nt0R0c834WQXaeeL9Rs:b-b4dty) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Networking/Networking.swift#L67)):*
> ```Swift
> 
>     /// Get configured ad serving domain as URL object
>     private static func getAdServingDomainURL() -> URL? {
> ```

Old code, but I think function should be based on `getAdServingDomain()` below, eg.:



_Code snippet:_
```Swift
guard let adServingDomain = getAdServingDomain(),
      let adServingDomainURL = URL(string: adServingDomain) else {
    XCTFail("Ad serving domain not configured")
    return nil
}
```
___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 29 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0aLbU2jhJNOIzGp9b:-Nt0aLbV5CupLGrf4xb-:bvhkrl1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L29)):*
> ```Swift
> 
>         let inAddressRow = app.otherElements[AccessibilityIdentifier.connectionPanelInAddressRow]
>         var capturedTexts: [String] = []
> ```

`capturedTexts` doesn't seem to be used for anything.
___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 42 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0_qx01AHxw2LaCPs6:-Nt0_qx01AHxw2LaCPs7:b-6ghb40) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L42)):*
> ```Swift
>             if let currentText = inAddressRow.value as? String {
>                 // Skip initial label value with IP address only - no port or protocol
>                 if currentText.contains(" ") == false {
> ```

Nit: More common to use `guard` when returning early.
___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 106 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0ZRUf19meNKTKx66D:-Nt0bMC-ES_dE6YI_pZp:bkvl4lj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L106)):*

Can we rename this to something more meaningful ?
</blockquote></details>

Perhaps the loops could be rewritten to include the index intead, eg.:



_Code snippet:_
```Swift
 for (attemptCount, attempt) in connectionAttempts {
    if attemptCount < 2 {
        XCTAssertEqual(attempt.protocolName, "UDP")
    } else if attemptCount == 2 {
        XCTAssertEqual(attempt.protocolName, "TCP")
    } else {
        XCTFail("Unexpected connection attempt")
    }
}
```
___
*[`ios/MullvadVPNUITests/Pages/VPNSettingsPage.swift` line 17 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0c3TQ13r3wHaLMduf:-Nt0c3TQ13r3wHaLMdug:b-53hq3t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/VPNSettingsPage.swift#L17)):*
> ```Swift
>     }
> 
>     private func cellExpandButton(_ cellAccessiblityIdentifier: AccessibilityIdentifier) -> XCUIElement {
> ```

Nit: We look for `AccessibilityIdentifier.collapseButton`, but the function name is `cellEXPANDButton` 🙃

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955)* status: 14 of 20 files reviewed, 15 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/Coordinators/ApplicationCoordinator.swift` line 770 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0LInl4fbtsg2xXwuV:-NtGwJUt7DiaLNTP5hCG:b-s6msf7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPN/Coordinators/ApplicationCoordinator.swift#L770)):*

How come this isn't set on the view controller inside `SettingsCoordinator` instead?
</blockquote></details>

I found it hard to follow the code here 😊 `SettingsCoordinator` sounds like a better place to set it but it's not a view. You mean it could be set inside `SettingsCoordinator` on `presentedViewController`?
___
*[`ios/MullvadVPN/View controllers/Settings/SettingsViewController.swift` line 58 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0MdVf8F2UUTccUWAI:-NtGyu6b0WCrhukHE446:bdq0ysl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPN/View%20controllers/Settings/SettingsViewController.swift#L58)):*

I know we discussed the issues with navigation bar buttons before, but do we really need to go through these hoops to make it work?
</blockquote></details>

This is to set accessibility identifier for the "Done" button in the upper right when in settings view 


`UIBarButtonItem` don't have any initialiser accepting accessibility identifier so need to store it as a variable and set `accessibilityIdentifier`
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nsx5S36-9sQnC1BQLnd:-Nt0rsXkAzZUYpugTIy9:b-5bezxx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/52dc88cd1c3a58735dcb7963254536e1e6d933a2/ios/MullvadVPNUITests/RelayTests.swift#L16)):*

I'd say run `removeRules()` only when you needed, unless it's needed often enough to warrant a general inclusion. In general, I think we shouldn't proactively add stuff that we don't actually use.
</blockquote></details>

@buggmagnet true, but I think it makes sense that it's failing because of the workaround we use to get around the local network restriction? `8.8.8.8` is pointing to a local IP address. So it is treated as an external IP address even though the machine is on the local network.

@rablador sounds good to me! Have updated it to run only before and after the test that actually create firewall rules.
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 59 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0Q2EO4YPTrUMQyY8x:-NtH03X3AdYwdV10ayBI:b28m7bk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/RelayTests.swift#L59)):*

Nit: I'd prefer omitting `Value` here, as I accidentally read it as "tap the wireguard value on the cell".
</blockquote></details>

I like it. Have removed "Value" from the names.
___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 80 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0I9o45D1ncdtgWx6f:-NtH07cV85xV_Wh5UdpE:b-xatl23) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/RelayTests.swift#L80)):*

We can't really do this because the server IP might change in the future.
It's okay to leave it for now, but what we should probably be doing is either leveraging the JSON file that contains all the relays IPs  that we automatically download when we build the app, or do something similar to avoid having a static IP defined.
</blockquote></details>

See updated implementation where the test first connects to the relay to extract its IP address from the UI
___
*[`ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift` line 83 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0J-tk2sNX1mw3jKML:-Nt0t3FnEN-RMbemr_vH:b-h7bvb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/75c053739a33b0314b13951a9dce04cfd1fb52da/ios/MullvadVPNUITests/Networking/FirewallAPIClient.swift#L83)):*

Why did we remove that line ? What was not working ?
</blockquote></details>

It was added by mistake before. This request don't have any body so there's no need to specify a content type.
___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 67 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0R0c834WQXaeeL9Rr:-NtH2yHhCh8WWL8Wq0Aw:b-cg8whm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Networking/Networking.swift#L67)):*

Old code, but I think function should be based on `getAdServingDomain()` below, eg.:
</blockquote></details>

Totally agree. I found that `getAdServingDomainURL` isn't even used though so removed it.
___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 89 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0JhI14ge9AXFukO4N:-NtH0hfa9ajbEqmN2LWq:byxv7r2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Networking/Networking.swift#L89)):*

nit
The name doesn't read super well, I think we can rename this `alwaysReachableDomain` and get the same intended idea.

With this addition, the checks become more readable too IMHO
```swift
XCTAssertTrue(try canConnectSocket(host: alwaysReachableDomain(), port: "80"))
```

What do you think ?
</blockquote></details>

I like it 👍 See updated suggestion
___
*[`ios/MullvadVPNUITests/Networking/Networking.swift` line 154 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0KfoM2juaRflg2N-I:-NtH3AC0B-CFuNz5Qa7O:b-gayzjn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Networking/Networking.swift#L154)):*

nit 
`does not have`
</blockquote></details>

Thanks, I've always been confused about do and does 😄 Will probably make this mistake again
___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 29 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0aLbU2jhJNOIzGp9b:-NtH3VNS1rPCGVKTqSk8:b-rkbpdg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L29)):*

`capturedTexts` doesn't seem to be used for anything.
</blockquote></details>

True, it was left from an earlier implementation and I forgot to remove it. Removed now.
___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 31 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0LihU44p2w4FkHgBb:-NtH4Zbc5aMyTZpS_-Q4:b-qcittk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L31)):*

I understand what you're trying to do, but it looks really convoluted.
It's not performance critical code, so we should do `Date().timeIntervalSince(startTime)` instead for readability's sake
</blockquote></details>

Good idea 👍 Much easier to read
___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 42 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0_qx01AHxw2LaCPs6:-NtH40IEAdEYpZDpyE2B:b-812wt5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L42)):*

Nit: More common to use `guard` when returning early.
</blockquote></details>

Updated to use `guard`
___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 103 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0ZaMW-IkoURR7pRxq:-NtHHUXsDXkxW5bYnPa9:b-p8uvlg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L103)):*

nit
The logic in this function is a bit hard to follow, so the comments are very appreciated.
I'm wondering if we can make it easier to follow. I'll think about it.
</blockquote></details>

I totally agree and was thinking about this too. One way to make it easier to follow could be breaking out the asserts to a separate function accepting an array of protocol names and their order and the attempts array. Something like `assertAttemptOrder(connectionAttempts, ["UDP", "UDP", "TCP"]`. Then in this function it would just be something like:
```
            if connectionAttempts.count == 3 { // Expected retries flow
                assertAttemptOrder(connectionAttempts, ["UDP", "UDP", "TCP"]
                }
            } else if connectionAttempts.count == 2 { // Most of the times this incorrect flow is shown
                assertAttemptOrder(connectionAttempts, ["UDP", "TCP"]
                }
            } else {
                XCTFail("Unexpected number of connection attempts")
            }
```
___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 106 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0ZRUf19meNKTKx66D:-NtHJA5P9qTN6eTDhaRD:b-hqy30e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L106)):*

Perhaps the loops could be rewritten to include the index intead, eg.:
</blockquote></details>

I'm all for longer descriptive names 👍 Renamed `i` to `attemptIndex`. @rablador I like it. See updated syntax. Is this what you meant?
___
*[`ios/MullvadVPNUITests/Pages/VPNSettingsPage.swift` line 17 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0c3TQ13r3wHaLMduf:-NtHGFzT7FBuMbotzD3Q:b-hd5q7y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPNUITests/Pages/VPNSettingsPage.swift#L17)):*

Nit: We look for `AccessibilityIdentifier.collapseButton`, but the function name is `cellEXPANDButton` 🙃
</blockquote></details>

Yea this bothers me too. Actually the button isn't a collapse button OR expand button but both 😊 I can't come up with a better convention than naming them "expand collapse button". There should be a better name?

buggmagnet (approved):
:lgtm:

> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-:-NtQFQ_2CeIaKmwluntV:bnfp4nl)* status: all files reviewed, 12 unresolved discussions (waiting on @niklasberglund and @rablador)

___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 106 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-NtQF6LF7IoPgpudMCjb:-NtQF6LF7IoPgpudMCjc:b3xij97) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0cb717062e40f45ed85f2eb48ca601b01623ae67/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L106)):*
> ```Swift
>         // Should do three connection attempts but due to UI bug sometimes only two are displayed, sometimes all three
>         if connectionAttempts.count == 3 { // Expected retries flow
>             for (attemptIndex, attempt) in connectionAttempts.enumerated() {
> ```

That is much better now :thumbsup:

rablador (changes requested):
:lgtm:

> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-:-NtUzU-C7VPBsGTl9z9a:bnfp4nl)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @niklasberglund)

___
*[`ios/MullvadVPN/Coordinators/ApplicationCoordinator.swift` line 770 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5955#-Nt0LInl4fbtsg2xXwuV:-NtUzy_e-UxiI4XfVSWA:b1keuvh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fddf8d50c69c1432a067b950032b9cdb1ec31c78/ios/MullvadVPN/Coordinators/ApplicationCoordinator.swift#L770)):*

I found it hard to follow the code here 😊 `SettingsCoordinator` sounds like a better place to set it but it's not a view. You mean it could be set inside `SettingsCoordinator` on `presentedViewController`?
</blockquote></details>

Either that or `SettingsViewController`, which also resides there.

PR #5962, opened 2024-03-15
Fix memory leak at LocationNode

mojganii:
It seems that maintaining a strong reference from a parent to a location node can cause a memory leak. Making that reference weak allows the automatic reference counting (ARC) system to release the memory when the object is no longer needed or exists.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5962)


rablador (approved):
Nice!

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5962#-:-Nt18ZhP-5jw3HDOQ4kG:b-83sjoh)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #5982, opened 2024-03-19
Replace duplicated cascades of case statements with utility methods on State

acb-mv:
This replaces recurring cascades of case statements that mutate `State`'s associated values with one utility method on State: `mutateAssociatedData`, which take a function that might modify the appropriate associated value, and if the current state has such a value, applies it and returns whether a modification took place.  The function is given a `StateAssociatedData` type, which is a protocol encompassing both the connection and blocked state data's common fields.

The benefit of this change is increased separation of concerns: code that cares only about the connection/error state no longer needs to know which states exist, which will make changing the logic of the state machine simpler, as there will be fewer places where changes are needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5982)

### DIFF
--- a/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift
+++ b/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift
@@ -108,6 +108,9 @@ class PacketTunnelProvider: NEPacketTunnelProvider {
                 if connectionState.connectionAttemptCount > 1 {
                     return
                 }
+            case let .negotiatingKey(connectionState):
+                try await startPostQuantumKeyExchange()
+                return
             default:
                 break
             }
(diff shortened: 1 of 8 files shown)


buggmagnet (changes requested):
> Reviewed 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5982)* status: all files reviewed, 2 unresolved discussions (waiting on @acb-mv)

___
*[`ios/PacketTunnelCore/Actor/State.swift` line 99 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5982#-NtWAp8SCDuWuUJEo8qT:-NtWAp8SCDuWuUJEo8qU:b3s20oe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7fe698fb3fdabb1e862a2e3f74bd31f3bafb8eb8/ios/PacketTunnelCore/Actor/State.swift#L99)):*
> ```Swift
> }
> 
> // perhaps this should have a better name?
> ```

nit
This looks fine, but we should (ideally) be able to understand where it's used from the name alone. 
We could rename it `ConnectionStateAssociatedData` for example, although that might feel a bit redundant.
___
*[`ios/PacketTunnelCore/Actor/State+Extensions.swift` line 131 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5982#-NtWIzy46EGbPkMJPJBq:-NtWIzy46EGbPkMJPJBr:b-eqshov) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7fe698fb3fdabb1e862a2e3f74bd31f3bafb8eb8/ios/PacketTunnelCore/Actor/State%2BExtensions.swift#L131)):*
> ```Swift
>             var associatedData: StateAssociatedData = connState
>             modifier(&associatedData)
>             self = self.replacingConnectionData(with: associatedData as! ConnectionData)
> ```

It's a bit annoying that we have to do this, but at least we should remove the `swiftlint` warning here and down in the blocked state.

Let's add a comment that specifices that both `ConnectionData` and `BlockingData` do implement `StateAssociatedData` so the cast is guaranteed to work.

acb-mv (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5982)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/PacketTunnelCore/Actor/State.swift` line 99 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5982#-NtWAp8SCDuWuUJEo8qT:-NtWOKs-E-rz4RjhP35K:b-wq6gjh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7fe698fb3fdabb1e862a2e3f74bd31f3bafb8eb8/ios/PacketTunnelCore/Actor/State.swift#L99)):*

nit
This looks fine, but we should (ideally) be able to understand where it's used from the name alone. 
We could rename it `ConnectionStateAssociatedData` for example, although that might feel a bit redundant.
</blockquote></details>

Except that this is not specific to a connection state, it can also apply to a blockage/error state. While the type this belongs to is named `State` and protocols cannot live under namespaces, I'm using `State` as a prefix for the protocol name.  Perhaps in future, it may do to move `State` under `PacketTunnelActor` and call it `PacketTunnelActorStateAssociatedData` or some abbreviation thereof (`PacketTunnelActorStateData`?) 

Another option would be to make the name an adjective, which is something that goes well with protocols, though I haven't thought of a good one (`StateDataBearing` perhaps? Though that sounds awkward).

In any case, the comment is a leftover from before I renamed it and the name was worse, and I can delete it.
___
*[`ios/PacketTunnelCore/Actor/State+Extensions.swift` line 131 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5982#-NtWIzy46EGbPkMJPJBq:-NtWPU5W2Gux8qNW0Xe0:b57948q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7fe698fb3fdabb1e862a2e3f74bd31f3bafb8eb8/ios/PacketTunnelCore/Actor/State%2BExtensions.swift#L131)):*

It's a bit annoying that we have to do this, but at least we should remove the `swiftlint` warning here and down in the blocked state.

Let's add a comment that specifices that both `ConnectionData` and `BlockingData` do implement `StateAssociatedData` so the cast is guaranteed to work.
</blockquote></details>

Is there a format for this comment that will shut swiftlint up, or is this just for human consumption?

PR #5997, opened 2024-03-20
Improve git commit message check

faern:
The combined diff is not the easiest way to view this PR. Per commit is way more logical.

Commit 1: Change `max-subject-line-length` from 74 to 72. This was just an error when I initially added the job. I invalidly thought 74 was the default body line length (this is where vim and other git aware editors automatically insert a line break), but it is of course 72! So I'm just correcting this. Nothing major really.

Commit 2: Move the `Check for unicode whitespaces` step to run first. This is because the mristin/opinionated-commit-message action can sometimes fail on things we don't consider blocking/hard errors. Such as a correct imperative verb that does not happen to be in the action's wordlist. In these cases the job is aborted and the whitespace checking step is never executed, leading to a larger risk of accidentally merging invalid whitespaces.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5997)

### DIFF
--- a/.github/workflows/git-commit-message-style.yml
+++ b/.github/workflows/git-commit-message-style.yml
@@ -14,8 +14,8 @@ jobs:
         with:
           # Commit messages are allowed to be subject only, no body
           allow-one-liners: 'true'
-          # This action defaults to 50 char subjects, but 74 is fine.
-          max-subject-line-length: '74'
+          # This action defaults to 50 char subjects, but 72 is fine.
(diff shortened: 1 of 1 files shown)


raksooo (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/5997)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6009, opened 2024-03-22
Remove python scripts/dependencies from iOS directory

faern:
So, if I run the vulnerability scanner `osv-scanner` on our repository. It points out that `ios/requirements.txt` has a vulnerable version of `gdal` in it: https://osv.dev/vulnerability/PYSEC-2022-43065.

Upon further investigation my conclusion is that `ios/requirements.txt` has a lot of stuff in it that we don't actually need. As far as I understand, our only iOS python code is `ios/update_translations.py`, and this script only depends on `nslocalized`. `nslocalized` in turn only depend on `six`. So the remaining entries in `requirements.txt` seem to be just noise.

My theory on why they ended up there in the first place is that by default `pip freeze` that generate these files generate a list of all globally installed python packages. Or all packages installed in a venv if one is active. And the person generating this lockfile probably had a lot more packages installed than the ones purely needed for `update_translations.py`.

Can you please verify my assumptions by trying to use this script in whatever ways you usually do, but isolated to a venv with only the things from my updated requirements file are installed?

```
python3 -m venv ./python_env
source python_env/bin/activate
pip3 install -r requirements.txt

python3 update_translations.py
```

EDIT: Turns out we don't use this script at all. So the PR changed into just removing it altogether. Later, when we want to introduce translations, we can resurrect it if we feel like this solution is the best way to go. But we should also consider writing such tooling in other languages that we already use, to keep the tech stack a bit narrower.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6009)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6009)* status: 0 of 2 files reviewed, all discussions resolved

___
*[`.github/workflows/verify-locked-down-signatures.yml` line 13 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6009#-Nt_m7Ay98lOhNFm4-72:-Nt_m7Ay98lOhNFm4-73:b-sepppn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/df05999347398fbeff2756b1a82a5d325977e6a5/.github/workflows/verify-locked-down-signatures.yml#L13)):*
> ```YAML
>       - ci/keys/**
>       - ci/verify-locked-down-signatures.sh
>       - ios/requirements.txt
> ```

All lockfiles, or all files that affect supply chain security, must be listed here. Changes to them are not allowed by third parties, and commits changing them must be signed.

faern:
I now changed this PR to remove the iOS python script altogether. I learned that we never use it, and have no concrete plans to use it in the short term.

Whenever we implement translations, we should evaluate if we can use a different language than python to do this type of task, if it's still needed by then.

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6009#-:-Nta3c9TAuRMZhlnCGeq:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

buggmagnet (approved):
nice job !

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6009#-:-Ntu-JKu5qoV6RdGtBTq:b-jc3by8)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6015, opened 2024-03-22
Point the hickory-dns dependency to our own fork of the repo

faern:
We should avoid depending on git repositories outside our own control. The hickory-dns organization or repository can go away at any point in time. That will make it way harder to re-build old versions of the app.

I also asked the maintainers if they could publish a patch release, because we should avoid git dependencies altogether when we can :pray:

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6015)


dlon (approved):
:+1: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6015#-:-Ntag-b15LrjMcwfynfA:b-8fx2u7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6029, opened 2024-03-26
Update `deny.toml` files for new `version = 2` format, and other cleanups.

faern:
This PR does a bunch of housekeeping around our usage of `cargo deny`. I saw that running `cargo deny` produced a bunch of warnings, since we use deprecated fields. This PR fixes that and more.

This PR is probably easiest to review per commit!

* Add `version = 2` where suitable and remove deprecated fields
* Remove fields which were set to default values anyway. No need to explicitly set them, just noisy.
* Require signatures to changes to `test/deny.toml` - Arguably way less important than the main `deny.toml` but still in some ways affect supply chain security and licenses, so should be signed.
* Adds more targets. Since we do build this software for Android and iOS also, we should check it strictly
* No need to explicitly deny `atty` I feel. That crate has a CVE registered against it anyway, so any CVE checking tool will catch if we accidentally get it into our dependency tree again.
* Add comment about us not wanting to have git dependencies, and if we do they should be on our own github org.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6029)

### DIFF
--- a/.github/workflows/verify-locked-down-signatures.yml
+++ b/.github/workflows/verify-locked-down-signatures.yml
@@ -6,6 +6,7 @@ on:
       - .github/workflows/verify-locked-down-signatures.yml
       - Cargo.lock
       - deny.toml
+      - test/deny.toml
       - gui/package-lock.json
       - wireguard/libwg/go.sum
       - ci/keys/**
--- a/deny.toml
+++ b/deny.toml
@@ -103,7 +103,6 @@ skip-tree = []
 [sources]
 unknown-registry = "deny"
 unknown-git = "deny"
-allow-registry = ["https://github.com/rust-lang/crates.io-index"]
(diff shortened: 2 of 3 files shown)


MarkusPettersson98 (approved):
:lgtm: :sparkles:

> Reviewed 2 of 2 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, 2 of 2 files at r4, 2 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6029#-:-NtuCm147p-cFd-vnEcT:b-8pwgi6)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6034, opened 2024-03-27
Bump opinionated-commit-message github action to version with more verbs

faern:
Includes our contributions to the action with ~185 more allowed imperative verbs. That's also why we no longer need to specify any `additional-verbs`.

Pins the action to a commit hash rather than tag. For supply chain security. We might want to do this with way more third party actions, but this is mostly a test to try this out for now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6034)


raksooo (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6034)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6036, opened 2024-03-27
Add test for login to account with too many devices

niklasberglund:
Implementation of test logging in to account with too many devices. The test creates a temporary account, adds 5 devices to it and then verifies that the flow is functioning according to to Linear ticket.

To test this PR run the test `testLoginToAccountWithTooManyDevices` under `AccountTests`.


rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6036)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

buggmagnet (changes requested):
> Reviewed 6 of 8 files at r1, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6036)* status: all files reviewed, 2 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 21 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6036#-Nux6U455nG8jeTsYJ7w:-Nux6U455nG8jeTsYJ7x:b-cjd7bg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/310dda0e685f2d2cf641d6c84e688539dad0be76/ios/MullvadVPNUITests/AccountTests.swift#L21)):*
> ```Swift
>     }
> 
>     override func tearDown() {
> ```

A couple of things can be improved here
- We should call `super.tearDown` after we run our custom tear down logic to avoid unintended consequences
- The `if` check here doesn't make a lot of sense, we bind `temporaryAccountNumber` to clear it just after, but setting an optional to `nil` will never fail, so we don't need to do it in the if body
- If we used the `withError` variation of this function (which is empty below) we could  use`XCTUnwrap`instead.
- The `temporaryAccountNumber` is only ever used in `testLoginToAccountWithTooManyDevices` so it should stay there as a local variable instead, unless we plan to use it in other tests.

Here is what I am suggesting to do instead 
- Remove this custom `tearDown` function
- Remove the empty `tearDownWithError` function too (we can write it when we start using it)
- Remove the `temporaryAccountNumber` variable from this class
- Do the following change

```swift
func testLoginToAccountWithTooManyDevices() throws {
    let temporaryAccountNumber = try MullvadAPIWrapper().createAccount()
    defer {
        do {
            try MullvadAPIWrapper().deleteAccount(temporaryAccountNumber)
        } catch {
            XCTFail("Failed to delete account \(temporaryAccountNumber) using app API")
        }
    // Do the rest of the test here
    }

```
___
*[`ios/MullvadVPNUITests/Networking/MullvadAPIWrapper.swift` line 9 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6036#-Nux8oDm6ecy4SJORmIO:-Nux8oDm6ecy4SJORmIP:bzdho27) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/310dda0e685f2d2cf641d6c84e688539dad0be76/ios/MullvadVPNUITests/Networking/MullvadAPIWrapper.swift#L9)):*
> ```Swift
> //
> 
> import CryptoKit
> ```

nit
This makes me wonder why we don't use it instead of having a custom implementation in our codebase... :thinking: 
Good food for thought !

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6036)* status: 5 of 9 files reviewed, 2 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @rablador)

___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 21 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6036#-Nux6U455nG8jeTsYJ7w:-NuxXAOi-yjGN6lff5yx:b-s1im8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/310dda0e685f2d2cf641d6c84e688539dad0be76/ios/MullvadVPNUITests/AccountTests.swift#L21)):*
> We should call `super.tearDown` after we run our custom tear down logic to avoid unintended consequences

You mean run it after the code in this `tearDown` rather than before as it is now? Don't the current order make sense though because `super.tearDown()` will log out the user and the code in this `tearDown` implementation deletes the account? If we reverse the order then the account will be deleted while the user is still logged in which might lead to unintended consequences.


> The `if` check here doesn't make a lot of sense, we bind `temporaryAccountNumber` to clear it just after, but setting an optional to `nil` will never fail, so we don't need to do it in the if body

Yes very true. Should be `self.temporaryAccountNumber = nil` so that the member variable is reset before next test runs.

Since the temporary account number was only used by one test the code has been moved to that test method instead.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6036)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @niklasberglund)

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6036#-:-Nvk-5kF0ZYGsHiRKpqQ:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 21 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6036#-Nux6U455nG8jeTsYJ7w:-NvjwyDx5jN2TkV7XsyM:b-6ybk0s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/310dda0e685f2d2cf641d6c84e688539dad0be76/ios/MullvadVPNUITests/AccountTests.swift#L21)):*
> `super.tearDown()` will log out the user 
AFAICT we log out when a new test starts in the `setUp` function of `LoggedOutUITestCase` via `logoutIfLoggedIn`

> and the code in this `tearDown` implementation deletes the account?
We delete the account in `class func tearDown()`, so after the entire suite is ran. 

But in general, to prevent any weird behaviour, we should strive to always have the following structure (I know it sounds silly on paper, but I've seen many bugs in the past because this convention was not followed properly)
```swift
override func setUp() {
    super.setUp()
    // Our code
}

override func tearDown() {
    // Our code
    super.tearDown()
}
```
___
*[`ios/MullvadVPNUITests/AccountTests.swift` line 86 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6036#-NvjzTl10Xb1k3ls6l05:-NvjzTl10Xb1k3ls6l06:by7uni5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f9885902f35c1913fa0b0c85ab083284e42802e7/ios/MullvadVPNUITests/AccountTests.swift#L86)):*
> ```Swift
>                 XCTFail("Failed to delete account using app API")
>             }
>         }
> ```

`deleteAccount` already wraps itself in a `try/catch` block
We can change this to
```swift
        addTeardownBlock {
            try MullvadAPIWrapper().deleteAccount(temporaryAccountNumber)
        }

```



_Code quote:_
```Swift
        addTeardownBlock {
            do {
                try MullvadAPIWrapper().deleteAccount(temporaryAccountNumber)
            } catch {
                XCTFail("Failed to delete account using app API")
            }
        }
```

PR #6051, opened 2024-04-02
Lint against single use lifetimes

faern:
Removes superfluous lifetime definitions. Simplifying the code. Here is the docs for this lint: https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#single-use-lifetimes

Since we obviously had a pretty large number of these, I think the lint can help us keep code simpler. The only potential red flag I see is the lint docs saying "Also, there are some known false positives". I did not stumble over any false positives (that I noticed) when doing this small refactoring. If it turns out to be too troublesome we can disable the lint again. Or add `#[allow(single_use_lifetimes)]` to the places where it does not work as intended.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051)

### DIFF
--- a/talpid-core/src/dns/windows/tcpip.rs
+++ b/talpid-core/src/dns/windows/tcpip.rs
@@ -93,24 +93,24 @@ fn set_dns_inner(
         transaction,
         &guid_str,
         "Tcpip",
-        servers.iter().filter(|addr| addr.is_ipv4()),
+        servers.iter().filter(|addr| addr.is_ipv4()).copied(),
     )?;
 
     config_interface(
         transaction,
         &guid_str,
         "Tcpip6",
-        servers.iter().filter(|addr| addr.is_ipv6()),
+        servers.iter().filter(|addr| addr.is_ipv6()).copied(),
     )?;
 
     Ok(())
 }
 
-fn config_interface<'a>(
+fn config_interface(
     transaction: &Transaction,
     guid: &str,
     service: &str,
-    nameservers: impl Iterator<Item = &'a IpAddr>,
+    nameservers: impl Iterator<Item = IpAddr>,
 ) -> io::Result<()> {
     let nameservers = nameservers
         .map(|addr| addr.to_string())
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ explicit_outlives_requirements = "warn"
 missing_abi = "deny"
 unused_lifetimes = "warn"
 unused_macro_rules = "warn"
+single_use_lifetimes = "warn"
 
 
 [workspace.lints.clippy]
--- a/android/translations-converter/src/gettext/msg_string.rs
+++ b/android/translations-converter/src/gettext/msg_string.rs
@@ -78,7 +78,7 @@ where
     }
 }
 
-impl<'l, 'r> Add<&'r MsgString> for &'l MsgString {
+impl<'r> Add<&'r MsgString> for &MsgString {
     type Output = MsgString;
 
     fn add(self, other: &'r MsgString) -> Self::Output {
--- a/mullvad-cli/src/cmds/api_access.rs
+++ b/mullvad-cli/src/cmds/api_access.rs
@@ -439,7 +439,7 @@ mod pp {
         }
     }
 
-    impl<'a> std::fmt::Display for ApiAccessMethodFormatter<'a> {
+    impl std::fmt::Display for ApiAccessMethodFormatter<'_> {
         fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
             let write_status = |f: &mut std::fmt::Formatter<'_>, enabled: bool| {
                 if enabled {
--- a/mullvad-cli/src/cmds/custom_list.rs
+++ b/mullvad-cli/src/cmds/custom_list.rs
@@ -216,7 +216,7 @@ impl<'a> GeographicLocationConstraintFormatter<'a> {
     }
 }
 
-impl<'a> std::fmt::Display for GeographicLocationConstraintFormatter<'a> {
+impl std::fmt::Display for GeographicLocationConstraintFormatter<'_> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
         let unwrap_country = |country: Option<String>, constraint: &str| {
             country.unwrap_or(format!("{constraint} <invalid country>"))
--- a/mullvad-cli/src/cmds/proxies.rs
+++ b/mullvad-cli/src/cmds/proxies.rs
@@ -182,7 +182,7 @@ pub mod pp {
         pub custom_proxy: &'a CustomProxy,
     }
 
-    impl<'a> std::fmt::Display for CustomProxyFormatter<'a> {
+    impl std::fmt::Display for CustomProxyFormatter<'_> {
         fn fmt(&self, _: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
             match self.custom_proxy {
                 CustomProxy::Shadowsocks(shadowsocks) => {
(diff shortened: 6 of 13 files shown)


MarkusPettersson98 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051#-:-NuThIfFE4JWN-s1IyWJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051)* status: 12 of 13 files reviewed, all discussions resolved (waiting on @MarkusPettersson98)

___
*[`talpid-core/src/dns/windows/tcpip.rs` line 113 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051#-NuTtvSj2XlC2S_2C3Zp:-NuTtvSj2XlC2S_2C3Zq:b-eyzy3v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8055f26560e63c74120b999f4132c21e1066e170/talpid-core/src/dns/windows/tcpip.rs#L113)):*
> ```XML
>     guid: &str,
>     service: &str,
>     nameservers: impl Iterator<Item = &IpAddr>,
> ```

This is probably a false positive of this lint. I could not solve it. Since `impl Trait` requires named lifetimes, but that lifetime becomes a single use lifetime, and the lint will trigger.. But since `IpAddr` is a `Copy` type, the more idiomatic thing to do would be to iterate over them by value. So I decided to solve the problem like this instead.

If we in the future hit the same issue with a non-copy type we can just ignore the lint on a case by case basis with `#[allow(single_use_lifetimes)]`

MarkusPettersson98 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051#-:-NuU4NDC390cB7scgPhE:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`talpid-core/src/dns/windows/tcpip.rs` line 113 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051#-NuTtvSj2XlC2S_2C3Zp:-NuU46Nz8BryOUON7PBm:bug3vu5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8055f26560e63c74120b999f4132c21e1066e170/talpid-core/src/dns/windows/tcpip.rs#L113)):*

This is probably a false positive of this lint. I could not solve it. Since `impl Trait` requires named lifetimes, but that lifetime becomes a single use lifetime, and the lint will trigger.. But since `IpAddr` is a `Copy` type, the more idiomatic thing to do would be to iterate over them by value. So I decided to solve the problem like this instead.

If we in the future hit the same issue with a non-copy type we can just ignore the lint on a case by case basis with `#[allow(single_use_lifetimes)]`
</blockquote></details>

I think this is a reasonable take :shipit:

talpid-core/src/dns/windows/tcpip.rs
```diff
@@ -93,24 +93,24 @@ fn set_dns_inner(
         transaction,
         &guid_str,
         "Tcpip",
-        servers.iter().filter(|addr| addr.is_ipv4()),
+        servers.iter().copied().filter(|addr| addr.is_ipv4()),
```
Serock3, line 96: Super tiny nit:
```suggestion
        servers.iter().filter(|addr| addr.is_ipv4()).copied(),
```

talpid-core/src/dns/windows/tcpip.rs
```diff
@@ -93,24 +93,24 @@ fn set_dns_inner(
...
     )?;
 
     config_interface(
         transaction,
         &guid_str,
         "Tcpip6",
-        servers.iter().filter(|addr| addr.is_ipv6()),
+        servers.iter().copied().filter(|addr| addr.is_ipv6()),
```
Serock3, line 103: Same here
```suggestion
        servers.iter().filter(|addr| addr.is_ipv6()).copied(),
```

Serock3 (approved):
The changes made here seem to make the code more readable, so I am in favor

Serock3 (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051#-:-NuUAVIO4rsFxHGJUGXn:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[`talpid-core/src/dns/windows/tcpip.rs` line 113 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051#-NuTtvSj2XlC2S_2C3Zp:-NuUAWmf2gkAowXx41UY:b-cstnkg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8055f26560e63c74120b999f4132c21e1066e170/talpid-core/src/dns/windows/tcpip.rs#L113)):*

I think this is a reasonable take :shipit:
</blockquote></details>

I agree

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051)* status: all files reviewed, 2 unresolved discussions (waiting on @Serock3)

___
*[`talpid-core/src/dns/windows/tcpip.rs` line 96 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051#gh-1547975878:-NuUcXuBBW5WiQLWeCkC:b-vq67vj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/da24e37dcd621a2b11241a1aa672045ae53d84a8/talpid-core/src/dns/windows/tcpip.rs#L96)):*

Super tiny nit:
```suggestion
        servers.iter().filter(|addr| addr.is_ipv4()).copied(),
```
</blockquote></details>

Why is this better? I suppose you mean because then we don't need to copy the items that don't match the filter? But these types are super tiny, and the program need to read their memory anyway to compute `is_ipv4`, so I'm sure the iterator chain will dereference the memory of all items not matching the filter anyway.

`Copy` in general is usually for small simple types, so copying them around is usually not a noticeable cost. I would very much understand the reasoning if it was `.cloned()` we were talking about.

I can change the code, because I don't have a strong opinion. I just want to know what your rationale is here.

Serock3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051)* status: all files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[`talpid-core/src/dns/windows/tcpip.rs` line 96 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051#gh-1547975878:-NuYJVjJBypIMVuD3mDt:b-2ope7b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/da24e37dcd621a2b11241a1aa672045ae53d84a8/talpid-core/src/dns/windows/tcpip.rs#L96)):*

Why is this better? I suppose you mean because then we don't need to copy the items that don't match the filter? But these types are super tiny, and the program need to read their memory anyway to compute `is_ipv4`, so I'm sure the iterator chain will dereference the memory of all items not matching the filter anyway.

`Copy` in general is usually for small simple types, so copying them around is usually not a noticeable cost. I would very much understand the reasoning if it was `.cloned()` we were talking about.

I can change the code, because I don't have a strong opinion. I just want to know what your rationale is here.
</blockquote></details>

My rationale is that I thought it said`.cloned()` :slightly_smiling_face:, if it's a `Copy` type then it doesn't really matter. Putting the filter first looks a bit more natural to me but that's just taste I guess.

Serock3 (approved):
> Reviewed 12 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6051)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6064, opened 2024-04-03
Add connection retry logic test

niklasberglund:
Implemented connection retry logic test which blocks all connections to a relay then let the app connect to it and observes the connection attempts made by the app in order to verify that they're done in the correct order.

Note that there's a UI bug which most of the time prevent one of the first two UDP connection attempts from being shown in the app even though the attempt is actually made. Therefor the test also allows for this incorrect flow of attempts where only one UDP connection attempt is displayed instead of two.

To test the changes in this PR your phone must be on the `app-team-ios-tests` wifi. Run `testConnectionRetryLogic` under `RelayTests`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6064)

### DIFF
--- a/ios/MullvadVPN/Classes/AccessbilityIdentifier.swift
+++ b/ios/MullvadVPN/Classes/AccessbilityIdentifier.swift
@@ -17,6 +17,7 @@ public enum AccessibilityIdentifier: String {
     case cancelButton
     case connectionPanelButton
     case collapseButton
+    case expandButton
     case createAccountButton
     case deleteButton
     case disconnectButton
@@ -57,7 +58,8 @@ public enum AccessibilityIdentifier: String {
 
     // Labels
     case headerDeviceNameLabel
-    case connectionStatusLabel
+    case connectionStatusConnectedLabel
+    case connectionStatusNotConnectedLabel
     case welcomeAccountNumberLabel
     case connectionPanelDetailLabel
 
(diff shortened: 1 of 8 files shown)


niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6064)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadVPNUITests/Pages/TunnelControlPage.swift` line 148 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6064#-NucQu5b2ldhYLW5SvPr:-NucimJk2sljAz1XVf1X:bqg3wep) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/29fdd0fba7fc73935b6e702b6f2f38c3054a8d81/ios/MullvadVPNUITests/Pages/TunnelControlPage.swift#L148)):*

Nit: The last three checks do the same thing. Extact to function or closure? Or perhaps not, not sure if it'd make it better... :)
</blockquote></details>

In a way it'd be smart to break it out to a function but it's also important that it's easy to see what is tested. If breaking things out to functions it's more difficult to quickly glance and see what tests are doing. This first for loop will hopefully be removed if the bug is fixed :smile:

PR #6082, opened 2024-04-08
Fix all bridges being discarded before selection

MarkusPettersson98:
Bridges are no longer filtered out solely based on their distance from the 'origin'. Instead, we simply order them by distance from 'origin' and consider the closest 5 bridges. They will still be assigned weights proportional to the distance from the origin.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6082)

### DIFF
--- a/mullvad-relay-selector/src/relay_selector/mod.rs
+++ b/mullvad-relay-selector/src/relay_selector/mod.rs
@@ -958,10 +958,8 @@ impl RelaySelector {
         relays: Vec<Relay>,
         location: T,
     ) -> Option<Relay> {
-        /// Minimum number of bridges to keep for selection when filtering by distance.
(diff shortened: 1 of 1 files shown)


Serock3 (approved):
As we discussed earlier, I am not a fan of this weight fn. If we remove the max distance filter, we should probably not also bump the min count to 10, as that likely will change the relative weight of the closes bridges in practice. I suggest keeping it at 5 and figuring out a smarter weight fn/algorithm in the future.
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6082#-:-NuwnsZy64B7wuSuTLi-:bimqeyc)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6089, opened 2024-04-08
Include 100.64.0.0/10 in ALLOWED_LAN_NETS

### LINKED ISSUE #6086: 100.64.0.0/10 is not included in allowed LAN ranges
### Is it a bug?

- [X] I know this is an issue with the app, and contacting Mullvad support is not relevant.

### I have checked if others have reported this already

- [X] I have checked the issue tracker to see if others have reported similar issues.

### Current Behavior

Although 100.64.0.0/100 is part of the IPv4 reserved address blocks used for internal networks, it is not included in the list of allowed addresses when LAN access is unblocked.

Strictly speaking that range is for carrier-grade NAT, so it's debatable whether or not that should be considered LAN.

However, not adding this manually to to nftables means I can't use tailscale and mullvad concurrently.

I'm aware of tailscale's mullvad integration, but unfortunately since I use headscale and not the tailscale service, I can't yet utilize this functionality.

### Expected Behavior

With tailscale enabled and connected, `mullvad lan set allow`, when connecting to mullvad traffic on the tailscale interface should not be blocked.

I can work around this using the following firewall rule:

```
    table inet mullvadExclusions {
      chain mullvadAllowTailsacle {
        type route hook output priority -1; policy accept;
        ip daddr 100.64.0.0/10 ct mark set 0x00000f41 meta mark set 0x6d6f6c65;
      }
```

but it would be less brittle and more secure if I would not have to do that statically, since that would still be allowed if `mullvad lan set block` is configured.

### Steps to Reproduce

1. `tailscale up`
2. `mullvad lan set allow`
3. `tailscale ping ...` (works)
4. `mullvad connect`
5. `tailscale ping ...` (no response)
6. `mullvad disconnect`
7. `tailscale ping ...` (works)

### Failure Logs

_No response_

### Operating system version

NixOS unstable (pre release 24.05)

### Mullvad VPN app version

2023.6

### Additional Information

_No response_

chaeynz:
I added the 100.64.0.0/10 address space to ALLOWED_LAN_NETS.

100.64.0.0/10 is reserved by IANA as Shared Address Space, is filtered on the public Internet through routing filters (May not be advertised) and can therefore be treated as local.
https://datatracker.ietf.org/doc/html/rfc6598

It mainly serves ISPs, but can also be used on the internal network by organizations of any kind.

Please merge my PR, so that users will be able to access parts of their network where this address space is used, while still being connected to the Mullvad VPN.

Closes #6086

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6089)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Line wrap the file at 100 chars.                                              Th
 - Add automatic MTU detection for desktop platforms. This currently only uses information about
   dropped packets and does not take fragmentation into account.
 - Add ability to import server IP overrides in GUI.
+- Allow the 100.64.0.0/10 RFC6598 shared address space in addition to the other networks allowed when local network sharing is enabled.
 
 #### Android
 - Add support for all screen orientations.
--- a/talpid-core/src/firewall/mod.rs
+++ b/talpid-core/src/firewall/mod.rs
@@ -25,9 +25,10 @@ mod imp;
 pub use self::imp::Error;
 
 /// When "allow local network" is enabled the app will allow traffic to and from these networks.
-pub(crate) static ALLOWED_LAN_NETS: Lazy<[IpNetwork; 6]> = Lazy::new(|| {
+pub(crate) static ALLOWED_LAN_NETS: Lazy<[IpNetwork; 7]> = Lazy::new(|| {
     [
         IpNetwork::V4(Ipv4Network::new(Ipv4Addr::new(10, 0, 0, 0), 8).unwrap()),
+        IpNetwork::V4(Ipv4Network::new(Ipv4Addr::new(100, 64, 0, 0), 10).unwrap()),
         IpNetwork::V4(Ipv4Network::new(Ipv4Addr::new(172, 16, 0, 0), 12).unwrap()),
         IpNetwork::V4(Ipv4Network::new(Ipv4Addr::new(192, 168, 0, 0), 16).unwrap()),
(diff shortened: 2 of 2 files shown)


0xBA5E64:
Really hoping to see this get merged. Mullvad's lack of interoperability with Tailscale is very frustrating, especially since the two projects *do* already have interoperability though Mullvad exit-nodes on Tailscale. as a regular Mullvad user however, I really hope to be able to use this instead.

faern:
I'm not comfortable just calling this network a "LAN" and blindly allow it for everyone who wants local network traffic. This network is primarily intended for CGN. This would be a risk to most customers, and a benefit to probably very few.

If you run Linux, you can apply custom Firewall rules and allow `100.64.0.0/10` by yourself. See this guide: https://mullvad.net/en/help/split-tunneling-with-linux-advanced

Maybe we should consider adding a feature where the user can enter a list of networks that they want unblocked?

chaeynz:
> I'm not comfortable just calling this network a "LAN" and blindly allow it for everyone who wants local network traffic. This network is primarily intended for CGN. This would be a risk to most customers, and a benefit to probably very few.
> 
> If you run Linux, you can apply custom Firewall rules and allow `100.64.0.0/10` by yourself. See this guide: https://mullvad.net/en/help/split-tunneling-with-linux-advanced
> 
> Maybe we should consider adding a feature where the user can enter a list of networks that they want unblocked?

Yes, I think the solution you proposed would be great

PR #6101, opened 2024-04-09
Add app version to start of each log file

acb-mv:
This PR adds a header line to the start of each log file at creation time; this contains the product version.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)

### DIFF
--- a/ios/MullvadLogging/LogFileOutputStream.swift
+++ b/ios/MullvadLogging/LogFileOutputStream.swift
@@ -19,6 +19,7 @@ class LogFileOutputStream: TextOutputStream {
     private let fileURL: URL
     private let encoding: String.Encoding
     private let maxBufferCapacity: Int
+    private let fileHeader: String
 
     private var state: State = .closed {
         didSet {
@@ -44,10 +45,11 @@ class LogFileOutputStream: TextOutputStream {
         case waitingToReopen
     }
 
-    init(fileURL: URL, encoding: String.Encoding = .utf8, maxBufferCapacity: Int = 16 * 1024) {
+    init(fileURL: URL, header: String, encoding: String.Encoding = .utf8, maxBufferCapacity: Int = 16 * 1024) {
         self.fileURL = fileURL
         self.encoding = encoding
         self.maxBufferCapacity = maxBufferCapacity
+        self.fileHeader = header
     }
 
     deinit {
@@ -66,7 +68,7 @@ class LogFileOutputStream: TextOutputStream {
         switch state {
         case .closed:
             do {
-                let fileHandle = try openFile()
+                let fileHandle = try openFileWithHeader(fileHeader)
                 state = .opened(fileHandle)
                 try write(fileHandle: fileHandle, data: data)
             } catch {
@@ -137,15 +139,22 @@ class LogFileOutputStream: TextOutputStream {
         timer = nil
     }
 
+    private func openFileWithHeader(_ header: String) throws -> FileHandle {
+        let fileHandle = try openFile()
+
+        let messageData =
+            "\(header)\n"
+                .data(using: encoding, allowLossyConversion: true)!
+        try write(fileHandle: fileHandle, data: messageData)
+
+        return fileHandle
+    }
+
     private func reopenFile() {
         do {
-            let fileHandle = try openFile()
-
             // Write a message indicating that the file was reopened.
-            let messageData =
-                "<Log file re-opened after failure. Buffered \(buffer.count) bytes of messages>\n"
-                    .data(using: encoding, allowLossyConversion: true)!
-            try write(fileHandle: fileHandle, data: messageData)
+            let fileHandle =
+                try openFileWithHeader("<Log file re-opened after failure. Buffered \(buffer.count) bytes of messages>")
 
             // Write all buffered messages.
             if !buffer.isEmpty {
--- a/ios/MullvadLogging/Logging.swift
+++ b/ios/MullvadLogging/Logging.swift
@@ -21,8 +21,11 @@ public struct LoggerBuilder {
 
     public var metadata: Logger.Metadata = [:]
     public var logLevel: Logger.Level = .debug
+    public var header: String
 
-    public init() {}
+    public init(header: String) {
+        self.header = header
+    }
 
     public mutating func addFileOutput(fileURL: URL) {
         let logsDirectoryURL = fileURL.deletingLastPathComponent()
@@ -42,19 +45,18 @@ public struct LoggerBuilder {
             logRotationErrors.append(error)
         }
 
-        outputs.append(.fileOutput(LogFileOutputStream(fileURL: fileURL)))
+        outputs.append(.fileOutput(LogFileOutputStream(fileURL: fileURL, header: header)))
     }
 
     public mutating func addOSLogOutput(subsystem: String) {
         outputs.append(.osLogOutput(subsystem))
     }
 
-    public func install(header: String) {
+    public func install() {
         LoggingSystem.bootstrap { label -> LogHandler in
             let logHandlers: [LogHandler] = outputs.map { output in
                 switch output {
                 case let .fileOutput(stream):
-                    stream.write("\(header)\n")
                     return CustomFormatLogHandler(label: label, streams: [stream])
 
                 case let .osLogOutput(subsystem):
--- a/ios/MullvadLogging/Logging.swift
+++ b/ios/MullvadLogging/Logging.swift
@@ -49,11 +49,12 @@ public struct LoggerBuilder {
         outputs.append(.osLogOutput(subsystem))
     }
 
-    public func install() {
+    public func install(header: String) {
         LoggingSystem.bootstrap { label -> LogHandler in
             let logHandlers: [LogHandler] = outputs.map { output in
                 switch output {
                 case let .fileOutput(stream):
+                    stream.write("\(header)\n")
                     return CustomFormatLogHandler(label: label, streams: [stream])
 
                 case let .osLogOutput(subsystem):
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -44,6 +44,7 @@
 		449EB9FD2B95F8AD00DFA4EB /* DeviceMock.swift in Sources */ = {isa = PBXBuildFile; fileRef = 449EB9FC2B95F8AD00DFA4EB /* DeviceMock.swift */; };
 		449EB9FF2B95FF2500DFA4EB /* AccountMock.swift in Sources */ = {isa = PBXBuildFile; fileRef = 449EB9FE2B95FF2500DFA4EB /* AccountMock.swift */; };
 		449EBA262B975B9700DFA4EB /* PostQuantumKeyReceiving.swift in Sources */ = {isa = PBXBuildFile; fileRef = 449EBA252B975B9700DFA4EB /* PostQuantumKeyReceiving.swift */; };
+		44B02E3B2BC5732D008EDF34 /* LoggingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44B02E3A2BC5732D008EDF34 /* LoggingTests.swift */; };
 		44DD7D242B6CFFD70005F67F /* StartTunnelOperationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44DD7D232B6CFFD70005F67F /* StartTunnelOperationTests.swift */; };
 		44DD7D272B6D18FB0005F67F /* MockTunnelInteractor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44DD7D262B6D18FB0005F67F /* MockTunnelInteractor.swift */; };
 		44DD7D292B7113CA0005F67F /* MockTunnel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44DD7D282B7113CA0005F67F /* MockTunnel.swift */; };
@@ -1328,6 +1329,7 @@
 		449EB9FC2B95F8AD00DFA4EB /* DeviceMock.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = DeviceMock.swift; sourceTree = "<group>"; };
 		449EB9FE2B95FF2500DFA4EB /* AccountMock.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountMock.swift; sourceTree = "<group>"; };
 		449EBA252B975B9700DFA4EB /* PostQuantumKeyReceiving.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PostQuantumKeyReceiving.swift; sourceTree = "<group>"; };
+		44B02E3A2BC5732D008EDF34 /* LoggingTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = LoggingTests.swift; sourceTree = "<group>"; };
 		44DD7D232B6CFFD70005F67F /* StartTunnelOperationTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StartTunnelOperationTests.swift; sourceTree = "<group>"; };
 		44DD7D262B6D18FB0005F67F /* MockTunnelInteractor.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MockTunnelInteractor.swift; sourceTree = "<group>"; };
 		44DD7D282B7113CA0005F67F /* MockTunnel.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MockTunnel.swift; sourceTree = "<group>"; };
@@ -2948,7 +2950,6 @@
 				7A83A0C52B29A750008B5CE7 /* APIAccessMethodsTests.swift */,
 				A900E9BD2ACC654100C95F67 /* APIProxy+Stubs.swift */,
 				A9EC20E72A5D3A8C0040D56E /* CoordinatesTests.swift */,
-				01168B192BBEE7F800D5F382 /* LoggingTests.swift */,
 				5896AE85246D6AD8005B36CB /* CustomDateComponentsFormattingTests.swift */,
 				58915D622A25F8400066445B /* DeviceCheckOperationTests.swift */,
 				A900E9BB2ACC609200C95F67 /* DevicesProxy+Stubs.swift */,
@@ -2963,6 +2964,7 @@
 				7AB4CCB82B69097E006037F5 /* IPOverrideTests.swift */,
 				7A516C3B2B712F0B00BBD33D /* IPOverrideWrapperTests.swift */,
 				7AA513852BC91C6B00D081A4 /* LogRotationTests.swift */,
+				44B02E3A2BC5732D008EDF34 /* LoggingTests.swift */,
 				A9B6AC172ADE8F4300F7802A /* MigrationManagerTests.swift */,
 				58C3FA652A38549D006A450A /* MockFileCache.swift */,
 				F09D04B42AE93CB6003D4F89 /* OutgoingConnectionProxy+Stub.swift */,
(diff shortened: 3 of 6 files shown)


buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: all files reviewed, 6 unresolved discussions (waiting on @acb-mv)

___
*[`ios/MullvadLogging/Logging.swift` line 54 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5kGg06Dzqe5d7pK2T:-Nv5kGg06Dzqe5d7pK2U:b-382ed8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadLogging/Logging.swift#L54)):*
> ```Swift
>                 switch output {
>                 case let .fileOutput(stream):
>                     stream.write("\(header)\n")
> ```

From my preliminary tests, it looks like this function is called many times in a row. I'm not sure why but we should investigate.
___
*[`ios/MullvadVPN/AppDelegate.swift` line 364 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5kNyp261bzBScjHed:-Nv5kNyp261bzBScjHee:b-2nxdok) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPN/AppDelegate.swift#L364)):*
> ```Swift
>         logger = Logger(label: "AppDelegate")
> 
>         loggerBuilder.logLevel = .debug
> ```

The [default level](https://github.com/mullvad/mullvadvpn-app/blob/main/ios/MullvadLogging/Logging.swift#L22) in `LoggerBuilder` is already set to`.debug` 
We can remove this line 
___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 4934 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5kyCv2cHXra1yCPeJ:-Nv5kyCv2cHXra1yCPeK:b-ta6x7h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPN.xcodeproj/project.pbxproj#L4934)):*
> ```
> 				A9A5FA2F2ACB05160083449F /* FixedWidthIntegerArithmeticsTests.swift in Sources */,
> 				A9A5FA302ACB05160083449F /* InputTextFormatterTests.swift in Sources */,
> 				44B02E3B2BC5732D008EDF34 /* LoggingTests.swift in Sources */,
> ```

Let's `sort by names` on the `MullvadVPNTests` folder 
___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5kk-q3pjBjB4W6ej-:-Nv5kk-q3pjBjB4W6ej0:bj0rduo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPNTests/LoggingTests.swift#L16)):*
> ```Swift
>     
>     func temporaryFileURL() -> URL {
>         
> ```

Don't forget to `swiftlint` before committing files 
___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5l_zXCAsBwKV_BuVR:-Nv5l_zXCAsBwKV_BuVS:b-igzvyw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPNTests/LoggingTests.swift#L27)):*
> ```Swift
>                 let fileManager = FileManager.default
>                 // Check that the file exists before trying to delete it.
>                 if fileManager.fileExists(atPath: fileURL.path) {
> ```

In general, it's better practice (on Apple platforms) to not detect where a file might be, but just go and delete it immediately.
Especially when the file can be located on a remote storage where the presence of files is never guaranteed.
As such, the recommendation is usually to try to delete the file, and fail gracefully, rather than guess whether the file exists before attempting deletion.

That being said here it doesn't matter too much since it's a test, but I'm writing for our future selves to not be tempted to copy this pattern in other parts of the application where it might be more important.

TL;DR
`try? fileManager.removeItem(at: fileURL)` is plenty good for this test
___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 50 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5nVU40olIOMxE_X2J:-Nv5nVU40olIOMxE_X2K:bd1o6nt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPNTests/LoggingTests.swift#L50)):*
> ```Swift
>         builder.install(header: expectedHeader)
>                 
>         let contents = String(decoding: try! Data(contentsOf: fileURL), as: UTF8.self)
> ```

`let contents = try XCTUnwrap(String(contentsOf: fileURL))`
This will require marking the test as `throws` but is much less convoluted 



_Code quote:_
```Swift
let contents = String(decoding: try! Data(contentsOf: fileURL), as: UTF8.self)
```

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: all files reviewed, 7 unresolved discussions (waiting on @acb-mv)

___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 23 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5oPR70kI979CGDezc:-Nv5oPR70kI979CGDezd:b3juepu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPNTests/LoggingTests.swift#L23)):*
> ```Swift
>         
>         // Add a teardown block to delete any file at `fileURL`.
>         addTeardownBlock {
> ```

Really cool use of `addTeardownBlock` !

acb-mv (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: 1 of 6 files reviewed, 7 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadLogging/Logging.swift` line 54 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5kGg06Dzqe5d7pK2T:-NvC790n6VLVkHjCpOkw:b-ze3ni5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadLogging/Logging.swift#L54)):*

From my preliminary tests, it looks like this function is called many times in a row. I'm not sure why but we should investigate.
</blockquote></details>

This should be fixed in the new version
___
*[`ios/MullvadVPN/AppDelegate.swift` line 364 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5kNyp261bzBScjHed:-NvC7WJ5EcRuLRtYLfdC:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPN/AppDelegate.swift#L364)):*

The [default level](https://github.com/mullvad/mullvadvpn-app/blob/main/ios/MullvadLogging/Logging.swift#L22) in `LoggerBuilder` is already set to`.debug` 
We can remove this line
</blockquote></details>

Done.
___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 4934 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5kyCv2cHXra1yCPeJ:-NvC7eJn1Ul3MGc_ChwE:bjjb5k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPN.xcodeproj/project.pbxproj#L4934)):*

Let's `sort by names` on the `MullvadVPNTests` folder
</blockquote></details>

They're sorted by name in the project as shown in XCode
___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5kk-q3pjBjB4W6ej-:-NvC7jtj6uHJ1A4QdQ_I:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPNTests/LoggingTests.swift#L16)):*

Don't forget to `swiftlint` before committing files
</blockquote></details>

Done.
___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 23 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5oPR70kI979CGDezc:-NvC7l3VDHeJqps6SWkK:b-zcjvte) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPNTests/LoggingTests.swift#L23)):*

Really cool use of `addTeardownBlock` !
</blockquote></details>

Thanks
___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5l_zXCAsBwKV_BuVR:-NvC88QL5AqYODGctAvh:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPNTests/LoggingTests.swift#L27)):*

In general, it's better practice (on Apple platforms) to not detect where a file might be, but just go and delete it immediately.
Especially when the file can be located on a remote storage where the presence of files is never guaranteed.
As such, the recommendation is usually to try to delete the file, and fail gracefully, rather than guess whether the file exists before attempting deletion.

That being said here it doesn't matter too much since it's a test, but I'm writing for our future selves to not be tempted to copy this pattern in other parts of the application where it might be more important.

TL;DR
`try? fileManager.removeItem(at: fileURL)` is plenty good for this test
</blockquote></details>

Done.
___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 50 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5nVU40olIOMxE_X2J:-NvC8Hbr4oTBquinI6Kh:bwygxrx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPNTests/LoggingTests.swift#L50)):*

`let contents = try XCTUnwrap(String(contentsOf: fileURL))`
This will require marking the test as `throws` but is much less convoluted
</blockquote></details>

Noted

buggmagnet (commented):
> Reviewed 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: 2 of 6 files reviewed, 5 unresolved discussions (waiting on @acb-mv)

___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 41 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-NvVkBvC5Xp63Ze7xqrl:-NvVkBvC5Xp63Ze7xqrm:bdid8z3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5e3d971a5bcf534df6f0d3ff879d01f9f2ae3bf/ios/MullvadVPNTests/LoggingTests.swift#L41)):*
> ```Swift
> 
>         // For some reason, reading the file fails here, despite the file existing. Manual inspection reveals the file to have the correct header. ¯\_(ツ)_/¯
> //        let contents = String(decoding: try! Data(contentsOf: fileURL), as: UTF8.self)
> ```

It's not "for some reason", it's because logging happens asynchronously. 
If you look at `LogFileOutputStream`, which is used when we call `addFileOutput`, you will see that it does the following when logging 
```swift
    func write(_ string: String) {
        queue.async {
            self.writeNoQueue(string)
        }
    }
```

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: 2 of 6 files reviewed, 5 unresolved discussions (waiting on @acb-mv)

___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 41 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-NvVkBvC5Xp63Ze7xqrl:-NvVki3o7bUhyb9rZBkl:b-jk7abb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5e3d971a5bcf534df6f0d3ff879d01f9f2ae3bf/ios/MullvadVPNTests/LoggingTests.swift#L41)):*

It's not "for some reason", it's because logging happens asynchronously. 
If you look at `LogFileOutputStream`, which is used when we call `addFileOutput`, you will see that it does the following when logging 
```swift
    func write(_ string: String) {
        queue.async {
            self.writeNoQueue(string)
        }
    }
```
</blockquote></details>

If you replace this with a call to `sync` the test works all the time now instead of randomly failing.
That doesn't mean we should make writing synchronous, but it helps understanding why this test was unreliable.

acb-mv (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: 2 of 6 files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 364 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-Nv5kNyp261bzBScjHed:-NvVruwMBHgbyQb0AV3k:b-bs2ej7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/417f9f2294606f84baa9362094064892db7360df/ios/MullvadVPN/AppDelegate.swift#L364)):*

The line is still there.
</blockquote></details>

It should be updated now.
___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 41 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-NvVkBvC5Xp63Ze7xqrl:-NvVuXV02nxIUv4DKWeX:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b5e3d971a5bcf534df6f0d3ff879d01f9f2ae3bf/ios/MullvadVPNTests/LoggingTests.swift#L41)):*

If you replace this with a call to `sync` the test works all the time now instead of randomly failing.
That doesn't mean we should make writing synchronous, but it helps understanding why this test was unreliable.
</blockquote></details>

Done

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: 5 of 6 files reviewed, 6 unresolved discussions (waiting on @acb-mv and @pinkisemils)

___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 41 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-NvWlAKS25CBlLmz1S0j:-NvWlAKS25CBlLmz1S0k:b-tzw7n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5f9a07ea25f08c2e236331927863765eac1bfa8a/ios/MullvadVPNTests/LoggingTests.swift#L41)):*
> ```Swift
> 
>         // Wait for the log file to settle
>         usleep(100000)
> ```

This is prone to be flaky and is not a good fix.
Instead here's what we should do:
- Dependency inject an implementation of a custom protocol that implements
- `init(fileURL: URL, header: String, encoding: String.Encoding = .utf8, maxBufferCapacity: Int = 16 * 1024)`
- `TextOutputStream` 
- Into the `LoggerBuilder` struct
Then:
- Make it default to `LogFileOutputStream` 
- Use a stub implementation that writes synchronously in a unit test context

We will be writing more tests in the future that will do calls to the logging system, this is not a solution that will be future prone.

acb-mv (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: 5 of 6 files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 41 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-NvWlAKS25CBlLmz1S0j:-NvWspT17KPUTK4lMbLo:byvz8hc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5f9a07ea25f08c2e236331927863765eac1bfa8a/ios/MullvadVPNTests/LoggingTests.swift#L41)):*

This is prone to be flaky and is not a good fix.
Instead here's what we should do:
- Dependency inject an implementation of a custom protocol that implements
- `init(fileURL: URL, header: String, encoding: String.Encoding = .utf8, maxBufferCapacity: Int = 16 * 1024)`
- `TextOutputStream` 
- Into the `LoggerBuilder` struct
Then:
- Make it default to `LogFileOutputStream` 
- Use a stub implementation that writes synchronously in a unit test context

We will be writing more tests in the future that will do calls to the logging system, this is not a solution that will be future prone.
</blockquote></details>

Except that that won't be useful as `LogFileOutputStream` is where the header writing logic is.

acb-mv (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: 5 of 6 files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 41 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-NvWlAKS25CBlLmz1S0j:-NvWvzM_DaXWNXywUqvq:bawjuh4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5f9a07ea25f08c2e236331927863765eac1bfa8a/ios/MullvadVPNTests/LoggingTests.swift#L41)):*

Except that that won't be useful as `LogFileOutputStream` is where the header writing logic is.
</blockquote></details>

The same happens for testing the `LogFileOutputStream` itself. In the end, adding a `sync()` call to the test seems like the least-worst solution for actually testing the code itself (short of refactoring it to separate the file writer from the rest of the logic).

buggmagnet (approved):
:lgtm:

> Reviewed 3 of 5 files at r2, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-:-NvaKq0UEVp38jLwSWk9:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @acb-mv)

___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 41 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101#-NvWlAKS25CBlLmz1S0j:-NvaK6G21zWOb7VgYsSm:b-lm26vd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5f9a07ea25f08c2e236331927863765eac1bfa8a/ios/MullvadVPNTests/LoggingTests.swift#L41)):*

The same happens for testing the `LogFileOutputStream` itself. In the end, adding a `sync()` call to the test seems like the least-worst solution for actually testing the code itself (short of refactoring it to separate the file writer from the rest of the logic).
</blockquote></details>

Oh, yes you're right, that was a bad suggestion from me, sorry.
The call to `sync` feels a bit hacky, (or rather, I wouldn't want to have that class of fixes in tests in general) but I think that's okay an acceptable tradeoff for now.

rablador (commented):
> Reviewed 1 of 5 files at r1, 3 of 5 files at r2, 1 of 2 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: all files reviewed, 3 unresolved discussions (waiting on @acb-mv and @buggmagnet)

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: all files reviewed, 3 unresolved discussions (waiting on @acb-mv and @buggmagnet)

pinkisemils (approved):
> Reviewed 1 of 1 files at r5, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6101)* status: all files reviewed, 3 unresolved discussions (waiting on @acb-mv and @buggmagnet)

PR #6108, opened 2024-04-11
Fix flaky tests caused by messed up daemon environment

MarkusPettersson98:
## What? :detective:
This PR addresses an issue in the test framework where some test would mess with the daemon's environment variables without resetting them to their expected default values. 
## Why? 🤔
This could lead to subsequent tests failing for unforeseen reasons, which was very confusing.
## How? :hammer:  :carpentry_saw: 
After each test is finished, we check whether the daemon's environment variables deviate from the one's specified by the current test configuration. If so, we restart the daemon with the 'expected values'. This all happens in the existing clean up routine, which means that no tests *should* have to clean up this particular mess themselves.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108)

### DIFF
--- a/test/Cargo.lock
+++ b/test/Cargo.lock
@@ -2123,15 +2123,16 @@ dependencies = [
 
 [[package]]
 name = "pcap"
-version = "0.10.1"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2da544c8115cc65b474554569c7654fc94a4f6a167f79192536e148fd654e17a"
+checksum = "99e935fc73d54a89fff576526c2ccd42bbf8247aae05b358693475b14fd4ff79"
 dependencies = [
  "bitflags 1.3.2",
  "errno 0.2.8",
  "futures",
  "libc",
  "libloading",
+ "pkg-config",
  "regex",
  "tokio",
  "windows-sys 0.36.1",
--- a/test/socks-server/src/lib.rs
+++ b/test/socks-server/src/lib.rs
@@ -1,3 +1,4 @@
+use fast_socks5::server::{AcceptAuthentication, Socks5Server};
 use futures::StreamExt;
 use std::{io, net::SocketAddr};
 
@@ -13,10 +14,9 @@ pub struct Handle {
 
 /// Spawn a SOCKS server bound to `bind_addr`
 pub async fn spawn(bind_addr: SocketAddr) -> Result<Handle, Error> {
-    let socks_server: fast_socks5::server::Socks5Server =
-        fast_socks5::server::Socks5Server::bind(bind_addr)
-            .await
-            .map_err(Error::StartSocksServer)?;
+    let socks_server: Socks5Server<AcceptAuthentication> = Socks5Server::bind(bind_addr)
+        .await
+        .map_err(Error::StartSocksServer)?;
 
     let handle = tokio::spawn(async move {
         let mut incoming = socks_server.incoming();
--- a/test/test-manager/Cargo.toml
+++ b/test/test-manager/Cargo.toml
@@ -39,7 +39,7 @@ serde_json = { workspace = true }
 tokio-serde = { workspace = true }
 log = { workspace = true }
 
-pcap = { version = "0.10.1", features = ["capture-stream"] }
+pcap = { version = "1.3", features = ["capture-stream"] }
 pnet_packet = "0.31.0"
 
 test-rpc = { path = "../test-rpc" }
--- a/test/test-manager/src/mullvad_daemon.rs
+++ b/test/test-manager/src/mullvad_daemon.rs
@@ -51,14 +51,17 @@ pub struct RpcClientProvider {
     service: DummyService,
 }
 
+pub enum MullvadClientArgument {
+    WithClient(MullvadProxyClient),
+    None,
+}
+
 impl RpcClientProvider {
-    pub async fn as_type(
-        &self,
-        client_type: MullvadClientVersion,
-    ) -> Box<dyn std::any::Any + Send> {
+    /// Whether a [test case](test_macro::test_function) needs a [`MullvadProxyClient`].
+    pub async fn mullvad_client(&self, client_type: MullvadClientVersion) -> MullvadClientArgument {
         match client_type {
-            MullvadClientVersion::New => Box::new(self.new_client().await),
-            MullvadClientVersion::None => Box::new(()),
+            MullvadClientVersion::New => MullvadClientArgument::WithClient(self.new_client().await),
+            MullvadClientVersion::None => MullvadClientArgument::None,
         }
     }
 
--- a/test/test-manager/src/run_tests.rs
+++ b/test/test-manager/src/run_tests.rs
@@ -105,8 +105,8 @@ pub async fn run(
             // Try to reset the daemon state if the test failed OR if the test doesn't explicitly
             // disabled cleanup.
             if test.cleanup || matches!(test_result.result, Err(_) | Ok(Err(_))) {
-                let mut client = test_context.rpc_provider.new_client().await;
-                crate::tests::cleanup_after_test(&mut client).await?;
+                crate::tests::cleanup_after_test(client.clone(), &test_context.rpc_provider)
+                    .await?;
             }
         }
 
--- a/test/test-manager/src/run_tests.rs
+++ b/test/test-manager/src/run_tests.rs
@@ -1,6 +1,6 @@
 use crate::{
     logging::{panic_as_string, TestOutput},
-    mullvad_daemon,
+    mullvad_daemon::{self, MullvadClientArgument},
     summary::{self, maybe_log_test_result},
     tests::{self, config::TEST_CONFIG, get_tests, TestContext},
     vm,
@@ -80,9 +80,9 @@ pub async fn run(
     let logger = super::logging::Logger::get_or_init();
 
     for test in tests {
-        let mclient = test_context
+        let mullvad_client = test_context
             .rpc_provider
-            .as_type(test.mullvad_client_version)
+            .mullvad_client(test.mullvad_client_version)
             .await;
 
         log::info!("Running {}", test.name);
@@ -94,7 +94,7 @@ pub async fn run(
 
         let test_result = run_test(
             client.clone(),
-            mclient,
+            mullvad_client,
             &test.func,
             test.name,
             test_context.clone(),
@@ -175,15 +175,15 @@ pub async fn run(
     final_result
 }
 
-pub async fn run_test<F, R, MullvadClient>(
+pub async fn run_test<F, R>(
     runner_rpc: ServiceClient,
-    mullvad_rpc: MullvadClient,
+    mullvad_rpc: MullvadClientArgument,
     test: &F,
     test_name: &'static str,
     test_context: super::tests::TestContext,
 ) -> TestOutput
 where
-    F: Fn(super::tests::TestContext, ServiceClient, MullvadClient) -> R,
+    F: Fn(super::tests::TestContext, ServiceClient, MullvadClientArgument) -> R,
     R: Future<Output = anyhow::Result<()>>,
 {
     let _flushed = runner_rpc.try_poll_output().await;
--- a/test/test-manager/src/tests/access_methods.rs
+++ b/test/test-manager/src/tests/access_methods.rs
@@ -1,50 +1,30 @@
 //! Integration tests for API access methods.
-use super::{Error, TestContext};
+//!
+//! The tested access methods are:
+//! * Shadowsocks
+//! * SOCKS5 in remote mode
+//!
+//! These tests rely on working proxies to exist *somewhere* for all tested protocols.
+//! If the proxies themselves are bad/not running, this test will fail due to issues
+//! that are out of the test manager's control.
+use anyhow::{anyhow, ensure, Context};
+
 use mullvad_management_interface::MullvadProxyClient;
+use talpid_types::net::proxy::CustomProxy;
 use test_macro::test_function;
 use test_rpc::ServiceClient;
 
-/// Assert that custom access methods may be used to access the Mullvad API.
-///
-/// The tested access methods are:
-/// * Shadowsocks
-/// * Socks5 in remote mode
-///
-/// # Note
-///
-/// This tests assume that there exists working proxies *somewhere* for all
-/// tested protocols. If the proxies themselves are bad/not running, this test
-/// will fail due to issues that are out of the test manager's control.
+use super::TestContext;
+
+/// Assert that API traffic can be proxied via a custom Shadowsocks proxy.
 #[test_function]
-pub async fn test_custom_access_methods(
+async fn test_access_method_shadowsocks(
     _: TestContext,
     _rpc: ServiceClient,
-    mullvad_client: MullvadProxyClient,
-) -> Result<(), Error> {
-    log::info!("Testing Shadowsocks access method");
-    test_shadowsocks(mullvad_client.clone()).await?;
-    log::info!("Testing SOCKS5 (Remote) access method");
-    test_socks_remote(mullvad_client.clone()).await?;
-    Ok(())
-}
-
-macro_rules! assert_access_method_works {
-    ($mullvad_client:expr, $access_method:expr) => {
-        let successful = $mullvad_client
-            .test_custom_api_access_method($access_method.clone().into())
-            .await
-            .expect("Failed to test custom API access method");
-
-        assert!(
-            successful,
-            "Failed while testing access method - {:?}",
-            $access_method
-        );
-    };
-}
-
-async fn test_shadowsocks(mut mullvad_client: MullvadProxyClient) -> Result<(), Error> {
+    mut mullvad_client: MullvadProxyClient,
+) -> anyhow::Result<()> {
     use mullvad_relay_selector::{RelaySelector, SelectorConfig};
+    log::info!("Testing Shadowsocks access method");
     // Set up all the parameters needed to create a custom Shadowsocks access method.
     //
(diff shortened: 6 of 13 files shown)


dlon (approved):
Good stuff!

> Reviewed 9 of 11 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-:-NvEZX6jAoZtQWnkpofH:beuyiaz)* status: all files reviewed, 3 unresolved discussions (waiting on @MarkusPettersson98)

___
*[`test/test-manager/src/mullvad_daemon.rs` line 61 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvDkmT006EArUNJ6pxg:-NvDkmT006EArUNJ6pxh:b-qb1ztz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/50161a5dd8b9813dc53481c03784a6712781c30b/test/test-manager/src/mullvad_daemon.rs#L61)):*
> ```XML
> impl RpcClientProvider {
>     /// Whether a [test case](test_macro::test_function) needs a [`MullvadProxyClient`].
>     pub async fn mullvad_client(&self, client_type: MullvadClientVersion) -> MullvadClientArgument {
> ```

Maybe there's no real point to `MullvadClientArgument` now. It could probably be removed in favor of `Option`, and the caller could be expected to call `Self::new_client`
___
*[`test/test-manager/src/tests/mod.rs` line 149 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvE_MnoDUB-mVtefHfh:-NvE_MnoDUB-mVtefHfi:bwqjf1f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/50161a5dd8b9813dc53481c03784a6712781c30b/test/test-manager/src/tests/mod.rs#L149)):*
> ```XML
>         rpc.set_daemon_environment(default_env).await?;
>     }
>     let mullvad_client = rpc_provider.new_client().await;
> ```

Contradicting myself, probably, but maybe we should pass the client to `cleanup_after_test`, and do `mem::replace(client, rpc_provider.new_client().await)` here if the daemon had to be restarted.
___
*[`test/test-runner/src/sys.rs` line 462 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvEYsRZA6gCLK5le4W3:-NvEYsRZA6gCLK5le4W4:b5fhv6c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/50161a5dd8b9813dc53481c03784a6712781c30b/test/test-runner/src/sys.rs#L462)):*
> ```XML
>         .await
>         .map_err(|e| test_rpc::Error::Service(e.to_string()))?
>         .write_all(override_content.as_bytes())
> ```

Nit: `fs::write` would be simpler.

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108)* status: 7 of 13 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`test/test-manager/src/mullvad_daemon.rs` line 61 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvDkmT006EArUNJ6pxg:-NvGKVmS0EYyoi0fjxXQ:bvyrkc8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/50161a5dd8b9813dc53481c03784a6712781c30b/test/test-manager/src/mullvad_daemon.rs#L61)):*

Maybe there's no real point to `MullvadClientArgument` now. It could probably be removed in favor of `Option`, and the caller could be expected to call `Self::new_client`
</blockquote></details>

I agree that `MullvadClientArgument` is not strictly necessary, but I think it serves as better documentation compared to using an `Option`. _If and when_ we expect the caller to invoke `RpcClientProvider::new_client` themselves, then I'm all for removing `MullvadClientArgument` :blush:
___
*[`test/test-runner/src/sys.rs` line 462 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvEYsRZA6gCLK5le4W3:-NvGKNzFC-ePbkiiVzeK:b-qiachz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/50161a5dd8b9813dc53481c03784a6712781c30b/test/test-runner/src/sys.rs#L462)):*

Nit: `fs::write` would be simpler.
</blockquote></details>

Done Nice, definitely using that! :blush:

dlon (approved):
> Reviewed 5 of 6 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108)* status: all files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`test/test-manager/src/mullvad_daemon.rs` line 61 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvDkmT006EArUNJ6pxg:-NvGRKYGD51CN108n5PX:b-8qvd9j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/50161a5dd8b9813dc53481c03784a6712781c30b/test/test-manager/src/mullvad_daemon.rs#L61)):*

I agree that `MullvadClientArgument` is not strictly necessary, but I think it serves as better documentation compared to using an `Option`. _If and when_ we expect the caller to invoke `RpcClientProvider::new_client` themselves, then I'm all for removing `MullvadClientArgument` :blush:
</blockquote></details>

All right! Let's keep it for now.
___
*[`test/test-manager/src/tests/mod.rs` line 149 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvE_MnoDUB-mVtefHfh:-NvGQm7PDnkVCiJqf6Su:bpp0mjs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/50161a5dd8b9813dc53481c03784a6712781c30b/test/test-manager/src/tests/mod.rs#L149)):*

Contradicting myself, probably, but maybe we should pass the client to `cleanup_after_test`, and do `mem::replace(client, rpc_provider.new_client().await)` here if the daemon had to be restarted.
</blockquote></details>

I'm overcomplicating this. Could we not just create the client after calling `restart_daemon` and return `Ok(())` here?

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108)* status: all files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`test/test-runner/src/sys.rs` line 659 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGRhdT5KAuqc96B7SL:-NvGRhdT5KAuqc96B7SM:bj90gko) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/df111c2d7dfa26f5739690e3a9b3a08fe6c14da7/test/test-runner/src/sys.rs#L659)):*
> ```XML
>         })
>         .await
>         .map_err(test_rpc::Error::from_tokio_join_error)??;
> ```

I slightly prefer panicking, as this should only panic if the task itself panics anyway, right?

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`test/test-runner/src/sys.rs` line 659 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGRhdT5KAuqc96B7SL:-NvGpr4P1ist1BQaeohZ:b5h34le) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/df111c2d7dfa26f5739690e3a9b3a08fe6c14da7/test/test-runner/src/sys.rs#L659)):*

Should we bring down the entire `test-runner` because one call to `get_daemon_environment` panics? Imo, we should leave that decision to the `test-manager` (or some other supervisor), because we will halt the test execution in place if we panic :blush:
</blockquote></details>

I totally agree that panicking should not halt the test runner. :slightly_smiling_face:  We definitely shouldn't panic if a "recoverable" error occurs. I think that isn't that case here, because `JoinHandle` only fails if the task itself panics, which it shouldn't.

So I'm thinking that this is handling a specific panic without solving the general problem. We're effectively wrapping a single piece of code in `catch_unwind`, but not handling panics that occur outside of this call to `spawn_blocking`, etc. IMO, the general solution might just be to restart `test-runner` if it crashes, or even catch panics at some point. :see_no_evil: 

Anyway, this isn't that important. :smile:

hulthe (commented):
> Reviewed 6 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108)* status: all files reviewed, 3 unresolved discussions (waiting on @MarkusPettersson98)

___
*[`test/test-manager/src/tests/mod.rs` line 141 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGxqCo9OtQYidB7e82:-NvGxqCp-dFQfC2pavf3:bjtpvax) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d0225384fb9c67c14c535f00559352b53bdbbbe/test/test-manager/src/tests/mod.rs#L141)):*
> ```XML
> }
> 
> /// Conditonally restart the running daemon
> ```

Nit: Conditionally
___
*[`test/test-runner/src/sys.rs` line 1 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvCkPgi0weifkGzMWSh:-NvCkPgi0weifkGzMWSi:be7menq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5a7abfe34fae91317a56e2ea93219d3c24dc26a2/test/test-runner/src/sys.rs#L1)):*
> ```XML
> use std::collections::HashMap;
> ```

BTW seems everything in this file seems to be target specific, maybe we should consider splitting into one file per target. Not strictly related to this PR of course.
___
*[`test/test-runner/src/sys.rs` line 641 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGffxiA3clzr8ITzWU:-NvGffxiA3clzr8ITzWV:b4lklmv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/df111c2d7dfa26f5739690e3a9b3a08fe6c14da7/test/test-runner/src/sys.rs#L641)):*
> ```XML
>                 string
>                     .trim_start_matches('"')
>                     .trim_end_matches('"')
> ```

Tip: [trim_matches](https://doc.rust-lang.org/std/primitive.str.html#method.trim_matches)
___
*[`test/test-runner/src/sys.rs` line 423 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGieVr6sY6lHYMRv0v:-NvGieVr6sY6lHYMRv0w:bq06aw9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d0225384fb9c67c14c535f00559352b53bdbbbe/test/test-runner/src/sys.rs#L423)):*
> ```XML
>             }
>             _ => Err(error),
>         }
> ```

Hmm.. this only parses a subset of ways to specify environment variables. Might be worth considering the cost of handling those cases vs the risk of this parsing breaking down, at a later date.
___
*[`test/test-runner/src/sys.rs` line 601 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGwXtVAv6oFE8VBg6f:-NvGwXtVAv6oFE8VBg6g:b-wpcr2s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d0225384fb9c67c14c535f00559352b53bdbbbe/test/test-runner/src/sys.rs#L601)):*
> ```XML
>             .map_err(|err| test_rpc::Error::FileSystem(err.to_string()))?;
>         buf
>     };
> ```

Tip: `tokio::fs::read_to_string` does all of this

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108)* status: all files reviewed, 2 unresolved discussions

___
*[`test/test-manager/src/tests/mod.rs` line 141 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGxqCo9OtQYidB7e82:-NvH7t6REbG6g67HRotW:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d0225384fb9c67c14c535f00559352b53bdbbbe/test/test-manager/src/tests/mod.rs#L141)):*

Nit: Conditionally
</blockquote></details>

Done
___
*[`test/test-runner/src/sys.rs` line 641 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGffxiA3clzr8ITzWU:-NvH9f3W7Xsk_QV8s7Lz:b-tovvud) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/df111c2d7dfa26f5739690e3a9b3a08fe6c14da7/test/test-runner/src/sys.rs#L641)):*

Tip: [trim_matches](https://doc.rust-lang.org/std/primitive.str.html#method.trim_matches)
</blockquote></details>

Done Nice!
___
*[`test/test-runner/src/sys.rs` line 601 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGwXtVAv6oFE8VBg6f:-NvH7zMkCoB0-Enx0KP5:b-4gyknk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d0225384fb9c67c14c535f00559352b53bdbbbe/test/test-runner/src/sys.rs#L601)):*

Tip: `tokio::fs::read_to_string` does all of this
</blockquote></details>

:raised_hands: Nice, thought that their ought to exists something similar already :blush:

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108)* status: all files reviewed, 2 unresolved discussions (waiting on @hulthe)

___
*[`test/test-runner/src/sys.rs` line 1 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvCkPgi0weifkGzMWSh:-NvHKH4L4N6wXmjS0LjG:b-eigf3l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5a7abfe34fae91317a56e2ea93219d3c24dc26a2/test/test-runner/src/sys.rs#L1)):*

BTW seems everything in this file seems to be target specific, maybe we should consider splitting into one file per target. Not strictly related to this PR of course.
</blockquote></details>

Yes, we definitely should! This PR has already grown a bit past its intended scope, so lets clean this up in another PR :blush: 
___
*[`test/test-runner/src/sys.rs` line 659 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGRhdT5KAuqc96B7SL:-NvHJyhT9VNiztBQrGE_:b-xe3tye) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/df111c2d7dfa26f5739690e3a9b3a08fe6c14da7/test/test-runner/src/sys.rs#L659)):*

I totally agree that panicking should not halt the test runner. :slightly_smiling_face:  I'll try to translate my thoughts into words.

We definitely shouldn't panic if a "recoverable" error occurs. I think that isn't that case here, because `JoinHandle` only fails if the task itself panics, which it shouldn't.

So I'm thinking that this is handling a specific panic without solving the general problem. We're effectively wrapping a single piece of code in `catch_unwind`, but not handling panics that occur outside of this call to `spawn_blocking`, etc. IMO, the general solution might just be to restart `test-runner` if it crashes, or even catch panics at some point. :see_no_evil: 

Anyway, this isn't that important. :smile:
</blockquote></details>

Yes, we should have a more in-depth discussion on this topic / the architecture of `test-runner` to see what approach is the most fitting :blush:

If it's okay, I'll keep it as it is. Otherwise, I'll just go ahead and add back the `unwrap`s :blush:

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108)* status: all files reviewed, 2 unresolved discussions (waiting on @hulthe)

___
*[`test/test-runner/src/sys.rs` line 423 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGieVr6sY6lHYMRv0v:-NvHKh5R8ClxidwV97kP:b-egzse5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d0225384fb9c67c14c535f00559352b53bdbbbe/test/test-runner/src/sys.rs#L423)):*

Hmm.. this only parses a subset of ways to specify environment variables. Might be worth considering the cost of handling those cases vs the risk of this parsing breaking down, at a later date.
</blockquote></details>

Yes, the parsing is limited. I think it is okay for now, as we also control the way that the environment variables are written to the override file :blush:

dlon (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-:-NvHW-JAEf-ZclL9i-2c:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @hulthe)

___
*[`test/test-runner/src/sys.rs` line 659 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6108#-NvGRhdT5KAuqc96B7SL:-NvHVvsJA3_cf5sZfaJY:b-r9szl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/df111c2d7dfa26f5739690e3a9b3a08fe6c14da7/test/test-runner/src/sys.rs#L659)):*

Yes, we should have a more in-depth discussion on this topic / the architecture of `test-runner` to see what approach is the most fitting :blush:

If it's okay, I'll keep it as it is. Otherwise, I'll just go ahead and add back the `unwrap`s :blush:
</blockquote></details>

That's fine!

PR #6110, opened 2024-04-11
Change log rotation to a quota based system

rablador:
Currently we only keep one .old.log, meaning we have the current log and the previous log. The logs are rotated on every app/process start, which is often on iOS. It would really help debugging thing if we had logs a bit further back in time.

We should therefore improve the current state by keeping as many logs as our filesystem budget allows for, eg.:

- never have more than 5MB of log files of any one type (application or packet tunnel)
- never have any single log file be larger than 500KB

Instead of suffixing log files with an .old.log when they're rotated, all log files should be named $loggingTarget.$currentDate.log, eg. PacketTunnel.2024.03.03-21:32:01.log.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110)

### DIFF
--- a/ios/MullvadLogging/Date+LogFormat.swift
+++ b/ios/MullvadLogging/Date+LogFormat.swift
@@ -15,4 +15,11 @@ extension Date {
 
         return formatter.string(from: self)
     }
+
+    public func logFormatFilename() -> String {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "dd-MM-yyyy'T'HH:mm:ss"
+
+        return formatter.string(from: self)
+    }
 }
--- a/ios/MullvadLogging/LogRotation.swift
+++ b/ios/MullvadLogging/LogRotation.swift
@@ -10,48 +10,96 @@ import Foundation
 import MullvadTypes
 
 public enum LogRotation {
+    private struct LogData {
+        var path: URL
+        var size: UInt64
+        var creationDate: Date
+    }
+
+    public struct Options {
+        let storageSizeLimit: Int
+        let oldestAllowedDate: Date
+
+        /// Options for log rotation, defining how logs should be retained.
+        ///
+        /// - Parameter storageSizeLimit: Storage size limit in bytes.
+        /// - Parameter oldestAllowedDate: Oldest allowed date.
+        public init(storageSizeLimit: Int, oldestAllowedDate: Date) {
+            self.storageSizeLimit = storageSizeLimit
+            self.oldestAllowedDate = oldestAllowedDate
+        }
+    }
+
     public enum Error: LocalizedError, WrappingError {
-        case noSourceLogFile
-        case moveSourceLogFile(Swift.Error)
+        case rotateLogFiles(Swift.Error)
 
         public var errorDescription: String? {
             switch self {
-            case .noSourceLogFile:
-                return "Source log file does not exist."
-            case .moveSourceLogFile:
-                return "Failure to move the source log file to backup."
+            case .rotateLogFiles:
+                return "Failure to rotate the source log file to backup."
             }
         }
 
         public var underlyingError: Swift.Error? {
             switch self {
-            case .noSourceLogFile:
-                return nil
-            case let .moveSourceLogFile(error):
+            case let .rotateLogFiles(error):
                 return error
             }
         }
     }
 
-    public static func rotateLog(logsDirectory: URL, logFileName: String) throws {
-        let source = logsDirectory.appendingPathComponent(logFileName)
-        let backup = source.deletingPathExtension().appendingPathExtension("old.log")
+    public static func rotateLogs(logDirectory: URL, options: Options) throws {
+        let fileManager = FileManager.default
 
         do {
-            _ = try FileManager.default.replaceItemAt(backup, withItemAt: source)
-        } catch {
-            // FileManager returns a very obscure error chain so we need to traverse it to find
-            // the root cause of the error.
-            for case let fileError as CocoaError in error.underlyingErrorChain {
-                // .fileNoSuchFile is returned when both backup and source log files do not exist
-                // .fileReadNoSuchFile is returned when backup exists but source log file does not
-                if fileError.code == .fileNoSuchFile || fileError.code == .fileReadNoSuchFile,
-                   fileError.url == source {
-                    throw Error.noSourceLogFile
+            // Filter out all log files in directory.
+            let logPaths: [URL] = (try fileManager.contentsOfDirectory(
+                atPath: logDirectory.relativePath
+            )).compactMap { file in
+                if file.split(separator: ".").last == "log" {
+                    logDirectory.appendingPathComponent(file)
+                } else {
+                    nil
                 }
             }
 
-            throw Error.moveSourceLogFile(error)
+            // Convert logs into objects with necessary meta data.
+            let logs = try logPaths.map { logPath in
+                let attributes = try fileManager.attributesOfItem(atPath: logPath.relativePath)
+                let size = (attributes[.size] as? UInt64) ?? 0
+                let creationDate = (attributes[.creationDate] as? Date) ?? Date.distantPast
+
+                return LogData(path: logPath, size: size, creationDate: creationDate)
+            }.sorted { log1, log2 in
+                log1.creationDate > log2.creationDate
+            }
+
+            try deleteLogsOlderThan(options.oldestAllowedDate, in: logs)
+            try deleteLogsWithCombinedSizeLargerThan(options.storageSizeLimit, in: logs)
+        } catch {
+            throw Error.rotateLogFiles(error)
+        }
+    }
+
+    private static func deleteLogsOlderThan(_ dateThreshold: Date, in logs: [LogData]) throws {
+        let fileManager = FileManager.default
+
+        for log in logs where log.creationDate < dateThreshold {
+            try fileManager.removeItem(at: log.path)
+        }
+    }
+
+    private static func deleteLogsWithCombinedSizeLargerThan(_ sizeThreshold: Int, in logs: [LogData]) throws {
+        let fileManager = FileManager.default
+
+        // Delete all logs outside maximum capacity (ordered newest to oldest).
+        var fileSizes = UInt64.zero
+        for log in logs {
+            fileSizes += log.size
+
+            if fileSizes > sizeThreshold {
+                try fileManager.removeItem(at: log.path)
+            }
         }
     }
 }
--- a/ios/MullvadLogging/Logging.swift
+++ b/ios/MullvadLogging/Logging.swift
@@ -8,6 +8,7 @@
 
 import Foundation
 @_exported import Logging
+import MullvadTypes
 
 private enum LoggerOutput {
     case fileOutput(_ fileOutput: LogFileOutputStream)
@@ -24,7 +25,6 @@ public struct LoggerBuilder {
     public init() {}
 
     public mutating func addFileOutput(fileURL: URL) {
-        let logFileName = fileURL.lastPathComponent
         let logsDirectoryURL = fileURL.deletingLastPathComponent()
 
         try? FileManager.default.createDirectory(
@@ -34,7 +34,10 @@ public struct LoggerBuilder {
         )
 
         do {
-            try LogRotation.rotateLog(logsDirectory: logsDirectoryURL, logFileName: logFileName)
+            try LogRotation.rotateLogs(logDirectory: logsDirectoryURL, options: LogRotation.Options(
+                storageSizeLimit: 5_242_880, // 5 MB
+                oldestAllowedDate: Date(timeIntervalSinceNow: Duration.days(7).timeInterval)
+            ))
         } catch {
             logRotationErrors.append(error)
         }
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -582,6 +582,7 @@
 		7A9CCCC42A96302800DD6A34 /* TunnelCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A9CCCB22A96302800DD6A34 /* TunnelCoordinator.swift */; };
 		7A9FA1422A2E3306000B728D /* CheckboxView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A9FA1412A2E3306000B728D /* CheckboxView.swift */; };
 		7A9FA1442A2E3FE5000B728D /* CheckableSettingsCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A9FA1432A2E3FE5000B728D /* CheckableSettingsCell.swift */; };
+		7AA513862BC91C6B00D081A4 /* LogRotationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AA513852BC91C6B00D081A4 /* LogRotationTests.swift */; };
 		7AB2B6702BA1EB8C00B03E3B /* ListCustomListViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AB2B66E2BA1EB8C00B03E3B /* ListCustomListViewController.swift */; };
 		7AB2B6712BA1EB8C00B03E3B /* ListCustomListCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AB2B66F2BA1EB8C00B03E3B /* ListCustomListCoordinator.swift */; };
 		7AB4CCB92B69097E006037F5 /* IPOverrideTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AB4CCB82B69097E006037F5 /* IPOverrideTests.swift */; };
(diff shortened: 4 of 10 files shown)


acb-mv (approved):
> Reviewed 4 of 8 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

buggmagnet (changes requested):
> Reviewed 3 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110)* status: all files reviewed, 7 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadLogging/LogRotation.swift` line 20 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvWpA509fqPfIwOLq4v:-NvWpA509fqPfIwOLq4w:b-jvf2vg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L20)):*
> ```Swift
> 
>     public struct Options {
>         let storageSizeLimit: Int?
> ```

nit
I feel that these parameters are optional just because it made it easier to write tests.
I'm not sure how I feel about this. But IMHO they shouldn't be optional.
___
*[`ios/MullvadLogging/LogRotation.swift` line 89 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX0dhe8z1pjrSlAMHK:-NvX0dhe8z1pjrSlAMHL:b-dn56rv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L89)):*
> ```Swift
>     }
> 
>     private static func rotateLogsByDate(logs: [LogData], oldestAllowedDate: Date) throws {
> ```

nit
What do you think of the following name ?
`private static func deleteLogsOlderThan(_ dateThreshold: Date, in logs[LogData]) throws`
___
*[`ios/MullvadLogging/LogRotation.swift` line 92 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX-moM7wDPjmboUgjL:-NvX-moM7wDPjmboUgjM:b-xqc9q5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L92)):*
> ```Swift
>         let fileManager = FileManager.default
> 
>         for log in logs where log.creationDate < oldestAllowedDate {
> ```

Cool usage of a `where` clause in a for loop !
___
*[`ios/MullvadLogging/LogRotation.swift` line 100 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX-i4VBIMrbTUZ9Qbd:-NvX-i4VBIMrbTUZ9Qbe:bq94j1a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L100)):*
> ```Swift
>         let fileManager = FileManager.default
> 
>         // From newest to oldest, delete all logs outside maximum capacity.
> ```

This comment is outdated, it should only refer to log size now.
___
*[`ios/MullvadLogging/LogRotation.swift` line 102 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX13vG7r9MGtisaeQk:-NvX13vHBy1vTrwp92gN:bv41zht) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L102)):*
> ```Swift
>         // From newest to oldest, delete all logs outside maximum capacity.
>         var fileSizes = UInt64.zero
>         for log in logs {
> ```

This doesn't look to be doing the right thing

The following test will fail
```swift
    func testRotateSize() throws {
        let logPaths = [
            directoryPath.appendingPathComponent("test1.log"),
            directoryPath.appendingPathComponent("test2.log"),
        ]

        try writeDataToDisk(path: logPaths[0], byteSize: 3)
        try writeDataToDisk(path: logPaths[1], byteSize: 10_000)

        try LogRotation.rotateLogs(logDirectory: directoryPath, options: LogRotation.Options(storageSizeLimit: 3999))
        let logFileCount = try fileManager.contentsOfDirectory(atPath: directoryPath.relativePath).count
        XCTAssertEqual(logFileCount, 1)
    }
```
___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 16 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvWqm3WAE84md1sBo09:-NvWqm3WAE84md1sBo0A:b-3v3y6a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadVPNTests/LogRotationTests.swift#L16)):*
From the docs:
> Override `setUp() async throws` for asynchronous state preparation with error handling before each test. Do not override `setUp(completion:)` for synchronous state preparation.

The doc is not exactly great, but the take away is that if we're doing synchronous preparation, we should use `setUpWithError` instead.
___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 89 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX1GZP524TH-BggXg0:-NvX1GZP524TH-BggXg1:b-s6xhwv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadVPNTests/LogRotationTests.swift#L89)):*
> ```Swift
> extension LogRotationTests {
>     private func writeDataToDisk(path: URL, byteSize: Int) throws {
>         let data = Data((0 ..< 1000).map { UInt8($0 & 0xff) })
> ```

Did you mean to use the `byteSize` parameter ?
Also we should probably rename it to `fileSize` since the size of a byte is always 8 bits.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadLogging/LogRotation.swift` line 20 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvWpA509fqPfIwOLq4v:-NvX8RbD7otixmO4Qybp:b-yamqnm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L20)):*

nit
I feel that these parameters are optional just because it made it easier to write tests.
I'm not sure how I feel about this. But IMHO they shouldn't be optional.
</blockquote></details>

The alternative would be to add fake and unrelated size constraints when testing dates and vice versa.
___
*[`ios/MullvadLogging/LogRotation.swift` line 89 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX0dhe8z1pjrSlAMHK:-NvXDkjS9PpPB2ZFVF-D:brxp9sd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L89)):*

nit
What do you think of the following name ?
`private static func deleteLogsOlderThan(_ dateThreshold: Date, in logs[LogData]) throws`
</blockquote></details>

Sounds good!
___
*[`ios/MullvadLogging/LogRotation.swift` line 100 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX-i4VBIMrbTUZ9Qbd:-NvX7D-24bZ5tIuLCehM:btr1oqo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L100)):*

This comment is outdated, it should only refer to log size now.
</blockquote></details>

We still want to delete them from newest to oldest so that we don't shave off new newest logs rather than the oldest.
___
*[`ios/MullvadLogging/LogRotation.swift` line 102 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX13vG7r9MGtisaeQk:-NvX7afZCN2x8X_zthMH:b-axhjml) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L102)):*

This doesn't look to be doing the right thing

The following test will fail
```swift
    func testRotateSize() throws {
        let logPaths = [
            directoryPath.appendingPathComponent("test1.log"),
            directoryPath.appendingPathComponent("test2.log"),
        ]

        try writeDataToDisk(path: logPaths[0], byteSize: 3)
        try writeDataToDisk(path: logPaths[1], byteSize: 10_000)

        try LogRotation.rotateLogs(logDirectory: directoryPath, options: LogRotation.Options(storageSizeLimit: 3999))
        let logFileCount = try fileManager.contentsOfDirectory(atPath: directoryPath.relativePath).count
        XCTAssertEqual(logFileCount, 1)
    }
```
</blockquote></details>

It depends on if log age is relevant. First log is older and should probably be deleted first, then on to the next and so on. 
___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 16 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvWqm3WAE84md1sBo09:-NvX8O1y8flw165qm6ux:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadVPNTests/LogRotationTests.swift#L16)):*

From the docs:
> Override `setUp() async throws` for asynchronous state preparation with error handling before each test. Do not override `setUp(completion:)` for synchronous state preparation.

The doc is not exactly great, but the take away is that if we're doing synchronous preparation, we should use `setUpWithError` instead.
</blockquote></details>

Done.
___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 89 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX1GZP524TH-BggXg0:-NvX83HC0anNDrLKDVbF:b-n486q8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadVPNTests/LogRotationTests.swift#L89)):*

Did you mean to use the `byteSize` parameter ?
Also we should probably rename it to `fileSize` since the size of a byte is always 8 bits.
</blockquote></details>

I did...
Size in bytes is what I wanted to convey. Filesize doesn't say anything about the size unit.

buggmagnet (changes requested):
> Reviewed 4 of 8 files at r1, 1 of 7 files at r2, 3 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110)* status: 10 of 11 files reviewed, 4 unresolved discussions (waiting on @acb-mv and @rablador)

___
*[`ios/MullvadLogging/LogRotation.swift` line 20 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvWpA509fqPfIwOLq4v:-Nv_qngo55v0bkzHiw64:bh00d8k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L20)):*

The alternative would be to add fake and unrelated size constraints when testing dates and vice versa.
</blockquote></details>

I think that's an okay tradeoff. We don't want to have a situation where we mistakenly use an API because it made testing easier.
Both the storage size limit, and the oldest allowed date are a mandatory part of it.

Additionally, we should also have the maximum file size for a single item here, as per the specs.
___
*[`ios/MullvadLogging/LogRotation.swift` line 100 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX-i4VBIMrbTUZ9Qbd:-Nv_rJXvBOgNv97LIaj1:bjdosfy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L100)):*

We still want to delete them from newest to oldest so that we don't shave off new newest logs rather than the oldest.
</blockquote></details>

Yes we do, but we delete them by age in a different function
___
*[`ios/MullvadLogging/LogRotation.swift` line 102 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX13vG7r9MGtisaeQk:-Nv_nqyo248ZzwYdr_te:b-num2gb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L102)):*

It depends on if log age is relevant. First log is older and should probably be deleted first, then on to the next and so on.
</blockquote></details>

This is not about the age of the log, it's about the implementation being incorrect.
In the test, _only_ the file that is 10_000 bytes should be deleted because we want to filter file sizes over 3999 bytes.
However, both files get deleted. 

We want a limit on both
- The size of a single log file
- The total size of all files combined.

i.e. let's say the total limit of logs is 500KB, if we have 300 files at 1KB, that is valid.
If we have 1 file at 200KB and a file at 400KB, the 400KB file should get deleted in the next rotation.
___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 89 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX1GZP524TH-BggXg0:-Nv_pMjy5-8iXy7_ceYN:b-cesuxg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadVPNTests/LogRotationTests.swift#L89)):*

I did...
Size in bytes is what I wanted to convey. Filesize doesn't say anything about the size unit.
</blockquote></details>

Fair enough. Although typically file sizes are always expressed in bytes, it wouldn't make too much sense to express them in another base unit.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110)* status: 9 of 11 files reviewed, 4 unresolved discussions (waiting on @acb-mv and @buggmagnet)

___
*[`ios/MullvadLogging/LogRotation.swift` line 20 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvWpA509fqPfIwOLq4v:-NvaEwU1D64VK7HoRMUU:bqu88wv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L20)):*

I think that's an okay tradeoff. We don't want to have a situation where we mistakenly use an API because it made testing easier.
Both the storage size limit, and the oldest allowed date are a mandatory part of it.

Additionally, we should also have the maximum file size for a single item here, as per the specs.
</blockquote></details>

Fair enough.
As discussed offline, individual file size limit will be added in another PR.
___
*[`ios/MullvadLogging/LogRotation.swift` line 100 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX-i4VBIMrbTUZ9Qbd:-Nva-RpK5locOaN48JwW:b-8b85wc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L100)):*

Yes we do, but we delete them by age in a different function
</blockquote></details>

Discussed and resolved offline.
___
*[`ios/MullvadLogging/LogRotation.swift` line 102 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX13vG7r9MGtisaeQk:-Nval1qU3dlhI4qAzwVP:b-8b85wc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadLogging/LogRotation.swift#L102)):*

This is not about the age of the log, it's about the implementation being incorrect.
In the test, _only_ the file that is 10_000 bytes should be deleted because we want to filter file sizes over 3999 bytes.
However, both files get deleted. 

We want a limit on both
- The size of a single log file
- The total size of all files combined.

i.e. let's say the total limit of logs is 500KB, if we have 300 files at 1KB, that is valid.
If we have 1 file at 200KB and a file at 400KB, the 400KB file should get deleted in the next rotation.
</blockquote></details>

Discussed and resolved offline.
___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 89 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-NvX1GZP524TH-BggXg0:-NvaEq3jCPdGdSPz9QnQ:b-cox9bo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c19d2ef72785e195bfaa7cb2566e960a6e54f7ee/ios/MullvadVPNTests/LogRotationTests.swift#L89)):*

Fair enough. Although typically file sizes are always expressed in bytes, it wouldn't make too much sense to express them in another base unit.
</blockquote></details>

You might be right. Let's go with simply `fileSize`.

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6110#-:-Nvf0GqA5tjDWLiufMEN:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6123, opened 2024-04-15
Add default settings when app is uninstalled

rablador:
We wipe all settings when the app is deleted and then installed again. Currently no default settings are stored after that, which puts the tunnel into blocked state since it cannot read any settings. We should add the default settings right after the wipe to remedy this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6123)

### DIFF
--- a/ios/MullvadVPN/AppDelegate.swift
+++ b/ios/MullvadVPN/AppDelegate.swift
@@ -502,7 +502,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
                 }
 
                 SettingsManager.resetStore(completely: true)
+                try? SettingsManager.writeSettings(LatestTunnelSettings())
 
+                // Default access methods need to be repopulated again after settings wipe.
                 self.accessMethodRepository.reloadWithDefaultsAfterDataRemoval()
                 // At app startup, the relay cache tracker will get populated with a list of overriden IPs.
                 // The overriden IPs will get wiped, therefore, the cache needs to be pruned as well.


mojganii (approved):
:lgtm:
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6123#-:-NvX28P88x7aYEjMFuGa:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 500 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6123#-NvX29BeBb1J1EoU6rYR:-NvX29BeBb1J1EoU6rYS:b-lz8l9o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/89b0c94a52b82cf5d94f211605bbd5f3c9406e10/ios/MullvadVPN/AppDelegate.swift#L500)):*
> ```Swift
>                         retryStrategy: .noRetry
>                     ) { _ in
>                         // Do nothing.
> ```

nit : maybe it's better to make the closure `optional` instead of  putting the comment `Do nothing.`

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6123)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @mojganii)

___
*[`ios/MullvadVPN/AppDelegate.swift` line 500 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6123#-NvX29BeBb1J1EoU6rYR:-NvX6t9z0R1MYCK6nF3n:b-ynzvdk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/89b0c94a52b82cf5d94f211605bbd5f3c9406e10/ios/MullvadVPN/AppDelegate.swift#L500)):*

nit : maybe it's better to make the closure `optional` instead of  putting the comment `Do nothing.`
</blockquote></details>

I think I agree, but I'll let this one be for now since it's not really a part of this PR and we need to complete the review as soon as we can.

PR #6148, opened 2024-04-18
Add UITest for creating a custom list Part 1

buggmagnet:
This PR adds support for writing UI Tests for custom lists.
As this is a rather large feature, this PR only adds a simple, single test case for the moment, creating a custom list, and verify that it still exists.

This PR also disables animations in the application for all UI Tests scenarios, and adds a new shorthand syntax for accessing elements from UITests.

More refactor to come later.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)

### DIFF
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -703,6 +703,8 @@
 		A988A3E22AFE54AC0008D2C7 /* AccountExpiry.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7A6F2FA62AFBB9AE006D0856 /* AccountExpiry.swift */; };
 		A988DF272ADE86ED00D807EF /* WireGuardObfuscationSettings.swift in Sources */ = {isa = PBXBuildFile; fileRef = A988DF252ADE86ED00D807EF /* WireGuardObfuscationSettings.swift */; };
 		A988DF2A2ADE880300D807EF /* TunnelSettingsV3.swift in Sources */ = {isa = PBXBuildFile; fileRef = A988DF282ADE880300D807EF /* TunnelSettingsV3.swift */; };
+		A998DA812BD147AD001D61A2 /* ListCustomListsPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = A998DA802BD147AD001D61A2 /* ListCustomListsPage.swift */; };
+		A998DA832BD2B055001D61A2 /* EditCustomListLocationsPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = A998DA822BD2B055001D61A2 /* EditCustomListLocationsPage.swift */; };
 		A99E5EE02B7628150033F241 /* ProblemReportViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = A99E5EDF2B7628150033F241 /* ProblemReportViewModel.swift */; };
 		A99E5EE22B762ED30033F241 /* ProblemReportViewController+ViewManagement.swift in Sources */ = {isa = PBXBuildFile; fileRef = A99E5EE12B762ED30033F241 /* ProblemReportViewController+ViewManagement.swift */; };
 		A9A1DE792AD5708E0073F689 /* TransportStrategy.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9A1DE782AD5708E0073F689 /* TransportStrategy.swift */; };
@@ -808,6 +810,8 @@
 		A9B6AC1B2ADEA3AD00F7802A /* MemoryCache.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58BDEB9C2A98F69E00F578F2 /* MemoryCache.swift */; };
 		A9BA08312BA32FA9005A7A2D /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = A92962582B1F4FDB00DFB93B /* PrivacyInfo.xcprivacy */; };
 		A9BA08322BA32FB6005A7A2D /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = A92962582B1F4FDB00DFB93B /* PrivacyInfo.xcprivacy */; };
+		A9BFAFFF2BD004ED00F2BCA1 /* CustomListsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9BFAFFE2BD004ED00F2BCA1 /* CustomListsTests.swift */; };
+		A9BFB0012BD00B7F00F2BCA1 /* CustomListPage.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9BFB0002BD00B7F00F2BCA1 /* CustomListPage.swift */; };
 		A9C342C12ACC37E30045F00E /* TunnelStatusBlockObserver.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9E0317D2ACC32920095D843 /* TunnelStatusBlockObserver.swift */; };
 		A9C342C32ACC3EE90045F00E /* RelayCacheTracker+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9C342C22ACC3EE90045F00E /* RelayCacheTracker+Stubs.swift */; };
 		A9C342C52ACC42130045F00E /* ServerRelaysResponse+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9C342C42ACC42130045F00E /* ServerRelaysResponse+Stubs.swift */; };
@@ -1966,6 +1970,8 @@
 		A98502022B627B120061901E /* LocalNetworkProbe.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LocalNetworkProbe.swift; sourceTree = "<group>"; };
 		A988DF252ADE86ED00D807EF /* WireGuardObfuscationSettings.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = WireGuardObfuscationSettings.swift; sourceTree = "<group>"; };
 		A988DF282ADE880300D807EF /* TunnelSettingsV3.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TunnelSettingsV3.swift; sourceTree = "<group>"; };
+		A998DA802BD147AD001D61A2 /* ListCustomListsPage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ListCustomListsPage.swift; sourceTree = "<group>"; };
+		A998DA822BD2B055001D61A2 /* EditCustomListLocationsPage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = EditCustomListLocationsPage.swift; sourceTree = "<group>"; };
 		A99E5EDF2B7628150033F241 /* ProblemReportViewModel.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ProblemReportViewModel.swift; sourceTree = "<group>"; };
 		A99E5EE12B762ED30033F241 /* ProblemReportViewController+ViewManagement.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "ProblemReportViewController+ViewManagement.swift"; sourceTree = "<group>"; };
 		A9A1DE782AD5708E0073F689 /* TransportStrategy.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = TransportStrategy.swift; sourceTree = "<group>"; };
@@ -1973,6 +1979,8 @@
 		A9A8A8EA2A262AB30086D569 /* FileCache.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = FileCache.swift; sourceTree = "<group>"; };
 		A9B6AC172ADE8F4300F7802A /* MigrationManagerTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MigrationManagerTests.swift; sourceTree = "<group>"; };
 		A9B6AC192ADE8FBB00F7802A /* InMemorySettingsStore.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = InMemorySettingsStore.swift; sourceTree = "<group>"; };
+		A9BFAFFE2BD004ED00F2BCA1 /* CustomListsTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CustomListsTests.swift; sourceTree = "<group>"; };
+		A9BFB0002BD00B7F00F2BCA1 /* CustomListPage.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CustomListPage.swift; sourceTree = "<group>"; };
 		A9C342C22ACC3EE90045F00E /* RelayCacheTracker+Stubs.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "RelayCacheTracker+Stubs.swift"; sourceTree = "<group>"; };
 		A9C342C42ACC42130045F00E /* ServerRelaysResponse+Stubs.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "ServerRelaysResponse+Stubs.swift"; sourceTree = "<group>"; };
 		A9CF11FC2A0518E7001D9565 /* AddressCacheTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AddressCacheTests.swift; sourceTree = "<group>"; };
@@ -3623,19 +3631,20 @@
 		852969262B4D9C1F007EAD4C /* MullvadVPNUITests */ = {
 			isa = PBXGroup;
 			children = (
-				85B267602B849ADB0098E3CD /* mullvad-api.h */,
 				852969272B4D9C1F007EAD4C /* AccountTests.swift */,
 				8556EB532B9A1D7100D26DD4 /* BridgingHeader.h */,
 				85557B112B594FC900795FE1 /* ConnectivityTests.swift */,
+				A9BFAFFE2BD004ED00F2BCA1 /* CustomListsTests.swift */,
 				852969372B4ED20E007EAD4C /* Info.plist */,
+				85B267602B849ADB0098E3CD /* mullvad-api.h */,
 				85557B0C2B591B0F00795FE1 /* Networking */,
 				852969312B4E9220007EAD4C /* Pages */,
 				850201DA2B503D7700EF8C96 /* RelayTests.swift */,
-				8518F6392B601910009EB113 /* Test base classes */,
-				85557B152B5ABBBE00795FE1 /* XCUIElementQuery+Extensions.swift */,
-				85A42B852BB1D627007BABF7 /* XCUIElement+Extensions.swift */,
-				85C7A2E82B89024B00035D5A /* SettingsTests.swift */,
 				85D039972BA4711800940E7F /* SettingsMigrationTests.swift */,
+				85C7A2E82B89024B00035D5A /* SettingsTests.swift */,
+				8518F6392B601910009EB113 /* Test base classes */,
+				85A42B852BB1D627007BABF7 /* XCUIElement+Extensions.swift */,
+				85557B152B5ABBBE00795FE1 /* XCUIElementQuery+Extensions.swift */,
 			);
 			path = MullvadVPNUITests;
 			sourceTree = "<group>";
@@ -3647,9 +3656,12 @@
 				85557B1F2B5FBBD700795FE1 /* AccountPage.swift */,
 				8529693B2B4F0257007EAD4C /* Alert.swift */,
 				8587A05C2B84D43100152938 /* ChangeLogAlert.swift */,
-				852A26452BA9C9CB006EB9C8 /* DNSSettingsPage.swift */,
+				A9BFB0002BD00B7F00F2BCA1 /* CustomListPage.swift */,
 				85A42B872BB44D31007BABF7 /* DeviceManagementPage.swift */,
+				852A26452BA9C9CB006EB9C8 /* DNSSettingsPage.swift */,
+				A998DA822BD2B055001D61A2 /* EditCustomListLocationsPage.swift */,
 				85557B1D2B5FB8C700795FE1 /* HeaderBar.swift */,
+				A998DA802BD147AD001D61A2 /* ListCustomListsPage.swift */,
 				852969342B4E9270007EAD4C /* LoginPage.swift */,
 				85139B2C2B84B4A700734217 /* OutOfTimePage.swift */,
 				852969322B4E9232007EAD4C /* Page.swift */,
@@ -5639,6 +5651,7 @@
 			isa = PBXSourcesBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				A9BFB0012BD00B7F00F2BCA1 /* CustomListPage.swift in Sources */,
 				8556EB522B9A1C6900D26DD4 /* MullvadApi.swift in Sources */,
 				85EC620C2B838D10005AFFB5 /* MullvadAPIWrapper.swift in Sources */,
 				A9DF789D2B7D1E8B0094E4AD /* LoggedInWithTimeUITestCase.swift in Sources */,
@@ -5655,7 +5668,9 @@
 				8590896F2B61763B003AF5F5 /* LoggedOutUITestCase.swift in Sources */,
 				85557B202B5FBBD700795FE1 /* AccountPage.swift in Sources */,
 				852969352B4E9270007EAD4C /* LoginPage.swift in Sources */,
+				A998DA832BD2B055001D61A2 /* EditCustomListLocationsPage.swift in Sources */,
 				8556EB562B9B0AC500D26DD4 /* RevokedDevicePage.swift in Sources */,
+				A9BFAFFF2BD004ED00F2BCA1 /* CustomListsTests.swift in Sources */,
 				85557B102B59215F00795FE1 /* FirewallRule.swift in Sources */,
 				85557B0E2B591B2600795FE1 /* FirewallAPIClient.swift in Sources */,
 				852969282B4D9C1F007EAD4C /* AccountTests.swift in Sources */,
@@ -5668,6 +5683,7 @@
 				8532E6872B8CCED600ACECD1 /* ProblemReportSubmittedPage.swift in Sources */,
 				85FB5A0C2B6903990015DCED /* WelcomePage.swift in Sources */,
 				852A26462BA9C9CB006EB9C8 /* DNSSettingsPage.swift in Sources */,
+				A998DA812BD147AD001D61A2 /* ListCustomListsPage.swift in Sources */,
 				850201DF2B5040A500EF8C96 /* TunnelControlPage.swift in Sources */,
 				8542CE242B95F7B9006FCA14 /* VPNSettingsPage.swift in Sources */,
 				85557B1E2B5FB8C700795FE1 /* HeaderBar.swift in Sources */,
--- a/ios/MullvadVPN/Classes/AccessbilityIdentifier.swift
+++ b/ios/MullvadVPN/Classes/AccessbilityIdentifier.swift
@@ -37,12 +37,21 @@ public enum AccessibilityIdentifier: String {
     case restorePurchasesButton
     case secureConnectionButton
     case selectLocationButton
+    case closeSelectLocationButton
     case settingsButton
     case startUsingTheAppButton
     case problemReportAppLogsButton
     case problemReportSendButton
     case relayStatusCollapseButton
     case settingsDoneButton
+    case openCustomListsMenuButton
+    case addNewCustomListButton
+    case editCustomListButton
+    case saveCreateCustomListButton
+    case confirmDeleteCustomListButton
+    case cancelDeleteCustomListButton
+    case customListLocationCheckmarkButton
+    case listCustomListDoneButton
 
     // Cells
     case deviceCell
@@ -60,6 +69,9 @@ public enum AccessibilityIdentifier: String {
     case wireGuardObfuscationCell
     case udpOverTCPPortCell
     case quantumResistantTunnelCell
+    case customListEditNameFieldCell
+    case customListEditAddOrEditLocationCell
+    case customListEditDeleteListCell
 
     // Labels
     case accountPagePaidUntilLabel
@@ -71,7 +83,6 @@ public enum AccessibilityIdentifier: String {
 
     // Views
     case accountView
-    case addLocationsView
     case alertContainerView
     case alertTitle
     case changeLogAlert
@@ -91,6 +102,12 @@ public enum AccessibilityIdentifier: String {
     case welcomeView
     case deleteAccountView
     case settingsContainerView
+    case newCustomListView
+    case customListEditTableView
+    case listCustomListsView
+    case listCustomListsTableView
+    case editCustomListEditLocationsView
+    case editCustomListEditLocationsTableView
 
     // Other UI elements
     case connectionPanelInAddressRow
--- a/ios/MullvadVPN/Coordinators/CustomLists/AddLocationsViewController.swift
+++ b/ios/MullvadVPN/Coordinators/CustomLists/AddLocationsViewController.swift
@@ -27,7 +27,7 @@ class AddLocationsViewController: UIViewController {
         tableView.separatorInset = .zero
         tableView.rowHeight = 56
         tableView.indicatorStyle = .white
-        tableView.accessibilityIdentifier = .addLocationsView
+        tableView.accessibilityIdentifier = .editCustomListEditLocationsTableView
         return tableView
     }()
 
@@ -46,6 +46,7 @@ class AddLocationsViewController: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        view.accessibilityIdentifier = .editCustomListEditLocationsView
         tableView.backgroundColor = view.backgroundColor
         view.backgroundColor = .secondaryColor
         addConstraints()
--- a/ios/MullvadVPN/Coordinators/CustomLists/CustomListCellConfiguration.swift
+++ b/ios/MullvadVPN/Coordinators/CustomLists/CustomListCellConfiguration.swift
@@ -78,6 +78,7 @@ struct CustomListCellConfiguration {
         contentConfiguration.maxLength = 30
         contentConfiguration.editingEvents.onChange = subject.bindTextAction(to: \.name)
 
+        cell.accessibilityIdentifier = AccessibilityIdentifier.customListEditNameFieldCell
         cell.contentConfiguration = contentConfiguration
     }
 
@@ -86,6 +87,7 @@ struct CustomListCellConfiguration {
 
         contentConfiguration.text = itemIdentifier.text
         cell.contentConfiguration = contentConfiguration
+        cell.accessibilityIdentifier = AccessibilityIdentifier.customListEditAddOrEditLocationCell
 
         if let cell = cell as? CustomCellDisclosureHandling {
             cell.disclosureType = .chevron
@@ -101,6 +103,7 @@ struct CustomListCellConfiguration {
             onDelete?()
         }
 
+        cell.accessibilityIdentifier = AccessibilityIdentifier.customListEditDeleteListCell
         cell.contentConfiguration = contentConfiguration
     }
 }
--- a/ios/MullvadVPN/Coordinators/CustomLists/CustomListViewController.swift
+++ b/ios/MullvadVPN/Coordinators/CustomLists/CustomListViewController.swift
@@ -58,6 +58,7 @@ class CustomListViewController: UIViewController {
             }
         )
         barButtonItem.style = .done
+        barButtonItem.accessibilityIdentifier = AccessibilityIdentifier.saveCreateCustomListButton
 
         return barButtonItem
     }()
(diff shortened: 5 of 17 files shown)


niklasberglund (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)* status: 0 of 11 files reviewed, 8 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/Coordinators/LocationCoordinator.swift` line 215 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl2MXH-yoKuDZweYRn:-Nvl2MXH-yoKuDZweYRo:b-9tkcp8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPN/Coordinators/LocationCoordinator.swift#L215)):*
> ```Swift
>             }
>         )
>         addCustomListAction.accessibilityIdentifier = AccessibilityIdentifier.addNewCustomListButton
> ```

Since the property is of `AccessibilityIdentifier` it can be simplified to just `.addNewCustomListButton`. I saw you used this syntax in some places but some not.
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl8ySHCCnJXXRc5wxm:-Nvl8ySHCCnJXXRc5wxn:bay3vly) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/CustomListsTests.swift#L17)):*
> ```Swift
> 
>         SelectLocationPage(app)
>             .openCustomListsActions()
> ```

I think it's a very useful convention to only have functions doing *one* user action each in test case level. It makes it easy to glance the test case code and see what the test is doing without having to look in multiple functions. This way it's easier to make sure the tests are testing what they should, and it reduces(or eliminates) the need of documenting test case user actions elsewhere.

So for example `openCustomListsActions()` could split into:
```
.scrollToCustomListsSection()
.tapCustomListsEllipsisButton()
```
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-NvlAiEb3bhEjguQEAfA:-NvlAiEb3bhEjguQEAfB:bxp9k68) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/CustomListsTests.swift#L24)):*
> ```Swift
>         CustomListPage(app)
>             .verifyCreateButtonIs(enabled: false)
>             .editCustomList(name: customListName)
> ```

Same comment as a above, IMO it's very helpful to not combine user actions in methods but doing one user action per function in test case level. So this could be for example
```
.tapNameTextField()
.enterText(customListName)
```

(The `Page` superclass have some generic functions like `enterText()`)

After entering the custom list name the keyboard is still visible. I'm not sure what a real user is more likely to do here and what's better to test but could either continue navigation without dismissing keyboard or dismiss keyboard first. Not sure if it's important to dismiss keyboard here but the page superclass `Page` has a function `dimissKeyboard()` if deciding to dismiss keyboard.
___
*[`ios/MullvadVPNUITests/XCUIElementQuery+Extensions.swift` line 19 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl3-tx-ypR70P2crY_:-Nvl3-tx-ypR70P2crYa:b-3gbqy1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/XCUIElementQuery%2BExtensions.swift#L19)):*
> ```Swift
>     subscript(key: AccessibilityIdentifier) -> XCUIElement {
>         self[key.rawValue]
>     }
> ```

nit: super neat 👍👍
___
*[`ios/MullvadVPNUITests/Pages/CustomListPage.swift` line 19 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl6f1t8d1zzGc48_Ls:-Nvl6f1t8d1zzGc48_Lt:b-78s1wl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Pages/CustomListPage.swift#L19)):*
> ```Swift
>     }
> 
>     @discardableResult
> ```

nit: Seems like `swiftformat` allows both `@discardableResult` on same line as function declaration and on separate line above. I have doing it on same line previously but don't have any strong preference, kinda like both. Should we try to agree on one way of doing it and sticking to it?
___
*[`ios/MullvadVPNUITests/Pages/SelectLocationPage.swift` line 52 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl5c8T3HuUhcD_nWHY:-Nvl5c8T3HuUhcD_nWHZ:b-pkieij) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Pages/SelectLocationPage.swift#L52)):*
> ```Swift
>     func scrollToCustomListsSection() -> Self {
>         let selectLocationTableView = app.tables[AccessibilityIdentifier.selectLocationTableView]
>         selectLocationTableView.swipeDown(velocity: XCUIGestureVelocity(floatLiteral: 9999))
> ```

`XCUIGestureVelociy` have some constants that make it easier to read out how fast the velocity is(`.default`, `.slow`, `.fast`) https://developer.apple.com/documentation/xctest/xcuigesturevelocity#3638600

9999 pixels per second seems very fast. I think if not adding much time or complexity it's nice to do it as human-like as possible. Another fast and reliable way of scrolling to the top is tapping the iOS status bar.
___
*[`ios/MullvadVPNUITests/Pages/SelectLocationPage.swift` line 56 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl7Jgp2ByGGZ0xyxPi:-Nvl7Jgp2ByGGZ0xyxPj:b-8fbp4o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Pages/SelectLocationPage.swift#L56)):*
> ```Swift
>     }
> 
>     func tapAddNewCustomList() {
> ```

I have made all page functions chainable by returning `self` regardless of whether another action on the same page might follow the function invocation or not, because the app's flow might change and I've been treating it as a protocol to be followed even though there is no such protocol. An argument could be made that it is invalid to continue chaining after these functions though, so your approach here also makes sense. What's your thoughts?
___
*[`ios/MullvadVPNUITests/Test base classes/BaseUITestCase.swift` line 69 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl4Drt8m8IOXK-mQ-g:-Nvl4Drt8m8IOXK-mQ-h:bynmst9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Test%20base%20classes/BaseUITestCase.swift#L69)):*
> ```Swift
>     override func setUp() {
>         continueAfterFailure = false
>         app.launchArguments = ["DisableAnimations"]
> ```

Should we disable animations? It will make the tests run faster but at the cost of the app being tested differing slightly from the app users use.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)* status: 0 of 11 files reviewed, 8 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPN/Coordinators/LocationCoordinator.swift` line 215 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl2MXH-yoKuDZweYRn:-Nvp-EGpCABDnpZ56iDu:b1dl8ew) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPN/Coordinators/LocationCoordinator.swift#L215)):*

Since the property is of `AccessibilityIdentifier` it can be simplified to just `.addNewCustomListButton`. I saw you used this syntax in some places but some not.
</blockquote></details>

Yes that's correct, I want to do another pull request where we only modify all those `AccessibilityIdentifier` in one swoop, because otherwise the changes in the codebase will be harder to review
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl8ySHCCnJXXRc5wxm:-Nvp24Ez-Qv-za-BWCEc:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/CustomListsTests.swift#L17)):*

I think it's a very useful convention to only have functions doing *one* user action each in test case level. It makes it easy to glance the test case code and see what the test is doing without having to look in multiple functions. This way it's easier to make sure the tests are testing what they should, and it reduces(or eliminates) the need of documenting test case user actions elsewhere.

So for example `openCustomListsActions()` could split into:
```
.scrollToCustomListsSection()
.tapCustomListsEllipsisButton()
```
</blockquote></details>

Done.
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-NvlAiEb3bhEjguQEAfA:-Nvp2K7k2SIke7GazYnm:b6semjz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/CustomListsTests.swift#L24)):*

Same comment as a above, IMO it's very helpful to not combine user actions in methods but doing one user action per function in test case level. So this could be for example
```
.tapNameTextField()
.enterText(customListName)
```

(The `Page` superclass have some generic functions like `enterText()`)

After entering the custom list name the keyboard is still visible. I'm not sure what a real user is more likely to do here and what's better to test but could either continue navigation without dismissing keyboard or dismiss keyboard first. Not sure if it's important to dismiss keyboard here but the page superclass `Page` has a function `dimissKeyboard()` if deciding to dismiss keyboard.
</blockquote></details>

On principle I agree with you, but I'd like to argue that in this specific case, we are doing one action only.
I agree with the general idea that we want each action to do _one thing only_ however, one cannot edit a field without tapping it first. 
Imagine we had to write a test for a form with several text fields, it would be extremely redundant to see the following 

```swift
form.tapTextField1()
form.enterText(someText1)

form.tapTextField2()
form.enterText(someText2)

form.tapTextField3()
form.enterText(someText3)
```

This is also why I tried to make the method name read as naturally as possible, there is no possible confusion here as to what the single action is doing.
___
*[`ios/MullvadVPNUITests/Pages/CustomListPage.swift` line 19 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl6f1t8d1zzGc48_Ls:-Nvp-b-G5ddhKfZKW5PH:b-x8nveg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Pages/CustomListPage.swift#L19)):*

nit: Seems like `swiftformat` allows both `@discardableResult` on same line as function declaration and on separate line above. I have doing it on same line previously but don't have any strong preference, kinda like both. Should we try to agree on one way of doing it and sticking to it?
</blockquote></details>

That's a good point, I'll put it on the same line
___
*[`ios/MullvadVPNUITests/Pages/SelectLocationPage.swift` line 52 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl5c8T3HuUhcD_nWHY:-Nvp-iDr1Hu93kTUD48z:b-ofh3mu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Pages/SelectLocationPage.swift#L52)):*

`XCUIGestureVelociy` have some constants that make it easier to read out how fast the velocity is(`.default`, `.slow`, `.fast`) https://developer.apple.com/documentation/xctest/xcuigesturevelocity#3638600

9999 pixels per second seems very fast. I think if not adding much time or complexity it's nice to do it as human-like as possible. Another fast and reliable way of scrolling to the top is tapping the iOS status bar.
</blockquote></details>

This was discussed offline, I will try to find a way to tap the status bar instead.

EDIT:
It used to be possible before iOS 12, since then, the status bar is not considered part of the application's UI anymore, therefore it's not a good method.

I will leave it like this as it's reliable enough.
___
*[`ios/MullvadVPNUITests/Pages/SelectLocationPage.swift` line 56 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl7Jgp2ByGGZ0xyxPi:-Nvp-sXE6jA-dHjjQ2Qg:bmsvcr6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Pages/SelectLocationPage.swift#L56)):*

I have made all page functions chainable by returning `self` regardless of whether another action on the same page might follow the function invocation or not, because the app's flow might change and I've been treating it as a protocol to be followed even though there is no such protocol. An argument could be made that it is invalid to continue chaining after these functions though, so your approach here also makes sense. What's your thoughts?
</blockquote></details>

I've added chaining when I needed it, but I think it's useful as a convention to have, especially because we can use `@discardableResult` 
I will add it too here 
___
*[`ios/MullvadVPNUITests/Test base classes/BaseUITestCase.swift` line 69 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl4Drt8m8IOXK-mQ-g:-Nvp044p92ZvUfI8-lO9:b-e5zl10) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Test%20base%20classes/BaseUITestCase.swift#L69)):*

Should we disable animations? It will make the tests run faster but at the cost of the app being tested differing slightly from the app users use.
</blockquote></details>

That's true. Also I noticed that even with animation disabled, we still show some of the views with an animation, which is not great.
Ultimately, we want the UITests to run as fast as possible, and since animations are not really easy to verify with UITests in the first place, I think it's okay to have them disabled in UITests as well.

niklasberglund (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)* status: 0 of 11 files reviewed, 8 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/Coordinators/LocationCoordinator.swift` line 215 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl2MXH-yoKuDZweYRn:-NvpOeC01EV5vvbC5jM7:b-2zwhok) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPN/Coordinators/LocationCoordinator.swift#L215)):*

Yes that's correct, I want to do another pull request where we only modify all those `AccessibilityIdentifier` in one swoop, because otherwise the changes in the codebase will be harder to review
</blockquote></details>

Sounds good 👍
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-NvlAiEb3bhEjguQEAfA:-NvpOymH7koGV4TNEgJJ:byh1ges) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/CustomListsTests.swift#L24)):*

On principle I agree with you, but I'd like to argue that in this specific case, we are doing one action only.
I agree with the general idea that we want each action to do _one thing only_ however, one cannot edit a field without tapping it first. 
Imagine we had to write a test for a form with several text fields, it would be extremely redundant to see the following 

```swift
form.tapTextField1()
form.enterText(someText1)

form.tapTextField2()
form.enterText(someText2)

form.tapTextField3()
form.enterText(someText3)
```

This is also why I tried to make the method name read as naturally as possible, there is no possible confusion here as to what the single action is doing.
</blockquote></details>

Agree with you that it is a quite specific action to enter a text into a textfield, but there is room for confusion I think. To me it's not obvious that `editCustomList(name: customListName)` enters this name for the custom list by just reading the code. To edit a custom list could mean other things too.
___
*[`ios/MullvadVPNUITests/Pages/SelectLocationPage.swift` line 52 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl5c8T3HuUhcD_nWHY:-NvpQ2u9DFADu5pfh8F2:b-m78zxv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Pages/SelectLocationPage.swift#L52)):*

This was discussed offline, I will try to find a way to tap the status bar instead.

EDIT:
It used to be possible before iOS 12, since then, the status bar is not considered part of the application's UI anymore, therefore it's not a good method.

I will leave it like this as it's reliable enough.
</blockquote></details>

Then the current solution sounds good to me 👍
___
*[`ios/MullvadVPNUITests/Test base classes/BaseUITestCase.swift` line 69 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl4Drt8m8IOXK-mQ-g:-NvpQNuN7AZS0RsT-cMH:b4wxzft) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Test%20base%20classes/BaseUITestCase.swift#L69)):*

That's true. Also I noticed that even with animation disabled, we still show some of the views with an animation, which is not great.
Ultimately, we want the UITests to run as fast as possible, and since animations are not really easy to verify with UITests in the first place, I think it's okay to have them disabled in UITests as well.
</blockquote></details>

Personally I'm not really for or against disabling animations. Both have their benefits. Check with the team after standup?

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)* status: 0 of 18 files reviewed, 7 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPNUITests/Pages/SelectLocationPage.swift` line 52 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl5c8T3HuUhcD_nWHY:-Nw47ZPm1jZkT1XLrI49:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Pages/SelectLocationPage.swift#L52)):*

Then the current solution sounds good to me 👍
</blockquote></details>

Done.
___
*[`ios/MullvadVPNUITests/Test base classes/BaseUITestCase.swift` line 69 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nvl4Drt8m8IOXK-mQ-g:-Nw47Sot00wZLbAR-s_X:b-teivys) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/Test%20base%20classes/BaseUITestCase.swift#L69)):*

Personally I'm not really for or against disabling animations. Both have their benefits. Check with the team after standup?
</blockquote></details>

We decided during the standup to disable animations for UI Tests

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)* status: 0 of 18 files reviewed, 7 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 24 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-NvlAiEb3bhEjguQEAfA:-Nw4eIqX9uGRrMxzihMC:b-je3jpt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5bf738aba5f3a857299de760d1d3c83f95c2fa44/ios/MullvadVPNUITests/CustomListsTests.swift#L24)):*

Agree with you that it is a quite specific action to enter a text into a textfield, but there is room for confusion I think. To me it's not obvious that `editCustomList(name: customListName)` enters this name for the custom list by just reading the code. To edit a custom list could mean other things too.
</blockquote></details>

This was discussed offline, we will rename this `renameCustomList`

niklasberglund (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)* status: 0 of 18 files reviewed, 16 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/View controllers/SelectLocation/LocationCell.swift` line 143 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4hjdc6WsMLb8djPYv:-Nw4hjdc6WsMLb8djPYw:b-2eovha) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPN/View%20controllers/SelectLocation/LocationCell.swift#L143)):*
> ```Swift
> 
>         checkboxButton.addTarget(self, action: #selector(toggleCheckboxButton(_:)), for: .touchUpInside)
> //        checkboxButton.accessibilityIdentifier
> ```

Should this be removed?
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 28 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4iYYd4XI7e6VzpJry:-Nw4iYYd4XI7e6VzpJrz:b-2r8zz1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L28)):*
> ```Swift
>             .tapWhereStatusBarShouldBeToScrollToTopMostPosition()
> 
>         XCTAssertTrue(app.staticTexts[customListName].exists)
> ```

nit: it would be nice to specifically look for custom list cell with this name. In the current implementation I think this check is fine, but there is a risk that the custom list name in the future could appear elsewhere in the view hierarchy.
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 38 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4lADRASPX17jV50Ow:-Nw4lADRASPX17jV50Ox:b-xpsruf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L38)):*
> ```Swift
>         createCustomList(named: customListName)
>         workaroundOpenCustomListMenuBug()
>         deleteCustomList(named: customListName)
> ```

Since deleting the custom list is part of the actual test I suggest breaking it out to represent user interactions. IMO automated test cases should read like manual test cases. If a manual test says "delete custom list" its leaving much room for interpretation of the tester. Maybe later for example it's possible to delete by swiping left on the cell then tapping a delete button. It's useful to see how the item is deleted directly in the test case. 


___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 43 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw52Ebb6B0eZm3X5_4P:-Nw52Ebb6B0eZm3X5_4Q:b-8vhp32) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L43)):*
> ```Swift
>             .tapWhereStatusBarShouldBeToScrollToTopMostPosition()
> 
>         XCTAssertFalse(app.staticTexts[customListName].exists)
> ```

When running `testDeleteCustomList` its failing here for me because the select location table view is in the view hierarchy in the background and has not been updated. See screenshot where I created custom lists "1" and "2" then removed "1". I think the assert should be more specific and look for custom list cell title with this text. 

___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 53 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4nVZW7Be2Xfj7d0Cr:-Nw4nVZW7Be2Xfj7d0Cs:b-ia7xr8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L53)):*
> ```Swift
>         createCustomList(named: customListName)
>         workaroundOpenCustomListMenuBug()
>         startEditingCustomList(named: customListName)
> ```

Same comment as above(https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4lADRASPX17jV50Ow) since this is part of the actual test. In setup and teardown it's good to use functions combining actions.
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 64 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4nnxeDB8FBneBT7EY:-Nw4nnxeDB8FBneBT7EZ:b-awmcbb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L64)):*
> ```Swift
> 
>         ListCustomListsPage(app)
>             .finishEditingCustomLists()
> ```

The function does one user action but the name is ambiguous about specifically how editing custom lists is finished. Suggest for example `tapDoneButton()`
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 69 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4paHF0bw6WWoGcc7t:-Nw4paHF0bw6WWoGcc7u:bqi172g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L69)):*
> ```Swift
> 
>         workaroundOpenCustomListMenuBug()
>         deleteCustomList(named: customListName)
> ```

Teardown code should be in teardown when possible so that it always run. `addTeardownBlock` is useful when a teardown is specific to a test and not all tests in the suite. Could for example do something like this at the top if this test:
```
addTeardownBlock {
    TunnelControlPage(app)
        .tapSelectLocationButton()
    
    SelectLocationPage(app)
            .tapWhereStatusBarShouldBeToScrollToTopMostPosition()
            .tapCustomListEllipsisButton()

    deleteCustomList(named: customListName)
}
```
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 86 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw52dupDHx5zvZAgGAp:-Nw52dupDHx5zvZAgGAq:bdfmqdx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L86)):*
> ```Swift
>     }
> 
>     func workaroundOpenCustomListMenuBug() {
> ```

Should this be addressed in the app? It might not only affect tests but also users relying on accessibility features.
___
*[`ios/MullvadVPNUITests/Pages/Page.swift` line 51 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4rc-V8hnTJjhO020N:-Nw4rc-V8hnTJjhO020O:bzd4yw5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/Pages/Page.swift#L51)):*
> ```Swift
> 
>     @discardableResult func tapWhereStatusBarShouldBeToScrollToTopMostPosition() -> Self {
>         app.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0)).tap()
> ```

Will this tap be in the upper left corner? Wondering if it can cause issues in some scenarios for example with dynamic island or when iOS is showing that you're sharing location data, are in a call or similar. Then it's a tappable button in the upper left corner.

Would this tap in the middle of the status bar? 
`app.coordinate(withNormalizedOffset: CGVector(dx: app.frame.width/2, dy: 10)).tap()`

I think status bar always(or in almost all cases?) are 20 points high

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)* status: 0 of 18 files reviewed, 16 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPN/View controllers/SelectLocation/LocationCell.swift` line 143 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4hjdc6WsMLb8djPYv:-Nw8r1N7AbisesEx1bAQ:b-5bamig) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPN/View%20controllers/SelectLocation/LocationCell.swift#L143)):*

Should this be removed?
</blockquote></details>

yes it should :see_no_evil: 
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 38 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4lADRASPX17jV50Ow:-Nw8rKqL5fwRtNe98vEg:b-7gpu2j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L38)):*

Since deleting the custom list is part of the actual test I suggest breaking it out to represent user interactions. IMO automated test cases should read like manual test cases. If a manual test says "delete custom list" its leaving much room for interpretation of the tester. Maybe later for example it's possible to delete by swiping left on the cell then tapping a delete button. It's useful to see how the item is deleted directly in the test case. 


</blockquote></details>

I hear what you're saying, but we're using this function in all the tests so far, and it's less error prone to have one function that does it than repeat all the steps in all the tests.

We are not supporting swipe to delete yet, so we can modify the code when we do
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 43 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw52Ebb6B0eZm3X5_4P:-Nw8rh7IEe9mN_fxk8Y7:bmlqob4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L43)):*

When running `testDeleteCustomList` its failing here for me because the select location table view is in the view hierarchy in the background and has not been updated. See screenshot where I created custom lists "1" and "2" then removed "1". I think the assert should be more specific and look for custom list cell title with this text. 

</blockquote></details>

This was discussed offline, the reason was because there already was a custom list present, which meant the navigation was not what the framework expected.
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 64 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4nnxeDB8FBneBT7EY:-Nw8vJ73EllRIFikBNBH:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L64)):*

The function does one user action but the name is ambiguous about specifically how editing custom lists is finished. Suggest for example `tapDoneButton()`
</blockquote></details>

Done
___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 86 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw52dupDHx5zvZAgGAp:-Nw8tCNqEztuAXvbyaav:bovx3kr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L86)):*

Should this be addressed in the app? It might not only affect tests but also users relying on accessibility features.
</blockquote></details>

Currently accessibility is broken in LocationSelection, let's discuss this with the team.

mojganii (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-:-NwA0O4L7iXi0KQSgvrq:bnfp4nl)* status: 0 of 18 files reviewed, 16 unresolved discussions (waiting on @buggmagnet and @niklasberglund)

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)* status: 0 of 18 files reviewed, 16 unresolved discussions (waiting on @niklasberglund)

___
*[`ios/MullvadVPNUITests/Pages/Page.swift` line 51 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4rc-V8hnTJjhO020N:-NwE7if_5BZcVHOkEDPV:b-ndqj68) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/Pages/Page.swift#L51)):*

Will this tap be in the upper left corner? Wondering if it can cause issues in some scenarios for example with dynamic island or when iOS is showing that you're sharing location data, are in a call or similar. Then it's a tappable button in the upper left corner.

Would this tap in the middle of the status bar? 
`app.coordinate(withNormalizedOffset: CGVector(dx: app.frame.width/2, dy: 10)).tap()`

I think status bar always(or in almost all cases?) are 20 points high
</blockquote></details>

I hear what you're saying, but why would a phone used for automation testing be doing that ?

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)* status: 0 of 18 files reviewed, 16 unresolved discussions (waiting on @buggmagnet and @niklasberglund)

___
*[`ios/MullvadVPNUITests/Pages/Page.swift` line 51 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4rc-V8hnTJjhO020N:-NwFo3pi0jgkYeE-Pt37:b-jq89ow) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/Pages/Page.swift#L51)):*

I hear what you're saying, but why would a phone used for automation testing be doing that ?
</blockquote></details>

Could we not trigger the same automation that is used when you use accesibility tools to swipe to the top?
Ultimately, I don't feel like this should be a blocker, but ensuring that these tests run as smooth as possible on dev phones and test phones alike should be a priority.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)* status: 0 of 18 files reviewed, 16 unresolved discussions (waiting on @buggmagnet and @niklasberglund)

___
*[`ios/MullvadVPNUITests/Pages/Page.swift` line 51 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4rc-V8hnTJjhO020N:-NwFoOy54TfMJg7nxCak:bk082zk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/Pages/Page.swift#L51)):*

Could we not trigger the same automation that is used when you use accesibility tools to swipe to the top?
Ultimately, I don't feel like this should be a blocker, but ensuring that these tests run as smooth as possible on dev phones and test phones alike should be a priority.
</blockquote></details>

We could also go with Niklas' suggestion and be done with it -if it works.

niklasberglund (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148)* status: 0 of 18 files reviewed, 16 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadVPNUITests/CustomListsTests.swift` line 38 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4lADRASPX17jV50Ow:-NwFoeqW4rEy7GJZxXEC:bdy6qoy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/CustomListsTests.swift#L38)):*

I hear what you're saying, but we're using this function in all the tests so far, and it's less error prone to have one function that does it than repeat all the steps in all the tests.

We are not supporting swipe to delete yet, so we can modify the code when we do
</blockquote></details>

Agree, I was just concerned it might not be easy to get an overview but it is easy to get an overview since it's just a click-through away and then all the user actions are there 👍
___
*[`ios/MullvadVPNUITests/Pages/Page.swift` line 51 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-Nw4rc-V8hnTJjhO020N:-NwEkhrt2QX4CwoCGQQw:b-x0m9sa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/71b196abdb4f0f05c0b6fe962d0f6affc238361f/ios/MullvadVPNUITests/Pages/Page.swift#L51)):*

We could also go with Niklas' suggestion and be done with it -if it works.
</blockquote></details>

With the phone dedicated to testing it shouldn't be an issue but could be an issue with developers' phones. I don't think this is very important and wanna downgrade this to a nit 😊

niklasberglund (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6148#-:-NwJMhtY1hhaG2kBEZVs:bnfp4nl)* status: 0 of 18 files reviewed, 16 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

PR #6151, opened 2024-04-18
Ensure log file sizes stay below a certain threshold

rablador:
We should limit log files to not be larger than a certain threshold (currently 16kB) when writing to disk. If limit is exceeded, create a new file with a "_<partial count>" suffix and keep writing to it instead. Repeat if necessary.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151)

### DIFF
--- a/ios/MullvadLogging/CustomFormatLogHandler.swift
+++ b/ios/MullvadLogging/CustomFormatLogHandler.swift
@@ -16,14 +16,6 @@ public struct CustomFormatLogHandler: LogHandler {
     private let label: String
     private let streams: [TextOutputStream]
 
-    private let dateFormatter = Self.makeDateFormatter()
-
-    public static func makeDateFormatter() -> DateFormatter {
-        let dateFormatter = DateFormatter()
-        dateFormatter.dateFormat = "YYYY-MM-dd HH:mm:ss.SSS"
-        return dateFormatter
-    }
-
     public init(label: String, streams: [TextOutputStream]) {
         self.label = label
         self.streams = streams
@@ -54,7 +46,7 @@ public struct CustomFormatLogHandler: LogHandler {
             }
         let prettyMetadata = Self.formatMetadata(mergedMetadata)
         let metadataOutput = prettyMetadata.isEmpty ? "" : " \(prettyMetadata)"
-        let timestamp = dateFormatter.string(from: Date())
+        let timestamp = Date().logFormatted
         let formattedMessage = "[\(timestamp)][\(label)][\(level)]\(metadataOutput) \(message)\n"
 
         for var stream in streams {
--- a/ios/MullvadLogging/Date+LogFormat.swift
+++ b/ios/MullvadLogging/Date+LogFormat.swift
@@ -9,16 +9,22 @@
 import Foundation
 
 extension Date {
-    public func logFormatDate() -> String {
+    public var logFormatted: String {
         let formatter = DateFormatter()
+
         formatter.dateFormat = "dd/MM/yyyy @ HH:mm:ss"
+        formatter.locale = Locale(identifier: "en_US_POSIX")
+        formatter.timeZone = TimeZone(abbreviation: "UTC")
 
         return formatter.string(from: self)
     }
 
-    public func logFormatFilename() -> String {
+    public var logFileFormatted: String {
         let formatter = DateFormatter()
+
         formatter.dateFormat = "dd-MM-yyyy'T'HH:mm:ss"
+        formatter.locale = Locale(identifier: "en_US_POSIX")
+        formatter.timeZone = TimeZone(abbreviation: "UTC")
 
         return formatter.string(from: self)
     }
--- a/ios/MullvadLogging/LogFileOutputStream.swift
+++ b/ios/MullvadLogging/LogFileOutputStream.swift
@@ -16,10 +16,14 @@ private let reopenFileLogInterval: Duration = .seconds(5)
 class LogFileOutputStream: TextOutputStream {
     private let queue = DispatchQueue(label: "LogFileOutputStreamQueue", qos: .utility)
 
-    private let fileURL: URL
+    private let baseFileURL: URL
+    private var fileURL: URL
     private let encoding: String.Encoding
-    private let maxBufferCapacity: Int
+    private let maximumBufferCapacity: Int
     private let fileHeader: String
+    private let fileSizeLimit: UInt64
+    private var partialFileSizeCounter: UInt64 = 0
+    private var partialFileNameCounter = 1
 
     private var state: State = .closed {
         didSet {
@@ -45,11 +49,20 @@ class LogFileOutputStream: TextOutputStream {
         case waitingToReopen
     }
 
-    init(fileURL: URL, header: String, encoding: String.Encoding = .utf8, maxBufferCapacity: Int = 16 * 1024) {
+    init(
+        fileURL: URL,
+        header: String,
+        fileSizeLimit: UInt64 = ApplicationConfiguration.logMaximumFileSize,
+        encoding: String.Encoding = .utf8,
+        maxBufferCapacity: Int = 16 * 1024
+    ) {
         self.fileURL = fileURL
-        self.encoding = encoding
-        self.maxBufferCapacity = maxBufferCapacity
         self.fileHeader = header
+        self.fileSizeLimit = fileSizeLimit
+        self.encoding = encoding
+        self.maximumBufferCapacity = maxBufferCapacity
+
+        baseFileURL = fileURL.deletingPathExtension()
     }
 
     deinit {
@@ -58,11 +71,11 @@ class LogFileOutputStream: TextOutputStream {
 
     func write(_ string: String) {
         queue.async {
-            self.writeNoQueue(string)
+            self.writeOnQueue(string)
         }
     }
 
-    private func writeNoQueue(_ string: String) {
+    private func writeOnQueue(_ string: String) {
         guard let data = string.data(using: encoding) else { return }
 
         switch state {
@@ -89,7 +102,26 @@ class LogFileOutputStream: TextOutputStream {
         }
     }
 
-    @discardableResult private func write(fileHandle: FileHandle, data: Data) throws -> Int {
+    private func write(fileHandle: FileHandle, data: Data) throws {
+        dispatchPrecondition(condition: .onQueue(queue))
+
+        let incomingDataSize = UInt64(data.count)
+
+        // Make sure incoming data chunks are not larger than the file size limit.
+        // Failure to handle this leads to data neither being written nor buffered/trimmed.
+        guard incomingDataSize <= fileSizeLimit else {
+            throw POSIXError(.EDQUOT)
+        }
+
+        let predictedFileSize = partialFileSizeCounter + incomingDataSize
+
+        // Rotate file if threshold has been met, then rerun the write operation.
+        guard predictedFileSize <= fileSizeLimit else {
+            try rotateFile(handle: fileHandle)
+            write(String(data: data, encoding: encoding) ?? "")
+            return
+        }
+
         let bytesWritten = data.withUnsafeBytes { buffer -> Int in
             guard let ptr = buffer.baseAddress else { return 0 }
 
@@ -99,9 +131,17 @@ class LogFileOutputStream: TextOutputStream {
         if bytesWritten == -1 {
             let code = POSIXErrorCode(rawValue: errno)!
             throw POSIXError(code)
-        } else {
-            return bytesWritten
         }
+
+        partialFileSizeCounter += UInt64(bytesWritten)
+    }
+
+    private func rotateFile(handle: FileHandle) throws {
+        try handle.close()
+
+        state = .closed
+        partialFileSizeCounter = 0
+        fileURL = try incrementFileName()
     }
 
     private func openFile() throws -> FileHandle {
@@ -171,8 +211,18 @@ class LogFileOutputStream: TextOutputStream {
     private func bufferData(_ data: Data) {
         buffer.append(data)
 
-        if buffer.count > maxBufferCapacity {
-            buffer.removeFirst(buffer.count - maxBufferCapacity)
+        if buffer.count > maximumBufferCapacity {
+            buffer.removeFirst(buffer.count - maximumBufferCapacity)
+        }
+    }
+
+    private func incrementFileName() throws -> URL {
+        partialFileNameCounter += 1
+
+        if let url = URL(string: baseFileURL.relativePath + "_\(partialFileNameCounter).log") {
+            return url
+        } else {
+            throw POSIXError(.ENOENT)
         }
     }
 }
--- a/ios/MullvadLogging/LogRotation.swift
+++ b/ios/MullvadLogging/LogRotation.swift
@@ -36,7 +36,7 @@ public enum LogRotation {
         public var errorDescription: String? {
             switch self {
             case .rotateLogFiles:
-                return "Failure to rotate the source log file to backup."
+                return "Failure to rotate logs"
             }
         }
 
@@ -74,30 +74,23 @@ public enum LogRotation {
                 log1.creationDate > log2.creationDate
             }
 
-            try deleteLogsOlderThan(options.oldestAllowedDate, in: logs)
-            try deleteLogsWithCombinedSizeLargerThan(options.storageSizeLimit, in: logs)
+            try deleteLogs(dateThreshold: options.oldestAllowedDate, sizeThreshold: options.storageSizeLimit, in: logs)
         } catch {
             throw Error.rotateLogFiles(error)
         }
     }
 
-    private static func deleteLogsOlderThan(_ dateThreshold: Date, in logs: [LogData]) throws {
+    private static func deleteLogs(dateThreshold: Date, sizeThreshold: Int, in logs: [LogData]) throws {
         let fileManager = FileManager.default
 
-        for log in logs where log.creationDate < dateThreshold {
-            try fileManager.removeItem(at: log.path)
-        }
-    }
-
-    private static func deleteLogsWithCombinedSizeLargerThan(_ sizeThreshold: Int, in logs: [LogData]) throws {
-        let fileManager = FileManager.default
-
-        // Delete all logs outside maximum capacity (ordered newest to oldest).
         var fileSizes = UInt64.zero
         for log in logs {
             fileSizes += log.size
 
-            if fileSizes > sizeThreshold {
+            let logIsTooOld = log.creationDate < dateThreshold
+            let logCapacityIsExceeded = fileSizes > sizeThreshold
+
+            if logIsTooOld || logCapacityIsExceeded {
                 try fileManager.removeItem(at: log.path)
             }
         }
--- a/ios/MullvadLogging/Logging.swift
+++ b/ios/MullvadLogging/Logging.swift
@@ -78,7 +78,7 @@ public struct LoggerBuilder {
             let rotationLogger = Logger(label: "LogRotation")
 
             for error in logRotationErrors {
-                rotationLogger.error(error: error, message: "Failed to rotate log")
+                rotationLogger.error(error: error, message: error.localizedDescription)
             }
         }
     }
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -603,6 +603,8 @@
 		7ADCB2D82B6A6EB300C88F89 /* AnyRelay.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7ADCB2D72B6A6EB300C88F89 /* AnyRelay.swift */; };
 		7ADCB2DA2B6A730400C88F89 /* IPOverrideRepositoryStub.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7ADCB2D92B6A730400C88F89 /* IPOverrideRepositoryStub.swift */; };
 		7AE044BB2A935726003915D8 /* Routing.h in Headers */ = {isa = PBXBuildFile; fileRef = 7A88DCD02A8FABBE00D2FF0E /* Routing.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		7AED35CC2BD13F60002A67D1 /* ApplicationConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58BFA5CB22A7CE1F00A6173D /* ApplicationConfiguration.swift */; };
+		7AED35CD2BD13FC4002A67D1 /* ApplicationTarget.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58C76A072A33850E00100D75 /* ApplicationTarget.swift */; };
 		7AEF7F1A2AD00F52006FE45D /* AppMessageHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AEF7F192AD00F52006FE45D /* AppMessageHandler.swift */; };
 		7AF10EB22ADE859200C090B9 /* AlertViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AF10EB12ADE859200C090B9 /* AlertViewController.swift */; };
 		7AF10EB42ADE85BC00C090B9 /* RelayFilterCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7AF10EB32ADE85BC00C090B9 /* RelayFilterCoordinator.swift */; };
@@ -5732,6 +5734,8 @@
 			isa = PBXSourcesBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				7AED35CC2BD13F60002A67D1 /* ApplicationConfiguration.swift in Sources */,
+				7AED35CD2BD13FC4002A67D1 /* ApplicationTarget.swift in Sources */,
 				58D22402294C90050029F5F8 /* Logger+Errors.swift in Sources */,
 				58D22404294C90050029F5F8 /* Date+LogFormat.swift in Sources */,
 				58D22403294C90050029F5F8 /* Logging.swift in Sources */,
--- a/ios/MullvadVPN/AddressCacheTracker/AddressCacheTracker.swift
+++ b/ios/MullvadVPN/AddressCacheTracker/AddressCacheTracker.swift
@@ -65,7 +65,7 @@ final class AddressCacheTracker {
 
         let scheduleDate = _nextScheduleDate()
 
-        logger.debug("Schedule address cache update at \(scheduleDate.logFormatDate()).")
+        logger.debug("Schedule address cache update at \(scheduleDate.logFormatted).")
 
         scheduleEndpointsUpdate(startTime: .now() + scheduleDate.timeIntervalSinceNow)
     }
@@ -165,7 +165,7 @@ final class AddressCacheTracker {
             let scheduleDate = self._nextScheduleDate()
 
             self.logger
-                .debug("Schedule next address cache update at \(scheduleDate.logFormatDate()).")
+                .debug("Schedule next address cache update at \(scheduleDate.logFormatted).")
 
             self.scheduleEndpointsUpdate(startTime: .now() + scheduleDate.timeIntervalSinceNow)
         }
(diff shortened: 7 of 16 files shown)


rablador (commented):
Will look at creating tests shortly.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-:-NvpJt8bBAlHhCzkG0cP:b35v7li)* status: 0 of 13 files reviewed, all discussions resolved

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151)* status: 0 of 13 files reviewed, 3 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadLogging/Date+LogFormat.swift` line 13 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-NvpEhiFD2Yjfo36Hw9u:-NvpEhiFD2Yjfo36Hw9v:bdpkc6t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/890163db3b5b1accfcdc6b6174ea848964aaedd9/ios/MullvadLogging/Date%2BLogFormat.swift#L13)):*
> ```Swift
> extension Date {
>     public var logFormatted: String {
>         logFileFormatted.replacingOccurrences(of: "T", with: " ")
> ```

we are doing some string manipulation in `Date` extension that has nothing to do with that.Moreover,  the using formatter is much more sensible to me instead of replacing.



_Code snippet:_
```Swift
let formatter = DateFormatter()
formatter.dateFormat = "dd/MM/yyyy HH:mm:ss"
return formatter.string(from: self)
```
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 101 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-NvpL86FBwqA0i4gYlmS:-NvpL86FBwqA0i4gYlmT:b40ohl5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/890163db3b5b1accfcdc6b6174ea848964aaedd9/ios/MullvadLogging/LogFileOutputStream.swift#L101)):*
> ```Swift
> 
>         guard partialFileSizeCounter <= ApplicationConfiguration.logMaximumFileSize else {
>             try fileHandle.close()
> ```

nit : for having more readability I suggest to put this scope of the file in another function which is called `writeInNewFile` or something like that
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 196 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-NvpLneC60uHILS8zemY:-NvpLneC60uHILS8zemZ:b-mm8j68) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/890163db3b5b1accfcdc6b6174ea848964aaedd9/ios/MullvadLogging/LogFileOutputStream.swift#L196)):*
> ```Swift
>     }
> 
>     private func incrementFileName(partialFileCounter: Int) throws -> URL {
> ```

we have the `partialFileCounter` as a global value, there is no need to pass it to function. the name of functions doesn't make sense the return value is a file url for new file.



_Code snippet:_
```Swift
    private func newFileUrl() throws -> URL {
        if let url = URL(string: baseFileURL.relativePath + "_\(self.partialFileNameCounter).log") {
            return url
        } else {
            throw POSIXError(POSIXErrorCode(rawValue: 2)!) // "No such file or directory"
        }
    }
```

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadLogging/Date+LogFormat.swift` line 13 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-NvpEhiFD2Yjfo36Hw9u:-NvqelGs5LpLmHO066N-:b4hco86) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/890163db3b5b1accfcdc6b6174ea848964aaedd9/ios/MullvadLogging/Date%2BLogFormat.swift#L13)):*

we are doing some string manipulation in `Date` extension that has nothing to do with that.Moreover,  the using formatter is much more sensible to me instead of replacing.
</blockquote></details>

Alright, I'll revert the change and apply locale and timezone instead.
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 101 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-NvpL86FBwqA0i4gYlmS:-NvqeY4s5iORZbPFfEX3:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/890163db3b5b1accfcdc6b6174ea848964aaedd9/ios/MullvadLogging/LogFileOutputStream.swift#L101)):*

nit : for having more readability I suggest to put this scope of the file in another function which is called `writeInNewFile` or something like that
</blockquote></details>

Done.
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 196 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-NvpLneC60uHILS8zemY:-NvqdlDc6MRWmOfEku3l:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/890163db3b5b1accfcdc6b6174ea848964aaedd9/ios/MullvadLogging/LogFileOutputStream.swift#L196)):*

we have the `partialFileCounter` as a global value, there is no need to pass it to function. the name of functions doesn't make sense the return value is a file url for new file.
</blockquote></details>

Done.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151)* status: 0 of 16 files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 109 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3jCq_7xH-b2AzBbi2:-Nw3jCq_7xH-b2AzBbi3:b-hgsg3d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L109)):*
> ```Swift
> 
>         // Rotate file if threshold has been met, then rerun the write operation.
>         guard partialFileSizeCounter <= fileSizeLimit else {
> ```

There is a caveat here. If the incoming data chunks themselves are larger than `fileSizeLimit`, no logs will be written at all. AND many many new log files will be created.

buggmagnet (changes requested):
> Reviewed 11 of 13 files at r1, 4 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151)* status: 15 of 16 files reviewed, 11 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 106 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3mSjgCyCbbSLXPdgy:-Nw3mSjgCyCbbSLXPdgz:b-qeg67n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L106)):*
> ```Swift
> 
>     private func write(fileHandle: FileHandle, data: Data) throws {
>         partialFileSizeCounter += UInt64(data.count)
> ```

This isn't correct.
We should only increase the `partialFileSizeCounter` with the number of bytes written (returned by `Darwin.write`)

Otherwise, we'll have a mismatch when we fail to write below.
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 118 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3iX1O3VQBoY_djLRN:-Nw3iX1O3VQBoY_djLRO:b5v7u3p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L118)):*
> ```Swift
>             guard let ptr = buffer.baseAddress else { return 0 }
> 
>             return Darwin.write(fileHandle.fileDescriptor, ptr, buffer.count)
> ```

nit
The small downside of this is that we can't easily test scenarios where we have to `bufferData` 
Fortunately, the data we buffer is way lower than a maximum file size, so we shouldn't have any scenario where we end up
in a loop where we can't write to a new file because there is too much buffered data. 
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 127 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3kqpT22ZWTGci0dt9:-Nw3kqpT22ZWTGci0dtA:b-rnr0jx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L127)):*
> ```Swift
>     }
> 
>     fileprivate func rotateFile(handle: FileHandle, data: Data) throws {
> ```

That `data` parameter is not used
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 213 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3WEfkE5Ue3CSqFrj4:-Nw3WEfkE5Ue3CSqFrj5:b-qb4sw4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L213)):*
> ```Swift
>             return url
>         } else {
>             throw POSIXError(POSIXErrorCode(rawValue: 2)!) // "No such file or directory"
> ```

`throw POSIXError(.ENOENT)`
___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 6504 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3cOpiA3My5mujmLYs:-Nw3cOpiA3My5mujmLYt:b-qy4f1f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadVPN.xcodeproj/project.pbxproj#L6504)):*
> ```
> 				PRODUCT_NAME = "$(TARGET_NAME)";
> 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
> 				SWIFT_STRICT_CONCURRENCY = minimal;
> ```

Did you mean to check this in ?
___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 37 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3czQA1lQ8kJUn31Wt:-Nw3czQA1lQ8kJUn31Wu:b-gxdnx0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadVPNTests/LoggingTests.swift#L37)):*
> ```Swift
>         sync()
> 
>         let contents = String(decoding: (try? Data(contentsOf: fileURL)) ?? Data(), as: UTF8.self)
> ```

`let contents = try XCTUnwrap(String(contentsOf: fileURL))` is much easier to read (we also need to mark the test as `throws` 
Same for the test below 
___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 35 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3eYmJ3c3NRAhK2y7d:-Nw3eYmJ3c3NRAhK2y7e:bnogvz2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadVPNTests/LogRotationTests.swift#L35)):*
> ```Swift
>         let logChunkSize = 20
> 
>         let exptectedLogCount = Int(ceil(Double(totalLogTestSize) / Double(totalLogSizeLimit)))
> ```

`exptectedLogCount` -> `expectedLogCount`
___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 138 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3himVCLxhkyCmYL7g:-Nw3himVCLxhkyCmYL7h:bvcfy50) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadVPNTests/LogRotationTests.swift#L138)):*
> ```Swift
> extension LogRotationTests {
>     private func stringOfSize(_ size: Int) -> String {
>         String(repeating: " ", count: size)
> ```

nit
if you look at the files generated during the tests, it just looks like empty space, that's quite fun !
But it's not a big deal

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151)* status: 15 of 16 files reviewed, 11 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 106 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3mSjgCyCbbSLXPdgy:-Nw46aHf-irmTgTAUFBO:blzmu3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L106)):*

This isn't correct.
We should only increase the `partialFileSizeCounter` with the number of bytes written (returned by `Darwin.write`)

Otherwise, we'll have a mismatch when we fail to write below.
</blockquote></details>

You're right, I moved the counter to the bottom. We should, however, still check for size limitation before attempting to write, so introduced a temporary prediction for that instead.
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 118 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3iX1O3VQBoY_djLRN:-Nw46eb77AiPSBJly6gj:b-36nf6l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L118)):*

nit
The small downside of this is that we can't easily test scenarios where we have to `bufferData` 
Fortunately, the data we buffer is way lower than a maximum file size, so we shouldn't have any scenario where we end up
in a loop where we can't write to a new file because there is too much buffered data.
</blockquote></details>

Yes, in realty this should not be a problem, but it still feels iffy that we could end up with in a bad state if not configured properly.
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 127 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3kqpT22ZWTGci0dt9:-Nw49C5G2r8LR-c_Stga:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L127)):*

That `data` parameter is not used
</blockquote></details>

Done.
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 213 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3WEfkE5Ue3CSqFrj4:-Nw49HOi9_3S3_frZtxJ:b8nksgm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L213)):*

`throw POSIXError(.ENOENT)`
</blockquote></details>

That's what I wanted, yes :)
___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 6504 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3cOpiA3My5mujmLYs:-Nw49MnCCa397UgkyX7D:br6s1v7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadVPN.xcodeproj/project.pbxproj#L6504)):*

Did you mean to check this in ?
</blockquote></details>

Nope...
___
*[`ios/MullvadVPNTests/LoggingTests.swift` line 37 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3czQA1lQ8kJUn31Wt:-Nw4A02e3RH6XDmBihHq:b3sb5nz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadVPNTests/LoggingTests.swift#L37)):*

`let contents = try XCTUnwrap(String(contentsOf: fileURL))` is much easier to read (we also need to mark the test as `throws` 
Same for the test below
</blockquote></details>

Didn't want to intrude too much on the existing code, just fix the lint error. But let's change it now that it's been pointed out.
___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 35 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3eYmJ3c3NRAhK2y7d:-Nw4AUTZ8LiEVaRcf-JK:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadVPNTests/LogRotationTests.swift#L35)):*

`exptectedLogCount` -> `expectedLogCount`
</blockquote></details>

Done.
___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 138 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3himVCLxhkyCmYL7g:-Nw4AWNBCtI1e-8-rSue:b-ycwiiv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadVPNTests/LogRotationTests.swift#L138)):*

nit
if you look at the files generated during the tests, it just looks like empty space, that's quite fun !
But it's not a big deal
</blockquote></details>

Felt like the most secure way to generate one byte per character. But I guess I could go with an "a" or something.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151)* status: 12 of 16 files reviewed, 10 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 109 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3jCq_7xH-b2AzBbi2:-Nw3lKPM85mnmdnaSIUT:b-n97lsq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L109)):*

There is a caveat here. If the incoming data chunks themselves are larger than `fileSizeLimit`, no logs will be written at all. AND many many new log files will be created.
</blockquote></details>

Can't we truncate the incoming data into different smaller chunks and write them in different files throughs a loop if the data size is bigger the limit?



_Code snippet:_
```Swift
    fileprivate func rotateFile(handle: FileHandle, data: Data) throws {
        var limit = fileSizeLimit
        var offset: UInt64 = 0

        let size = UInt64(data.count)

        while offset < data.count {
            try handle.close()
            state = .closed
            fileURL = try incrementFileName()
            limit = offset + limit < size ? limit : (size - offset)
            write(String(data: data[Data.Index(offset) ..< Data.Index(offset + limit)], encoding: encoding) ?? "")
            offset += limit
        }

        partialFileSizeCounter = 0
    }
```

acb-mv (approved):
> Reviewed 1 of 5 files at r2, 2 of 4 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151)* status: all files reviewed, 10 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 138 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3himVCLxhkyCmYL7g:-NwJRFf90ZpJL5Y_yEz-:b-wer2aq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadVPNTests/LogRotationTests.swift#L138)):*

Felt like the most secure way to generate one byte per character. But I guess I could go with an "a" or something.
</blockquote></details>

If you want a sequence that's less featureless, you could do something like 

```
(0..<size).map { "\($0%10)" }.joined(separator: '')
```
which gives you `0123456789012...`, or something similar.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151)* status: all files reviewed, 10 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @mojganii)

___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 109 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3jCq_7xH-b2AzBbi2:-NwK49ya3upzCL9URRBs:b-p5lxro) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L109)):*

Can't we truncate the incoming data into different smaller chunks and write them in different files throughs a loop if the data size is bigger the limit?
</blockquote></details>

`write()` calls to an async queue, so the while loop here gets interweaved with calls from the re-queued items as well as the current item being processed. Perhaps we can do something of similar logic though. What do @acb-mv or @buggmagnet think? Worth the extra effort?
___
*[`ios/MullvadVPNTests/LogRotationTests.swift` line 138 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3himVCLxhkyCmYL7g:-NwJsb4X4WfZo-_hHV4S:b-rqhyyw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadVPNTests/LogRotationTests.swift#L138)):*

If you want a sequence that's less featureless, you could do something like 

```
(0..<size).map { "\($0%10)" }.joined(separator: '')
```
which gives you `0123456789012...`, or something similar.
</blockquote></details>

I like it 👍

buggmagnet (approved):
> Reviewed 1 of 4 files at r3, 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151)* status: all files reviewed, 5 unresolved discussions (waiting on @acb-mv, @mojganii, and @rablador)

___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 109 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3jCq_7xH-b2AzBbi2:-Nwcfgyy4p_6yEEyDl4I:bgko8o7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L109)):*
> If the incoming data chunks themselves are larger than `fileSizeLimit`, no logs will be written at all. AND many many new log files will be created.

You are technically correct, but it's not a realistic problem to solve IMHO, 128k data to write is a truckload of data, especially given that we control most of the data written to the log files.

I would say in that case, throw an error and drop all the data. We shouldn't make this more complicated than it already is.
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 78 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nwcdd-T-yacH_eg94fv:-Nwcdd-T-yacH_eg94fw:b6tvt98) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2091c4353ecd16076f3a74a86565a95c4fcc3a42/ios/MullvadLogging/LogFileOutputStream.swift#L78)):*
> ```Swift
>     }
> 
>     private func writeNoQueue(_ string: String) {
> ```

nit
Everytime I read this method name, I'm confused because it's only ever called from the dispatch queue we have here. 
Can we rename this `writeOnQueue` instead ?

Also we should sprinkle some `dispatchPrecondition(condition: .onQueue(queue))` in every function that is called on the queue to help get more contextual information as to what happens where IMHO
___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 6539 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-NwcgUvh7wkJOcuYVmM4:-NwcgUvh7wkJOcuYVmM5:bsdq4t0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2091c4353ecd16076f3a74a86565a95c4fcc3a42/ios/MullvadVPN.xcodeproj/project.pbxproj#L6539)):*
> ```
> 				PRODUCT_BUNDLE_IDENTIFIER = "$(APPLICATION_IDENTIFIER)";
> 				PRODUCT_NAME = "$(TARGET_NAME)";
> 				SUPPORTED_PLATFORMS = "iphoneos iphonesimulator";
> ```

What about these changes, did you mean to check them in ?

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 109 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nw3jCq_7xH-b2AzBbi2:-Nwcxk-e87MPSYKf__O_:bqrl0yv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/26fbccb8ed957621783d0c80ca245a3b16a50a0f/ios/MullvadLogging/LogFileOutputStream.swift#L109)):*

> If the incoming data chunks themselves are larger than `fileSizeLimit`, no logs will be written at all. AND many many new log files will be created.

You are technically correct, but it's not a realistic problem to solve IMHO, 128k data to write is a truckload of data, especially given that we control most of the data written to the log files.

I would say in that case, throw an error and drop all the data. We shouldn't make this more complicated than it already is.
</blockquote></details>

I think you're right. If we throw an error, the data will be buffered (and trimmed), thus solving the immediate problem of chunk size being too large. It doesn't really fix anything, but it handles deviations gracefully, which is good enough I think.
___
*[`ios/MullvadLogging/LogFileOutputStream.swift` line 78 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-Nwcdd-T-yacH_eg94fv:-Nwcu_lc2aR5EwKAx0pS:b-ne7w75) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2091c4353ecd16076f3a74a86565a95c4fcc3a42/ios/MullvadLogging/LogFileOutputStream.swift#L78)):*

nit
Everytime I read this method name, I'm confused because it's only ever called from the dispatch queue we have here. 
Can we rename this `writeOnQueue` instead ?

Also we should sprinkle some `dispatchPrecondition(condition: .onQueue(queue))` in every function that is called on the queue to help get more contextual information as to what happens where IMHO
</blockquote></details>

Added a `dispatchPrecondition` to `write(fileHandle: FileHandle, data: Data)` since that's were the actual writing happens. We can end up there either directly from a function call or indirectly from a timer call.
___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 6539 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-NwcgUvh7wkJOcuYVmM4:-NwckaC18OaDzfMojkAh:b-wngnuj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2091c4353ecd16076f3a74a86565a95c4fcc3a42/ios/MullvadVPN.xcodeproj/project.pbxproj#L6539)):*

What about these changes, did you mean to check them in ?
</blockquote></details>

The changes you see are probably between revisions where I either messed up or restored the mess. If you compare first and last revision there will be no changes here.

buggmagnet (approved):
:lgtm:

> Reviewed 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6151#-:-NxCErAaEyaoimskBxNk:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @mojganii)

PR #6167, opened 2024-04-23
Upgrade rustls to avoid RUSTSEC-2024-0336

faern:
https://github.com/mullvad/mullvadvpn-app/actions/runs/8763182344

Reported CVE against rustls. We are not vulnerable since we only use rustls via tokio-rustls which do not call complete_io, but we should still always upgrade instead of ignore reported issues.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6167)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1897,7 +1897,7 @@ dependencies = [
  "libflate",
  "rand 0.8.5",
  "rand_distr",
- "ring",
+ "ring 0.16.20",
  "serde",
  "simple-error",
 ]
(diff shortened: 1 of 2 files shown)


dlon (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6167#-:-Nw8zp_m2RdtAH8ajVSe:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6175, opened 2024-04-24
Upgrade go dependencies used by wireguard-go (and remove unused golang monotonic time patch)

faern:
Upgrade golang.org/x/net, golang.org/x/crypto and golang.org/x/sys. Fixes issues GO-2023-2402, GO-2022-0969, GO-2023-1495, GO-2023-1571, GO-2023-1988, GO-2023-2102 and GO-2024-2687.

None of these issues affected our app (they were all about HTTP/2 or other parts of golang that wireguard-go don't use). But we have not upgrade in a long time, so it's time. It also helps silence vulnerability scanners. It's either upgrade or add to config files as ignored. Better to upgrade if there are no blockers in doing so.

This PR also does something sort of unrelated: Remove the completely unused `goruntime-boottime-over-monotonic.diff` file. This file is only used on Android, and they have their own version of the patch vendored down under `android/`.

Fixes DES-721

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6175)


faern:
I have tried this on Linux (with `TALPID_FORCE_USERSPACE_WIREGUARD=1`) and macOS. Seems to work. Only did smoke tests where I verified I could establish a wg tunnel and that traffic passed through it.

albin-mullvad (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6175#-:-NwEzQXV96z_brHdwsY6:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6177, opened 2024-04-24
Run `cargo update`, fix base64 + chrono

hulthe:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6177)

PR #6178, opened 2024-04-24
Upgrade winapi related deps

faern:
I just saw that `winapi-util` just released a version finally migrating from `winapi` to `windows-sys` (https://github.com/BurntSushi/winapi-util/pull/13#issuecomment-2072513938). So I created this branch to bump the dependency. Not that it gets rid of `winapi` from our dependency tree, but one step in that direction.

I then looked at `cargo tree --all-targets` and identified another place where we could actually shrink our dependency tree a tiny bit. Bumping `hyper` removes the last dependency on `socket2 0.4` which gets rid of that completely, and in turn removes another link to `winapi` in the meantime.

So this is just a small dependency tree maintenance PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6178)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1442,9 +1442,9 @@ checksum = "9a3a5bfb195931eeb336b2a7b4d761daec841b97f947d34394601737a7bba5e4"
 
 [[package]]
 name = "hyper"
-version = "0.14.27"
+version = "0.14.28"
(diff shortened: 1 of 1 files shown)


faern:
There is some build error here that I have to look into...

faern:
This PR was made obsolete by @hulthe in #6177

PR #6193, opened 2024-04-29
Trim whitespace from list names on save

rablador:
List names should have their trailing and leading whitespace removed when they are saved. Trailing and leading whitespace should also be ignored when checking if a list already exists.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6193)


acb-mv (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6193)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6194, opened 2024-04-29
Refactor PacketTunnelActor's state logic to use a Reducer

acb-mv:
This is an experimental refactoring of the PacketTunnelActor state machine, which splits the state processing logic into two stages: a stateless function (a Reducer) which takes incoming events (`Command`s) and existing state and produces updated state and zero or more `Effect` values, and stateful code within the actor that executes the `Effect` values, producing effects.  This is an instance of the Reducer architecture, and has the effect of making the state transition logic purely stateless and allowing it to be tested and reasoned about without concern for side effects.

This branch is not on `main` but on `features/ios-post-quantum`, as to incorporate extensions to the state flow used in the PQ work.  Despite this, it is not part of the PQ work as such.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6194)


rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6194)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6196, opened 2024-04-29
Add apparmor profile on Ubuntu 24.04 or later

### LINKED ISSUE #6128: Impossible to launch Mullvad GUI | The SUID sandbox helper binary was found, but is not configured correctly.
It works using `--no-sandbox`.

dlon:
Disable restrictions for unprivileged user namespaces. See [this](https://discourse.ubuntu.com/t/noble-numbat-release-notes/39890) for details.

Fix #6128.
Fix DES-924.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6196)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ Line wrap the file at 100 chars.                                              Th
 ### Added
 - Add custom bridge settings in GUI.
 
+### Fixed
+#### Linux
+- Fix GUI not working on Ubuntu 24.04 by adding an AppArmor profile.
+
 
 ## [2024.2] - 2024-04-29
 ### Fixed
--- a/dist-assets/linux/after-install.sh
+++ b/dist-assets/linux/after-install.sh
@@ -6,3 +6,30 @@ chmod u+s "/usr/bin/mullvad-exclude"
 systemctl enable "/usr/lib/systemd/system/mullvad-daemon.service"
 systemctl start mullvad-daemon.service || echo "Failed to start mullvad-daemon.service"
 systemctl enable "/usr/lib/systemd/system/mullvad-early-boot-blocking.service"
+
+# return 0 if version $1 is greater than or equal to $2
+function version_is_ge {
+    [ "$1" = "$2" ] && return 0
+    printf '%s\n' "$2" "$1" | sort -C -V
+}
+
+# Ubuntu 24.04 or newer: Install apparmor profile to allow Electron sandbox to work
+# This disables user namespace restrictions
+os=$(grep -oP '^ID=\K.+' /etc/os-release | tr -d '"')
+version=$(grep -oP '^VERSION_ID=\K.+' /etc/os-release | tr -d '"')
+
+if [[ "$os" == "ubuntu" ]] && version_is_ge "$version" "24.04"; then
+    echo "Creating apparmor profile"
+
+    cat <<EOF > /etc/apparmor.d/mullvad
+abi <abi/4.0>,
+include <tunables/global>
+
+profile mullvad /opt/Mullvad\ VPN/mullvad-gui flags=(unconfined) {
+  userns,
+  # Site-specific additions and overrides. See local/README for details.
+  include if exists <local/mullvad>
+}
+EOF
+    apparmor_parser -r /etc/apparmor.d/mullvad || echo "Failed to reload apparmor profile"
+fi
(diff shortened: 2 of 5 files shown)


MarkusPettersson98 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6196)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`dist-assets/linux/after-install.sh` line 33 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6196#-NwhrgR8-UwXBMumeLUT:-NwhrgR8-UwXBMumeLUU:b18ru4n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78c9b38e937dbaa3390e69753b0f667a1bb049b6/dist-assets/linux/after-install.sh#L33)):*
> ```Shell
>   include if exists <local/mullvad>
> }
> EOF
> ```

Are you opposed to shipping this AppArmor profile in a separate file (in `dist-assets/linux/`) instead? We already do this with our systemd services, and I feel like it will be slightly easier to see what we actually bundle with the app :blush:



_Code quote:_
```Shell
    cat <<EOF > /etc/apparmor.d/mullvad
abi <abi/4.0>,
include <tunables/global>

profile mullvad /opt/Mullvad\ VPN/mullvad-gui flags=(unconfined) {
  userns,
  # Site-specific additions and overrides. See local/README for details.
  include if exists <local/mullvad>
}
EOF
```

MarkusPettersson98 (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6196)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`dist-assets/linux/after-install.sh` line 33 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6196#-NwhrgR8-UwXBMumeLUT:-Nwi24Z8B7oj3sZb7y-Y:bcv502w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/78c9b38e937dbaa3390e69753b0f667a1bb049b6/dist-assets/linux/after-install.sh#L33)):*

Are you opposed to shipping this AppArmor profile in a separate file (in `dist-assets/linux/`) instead? We already do this with our systemd services, and I feel like it will be slightly easier to see what we actually bundle with the app :blush:
</blockquote></details>

Nice :+1:

greengem70:
Installed 24.02 hoping fixed but still fails.     GUI does not launch from Applications

From terminal:
/opt/Mullvad\ VPN/mullvad-gui 

gives.
[314073:0506/195002.808568:FATAL:setuid_sandbox_host.cc(158)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /opt/Mullvad VPN/chrome-sandbox is owned by root and has mode 4755.

Workaround still functions.

dlon:
@greengem70 This was included in the 2024.3-beta1 release. Did you try that one?

gamer191:
If this script only runs when installing Mullvad on 24.04, doesn't that mean that Mullvad VPN will need to be re-installed when upgrading to 24.04?

MarkusPettersson98:
Yes, that is an unfortunate consequence.

gamer191:
@MarkusPettersson98 would it make more sense to  always copy the profile if the OS is Ubuntu?

MarkusPettersson98:
I don't think it makes too much sense _any more_. The number of people upgrading from an earlier version of Ubuntu to 24.04 is decreasing all the time, and deploying the suggested fix would already run our apparmor deployment since people would be upgrading their app version. If we would have been quicker to pick up this fix on our side, then I would be all for it.

Thanks for making us aware about this potential issue, such efforts are always appreciated :yellow_heart:

gamer191:
@MarkusPettersson98 According to https://discourse.ubuntu.com/t/ubuntu-24-04-lts-noble-numbat-release-notes/39890

>Users of 22.04 LTS however will be offered the automatic upgrade when 24.04.1 LTS is released, which is scheduled for the 15th of August.

As such, I suspect there will be a large increase in the number of users experiencing this issue in about 2 days (sorry, I just noticed that now)

MarkusPettersson98:
Thanks for making us aware of those release notes, we'll see what we can do to prepare for this event.

PR #6211, opened 2024-05-02
Don't rebuild daemon every time new git refs are fetched

Serock3:
Fixes DES-950

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)

### DIFF
--- a/mullvad-version/build.rs
+++ b/mullvad-version/build.rs
@@ -75,6 +75,40 @@ fn get_dev_suffix(target: Target, product_version: &str) -> Option<String> {
         Target::Desktop => product_version.to_owned(),
     };
 
+    let git_dir = Path::new("..").join(".git");
+
+    // If we build our output on information about HEAD we need to re-run if HEAD moves
+    let head_path = git_dir.join("HEAD");
+    if head_path.exists() {
+        println!("cargo:rerun-if-changed={}", head_path.display());
+    }
+
+    let output = Command::new("git")
+        .arg("branch")
+        .arg("--show-current")
+        .output()
+        .ok()?;
+    let current_branch = String::from_utf8(output.stdout).unwrap();
+    // If we build our output on information about a git reference, we need to re-run
+    // if it moves. Instead of trying to be smart, just re-run if any git reference moves.
+    let git_current_branch_ref = git_dir
+        .join("refs")
+        .join("heads")
+        .join(current_branch.trim());
+    if git_current_branch_ref.exists() {
+        println!(
+            "cargo:rerun-if-changed={}",
+            git_current_branch_ref.display()
+        );
+    }
+    let git_current_branch_ref = git_dir.join("refs").join("tags").join(&release_tag);
+    if git_current_branch_ref.exists() {
+        println!(
+            "cargo:rerun-if-changed={}",
+            git_current_branch_ref.display()
+        );
+    }
+
     // Get the git commit hashes for the latest release and current HEAD
     // Return `None` if unable to find the hash for HEAD.
     let head_commit_hash = git_rev_parse_commit_hash("HEAD")?;
@@ -94,21 +128,6 @@ fn get_dev_suffix(target: Target, product_version: &str) -> Option<String> {
 ///
 /// Returns `None` if executing the `git rev-parse` command fails for some reason.
 fn git_rev_parse_commit_hash(git_ref: &str) -> Option<String> {
-    let git_dir = Path::new("..").join(".git");
-    // If we build our output on information about HEAD we need to re-run if HEAD moves
-    if git_ref == "HEAD" {
-        let head_path = git_dir.join("HEAD");
-        if head_path.exists() {
-            println!("cargo:rerun-if-changed={}", head_path.display());
-        }
-    }
-    // If we build our output on information about a git reference, we need to re-run
-    // if it moves. Instead of trying to be smart, just re-run if any git reference moves.
-    let git_refs_dir = git_dir.join("refs");
-    if git_refs_dir.exists() {
-        println!("cargo:rerun-if-changed={}", git_refs_dir.display());
-    }
-
     let output = Command::new("git")
         .arg("rev-parse")
         .arg(format!("{git_ref}^{{commit}}"))
--- a/mullvad-version/build.rs
+++ b/mullvad-version/build.rs
@@ -101,14 +101,17 @@ fn get_dev_suffix(release_tag: &str) -> Option<String> {
 fn rerun_if_git_ref_changed(release_tag: &str) -> std::io::Result<()> {
     let git_dir = Path::new("..").join(".git");
 
-    // If we build our output on information about HEAD we need to re-run if HEAD moves
+    // The `.git/HEAD` file contains the position of the current head. If in 'detached HEAD' state,
+    // this will be the ref of the current commit. If on a branch it will just point to it, e.g.
+    // `ref: refs/heads/main`. Tracking changes to this file will tell us if we change branch, or
+    // modify the current detached HEAD state (e.g. committing or rebasing).
     let head_path = git_dir.join("HEAD");
     if head_path.exists() {
         println!("cargo:rerun-if-changed={}", head_path.display());
     }
 
-    // If we build our output on information about the ref of the current branch, we need to re-run
-    // on changes to it
+    // The above check will not cause a rebuild when modifying commits on a currently checked out
+    // branch. To catch this, we need to track the `.git/refs/heads/$current_branch` file.
     let output = Command::new("git")
         .arg("branch")
         .arg("--show-current")
@@ -135,7 +138,15 @@ fn rerun_if_git_ref_changed(release_tag: &str) -> std::io::Result<()> {
     if git_release_tag_ref.exists() {
         println!("cargo:rerun-if-changed={}", git_release_tag_ref.display());
     };
-    Some(())
+
+    // NOTE: As the repository has gotten quite large, you may find the contents of the
+    // `.git/refs/heads` and `.git/refs/tags` empty. This happens because `git pack-refs` compresses
+    // and moves the information into the `.git/packed-refs` file to save storage. We do not have to
+    // track this file, however, as any changes to the current branch, 'detached HEAD' state
+    // or tags will update the corresponding `.git/refs` file we are tracking, even if it had
+    // previously been pruned.
+
+    Ok(())
 }
 
 /// Returns the commit hash for the commit that `git_ref` is pointing to.
--- a/mullvad-version/build.rs
+++ b/mullvad-version/build.rs
@@ -53,53 +53,53 @@ fn get_product_version(target: Target) -> String {
         Target::Desktop => DESKTOP_VERSION_FILE_PATH,
     };
     println!("cargo:rerun-if-changed={version_file_path}");
-    let product_version = fs::read_to_string(version_file_path)
+
+    let release_version = fs::read_to_string(version_file_path)
         .unwrap_or_else(|_| panic!("Failed to read {version_file_path}"))
         .trim()
         .to_owned();
 
     // Compute the expected tag name for the release named `product_version`
     let release_tag = match target {
-        Target::Android => format!("android/{product_version}"),
-        Target::Desktop => product_version.to_owned(),
+        Target::Android => format!("android/{release_version}"),
+        Target::Desktop => release_version.to_owned(),
     };
 
-    // Rerun this build script on changes to the git ref that affects the build version.
-    // NOTE: This must be kept up to date with the behavior of `git_rev_parse_commit_hash`.
-    rerun_if_git_ref_changed(&release_tag)
-        .expect("Failed to set 'cargo:rerun-if-changed' on git ref changes");
-
-    if let Some(dev_suffix) = get_dev_suffix(&product_version) {
-        format!("{product_version}{dev_suffix}")
-    } else {
-        product_version
-    }
+    format!("{release_version}{}", get_suffix(&release_tag))
 }
 
-/// Returns the development suffix for the current build. A build has a development
-/// suffix if the build is not done on a git tag named `product_version`.
-/// This also returns `None` if the `git` command can't run, or the code does
-/// not live in a git repository.
-fn get_dev_suffix(release_tag: &str) -> Option<String> {
-    // Get the git commit hashes for the latest release and current HEAD
-    // Return `None` if unable to find the hash for HEAD.
-    let head_commit_hash = git_rev_parse_commit_hash("HEAD")?;
+/// Returns the suffix for the current build. If the build is done on a git tag named
+/// `product_version` or a git repository cannot be found, the suffix is empty. Otherwise,
+/// `-dev-$hash` is appended to the release version.
+fn get_suffix(release_tag: &str) -> String {
+    if !valid_git_repo() {
+        return String::new();
+    };
+    // Rerun this build script on changes to the git ref that affects the build version.
+    // NOTE: This must be kept up to date with the behavior of `git_rev_parse_commit_hash`.
+    rerun_if_git_ref_changed(release_tag);
+    let head_commit_hash =
+        git_rev_parse_commit_hash("HEAD").expect("Failed to run `git rev-parse HEAD^{{commit}}`");
     let product_version_commit_hash = git_rev_parse_commit_hash(release_tag);
 
     // If we are currently building the release tag, there is no dev suffix
     if Some(&head_commit_hash) == product_version_commit_hash.as_ref() {
-        return None;
+        String::new()
+    } else {
+        format!("-dev-{}", &head_commit_hash[..GIT_HASH_DEV_SUFFIX_LEN])
     }
-    Some(format!(
-        "-dev-{}",
-        &head_commit_hash[..GIT_HASH_DEV_SUFFIX_LEN]
-    ))
+}
+
+fn valid_git_repo() -> bool {
+    matches!(Command::new("git").arg("status").status(), Ok(status) if status.success())
 }
 
 /// Trigger rebuild of `mullvad-version` on changing branch (`.git/HEAD`), on changes to the ref of
 /// the current branch (`.git/refs/heads/$current_branch`) and on changes to the ref of the current
 /// release tag (`.git/refs/tags/$current_release_tag`).
-fn rerun_if_git_ref_changed(release_tag: &str) -> std::io::Result<()> {
+///
+/// Returns an error if not in a git repository, or the git binary is not in `PATH`.
+fn rerun_if_git_ref_changed(release_tag: &str) {
     let git_dir = Path::new("..").join(".git");
 
     // The `.git/HEAD` file contains the position of the current head. If in 'detached HEAD' state,
@@ -116,7 +116,8 @@ fn rerun_if_git_ref_changed(release_tag: &str) -> std::io::Result<()> {
     let output = Command::new("git")
         .arg("branch")
         .arg("--show-current")
-        .output()?;
+        .output()
+        .expect("Failed to execute `git branch --show-current`");
 
     let current_branch = String::from_utf8(output.stdout).unwrap();
     let current_branch = current_branch.trim();
@@ -146,19 +147,17 @@ fn rerun_if_git_ref_changed(release_tag: &str) -> std::io::Result<()> {
     // track this file, however, as any changes to the current branch, 'detached HEAD' state
     // or tags will update the corresponding `.git/refs` file we are tracking, even if it had
     // previously been pruned.
-
-    Ok(())
 }
 
 /// Returns the commit hash for the commit that `git_ref` is pointing to.
 ///
-/// Returns `None` if executing the `git rev-parse` command fails for some reason.
+/// Returns `None` if the git reference cannot be found.
 fn git_rev_parse_commit_hash(git_ref: &str) -> Option<String> {
     let output = Command::new("git")
         .arg("rev-parse")
         .arg(format!("{git_ref}^{{commit}}"))
         .output()
-        .ok()?;
+        .expect("Failed to run `git rev-parse`");
     if !output.status.success() {
         return None;
     }
--- a/mullvad-version/build.rs
+++ b/mullvad-version/build.rs
@@ -52,15 +52,26 @@ fn get_product_version(target: Target) -> String {
         Target::Desktop => DESKTOP_VERSION_FILE_PATH,
     };
     println!("cargo:rerun-if-changed={version_file_path}");
-    let version = fs::read_to_string(version_file_path)
+    let product_version = fs::read_to_string(version_file_path)
         .unwrap_or_else(|_| panic!("Failed to read {version_file_path}"))
         .trim()
         .to_owned();
 
-    if let Some(dev_suffix) = get_dev_suffix(target, &version) {
-        format!("{version}{dev_suffix}")
+    // Compute the expected tag name for the release named `product_version`
+    let release_tag = match target {
+        Target::Android => format!("android/{product_version}"),
+        Target::Desktop => product_version.to_owned(),
+    };
+
+    // Rerun this build script on changes to the git ref that affects the build version.
+    // NOTE: This must be kept up to date with the behavior of `git_rev_parse_commit_hash`.
+    rerun_if_git_ref_changed(&release_tag)
+        .expect("Failed to set 'cargo:rerun-if-changed' on git ref changes");
+
+    if let Some(dev_suffix) = get_dev_suffix(&product_version) {
+        format!("{product_version}{dev_suffix}")
     } else {
-        version
+        product_version
     }
 }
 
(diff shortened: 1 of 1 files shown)


faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @Serock3)

___
*[`mullvad-version/build.rs` line 97 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NwwtcVs84GTbTYfe6Y6:-NwwtcVs84GTbTYfe6Y7:bcovqlz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ee773f074b98c45b2bde07915a4bc15b0a281/mullvad-version/build.rs#L97)):*
> ```XML
>         .join("refs")
>         .join("heads")
>         .join(current_branch.trim());
> ```

Please mind detached head state. Seems like `git branch --show-current` just outputs nothing (empty string) in this case. I think this logic needs to be behind some `if !current_branch.trim().is_empty() {...}` because I guess the `.exists()` check will succeed because `.git/refs/head` exists and is a directory.
___
*[`mullvad-version/build.rs` line 110 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NwwuEP2E66lYuCigHo8:-NwwuEP2E66lYuCigHo9:b1t2j6v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ee773f074b98c45b2bde07915a4bc15b0a281/mullvad-version/build.rs#L110)):*
> ```XML
>             git_current_branch_ref.display()
>         );
>     }
> ```

`get_dev_suffix` now became significantly longer. Would it be possible to extract all this `cargo:rerun-if-changed` logic to a separate function(s)? Or some other way to massage this into shorter more dedicated functions.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @Serock3)

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[`mullvad-version/build.rs` line 97 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NwwtcVs84GTbTYfe6Y6:-NwxiNguAPp20igXTnYZ:b-vh0vjr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ee773f074b98c45b2bde07915a4bc15b0a281/mullvad-version/build.rs#L97)):*

Please mind detached head state. Seems like `git branch --show-current` just outputs nothing (empty string) in this case. I think this logic needs to be behind some `if !current_branch.trim().is_empty() {...}` because I guess the `.exists()` check will succeed because `.git/refs/head` exists and is a directory.
</blockquote></details>

Good catch. I added the check. Note that`.git/HEAD` contains the current commit ref when in detached head state, so we don't need to track anyhing in `.git/refs/heads/` to rebuild on e.g. adding new commits.
___
*[`mullvad-version/build.rs` line 110 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NwwuEP2E66lYuCigHo8:-NwxV1XN53Ay0KHW8Xyj:b9an43d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ee773f074b98c45b2bde07915a4bc15b0a281/mullvad-version/build.rs#L110)):*

`get_dev_suffix` now became significantly longer. Would it be possible to extract all this `cargo:rerun-if-changed` logic to a separate function(s)? Or some other way to massage this into shorter more dedicated functions.
</blockquote></details>

I put it in a separate function and added a bunch of comments enplaning the reasoning for the files we track. I also added a section on the `.git/packed-refs` file, and why we don't need to track it

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @dlon and @Serock3)

___
*[`mullvad-version/build.rs` line 97 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NwwtcVs84GTbTYfe6Y6:-Nwy2SfkBn_1crYhTXQf:b-iogxti) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ee773f074b98c45b2bde07915a4bc15b0a281/mullvad-version/build.rs#L97)):*

Good catch. I added the check. Note that`.git/HEAD` contains the current commit ref when in detached head state, so we don't need to track anyhing in `.git/refs/heads/` to rebuild on e.g. adding new commits.
</blockquote></details>

You don't do any trimming. Have you verified this works as intended? I'm just afraid the git output can be a single newline or something like that.
___
*[`mullvad-version/build.rs` line 110 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NwwuEP2E66lYuCigHo8:-Nwy2mcvCgaGYqMBLWQB:b-83sjoh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ee773f074b98c45b2bde07915a4bc15b0a281/mullvad-version/build.rs#L110)):*

I put it in a separate function and added a bunch of comments enplaning the reasoning for the files we track. I also added a section on the `.git/packed-refs` file, and why we don't need to track it
</blockquote></details>

Nice!
___
*[`mullvad-version/build.rs` line 98 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy20APBrD2Gz_gJFzU:-Nwy20APBrD2Gz_gJFzV:b-y9d2ec) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L98)):*
> ```XML
> /// the current branch (`.git/refs/heads/$current_branch`) and on changes to the ref of the current
> /// release tag (`.git/refs/tags/$current_release_tag`).
> fn rerun_if_git_ref_changed(release_tag: &String) -> Option<()> {
> ```

Nit. But more idiomatic to accept `&str` than `&String` when you just need to immutably borrow.
___
*[`mullvad-version/build.rs` line 145 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy3eRUAQKVTPmnUxmm:-Nwy3eRUAQKVTPmnUxmn:b-qqumen) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L145)):*
> ```XML
>     // this folder, however, as any changes to the current branch, 'detached HEAD' state or tags
>     // will update the corresponding `.git/refs` file we are tracking, even if it had previously
>     // been pruned.
> ```

Nit, but this comment really looks like it belongs to the return value, which I don't think is the case? Maybe separate them by one or two extra newlines?
___
*[`mullvad-version/build.rs` line 146 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy3Kfa-Yt7nq0bJ9lJ:-Nwy3Kfa-Yt7nq0bJ9lK:b9m6kv3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L146)):*
> ```XML
>     // will update the corresponding `.git/refs` file we are tracking, even if it had previously
>     // been pruned.
>     Some(())
> ```

Can the return type here maybe be changed to `Result<(), ...>`? `Option<()>` is a bit weird. The only early return I can spot is the `git branch` subprocess. If you just remove the `.ok()` call you'll get the correct type directly. Returning `Ok(())` looks more idiomatic and familiar than `Some(())`.

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @dlon and @faern)

___
*[`mullvad-version/build.rs` line 97 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NwwtcVs84GTbTYfe6Y6:-NwyR5cnDVBzt1uaD8_V:bv55luy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ee773f074b98c45b2bde07915a4bc15b0a281/mullvad-version/build.rs#L97)):*

You don't do any trimming. Have you verified this works as intended? I'm just afraid the git output can be a single newline or something like that.
</blockquote></details>

Miss on my part, of course the trimming should be done before the empty-check. Although after inspecting the output, it does not seem to contain a newline when in detached head state, so maybe the trim can be removed.
___
*[`mullvad-version/build.rs` line 110 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NwwuEP2E66lYuCigHo8:-NwyNE8S35Jly5VdhgwR:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ee773f074b98c45b2bde07915a4bc15b0a281/mullvad-version/build.rs#L110)):*

Nice!
</blockquote></details>

Done.
___
*[`mullvad-version/build.rs` line 98 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy20APBrD2Gz_gJFzU:-NwyKIFi1GW18CmIJhm9:b-qcqxdy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L98)):*

Nit. But more idiomatic to accept `&str` than `&String` when you just need to immutably borrow.
</blockquote></details>

Ah, I used the `rust-analyzer` code action "Extract into function" and didn't notice the `@String`. Good catch again!
___
*[`mullvad-version/build.rs` line 145 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy3eRUAQKVTPmnUxmm:-NwyNIdr-VKNctWTssBO:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L145)):*

Nit, but this comment really looks like it belongs to the return value, which I don't think is the case? Maybe separate them by one or two extra newlines?
</blockquote></details>

Done
___
*[`mullvad-version/build.rs` line 146 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy3Kfa-Yt7nq0bJ9lJ:-NwyNcJE2NgUlMcxOZee:b-u8ooln) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L146)):*

Can the return type here maybe be changed to `Result<(), ...>`? `Option<()>` is a bit weird. The only early return I can spot is the `git branch` subprocess. If you just remove the `.ok()` call you'll get the correct type directly. Returning `Ok(())` looks more idiomatic and familiar than `Some(())`.
</blockquote></details>

Done. But note that since we only use `Option<T>` as return types throughout the module we still have to call `.ok()` later on

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: all files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[`mullvad-version/build.rs` line 78 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxCitIZ4YXWs2sSlh2s:-NxCitI_847eeW4bv-xB:bnzvw7a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/605d9f9e9cf0205be6d5671b639dab42ded2d70c/mullvad-version/build.rs#L78)):*
> ```XML
>     };
> 
>     rerun_if_git_ref_changed(&release_tag).ok()?;
> ```

Does this have to be a side-effect of calling `get_dev_suffix`?

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[`mullvad-version/build.rs` line 78 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxCitIZ4YXWs2sSlh2s:-NxHCjfy0YLq_6A7N1PS:bx4bcf3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/605d9f9e9cf0205be6d5671b639dab42ded2d70c/mullvad-version/build.rs#L78)):*

Does this have to be a side-effect of calling `get_dev_suffix`?
</blockquote></details>

No, I moved it to `get_product_version`, which ended up a bit cleaner.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @dlon)

___
*[`mullvad-version/build.rs` line 146 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy3Kfa-Yt7nq0bJ9lJ:-NxHzWZ844mCBm0hBglh:bjfbm3j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L146)):*

Done. But note that since we only use `Option<T>` as return types throughout the module we still have to call `.ok()` later on
</blockquote></details>

Of course! But we can still make each function as idiomatic as possible. Moving those things upwards to the level where they make the most sense.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @dlon and @Serock3)

___
*[`mullvad-version/build.rs` line 78 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxCitIZ4YXWs2sSlh2s:-NxLoFthCXTPBkNd86VS:blqydob) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/605d9f9e9cf0205be6d5671b639dab42ded2d70c/mullvad-version/build.rs#L78)):*

No, I moved it to `get_product_version`, which ended up a bit cleaner.
</blockquote></details>

This is fine, but I don't 100% agree it is better (no need to change anything).

Now the "dependencies" for computing the version string are moved further away from where they are used. If we now were to start depending on `.git/foo/bar` for our version computation it could be easy to forget to emit the rerun-if since it's in a different place to where the reading of the file happens.
___
*[`mullvad-version/build.rs` line 80 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxLloroARJFghiGKmiW:-NxLloroARJFghiGKmiX:b-ycdhsy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ae75d413e05af84ec263ba1abba8d6b43481f139/mullvad-version/build.rs#L80)):*
> ```XML
> /// This also returns `None` if the `git` command can't run, or the code does
> /// not live in a git repository.
> fn get_dev_suffix(release_tag: &str) -> Option<String> {
> ```

The documentation for this function is now wrong, please update it :pray: 
___
*[`mullvad-version/build.rs` line 134 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxLnT7-BBzK9G85i9Mp:-NxLnT7-BBzK9G85i9Mq:bhjezz1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ae75d413e05af84ec263ba1abba8d6b43481f139/mullvad-version/build.rs#L134)):*
Nit. Tags rarely move, so even if the comment is not wrong it reads strangely to me. In practice this would trigger if the release tag was *created* most of the time. How about something like:

> Since the product version depends on if the build is done on the commit with the corresponding release tag or not, we must track creation of/changes to said tag

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

___
*[`mullvad-version/build.rs` line 146 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy3Kfa-Yt7nq0bJ9lJ:-NxMqdmf51X1KVm6Vg3G:b-h38meh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L146)):*

Of course! But we can still make each function as idiomatic as possible. Moving those things upwards to the level where they make the most sense.
</blockquote></details>

I ended up taking this idea a bit further and also changed the return type of `git_rev_parse_commit_hash` from an `Option<String>` to just `String` and panicking on failed calls to `git rev-parse`. The previous behavior was to fail silently which ended up producing a release version string, which seemed wrong to me. This could be triggered if, e.g., the text in the `product-version.txt` file was changed to a tag that doesn't exist. Say that I change it to `2042.3` by mistake, then the computed version would just become `2042.3` regardless of the git ref of HEAD, instead of emitting an error.
___
*[`mullvad-version/build.rs` line 78 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxCitIZ4YXWs2sSlh2s:-NxMTbxb1JbHNlLXPQxf:b-ohm012) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/605d9f9e9cf0205be6d5671b639dab42ded2d70c/mullvad-version/build.rs#L78)):*

This is fine, but I don't 100% agree it is better (no need to change anything).

Now the "dependencies" for computing the version string are moved further away from where they are used. If we now were to start depending on `.git/foo/bar` for our version computation it could be easy to forget to emit the rerun-if since it's in a different place to where the reading of the file happens.
</blockquote></details>

I agree with your sentiment. I ended up removing and inlining the `get_dev_suffix` fn which simplified the code a bit and I added a comment about keeping `rerun_if_git_ref_changed` updated with `git_rev_parse_commit_hash`.
___
*[`mullvad-version/build.rs` line 80 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxLloroARJFghiGKmiW:-NxMqTBr0nVWqkDAlfPP:bmz2ke8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ae75d413e05af84ec263ba1abba8d6b43481f139/mullvad-version/build.rs#L80)):*

The documentation for this function is now wrong, please update it :pray:
</blockquote></details>

See comment above. I updated the text and moved it to a normal comment where `get_dev_suffix` previously was called.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @dlon and @Serock3)

___
*[`mullvad-version/build.rs` line 146 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy3Kfa-Yt7nq0bJ9lJ:-NxO4pmoERkPfn4_sGCv:b-co5arm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L146)):*

I ended up taking this idea a bit further and also changed the return type of `git_rev_parse_commit_hash` from an `Option<String>` to just `String` and panicking on failed calls to `git rev-parse`. The previous behavior was to fail silently which ended up producing a release version string, which seemed wrong to me. This could be triggered if, e.g., the text in the `product-version.txt` file was changed to a tag that doesn't exist. Say that I change it to `2042.3` by mistake, then the computed version would just become `2042.3` regardless of the git ref of HEAD, instead of emitting an error.
</blockquote></details>

Please mind that `mullvad-version` must still build even if `git` is not installed or if the `.git` directory is removed. So this change seems wrong to me

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @dlon and @Serock3)

___
*[`mullvad-version/build.rs` line 146 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy3Kfa-Yt7nq0bJ9lJ:-NxO5bC_ED-HDONQjh7u:bt5dpo0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L146)):*

Please mind that `mullvad-version` must still build even if `git` is not installed or if the `.git` directory is removed. So this change seems wrong to me
</blockquote></details>

This is documented in the documentation for `get_dev_suffix` and that was the correct behavior. The code living in a git repository should not be a prerequisite for building the app.

dlon (commented):
> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: all files reviewed, 2 unresolved discussions (waiting on @faern and @Serock3)

___
*[`mullvad-version/build.rs` line 78 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxCitIZ4YXWs2sSlh2s:-NxMBEgL8P424BHFM0I-:b6v5uov) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/605d9f9e9cf0205be6d5671b639dab42ded2d70c/mullvad-version/build.rs#L78)):*

I agree with your sentiment. I ended up removing and inlining the `get_dev_suffix` fn which simplified the code a bit and I added a comment about keeping `rerun_if_git_ref_changed` updated with `git_rev_parse_commit_hash`.
</blockquote></details>

In my opinion, functions with a `get` prefix should not have any meaningful side effects, such as emitting `rerun-if-changed`.

Since this is a small build script, I don't think the point is strong. But that's what I was initially hinting at.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: all files reviewed, 1 unresolved discussion (waiting on @Serock3)

___
*[`mullvad-version/build.rs` line 78 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxCitIZ4YXWs2sSlh2s:-NxWT1pX2mRuB-42amaN:bxuvek5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/605d9f9e9cf0205be6d5671b639dab42ded2d70c/mullvad-version/build.rs#L78)):*

In my opinion, functions with a `get` prefix should not have any meaningful side effects, such as emitting `rerun-if-changed`.

Since this is a small build script, I don't think the point is strong. But that's what I was initially hinting at.
</blockquote></details>

If the *naming* is the issue here we can rename it to `compute` :shrug: I think a design that prevents easy mistakes and keep related things close together is more important.

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`mullvad-version/build.rs` line 146 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy3Kfa-Yt7nq0bJ9lJ:-Nxkzn1eD-7hmb39BaVg:b-5bpnin) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L146)):*

This is documented in the documentation for `get_dev_suffix` and that was the correct behavior. The code living in a git repository should not be a prerequisite for building the app.
</blockquote></details>

Ok, but falling back to assuming a release tag seems like the wrong behavior to me. What about having a custom release tag for this scenario? "Dirty" (as in `2024.2-dev-dirty`) is commonly used to indicate that there are modified files that are not commited, which is not quite correct in this case. Maybe `2024.2-dev-unversion`?
___
*[`mullvad-version/build.rs` line 78 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxCitIZ4YXWs2sSlh2s:-NxkyXsTAGZke9JXqslY:brpwssp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/605d9f9e9cf0205be6d5671b639dab42ded2d70c/mullvad-version/build.rs#L78)):*

If the *naming* is the issue here we can rename it to `compute` :shrug: I think a design that prevents easy mistakes and keep related things close together is more important.
</blockquote></details>

Renamed the fn to `compute_product_version`.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`mullvad-version/build.rs` line 78 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxCitIZ4YXWs2sSlh2s:-Nxl3DOu3t0zaRWbse07:b-ldamgq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/605d9f9e9cf0205be6d5671b639dab42ded2d70c/mullvad-version/build.rs#L78)):*

Renamed the fn to `compute_product_version`.
</blockquote></details>

I'm ok with this, but, to be honest, changing the prefix to `compute` doesn't change anything

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

___
*[`mullvad-version/build.rs` line 78 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-NxCitIZ4YXWs2sSlh2s:-Nxl3p3ZBbGr2Qh0C1n3:b-1dzc37) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/605d9f9e9cf0205be6d5671b639dab42ded2d70c/mullvad-version/build.rs#L78)):*

I'm ok with this, but, to be honest, changing the prefix to `compute` doesn't change anything
</blockquote></details>

Sure doesn't. Not sure if there are idiomatic names for non-pure functions 🤷. I updated the docstring to explicitly mention the side effects.

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @dlon and @faern)

___
*[`mullvad-version/build.rs` line 146 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-Nwy3Kfa-Yt7nq0bJ9lJ:-Nxm9-xJ7Pg_mCFb7fBJ:boamvtt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a90870edac3371e7c21f7f53790f41d42cb18550/mullvad-version/build.rs#L146)):*

Ok, but falling back to assuming a release tag seems like the wrong behavior to me. What about having a custom release tag for this scenario? "Dirty" (as in `2024.2-dev-dirty`) is commonly used to indicate that there are modified files that are not commited, which is not quite correct in this case. Maybe `2024.2-dev-unversion`?
</blockquote></details>

I ended up reverting this change and the one to `git_rev_parse_commit_hash`. Although I like the idea of validating that the release tag matches an existing tag and emitting a nice error otherwise, it caused problems with the CI where tags are not currently fetched. Fetching release tags also may not be a requirement we want to force on users that want to build the app. I did however add an explicit check for being in a valid repository, otherwise `-dev-unknown` is appended to the version. If the repo is valid but the release tag still cannot be found, we now default to appending a `-dev-$hash` suffix to make it slightly harder to accidentally produce something that looks like a release build.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

Serock3:
> I did however add an explicit check for being in a valid repository, otherwise `-dev-unknown` is appended to the version.

I reverted this behavior, as per our discussion.

>  If the repo is valid but the release tag still cannot be found, we now default to appending a -dev-$hash suffix to make it slightly harder to accidentally produce something that looks like a release build.

This was how it worked from the start. The behavior is unchanged, sorry for the confusion on my part.

faern (approved):
:lgtm: Good job! I could have been clearer with the requirements of this file up front. This is probably going to be a good time saver with regards to compile times!

> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6211#-:-Nxw_Y3W1jJy4zvLLYf3:biqwz3i)* status: all files reviewed, 1 unresolved discussion (waiting on @Serock3)

PR #6257, opened 2024-05-15
Upgrade settings to associate with multi-hop

mojganii:
This PR introduces the ability to store entry and exit locations in RelayConstraints to implemnt  multi-hop feature. Additionally, users can enable or disable this feature through the settings, which will also upgrade the tunnel settings accordingly. 
Note that, for backward compatibility, the entry location functionality is not yet implemented. However, in an upcoming PR, we will refactor the related code sections to incorporate entry locations, either.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)

### DIFF
--- a/ios/MullvadREST/Relay/NoRelaysSatisfyingConstraintsError.swift
+++ b/ios/MullvadREST/Relay/NoRelaysSatisfyingConstraintsError.swift
@@ -1,0 +1,15 @@
+//
+//  NoRelaysSatisfyingConstraintsError.swift
+//  MullvadREST
+//
+//  Created by Mojgan on 2024-04-26.
+//  Copyright © 2024 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+
+public struct NoRelaysSatisfyingConstraintsError: LocalizedError {
+    public var errorDescription: String? {
+        "No relays satisfying constraints."
+    }
+}
--- a/ios/MullvadREST/Relay/RelaySelector+Shadowsocks.swift
+++ b/ios/MullvadREST/Relay/RelaySelector+Shadowsocks.swift
@@ -1,0 +1,91 @@
+//
+//  RelaySelector+Shadowsocks.swift
+//  MullvadREST
+//
+//  Created by Mojgan on 2024-05-17.
+//  Copyright © 2024 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+import MullvadTypes
+
+extension RelaySelector {
+    public enum Shadowsocks {
+        /**
+         Returns random shadowsocks TCP bridge, otherwise `nil` if there are no shadowdsocks bridges.
+         */
+        public static func tcpBridge(from relays: REST.ServerRelaysResponse) -> REST.ServerShadowsocks? {
+            relays.bridge.shadowsocks.filter { $0.protocol == "tcp" }.randomElement()
+        }
+
+        /// Return a random Shadowsocks bridge relay, or `nil` if no relay were found.
+        ///
+        /// Non `active` relays are filtered out.
+        /// - Parameter relays: The list of relays to randomly select from.
+        /// - Returns: A Shadowsocks relay or `nil` if no active relay were found.
+        public static func relay(from relaysResponse: REST.ServerRelaysResponse) -> REST.BridgeRelay? {
+            relaysResponse.bridge.relays.filter { $0.active }.randomElement()
+        }
+
+        /// Returns the closest Shadowsocks relay using the given `location`, or a random relay if `constraints` were
+        /// unsatisfiable.
+        ///
+        /// - Parameters:
+        ///   - location: The user selected `location`
+        ///   - port: The user selected port
+        ///   - filter: The user filtered criteria
+        ///   - relays: The list of relays to randomly select from.
+        /// - Returns: A Shadowsocks relay or `nil` if no active relay were found.
+        public static func closestRelay(
+            location: RelayConstraint<UserSelectedRelays>,
+            port: RelayConstraint<UInt16>,
+            filter: RelayConstraint<RelayFilter>,
+            in relaysResponse: REST.ServerRelaysResponse
+        ) -> REST.BridgeRelay? {
+            let mappedBridges = mapRelays(relays: relaysResponse.bridge.relays, locations: relaysResponse.locations)
+            let filteredRelays = applyConstraints(
+                location,
+                portConstraint: port,
+                filterConstraint: filter,
+                relays: mappedBridges
+            )
+            guard filteredRelays.isEmpty == false else { return relay(from: relaysResponse) }
+
+            // Compute the midpoint location from all the filtered relays
+            // Take *either* the first five relays, OR the relays below maximum bridge distance
+            // sort all of them by Haversine distance from the computed midpoint location
+            // then use the roulette selection to pick a bridge
+
+            let midpointDistance = Midpoint.location(in: filteredRelays.map { $0.serverLocation.geoCoordinate })
+            let maximumBridgeDistance = 1500.0
+            let relaysWithDistance = filteredRelays.map {
+                RelayWithDistance(
+                    relay: $0.relay,
+                    distance: Haversine.distance(
+                        midpointDistance.latitude,
+                        midpointDistance.longitude,
+                        $0.serverLocation.latitude,
+                        $0.serverLocation.longitude
+                    )
+                )
+            }.sorted {
+                $0.distance < $1.distance
+            }.filter {
+                $0.distance <= maximumBridgeDistance
+            }.prefix(5)
+
+            var greatestDistance = 0.0
+            relaysWithDistance.forEach {
+                if $0.distance > greatestDistance {
+                    greatestDistance = $0.distance
+                }
+            }
+
+            let randomRelay = rouletteSelection(relays: Array(relaysWithDistance), weightFunction: { relay in
+                UInt64(1 + greatestDistance - relay.distance)
+            })
+
+            return randomRelay?.relay ?? filteredRelays.randomElement()?.relay
+        }
+    }
+}
--- a/ios/MullvadREST/Relay/RelaySelector+Wireguard.swift
+++ b/ios/MullvadREST/Relay/RelaySelector+Wireguard.swift
@@ -1,0 +1,78 @@
+//
+//  RelaySelector+Wireguard.swift
+//  MullvadREST
+//
+//  Created by Mojgan on 2024-05-17.
+//  Copyright © 2024 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+import MullvadTypes
+
+extension RelaySelector {
+    public enum WireGuard {
+        /**
+         Filters relay list using given constraints and selects random relay for exit relay.
+         Throws an error if there are no relays satisfying the given constraints.
+         */
+        public static func evaluate(
+            by constraints: RelayConstraints,
+            in relaysResponse: REST.ServerRelaysResponse,
+            numberOfFailedAttempts: UInt
+        ) throws -> RelaySelectorResult {
+            let exitCandidates = try findBestMatch(
+                relays: relaysResponse,
+                relayConstraint: constraints.exitLocations,
+                portConstraint: constraints.port,
+                filterConstraint: constraints.filter,
+                numberOfFailedAttempts: numberOfFailedAttempts
+            )
+
+            return exitCandidates
+        }
+
+        // MARK: - private functions
+
+        private static func findBestMatch(
+            relays: REST.ServerRelaysResponse,
+            relayConstraint: RelayConstraint<UserSelectedRelays>,
+            portConstraint: RelayConstraint<UInt16>,
+            filterConstraint: RelayConstraint<RelayFilter>,
+            numberOfFailedAttempts: UInt
+        ) throws -> RelaySelectorMatch {
+            let mappedRelays = mapRelays(relays: relays.wireguard.relays, locations: relays.locations)
+            let filteredRelays = applyConstraints(
+                relayConstraint,
+                portConstraint: portConstraint,
+                filterConstraint: filterConstraint,
+                relays: mappedRelays
+            )
+            let port = applyPortConstraint(
+                portConstraint,
+                rawPortRanges: relays.wireguard.portRanges,
+                numberOfFailedAttempts: numberOfFailedAttempts
+            )
+
+            guard let relayWithLocation = pickRandomRelayByWeight(relays: filteredRelays), let port else {
+                throw NoRelaysSatisfyingConstraintsError()
+            }
+
+            let endpoint = MullvadEndpoint(
+                ipv4Relay: IPv4Endpoint(
+                    ip: relayWithLocation.relay.ipv4AddrIn,
+                    port: port
+                ),
+                ipv6Relay: nil,
+                ipv4Gateway: relays.wireguard.ipv4Gateway,
+                ipv6Gateway: relays.wireguard.ipv6Gateway,
+                publicKey: relayWithLocation.relay.publicKey
+            )
+
+            return RelaySelectorMatch(
+                endpoint: endpoint,
+                relay: relayWithLocation.relay,
+                location: relayWithLocation.serverLocation
+            )
+        }
+    }
+}
--- a/ios/MullvadREST/Relay/RelaySelector.swift
+++ b/ios/MullvadREST/Relay/RelaySelector.swift
@@ -12,112 +12,7 @@ import MullvadTypes
 private let defaultPort: UInt16 = 53
 
 public enum RelaySelector {
-    /**
-     Returns random shadowsocks TCP bridge, otherwise `nil` if there are no shadowdsocks bridges.
-     */
-    public static func shadowsocksTCPBridge(from relays: REST.ServerRelaysResponse) -> REST.ServerShadowsocks? {
-        relays.bridge.shadowsocks.filter { $0.protocol == "tcp" }.randomElement()
-    }
-
-    /// Return a random Shadowsocks bridge relay, or `nil` if no relay were found.
-    ///
-    /// Non `active` relays are filtered out.
-    /// - Parameter relays: The list of relays to randomly select from.
-    /// - Returns: A Shadowsocks relay or `nil` if no active relay were found.
-    public static func shadowsocksRelay(from relaysResponse: REST.ServerRelaysResponse) -> REST.BridgeRelay? {
-        relaysResponse.bridge.relays.filter { $0.active }.randomElement()
-    }
-
-    /// Returns the closest Shadowsocks relay using the given `constraints`, or a random relay if `constraints` were
-    /// unsatisfiable.
-    ///
-    /// - Parameters:
-    ///   - constraints: The user selected `constraints`
-    ///   - relays: The list of relays to randomly select from.
-    /// - Returns: A Shadowsocks relay or `nil` if no active relay were found.
-    public static func closestShadowsocksRelayConstrained(
-        by constraints: RelayConstraints,
-        in relaysResponse: REST.ServerRelaysResponse
-    ) -> REST.BridgeRelay? {
-        let mappedBridges = mapRelays(relays: relaysResponse.bridge.relays, locations: relaysResponse.locations)
-        let filteredRelays = applyConstraints(constraints, relays: mappedBridges)
-        guard filteredRelays.isEmpty == false else { return shadowsocksRelay(from: relaysResponse) }
-
-        // Compute the midpoint location from all the filtered relays
-        // Take *either* the first five relays, OR the relays below maximum bridge distance
-        // sort all of them by Haversine distance from the computed midpoint location
-        // then use the roulette selection to pick a bridge
-
-        let midpointDistance = Midpoint.location(in: filteredRelays.map { $0.serverLocation.geoCoordinate })
-        let maximumBridgeDistance = 1500.0
-        let relaysWithDistance = filteredRelays.map {
-            RelayWithDistance(
-                relay: $0.relay,
-                distance: Haversine.distance(
-                    midpointDistance.latitude,
-                    midpointDistance.longitude,
-                    $0.serverLocation.latitude,
-                    $0.serverLocation.longitude
-                )
-            )
-        }.sorted {
-            $0.distance < $1.distance
-        }.filter {
-            $0.distance <= maximumBridgeDistance
-        }.prefix(5)
-
-        var greatestDistance = 0.0
-        relaysWithDistance.forEach {
-            if $0.distance > greatestDistance {
-                greatestDistance = $0.distance
-            }
-        }
-
-        let randomRelay = rouletteSelection(relays: Array(relaysWithDistance), weightFunction: { relay in
-            UInt64(1 + greatestDistance - relay.distance)
-        })
-
-        return randomRelay?.relay ?? filteredRelays.randomElement()?.relay
-    }
-
-    /**
-     Filters relay list using given constraints and selects random relay.
-     Throws an error if there are no relays satisfying the given constraints.
-     */
-    public static func evaluate(
-        relays: REST.ServerRelaysResponse,
-        constraints: RelayConstraints,
-        numberOfFailedAttempts: UInt
-    ) throws -> RelaySelectorResult {
-        let mappedRelays = mapRelays(relays: relays.wireguard.relays, locations: relays.locations)
-        let filteredRelays = applyConstraints(constraints, relays: mappedRelays)
-        let port = applyConstraints(
-            constraints,
-            rawPortRanges: relays.wireguard.portRanges,
-            numberOfFailedAttempts: numberOfFailedAttempts
-        )
-
-        guard let relayWithLocation = pickRandomRelayByWeight(relays: filteredRelays), let port else {
-            throw NoRelaysSatisfyingConstraintsError()
-        }
-
-        let endpoint = MullvadEndpoint(
-            ipv4Relay: IPv4Endpoint(
-                ip: relayWithLocation.relay.ipv4AddrIn,
-                port: port
-            ),
-            ipv6Relay: nil,
-            ipv4Gateway: relays.wireguard.ipv4Gateway,
-            ipv6Gateway: relays.wireguard.ipv6Gateway,
-            publicKey: relayWithLocation.relay.publicKey
-        )
-
-        return RelaySelectorResult(
-            endpoint: endpoint,
-            relay: relayWithLocation.relay,
-            location: relayWithLocation.serverLocation
-        )
-    }
+    // MARK: - public
 
     /// Determines whether a `REST.ServerRelay` satisfies the given relay filter.
     public static func relayMatchesFilter(_ relay: AnyRelay, filter: RelayFilter) -> Bool {
@@ -135,86 +30,37 @@ public enum RelaySelector {
         }
     }
 
-    /// Produce a list of `RelayWithLocation` items satisfying the given constraints
-    static func applyConstraints<T: AnyRelay>(
-        _ constraints: RelayConstraints,
-        relays: [RelayWithLocation<T>]
-    ) -> [RelayWithLocation<T>] {
-        // Filter on active status, filter, and location.
-        let filteredRelays = relays.filter { relayWithLocation -> Bool in
-            guard relayWithLocation.relay.active else {
-                return false
-            }
+    // MARK: - private
 
-            switch constraints.filter {
-            case .any:
-                break
-            case let .only(filter):
-                if !relayMatchesFilter(relayWithLocation.relay, filter: filter) {
-                    return false
-                }
-            }
-
-            return switch constraints.locations {
-            case .any:
-                true
-            case let .only(relayConstraint):
-                // At least one location must match the relay under test.
-                relayConstraint.locations.contains { location in
-                    relayWithLocation.matches(location: location)
-                }
-            }
-        }
-
-        // Filter on country inclusion.
-        let includeInCountryFilteredRelays = filteredRelays.filter { relayWithLocation in
-            return switch constraints.locations {
-            case .any:
-                true
-            case let .only(relayConstraint):
-                relayConstraint.locations.contains { location in
-                    if case .country = location {
-                        return relayWithLocation.relay.includeInCountry
-                    }
-                    return false
-                }
-            }
-        }
-
-        // If no relays should be included in the matched country, instead accept all.
-        if includeInCountryFilteredRelays.isEmpty {
-            return filteredRelays
-        } else {
-            return includeInCountryFilteredRelays
-        }
-    }
-
-    /// Produce a port that is either user provided or randomly selected, satisfying the given constraints.
-    private static func applyConstraints(
-        _ constraints: RelayConstraints,
-        rawPortRanges: [[UInt16]],
-        numberOfFailedAttempts: UInt
-    ) -> UInt16? {
-        switch constraints.port {
-        case let .only(port):
-            return port
-
-        case .any:
-            // 1. First two attempts should pick a random port.
-            // 2. The next two should pick port 53.
-            // 3. Repeat steps 1 and 2.
-            let useDefaultPort = (numberOfFailedAttempts % 4 == 2) || (numberOfFailedAttempts % 4 == 3)
-
-            return useDefaultPort ? defaultPort : pickRandomPort(rawPortRanges: rawPortRanges)
-        }
-    }
-
-    private static func pickRandomRelayByWeight<T: AnyRelay>(relays: [RelayWithLocation<T>])
+    static func pickRandomRelayByWeight<T: AnyRelay>(relays: [RelayWithLocation<T>])
         -> RelayWithLocation<T>? {
         rouletteSelection(relays: relays, weightFunction: { relayWithLocation in relayWithLocation.relay.weight })
     }
 
-    private static func rouletteSelection<T>(relays: [T], weightFunction: (T) -> UInt64) -> T? {
+    private static func pickRandomPort(rawPortRanges: [[UInt16]]) -> UInt16? {
+        let portRanges = parseRawPortRanges(rawPortRanges)
+        let portAmount = portRanges.reduce(0) { partialResult, closedRange in
+            partialResult + closedRange.count
+        }
+
+        guard var portIndex = (0 ..< portAmount).randomElement() else {
+            return nil
+        }
+
+        for range in portRanges {
+            if portIndex < range.count {
+                return UInt16(portIndex) + range.lowerBound
+            } else {
+                portIndex -= range.count
+            }
+        }
+
+        assertionFailure("Port selection algorithm is broken!")
+
+        return nil
+    }
+
+    static func rouletteSelection<T>(relays: [T], weightFunction: (T) -> UInt64) -> T? {
         let totalWeight = relays.map { weightFunction($0) }.reduce(0) { accumulated, weight in
             accumulated + weight
         }
@@ -241,45 +87,7 @@ public enum RelaySelector {
         return randomRelay
     }
 
-    private static func pickRandomPort(rawPortRanges: [[UInt16]]) -> UInt16? {
-        let portRanges = parseRawPortRanges(rawPortRanges)
-        let portAmount = portRanges.reduce(0) { partialResult, closedRange in
-            partialResult + closedRange.count
-        }
-
-        guard var portIndex = (0 ..< portAmount).randomElement() else {
-            return nil
-        }
-
-        for range in portRanges {
-            if portIndex < range.count {
-                return UInt16(portIndex) + range.lowerBound
-            } else {
-                portIndex -= range.count
-            }
-        }
-
-        assertionFailure("Port selection algorithm is broken!")
-
-        return nil
-    }
-
-    private static func parseRawPortRanges(_ rawPortRanges: [[UInt16]]) -> [ClosedRange<UInt16>] {
-        rawPortRanges.compactMap { inputRange -> ClosedRange<UInt16>? in
-            guard inputRange.count == 2 else { return nil }
-
-            let startPort = inputRange[0]
-            let endPort = inputRange[1]
-
-            if startPort <= endPort {
-                return startPort ... endPort
-            } else {
-                return nil
-            }
-        }
-    }
-
-    private static func mapRelays<T: AnyRelay>(
+    static func mapRelays<T: AnyRelay>(
         relays: [T],
         locations: [String: REST.ServerLocation]
     ) -> [RelayWithLocation<T>] {
@@ -307,42 +115,95 @@ public enum RelaySelector {
 
         return RelayWithLocation(relay: relay, serverLocation: location)
     }
-}
 
-public struct NoRelaysSatisfyingConstraintsError: LocalizedError {
-    public var errorDescription: String? {
-        "No relays satisfying constraints."
-    }
-}
+    private static func parseRawPortRanges(_ rawPortRanges: [[UInt16]]) -> [ClosedRange<UInt16>] {
+        rawPortRanges.compactMap { inputRange -> ClosedRange<UInt16>? in
+            guard inputRange.count == 2 else { return nil }
 
-public struct RelaySelectorResult: Codable, Equatable {
-    public var endpoint: MullvadEndpoint
-    public var relay: REST.ServerRelay
-    public var location: Location
-}
+            let startPort = inputRange[0]
+            let endPort = inputRange[1]
 
-struct RelayWithLocation<T: AnyRelay> {
-    let relay: T
-    let serverLocation: Location
-
-    func matches(location: RelayLocation) -> Bool {
-        return switch location {
-        case let .country(countryCode):
-            serverLocation.countryCode == countryCode
-
-        case let .city(countryCode, cityCode):
-            serverLocation.countryCode == countryCode &&
-                serverLocation.cityCode == cityCode
-
-        case let .hostname(countryCode, cityCode, hostname):
-            serverLocation.countryCode == countryCode &&
-                serverLocation.cityCode == cityCode &&
-                relay.hostname == hostname
+            if startPort <= endPort {
+                return startPort ... endPort
+            } else {
+                return nil
+            }
         }
     }
-}
 
-private struct RelayWithDistance<T: AnyRelay> {
-    let relay: T
-    let distance: Double
+    /// Produce a list of `RelayWithLocation` items satisfying the given constraints
+    static func applyConstraints<T: AnyRelay>(
+        _ relayConstraint: RelayConstraint<UserSelectedRelays>,
+        portConstraint: RelayConstraint<UInt16>,
+        filterConstraint: RelayConstraint<RelayFilter>,
+        relays: [RelayWithLocation<T>]
+    ) -> [RelayWithLocation<T>] {
+        // Filter on active status, filter, and location.
+        let filteredRelays = relays.filter { relayWithLocation -> Bool in
+            guard relayWithLocation.relay.active else {
+                return false
+            }
+
+            switch filterConstraint {
+            case .any:
+                break
+            case let .only(filter):
+                if !relayMatchesFilter(relayWithLocation.relay, filter: filter) {
+                    return false
+                }
+            }
+
+            return switch relayConstraint {
+            case .any:
+                true
+            case let .only(relayConstraint):
+                // At least one location must match the relay under test.
+                relayConstraint.locations.contains { location in
+                    relayWithLocation.matches(location: location)
+                }
+            }
+        }
+
+        // Filter on country inclusion.
+        let includeInCountryFilteredRelays = filteredRelays.filter { relayWithLocation in
+            return switch relayConstraint {
+            case .any:
+                true
+            case let .only(relayConstraint):
+                relayConstraint.locations.contains { location in
+                    if case .country = location {
+                        return relayWithLocation.relay.includeInCountry
+                    }
+                    return false
+                }
+            }
+        }
+
+        // If no relays should be included in the matched country, instead accept all.
+        if includeInCountryFilteredRelays.isEmpty {
+            return filteredRelays
+        } else {
+            return includeInCountryFilteredRelays
+        }
+    }
+
+    /// Produce a port that is either user provided or randomly selected, satisfying the given constraints.
+    static func applyPortConstraint(
+        _ portConstraint: RelayConstraint<UInt16>,
+        rawPortRanges: [[UInt16]],
+        numberOfFailedAttempts: UInt
+    ) -> UInt16? {
+        switch portConstraint {
+        case let .only(port):
+            return port
+
+        case .any:
+            // 1. First two attempts should pick a random port.
+            // 2. The next two should pick port 53.
+            // 3. Repeat steps 1 and 2.
+            let useDefaultPort = (numberOfFailedAttempts % 4 == 2) || (numberOfFailedAttempts % 4 == 3)
+
+            return useDefaultPort ? defaultPort : pickRandomPort(rawPortRanges: rawPortRanges)
+        }
+    }
 }
--- a/ios/MullvadREST/Relay/RelaySelectorResult.swift
+++ b/ios/MullvadREST/Relay/RelaySelectorResult.swift
@@ -1,0 +1,18 @@
+//
+//  RelaySelectorResult.swift
+//  MullvadREST
+//
+//  Created by Mojgan on 2024-05-14.
+//  Copyright © 2024 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+import MullvadTypes
+
+public typealias RelaySelectorResult = RelaySelectorMatch
+
+public struct RelaySelectorMatch: Codable, Equatable {
+    public var endpoint: MullvadEndpoint
+    public var relay: REST.ServerRelay
+    public var location: Location
+}
--- a/ios/MullvadREST/Relay/RelayWithDistance.swift
+++ b/ios/MullvadREST/Relay/RelayWithDistance.swift
@@ -1,0 +1,13 @@
+//
+//  RelayWithDistance.swift
+//  MullvadREST
+//
+//  Created by Mojgan on 2024-05-17.
+//  Copyright © 2024 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+struct RelayWithDistance<T: AnyRelay> {
+    let relay: T
+    let distance: Double
+}
--- a/ios/MullvadREST/Relay/RelayWithLocation.swift
+++ b/ios/MullvadREST/Relay/RelayWithLocation.swift
@@ -1,0 +1,31 @@
+//
+//  RelayWithLocation.swift
+//  MullvadREST
+//
+//  Created by Mojgan on 2024-05-17.
+//  Copyright © 2024 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+import MullvadTypes
+
+struct RelayWithLocation<T: AnyRelay> {
+    let relay: T
+    let serverLocation: Location
+
+    func matches(location: RelayLocation) -> Bool {
+        return switch location {
+        case let .country(countryCode):
+            serverLocation.countryCode == countryCode
+
+        case let .city(countryCode, cityCode):
+            serverLocation.countryCode == countryCode &&
+                serverLocation.cityCode == cityCode
+
+        case let .hostname(countryCode, cityCode, hostname):
+            serverLocation.countryCode == countryCode &&
+                serverLocation.cityCode == cityCode &&
+                relay.hostname == hostname
+        }
+    }
+}
--- a/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksConfigurationCache.swift
+++ b/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksConfigurationCache.swift
@@ -9,8 +9,14 @@
 import Foundation
 import MullvadTypes
 
+public protocol ShadowsocksConfigurationCacheProtocol {
+    func read() throws -> ShadowsocksConfiguration
+    func write(_ configuration: ShadowsocksConfiguration) throws
+    func clear() throws
+}
+
 /// Holds a shadowsocks configuration object backed by a caching mechanism shared across processes
-public final class ShadowsocksConfigurationCache {
+public final class ShadowsocksConfigurationCache: ShadowsocksConfigurationCacheProtocol {
     private let configurationLock = NSLock()
     private var cachedConfiguration: ShadowsocksConfiguration?
     private let fileCache: FileCache<ShadowsocksConfiguration>
@@ -43,5 +49,13 @@ public final class ShadowsocksConfigurationCache {
 
         cachedConfiguration = configuration
         try fileCache.write(configuration)
+    }
+
+    /// Clear cached configuration.
+    public func clear() throws {
+        configurationLock.lock()
+        defer { configurationLock.unlock() }
+        cachedConfiguration = nil
+        try fileCache.clear()
     }
 }
--- a/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift
+++ b/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift
@@ -7,61 +7,63 @@
 //
 
 import Foundation
+import MullvadSettings
 import MullvadTypes
 
 public protocol ShadowsocksLoaderProtocol {
     func load() throws -> ShadowsocksConfiguration
-    func reloadConfiguration() throws
+    func clear() throws
 }
 
 public class ShadowsocksLoader: ShadowsocksLoaderProtocol {
-    private let shadowsocksCache: ShadowsocksConfigurationCache
-    private let relayCache: RelayCacheProtocol
+    let cache: ShadowsocksConfigurationCacheProtocol
+    let relaySelector: ShadowsocksRelaySelectorProtocol
+    let constraintsUpdater: RelayConstraintsUpdater
+
     private var relayConstraints = RelayConstraints()
-    private let constraintsUpdater: RelayConstraintsUpdater
 
     public init(
-        shadowsocksCache: ShadowsocksConfigurationCache,
-        relayCache: RelayCacheProtocol,
+        cache: ShadowsocksConfigurationCacheProtocol,
+        relaySelector: ShadowsocksRelaySelectorProtocol,
         constraintsUpdater: RelayConstraintsUpdater
     ) {
-        self.shadowsocksCache = shadowsocksCache
-        self.relayCache = relayCache
+        self.cache = cache
+        self.relaySelector = relaySelector
         self.constraintsUpdater = constraintsUpdater
+
+        // The constraints gets updated a lot when observing the tunnel, avoid clearing the cache if the constraints haven't changed.
         constraintsUpdater.onNewConstraints = { [weak self] newConstraints in
-            self?.relayConstraints = newConstraints
+            if self?.relayConstraints != newConstraints {
+                self?.relayConstraints = newConstraints
+                try? self?.clear()
+            }
         }
     }
 
-    public func reloadConfiguration() throws {
-        let newConfiguration = try create()
-        try shadowsocksCache.write(newConfiguration)
+    public func clear() throws {
+        try self.cache.clear()
     }
 
     /// Returns the last used shadowsocks configuration, otherwise a new randomized configuration.
     public func load() throws -> ShadowsocksConfiguration {
         do {
             // If a previous shadowsocks configuration was in cache, return it directly.
-            return try shadowsocksCache.read()
+            return try cache.read()
         } catch {
             // There is no previous configuration either if this is the first time this code ran
-            // Or because the previous shadowsocks configuration was invalid, therefore generate a new one.
             let newConfiguration = try create()
-            try shadowsocksCache.write(newConfiguration)
+            try cache.write(newConfiguration)
             return newConfiguration
         }
     }
 
     /// Returns a randomly selected shadowsocks configuration.
     private func create() throws -> ShadowsocksConfiguration {
-        let cachedRelays = try relayCache.read()
-        let bridgeConfiguration = RelaySelector.shadowsocksTCPBridge(from: cachedRelays.relays)
-        let closestRelay = RelaySelector.closestShadowsocksRelayConstrained(
-            by: relayConstraints,
-            in: cachedRelays.relays
-        )
+        let bridgeConfiguration = try relaySelector.getBridges()
+        let closestRelay = try relaySelector.selectRelay(with: relayConstraints)
 
-        guard let bridgeAddress = closestRelay?.ipv4AddrIn, let bridgeConfiguration else { throw POSIXError(.ENOENT) }
+        guard let bridgeAddress = closestRelay?.ipv4AddrIn,
+              let bridgeConfiguration else { throw POSIXError(.ENOENT) }
 
         return ShadowsocksConfiguration(
             address: .ipv4(bridgeAddress),
--- a/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksRelaySelector.swift
+++ b/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksRelaySelector.swift
@@ -1,0 +1,71 @@
+//
+//  ShadowsocksRelaySelector.swift
+//  MullvadREST
+//
+//  Created by Mojgan on 2024-05-23.
+//  Copyright © 2024 Mullvad VPN AB. All rights reserved.
+//
+
+import Foundation
+import MullvadSettings
+import MullvadTypes
+
+public protocol ShadowsocksRelaySelectorProtocol {
+    func selectRelay(
+        with constraints: RelayConstraints
+    ) throws -> REST.BridgeRelay?
+
+    func getBridges() throws -> REST.ServerShadowsocks?
+}
+
+final public class ShadowsocksRelaySelector: ShadowsocksRelaySelectorProtocol {
+    let relayCache: RelayCacheProtocol
+    let multihopUpdater: MultihopUpdater
+    private var multihopState: MultihopState
+    private var observer: MultihopObserverBlock!
+
+    deinit {
+        self.multihopUpdater.removeObserver(observer)
+    }
+
+    public init(
+        relayCache: RelayCacheProtocol,
+        multihopUpdater: MultihopUpdater,
+        multihopState: MultihopState
+    ) {
+        self.relayCache = relayCache
+        self.multihopUpdater = multihopUpdater
+        self.multihopState = multihopState
+        self.addObserver()
+    }
+
+    private func addObserver() {
+        self.observer = MultihopObserverBlock(didUpdateMultihop: { [weak self] _, multihopState in
+            self?.multihopState = multihopState
+        })
+        multihopUpdater.addObserver(observer)
+    }
+
+    public func selectRelay(
+        with constraints: RelayConstraints
+    ) throws -> REST.BridgeRelay? {
+        let cachedRelays = try relayCache.read().relays
+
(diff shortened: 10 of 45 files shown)


mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: 0 of 20 files reviewed, 1 unresolved discussion

___
*[`ios/MullvadREST/Relay/RelaySelectorResult.swift` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Nxwjshd0k1KQyIfNHIn:-Nxwjshd0k1KQyIfNHIo:b-86fus1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadREST/Relay/RelaySelectorResult.swift#L12)):*
> ```Swift
> import MullvadTypes
> 
> public typealias RelaySelectorResult = RelaySelectorMatch
> ```

when tweaking `RelaySelector` to work with entry and exit is done then this type would be something like below :



_Code snippet:_
```Swift
struct RelaySelectorResult {
    let entryRelay : RelaySelectorMatch?
    let exitRelay : RelaySelectorMatch
}
```

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: 0 of 20 files reviewed, 2 unresolved discussions

___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 45 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-cgnd5lji8JJKXmKt:-Ny-cgnd5lji8JJKXmKu:bj43jmm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadREST/Relay/RelaySelector.swift#L45)):*
> ```Swift
>             let mappedBridges = mapRelays(relays: relaysResponse.bridge.relays, locations: relaysResponse.locations)
>             let filteredRelays = applyConstraint(
>                 constraints.entryLocations ?? constraints.exitLocations,
> ```

"When selecting a bridge for API communication, the relay selector should honor if multi hop is enabled or not - if it is, the entry constraints should be used to pick a bridge for API access, otherwise the exit constraints should be used to pick a bridge for API access."

https://linear.app/mullvad/issue/IOS-598/allow-relay-selector-to-select-an-entry-peer

rablador (changes requested):
> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 4 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 114 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-hbXLDRErYtKbfGXF:-Ny-hbXLDRErYtKbfGXG:bfemf30) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadREST/Relay/RelaySelector.swift#L114)):*
> ```Swift
> 
>         /**
>          Filters relay list using given constraints and selects random relay.
> ```

Move this comment to the public `evaluate` function above.
___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 282 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-mF979aGWmt3hSo5g:-Ny-mF979aGWmt3hSo5h:b-o4j7mi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadREST/Relay/RelaySelector.swift#L282)):*
> ```Swift
> 
>     /// Produce a list of `RelayWithLocation` items satisfying the given constraints
>     private static func applyConstraint<T: AnyRelay>(
> ```

`applyConstraints` seems more appropriate.
___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 338 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-mnmd7dEIYxf2m6Fz:-Ny-mnmd7dEIYxf2m6G-:b-nsaxro) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadREST/Relay/RelaySelector.swift#L338)):*
> ```Swift
> 
>     /// Produce a port that is either user provided or randomly selected, satisfying the given constraints.
>     private static func applyConstraint(
> ```

I think we should call this method `applyPortConstraint` or something similar to distinguish it from the similarly named function above.
___
*[`ios/MullvadVPNTests/MullvadREST/Relay/RelaySelectorTests.swift` line 62 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-vUgS34yLkpL0FYU6:-Ny-vUgS34yLkpL0FYU7:b-ijxw45) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b09ebf93a127f84573257f163dea612c193d678d/ios/MullvadVPNTests/MullvadREST/Relay/RelaySelectorTests.swift#L62)):*
> ```Swift
>     }
> 
>     func testMultipleLocationsConstraint() throws {
> ```

Why was this test removed? We should still test constraints consisting of multiple relays (ie custom lists).

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 5 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadTypes/RelayConstraints.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-zBYk6HUUQ--G8G8U:-Ny-zBYk6HUUQ--G8G8V:b26a5p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L41)):*
> ```Swift
> 
>     public init(
>         entryLocations: RelayConstraint<UserSelectedRelays>? = nil,
> ```

"Sweden" should be default for entry as well.

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 5 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadTypes/RelayConstraints.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-zBYk6HUUQ--G8G8U:-Ny00Jjm5_8Wn7binCpH:bg52o71) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L41)):*

"Sweden" should be default for entry as well.
</blockquote></details>

I don't think entry should be optional. Instead, depending on if multihop is enabled or not we can let `RelaySelector` return a `RelaySelectorResult` where entry is nil (like your proposal where `RelaySelectorResult` contains entry and exit variables of type `RelaySelectorMatch?`), which is then passed to wireguard config later down the line.

Other reviewers, please chime in. :)

buggmagnet (changes requested):
> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 8 unresolved discussions (waiting on @mojganii and @rablador)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny0IP8y-rVXd1mVZZeT:-Ny0IP8y-rVXd1mVZZeU:bcm9kmm) (no related file):*
Please add tests in `MigrationManagerTests` to migrate for the new version added, but also for the missing migrations (`v3` and `v4`)

___
___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 90 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny0F89yAKiQkuH07phD:-Ny0F89yAKiQkuH07phE:b-jgms3p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadREST/Relay/RelaySelector.swift#L90)):*
> ```Swift
>     }
> 
>     public enum WireGuard {
> ```

What was the reason to add this and the `Shadowsocks` namespaces ?
___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 338 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-mnmd7dEIYxf2m6Fz:-Ny0F-6EDVpa8GF4Ln9E:buhcg4l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadREST/Relay/RelaySelector.swift#L338)):*

I think we should call this method `applyPortConstraint` or something similar to distinguish it from the similarly named function above.
</blockquote></details>

Totally agree, good catch !
___
*[`ios/MullvadTypes/RelayConstraints.swift` line 23 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny0GHxn9rUwhOxlwHfH:-Ny0GHxn9rUwhOxlwHfI:b-15bvfx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L23)):*
> ```Swift
> }
> 
> public struct RelayConstraints: Codable, Equatable {
> ```

I think we should leave the `CustomDebugStringConvertible` conformance, it makes debugging a much nicer experience.
___
*[`ios/MullvadTypes/RelayConstraints.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-zBYk6HUUQ--G8G8U:-Ny0GlzG4GGAJ_MuOT2F:b-j4g386) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L41)):*

I don't think entry should be optional. Instead, depending on if multihop is enabled or not we can let `RelaySelector` return a `RelaySelectorResult` where entry is nil (like your proposal where `RelaySelectorResult` contains entry and exit variables of type `RelaySelectorMatch?`), which is then passed to wireguard config later down the line.

Other reviewers, please chime in. :)
</blockquote></details>

Agreed, especially if we have a default entry, it doesn't make sense to make the entry point optional. 
___
*[`ios/MullvadVPNTests/MullvadREST/Relay/RelaySelectorTests.swift` line 62 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-vUgS34yLkpL0FYU6:-Ny0Hp7u0nayeIqmWRB3:b32v57) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b09ebf93a127f84573257f163dea612c193d678d/ios/MullvadVPNTests/MullvadREST/Relay/RelaySelectorTests.swift#L62)):*

Why was this test removed? We should still test constraints consisting of multiple relays (ie custom lists).
</blockquote></details>

+1

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: 11 of 20 files reviewed, 8 unresolved discussions (waiting on @buggmagnet and @rablador)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny0IP8y-rVXd1mVZZeT:-Ny1-44M5UjcgjbfHY_T:b-896fix) (no related file):*

Please add tests in `MigrationManagerTests` to migrate for the new version added, but also for the missing migrations (`v3` and `v4`)
</blockquote></details>

Done.

___
___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 90 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny0F89yAKiQkuH07phD:-Ny0PAJi3-CjqA6IXByI:b-35g6ul) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadREST/Relay/RelaySelector.swift#L90)):*

What was the reason to add this and the `Shadowsocks` namespaces ?
</blockquote></details>

functions tracking were getting hard, I tried to separate them into different namespaces.
___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 114 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-hbXLDRErYtKbfGXF:-Ny0QPDc45wXtHahN5tN:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadREST/Relay/RelaySelector.swift#L114)):*

Move this comment to the public `evaluate` function above.
</blockquote></details>

Done.
___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 282 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-mF979aGWmt3hSo5g:-Ny0QhbVArYwDUN6pQHz:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadREST/Relay/RelaySelector.swift#L282)):*

`applyConstraints` seems more appropriate.
</blockquote></details>

Done.
___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 338 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-mnmd7dEIYxf2m6Fz:-Ny0Qib45b1fG1LZwS2J:b-jnts5i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadREST/Relay/RelaySelector.swift#L338)):*

Totally agree, good catch !
</blockquote></details>

fair enough!
___
*[`ios/MullvadTypes/RelayConstraints.swift` line 23 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny0GHxn9rUwhOxlwHfH:-Ny0V8Py-oVaFFQEW2v3:bvm5b5u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L23)):*

I think we should leave the `CustomDebugStringConvertible` conformance, it makes debugging a much nicer experience.
</blockquote></details>

having this protocol wouldn't give a trunstable log when there is no acknowledge  here if multi-hop is enabled or not
___
*[`ios/MullvadTypes/RelayConstraints.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-zBYk6HUUQ--G8G8U:-Ny0lGPSCH08KFQHJG6s:bxzz458) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L41)):*

Agreed, especially if we have a default entry, it doesn't make sense to make the entry point optional.
</blockquote></details>

Making this element required led to having different functions for selecting relays depending on whether multi-hop is enabled or not in the upper layers. The relaySelector should not and does not have any knowledge of whether multi-hop is active or not.
___
*[`ios/MullvadVPNTests/MullvadREST/Relay/RelaySelectorTests.swift` line 62 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-vUgS34yLkpL0FYU6:-Ny0oa2u2n7YzJ0Uy6nf:b-85ccb4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b09ebf93a127f84573257f163dea612c193d678d/ios/MullvadVPNTests/MullvadREST/Relay/RelaySelectorTests.swift#L62)):*

+1
</blockquote></details>

sorry, It was unintentional.

rablador (changes requested):
> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 408 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny4baKl20-8jRos1DX0:-Ny4baKl20-8jRos1DX1:bwegzn3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cf584c45e3681afde5e4402224a5679c42df3b74/ios/MullvadREST/Relay/RelaySelector.swift#L408)):*
> ```Swift
> }
> 
> // swiftlint:disable:this file_length
> ```

I think we should try to break out a few methods rather than allowing the file to grow. Could we perhaps break out the namespaced things? (shadowsocks, wireguard)

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: 18 of 27 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 49 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny4eKkA-rtK50-lj-sD:-Ny4eKkA-rtK50-lj-sE:b-5jjqbz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68703cf88d7ef010fd3e3c1452d2571883a3a60b/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L49)):*
> ```Swift
>     /// Returns the last used shadowsocks configuration, otherwise a new randomized configuration.
>     public func load() throws -> ShadowsocksConfiguration {
>         do {
> ```

@buggmagnet  I don't know  what it's the best place to inform shadow socks to choose `entryLocations` or `exitLocations` when settings gets updated. `shadowSocksConfiguration` should update itself upon the multi hop state changes.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: 16 of 31 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadREST/Relay/RelaySelector.swift` line 408 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny4baKl20-8jRos1DX0:-Ny4wZwa8qQoD9kJKdsx:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cf584c45e3681afde5e4402224a5679c42df3b74/ios/MullvadREST/Relay/RelaySelector.swift#L408)):*

I think we should try to break out a few methods rather than allowing the file to grow. Could we perhaps break out the namespaced things? (shadowsocks, wireguard)
</blockquote></details>

Done.

acb-mv (changes requested):
> Reviewed 8 of 9 files at r3, 6 of 6 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 7 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadREST/Relay/RelaySelector+Wiregaurd.swift` line 2 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny5XZOy-s0Njajxxo0o:-Ny5XZOy-s0Njajxxo0p:bqdcmwg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/98ce0dc7a4891dc1e8947c6af20e87606d170381/ios/MullvadREST/Relay/RelaySelector%2BWiregaurd.swift#L2)):*
> ```Swift
> //
> //  RelaySelector+Wiregaurd.swift
> ```

nit: "Wireguard" is misspelled
___
*[`ios/PacketTunnel/PacketTunnelProvider/RelaySelectorWrapper.swift` line 16 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny5XyW47vNDXG3dDEzN:-Ny5XyW47vNDXG3dDEzO:b2w87zd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/98ce0dc7a4891dc1e8947c6af20e87606d170381/ios/PacketTunnel/PacketTunnelProvider/RelaySelectorWrapper.swift#L16)):*
> ```Swift
> struct MultihopNotImplementedError: LocalizedError {
>     public var errorDescription: String? {
>         "The picking relays for Multihop is not implemented yet."
> ```

Maybe lose the "the"?

rablador (approved):
> Reviewed 7 of 9 files at r3, 4 of 6 files at r4, 1 of 1 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @mojganii)

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 6 unresolved discussions (waiting on @buggmagnet and @mojganii)

buggmagnet (changes requested):
> Reviewed 5 of 9 files at r2, 3 of 9 files at r3, 3 of 6 files at r4, 1 of 1 files at r5, 1 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 7 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 26 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyOsY3uEdCRyUJbcpLg:-NyOsY3uEdCRyUJbcpLh:bny35u9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6414a1fadc9580304110b985c89af36cc38dfff3/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L26)):*
The immediate problem with this approach is that now `ShadowsocksLoader` has to keep tabs on the settings,
and read from them whenever it needs them, which is prone to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issues.

If it were the only component that had this dependency, it could work. But as more and more components rely on `SettingsManager`, we can conclude that this approach does not scale well, and that we should do something different here.

Now, what I don't understand is why do we have a singled out `location` here, when the `relayConstraints` (which should automatically get updated via `RelayConstraintsUpdater` already has the entry and exit locations chosen ?

To reply to your other comment here: 
> I don't know what it's the best place to inform shadow socks to choose `entryLocations` or `exitLocations` when settings gets updated. `shadowSocksConfiguration` should update itself upon the multi hop state changes.

IMHO `ShadowsocksLoader` doesn't, and shouldn't care about multi-hop settings.
It looks to me that all the objects that depend on `RelaySelector` should instead depend on a new layer that sits on top of the relayselector, and that will decide on how to do multi-hop selection.

The reason is that the relay selector itself shouldn't care about multi-hop, but we still need to have the responsibility somewhere.
This approach lets us plug that component everywhere, without changing all the call sites (too much). 
___
*[`ios/MullvadTypes/RelayConstraints.swift` line 23 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny0GHxn9rUwhOxlwHfH:-NyOvWz8Bb9VEfnz1Tw2:b-ul1wen) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L23)):*

having this protocol wouldn't give a trunstable log when there is no acknowledge  here if multi-hop is enabled or not
</blockquote></details>

It's not just about multihop, this is also used when multi hop is not enabled, and that debug information is still very valuable.
___
*[`ios/MullvadTypes/RelayConstraints.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-zBYk6HUUQ--G8G8U:-NyOxER40jT6cjnmKdZ9:bxhsx88) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L41)):*

Making this element required led to having different functions for selecting relays depending on whether multi-hop is enabled or not in the upper layers. The relaySelector should not and does not have any knowledge of whether multi-hop is active or not.
</blockquote></details>

In this case, we should leave the `RelayConstraints` to have a single `location` in it. 
Then we create 2 typealiases `EntryConstraints` and `ExitConstraints`, which both map to `RelayConstraints` 
We let the `RelaySelector` work with `RelayConstraints`, and where we need to handle both entry and exit constraints, we already have those types. 

And we can have custom logic that sits on top of the `RelaySelector` that accepts both `EntryConstraints` and `ExitConstraints`

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 7 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @mojganii)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 49 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny4eKkA-rtK50-lj-sD:-NyZQ_snAts6foy55ebt:bjqkx6n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68703cf88d7ef010fd3e3c1452d2571883a3a60b/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L49)):*

@buggmagnet  I don't know  what it's the best place to inform shadow socks to choose `entryLocations` or `exitLocations` when settings gets updated. `shadowSocksConfiguration` should update itself upon the multi hop state changes.
</blockquote></details>

I believe this comment is not applicable anymore, right?
___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 26 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyOsY3uEdCRyUJbcpLg:-NyZTyfK1ohcqYyNYBM2:bk49l40) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6414a1fadc9580304110b985c89af36cc38dfff3/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L26)):*

I like the suggestion. what do you think to take this part of responsibility to another object like below :
</blockquote></details>

Yes, something like this I think. There should be no `evaluateForSingleHop` or similar methods on the relay selector class though. Instead we should add a param to `evaluate` for excluding a certain relay. That way relay selector knows nothing about multiphop as a concept.

Could this part be added to the next PR instead? I feel that this PR should only be concerned about settings changes and not the relay selector.
___
*[`ios/MullvadTypes/RelayConstraints.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-zBYk6HUUQ--G8G8U:-NyZRCCK3sXUZWCJdj0v:b-emykol) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L41)):*

we had discussed regarding the structure several weeks ago but it doesn't mean there is no room for improvement then before making any changes I would like to know others opinion.additionally, there are several points which need to attention according your suggestion if I didn't misunderstood  :

1- firstly, the settings should store 2 fields, one for entry and another for exit which both are `RelayConstraints` datatye

2- additionally, we need to update `port` and `filter` twice when any changes come up


3- lastly, do we need to apply this changes?
</blockquote></details>

I don't think we should make this change. I don't see much benefit, primary reasons being:

1. The need to update filter and port twice every time as you mention.
2. In singlehop port is only applicable for exit, in multihop it's the opposite. So in multihop mode the exit constraint would have a port attributre that is never supposed to be used.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 7 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @rablador)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 49 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny4eKkA-rtK50-lj-sD:-Ny_TnLV7d7IRhlctAo4:b-2p2upl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68703cf88d7ef010fd3e3c1452d2571883a3a60b/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L49)):*

I believe this comment is not applicable anymore, right?
</blockquote></details>

No,It's still there.when user enables `mutihop`,  then the shadowsocks should pick up  the closest one to entry which it means if the previous one is not the closest to entry, it should pick up  new one from `relaySelector` and replace that the . `load` method is fetching the last successful configuration from cache which probably is not valid anymore after changing the status of `multhop`.I believe caching last used `shadowsocks` configuration doesn't make any point. 
___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 26 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyOsY3uEdCRyUJbcpLg:-Ny_YPn59TqqFBTWlIL7:bkq46y0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6414a1fadc9580304110b985c89af36cc38dfff3/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L26)):*

Yes, something like this I think. There should be no `evaluateForSingleHop` or similar methods on the relay selector class though. Instead we should add a param to `evaluate` for excluding a certain relay. That way relay selector knows nothing about multiphop as a concept.

Could this part be added to the next PR instead? I feel that this PR should only be concerned about settings changes and not the relay selector.
</blockquote></details>

I agree this PR has nothing to do with how `RelaySelector` must choose the relays
___
*[`ios/MullvadTypes/RelayConstraints.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-zBYk6HUUQ--G8G8U:-Ny_TgJT3D39Y011Rltj:b32v57) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L41)):*

I don't think we should make this change. I don't see much benefit, primary reasons being:

1. The need to update filter and port twice every time as you mention.
2. In singlehop port is only applicable for exit, in multihop it's the opposite. So in multihop mode the exit constraint would have a port attributre that is never supposed to be used.
</blockquote></details>

+1

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: 15 of 33 files reviewed, 7 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @rablador)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 49 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny4eKkA-rtK50-lj-sD:-Ny_ePQ4749KbOEv1TFA:be2i650) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68703cf88d7ef010fd3e3c1452d2571883a3a60b/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L49)):*

No,It's still there.when user enables `mutihop`,  then the shadowsocks should pick up  the closest one to entry which it means if the previous one is not the closest to entry, it should pick up  new one from `relaySelector` and replace that the . `load` method is fetching the last successful configuration from cache which probably is not valid anymore after changing the status of `multhop`.I believe caching last used `shadowsocks` configuration doesn't make any point.
</blockquote></details>

or we need to invalidate cache when the value is changed

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: 15 of 33 files reviewed, 7 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @rablador)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 49 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny4eKkA-rtK50-lj-sD:-Nye9XPr6kLq9pz7fuQw:bysi6xy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/68703cf88d7ef010fd3e3c1452d2571883a3a60b/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L49)):*

or we need to invalidate cache when the value is changed
</blockquote></details>

this scenario goes for updating `RelayConstraints`, either. the bridge should be changed when it's updated

rablador (changes requested):
> Reviewed 18 of 18 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 7 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @mojganii)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 76 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyeDbMZDc8q_H9Qrub1:-NyeDbMZDc8q_H9Qrub2:b97242t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/efb0b3a72bb498d4d8aa66a39ebfe8b067626ac5/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L76)):*
> ```Swift
>     }
> 
>     private func invalidate() throws {
> ```

There's already a function for this: `reloadConfiguration() `

rablador (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-:-NyegI_L2tHjDYLKZVxp:bnfp4nl)* status: all files reviewed, 7 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @mojganii)

acb-mv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: 24 of 36 files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksRelaySelector.swift` line 57 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyuKq9xC0vSkMZm3bUQ:-NyuKq9xC0vSkMZm3bUR:b8664) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/485a40989253f526802676c9e077eee93ccd9828/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksRelaySelector.swift#L57)):*
> ```Swift
>                 in: cachedRelays
>             )
>         }
> ```

Given that the only parameter to the call that changes is `location`, this could be simplified to something like



_Code snippet:_
```Swift
let locationConstraint = switch multihopState {
    case .on:  constraints.entryLocations
    case .off: constraints.exitLocations
}
return RelaySelector.Shadowsocks.closestRelay(
                location: locationConstraint,
                port: constraints.port,
                filter: constraints.filter,
                in: cachedRelays
            )
```

rablador (changes requested):
> Reviewed 12 of 12 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 9 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @mojganii)

___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 37 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyuLpfS9A-ebxjnqkTt:-NyuLpfS9A-ebxjnqkTu:bjpnsz9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/485a40989253f526802676c9e077eee93ccd9828/ios/PacketTunnelCore/Actor/ObservedState.swift#L37)):*
> ```Swift
>     public var lastKeyRotation: Date?
>     public let isPostQuantum: Bool
>     public let isMultihop: Bool
> ```

I don't think the actor or state/observed state need to know about multihop. What would be the reason for it?

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 9 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @rablador)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 76 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyeDbMZDc8q_H9Qrub1:-Nyuekqh-NsEDZmKkXUd:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/efb0b3a72bb498d4d8aa66a39ebfe8b067626ac5/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L76)):*

There's already a function for this: `reloadConfiguration() `
</blockquote></details>

Done.
___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksRelaySelector.swift` line 57 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyuKq9xC0vSkMZm3bUQ:-Nyuczqa89OtHTMFW3Xf:b-wzidj8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/485a40989253f526802676c9e077eee93ccd9828/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksRelaySelector.swift#L57)):*

Given that the only parameter to the call that changes is `location`, this could be simplified to something like
</blockquote></details>

indeed. 
___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 37 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyuLpfS9A-ebxjnqkTt:-NyucPlV7GX3FbGpSrIJ:b-8tjzqf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/485a40989253f526802676c9e077eee93ccd9828/ios/PacketTunnelCore/Actor/ObservedState.swift#L37)):*

I don't think the actor or state/observed state need to know about multihop. What would be the reason for it?
</blockquote></details>

packet tunnel process is observing to the value for that regarding latest settings

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 8 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @mojganii)

___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 37 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyuLpfS9A-ebxjnqkTt:-NyyBKXN5mBM-Dv3-q2q:b2sxmrb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/485a40989253f526802676c9e077eee93ccd9828/ios/PacketTunnelCore/Actor/ObservedState.swift#L37)):*

packet tunnel process is observing to the value for that regarding latest settings
</blockquote></details>

There should be no reason for it to observe the multihop setting. The only thing the packet tunnel should care about regarding multihop are the relay constraints. Nil or value for entry, always value for exit.

buggmagnet (changes requested):
> Reviewed 1 of 6 files at r4, 6 of 18 files at r7, 6 of 12 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 10 unresolved discussions (waiting on @acb-mv, @mojganii, and @rablador)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 35 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Nyy0EuvCTx3tJRErbWs:-Nyy0EuvCTx3tJRErbWt:bx8e2ob) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d1124ecc964512eba478481ea7d6e6cf8bbd7bf0/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L35)):*
> ```Swift
> 
>         constraintsUpdater.onNewConstraints = { [weak self] newConstraints in
>             try? self?.reloadConfiguration()
> ```

We really should have tests for this class. 
Here we recreate the configuration before we get the new constraints, which should be done the other way,
get the new constraints, then create a new configuration. 
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 169 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Nyy8inG649U24w5cutF:-Nyy8inG649U24w5cutG:b-1etwxf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d1124ecc964512eba478481ea7d6e6cf8bbd7bf0/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L169)):*
> ```Swift
>         let shadowsocksRelaySelector = ShadowsocksRelaySelector(
>             relayCache: ipOverrideWrapper,
>             multihopStateUpdater: multihopUpdater
> ```

This design is problematic. 
Both `RelaySelectorWrapper` and `ShadowsocksRelaySelector` depend on the same instance of the `MultihopStateUpdater` 

And they both override its `onNewState` closure (meaning only one of them will ever get callbacks on new multihop state)

It looks like the design was inspired by the `RelayConstraintsUpdater`, which is fine, but it was a careful trade off.
In any case, this here won't work. 

If we want to keep this, we will need the `MultihopStateUpdater` to have multiple listeners.
And I insist, we **do not need** Combine for this
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 267 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyyCBeH-1ZahZr4XE9j:-NyyCBeH-1ZahZr4XE9k:b-qjlnkz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d1124ecc964512eba478481ea7d6e6cf8bbd7bf0/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L267)):*
> ```Swift
> 
>                     // Pass multi-hop state retrieved during the last read from setting into relay selector.
>                     multihopUpdater.onNewState?(connState.isMultihop ? .on : .off)
> ```

The `ConnectionData` structure should really use the `MultihopState` enum instead of being a boolean. 
Also it seems like this should logically be grouped with the new constraints propagation. 
___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 37 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyuLpfS9A-ebxjnqkTt:-NyyDoGs7nrLbc9qOkmk:brwd0nn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/485a40989253f526802676c9e077eee93ccd9828/ios/PacketTunnelCore/Actor/ObservedState.swift#L37)):*

There should be no reason for it to observe the multihop setting. The only thing the packet tunnel should care about regarding multihop are the relay constraints. Nil or value for entry, always value for exit.
</blockquote></details>

It's kinda messy at the moment, the packet tunnel actor will likely end up reading the multihop settings from the connection state when trying to make a connection instead of reading from the settings, at least, that's what I'm predicting on how it will be implemented.

That being said, we should make sure of the `MultihopState` enum here.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 8 unresolved discussions (waiting on @acb-mv, @mojganii, and @rablador)

___
*[`ios/MullvadTypes/RelayConstraints.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-zBYk6HUUQ--G8G8U:-NyyFQox2migNF1tZhOr:biidsev) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L41)):*

+1
</blockquote></details>

I'm confused about this comment now, @rablador we shouldn't make this change, but are making it anyway ?
Let's discuss that together.

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 8 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @mojganii)

___
*[`ios/MullvadTypes/RelayConstraints.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-zBYk6HUUQ--G8G8U:-NyyGXkt1XPBdNej6DAv:bb2ccj6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L41)):*

I'm confused about this comment now, @rablador we shouldn't make this change, but are making it anyway ?
Let's discuss that together.
</blockquote></details>

We are not making the proposed changes, ie we keep author's original suggestion.
___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 37 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyuLpfS9A-ebxjnqkTt:-NyyGI2N236CmBlEbC1N:blg53wk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/485a40989253f526802676c9e077eee93ccd9828/ios/PacketTunnelCore/Actor/ObservedState.swift#L37)):*

It's kinda messy at the moment, the packet tunnel actor will likely end up reading the multihop settings from the connection state when trying to make a connection instead of reading from the settings, at least, that's what I'm predicting on how it will be implemented.

That being said, we should make sure of the `MultihopState` enum here.
</blockquote></details>

Why? Didn't we agree that packet tunnel would know from the relay constraints what to do?

acb-mv (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 8 unresolved discussions (waiting on @buggmagnet)

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 8 unresolved discussions (waiting on @acb-mv and @mojganii)

___
*[`ios/MullvadTypes/RelayConstraints.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-zBYk6HUUQ--G8G8U:-Nyz7vOR2K-T_YfgHj-0:b-kf5nvk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L41)):*

We are not making the proposed changes, ie we keep author's original suggestion.
</blockquote></details>

This was discussed offline.

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 8 unresolved discussions (waiting on @acb-mv and @mojganii)

___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 267 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyyCBeH-1ZahZr4XE9j:-Nyz8MQ58KyZskibb-Hw:b-feq6q3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d1124ecc964512eba478481ea7d6e6cf8bbd7bf0/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L267)):*

The `ConnectionData` structure should really use the `MultihopState` enum instead of being a boolean. 
Also it seems like this should logically be grouped with the new constraints propagation.
</blockquote></details>

We discussed offline that the `ConnectionData` would not keep track of the multi hop settings. 
Because the packet tunnel actor already uses the relay selector via the `RelaySelectorWrapper`

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 8 unresolved discussions (waiting on @acb-mv, @mojganii, and @rablador)

___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 37 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyuLpfS9A-ebxjnqkTt:-Nyz8kcw5CUZRJvDdh-E:b-dwov5h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/485a40989253f526802676c9e077eee93ccd9828/ios/PacketTunnelCore/Actor/ObservedState.swift#L37)):*

Why? Didn't we agree that packet tunnel would know from the relay constraints what to do?
</blockquote></details>

We discussed offline, we will remove `isMultihop` from the `ObservedState`

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: 24 of 46 files reviewed, 7 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @rablador)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 35 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Nyy0EuvCTx3tJRErbWs:-Nz8F62b5nmQ9TnqbKs-:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d1124ecc964512eba478481ea7d6e6cf8bbd7bf0/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L35)):*

We really should have tests for this class. 
Here we recreate the configuration before we get the new constraints, which should be done the other way,
get the new constraints, then create a new configuration.
</blockquote></details>

Done.
___
*[`ios/MullvadTypes/RelayConstraints.swift` line 41 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Ny-zBYk6HUUQ--G8G8U:-Nz8F90b8VUYH-LzUHZw:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/27ea6e3b6cb547bd4276bc0b4bbfe64899ea0451/ios/MullvadTypes/RelayConstraints.swift#L41)):*

This was discussed offline.
</blockquote></details>

Done
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 169 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Nyy8inG649U24w5cutF:-NyyN-lOCdWkPkpY5m7e:bhjeemy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d1124ecc964512eba478481ea7d6e6cf8bbd7bf0/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L169)):*

I can't follow your suggestion. could you please elaborate?
</blockquote></details>

having a object which manages both changes?
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 267 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyyCBeH-1ZahZr4XE9j:-Nz8Ezm5CNKJaaQRFGLu:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d1124ecc964512eba478481ea7d6e6cf8bbd7bf0/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L267)):*

We discussed offline that the `ConnectionData` would not keep track of the multi hop settings. 
Because the packet tunnel actor already uses the relay selector via the `RelaySelectorWrapper`
</blockquote></details>

Done.
___
*[`ios/PacketTunnelCore/Actor/ObservedState.swift` line 37 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NyuLpfS9A-ebxjnqkTt:-Nz8F1p54hDkqGX_btkT:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/485a40989253f526802676c9e077eee93ccd9828/ios/PacketTunnelCore/Actor/ObservedState.swift#L37)):*

We discussed offline, we will remove `isMultihop` from the `ObservedState`
</blockquote></details>

Done

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 10 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @mojganii)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 36 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzSdXuIE3iLH759dKOn:-NzSmsjaBGQuMNFJpsvc:by2ds47) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L36)):*

why not call `try? self?.clear()` directly ?
</blockquote></details>

My thought too, but I figured that maybe clear might be updated at some point and then it would do other undesired things.
___
*[`ios/MullvadTypes/FileCache.swift` line 38 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzSduQ3DSbOmYA0TABr:-NzSn8XS6YkD_DK4QRAO:b32v57) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadTypes/FileCache.swift#L38)):*

This really should be doing `try FileManager.default.removeItem(at: fileURL)` instead
</blockquote></details>

+1

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: 41 of 46 files reviewed, 10 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @rablador)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 36 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzSdXuIE3iLH759dKOn:-NzSgXpOEt25OAGhgLq4:b-xkgx40) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L36)):*

why not call `try? self?.clear()` directly ?
</blockquote></details>

we have the same functionality which is exposed for the external calling and this block and that function should have done the same behaviour.
___
*[`ios/MullvadSettings/MultihopSettings.swift` line 43 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzS_7d9BQtEU5suosTf:-NzShbI06sVSClXXgHuQ:b-5adwp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadSettings/MultihopSettings.swift#L43)):*

We already have a class called `ObserverList` that does this functionality, we should reuse it instead of re-creating the same concept.
</blockquote></details>

the responsibility of the class is listening to the changes and propagate them to listeners but observer list is an array of observers which keeps their reference, In my opinion they have single/separate responsibility. If I couldn't follow your suggestion then elaborate more, please.
___
*[`ios/MullvadTypes/FileCache.swift` line 38 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzSduQ3DSbOmYA0TABr:-NzSixkI-cf3qvN1OKI4:b-g9zztm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadTypes/FileCache.swift#L38)):*

This really should be doing `try FileManager.default.removeItem(at: fileURL)` instead
</blockquote></details>

It can be but the deleting has more cost.I was waiting to know other's opinion to choose between clear content or delete file
___
*[`ios/MullvadVPNTests/MullvadREST/Shadowsocks/ShadowsocksLoaderTests.swift` line 64 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzSadjk1LQ6amjWUkhO:-NzSjKlg46oY3_CbVjSi:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadVPNTests/MullvadREST/Shadowsocks/ShadowsocksLoaderTests.swift#L64)):*

This test name is misleading. What this test is really doing is checking that changing constraints clears the cache.
Changing relay constraints doesn't reload the cached configuration anymore.

We should rename this `testConstraintsUpdateClearsCache` IMHO
</blockquote></details>

Done.
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 81 at r11](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-Nz8Vatu6vkmfrEQXHjc:-NzSdhFj7_0FUuCLkAO2:b-79jm9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/79c8762e74d5984b26aadd13bafd54531a7e27d2/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L81)):*

Nit: In `ShadowsocksRelaySelector` this is called `multihopUpdater`. I think we should settle on one definition.
</blockquote></details>

Right, It's slipped to be renamed.

rablador (approved):
> Reviewed 5 of 5 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 9 unresolved discussions (waiting on @acb-mv and @buggmagnet)

buggmagnet (changes requested):
> Reviewed 3 of 22 files at r10, 1 of 5 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 10 unresolved discussions (waiting on @acb-mv, @mojganii, and @rablador)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 36 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzSdXuIE3iLH759dKOn:-NzT3LHACHXNbXcgXc4s:b51gjox) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L36)):*

we have the same functionality which is exposed for the external calling and this block and that function should have done the same behaviour.
</blockquote></details>

I'm sorry I don't undertstand what you mean. 
We shouldn't have more than 1 way to `clear` the cache, whether it's internal or external.
So there's no reason to not do `try? self?.clear()` instead here.
___
*[`ios/MullvadTypes/FileCache.swift` line 38 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzSduQ3DSbOmYA0TABr:-NzSyKtvCftsEt_Sc42j:b-7dr2ur) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadTypes/FileCache.swift#L38)):*
> deleting has more cost
That's not correct. Without going too much into the details, "deleting a file" basically tells the system that the contents of that file can be overwritten later, and that the file can be marked as cleared, whereas "writing empty content" into a file needs the system to open the file, write the content, truncate it" etc...

In general, we shouln't make such assumptions without measuring first.
That being said,

We can't just clear the file like that, we still need to use the `FileCoordinator` to guarantee safe cross process access.

i.e what needs to be done here is : 
```swift
    public func clear() throws {
        let fileCoordinator = NSFileCoordinator(filePresenter: nil)
        try fileCoordinator.coordinate(writingItemAt: fileURL, options: [.forReplacing]) { fileURL in
            try FileManager.default.removeItem(at: fileURL)
        }
    }
```
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 176 at r13](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzT-E_Y2YNjmtzMaKu2:-NzT-E_Y2YNjmtzMaKu3:b-ea60np) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3bd67ef5460eccd8e87088d17de1c5f70f35bc20/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L176)):*
> ```Swift
>         let multihopState = try? settingsReader.read().multihopState
> 
>         multihopStateListener.onNewMultihop?(multihopState ?? .off)
> ```

Why are we doing this here ? That doesn't really make sense.
The listener updates should only happen when the user selects new relay constraints, if we're doing this so that `ShadowsocksRelaySelector` has the correct value, then we should dependency inject it when we create it instead of propagating fake updates.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 9 unresolved discussions (waiting on @acb-mv and @buggmagnet)

___
*[`ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift` line 36 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzSdXuIE3iLH759dKOn:-NzTJAfh80rj2oKz1oSh:bh9ar1f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadREST/Transport/Shadowsocks/ShadowsocksLoader.swift#L36)):*

I'm sorry I don't undertstand what you mean. 
We shouldn't have more than 1 way to `clear` the cache, whether it's internal or external.
So there's no reason to not do `try? self?.clear()` instead here.
</blockquote></details>

I made a mistake. yes, It should call `try? self?.clear()`. I mixed up
___
*[`ios/MullvadTypes/FileCache.swift` line 38 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzSduQ3DSbOmYA0TABr:-NzTLwsnCX6z9AQzMd3T:b-qqj6pk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadTypes/FileCache.swift#L38)):*

> deleting has more cost
That's not correct. Without going too much into the details, "deleting a file" basically tells the system that the contents of that file can be overwritten later, and that the file can be marked as cleared, whereas "writing empty content" into a file needs the system to open the file, write the content, truncate it" etc...

In general, we shouln't make such assumptions without measuring first.
That being said,

We can't just clear the file like that, we still need to use the `FileCoordinator` to guarantee safe cross process access.

i.e what needs to be done here is : 
```swift
    public func clear() throws {
        let fileCoordinator = NSFileCoordinator(filePresenter: nil)
        try fileCoordinator.coordinate(writingItemAt: fileURL, options: [.forReplacing]) { fileURL in
            try FileManager.default.removeItem(at: fileURL)
        }
    }
```
</blockquote></details>

shouldn't the writing option  be `forDeleting`?
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 176 at r13](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzT-E_Y2YNjmtzMaKu2:-NzTMCDv6atEE3ci1p7W:b4xhj1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3bd67ef5460eccd8e87088d17de1c5f70f35bc20/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L176)):*

Why are we doing this here ? That doesn't really make sense.
The listener updates should only happen when the user selects new relay constraints, if we're doing this so that `ShadowsocksRelaySelector` has the correct value, then we should dependency inject it when we create it instead of propagating fake updates.
</blockquote></details>

I did this before through DI, but I'll do that instead of faking update

buggmagnet (commented):
> Reviewed 1 of 3 files at r12.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 8 unresolved discussions (waiting on @acb-mv, @mojganii, and @rablador)

___
*[`ios/MullvadTypes/FileCache.swift` line 38 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzSduQ3DSbOmYA0TABr:-NzTPkq_6gm0GHIMqlMq:b-vgq5nt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadTypes/FileCache.swift#L38)):*

shouldn't the writing option  be `forDeleting`?
</blockquote></details>

Yes it should, good catch !

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 8 unresolved discussions (waiting on @acb-mv and @mojganii)

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: 36 of 52 files reviewed, 9 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @rablador)

___
*[`ios/MullvadSettings/MultihopSettings.swift` line 43 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzS_7d9BQtEU5suosTf:-NzUfzke1NMNnbx328Mn:b-z9qi7f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadSettings/MultihopSettings.swift#L43)):*

I didn't know or forgot that we have a class for this! The concept does look very similar to this implementation. If it's not too much work to use the existing one I think it might be a good idea to reuse it.
</blockquote></details>

done!
I've forgotten which we had this.
___
*[`ios/MullvadTypes/FileCache.swift` line 38 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzSduQ3DSbOmYA0TABr:-NzUgM6x4WRJ9rt0rzPZ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a86200c0880aec1d4abff095ba1350dacff851c0/ios/MullvadTypes/FileCache.swift#L38)):*

Yes it should, good catch !
</blockquote></details>

Done.
___
*[`ios/MullvadTypes/ObserverList.swift` line 65 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257#-NzUgb3T7M77u7Oz_23I:-NzUgb3T7M77u7Oz_23J:b-g5emj5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/884efe5e0ec1f3eea445aa1c9c9aabb1844e7325/ios/MullvadTypes/ObserverList.swift#L65)):*
> ```Swift
>     }
> 
>     public func notify(_ body: (T) -> Void) {
> ```

`forEach` was not well-known, I renamed that to `notify` for the clarification.

buggmagnet (approved):
> Reviewed 5 of 5 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6257)* status: all files reviewed, 4 unresolved discussions (waiting on @acb-mv)

PR #6279, opened 2024-05-24
Suggestion for spending less GitHub cloud minutes

niklasberglund:
A suggestion for how less cloud minutes could be spent by not running time consuming jobs for draft PRs. With these changes runs triggered by `pull_request` will only trigger when the PR is not a draft, and when a draft PR is ready for review.

Looked over job execution time and made sure jobs taking over 1 minute aren't triggered for draft PRs.
image

So when pushing commits to a draft PR branch actions triggered will look something like this:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6279)

### DIFF
--- a/.github/workflows/android-app.yml
+++ b/.github/workflows/android-app.yml
@@ -2,6 +2,7 @@
 name: Android - Build and test
 on:
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
     paths:
       - '**'
       - '!.github/workflows/**'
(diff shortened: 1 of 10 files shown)


niklasberglund:
Closing this for now. Many workflows use free runner images, and the solution to spending many cloud minutes when iterating on draft PRs could be to just be mindful and try to avoid it.

PR #6280, opened 2024-05-24
Refactor PacketTunnelActor to use a Reducer

acb-mv:
This is a refactoring of the PacketTunnelActor state machine, which splits the state processing logic into two stages: a stateless function (a Reducer) which takes incoming events (Events) and existing state and produces updated state and zero or more Effect values, and stateful code within the actor that executes the Effect values, producing effects. This is an instance of the Reducer architecture, and has the effect of making the state transition logic purely stateless and allowing it to be tested and reasoned about without concern for side effects.

As part of this, the `Command` type has been renamed to `Effect`, for accuracy: a value of this type can represent not only a user command (such as `.start`, `.stop` or `.reconnect`) but non-user events such as key rotations and network condition changes. (Additionally, renaming it to Event keeps it consistent with common nomenclature in the Reducer pattern.)

There is also a suite of unit tests to validate state transitions in the Actor through its Reducer.

(There was an earlier, unnumbered, PR,  but it fell by the wayside of extensive rebasing activity.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280)

### DIFF
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -45,6 +45,7 @@
 		44B02E3B2BC5732D008EDF34 /* LoggingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44B02E3A2BC5732D008EDF34 /* LoggingTests.swift */; };
 		44B02E3C2BC5B8A5008EDF34 /* Bundle+ProductVersion.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5891BF1B25E3E3EB006D6FB0 /* Bundle+ProductVersion.swift */; };
 		44B3C43A2BFE2C800079782C /* PacketTunnelActorReducer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44B3C4392BFE2C800079782C /* PacketTunnelActorReducer.swift */; };
+		44B3C43D2C00CBBD0079782C /* PacketTunnelActorReducerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44B3C43C2C00CBBC0079782C /* PacketTunnelActorReducerTests.swift */; };
 		44BB5F972BE527F4002520EB /* TunnelState+UI.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44BB5F962BE527F4002520EB /* TunnelState+UI.swift */; };
 		44BB5F982BE527F4002520EB /* TunnelState+UI.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44BB5F962BE527F4002520EB /* TunnelState+UI.swift */; };
 		44DD7D242B6CFFD70005F67F /* StartTunnelOperationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 44DD7D232B6CFFD70005F67F /* StartTunnelOperationTests.swift */; };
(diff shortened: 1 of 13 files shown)


rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280)* status: all files reviewed, 2 unresolved discussions (waiting on @acb-mv)

___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActor.swift` line 341 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280#-Nyt8G5Y71UoznMHIZim:-Nyt8G5Y71UoznMHIZin:b-cc5fpf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c05823c879b7bc137c3801c6bba92662169168d7/ios/PacketTunnelCore/Actor/PacketTunnelActor.swift#L341)):*
> ```Swift
>      - Returns: New connection state or `nil` if current state is at or past `.disconnecting` phase.
>      */
>     internal func makeConnectionState(
> ```

Swiftlint body length warning.
___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActorReducer.swift` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280#-Nyt8oHl5Mbh_jkO47MR:-Nyt8oHl5Mbh_jkO47MS:b-kwy2p2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c05823c879b7bc137c3801c6bba92662169168d7/ios/PacketTunnelCore/Actor/PacketTunnelActorReducer.swift#L55)):*
> ```Swift
>     }
> 
>     static func reducer(_ state: inout State, _ event: Event) -> [Effect] {
> ```

Perhaps this function could be broken down to smaller pieces to make it easier to follow.

acb-mv (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280)* status: all files reviewed, 2 unresolved discussions (waiting on @rablador)

___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActorReducer.swift` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280#-Nyt8oHl5Mbh_jkO47MR:-NytIDsy9evDVT3nzSPq:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c05823c879b7bc137c3801c6bba92662169168d7/ios/PacketTunnelCore/Actor/PacketTunnelActorReducer.swift#L55)):*

Perhaps this function could be broken down to smaller pieces to make it easier to follow.
</blockquote></details>

Done.

acb-mv:
*[`ios/PacketTunnelCore/Actor/PacketTunnelActor.swift` line 341 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280#-Nyt8G5Y71UoznMHIZim:-NytIQkqBsR6DfOryBrg:b4zlrax) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c05823c879b7bc137c3801c6bba92662169168d7/ios/PacketTunnelCore/Actor/PacketTunnelActor.swift#L341)):*

Swiftlint body length warning.
</blockquote></details>

That's not one of the functions touched in this PR

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

buggmagnet (commented):
> Reviewed 2 of 13 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280)* status: all files reviewed, 2 unresolved discussions (waiting on @acb-mv)

___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActorReducer.swift` line 12 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280#-Nz29GQ5AE0wymlV4H-r:-Nz29GQ5AE0wymlV4H-s:b46dvon) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d119ec0570d376f4079b782bd845aabc232e84de/ios/PacketTunnelCore/Actor/PacketTunnelActorReducer.swift#L12)):*
> ```Swift
> import WireGuardKitTypes
> 
> extension WireGuardKey where Self: Equatable {}
> ```

That shouldn't be needed, `WireGuardKey` types already implement `Equatable` 
___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActorReducer.swift` line 55 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280#-Nz2Bc54Dvrfvm0thK0f:-Nz2Bc54Dvrfvm0thK0g:berws6e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d119ec0570d376f4079b782bd845aabc232e84de/ios/PacketTunnelCore/Actor/PacketTunnelActorReducer.swift#L55)):*
> ```Swift
>     }
> 
>     static func reducer(_ state: inout State, _ event: Event) -> [Effect] {
> ```

Any specific reason why it's all static methods ? 
Unless I'm mistaken, a struct should be able to do the same job and we can avoid having to depend on `PacketTunnelActor` inside the tests equally well. 

I'm assuming this was done so we don't need to instantiate it for tests, and just use the static functions instead, but it will just make it harder to actually write logic tests for `PacketTunnelActor` now because with this, we can't avoid dealing with reduced states.

IMHO this shouldn't be an extension of the PacketTunnel at all, it should just consume it, and depend on it.

buggmagnet (approved):
:lgtm:

> Reviewed 10 of 13 files at r1, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280#-:-Nz36RRP9KuAvJVcVqvy:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @acb-mv)

___
*[`ios/PacketTunnelCore/Actor/PacketTunnelActorReducer.swift` line 55 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6280#-Nz2Bc54Dvrfvm0thK0f:-Nz2zNDr-7yF3MuICQYA:bpld2nf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d119ec0570d376f4079b782bd845aabc232e84de/ios/PacketTunnelCore/Actor/PacketTunnelActorReducer.swift#L55)):*

Any specific reason why it's all static methods ? 
Unless I'm mistaken, a struct should be able to do the same job and we can avoid having to depend on `PacketTunnelActor` inside the tests equally well. 

I'm assuming this was done so we don't need to instantiate it for tests, and just use the static functions instead, but it will just make it harder to actually write logic tests for `PacketTunnelActor` now because with this, we can't avoid dealing with reduced states.

IMHO this shouldn't be an extension of the PacketTunnel at all, it should just consume it, and depend on it.
</blockquote></details>

This was discussed offline, we decided to take an approach with both the benefits of having a struct, and use static functions to infer that there shouldn't be any statefulness in there.

PR #6284, opened 2024-05-24
Remove SUPPORTED_COMPOSITORS check for wayland

hulthe:
Effectively enable wayland for all linux distros that use it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6284)


faern:
Has this been tested well? Last time I checked the app worked really bad under at least Gnome. The PR description could be a bit more explanatory as to why this is considered safe now.

PR #6287, opened 2024-05-27
IOS-699: fix crash on deletion of access method

acb-mv:
`ListAccessMethodViewController`'s data source was getting out of sync with `fetchedItems`, which was caused by reloading items (useful when existing items have changed, catastrophic for when items have been removed from the middle of the list). Only reloading if the number of items is the same fixes the issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6287)

### DIFF
--- a/ios/MullvadVPN/Coordinators/Settings/APIAccess/List/ListAccessMethodViewController.swift
+++ b/ios/MullvadVPN/Coordinators/Settings/APIAccess/List/ListAccessMethodViewController.swift
@@ -180,7 +180,9 @@ class ListAccessMethodViewController: UIViewController, UITableViewDelegate {
         itemIdentifier: ListAccessMethodItemIdentifier
     ) -> UITableViewCell {
         let cell = tableView.dequeueReusableView(withIdentifier: CellReuseIdentifier.default, for: indexPath)
-        let item = fetchedItems[indexPath.row]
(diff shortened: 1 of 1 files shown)


rablador (commented):
The crash happens when the snapshot tries to reload items during removal. In `updateDataSource` we create a new snapshot and simultaneously populate and reload items from the repository. It seems the snapshot then crashes when it tries to compare the new items (3) to the old (4), making the index go out of range. A slight guess, but in a "pure" removal there is no comparison made and therefore no crash. Also, adding and reloading at the same time seems to be fine, since nothing is removed and range remains "safe".

The solution here - although more cumbersome - would be to avoid reloading when removing items. It was put there to make sure that non-destructive changes to existing api methods (eg. name change) would be propagated back to the list.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6287#-:-NyyH0-b5gxUU7e-YjnV:b-vq01jo)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6287)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6290, opened 2024-05-28
Fix filename in Android build instructions

dlon:
I also updated the suggested cmd tools to the latest version at the same time (and tested them).

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6290)

### DIFF
--- a/android/BuildInstructions.md
+++ b/android/BuildInstructions.md
@@ -105,8 +105,8 @@ Linux distro:
   cd /opt/android     # Or some other directory to place the Android SDK
   export ANDROID_HOME=$PWD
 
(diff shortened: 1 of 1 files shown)


albin-mullvad (approved):
:lgtm:

We should also update the ones in the container, but we can do that when updating other things such as the sdk

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6290#-:-Nz31qX65Dan9OlWWFC0:biiczs1)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6299, opened 2024-05-29
Add a safe FFI wrapper in wireguard-go-rs

hulthe:
### TODO
- [x] Resolve TODOs
- [x] Make sure android works (again)

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299)


talpid-wireguard/src/lib.rs
```diff
@@ -601,7 +601,7 @@ impl WireguardMonitor {
     /// Reconfigures the tunnel to use the provided config while potentially modifying the config
     /// and restarting the obfuscation provider. Returns the new config used by the new tunnel.
     async fn reconfigure_tunnel(
-        tunnel: &Arc<Mutex<Option<Box<dyn Tunnel>>>>,
+        tunnel: &Arc<AsyncMutex<Option<Box<dyn Tunnel>>>>,
```
Serock3, line 604: Unrelated to your PR, but isn't this arguement type a bit weird? Why is the tunnel wrapped in both an `Arc` and `Box`? And why is the option inside the mutex, wouldn't it make more sense with a `Option<Arc<AsyncMutex<dyn Tunnel>>>`? Actually, we don't seem to mutate the tunnel in any way, why not just take a `&dyn Tunnel`? :thinking_face:

talpid-wireguard/src/lib.rs
```diff
@@ -144,7 +144,7 @@ impl Error {
 pub struct WireguardMonitor {
     runtime: tokio::runtime::Handle,
     /// Tunnel implementation
-    tunnel: Arc<Mutex<Option<Box<dyn Tunnel>>>>,
+    tunnel: Arc<AsyncMutex<Option<Box<dyn Tunnel>>>>,
```
Serock3, line 147: Why is a tokio mutex needed here?

wireguard-go-rs/src/lib.rs
```diff
@@ -1,70 +1,265 @@
...
+}
 
-    // Android
-    #[cfg(target_os = "android")]
-    pub fn wgTurnOn(
-        settings: *const i8,
-        fd: Fd,
+// NOTE: Must be kept in sync with libwg.go
+// NOTE: must be kept in sync with `result_from_code`
+// INVARIANT: Will aways be represented as a negative i32
```
Serock3, line 25:
```suggestion
// INVARIANT: Will always be represented as a negative i32
```

talpid-wireguard/src/lib.rs
hulthe: Due to a small refactor in another part of the code, I needed to be able to hold the lock across an await point. We need a tokio mutex for that. Unfortunately that change kinda cascaded all the way up

talpid-wireguard/src/lib.rs
hulthe: Good questions

wireguard-go-rs/src/lib.rs
```diff
@@ -1,70 +1,265 @@
...
+    #[error("Something went wrong when enabling DAITA.")]
+    EnableDaita = -6,
+
+    #[error("`libwg` provided an unknown error code. This is a bug.")]
+    Other = i32::MIN,
+}
+
+impl Tunnel {
+    // TODO: this function is supposed to be a safe wrapper, but as clippy points out, 
+    // the logging_context is a *mut, which may unsafely be dereferenced by the callback.
```
Serock3, line 54: Why does the function have to take logging_context as a `*mut c_void` and not `logging_context: &mut LoggingContext`?

hulthe: can't pass rust references through ffi,e
i'm thinking we can drop the *mut though and just pass a u64 instead :thinking:

wireguard-go-rs/src/lib.rs
```diff
@@ -1,70 +1,265 @@
...
 
-    // Returns the file descriptor of the tunnel IPv4 socket.
-    pub fn wgGetConfig(handle: i32) -> *mut c_char;
+        result_from_code(code)?;
+        Ok(Tunnel { handle: code })
+    }
 
-    // Sets the config of the WireGuard interface.
-    pub fn wgSetConfig(handle: i32, settings: *const i8) -> i32;
+    pub fn turn_off(self) -> Result<(), Error> {
```
Serock3, line 81: It's a bit akward call `ffi::wgTurnOff` both on drop and also provide a manual version. Slightly outside the scope of this PR, but could the `stop` method on the `Tunnel` trait not just be removed in favour of just cleaning up on drop for all `Tunnel` implementations?

hulthe: The benefit of having a stop function is that we can explicitly handle the error in case wgTurnOff should fail

wireguard-go-rs/src/lib.rs
```diff
@@ -1,70 +1,259 @@
...
+    pub const fn as_raw(self) -> i32 {
+        self as i32
+    }
+}
+
+mod ffi {
+    use super::{Fd, LoggingCallback};
+    use core::ffi::{c_char, c_void};
+
+    extern "C" {
```
Serock3, line 199: Should the error cases for these function be documented here?

wireguard-go-rs/src/lib.rs
Serock3: The only place we call the stop function is in  `talpid_wireguard::WireguardMonitor::stop_tunnel`, and there we just print the error and then throw it away

wireguard-go-rs/src/lib.rs
Serock3: But can't the `turn_on` fn handle the conversion from `&mut LoggingContext` to `*mut c_void`?

talpid-wireguard/src/wireguard_go/mod.rs
```diff
@@ -191,64 +178,37 @@ impl WgGoTunnel {
...
 
-    fn set_config(
-        &self,
-        config: Config,
-    ) -> Pin<Box<dyn Future<Output = std::result::Result<(), super::TunnelError>> + Send>> {
-        let wg_config_str = config.to_userspace_format();
-        let handle = self.tunnel_handle;
-        #[cfg(target_os = "android")]
-        let tun_provider = self.tun_provider.clone();
+    // TODO: should probably be &mut to guard against concurrency issues?
```
Serock3, line 201: Yes, that seems reasonable

wireguard-go-rs/src/lib.rs
hulthe: If we're passing a `&mut` over ffi (even as a *mut), we would need to make sure the go-code doesn't violate any of the rust rules about references, which might be tricky.

Thing is though, right now we're not actually passing a pointer as the logging context, instead we're just passing a `u32` by casting it to a  `*mut c_void`. For our use-case, there's no need to use pointers.

wireguard-go-rs/src/lib.rs
```diff
@@ -1,70 +1,259 @@
...
+        {
+            // SAFETY:
+            // we checked for null, and wgGetConfig promises that this is a valid cstr.
+            // config_len comes from the CStr above, so it should be good.
+            let config_bytes = unsafe { slice::from_raw_parts_mut(ptr, config_len) };
+            config_bytes.zeroize();
+        }
+
+        // SAFETY: the pointer was created by wgGetConfig, and we are no longer using it.
+        unsafe { ffi::wgFreePtr(ptr.cast()) };
```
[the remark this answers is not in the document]
hulthe: Hmm... Good point. If `f` panics, the pointer will get leaked. Considering that it contains secrets, we should probably guard against that.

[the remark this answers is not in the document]
hulthe: Added a drop  guard, which should make sure that ptr is zeroed and freed ever if `f` panics

wireguard-go-rs/src/lib.rs
hulthe: Maybe, but that seems like a lot of work and hard to keep in sync

wireguard-go-rs/src/lib.rs
hulthe: True. Could be worth refactoring at some point

talpid-wireguard/src/connectivity_check.rs
```diff
@@ -214,8 +215,7 @@ impl ConnectivityMonitor {
     fn get_stats(&self) -> Option<Result<StatsMap, Error>> {
         self.tunnel_handle
             .upgrade()?
-            .lock()
-            .ok()?
+            .blocking_lock()
```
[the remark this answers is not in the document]
hulthe: I've checked, and using blocking_lock here is fine. Tokio will panic if this method is called from a tokio runtime

wireguard-go-rs/libwg/libwg.go
```diff
@@ -23,12 +23,27 @@ import (
...
+
+	// The provided tunnel handle did not refer to an existing tunnel.
 	ERROR_UNKNOWN_TUNNEL
+
+	// The provided public key did not refer to an existing peer.
 	ERROR_UNKNOWN_PEER
+
+	// Something went wrong when enabling DAITA.
+	// TODO: consider removing this? should probably be replaced by more specific errors
 	ERROR_ENABLE_DAITA
```
hulthe, line 47: Keep or remove? Whaddya think?

MarkusPettersson98 (changes requested):
> Reviewed 1 of 10 files at r1, 3 of 80 files at r3, 7 of 13 files at r6, 71 of 76 files at r7, 2 of 2 files at r8, 5 of 5 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299)* status: all files reviewed, 13 unresolved discussions (waiting on @hulthe and @Serock3)

___
*[`talpid-wireguard/src/connectivity_check.rs` line 218 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299#gh-1624260056:-NzXppRD9j-_TnkZNpMA:b-lqvovz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6d798bdf38ada7eba791de075dbdedcb2d4407c5/talpid-wireguard/src/connectivity_check.rs#L218)):*

I've checked, and using blocking_lock here is fine. Tokio will panic if this method is called from a tokio runtime
</blockquote></details>

Yeah, I also think it is fine. Maybe add a "Panics"-section to this function's doc comment which echos the same gotcha as [blocking_lock](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#method.blocking_lock)?:)
___
*[`wireguard-go-rs/libwg/libwg.go` line 47 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299#gh-1625717278:-NzXtUsA2nFdu_psj2Qo:b7o3f6h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b59c51f37cbee00d830be817684ce8bdd1db1bd/wireguard-go-rs/libwg/libwg.go#L47)):*

Keep or remove? Whaddya think?
</blockquote></details>

Keep if we currently can't replace it with (a) more specific error(s), otherwise remove it in favor of it/those:)
___
*[`wireguard-go-rs/src/lib.rs` line 68 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299#-NzXuY8C9bSzjuV9wvpr:-NzXuY8C9bSzjuV9wvps:b-2xtc17) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f4e660d1f79a9ebe8730f4b022af1eafe01b5c10/wireguard-go-rs/src/lib.rs#L68)):*
> ```XML
>     /// for the tunnel device and logging. For targets other than android, this also takes an MTU value.
>     ///
>     /// The `logging_callback` let's you provide a Rust function that receivec any logging output
> ```

receives*



_Code quote:_
```XML
receivec
```
___
*[`wireguard-go-rs/src/lib.rs` line 69 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299#-NzXubPxDEOE_EfRSB0l:-NzXubPxDEOE_EfRSB0m:b32wfi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f4e660d1f79a9ebe8730f4b022af1eafe01b5c10/wireguard-go-rs/src/lib.rs#L69)):*
> ```XML
>     ///
>     /// The `logging_callback` let's you provide a Rust function that receivec any logging output
>     /// from wireguard-go. `logging_context` is an value that will be passed to each invocation of
> ```

a*



_Code quote:_
```XML
an
```
___
*[`wireguard-go-rs/src/lib.rs` line 109 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299#-NzXwBLh1p0Z1CMs0Bhg:-NzXwBLh1p0Z1CMs0Bhh:b-wev2vo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f4e660d1f79a9ebe8730f4b022af1eafe01b5c10/wireguard-go-rs/src/lib.rs#L109)):*
> ```XML
>     ///
>     /// **NOTE:** You should take extra care to avoid copying any secrets from the config without zeroizing them afterwards.
>     // NOTE: this could return a guard type with a custom Drop impl instead, but me lazy.
> ```

Would be neat ;) No need to though



_Code quote:_
```XML
// NOTE: this could return a guard type with a custom Drop impl instead, but me lazy.
```

talpid-wireguard/src/lib.rs
```diff
@@ -1055,10 +1056,19 @@ pub enum TunnelError {
     #[error("Failed to start wireguard tunnel")]
     FatalStartWireguardError,
 
+    /// Failed to start the wireguard tunnel.
+    #[error("Failed to start wireguard-go tunnel: {status}")]
+    StartWireguardError {
```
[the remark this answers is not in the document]
hulthe: Done

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299)* status: 83 of 86 files reviewed, 12 unresolved discussions (waiting on @MarkusPettersson98 and @Serock3)

___
*[`wireguard-go-rs/libwg/libwg.go` line 47 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299#gh-1625717278:-NzY0WgPCeL-_-dk_YK-:b4ch3ad) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b59c51f37cbee00d830be817684ce8bdd1db1bd/wireguard-go-rs/libwg/libwg.go#L47)):*

Keep if we currently can't replace it with (a) more specific error(s), otherwise remove it in favor of it/those:)
</blockquote></details>

We could, but it would involve refactoring wireguard-go. I'm removing this TODO from here as I feel it's slightly out of scope of this PR

MarkusPettersson98 (changes requested):
> Reviewed 2 of 2 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299)* status: all files reviewed, 12 unresolved discussions (waiting on @hulthe and @Serock3)

___
*[`talpid-wireguard/src/wireguard_go/mod.rs` line 230 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299#-NzXt0TbCG9PJ1fO7Yyy:-NzXt0TbCG9PJ1fO7Yyz:bwxfciy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f4e660d1f79a9ebe8730f4b022af1eafe01b5c10/talpid-wireguard/src/wireguard_go/mod.rs#L230)):*
> ```XML
>             {
>                 let socket_v4 = unsafe { wgGetSocketV4(handle) };
>                 let socket_v6 = unsafe { wgGetSocketV6(handle) };
> ```

You did already wrap these in safe functions, right? :blush:



_Code quote:_
```XML
                let socket_v4 = unsafe { wgGetSocketV4(handle) };
                let socket_v6 = unsafe { wgGetSocketV6(handle) };
```

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299)* status: 83 of 86 files reviewed, 12 unresolved discussions (waiting on @MarkusPettersson98 and @Serock3)

___
*[`talpid-wireguard/src/connectivity_check.rs` line 218 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299#gh-1624260056:-NzY3iEE4fhXfwF1fHcw:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6d798bdf38ada7eba791de075dbdedcb2d4407c5/talpid-wireguard/src/connectivity_check.rs#L218)):*

Yeah, I also think it is fine. Maybe add a "Panics"-section to this function's doc comment which echos the same gotcha as [blocking_lock](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#method.blocking_lock)?:)
</blockquote></details>

Done.
___
*[`wireguard-go-rs/src/lib.rs` line 68 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299#-NzXuY8C9bSzjuV9wvpr:-NzY3Zs3EMA38R1i5KdB:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f4e660d1f79a9ebe8730f4b022af1eafe01b5c10/wireguard-go-rs/src/lib.rs#L68)):*

receives*
</blockquote></details>

Done.
___
*[`wireguard-go-rs/src/lib.rs` line 69 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299#-NzXubPxDEOE_EfRSB0l:-NzY3_7cA05vLRnXvN1l:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f4e660d1f79a9ebe8730f4b022af1eafe01b5c10/wireguard-go-rs/src/lib.rs#L69)):*

a*
</blockquote></details>

Done.

MarkusPettersson98 (approved):
> Reviewed 3 of 3 files at r13, 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299)* status: all files reviewed, 9 unresolved discussions (waiting on @hulthe and @Serock3)

wireguard-go-rs/src/lib.rs
Serock3: Alright, I suggest we add a TODO note and leave it for now

wireguard-go-rs/src/lib.rs
Serock3: Up to you. I'm ok with just refering to the libwg.go source code for details also.

wireguard-go-rs/libwg/libwg.go
Serock3: Why should we remove it? It's a possible error case. It's almost a shame we aren't more specific about the underlying error,

talpid-wireguard/src/lib.rs
Serock3: This should probably be refactored at some point IMO, but let's drop it for this PR.

wireguard-go-rs/src/lib.rs
Serock3: Nice!

MarkusPettersson98 (approved):
> Reviewed 1 of 2 files at r15, 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6299)* status: all files reviewed, 10 unresolved discussions (waiting on @hulthe and @Serock3)

PR #6316, opened 2024-06-04
Save executable path instead of application bundle path for manually added macOS split tunneling applications

raksooo:
I noticed that it is not possible to remove manually added applications in macOS split tunneling. This is because the path that is saved to `gui_settings.json` is the application bundle path and when we remove it we try to remove the executable path. This PR changes the behavior to save the executable path in `gui_settings.json` instead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6316)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6316)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6338, opened 2024-06-10
Add integration tests for WG over Shadowsocks

dlon:
Close DES-1025.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6338)

### DIFF
--- a/mullvad-relay-selector/src/relay_selector/query.rs
+++ b/mullvad-relay-selector/src/relay_selector/query.rs
@@ -32,9 +32,10 @@ use crate::{AdditionalWireguardConstraints, Error};
 use mullvad_types::{
     constraints::Constraint,
     relay_constraints::{
-        BridgeConstraints, LocationConstraint, ObfuscationSettings, OpenVpnConstraints, Ownership,
-        Providers, RelayConstraints, RelaySettings, SelectedObfuscation, ShadowsocksSettings,
-        TransportPort, Udp2TcpObfuscationSettings, WireguardConstraints,
+        BridgeConstraints, BridgeSettings, BridgeState, BridgeType, LocationConstraint,
+        ObfuscationSettings, OpenVpnConstraints, Ownership, Providers, RelayConstraints,
+        SelectedObfuscation, ShadowsocksSettings, TransportPort, Udp2TcpObfuscationSettings,
+        WireguardConstraints,
     },
     wireguard::QuantumResistantState,
     Intersection,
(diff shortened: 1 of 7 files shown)


test/test-manager/src/tests/tunnel.rs
```diff
@@ -203,6 +203,49 @@ pub async fn test_udp2tcp_tunnel(
...
+
+    let relay_settings = RelaySettings::Normal(RelayConstraints {
+        tunnel_protocol: Constraint::Only(TunnelType::Wireguard),
+        ..Default::default()
+    });
+
+    set_relay_settings(&mut mullvad_client, relay_settings)
+        .await
+        .context("Failed to update relay settings")?;
+
```
Serock3, line 233: Calling `set_obfuscation_settings` separetly from `set_relay_settings` is not necessary, right? We could also make use of the relay query builder here, which is nice if we change the underlying settings format

``suggestion
    set_relay_settings(
        &mut mullvad_client,
        RelaySettings::Normal(
            RelayQueryBuilder::new()
                .wireguard()
                .shadowsocks()
                .into_constraint(),
        ),
    )
    .await?;
```

File: test/test-manager/src/tests/tunnel.rs [hunk 2]
```diff
@@ -572,6 +615,54 @@ pub async fn test_quantum_resistant_multihop_udp2tcp_tunnel(
...
+    let relay_constraints = RelayQueryBuilder::new()
+        .wireguard()
+        .multihop()
+        .into_constraint();
+
+    mullvad_client
+        .set_relay_settings(RelaySettings::Normal(relay_constraints))
+        .await
+        .context("Failed to update relay settings")?;
+
```
Serock3, line 655: Same here, this should be replacable with
```suggestion
    set_relay_settings(
        &mut mullvad_client,
        RelaySettings::Normal(
            RelayQueryBuilder::new()
                .wireguard()
                .multihop()
                .shadowsocks()
                .into_constraint(),
        ),
    )
    .await?;
```

test/test-manager/src/tests/tunnel.rs
dlon: It is necessary to call `set_obfuscation_settings`. When the query is converted to a `RelaySettings`, it will not use any of the obfuscation constraints, as the relay settings struct doesn't contain any obfuscation constraints.

This is obviously a footgun, but that's how it works for the time being. Maybe our helper function `set_relay_settings` should take care of this? That doesn't solve the underlying problem, though.

dlon: Maybe instead of implementing `From<RelayQuery>` for `RelayConstraints`, we should add a method to `RelayQuery` that returns obfuscation settings (if there are any) as well as relay settings. That would force the caller to pay attention to both.

test/test-manager/src/tests/tunnel.rs
dlon: See above.

mullvad-relay-selector/src/relay_selector/query.rs
```diff
@@ -107,6 +108,37 @@ impl RelayQuery {
             openvpn_constraints: OpenVpnRelayQuery::new(),
         }
     }
+
+    /// The mapping from [`RelayQuery`] to all underlying settings types.
```
Serock3, line 112: What about making this a bit more explicit?
```suggestion
    /// The mapping from [`RelayQuery`] to all underlying settings types.
    /// 
    /// Useful in contexts where you cannot use the query directly but
    /// still want use of the builder for convenience. For example in
    /// end to end tests where you must use the management interface
    /// to apply settings to the daemon.
```

Serock3 (commented):
Just a minor nit comment. Would be nice to see that the tests pass before merging.

Serock3 (approved):

Serock3 (approved):

Serock3 (approved):
> Reviewed 2 of 7 files at r4, 3 of 3 files at r5, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6338)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

PR #6344, opened 2024-06-13
Fix missing device refresh

albin-mullvad:
This PR aims to fix an issue where the UI doesn't trigger automatic device updates/refresh. Instead, the app would rely on other mechanisms such as the tunnel connection or purchase flows to update/refresh the device. The consequence of this is that users could end up in strange scenarios such as a device being revoked immediately after purchasing time.

The new behavior is that we update/refresh the device similar to how we handle purchase verification as part of a few screens. This is a bit similar to how it's handled on desktop where they are even more aggressive on updating/refresh the device. On desktop it's done every time a gRPC connection is established.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6344)

### DIFF
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/ConnectViewModel.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/ConnectViewModel.kt
@@ -118,6 +118,7 @@ class ConnectViewModel(
                 viewModelScope.launch { accountRepository.getAccountData() }
             }
         }
+        viewModelScope.launch { deviceRepository.updateDevice() }
     }
 
     fun onDisconnectClick() {
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/OutOfTimeViewModel.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/OutOfTimeViewModel.kt
@@ -59,6 +59,7 @@ class OutOfTimeViewModel(
         }
         verifyPurchases()
         fetchPaymentAvailability()
+        viewModelScope.launch { deviceRepository.updateDevice() }
     }
 
     fun onSitePaymentClick() {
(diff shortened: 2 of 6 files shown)


Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6344)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @albin-mullvad)

___
*[`android/lib/shared/src/main/kotlin/net/mullvad/mullvadvpn/lib/shared/DeviceRepository.kt` line 39 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6344#-O-FXg3z5PeAmLVXuhvp:-O-FXg3z5PeAmLVXuhvq:b3bqfbo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fea82519738c9de7263d96ad106994f73a7d3b8e/android/lib/shared/src/main/kotlin/net/mullvad/mullvadvpn/lib/shared/DeviceRepository.kt#L39)):*
> ```Kotlin
> 
>     suspend fun updateDevice() {
>         Log.d("mullvad", "Update device")
> ```

This is to ensure we see that device is updated, e.g if we get new support ticket?

albin-mullvad (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6344)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @Rawa)

___
*[`android/lib/shared/src/main/kotlin/net/mullvad/mullvadvpn/lib/shared/DeviceRepository.kt` line 39 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6344#-O-FXg3z5PeAmLVXuhvp:-O-FYwvVAG0-VlN2tn1G:b7fx6x8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fea82519738c9de7263d96ad106994f73a7d3b8e/android/lib/shared/src/main/kotlin/net/mullvad/mullvadvpn/lib/shared/DeviceRepository.kt#L39)):*

This is to ensure we see that device is updated, e.g if we get new support ticket?
</blockquote></details>

Yup, I thought that could be valuable :thumbsup:

Pururun (approved):
:lgtm:

> Reviewed 1 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6344#-:-O-FchTV4pLoXd-6aJ3-:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Rawa)

PR #6374, opened 2024-06-18
Link statically against libwg

dlon:
Use `Cargo.toml` to link against `maybenot-ffi` on desktop. This solves the issue of LTO breaking when using a separate archive file, as well as some issues of some scripts failing due to `libwg.so` not being found.

Fix DES-1038.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6374)

### DIFF
--- a/wireguard-go-rs/Cargo.toml
+++ b/wireguard-go-rs/Cargo.toml
@@ -8,3 +8,10 @@ license.workspace = true
 thiserror.workspace = true
 log.workspace = true
 zeroize = "1.8.1"
+
+[target.'cfg(target_os = "linux")'.dependencies]
+# The app does not depend on maybenot-ffi itself, but adds it as a dependency to expose FFI symbols to wireguard-go. 
+# This is done, instead of using the makefile in wireguard-go to build maybenot-ffi into its archive, to prevent
+# name clashes induced by link-time optimization.
+# NOTE: the version of maybenot-ffi below must be the same as the version checked into the wireguard-go submodule
+maybenot-ffi = "1.0.0"
--- a/wireguard-go-rs/src/lib.rs
+++ b/wireguard-go-rs/src/lib.rs
@@ -26,6 +26,10 @@ pub type LoggingContext = u64;
 pub type LoggingCallback =
     unsafe extern "system" fn(level: WgLogLevel, msg: *const c_char, context: LoggingContext);
 
+// Make symbols from maybenot-ffi visible to wireguard-go
+#[cfg(daita)]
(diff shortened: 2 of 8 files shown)


wireguard-go-rs/src/lib.rs
```diff
@@ -283,3 +283,7 @@ mod ffi {
         pub fn wgGetSocketV6(handle: i32) -> Fd;
     }
 }
+
+// Make symbols from maybenot-ffi visible
+#[cfg(daita)]
+use maybenot_ffi as _;
```
Serock3, line 289: Should this be at the very bottom of the file?

Serock3: Perhaps we should even write more clearly (somewhere) that we are linking `maybenot-ffi` to the app only for the symbols to be visible to  `wireguard-go`, not for the app itself, instead of building it into the `wiregurad-go` archive using the makefile.

Serock3: Also, we should document that the `maybenot-ffi` version in the toml file has to match the version checked in to `wireguard-go`

wireguard-go-rs/Cargo.toml
```diff
@@ -8,3 +8,7 @@ license.workspace = true
 thiserror.workspace = true
 log.workspace = true
 zeroize = "1.8.1"
+
+[target.'cfg(target_os = "linux")'.dependencies]
+# wireguard-go depends on maybenot-ffi
```
Serock3, line 13:
```suggestion
# The app does not depend on maybenot-ffi itself, but adds it as a dependency to expose FFI symbols to wireguard-go. 
# This is done, instead of using the makefile in wireguard-go to build maybenot-ffi into its archive, to prevent
# name clashes induced by link-time optimization.
# NOTE: the version of maybenot-ffi below must be the same as the version checked into the wireguard-go submodule
```

Serock3 (commented):
> Reviewed 3 of 8 files at r1, 2 of 3 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6374)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

Serock3 (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6374)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6381, opened 2024-06-19
Restart `mullvad-daemon` after user grants Full Disk Access (macOS)

MarkusPettersson98:
This PR fixes an issue where the `mullvad-daemon` would not automatically restart after the user had granted "Mullvad VPN" application Full Disk Access. This is needed to make split tunneling work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6381)

### DIFF
--- a/gui/src/main/tunnel-state.ts
+++ b/gui/src/main/tunnel-state.ts
@@ -1,5 +1,5 @@
 import { connectEnabled, disconnectEnabled, reconnectEnabled } from '../shared/connect-helper';
-import { ILocation, TunnelState } from '../shared/daemon-rpc-types';
+import { ErrorStateCause, ILocation, TunnelState } from '../shared/daemon-rpc-types';
 import { Scheduler } from '../shared/scheduler';
 
 export interface TunnelStateProvider {
@@ -20,10 +20,15 @@ export default class TunnelStateHandler {
   // Scheduler for discarding the assumed next state.
   private tunnelStateFallbackScheduler = new Scheduler();
 
+  private receivedFullDiskAccessError = false;
+
   private lastKnownDisconnectedLocation: Partial<ILocation> | undefined;
 
   public constructor(private delegate: TunnelStateHandlerDelegate) {}
 
+  public get hasReceivedFullDiskAccessError() {
+    return this.receivedFullDiskAccessError;
+  }
   public get tunnelState() {
     return this.tunnelStateValue;
   }
@@ -53,6 +58,12 @@ export default class TunnelStateHandler {
   }
 
   public handleNewTunnelState(newState: TunnelState) {
+    if (newState.state === 'error' && newState.details) {
+      if (newState.details.cause === ErrorStateCause.needFullDiskPermissions) {
+        this.receivedFullDiskAccessError = true;
+      }
+    }
+
     // If there's a fallback state set then the app is in an assumed next state and need to check
     // if it's now reached or if the current state should be ignored and set as the fallback state.
     if (this.tunnelStateFallback) {
--- a/mullvad-cli/src/format.rs
+++ b/mullvad-cli/src/format.rs
@@ -225,6 +225,21 @@ fn print_error_state(error_state: &ErrorState) {
             println!("Blocked: {cause}");
             println!("Your kernel might be terribly out of date or missing nftables");
         }
+        #[cfg(target_os = "macos")]
+        cause @ talpid_types::tunnel::ErrorStateCause::NeedFullDiskPermissions => {
+            println!("Blocked: {cause}");
+            println!();
+            println!(
+                r#"Enable "Full Disk Access" for "Mullvad VPN" in the macOS system settings:"#
+            );
+            println!(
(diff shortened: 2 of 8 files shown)


gui/src/main/tunnel-state.ts
```diff
@@ -20,10 +20,15 @@ export default class TunnelStateHandler {
   // Scheduler for discarding the assumed next state.
   private tunnelStateFallbackScheduler = new Scheduler();
 
+  private receivedFullDiskError = false;
+
```
Serock3, line 24: Alternatively, we could just call this `shutdownDaemonWhenGuiCloses`

mullvad-cli/src/format.rs
```diff
@@ -225,6 +225,17 @@ fn print_error_state(error_state: &ErrorState) {
             println!("Blocked: {cause}");
             println!("Your kernel might be terribly out of date or missing nftables");
         }
+        #[cfg(target_os = "macos")]
+        cause @ talpid_types::tunnel::ErrorStateCause::NeedFullDiskPermissions => {
+            println!("Blocked: {cause}");
+            println!();
+            println!(r#"Enable "Full Disk Access" for "Mullvad VPN" in the macOS system settings:"#);
+            println!(r#"open "x-apple.systempreferences:com.apple.preference.security?Privacy_AllFiles"#);
```
Serock3, line 233: Is this formatted correctly? Seems like there is a `"` before `x-apple`, but not ending quote.

mullvad-cli/src/format.rs
```diff
@@ -225,6 +225,17 @@ fn print_error_state(error_state: &ErrorState) {
...
+            println!();
+            println!("Restart the Mullvad daemon for the change to take effect:");
+            println!("launchctl unload -w /Library/LaunchDaemons/net.mullvad.daemon.plist");
+            println!("launchctl load -w /Library/LaunchDaemons/net.mullvad.daemon.plist");
```
Serock3, line 237: Why not write it using a single `println!`?
```suggestion
            println!(r#"
            Enable "Full Disk Access" for "Mullvad VPN" in the macOS system settings:
            open "x-apple.systempreferences:com.apple.preference.security?Privacy_AllFiles

            Restart the Mullvad daemon for the change to take effect:
            launchctl unload -w /Library/LaunchDaemons/net.mullvad.daemon.plist
            launchctl load -w /Library/LaunchDaemons/net.mullvad.daemon.plist
            "#);
```

gui/src/main/tunnel-state.ts
dlon: IMO, that would be worse. As is, `TunnelStateHandler` doesn't need to care what the boolean is used for. If it were renamed to `shutdownDaemonWhenGuiCloses` then it would couple it (slightly) more with the main class.

mullvad-cli/src/format.rs
dlon: You're right! Fixed.

Serock3 (commented):
> Reviewed 2 of 5 files at r1, 1 of 3 files at r2, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6381)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`gui/src/main/tunnel-state.ts` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6381#gh-1646036962:-O-ktGDuAhsJ5rKK9i0I:biftvth) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/263d5cbf622c6f2f157b2762b890246d42312279/gui/src/main/tunnel-state.ts#L24)):*

IMO, that would be worse. As is, `TunnelStateHandler` doesn't need to care what the boolean is used for. If it were renamed to `shutdownDaemonWhenGuiCloses` then it would couple it (slightly) more with the main class.
</blockquote></details>

Sure, I agree with that argument

dlon (approved):
> Reviewed 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6381)* status: all files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

dlon (approved):
> Reviewed 1 of 5 files at r4, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6381)* status: all files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6381)* status: all files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

Serock3 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6381#-:-O-pYQP895GZ8A-zxGb7:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6383, opened 2024-06-19
Fix RUSTSEC-2024-0344

dlon:
Fix [RUSTSEC-2024-0344](https://rustsec.org/advisories/RUSTSEC-2024-0344) by bumping `curve25519-dalek` to the latest version.

> Timing variability of any kind is problematic when working with potentially secret values such as elliptic curve scalars, and such issues can potentially leak private keys and other secrets.

It doesn't appear to be problematic with respect to the app's threat model. An attacker would need local access to the machine.

Close DES-1049.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6383)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -681,15 +681,14 @@ dependencies = [
 
 [[package]]
 name = "curve25519-dalek"
-version = "4.1.2"
+version = "4.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


Serock3 (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6383)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6447, opened 2024-07-05
 Add E2E test of CVE-2019-14899 mitigation

hulthe:

### DIFF
--- a/test/Cargo.lock
+++ b/test/Cargo.lock
@@ -337,6 +337,48 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "11aade7a05aa8c3a351cedc44c3fc45806430543382fcc4743a9b757a2a0b4ed"
 
 [[package]]
+name = "bytecheck"
+version = "0.6.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "23cdc57ce23ac53c931e88a43d06d070a6fd142f2617be5855eb75efc9beb1c2"
+dependencies = [
+ "bytecheck_derive",
+ "ptr_meta",
+ "simdutf8",
+]
+
+[[package]]
+name = "bytecheck_derive"
+version = "0.6.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3db406d29fbcd95542e92559bed4d8ad92636d1ca8b3b72ede10b4bcc010e659"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
+name = "bytemuck"
+version = "1.16.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b236fc92302c97ed75b38da1f4917b5cdda4984745740f153a5d3059e48d725e"
+dependencies = [
+ "bytemuck_derive",
+]
+
+[[package]]
+name = "bytemuck_derive"
+version = "1.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4da9a32f3fed317401fa3c862968128267c3106685286e15d5aaa3d7389c2f60"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.60",
+]
+
+[[package]]
 name = "byteorder"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1963,20 +2005,6 @@ dependencies = [
 
 [[package]]
 name = "nix"
-version = "0.25.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f346ff70e7dbfd675fe90590b92d59ef2de15a8779ae305ebcbfd3f0caf59be4"
-dependencies = [
- "autocfg",
- "bitflags 1.3.2",
- "cfg-if",
- "libc",
- "memoffset 0.6.5",
- "pin-utils",
-]
-
-[[package]]
-name = "nix"
 version = "0.26.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "598beaf3cc6fdd9a5dfb1630c2800c7acd31df7aaf0f565796fba2b53ca1af1b"
@@ -2518,6 +2546,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "ptr_meta"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0738ccf7ea06b608c10564b31debd4f5bc5e197fc8bfe088f68ae5ce81e7a4f1"
+dependencies = [
+ "ptr_meta_derive",
+]
+
+[[package]]
+name = "ptr_meta_derive"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "16b845dbfca988fa33db069c0e230574d15a3088f147a87b64c7589eb662c9ac"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
 name = "quick-error"
 version = "1.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2669,6 +2717,16 @@ name = "regex-syntax"
 version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "adad44e29e4c806119491a7f06f03de4d1af22c3a680dd47f1e6e179439d1f56"
+
+[[package]]
+name = "rend"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "71fe3824f5629716b1589be05dacd749f6aa084c87e00e016714a8cdfccc997c"
+dependencies = [
+ "bytecheck",
+ "bytemuck",
+]
 
 [[package]]
 name = "reqwest"
@@ -3094,6 +3152,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "simdutf8"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f27f6278552951f1f2b8cf9da965d10969b2efdea95a6ec47987ab46edfe263a"
+
+[[package]]
 name = "slab"
 version = "0.4.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3361,6 +3425,7 @@ dependencies = [
  "anyhow",
  "async-tempfile",
  "async-trait",
+ "bytemuck",
  "bytes",
  "chrono",
  "clap",
@@ -3378,11 +3443,12 @@ dependencies = [
  "mullvad-management-interface",
  "mullvad-relay-selector",
  "mullvad-types",
- "nix 0.25.1",
+ "nix 0.26.4",
  "once_cell",
  "pcap",
  "pnet_packet 0.31.0",
  "regex",
+ "rend",
  "scopeguard",
  "serde",
  "serde_json",
@@ -3436,7 +3502,7 @@ dependencies = [
  "libc",
  "log",
  "mullvad-paths",
- "nix 0.25.1",
+ "nix 0.26.4",
  "once_cell",
  "parity-tokio-ipc",
  "plist",
--- a/test/test-manager/Cargo.toml
+++ b/test/test-manager/Cargo.toml
@@ -59,7 +59,9 @@ talpid-types = { path = "../../talpid-types" }
 
 ssh2 = "0.9.4"
 
-nix = { version = "0.25", features = ["net"] }
+nix = { version = "0.26", features = ["net"] }
+rend = { version = "0.4.2", features = ["bytemuck"] }
+bytemuck = { version = "1.16.1", features = ["derive"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 tun = "0.5.1"
--- a/test/test-manager/src/config.rs
+++ b/test/test-manager/src/config.rs
@@ -3,7 +3,7 @@
 use serde::{Deserialize, Serialize};
 use std::{
     collections::BTreeMap,
-    io,
+    env, io,
     ops::Deref,
     path::{Path, PathBuf},
 };
@@ -157,17 +157,20 @@ impl VmConfig {
         Some((self.ssh_user.as_ref()?, self.ssh_password.as_ref()?))
     }
 
-    pub fn get_runner_dir(&self) -> &Path {
-        match self.architecture {
-            None | Some(Architecture::X64) => self.get_x64_runner_dir(),
-            Some(Architecture::Aarch64) => self.get_aarch64_runner_dir(),
-        }
+    pub fn get_runner_dir(&self) -> PathBuf {
+        let target_dir = self.get_target_dir();
+        let subdir = match self.architecture {
+            None | Some(Architecture::X64) => self.get_x64_runner_subdir(),
+            Some(Architecture::Aarch64) => self.get_aarch64_runner_subdir(),
+        };
+
+        target_dir.join(subdir)
     }
 
-    fn get_x64_runner_dir(&self) -> &Path {
-        pub const X64_LINUX_TARGET_DIR: &str = "./target/x86_64-unknown-linux-gnu/release";
-        pub const X64_WINDOWS_TARGET_DIR: &str = "./target/x86_64-pc-windows-gnu/release";
-        pub const X64_MACOS_TARGET_DIR: &str = "./target/x86_64-apple-darwin/release";
+    fn get_x64_runner_subdir(&self) -> &Path {
+        pub const X64_LINUX_TARGET_DIR: &str = "x86_64-unknown-linux-gnu/release";
+        pub const X64_WINDOWS_TARGET_DIR: &str = "x86_64-pc-windows-gnu/release";
+        pub const X64_MACOS_TARGET_DIR: &str = "x86_64-apple-darwin/release";
 
         match self.os_type {
             OsType::Linux => Path::new(X64_LINUX_TARGET_DIR),
@@ -176,15 +179,21 @@ impl VmConfig {
         }
     }
 
-    fn get_aarch64_runner_dir(&self) -> &Path {
-        pub const AARCH64_LINUX_TARGET_DIR: &str = "./target/aarch64-unknown-linux-gnu/release";
-        pub const AARCH64_MACOS_TARGET_DIR: &str = "./target/aarch64-apple-darwin/release";
+    fn get_aarch64_runner_subdir(&self) -> &Path {
+        pub const AARCH64_LINUX_TARGET_DIR: &str = "aarch64-unknown-linux-gnu/release";
+        pub const AARCH64_MACOS_TARGET_DIR: &str = "aarch64-apple-darwin/release";
 
         match self.os_type {
             OsType::Linux => Path::new(AARCH64_LINUX_TARGET_DIR),
             OsType::Macos => Path::new(AARCH64_MACOS_TARGET_DIR),
             _ => unimplemented!(),
         }
+    }
+
+    fn get_target_dir(&self) -> PathBuf {
+        env::var("CARGO_TARGET_DIR")
+            .unwrap_or_else(|_| "./target".into())
+            .into()
     }
 }
 
--- a/test/test-manager/src/tests/cve_2019_14899.rs
+++ b/test/test-manager/src/tests/cve_2019_14899.rs
@@ -1,0 +1,403 @@
+use std::{
+    ffi::{c_char, c_int, OsString},
+    mem::{self, size_of, size_of_val},
+    net::IpAddr,
+    os::fd::RawFd,
+    time::Duration,
+};
+
+use anyhow::{bail, Context};
+use bytemuck::{bytes_of, cast_slice, from_bytes, Pod, Zeroable};
+use futures::{select, FutureExt};
+use libc::{c_void, sockaddr_ll, ETH_ALEN, ETH_P_IP};
+use mullvad_management_interface::MullvadProxyClient;
+use nix::{
+    errno::{errno, Errno},
+    ioctl_readwrite_bad,
+    sys::socket::{
+        setsockopt,
+        sockopt::{self},
+        AddressFamily, MsgFlags, SockFlag, SockProtocol, SockType,
+    },
+};
+use pnet_packet::{ip::IpNextHeaderProtocols, tcp::TcpFlags};
+use rend::{u16_be, u32_be};
+use test_macro::test_function;
+use test_rpc::ServiceClient;
+use tokio::{pin, task::yield_now, time::sleep};
+
+use crate::{
+    tests::helpers,
+    vm::network::{linux::TAP_NAME, NON_TUN_GATEWAY},
+};
+
+use super::TestContext;
+
+const IPV4_VERSION_IHL: u8 = 0x45;
+
+/// The port number we set in the malicious packet.
+const MALICIOUS_PACKET_PORT: u16 = 12345;
+
+#[repr(C, packed)]
+#[derive(Clone, Copy, Debug, Pod, Zeroable)]
+struct EthernetPacket<T> {
+    destination_mac: [u8; 6],
+    source_mac: [u8; 6],
+    ether_type: u16_be,
+    data: T,
+}
+
+#[repr(C, packed)]
+#[derive(Clone, Copy, Debug, Pod, Zeroable)]
+struct Ipv4Header {
+    /// Must be set to [IPV4_VERSION_IHL].
+    version_ihl: u8,
+    dscp_ecn: u8,
+    total_len: u16_be,
+    identification: u16_be,
+    flags_fragment_offset: u16_be,
+    ttl: u8,
+    protocol: u8,
+    header_checksum: u16_be,
+    source_addr: u32_be,
+    destination_addr: u32_be,
+}
+
+#[repr(C, packed)]
+#[derive(Clone, Copy, Debug, Pod, Zeroable)]
+struct Ipv4Packet<T> {
+    header: Ipv4Header,
+    data: T,
+}
+
+#[repr(C, packed)]
+#[derive(Clone, Copy, Debug, Pod, Zeroable)]
+struct TcpHeader {
+    source_port: u16_be,
+    destination_port: u16_be,
+    seq_num: u32_be,
+    ack_num: u32_be,
+    data_offset: u8,
+    tcp_flags: u8,
+    window_size: u16_be,
+    tcp_checksum: u16_be,
+    urgent_pointer: u16_be,
+}
+
+/// Test mitigation for cve-2019-14899.
+///
+/// The vulnerability allowed a malicious router learn the victims private mullvad tunnel IP.
+/// It is performed by sending a TCP packet to the victim with SYN and ACK flags set.
+///
+/// If the destination_addr of the packet was the same as the private IP, the victims computer
+/// would respond to the packet with the RST flag set.
+///
+/// This test simply gets the private tunnel IP from the test runner and sends the SYN/ACK packet
+/// targeted to that address. If the guest does not respond, the test passes.
+#[test_function(target_os = "linux")]
+pub async fn test_cve_2019_14899_mitigation(
+    _: TestContext,
+    rpc: ServiceClient,
+    mut mullvad_client: MullvadProxyClient,
+) -> anyhow::Result<()> {
+    helpers::connect_and_wait(&mut mullvad_client).await?;
+
+    // The vulnerability required local network sharing to be enabled.
+    mullvad_client
+        .set_allow_lan(true)
+        .await
+        .context("Failed to allow local network sharing")?;
+
+    // Create a raw socket which let's us send custom ethernet packets
+    log::info!("Creating raw socket.");
+    let socket: RawFd = nix::sys::socket::socket(
+        AddressFamily::Packet,
+        SockType::Raw,
+        SockFlag::empty(),
+        SockProtocol::EthAll,
+    )
+    .with_context(|| "Failed to create raw socket")?;
+
+    let host_interface = TAP_NAME;
+    let victim_interface = "wg0-mullvad";
+    let gateway_ip = NON_TUN_GATEWAY;
+
+    log::info!("Binding raw socket to tap interface.");
+    setsockopt(
+        socket,
+        sockopt::BindToDevice,
+        &OsString::from(host_interface),
+    )
+    .with_context(|| format!("Failed to bind the socket to {host_interface:?}"))?;
+
+    // Get the private IP address of the victims VPN tunnel
+    let victim_iface_ip = rpc
+        .get_interface_ip(victim_interface.to_string())
+        .await
+        .with_context(|| {
+            format!("Failed to get ip of guest tunnel interface {victim_interface:?}")
+        })?;
+
+    let IpAddr::V4(victim_iface_ip) = victim_iface_ip else {
+        bail!("I didn't ask for IPv6!");
+    };
+
+    // Get the MAC address and "index" of the tap interface.
+    let mut host_interface_mac = [0u8; 6];
+    let host_interface_index: c_int;
+    {
+        // Set up ioctl request structs.
+        let mut if_mac_request: libc::ifreq = unsafe { mem::zeroed() };
+        let host_interface: &[c_char] = cast_slice(host_interface.as_bytes());
+        if_mac_request.ifr_name[..host_interface.len()].copy_from_slice(host_interface);
+
+        let mut if_index_request: libc::ifreq = if_mac_request;
+
+        // call the netdev ioctl that gets the interface index
+        ioctl_readwrite_bad!(get_interface_index, libc::SIOCGIFINDEX, libc::ifreq);
+        unsafe { get_interface_index(socket, &mut if_index_request) }
+            .context("Failed to get index of host interface")?;
+        host_interface_index = unsafe { if_index_request.ifr_ifru.ifru_ifindex };
+
+        // call the netdev ioctl that gets the interface mac address
+        ioctl_readwrite_bad!(get_interface_mac, libc::SIOCGIFHWADDR, libc::ifreq);
+        let result = unsafe { get_interface_mac(socket, &mut if_mac_request) }
+            .context("Failed to get MAC address of host interface")?;
+        log::warn!("SIOCGIFHWADDR result code: {result}");
+
+        host_interface_mac.copy_from_slice(cast_slice(unsafe {
+            &if_mac_request.ifr_ifru.ifru_hwaddr.sa_data[..6]
+        }));
+    }
+
+    // craft a malicious packet.
+    let mut ip_packet = Ipv4Packet {
+        header: Ipv4Header {
+            version_ihl: IPV4_VERSION_IHL,
+            total_len: (size_of::<Ipv4Packet<TcpHeader>>() as u16).into(),
+            identification: 0x77.into(),
+            ttl: 0xff,
+            protocol: IpNextHeaderProtocols::Tcp.0,
+            source_addr: u32::from(gateway_ip).into(),
+
+            // set the destination_addr to the victims private tunnel IP
+            destination_addr: u32::from(victim_iface_ip).into(),
+
+            ..Ipv4Header::zeroed()
+        },
+        data: TcpHeader {
+            // We use the port value to help us identify the response packet.
+            source_port: MALICIOUS_PACKET_PORT.into(),
+            destination_port: MALICIOUS_PACKET_PORT.into(),
+
+            data_offset: 0x50.into(), // 5
+            window_size: 0xff.into(),
+
+            // Important: Set the SYN and ACK flags
+            tcp_flags: ((TcpFlags::SYN | TcpFlags::ACK) as u8).into(),
+
+            ..TcpHeader::zeroed()
+        },
+    };
+    ip_packet.calculate_checksum();
+
+    let eth_packet = EthernetPacket {
+        // TODO: query test runner for mac address
+        destination_mac: [0x52, 0x54, 0x00, 0x12, 0x34, 0x56],
+        source_mac: host_interface_mac,
+        ether_type: (ETH_P_IP as u16).into(),
+        data: ip_packet,
+    };
+
+    let filter = |packet: &Ipv4Packet<TcpHeader>| {
+        let source_port = packet.data.source_port;
+        let destination_port = packet.data.destination_port;
+        let reset_flag_set = (packet.data.tcp_flags & TcpFlags::RST as u8) != 0;
+
+        reset_flag_set
+            && source_port == MALICIOUS_PACKET_PORT
+            && destination_port == MALICIOUS_PACKET_PORT
+    };
+
+    let saw_packet = select! {
+        result = spam_packet(socket, host_interface_index, &eth_packet).fuse() => return result,
+        result = listen_for_packet(socket, filter, Duration::from_secs(3)).fuse() => result?,
+    };
+
+    if saw_packet {
+        bail!("Managed to leak private tunnel IP.");
+    }
+
+    Ok(())
+}
+
+/// Read from the socket and return true if we see a packet that passes the filter.
+/// Returns false if we don't see such a packet within the timeout.
+async fn listen_for_packet(
+    socket: RawFd,
+    filter: impl Fn(&Ipv4Packet<TcpHeader>) -> bool,
+    timeout: Duration,
+) -> anyhow::Result<bool> {
+    let mut buf = vec![0u8; 0xffff];
+
+    let listener = async {
+        loop {
+            // yield so we don't end up hogging the runtime polling the socket
+            yield_now().await;
+
+            match poll_for_packet(socket, &mut buf)? {
+                Some(packet) if filter(&packet.data) => break,
+                Some(_packet) => yield_now().await,
+                None => sleep(Duration::from_millis(100)).await,
+            }
+        }
+        anyhow::Ok(())
+    };
+    pin!(listener);
+
+    select! {
+        result = listener.fuse() => result.map(|_| true),
+        _ = sleep(timeout).fuse() => Ok(false),
+    }
+}
+
+/// Send `packet` on the socket in a loop.
+async fn spam_packet(
+    socket: RawFd,
+    interface_index: c_int,
+    packet: &EthernetPacket<Ipv4Packet<TcpHeader>>,
+) -> anyhow::Result<()> {
+    loop {
+        send_packet(socket, interface_index, packet)?;
+        sleep(Duration::from_millis(50)).await;
+    }
+}
+
+/// Send an ethernet packet on the socket.
+fn send_packet(
+    socket: RawFd,
+    interface_index: c_int,
+    packet: &EthernetPacket<Ipv4Packet<TcpHeader>>,
+) -> anyhow::Result<()> {
+    let result = {
+        let mut destination = sockaddr_ll {
+            sll_family: 0,
+            sll_protocol: 0,
+            sll_ifindex: interface_index,
+            sll_hatype: 0,
+            sll_pkttype: 0,
+            sll_halen: ETH_ALEN as u8,
+            sll_addr: [0; 8],
+        };
+        destination.sll_addr[..6].copy_from_slice(&packet.destination_mac);
+
+        unsafe {
+            libc::sendto(
+                socket,
+                bytes_of(packet).as_ptr() as *const c_void,
+                size_of_val(packet),
+                0,
+                (&destination as *const sockaddr_ll).cast(),
+                size_of::<sockaddr_ll>() as u32,
+            )
+        }
+    };
+
+    if result < 0 {
+        let err = errno();
+        bail!("Failed to send ethernet packet. code={err}");
+    }
+
+    Ok(())
+}
+
+fn poll_for_packet(
+    socket: RawFd,
+    buf: &mut [u8],
+) -> anyhow::Result<Option<EthernetPacket<Ipv4Packet<TcpHeader>>>> {
+    let n = match nix::sys::socket::recv(socket, &mut buf[..], MsgFlags::MSG_DONTWAIT) {
+        Ok(0) | Err(Errno::EWOULDBLOCK) => return Ok(None),
+        Err(e) => return Err(e).context("failed to read from socket"),
+        Ok(n) => n,
+    };
+
+    let packet = &buf[..n];
+    const LEN: usize = size_of::<EthernetPacket<Ipv4Packet<TcpHeader>>>();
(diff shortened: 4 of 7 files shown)


faern (commented):
> Reviewed 2 of 7 files at r1, 2 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 4 of 11 files reviewed, 5 unresolved discussions (waiting on @hulthe)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 72 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1MUClD-FZ4Y39fc63U:-O1MUClD-FZ4Y39fc63V:b8tfig1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fd18f0bf8ef8730ded7ef36723abd434c61c5231/test/test-manager/src/tests/cve_2019_14899.rs#L72)):*
> ```XML
> #[repr(C, packed)]
> #[derive(Clone, Copy, Debug, Pod, Zeroable)]
> struct TcpHeader {
> ```

Did you evaluate `pnet_packet`? It really feels like we should not implement network header parsing at all, even less directly in a test :sweat_smile: https://docs.rs/pnet_packet/0.35.0/pnet_packet/tcp/struct.Tcp.html
___
*[`test/test-runner/Cargo.toml` line 61 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1MSB-677iMdZ3bti2K:-O1MSB-73B4ERa-dRl7i:b-bmdrpw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fd18f0bf8ef8730ded7ef36723abd434c61c5231/test/test-runner/Cargo.toml#L61)):*
> ```TOML
> 
> [target.'cfg(unix)'.dependencies]
> nix = { version = "0.26", features = ["socket", "net"] }
> ```

If you upgrade it in one crate, can you upgrade it in all crates in the test workspace? Maybe you can even uplift it to a workspace dependency so we keep it in sync? To help reduce the number of different versions we depend on.

If you do upgrade it, is there a reason you don't upgrade it to the latest? That would be preferable if not more complex.
___
*[`test/test-runner/src/main.rs` line 239 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1MSSyd2nla5gnPg37_:-O1MSSyd2nla5gnPg37a:b-1nrhex) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fd18f0bf8ef8730ded7ef36723abd434c61c5231/test/test-runner/src/main.rs#L239)):*
> ```XML
>         self,
>         _: context::Context,
>         interface: String,
> ```

Does this really need to be `String` when it looks like the value is just borrowed? Then it seems like it would be more idiomatic to accept `&str`, but I have not tried.
___
*[`test/test-runner/src/net.rs` line 229 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1MT-DY6xAVo1BKJ6-G:-O1MT-DY6xAVo1BKJ6-H:b-ru65a2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fd18f0bf8ef8730ded7ef36723abd434c61c5231/test/test-runner/src/net.rs#L229)):*
> ```XML
>             if let Some(address) = addr.address {
>                 if let Some(sockaddr) = address.as_link_addr() {
>                     return Ok(sockaddr.addr());
> ```

When does this ever return `Ok(None)` and is that a legitimate code path? What does it symbolize? From the only callsite I can find it looks like we just convert that into an error anyway. So can we squash this to just a `Result<[u8; 6]>`?
___
*[`test/test-runner/src/net.rs` line 235 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1MTH3lEdht-Cnexoqr:-O1MTH3lEdht-Cnexoqs:bqs8ibi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fd18f0bf8ef8730ded7ef36723abd434c61c5231/test/test-runner/src/net.rs#L235)):*
> ```XML
>     }
> 
>     log::error!("Could not find tunnel interface");
> ```

Nit: Maybe log `"Could not find interface {interface}"` instead of assuming it's a tunnel interface, nor log the name?

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 2 of 12 files reviewed, 4 unresolved discussions (waiting on @faern)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 72 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1MUClD-FZ4Y39fc63U:-O1ROvrP7NkyUfJ89loZ:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fd18f0bf8ef8730ded7ef36723abd434c61c5231/test/test-manager/src/tests/cve_2019_14899.rs#L72)):*

> feels like we should not implement network header parsing

Agreed


> pnet_packet

I might be blind, but I couldn't figure out how to create packets using pnet, only parse them.
</blockquote></details>

Done
___
*[`test/test-runner/Cargo.toml` line 61 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1MSB-677iMdZ3bti2K:-O1ROu6jCZIK6RicLO2X:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fd18f0bf8ef8730ded7ef36723abd434c61c5231/test/test-runner/Cargo.toml#L61)):*

I'll upgrade it, no problem.

Using 0.26 was just a quick, "i need this to work, and 0.26 is already in our dependency tree"
</blockquote></details>

Done
___
*[`test/test-runner/src/net.rs` line 235 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1MTH3lEdht-Cnexoqr:-O1ROs3EBETcfQhy_4m8:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fd18f0bf8ef8730ded7ef36723abd434c61c5231/test/test-runner/src/net.rs#L235)):*

Nit: Maybe log `"Could not find interface {interface}"` instead of assuming it's a tunnel interface, nor log the name?
</blockquote></details>

Done

faern (commented):
> Reviewed 2 of 7 files at r1, 2 of 7 files at r2, 5 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 11 of 12 files reviewed, 9 unresolved discussions (waiting on @hulthe)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 46 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1RzNoe5vHnIPbEhK4R:-O1RzNoe5vHnIPbEhK4S:b-xr0h41) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L46)):*
> ```XML
> /// Test mitigation for cve-2019-14899.
> ///
> /// The vulnerability allowed a malicious router learn the victims private mullvad tunnel IP.
> ```

"router *to* learn"
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 46 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1RznKkAlKca7tlieZX:-O1RznKl-Vvj_dVDAHWc:b9c3a9x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L46)):*
> ```XML
> /// Test mitigation for cve-2019-14899.
> ///
> /// The vulnerability allowed a malicious router learn the victims private mullvad tunnel IP.
> ```

Maybe clarify that only Linux was/is susceptible to this vulnerability?
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 73 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1S-qlw6iTwZHgrbFI0:-O1S-qlw6iTwZHgrbFI1:bcinbuw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L73)):*
> ```XML
> 
>     // Create a raw socket which let's us send custom ethernet packets
>     log::info!("Creating raw socket.");
> ```

A bit nitpicky, but we usually never end logging with a period or other punctuation. Multiple places in this PR.
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 126 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1S0y057afEQIWEEDC2:-O1S0y057afEQIWEEDC3:b21x1kq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L126)):*
> ```XML
>         // call the netdev ioctl that gets the interface index
>         ioctl_readwrite_bad!(get_interface_index, libc::SIOCGIFINDEX, libc::ifreq);
>         unsafe { get_interface_index(socket.as_raw_fd(), &mut if_index_request) }
> ```

Can we use https://docs.rs/libc/latest/libc/fn.if_nametoindex.html instead?
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 132 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1S1fG05J9iG6OEnuoU:-O1S1fG05J9iG6OEnuoV:bhivofg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L132)):*
> ```XML
>         // call the netdev ioctl that gets the interface mac address
>         ioctl_readwrite_bad!(get_interface_mac, libc::SIOCGIFHWADDR, libc::ifreq);
>         unsafe { get_interface_mac(socket.as_raw_fd(), &mut if_mac_request) }
> ```

What is the reason we can't use the `nix` crate for this, like we do in the other place in this PR.
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 138 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1S0cpl66W37RzBHGTF:-O1S0cpl66W37RzBHGTG:bwb30p8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L138)):*
> ```XML
>             &if_mac_request.ifr_ifru.ifru_hwaddr.sa_data[..6]
>         }));
>     }
> ```

Can this part maybe be moved out to a separate function? To pollute the test a bit less.
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 141 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1S1pNA9CL3hxmnctJi:-O1S1pNA9CL3hxmnctJj:b-3ns98q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L141)):*
> ```XML
> 
>     // craft a malicious packet.
>     let mut tcp_packet =
> ```

Can the crafting of a malicious packet be factored out to a function? The current test function is very long. If we can move stuff out to make the test itself more readable, that would be awesome.
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 149 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1S2A8z9lqsEK9o94yl:-O1S2A8z9lqsEK9o94ym:b-sxhjut) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L149)):*
> ```XML
>     tcp_packet.set_flags(TcpFlags::SYN | TcpFlags::ACK);
> 
>     let mut ip4_packet =
> ```

Nit, but I would name this `ipv4_packet`. `ip4` reads strangely in my head. Is there a specific reason you left out the `v`?
___
*[`test/test-runner/src/net.rs` line 229 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1MT-DY6xAVo1BKJ6-G:-O1Rxl0sBYis6dAwlVtv:b-sxdmy0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fd18f0bf8ef8730ded7ef36723abd434c61c5231/test/test-runner/src/net.rs#L229)):*

sockaddr.addr() can return None, so that would be the code path.

Ok(None) is intended to symbolize that the interface doesn't have a MAC address, which applies to e.g. tun interfaces
</blockquote></details>

Ok! Sounds fine.
___
*[`test/test-runner/src/net.rs` line 175 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1RxTkB6ui74MYFOT6T:-O1RxTkB6ui74MYFOT6U:b-wup09w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-runner/src/net.rs#L175)):*
> ```XML
> #[cfg(unix)]
> pub fn get_interface_ip(interface: &str) -> Result<IpAddr, test_rpc::Error> {
>     // TODO: IPv6
> ```

This todo doesn't really make any sense when the line below in went away. What is it that should be done? :P

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 11 of 13 files reviewed, 5 unresolved discussions (waiting on @hulthe)

___
*[`test/test-manager/src/tests/helpers.rs` line 258 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1Vr1h39brhc1KUE9LL:-O1Vr1h39brhc1KUE9LM:b-y5cq69) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/helpers.rs#L258)):*
> ```XML
> }
> 
> /// Get the mac address (if any) of a network interface.
> ```

This fetches data about network interfaces on the test *manager*, right? Can that be clarified in the documentation? Since it can be a bit hard to follow along with the testing code since it runs partially here, partially there and there are network interfaces on both sides.

Same holds for `get_interface_index` below
___
*[`test/test-manager/src/tests/helpers.rs` line 296 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1VrXKBDDJ8G1cTvYcg:-O1VrXKBDDJ8G1cTvYch:b-4ef4jn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/helpers.rs#L296)):*
> ```XML
>         0 => {
>             let err = Errno::last();
>             Err(anyhow!("Failed to get interface index").context(err))
> ```

Does this not create the error chain in the wrong order. Something like:
```
Error: No such interface
Caused by: Failed to get interface index
```
or am I reading this wrong?

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 12 of 13 files reviewed, 8 unresolved discussions (waiting on @hulthe)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 67 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1VyVWM9xscij6tB1N8:-O1VyVWN0NnnvZQpUrc5:b-1ebwem) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L67)):*
> ```XML
>         .set_allow_lan(true)
>         .await
>         .context("Failed to allow local network sharing")?;
> ```

Local network access is not a setting that triggers a reconnect. So this is fine! But I had to double check. Many of our settings trigger a reconnect. If so, it would be more logical and less error prone to first set all settings and then `connect_and_wait`. But I have not studied the other tests here, not sure what pattern is followed here in general.
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 70 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1W-8e187QNbUfTkHrM:-O1W-8e187QNbUfTkHrN:b-am9djy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L70)):*
> ```XML
> 
>     let host_interface = TAP_NAME;
>     let victim_tunnel_if = "wg0-mullvad";
> ```

Nit: I don't think it's ideal to mix between abbreviating and not. Some variables are called `interface` and some `if`. I vote for spelling them all out. And potentially also adding `_name` everywhere. There are so many identifiers for interfaces, explicitly calling them `interface_name` would not hurt, except that it's a bit longer, which is probably fine?
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 71 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1Vz9dSEi-HtelFJhg8:-O1Vz9dSEi-HtelFJhg9:bqtlrvy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L71)):*
> ```XML
>     let host_interface = TAP_NAME;
>     let victim_tunnel_if = "wg0-mullvad";
>     let gateway_ip = NON_TUN_GATEWAY;
> ```

Is this the victim's gateway? If so, I think the variable name should reflect that. All this code is extremely tricky since it handles the network setup of two different machines intermixed.
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 299 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1VwybJ8rIAwwg8lTPj:-O1VwybKDFN0W4Au_ycZ:bngc1tq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L299)):*
> ```XML
> 
>     let mut eth_packet =
>         MutableEthernetPacket::owned(vec![0u8; ETH_LEN]).expect("ETH_LEN bytes is enough");
> ```

This is not a blocker, I just wanted to bring it up. This reads slightly backwards in my head, but that could be a bias I have. This function requires three allocations and copy payloads upwards. If it's constructed the other way around we would need fewer `*_LEN` constants and allocations:

```
let mut ethernet_packet = MutableEthernetPacket::owned(vec![0u8; ETH_LEN).expect(...);
// set ethernet fields

let mut ipv4_packet = MutableIpv4Packet::new(ethernet_packet.payload()).unwrap();
// set ipv4 fields

let mut tcp_packet = MutableTcpPacket::new(ipv4_packet.payload()).unwrap();
// set tcp fields
```

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 12 of 13 files reviewed, 12 unresolved discussions (waiting on @hulthe)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 154 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1WkzaI9Si8FppUjgRY:-O1WkzaI9Si8FppUjgRZ:b-g9lad0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L154)):*
> ```XML
>     timeout: Duration,
> ) -> anyhow::Result<bool> {
>     let mut buf = vec![0u8; 0xffff];
> ```

What does `0xffff` signify here? Why is that hex value large enough for `buf`? I would argue that `u16::MAX as usize` is easier to understand.
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 163 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1WmzxB1SHe6aCShXFJ:-O1WmzxB1SHe6aCShXFK:bxb8o03) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L163)):*
> ```XML
>             match poll_for_packet(socket, &mut buf)? {
>                 Some(eth_packet) => {
>                     let Some(ip4_packet) = Ipv4Packet::new(eth_packet.payload()) else {
> ```

This check is already performed in `poll_for_packet`. Do we really need to do it in both places? Feels like this can be simplified one way or the other. Since this test only ever seems to use the innermost `TcpPacket` and not really need the outer layers, can't we make `poll_for_packet` return the `TcpPacket` directly?
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 238 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1Wl_s011O-WwrENkxc:-O1Wl_s011O-WwrENkxd:b-82eu3q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L238)):*
> ```XML
> }
> 
> fn poll_for_packet<'a>(
> ```

Nit, but order matters. `spam_packet` and `send_packet` are closely related and placed right after each other. Awesome! But `listen_for_packet` and `poll_for_packet` have the same relationship, but they are not placed right after each other. I think that would provide better code structure to put them in descending order of abstraction and as close as possible to the callsite.
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 249 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1WnFjC5Cofjmg5OASY:-O1WnFjC5Cofjmg5OASZ:b-n6kja6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L249)):*
> ```XML
> 
>     let packet = &buf[..n];
>     if packet.len() >= ETH_LEN {
> ```

This check is already built into `EthernetPacket::new`, which you unwrap below. Can't this be combined so we rely on the check in `EthernetPacket::new` instead of duplicating it?
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 299 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1VwybJ8rIAwwg8lTPj:-O1Wk0CFBRAWRthG5bn1:b-blf92r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L299)):*

This is not a blocker, I just wanted to bring it up. This reads slightly backwards in my head, but that could be a bias I have. This function requires three allocations and copy payloads upwards. If it's constructed the other way around we would need fewer `*_LEN` constants and allocations:

```
let mut ethernet_packet = MutableEthernetPacket::owned(vec![0u8; ETH_LEN).expect(...);
// set ethernet fields

let mut ipv4_packet = MutableIpv4Packet::new(ethernet_packet.payload()).unwrap();
// set ipv4 fields

let mut tcp_packet = MutableTcpPacket::new(ipv4_packet.payload()).unwrap();
// set tcp fields
```
</blockquote></details>

I just want to point out that I'm **not concerned about the performance of this code**! This is not at all in any hot path and I'm not scared of allocations. That was mostly a side-note. Mostly that I build packages outside in in my head probably.

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 11 of 13 files reviewed, 6 unresolved discussions (waiting on @faern)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 46 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1RznKkAlKca7tlieZX:-O1Wlj1VCRXeJrutnfMP:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L46)):*

Maybe clarify that only Linux was/is susceptible to this vulnerability?
</blockquote></details>

Done.
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 126 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1S0y057afEQIWEEDC2:-O1WpKdV0Yr0G1ucgnqw:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L126)):*

Can we use https://docs.rs/libc/latest/libc/fn.if_nametoindex.html instead?
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 132 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1S1fG05J9iG6OEnuoU:-O1WpL9T-FtLyS2sTOTQ:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L132)):*

What is the reason we can't use the `nix` crate for this, like we do in the other place in this PR.
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 138 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1S0cpl66W37RzBHGTF:-O1WpMDh-h7I3d3vwFQ6:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L138)):*

Can this part maybe be moved out to a separate function? To pollute the test a bit less.
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 141 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1S1pNA9CL3hxmnctJi:-O1WpNIP0jOXBnXeBlo8:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L141)):*

Can the crafting of a malicious packet be factored out to a function? The current test function is very long. If we can move stuff out to make the test itself more readable, that would be awesome.
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 149 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1S2A8z9lqsEK9o94yl:-O1WpNqNBecDHBuf6qpx:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-manager/src/tests/cve_2019_14899.rs#L149)):*

Nit, but I would name this `ipv4_packet`. `ip4` reads strangely in my head. Is there a specific reason you left out the `v`?
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 71 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1Vz9dSEi-HtelFJhg8:-O1WwKtsEjxgQhAnnMxv:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L71)):*

Is this the victim's gateway? If so, I think the variable name should reflect that. All this code is extremely tricky since it handles the network setup of two different machines intermixed.
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 154 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1WkzaI9Si8FppUjgRY:-O1WwZ0x73JDIW6Dntb4:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L154)):*

What does `0xffff` signify here? Why is that hex value large enough for `buf`? I would argue that `u16::MAX as usize` is easier to understand.
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 163 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1WmzxB1SHe6aCShXFJ:-O1WpWda5NW4O5irMlyA:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L163)):*

This check is already performed in `poll_for_packet`. Do we really need to do it in both places? Feels like this can be simplified one way or the other. Since this test only ever seems to use the innermost `TcpPacket` and not really need the outer layers, can't we make `poll_for_packet` return the `TcpPacket` directly?
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 238 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1Wl_s011O-WwrENkxc:-O1WwCrI6GH4Jn3Xh6Qk:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L238)):*

Nit, but order matters. `spam_packet` and `send_packet` are closely related and placed right after each other. Awesome! But `listen_for_packet` and `poll_for_packet` have the same relationship, but they are not placed right after each other. I think that would provide better code structure to put them in descending order of abstraction and as close as possible to the callsite.
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 249 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1WnFjC5Cofjmg5OASY:-O1WwDXm0tro-wi4Eyay:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L249)):*

This check is already built into `EthernetPacket::new`, which you unwrap below. Can't this be combined so we rely on the check in `EthernetPacket::new` instead of duplicating it?
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 299 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1VwybJ8rIAwwg8lTPj:-O1WwF7x2AuI8GoA8eFC:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L299)):*

I just want to point out that I'm **not concerned about the performance of this code**! This is not at all in any hot path and I'm not scared of allocations. That was mostly a side-note. Mostly that I build packages outside in in my head probably.
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/helpers.rs` line 258 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1Vr1h39brhc1KUE9LL:-O1WyNOm4uNJepSIFI6X:bplppzj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/helpers.rs#L258)):*

This fetches data about network interfaces on the test *manager*, right? Can that be clarified in the documentation? Since it can be a bit hard to follow along with the testing code since it runs partially here, partially there and there are network interfaces on both sides.

Same holds for `get_interface_index` below
</blockquote></details>

Added a note to the docstring.
___
*[`test/test-manager/src/tests/helpers.rs` line 296 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1VrXKBDDJ8G1cTvYcg:-O1WxF633XKVwqZb53wL:b-r1zgaa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/helpers.rs#L296)):*

Does this not create the error chain in the wrong order. Something like:
```
Error: No such interface
Caused by: Failed to get interface index
```
or am I reading this wrong?
</blockquote></details>

It sure does. Fixed
___
*[`test/test-runner/src/net.rs` line 175 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1RxTkB6ui74MYFOT6T:-O1SAjJs3vIGsR10L3jd:bsv1gca) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-runner/src/net.rs#L175)):*

This todo doesn't really make any sense when the line below in went away. What is it that should be done? :P
</blockquote></details>

I think the todo refers to the function as a whole, which only deals in ipv4.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 10 of 13 files reviewed, 5 unresolved discussions (waiting on @hulthe)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 154 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1WkzaI9Si8FppUjgRY:-O1WyxRw43sxj1Jh6hcL:bnv3u4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L154)):*

Done
</blockquote></details>

https://github.com/mullvad/coding-guidelines/blob/0afd7c0acf4b1227f7924f783c2e459182a422de/rust.md#prefer-from-over-into

`usize` implements `From<u16>`.
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 160 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1WzMO0CzPb5xPtTIn0:-O1WzMO0CzPb5xPtTIn1:b-qn0jmy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f1e8cba2bb1b16e781df70cd16ca67baf5013a2/test/test-manager/src/tests/cve_2019_14899.rs#L160)):*
> ```XML
>                 Some(tcp_packet) if filter(&tcp_packet) => break,
>                 Some(_other_packet) => continue,
>                 None => sleep(Duration::from_millis(10)).await,
> ```

Since `poll_for_packet` returns `Ok(None)` when the read packet is not a TCP packet, this loop could end up in many sleeps if the machine receives a bunch of non-TCP packet. Say the test host receives 1000 ICMP packets while this test runs, then this test would get stuck for 10 seconds/time out.

I think the code should be designed to sleep only when the raw socket has no packet at all. That prevents this flakiness.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 10 of 13 files reviewed, 4 unresolved discussions (waiting on @hulthe)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 299 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1VwybJ8rIAwwg8lTPj:-O1X3yTS8YuQjPeMZDg-:b-83sjoh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L299)):*

Done
</blockquote></details>

Nice!

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 10 of 13 files reviewed, 4 unresolved discussions (waiting on @hulthe)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 211 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1X4g49E85dfJY30vLC:-O1X4g49E85dfJY30vLD:bkaaiu2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f1e8cba2bb1b16e781df70cd16ca67baf5013a2/test/test-manager/src/tests/cve_2019_14899.rs#L211)):*
> ```XML
>     interface_index: c_uint,
>     packet: &EthernetPacket<'_>,
> ) -> anyhow::Result<()> {
> ```

Strictly this should return `Result<Infallible, ...>`. Oh dear lord the day when we have stable never type I will cry of happiness :pray:

faern (commented):
> Reviewed 1 of 1 files at r7, 1 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 12 of 13 files reviewed, 2 unresolved discussions (waiting on @hulthe)

___
*[`test/test-runner/src/net.rs` line 175 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1RxTkB6ui74MYFOT6T:-O1aDPG-ExYJrenalsCx:bqsbzst) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9628d895818f27eef2835a868590cd3bdab6d03c/test/test-runner/src/net.rs#L175)):*

I think the todo refers to the function as a whole, which only deals in ipv4.
</blockquote></details>

Aha. I would then have put it above the entire function. But oh well, thanks for explaining, let's move on.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 12 of 13 files reviewed, 3 unresolved discussions (waiting on @hulthe)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 160 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1WzMO0CzPb5xPtTIn0:-O1b3E7X81yorw5C4qUe:bymegcs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f1e8cba2bb1b16e781df70cd16ca67baf5013a2/test/test-manager/src/tests/cve_2019_14899.rs#L160)):*

Since `poll_for_packet` returns `Ok(None)` when the read packet is not a TCP packet, this loop could end up in many sleeps if the machine receives a bunch of non-TCP packet. Say the test host receives 1000 ICMP packets while this test runs, then this test would get stuck for 10 seconds/time out.

I think the code should be designed to sleep only when the raw socket has no packet at all. That prevents this flakiness.
</blockquote></details>

This is much better IMO!
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 167 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1b3M9BA1npUF8_U_za:-O1b3M9BA1npUF8_U_zb:b-pmnnq8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d093e1f438ad011514d8edbb85a76849392eb906/test/test-manager/src/tests/cve_2019_14899.rs#L167)):*
> ```XML
>     select! {
>         result = wait_for_packet.fuse() => result.map(Some),
>         _ = sleep(timeout).fuse() => Ok(None),
> ```

Selecting over the desired result and a sleep is a pretty blunt way of adding a timeout to any future. Maybe look at `tokio::time::timeout` for implementing this with existing types?
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 206 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1b2hV61KUeOLh1XPpO:-O1b2hV61KUeOLh1XPpP:bsu35u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d093e1f438ad011514d8edbb85a76849392eb906/test/test-manager/src/tests/cve_2019_14899.rs#L206)):*
> ```XML
>         let protocol_is_tcp = ipv4_packet.get_next_level_protocol() == IpNextHeaderProtocols::Tcp;
> 
>         if !valid_ip_version || !protocol_is_tcp {
> ```

You might also want to check in the `eth_packet` so its header says the payload actually is IPv4. Otherwise you might end up parsing IPv4 headers from something that is not an IPv4 packet.

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 10 of 13 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 154 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1WkzaI9Si8FppUjgRY:-O1b4gWv3Zd43MEy53y_:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0213df3b5555b2f94c755a0edfa270cc41bbcb35/test/test-manager/src/tests/cve_2019_14899.rs#L154)):*

https://github.com/mullvad/coding-guidelines/blob/0afd7c0acf4b1227f7924f783c2e459182a422de/rust.md#prefer-from-over-into

`usize` implements `From<u16>`.
</blockquote></details>

Done
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 160 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1WzMO0CzPb5xPtTIn0:-O1b4jHTEL1Ix4sgF9qO:b-8fx2u7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f1e8cba2bb1b16e781df70cd16ca67baf5013a2/test/test-manager/src/tests/cve_2019_14899.rs#L160)):*

This is much better IMO!
</blockquote></details>

:+1: 
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 211 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1X4g49E85dfJY30vLC:-O1a2kUmDWmcOGoYHW__:b37l3mw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8f1e8cba2bb1b16e781df70cd16ca67baf5013a2/test/test-manager/src/tests/cve_2019_14899.rs#L211)):*

Strictly this should return `Result<Infallible, ...>`. Oh dear lord the day when we have stable never type I will cry of happiness :pray:
</blockquote></details>

I instinctively tried `Result<!>` when i wrote this, with the dream that, somehow, Never had been stabilized without me knowing. Alas, it was not to be :'(
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 167 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1b3M9BA1npUF8_U_za:-O1b9KQ8EwGiid8QwYlt:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d093e1f438ad011514d8edbb85a76849392eb906/test/test-manager/src/tests/cve_2019_14899.rs#L167)):*

Selecting over the desired result and a sleep is a pretty blunt way of adding a timeout to any future. Maybe look at `tokio::time::timeout` for implementing this with existing types?
</blockquote></details>

Done.
___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 206 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1b2hV61KUeOLh1XPpO:-O1b9Lio4gouRUHJ53pA:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d093e1f438ad011514d8edbb85a76849392eb906/test/test-manager/src/tests/cve_2019_14899.rs#L206)):*

You might also want to check in the `eth_packet` so its header says the payload actually is IPv4. Otherwise you might end up parsing IPv4 headers from something that is not an IPv4 packet.
</blockquote></details>

Done.

faern (commented):
> Reviewed 2 of 4 files at r10.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @hulthe)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 134 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1vIUloEdBGlsGTOglK:-O1vIUloEdBGlsGTOglL:beo3raz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3a4621280b933bf4ef1510261615af59a772c644/test/test-manager/src/tests/cve_2019_14899.rs#L134)):*
> ```XML
> 
>     let rst_packet = select! {
>         result = spam_packet(&socket, host_interface_index, &malicious_packet).fuse() => return result,
> ```

If `spam_packet` were to return `Ok(())`, this test would succeed. The current implementation of `spam_packet` probably won't do that, but it could be an easy mistake to make. I suggest making this match bails out with a hard no, unconditionally. Also potentially making `spam_packet` return `Result<Infallible>`.

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`test/test-manager/src/tests/cve_2019_14899.rs` line 134 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447#-O1vIUloEdBGlsGTOglK:-O1vRzfG01G92YcJLWjo:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3a4621280b933bf4ef1510261615af59a772c644/test/test-manager/src/tests/cve_2019_14899.rs#L134)):*

If `spam_packet` were to return `Ok(())`, this test would succeed. The current implementation of `spam_packet` probably won't do that, but it could be an easy mistake to make. I suggest making this match bails out with a hard no, unconditionally. Also potentially making `spam_packet` return `Result<Infallible>`.
</blockquote></details>

Done.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (commented):
> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (approved):
> Reviewed 2 of 2 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (approved):
> Reviewed 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6447)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6448, opened 2024-07-05
Add negative leak test

niklasberglund:
Add a negative leak test. This is a draft PR. The implementation should almost be there, but the main issue with the test right now seems to be that `getConnectedThroughRelayDateInterval` is returning a longer date interval for the relay connection than expected. I believe its a bug in the test code but it could also be that the dummy data is incorrect. Had to work with dummy data because of the ongoing move. 

To try it out run `testNegativeLeaks` under `LeakTests`. The test device must be connected to the `app-team-ios-tests` WiFi to be able to reach the packet capture API and firewall API.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6448)


niklasberglund:
Replaced by https://github.com/mullvad/mullvadvpn-app/pull/6449

PR #6455, opened 2024-07-09
Prevent user from selecting same entry and exit relay for multihop

rablador:
When displaying either the entry or exit location lists, any location (country, city or relay) shouldn't be selectable if, after apply all other constraints, the location item resolves to a single candidate that is already selected by the entry.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6455)

### DIFF
--- a/ios/MullvadVPN/Coordinators/CustomLists/AddLocationsDataSource.swift
+++ b/ios/MullvadVPN/Coordinators/CustomLists/AddLocationsDataSource.swift
@@ -161,10 +161,6 @@ extension AddLocationsDataSource {
     func nodeShouldBeSelected(_ node: LocationNode) -> Bool {
         customListLocationNode.children.contains(node)
     }
-
-    func excludedRelayTitle(_ node: LocationNode) -> String? {
-        nil // N/A
-    }
 }
 
 // MARK: - Toggle selection in table view
--- a/ios/MullvadVPN/View controllers/SelectLocation/LocationCell.swift
+++ b/ios/MullvadVPN/View controllers/SelectLocation/LocationCell.swift
@@ -209,7 +209,6 @@ class LocationCell: UITableViewCell {
 
     private func updateDisabled(_ isDisabled: Bool) {
         locationLabel.alpha = isDisabled ? 0.2 : 1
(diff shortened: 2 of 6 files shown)


rablador (commented):
The exclusion is all done in `tableView(cellForRowAt)`. The main reason for this is that the function is very reliable if you want to apply changes during every table update without rebuilding the underlying data set. Since we want to keep the UI state (unfolded cells, position etc) while toggling multihop this is by far the least complex place to "patch" cell contents. We could let the diffable data source handle it for us, but since 1) cell content isn't updated unless the underlying data is changed (which would require adding exclusion state in the nodes, which in turn requires more complexity when populating and manipulating the node tree) and 2) the view state needs to be tracked and rebuilt every time, I opted for the easiest method, knowing that we want to make changes to multihop in the near future anyway.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6455#-:-O1JtKc211DsXSGDgFte:bji9ug8)* status: 0 of 8 files reviewed, all discussions resolved

rablador (commented):
Also, looking at my local git history, Emils' `c86bcc3` commit is on top of main. I have rebased from main just now but the commit still wants to be included in this PR for some reason.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6455#-:-O1Jw0s53l6Vjb6h0xp6:b-qejaeb)* status: 0 of 8 files reviewed, all discussions resolved

acb-mv (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6455#-:-O1TT_eXEqhRIs3vvUyc:bnfp4nl)* status: 1 of 7 files reviewed, all discussions resolved

mojganii (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6455#-:-O1WDKSs9GYgVBfsbrsm:bnfp4nl)* status: 1 of 7 files reviewed, all discussions resolved

PR #6469, opened 2024-07-16
Fix browser download script

faern:
A missing `-f` made the `ln` command exit with a `File exists` error. I don't think this script was ever tested with upgrades when the destination already existed :sweat_smile:

Except from crippling the download script from actually publishing new browsers, this also left a ton of temporary downloads in `/tmp/mullvad-browser-tmp-.....`. I fixed this my moving the cleanup to a signal trap. This means the cleanup code always run on script exit, even if it exits prematurely due to a failing command.

I took the opportunity to do two other minor cleanups of the scripts here also:

1. Stop `wget` from printing very verbose progress. This filled up the systemd logs. The progress is not needed.
2. Store the latest known browser artifacts in the script dir instead of `/tmp`. This keeps the data closer to the script. This results in `/tmp` only being used for actual temporary files created and deleted during a single run of this script.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6469)

### DIFF
--- a/ci/mullvad-browser/download-mullvad-browser.sh
+++ b/ci/mullvad-browser/download-mullvad-browser.sh
@@ -65,7 +65,7 @@ function main() {
     fi
 
     echo "[#] $PACKAGE_FILENAME has changed"
-    ln -f "$PACKAGE_FILENAME" $WORKDIR/
+    ln -f "$PACKAGE_FILENAME" "$WORKDIR/"
 }
 
 if [[ ${1:-} == "-h" ]] || [[ ${1:-} == "--help" ]]; then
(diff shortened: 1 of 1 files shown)


Serock3 (approved):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6469)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6475, opened 2024-07-17
Make test-manager CLI simpler

Serock3:
Includes a few quality of life improvments, most importantly making previous app and GUI e2e test packages optional.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475)

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -115,7 +115,7 @@ cargo run --bin test-manager run-vm debian11
 cargo run --bin test-manager run-tests debian11 \
     --display \
     --account 0123456789 \
-    --current-app <git hash or tag> \
+    --app-package <git hash or tag> \
     --previous-app 2023.2
 ```
 
@@ -141,7 +141,7 @@ cargo run --bin test-manager set macos-ventura tart ventura-base macos \
 cargo run --bin test-manager run-tests macos-ventura \
     --display \
     --account 0123456789 \
-    --current-app <git hash or tag> \
+    --app-package <git hash or tag> \
     --previous-app 2023.2
 ```
 
--- a/test/README.md
+++ b/test/README.md
@@ -116,7 +116,7 @@ cargo run --bin test-manager run-tests debian11 \
     --display \
     --account 0123456789 \
     --app-package <git hash or tag> \
-    --previous-app 2023.2
+    --app-package-to-upgrade-from 2023.2
 ```
 
 ## macOS
@@ -142,7 +142,7 @@ cargo run --bin test-manager run-tests macos-ventura \
     --display \
     --account 0123456789 \
     --app-package <git hash or tag> \
-    --previous-app 2023.2
+    --app-package-to-upgrade-from 2023.2
 ```
 
 ## Note on `ci-runtests.sh`
--- a/test/ci-runtests.sh
+++ b/test/ci-runtests.sh
@@ -198,7 +198,7 @@ function run_tests_for_os {
         run-tests \
         --account "${ACCOUNT_TOKEN:?Error: ACCOUNT_TOKEN not set}" \
         --app-package "${cur_filename}" \
-        --previous-app "${prev_filename}" \
+        --app-package-to-upgrade-from "${prev_filename}" \
         --package-folder "$PACKAGES_DIR" \
         --test-report "$SCRIPT_DIR/.ci-logs/${os}_report" \
         "$os" 2>&1 | sed "s/${ACCOUNT_TOKEN}/\{ACCOUNT_TOKEN\}/g"
--- a/test/ci-runtests.sh
+++ b/test/ci-runtests.sh
@@ -197,8 +197,9 @@ function run_tests_for_os {
     RUST_LOG=debug cargo run --bin test-manager \
         run-tests \
         --account "${ACCOUNT_TOKEN:?Error: ACCOUNT_TOKEN not set}" \
-        --current-app "${cur_filename}" \
+        --app-package "${cur_filename}" \
         --previous-app "${prev_filename}" \
+        --package-folder "$PACKAGES_DIR" \
         --test-report "$SCRIPT_DIR/.ci-logs/${os}_report" \
         "$os" 2>&1 | sed "s/${ACCOUNT_TOKEN}/\{ACCOUNT_TOKEN\}/g"
 }
--- a/test/scripts/ssh-setup.sh
+++ b/test/scripts/ssh-setup.sh
@@ -6,7 +6,7 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd "$SCRIPT_DIR"
 
 RUNNER_DIR="$1"
-CURRENT_APP="$2"
+APP_PACKAGE="$2"
 PREVIOUS_APP="$3"
 UI_RUNNER="$4"
 
@@ -16,7 +16,7 @@ echo "Copying test-runner to $RUNNER_DIR"
 
 mkdir -p "$RUNNER_DIR"
 
-for file in test-runner connection-checker $CURRENT_APP $PREVIOUS_APP $UI_RUNNER openvpn.ca.crt; do
+for file in test-runner connection-checker $APP_PACKAGE $PREVIOUS_APP $UI_RUNNER openvpn.ca.crt; do
     echo "Moving $file to $RUNNER_DIR"
     cp -f "$SCRIPT_DIR/$file" "$RUNNER_DIR"
 done
--- a/test/test-manager/src/container.rs
+++ b/test/test-manager/src/container.rs
@@ -28,7 +28,9 @@ pub async fn relaunch_with_rootlesskit(vnc_port: Option<u16>) {
 
     cmd.args(std::env::args());
 
-    let status = cmd.status().await.unwrap();
+    let status = cmd.status().await.unwrap_or_else(|e| {
+        panic!("failed to execute [{:?}]: {}", cmd, e);
+    });
 
     std::process::exit(status.code().unwrap_or(1));
 }
--- a/test/test-manager/src/logging.rs
+++ b/test/test-manager/src/logging.rs
@@ -30,7 +30,7 @@ impl Logger {
             logger.filter_module("tower", log::LevelFilter::Info);
             logger.filter_module("hyper", log::LevelFilter::Info);
             logger.filter_module("rustls", log::LevelFilter::Info);
-            logger.filter_level(log::LevelFilter::Debug);
+            logger.filter_level(log::LevelFilter::Info);
             logger.parse_env(env_logger::DEFAULT_FILTER_ENV);
 
             let env_logger = logger.build();
--- a/test/test-manager/src/main.rs
+++ b/test/test-manager/src/main.rs
@@ -262,7 +262,6 @@ async fn main() -> Result<()> {
 
             let manifest =
                 package::get_app_manifest(vm_config, current_app, previous_app, package_folder)
-                    .await
                     .context("Could not find the specified app packages")?;
 
             let mut instance = vm::run(&config, &name)
--- a/test/test-manager/src/main.rs
+++ b/test/test-manager/src/main.rs
@@ -79,7 +79,15 @@ enum Commands {
         #[arg(long, short)]
         account: String,
 
-        /// App package to test.
+        /// App package to test. Can be a path to the package, just the package file name, git hash
+        /// or tag. If the direct path is not given, the package is assumed to be in the cache
+        /// directory for the host OS, given by the following table:
+        ///
+        /// |Platform | Value                               | Example                      |
+        /// | ------- | ----------------------------------- | ---------------------------- |
+        /// | Linux   | `$XDG_CACHE_HOME` or `$HOME`/.cache | /home/alice/.cache           |
+        /// | macOS   | `$HOME`/Library/Caches              | /Users/Alice/Library/Caches  |
+        /// | Windows | `{FOLDERID_LocalAppData}`           | C:\Users\Alice\AppData\Local |
         ///
         /// # Note
         ///
@@ -88,7 +96,8 @@ enum Commands {
         #[arg(long, short)]
         current_app: String,
 
-        /// App package to upgrade from.
+        /// App package to upgrade from when running `test_install_previous_app`, can be left empty
+        /// if this test is not ran. Parsed the same way as `--current-app`.
         ///
         /// # Note
         ///
--- a/test/test-manager/src/main.rs
+++ b/test/test-manager/src/main.rs
@@ -97,7 +97,7 @@ enum Commands {
         ///
         /// The CLI interface must be compatible with the upgrade test.
         #[arg(long)]
-        previous_app: Option<String>,
+        app_package_to_upgrade_from: Option<String>,
 
         /// Folder to search for packages. Defaults to current directory.
         #[arg(long, value_name = "DIR")]
@@ -225,7 +225,7 @@ async fn main() -> Result<()> {
             vnc,
             account,
             app_package,
-            previous_app,
+            app_package_to_upgrade_from,
             package_folder,
             test_filters,
             verbose,
@@ -260,9 +260,13 @@ async fn main() -> Result<()> {
                 None => None,
             };
 
-            let manifest =
-                package::get_app_manifest(vm_config, app_package, previous_app, package_folder)
-                    .context("Could not find the specified app packages")?;
+            let manifest = package::get_app_manifest(
+                vm_config,
+                app_package,
+                app_package_to_upgrade_from,
+                package_folder,
+            )
+            .context("Could not find the specified app packages")?;
 
             let mut instance = vm::run(&config, &name)
                 .await
@@ -290,8 +294,8 @@ async fn main() -> Result<()> {
                         .unwrap()
                         .to_string_lossy()
                         .into_owned(),
-                    previous_app_filename: manifest
-                        .previous_app_path
+                    app_package_to_upgrade_from_filename: manifest
+                        .app_package_to_upgrade_from_path
                         .map(|path| path.file_name().unwrap().to_string_lossy().into_owned()),
                     ui_e2e_tests_filename: manifest
                         .ui_e2e_tests_path
--- a/test/test-manager/src/main.rs
+++ b/test/test-manager/src/main.rs
@@ -80,14 +80,8 @@ enum Commands {
         account: String,
 
         /// App package to test. Can be a path to the package, just the package file name, git hash
-        /// or tag. If the direct path is not given, the package is assumed to be in the cache
-        /// directory for the host OS, given by the following table:
-        ///
-        /// |Platform | Value                               | Example                      |
-        /// | ------- | ----------------------------------- | ---------------------------- |
-        /// | Linux   | `$XDG_CACHE_HOME` or `$HOME`/.cache | /home/alice/.cache           |
-        /// | macOS   | `$HOME`/Library/Caches              | /Users/Alice/Library/Caches  |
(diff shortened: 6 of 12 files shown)


faern (changes requested):
> Reviewed 6 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475)* status: 4 of 11 files reviewed, 6 unresolved discussions (waiting on @Serock3)

___
*[`test/test-manager/src/main.rs` line 97 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24QHCt7VpM-LkeTDlW:-O24QHCt7VpM-LkeTDlX:b-1pbeia) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/main.rs#L97)):*
> ```XML
>         /// The CLI interface must be compatible with the upgrade test.
>         #[arg(long, short)]
>         previous_app: Option<String>,
> ```

Can the docs be updated to reflect what happens if this is not specified? Given that it's part of the CLI interface docs, it makes sense to document it pretty well.
___
*[`test/test-manager/src/package.rs` line 38 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24UcwJ4OFa51YPSoxz:-O24UcwJ4OFa51YPSoy-:b-axat9l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/package.rs#L38)):*
> ```XML
>         log::warn!("No previous app version specified");
>         None
>     };
> ```

If the logging is not considered important, this could be simplified to just `previous_app.map(|app| find_app(app, false, package_type)`
___
*[`test/test-manager/src/package.rs` line 51 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24VHnb8fZTV84spEZu:-O24VHnb8fZTV84spEZv:b-nck8cq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/package.rs#L51)):*
I find "graphics" to be a suboptimal wording here. It's the electron app e2e tests that can't run, right? That they might contain graphics is just an unrelated property is it not? Any e2e test we write in our electron app will be unable to run whether or not they test something that is primarily visual or not.

How about just:
> Could not find UI e2e test binary.
___
*[`test/test-manager/src/vm/provision.rs` line 124 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24YKz_3aluin5MTR_C:-O24YKz_3aluin5MTR_D:b3jlh5s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/vm/provision.rs#L124)):*
The question I ask here is "Why is it not copied?!". Saying not copying X implies there is an X and it was decided to not copy it. But here it does not exist. How about something like:

> No previous app to send to remote
___
*[`test/test-manager/src/vm/provision.rs` line 130 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24Yb_oCtH9v6wsV-hK:-O24Yb_oCtH9v6wsV-hL:b2nreh6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/vm/provision.rs#L130)):*
> ```XML
>             .context("Failed to send ui_e2e_tests_path to remote")?;
>     } else {
>         log::warn!("Not copying GUI e2e test to remote")
> ```

Same as above.
___
*[`test/test-manager/src/vm/provision.rs` line 173 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24ZbWq26-_oB6thz_d:-O24ZbWq26-_oB6thz_e:b-m9i3uz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/vm/provision.rs#L173)):*
> ```XML
>             .ui_e2e_tests_path
>             .map(|path| path.file_name().unwrap().to_string_lossy().into_owned())
>             .unwrap_or_default(),
> ```

Can we lift these out to variables instead of inlining them as arguments directly in `format!`? They are pretty long expressions and it's not trivial to see how they differ/what they mean. By assigning them to a variable with a nice name it becomes easier to see what `args` will end up containing/meaning.

faern (changes requested):
> Reviewed 1 of 7 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475)* status: 7 of 11 files reviewed, 9 unresolved discussions (waiting on @Serock3)

___
*[`test/ci-runtests.sh` line 201 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24kU3N4B6QUUb6ysnp:-O24kU3O9YBCxfYLDpBO:bkyz0ct) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d5bd432ecaadd926aa60e6f5e6264aa2efa4d40b/test/ci-runtests.sh#L201)):*
> ```Shell
>         --account "${ACCOUNT_TOKEN:?Error: ACCOUNT_TOKEN not set}" \
>         --current-app "${cur_filename}" \
>         --old-app "${prev_filename}" \
> ```

Why is "old" better than "previous"? I'd argue that old is less specific and more vague. I guess the meaning of this in reality is "For tests that involve upgrading from one version of the app to `current-app`, this specifies the version of the app to upgrade *from*"?
___
*[`test/test-manager/src/container.rs` line 32 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24jw2AE5c_jMllxnQn:-O24jw2AE5c_jMllxnQo:b-qj730m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/288168705ebaf6d53cf23d820258588719e0933a/test/test-manager/src/container.rs#L32)):*
> ```XML
> 
>     let status = cmd.status().await.unwrap_or_else(|e| {
>         panic!("failed to execute [{:?}]: {}", cmd, e);
> ```

Is this not equivalent to using `.expect()` with the same message?
___
*[`test/test-manager/src/logging.rs` line 33 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24k7A297O8t_9-K8z5:-O24k7A297O8t_9-K8z6:bkxvoxp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/542bdf77975eae321f40fedb15dd06ebf222614e/test/test-manager/src/logging.rs#L33)):*
> ```XML
>             logger.filter_module("hyper", log::LevelFilter::Info);
>             logger.filter_module("rustls", log::LevelFilter::Info);
>             logger.filter_level(log::LevelFilter::Info);
> ```

This seems to change the logging level of the entire test-manager? Why is this desired? It seems unrelated to simplifying the CLI
___
*[`test/test-manager/src/main.rs` line 97 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24QHCt7VpM-LkeTDlW:-O24jee51awAWOj7Luzf:bdjz9b9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/main.rs#L97)):*

Can the docs be updated to reflect what happens if this is not specified? Given that it's part of the CLI interface docs, it makes sense to document it pretty well.
</blockquote></details>

Sorry. I did review this per commit. So when I left this comment I did not see that you later added these docs.

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475)* status: 6 of 11 files reviewed, 8 unresolved discussions (waiting on @faern)

___
*[`test/ci-runtests.sh` line 201 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24kU3N4B6QUUb6ysnp:-O24uVLhB9Kbya8IMkeo:b-hr4cpw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d5bd432ecaadd926aa60e6f5e6264aa2efa4d40b/test/ci-runtests.sh#L201)):*

Why is "old" better than "previous"? I'd argue that old is less specific and more vague. I guess the meaning of this in reality is "For tests that involve upgrading from one version of the app to `current-app`, this specifies the version of the app to upgrade *from*"?
</blockquote></details>

I don't prefer "old", but clap doesn't allow short forms for flags that start with the same letter. Alternatively we could rename `--package-folder` to something starting with another letter, but I couldn't come up with anything good.
___
*[`test/test-manager/src/container.rs` line 32 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24jw2AE5c_jMllxnQn:-O24uhmMC-Ur0kNBrhAf:b-dtjdam) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/288168705ebaf6d53cf23d820258588719e0933a/test/test-manager/src/container.rs#L32)):*

Is this not equivalent to using `.expect()` with the same message?
</blockquote></details>

If you put a `format!` inside `.expect()` then clippy complains about `use of 'expect' followed by a function call`, since it's eagerly evaluated.
___
*[`test/test-manager/src/logging.rs` line 33 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24k7A297O8t_9-K8z5:-O24vq5y9qmT8D3daoM-:b-895tk9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/542bdf77975eae321f40fedb15dd06ebf222614e/test/test-manager/src/logging.rs#L33)):*

This seems to change the logging level of the entire test-manager? Why is this desired? It seems unrelated to simplifying the CLI
</blockquote></details>

You can still specify the log level using `RUST_LOG`, which is done by `ci-runtests.sh` so its output will be the same. I just found the logs too verbose for my taste when running manually. The debug level logs are not really relevant to me when I am working on new e2e tests, so I consider it a more sane default. In general I think `INFO` should be the default for logs.
___
*[`test/test-manager/src/package.rs` line 38 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24UcwJ4OFa51YPSoxz:-O24n89P21XC8XzFTcbk:b-o8agfo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/package.rs#L38)):*

If the logging is not considered important, this could be simplified to just `previous_app.map(|app| find_app(app, false, package_type)`
</blockquote></details>

I opted for the cautious approach and added a warning, just in case someone forgets to set the previous app version now that it's optional. It will eventually give a pretty clear error message anyway, so it may be superfluous. I'll keep it unless you strongly think logging is unimportant.
___
*[`test/test-manager/src/package.rs` line 51 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24VHnb8fZTV84spEZu:-O24oqUr2dlR0jVy-fLt:bb74ce2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/package.rs#L51)):*

I find "graphics" to be a suboptimal wording here. It's the electron app e2e tests that can't run, right? That they might contain graphics is just an unrelated property is it not? Any e2e test we write in our electron app will be unable to run whether or not they test something that is primarily visual or not.

How about just:
> Could not find UI e2e test binary.
</blockquote></details>

Sure
___
*[`test/test-manager/src/vm/provision.rs` line 124 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24YKz_3aluin5MTR_C:-O24p_eF0y2-X3UQwrUj:b-8ayu6x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/vm/provision.rs#L124)):*

The question I ask here is "Why is it not copied?!". Saying not copying X implies there is an X and it was decided to not copy it. But here it does not exist. How about something like:

> No previous app to send to remote
</blockquote></details>

Agree
___
*[`test/test-manager/src/vm/provision.rs` line 130 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24Yb_oCtH9v6wsV-hK:-O24pmDs7PfXso4_3FKH:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/vm/provision.rs#L130)):*

Same as above.
</blockquote></details>

Done.
___
*[`test/test-manager/src/vm/provision.rs` line 173 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24ZbWq26-_oB6thz_d:-O24tMK55F1b-r0xiXkK:b-9f6xi7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/vm/provision.rs#L173)):*

Can we lift these out to variables instead of inlining them as arguments directly in `format!`? They are pretty long expressions and it's not trivial to see how they differ/what they mean. By assigning them to a variable with a nice name it becomes easier to see what `args` will end up containing/meaning.
</blockquote></details>

Sure. I did that and refactored a bit.

faern (changes requested):
> Reviewed 1 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475)* status: 7 of 11 files reviewed, 1 unresolved discussion (waiting on @Serock3)

___
*[`test/ci-runtests.sh` line 201 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24kU3N4B6QUUb6ysnp:-O24yrwK8EePcpMXbfyE:bkzohem) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d5bd432ecaadd926aa60e6f5e6264aa2efa4d40b/test/ci-runtests.sh#L201)):*

I don't prefer "old", but clap doesn't allow short forms for flags that start with the same letter. Alternatively we could rename `--package-folder` to something starting with another letter, but I couldn't come up with anything good.
</blockquote></details>

You can manually specify the letter used for the short form. Or just not let every argument have a short form. I much prefer doing one of those over arbitrarily renaming stuff just to have unique first letters. Always prefer descriptive good names over submitting to trivial technical limitations IMO :P

Otherwise, what happens when you have one flag per letter in the alphabet except one... Will whatever flag that comes next have to find a decent word on that single letter to be allowed to be added? :D

Except for a few well known commonly used abbreviations like `-o --output`, `-i --input` or `-i --ignore-case`, `-r --recursive`, `-f --force` I actually personally think short flags is a bit of an anti-pattern. It makes reading commands shorter, but way harder :upside_down_face: 

`run-tests -o 2024.1 -c 2024.5 -a 123423423` is not very descriptive! I would not have any idea what that did :sweat_smile: 
___
*[`test/test-manager/src/logging.rs` line 33 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24k7A297O8t_9-K8z5:-O24yjmIAz_czkizHv4I:b55u2dz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/542bdf77975eae321f40fedb15dd06ebf222614e/test/test-manager/src/logging.rs#L33)):*

You can still specify the log level using `RUST_LOG`, which is done by `ci-runtests.sh` so its output will be the same. I just found the logs too verbose for my taste when running manually. The debug level logs are not really relevant to me when I am working on new e2e tests, so I consider it a more sane default. In general I think `INFO` should be the default for logs.
</blockquote></details>

Ok! Yeah, makes sense IMO
___
*[`test/test-manager/src/package.rs` line 38 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24UcwJ4OFa51YPSoxz:-O24y8_p8etuOPeTXSf2:bmgv0pc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/package.rs#L38)):*

I opted for the cautious approach and added a warning, just in case someone forgets to set the previous app version now that it's optional. It will eventually give a pretty clear error message anyway, so it may be superfluous. I'll keep it unless you strongly think logging is unimportant.
</blockquote></details>

We can keep it if you think it's important. I just figured this is probably more suitably logged somewhere else earlier/further up? Let's keep it.
___
*[`test/test-manager/src/vm/provision.rs` line 173 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24ZbWq26-_oB6thz_d:-O24yPJtCjW_xEA5GpF_:b-83sjoh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c79b6493abd6240ba63ce24fb3aa022cce74fe7/test/test-manager/src/vm/provision.rs#L173)):*

Sure. I did that and refactored a bit.
</blockquote></details>

Nice!

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475)* status: 7 of 11 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`test/ci-runtests.sh` line 201 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475#-O24kU3N4B6QUUb6ysnp:-O25CR43D-zV1m_dh60d:bou6tt2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d5bd432ecaadd926aa60e6f5e6264aa2efa4d40b/test/ci-runtests.sh#L201)):*

You can manually specify the letter used for the short form. Or just not let every argument have a short form. I much prefer doing one of those over arbitrarily renaming stuff just to have unique first letters. Always prefer descriptive good names over submitting to trivial technical limitations IMO :P

Otherwise, what happens when you have one flag per letter in the alphabet except one... Will whatever flag that comes next have to find a decent word on that single letter to be allowed to be added? :D

Except for a few well known commonly used abbreviations like `-o --output`, `-i --input` or `-i --ignore-case`, `-r --recursive`, `-f --force` I actually personally think short flags is a bit of an anti-pattern. It makes reading commands shorter, but way harder :upside_down_face: 

`run-tests -o 2024.1 -c 2024.5 -a 123423423` is not very descriptive! I would not have any idea what that did :sweat_smile:
</blockquote></details>

I agree with your point about short forms, I don't actually use them very often myself. It's just that my explicit intent with the PR was to make running `test-manager` manually easier, so it seemed in line with that goal to have a short form for the new flag. Maybe a better solution is to remove the short form of `--previous-app` since most likely you want to skip specifying it anyway.

Ideally, I would like to change the name of both `--current-app` and `--previous-app`. I don't think those names make sense anymore. `previous` kind of implies that it had to be the version that's one release older than `current-app`. Since you likely will skip `previous-app` the flags could be changed to `--app-package` and `--app-package-to-upgrade-from` or something, but that causes conflicts with `--account` :upside_down_face:

faern (commented):
> Reviewed 2 of 4 files at r5, 1 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475)* status: 10 of 11 files reviewed, 1 unresolved discussion

faern (approved):
> Reviewed 7 of 8 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6475)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6481, opened 2024-07-18
Implement feature indicators in daemon

hulthe:
- Add a GetFeatureIndicators gRPC call that get's the current set of active "features" that should be shown in the UI.
- Extend the DaemonEvent gRPC stream with a FeatureIndicators variant that has the same info.


mullvad-management-interface/proto/management_interface.proto
```diff
@@ -235,6 +238,24 @@ message TunnelEndpoint {
...
+  LOCKDOWN_MODE = 4;
+  UDP_2_TCP = 5;
+  LAN_SHARING = 6;
+  DNS_CONTENT_BLOCKERS = 7;
+  CUSTOM_DNS = 8;
+  SERVER_IP_OVERRIDE = 9;
+  CUSTOM_MTU = 10;
+  CUSTOM_MSS_FIX = 11;
+  DAITA = 12;
+}
```
Serock3, line 257: Just a thought, but `FeatureIndicators` and `FeatureIndicator` have very similar names. What do you think about replacing `FeatureIndicators` with e.g. `FeatureIndicatorSet` or something similar?

mullvad-types/src/features.rs
```diff
@@ -0,0 +1,31 @@
...
+
+/// Feature indicators are active settings that should be shown to the user to make them aware of
+/// what is affecting their connection at any given time.
+///
+/// Note that the feature indicators are not ordered.
+#[derive(Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct FeatureIndicators {
+    pub active_features: HashSet<FeatureIndicator>,
+}
+
```
Serock3, line 13: Maybe this should just be a tuple struct so we avoid having to name the only field? I.e. FeatureIndicators(HashSet<FeatureIndicator>).

mullvad-management-interface/src/client.rs
```diff
@@ -50,6 +51,7 @@ pub enum DaemonEvent {
     Device(DeviceEvent),
     RemoveDevice(RemoveDeviceEvent),
     NewAccessMethod(AccessMethodSetting),
+    FeatureIndicators(FeatureIndicators),
```
Serock3, line 54: I discussed with @hulthe before his vacation about whether this should be its own entry in `DaemonEvent`, or a part of the `TunnelState`. With this implementation, won't we very frequenty get double `DaemonEvent`s notifying the same thing? what are your thoughts?

Serock3: Also, since the indicators are not really meaningful in the `Disconnected` state, we could simply leave them out of that enum variant, which would reduce some ambiguity.

mullvad-daemon/src/lib.rs
```diff
@@ -1399,6 +1417,11 @@ where
         let _ = tx.send(save_result.map(|_| ()));
     }
 
+    fn handle_settings_changed(&mut self) {
+        self.event_listener
+            .notify_feature_indicators(self.get_feature_indicators())
+    }
```
Serock3, line 1423: This seems a bit verbose to me. Should we really send a feature indicator nitification on every single settings change? Would it not be preferable to only send the event if an actual indicator was affected?

mullvad-management-interface/proto/management_interface.proto
MarkusPettersson98: I don't really mind the close resemblance, but I see how it could be confusing. I like the relative brevity of it though.  I'll think about it :blush:

mullvad-management-interface/src/client.rs
MarkusPettersson98: I like it! `mullvad_types::TunnelState` is a representation of the actual tunnel state, so it should be fine to put non-critical pieces of user-facing information there.

mullvad-daemon/src/lib.rs
MarkusPettersson98: Hmm, I'll see what I can do about it. Should be resolved by itself if `FeatureIndicators` is moved to `TunnelState`.

MarkusPettersson98: Yup, `handle_settings_changed` could be removed :tada:

mullvad-management-interface/src/client.rs
MarkusPettersson98: Dropped `FeatureIndicators` from the `DaemonEvent` enum, think it turned out better!

Serock3 (changes requested):
There seems to be no code for triggering a new `TunnelState` when a setting changes that affects feature indicators, but not the tunnelstate itself, e.g. lockdown mode. You likely still have to listen to changes to the settings to capture this.

> Reviewed 8 of 11 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6481#-:-O36sI0u--8uv1YZpv-Y:b5a1qpc)* status: all files reviewed, 2 unresolved discussions (waiting on @hulthe)

___
*[`mullvad-cli/src/format.rs` line 25 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6481#-O36omDxEYDUeICxM6PM:-O36omDxEYDUeICxM6PN:b-2f4of5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b9ab1f82aea2a4aba6f0ce3320a931bf2a7981c1/mullvad-cli/src/format.rs#L25)):*
> ```XML
>             endpoint,
>             location,
>             feature_indicators: _,
> ```

What do you think about printing the feature indicators as a part of the CLI?
___
*[`mullvad-management-interface/src/client.rs` line 54 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6481#gh-1695057547:-O36pVSM8pyYdzFfpMhG:bmusbp8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a07dc725e9dcc7765cc578460f3049d5b3f22b56/mullvad-management-interface/src/client.rs#L54)):*

Dropped `FeatureIndicators` from the `DaemonEvent` enum, think it turned out better!
</blockquote></details>

Great work! :smile:

Serock3 (changes requested):
> Reviewed 1 of 2 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6481)* status: all files reviewed, 2 unresolved discussions (waiting on @hulthe)

___
*[`mullvad-daemon/src/lib.rs` line 1116 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6481#-O38332gB-edV5dYMR1O:-O38332gB-edV5dYMR1P:bek1olj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/afd058858d5cee4037ca5f05d71e3590ad0fe1f4/mullvad-daemon/src/lib.rs#L1116)):*
> ```XML
> 
>     /// Update the set of feature indicators.
>     fn handle_feature_indicator_event(&self) {
> ```

Perhaps this function could take the new settings, convert them to feature indicators and only notify about a new tunnel state if they are different from `new_indicators`?

MarkusPettersson98 (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6481#-:-O3CJBj_-6cVIP7P-wd7:bb6un1j)* status: 13 of 15 files reviewed, 2 unresolved discussions (waiting on @hulthe and @Serock3)

Serock3 (approved):
> Reviewed 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6481)* status: all files reviewed, 1 unresolved discussion (waiting on @hulthe)

Serock3 (approved):
> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6481)* status: all files reviewed, 1 unresolved discussion (waiting on @hulthe)

PR #6491, opened 2024-07-23
Fix desktop e2e GitHub action false negatives

MarkusPettersson98:
Add a precondition which checks if there are any VMs to test for the given OS. Without this check, the parsing of the VM names assumed that there would exist at least one, causing the parsing to fail if there were none. This would show up in the GitHub summary view as a failure.

Also, add some more descriptive names to other steps

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6491)

### DIFF
--- a/.github/workflows/desktop-e2e.yml
+++ b/.github/workflows/desktop-e2e.yml
@@ -22,6 +22,7 @@ on:
         type: string
 jobs:
   prepare-matrices:
+    name: Prepare virtual machines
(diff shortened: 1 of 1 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6491)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6499, opened 2024-07-24
Add GitHub actions permissions

faern:
Locks down what permissions some of our Github actions has. Limits the risk of them doing something malicious against us. We don't build anything trusted in any of our actions. But with enough permissions set, the actions can write stuff back into the repository or cause other issues. So by default we should try to limit them as much as possible.

I wish Github defaulted to no permissions and explicitly required an action to specify if they needed more permissions than that.

This is in accordance with OpenSSF best practices: https://github.com/ossf/scorecard/blob/b48bdbf250dedadedb42934480bb885d756ead0c/docs/checks.md#token-permissions

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6499)

### DIFF
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -3,9 +3,13 @@ name: Check changelog format
 on:
   pull_request:
     paths:
+      - .github/workflows/check-changelog.yml
       - 'CHANGELOG.md'
       - 'ios/CHANGELOG.md'
       - 'android/CHANGELOG.md'
+
+permissions: {}
+
 env:
   LINE_LIMIT: 100
 jobs:
--- a/.github/workflows/git-commit-message-style.yml
+++ b/.github/workflows/git-commit-message-style.yml
@@ -4,6 +4,8 @@ on:
   push:
   workflow_dispatch:
 
+permissions: {}
+
 jobs:
   check-commit-message-style:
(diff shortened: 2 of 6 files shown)


faern:
Converting back to draft... I think these permissions screwed with some of the action runs. Will have to check more carefully that this PR does not break any CI job... https://github.com/mullvad/mullvadvpn-app/actions/runs/10076755615

dlon (approved):
:lgtm:

We should probably lock down the remaining workflows at some point? It seems like most permissions are unnecessary for nearly all of them.

> Reviewed 1 of 5 files at r2, 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6499#-:-O2dQEAV7lEBkmSZm39K:b-m3t13n)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (commented):
Yes, absolutely! That's the plan. Starting off slow, in case it creates some issues.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6499#-:-O2dRrB_7a6O3AxC-Ey1:bmmwjft)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6547, opened 2024-07-30
Fix crash in request vpn permission

Pururun:
This crash could occur if create intent in request vpn permission was called while vpn permission was already approved

Fixed by returning true immediately in those cases

### DIFF
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/util/RequestVpnPermission.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/util/RequestVpnPermission.kt
@@ -8,13 +8,21 @@ import androidx.activity.result.contract.ActivityResultContract
 
 class RequestVpnPermission : ActivityResultContract<Unit, Boolean>() {
     override fun createIntent(context: Context, input: Unit): Intent {
-        // We expect this permission to only be requested when the permission is missing, however,
-        // if it for some reason is called incorrectly we should return an empty intent so we avoid
-        // a crash.
-        return VpnService.prepare(context) ?: Intent()
+        return VpnService.prepare(context)
     }
 
     override fun parseResult(resultCode: Int, intent: Intent?): Boolean {
         return resultCode == Activity.RESULT_OK
+    }
+
+    // We expect this permission to only be requested when the permission is missing, however,
+    // if it for some reason is called incorrectly we will skip the call to create intent
+    // to avoid crashing. The app will then proceed as the user accepted permission.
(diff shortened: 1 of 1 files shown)


Rawa (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6547)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Pururun)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/util/RequestVpnPermission.kt` line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6547#-O36hZCX9bXvtmAgvWCL:-O36hZCX9bXvtmAgvWCM:b-389u23) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bbf2c8702c583df84df49a1825b0a720bd7ad330/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/util/RequestVpnPermission.kt#L11)):*
> ```Kotlin
> class RequestVpnPermission : ActivityResultContract<Unit, Boolean>() {
>     override fun createIntent(context: Context, input: Unit): Intent {
>         return VpnService.prepare(context)
> ```

This is a bit unclear, return type is `Intent` but `VpnService.prepare(context)` can return null right? We should at least do `!!` to make it clear that we don't expect a null.

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6547)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Rawa)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/util/RequestVpnPermission.kt` line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6547#-O36hZCX9bXvtmAgvWCL:-O36tujT4kDor8_l9H7h:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bbf2c8702c583df84df49a1825b0a720bd7ad330/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/util/RequestVpnPermission.kt#L11)):*

This is a bit unclear, return type is `Intent` but `VpnService.prepare(context)` can return null right? We should at least do `!!` to make it clear that we don't expect a null.
</blockquote></details>

Done.

Rawa (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6547)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6559, opened 2024-08-02
Gracefully shutdown gRPC server

MarkusPettersson98:
This PR refactors the dameon a bit to properly shut down the management interface server. Previously, it would die when the tokio runtime was tore down due to the tokio task it was run on being dropped. Since we already spawn the `tonic` server with [`serve_with_incoming_shutdown`](https://docs.rs/tonic/latest/tonic/transport/server/struct.Router.html#method.serve_with_incoming_shutdown), it was just a matter of actually using the shutdown signal future that we already had before awaiting the management interface server future.

This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6559)

### DIFF
--- a/mullvad-daemon/src/access_method.rs
+++ b/mullvad-daemon/src/access_method.rs
@@ -1,4 +1,4 @@
-use crate::{api, settings, Daemon, EventListener};
+use crate::{api, settings, Daemon};
 use mullvad_api::{proxy::ApiConnectionMode, rest, ApiProxy};
 use mullvad_types::{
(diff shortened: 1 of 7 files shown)


dlon (approved):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6559)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6563, opened 2024-08-05
Fix potential crash in connection state flow

Pururun:
I am not sure this happens in the actual app, but it is good to prevent future issues.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6563)

### DIFF
--- a/android/lib/daemon-grpc/src/main/kotlin/net/mullvad/mullvadvpn/lib/daemon/grpc/util/ManagedChannel.kt
+++ b/android/lib/daemon-grpc/src/main/kotlin/net/mullvad/mullvadvpn/lib/daemon/grpc/util/ManagedChannel.kt
@@ -1,24 +1,28 @@
 package net.mullvad.mullvadvpn.lib.daemon.grpc.util
 
+import co.touchlab.kermit.Logger
 import io.grpc.ConnectivityState
 import io.grpc.ManagedChannel
-import kotlin.coroutines.resume
-import kotlin.coroutines.suspendCoroutine
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.callbackFlow
 import kotlinx.coroutines.isActive
+import kotlinx.coroutines.suspendCancellableCoroutine
 
+@OptIn(ExperimentalCoroutinesApi::class)
 internal fun ManagedChannel.connectivityFlow(): Flow<ConnectivityState> {
     return callbackFlow {
         var currentState = getState(false)
-        send(currentState)
 
         while (isActive) {
-            currentState =
-                suspendCoroutine<ConnectivityState> {
-                    notifyWhenStateChanged(currentState) { it.resume(getState(false)) }
-                }
+            // Check that we are active before sending
             send(currentState)
+            currentState = suspendCancellableCoroutine {
+                notifyWhenStateChanged(currentState) {
+                    // If we are cancelled we will just log
+                    it.resume(getState(false)) { Logger.w("Resume while cancelled", it) }
+                }
+            }
         }
     }
 }


Rawa (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6563)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Pururun)

___
*[`android/lib/daemon-grpc/src/main/kotlin/net/mullvad/mullvadvpn/lib/daemon/grpc/util/ManagedChannel.kt` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6563#-O3YOMHA0uGMMN5PIifJ:-O3YOMHC38KWdlU56qFA:b-8lten5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2649ec774823d15d91415bf73eeafda2ddb6cbc/android/lib/daemon-grpc/src/main/kotlin/net/mullvad/mullvadvpn/lib/daemon/grpc/util/ManagedChannel.kt#L26)):*
> ```Kotlin
>         }
> 
>         awaitClose { cancel() }
> ```

I don't think this should be necessary? `awaitClose` block is called when the flow is cancelled, I believe this block should rather be to used to remove listeners etc.

I believe the problem we faced is that we don't have a way to unregister the `notifyWhenStateChanged` callback. Thus it would invoke `it.resume(...)` and possibly `send(currentState)` on a already cancelled flow.

So I believe the correct solution would be something along the lines of:
```
internal fun ManagedChannel.connectivityFlow(): Flow<ConnectivityState> {
    return callbackFlow {
        var currentState = getState(false)

        while (isActive) {
            // Check that we are active before sending
            send(currentState)
            currentState = suspendCancellableCoroutine {
                notifyWhenStateChanged(currentState) {
                    // If we are cancelled we don't try and invoke the it.resume
                    if (it.isActive) {
                        it.resume(getState(false))
                    }
                }
            }
        }
    }
}
```

Alternatively I believe the callback could possibly be, where lambda is the onCancellation callback:
```
                        it.resume(getState(false), { _ -> })
```

Rawa (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6563)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Pururun)

___
*[`android/lib/daemon-grpc/src/main/kotlin/net/mullvad/mullvadvpn/lib/daemon/grpc/util/ManagedChannel.kt` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6563#-O3YOMHA0uGMMN5PIifJ:-O3YRV7D92x0w2AFYijf:b4h6w42) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a2649ec774823d15d91415bf73eeafda2ddb6cbc/android/lib/daemon-grpc/src/main/kotlin/net/mullvad/mullvadvpn/lib/daemon/grpc/util/ManagedChannel.kt#L26)):*

I don't think this should be necessary? `awaitClose` block is called when the flow is cancelled, I believe this block should rather be to used to remove listeners etc.

I believe the problem we faced is that we don't have a way to unregister the `notifyWhenStateChanged` callback. Thus it would invoke `it.resume(...)` and possibly `send(currentState)` on a already cancelled flow.

So I believe the correct solution would be something along the lines of:
```
internal fun ManagedChannel.connectivityFlow(): Flow<ConnectivityState> {
    return callbackFlow {
        var currentState = getState(false)

        while (isActive) {
            // Check that we are active before sending
            send(currentState)
            currentState = suspendCancellableCoroutine {
                notifyWhenStateChanged(currentState) {
                    // If we are cancelled we don't try and invoke the it.resume
                    if (it.isActive) {
                        it.resume(getState(false))
                    }
                }
            }
        }
    }
}
```

Alternatively I believe the callback could possibly be, where lambda is the onCancellation callback:
```
                        it.resume(getState(false), { _ -> })
```
</blockquote></details>

Clarification: With second alternative I don't believe the `if (it.isActive)` is needed.

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6563)* status: all files reviewed, 1 unresolved discussion (waiting on @Pururun)

Rawa (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6563)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

albin-mullvad (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6563#-:-O45MVaX6f3dydD6RmzI:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6571, opened 2024-08-06
Add better Voucher Code validation

Rawa:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571)

### DIFF
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/dialog/RedeemVoucherDialog.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/dialog/RedeemVoucherDialog.kt
@@ -274,10 +274,23 @@ private fun EnterVoucherBody(
             )
         }
     }
+    if (
+        state.voucherState is VoucherDialogState.Error &&
+            state.voucherState.error is RedeemVoucherError.EnteredAccountNumber
+    ) {
+        Text(
+            modifier = Modifier.padding(top = Dimens.smallPadding),
+            text = stringResource(id = R.string.voucher_is_account_number),
+            color = MaterialTheme.colorScheme.onPrimary,
+            style = MaterialTheme.typography.bodySmall
+        )
+    }
 }
 
 private fun RedeemVoucherError.message(): Int =
     when (this) {
+        RedeemVoucherError.TooShortVoucher,
+        RedeemVoucherError.EnteredAccountNumber,
         RedeemVoucherError.InvalidVoucher -> R.string.invalid_voucher
         RedeemVoucherError.VoucherAlreadyUsed -> R.string.voucher_already_used
         RedeemVoucherError.RpcError,
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/VoucherDialogViewModel.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/VoucherDialogViewModel.kt
@@ -2,6 +2,7 @@ package net.mullvad.mullvadvpn.viewmodel
 
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
+import arrow.core.raise.either
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.combine
@@ -11,7 +12,9 @@ import kotlinx.coroutines.launch
 import net.mullvad.mullvadvpn.compose.state.VoucherDialogState
 import net.mullvad.mullvadvpn.compose.state.VoucherDialogUiState
 import net.mullvad.mullvadvpn.constant.VOUCHER_LENGTH
+import net.mullvad.mullvadvpn.lib.model.ParseVoucherCodeError
 import net.mullvad.mullvadvpn.lib.model.RedeemVoucherError
+import net.mullvad.mullvadvpn.lib.model.VoucherCode
 import net.mullvad.mullvadvpn.lib.shared.VoucherRepository
 import net.mullvad.mullvadvpn.util.VoucherRegexHelper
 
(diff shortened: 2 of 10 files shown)


Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571)* status: 0 of 9 files reviewed, 2 unresolved discussions

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/VoucherDialogViewModel.kt` line 36 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571#-O3abJX49n3DrjbbCusQ:-O3abJX49n3DrjbbCusR:bzhd80b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b87e015ff4c13bbf018066340e2ae628eab25dc7/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/VoucherDialogViewModel.kt#L36)):*
> ```Kotlin
>         viewModelScope.launch {
>             either {
>                     val voucherCode =
> ```

There is two ways about this, right now we try and parse it as VoucherCode and if it returns AllDigit (which is not valid according to spec) we show the AccountNumber error message. Another way about it would be to try and parse it as an AccountNumber and if successful print the error message. What do you guys think?
___
*[`android/lib/resource/src/main/res/values/strings.xml` line 36 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571#-O3abikTDNs-iVQakTlR:-O3abikTDNs-iVQakTlS:bipg7k5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b87e015ff4c13bbf018066340e2ae628eab25dc7/android/lib/resource/src/main/res/values/strings.xml#L36)):*
> ```XML
>     <string name="invalid_voucher">Voucher code is invalid.</string>
>     <string name="voucher_already_used">Voucher code has already been used.</string>
>     <string name="voucher_is_account_number">Looks like you entered an account number</string>
> ```

Any ideas of what we want the error message to be?

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571)* status: 0 of 9 files reviewed, 3 unresolved discussions

___
*[`android/lib/model/src/main/kotlin/net/mullvad/mullvadvpn/lib/model/VoucherCode.kt` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571#-O3ac_eyCrTC_V_6S_Ra:-O3ac_eyCrTC_V_6S_Rb:b-h28vdm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b87e015ff4c13bbf018066340e2ae628eab25dc7/android/lib/model/src/main/kotlin/net/mullvad/mullvadvpn/lib/model/VoucherCode.kt#L12)):*
> ```Kotlin
>     companion object {
>         // Parsing reference: docs/adr/0018-distinguish-voucher-codes-from-account-numbers.md
>         fun fromString(value: String): Either<ParseVoucherCodeError, VoucherCode> = either {
> ```

We could make even stricter validation, e.g only certain characters are allowed or max length, but I opted against it if the schema is subject to change in the future. So what is implemented is the lower bounds. (min length, not only digits)

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571)* status: all files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @Pururun)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/VoucherDialogViewModel.kt` line 36 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571#-O3abJX49n3DrjbbCusQ:-O3au0bV8RcEUgagWTca:bf99z0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b87e015ff4c13bbf018066340e2ae628eab25dc7/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/VoucherDialogViewModel.kt#L36)):*

In which cases would we fail to parse it as an account number but it would be only digits and over 16 characters? If it is too many? It feels fine to me as it is.
</blockquote></details>

Well yeah it is a bit about interpretation of how when we logically should show the error.
A. When we are successfully parsing it as an account number
B. When we notice that the Voucher Code is consisting of only digits and is long enough.

Right now we have the B approach, we could technically miss cases were they enter a valid account number that is shorter than 16 characters and then would say it is too short rather than it is an account number. By doing the A approach we know it fulfills all our requirements for an AccountNumber (could be good and bad).

I agree, I think our current solution is fine for now, if it becomes a problem we can refine it later.
___
*[`android/lib/resource/src/main/res/values/strings.xml` line 36 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571#-O3abikTDNs-iVQakTlR:-O3aurbM3ifgATHbgoSx:b-y4nqut) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b87e015ff4c13bbf018066340e2ae628eab25dc7/android/lib/resource/src/main/res/values/strings.xml#L36)):*

I guess it should be based on what the normal use case is. Maybe something like "It looks like you entered an account number. If that is the case, log out and log in again using your account number" ?
</blockquote></details>

Yeah it is hard to tell. Also "Looks like you entered an account number" doesn't sound like an error message. Maybe it should say, "Voucher code is invalid, looks like you enter an account number, if you wish to change account please logout first", but it would be quite a long error message :see_no_evil: . Maybe @albin-mullvad has some input?

Pururun (approved):
:lgtm:

> Reviewed 1 of 1 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571#-:-O3rEbKqBh-jdtDQefs1:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

albin-mullvad (approved):
:lgtm: 

> Reviewed 2 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571#-:-O4o1eE9DxYFRZXo0CXa:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`android/lib/model/src/main/kotlin/net/mullvad/mullvadvpn/lib/model/VoucherCode.kt` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571#-O3ac_eyCrTC_V_6S_Ra:-O4o1U8d3XjK0UlLvJka:b-ux0ur8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b87e015ff4c13bbf018066340e2ae628eab25dc7/android/lib/model/src/main/kotlin/net/mullvad/mullvadvpn/lib/model/VoucherCode.kt#L12)):*

We could make even stricter validation, e.g only certain characters are allowed or max length, but I opted against it if the schema is subject to change in the future. So what is implemented is the lower bounds. (min length, not only digits)
</blockquote></details>

Sounds good :thumbsup:

albin-mullvad (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

albin-mullvad (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6571)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6603, opened 2024-08-12
Fix view logs crash on TV

Pururun:
This seems to be caused by the app going through all the different items in the lazy list and then checking them one by one if they can be focused.

This is a workaround for this, I guess we should check if this is fixed by future updates to compose.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6603)


albin-mullvad (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6603#-:-O45s7q09MchSSb41DQr:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6643, opened 2024-08-19
Fix feature indicators missing during connecting state

Serock3:
Feature indicators were missing during the connecting state. This was
because they were calculated using the endpoint of the previous tunnel
state.

This commit fixes the bug and restructures the access to feature
indicators to be more readable.

Fixes DES-1140

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6643)

### DIFF
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -53,13 +53,13 @@ use mullvad_types::{
     auth_failed::AuthFailed,
     custom_list::CustomList,
     device::{Device, DeviceEvent, DeviceEventCause, DeviceId, DeviceState, RemoveDeviceEvent},
-    features::{FeatureIndicator, FeatureIndicators},
+    features::{compute_feature_indicators, FeatureIndicators},
     location::{GeoIpLocation, LocationEventData},
     relay_constraints::{
         BridgeSettings, BridgeState, BridgeType, ObfuscationSettings, RelayOverride, RelaySettings,
     },
     relay_list::RelayList,
-    settings::{DnsOptions, DnsState, Settings},
+    settings::{DnsOptions, Settings},
     states::{Secured, TargetState, TargetStateStrict, TunnelState},
     version::{AppVersion, AppVersionInfo},
     wireguard::{PublicKey, QuantumResistantState, RotationInterval},
@@ -87,7 +87,7 @@ use talpid_types::android::AndroidContext;
 #[cfg(target_os = "windows")]
 use talpid_types::split_tunnel::ExcludedProcess;
 use talpid_types::{
-    net::{IpVersion, ObfuscationType, TunnelType},
+    net::{IpVersion, TunnelType},
     tunnel::{ErrorStateCause, TunnelStateTransition},
     ErrorExt,
 };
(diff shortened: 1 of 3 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6643)* status: all files reviewed, 2 unresolved discussions (waiting on @Serock3)

___
*[`mullvad-daemon/src/lib.rs` line 1136 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6643#-O4ewCznDu629onr9eOg:-O4ewCznDu629onr9eOh:b-xkyk5e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ecb7036a1497b63cfcbf86d93a654cad943ecc7/mullvad-daemon/src/lib.rs#L1136)):*
> ```XML
>         // state (e.g. activating lockdown mode). Note that only the connected and connecting states
>         // have feature indicators.
>         match &mut self.tunnel_state {
> ```

What do you think of adding a getter in `TunnelState` that returns a mutable optional reference to `feature_indicators`?
___
*[`mullvad-daemon/src/lib.rs` line 2837 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6643#-O4evxVFEFRCrDG0wsR3:-O4evxVFEFRCrDG0wsR4:bwuzmi1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ecb7036a1497b63cfcbf86d93a654cad943ecc7/mullvad-daemon/src/lib.rs#L2837)):*
> ```XML
> 
>     fn on_get_feature_indicators(&self, tx: oneshot::Sender<FeatureIndicators>) {
>         let feature_indicators = match &self.tunnel_state {
> ```

What do you think of keeping the getter (`TunnelState::get_feature_indicators`) and just calling `unwrap_or_default()` on that?

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6643)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-daemon/src/lib.rs` line 1136 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6643#-O4ewCznDu629onr9eOg:-O4iTyqY3X7gxtb1zfu5:b-gre2tw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ecb7036a1497b63cfcbf86d93a654cad943ecc7/mullvad-daemon/src/lib.rs#L1136)):*

What do you think of adding a getter in `TunnelState` that returns a mutable optional reference to `feature_indicators`?
</blockquote></details>

I am not strongly against the idea of keeping the getter. In this case I also needed the endpoint so it was simpler with just a match statement, so I prefer to keep it this way here.
___
*[`mullvad-daemon/src/lib.rs` line 2837 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6643#-O4evxVFEFRCrDG0wsR3:-O4iV_Xp64U7AANwVIrb:bxd1iut) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0ecb7036a1497b63cfcbf86d93a654cad943ecc7/mullvad-daemon/src/lib.rs#L2837)):*

What do you think of keeping the getter (`TunnelState::get_feature_indicators`) and just calling `unwrap_or_default()` on that?
</blockquote></details>

Here, keeping the getter would be a little simpler, or at least it would hide away the logic. It would be the only place where it is used though, so I am stil not convinced. Also, the fact that feature indicators only exist in the connected and connecting states is more readable this way.

PR #6654, opened 2024-08-21
Fix IP-override feature indicator

Serock3:
It was trigger by any overrides existing in the settings, not by the current endpoint being overridden.

Add flag to `Relay` to specify if its IPv4 and/or IPv6 has been overridden and use that in combination with the endpoint IP version to derive if the current connection is overridden.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6654)

### DIFF
--- a/mullvad-api/src/relay_list.rs
+++ b/mullvad-api/src/relay_list.rs
@@ -154,6 +154,8 @@ fn into_mullvad_relay(
         hostname: relay.hostname,
         ipv4_addr_in: relay.ipv4_addr_in,
         ipv6_addr_in: relay.ipv6_addr_in,
+        overridden_ipv4: false,
+        overridden_ipv6: false,
         include_in_country: relay.include_in_country,
         active: relay.active,
         owned: relay.owned,
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -53,7 +53,7 @@ use mullvad_types::{
     auth_failed::AuthFailed,
     custom_list::CustomList,
     device::{Device, DeviceEvent, DeviceEventCause, DeviceId, DeviceState, RemoveDeviceEvent},
-    features::{compute_feature_indicators, FeatureIndicators},
+    features::{compute_feature_indicators, FeatureIndicator, FeatureIndicators},
     location::{GeoIpLocation, LocationEventData},
     relay_constraints::{
         BridgeSettings, BridgeState, BridgeType, ObfuscationSettings, RelayOverride, RelaySettings,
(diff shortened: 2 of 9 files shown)


Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6654)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[`mullvad-relay-selector/tests/relay_selector.rs` line 851 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6654#-O4tCnu9AnAM2VVvoU2N:-O4u35LF6XlWfiTLosX9:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/74f2df2efb0eeb63ab2f93d7bb91ec9ebd1b1bb4/mullvad-relay-selector/tests/relay_selector.rs#L851)):*

An assertion is missing from the `_v6` test.
</blockquote></details>

Done.
___
*[`mullvad-relay-selector/tests/relay_selector.rs` line 851 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6654#-O4tAU7o9USAJmt0LrLY:-O4u35UmA-rpQ51dAxOx:b2eu61y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/74f2df2efb0eeb63ab2f93d7bb91ec9ebd1b1bb4/mullvad-relay-selector/tests/relay_selector.rs#L851)):*

Could we also assert that it is *false* when not using overrides? Also, we could assert that `!overridden_ipv6`.
</blockquote></details>

Alright. I added the assertion of no overrides for either v4 or v6 to `test_selecting_wireguard_over_shadowsocks_extra_ips`

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6654)* status: all files reviewed, 1 unresolved discussion (waiting on @Serock3)

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6654)* status: 8 of 9 files reviewed, all discussions resolved (waiting on @dlon)

___
*[`mullvad-daemon/src/lib.rs` line 1153 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6654#-O4t2q1wAG1je2sz1hyw:-O4u_3VV2UrWd8SCjjDd:b-4q8ddl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/74f2df2efb0eeb63ab2f93d7bb91ec9ebd1b1bb4/mullvad-daemon/src/lib.rs#L1153)):*

It would probably be nicer to only rely on `parameters_generator` during transitions, to make fewer assumptions about when it's updated, right? Maybe `ip_override` could be derived from the tunnel state  (eg by adding `Relay` to it)?

When constructing the `TunnelState`s, we could save the last relay and obtain the override from `tunnel_state.ip_overridden()  -> bool`.
</blockquote></details>

I re-used the previous value for the indicator as you suggested. Take a look at the comment and see if you think it makes sense.

dlon (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6654#-:-O4vKDMn3NaC4yXCsqvy:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6660, opened 2024-08-22
Add top level `permissions` to all Github Actions workflows

faern:
The default permission on the repository is already set to read only. So in practice this makes no difference. But this makes that more explicit, and less relying on the repository being correctly configured.

This also makes security scanning tools such as OpenSSF scorecard happier about the overall security of our repository.

I think we should start *always* defining `permissions` on our actions. However, we currently don't have anything that enforces it, and given how little actual security it gives us right now it's definitely not worth automating. I just wanted to do this little manual sweep to "improve" the current state.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660)

### DIFF
--- a/.github/workflows/android-app.yml
+++ b/.github/workflows/android-app.yml
@@ -49,6 +49,9 @@ on:
   # Build if main is updated to ensure up-to-date caches are available
   push:
     branches: [main]
+
+permissions: {}
+
 jobs:
   prepare:
     name: Prepare
--- a/.github/workflows/android-audit.yml
+++ b/.github/workflows/android-audit.yml
@@ -21,6 +21,9 @@ on:
         description: Override container image
         type: string
         required: false
+
+permissions: {}
+
 jobs:
   prepare:
     name: Prepare
--- a/.github/workflows/android-kotlin-format-check.yml
+++ b/.github/workflows/android-kotlin-format-check.yml
@@ -9,6 +9,9 @@ on:
         description: Override container image
         type: string
         required: false
+
+permissions: {}
+
 jobs:
   prepare:
     name: Prepare
--- a/.github/workflows/android-static-analysis.yml
+++ b/.github/workflows/android-static-analysis.yml
@@ -13,6 +13,9 @@ on:
     # Github Actions enabled, so these don't go unnoticed.
     # https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs
     - cron: '20 6 * * *'
+
+permissions: {}
+
 jobs:
   mobsfscan:
     name: Code scanning using mobsfscan
--- a/.github/workflows/android-xml-format-check.yml
+++ b/.github/workflows/android-xml-format-check.yml
@@ -6,6 +6,9 @@ on:
       - .github/workflows/android-xml-format-check.yml
       - android/**/*.xml
   workflow_dispatch:
+
+permissions: {}
+
 jobs:
   prepare:
     name: Prepare
--- a/.github/workflows/cargo-audit.yml
+++ b/.github/workflows/cargo-audit.yml
@@ -10,11 +10,13 @@ on:
     # At 06:20 UTC every day. Will create an issue if a CVE is found.
     - cron: '20 6 * * *'
   workflow_dispatch:
+
+permissions:
+  issues: write
+
 jobs:
   audit:
     runs-on: ubuntu-latest
-    permissions:
-      issues: write
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
--- a/.github/workflows/cargo-vendor.yml
+++ b/.github/workflows/cargo-vendor.yml
@@ -11,6 +11,9 @@ on:
       - Cargo.lock
       - '**/Cargo.toml'
   workflow_dispatch:
+
+permissions: {}
+
 jobs:
   cargo-vendor:
     runs-on: ubuntu-latest
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -7,6 +7,9 @@ on:
       - clippy.toml
       - '**/*.rs'
   workflow_dispatch:
+
+permissions: {}
+
 jobs:
   prepare-android:
     name: Prepare Android container
--- a/.github/workflows/daemon.yml
+++ b/.github/workflows/daemon.yml
@@ -33,6 +33,9 @@ on:
         description: Override container image
         type: string
         required: false
+
+permissions: {}
+
 jobs:
   prepare-linux:
     runs-on: ubuntu-latest
--- a/.github/workflows/desktop-e2e.yml
+++ b/.github/workflows/desktop-e2e.yml
@@ -25,6 +25,9 @@ on:
         default: ''
         required: false
         type: string
+
+permissions: {}
+
 jobs:
   prepare-matrices:
     name: Prepare virtual machines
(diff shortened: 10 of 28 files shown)


albin-mullvad (changes requested):
> Reviewed 26 of 28 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660)* status: 26 of 28 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`.github/workflows/ios-end-to-end-tests-merge-to-main.yml` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660#-O4t3qsHAiSvwbz8BWut:-O4t3qsHAiSvwbz8BWuu:bqv477t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02e67de540a99cd95687dcebe3358a020b7cf7aa/.github/workflows/ios-end-to-end-tests-merge-to-main.yml#L14)):*
> ```YAML
>       - ios/**
> 
> permissions:
> ```

I'm not sure it's a good idea to move these permissions outside the job, since that would make any (potential future) job in this workflow gain those permissions. I believe it's better to put `permissions: {}` on the top level and then let each job specific exactly what they need.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660)* status: 26 of 28 files reviewed, 1 unresolved discussion (waiting on @albin-mullvad)

___
*[`.github/workflows/ios-end-to-end-tests-merge-to-main.yml` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660#-O4t3qsHAiSvwbz8BWut:-O4t80ow2bhV6gcpsRp8:b-920g52) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02e67de540a99cd95687dcebe3358a020b7cf7aa/.github/workflows/ios-end-to-end-tests-merge-to-main.yml#L14)):*

I'm not sure it's a good idea to move these permissions outside the job, since that would make any (potential future) job in this workflow gain those permissions. I believe it's better to put `permissions: {}` on the top level and then let each job specific exactly what they need.
</blockquote></details>

In general I agree. I was just being pragmatic since this workflow has a single job. The same pattern exists in two other files in this PR. For example `cargo-audit` is very unlikely to ever get more jobs. And if it does, creating permissions is pretty core to a CVE checking job. So do you think the same logic holds there?

If so, is there *ever* a reason to specify non-empty top level permissions? I don't see the point in always specifying empty top level permissions and then re-define them with more permissions for workflows that very clearly do one thing and have one job.

albin-mullvad (approved):
:lgtm:

> Reviewed 2 of 28 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660#-:-O4tqZ2BC2CfUh45Mu1B:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`.github/workflows/ios-end-to-end-tests-merge-to-main.yml` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660#-O4t3qsHAiSvwbz8BWut:-O4tEjQaBadLN12fJDbv:bybnl3s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02e67de540a99cd95687dcebe3358a020b7cf7aa/.github/workflows/ios-end-to-end-tests-merge-to-main.yml#L14)):*

In general I agree. I was just being pragmatic since this workflow has a single job. The same pattern exists in two other files in this PR. For example `cargo-audit` is very unlikely to ever get more jobs. And if it does, creating permissions is pretty core to a CVE checking job. So do you think the same logic holds there?

If so, is there *ever* a reason to specify non-empty top level permissions? I don't see the point in always specifying empty top level permissions and then re-define them with more permissions for workflows that very clearly do one thing and have one job.
</blockquote></details>

Alright, good point,  as long as we are aware of the impact of this PR. I'm fine with keeping this change as-is, but I think we need to be careful doing the same in bigger workflows.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660)* status: all files reviewed, 1 unresolved discussion (waiting on @albin-mullvad and @faern)

___
*[`.github/workflows/ios-end-to-end-tests-merge-to-main.yml` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660#-O4t3qsHAiSvwbz8BWut:-O4tSJuZAf1EVo3tPR9R:b1mnmyb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02e67de540a99cd95687dcebe3358a020b7cf7aa/.github/workflows/ios-end-to-end-tests-merge-to-main.yml#L14)):*

Alright, good point,  as long as we are aware of the impact of this PR. I'm fine with keeping this change as-is, but I think we need to be careful doing the same in bigger workflows.
</blockquote></details>

I had a small chat with @albin-mullvad on this, I also agree with his conclusion.
AFAICT, as long as the [default permissions](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) for the `GITHUB_TOKEN` are set to restrictive, most jobs should have no permissions. (Please take this with a grain of salt, this is my understanding of the documentation, but I'm not 100% sure whether I got it right)

As long as we don't define any new jobs in this `yml` file, the change should have no impact.

With that being said, I don't have any strong opinions on the changes in this PR, so I'm fine with them.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660)* status: all files reviewed, 1 unresolved discussion (waiting on @albin-mullvad and @buggmagnet)

___
*[`.github/workflows/ios-end-to-end-tests-merge-to-main.yml` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660#-O4t3qsHAiSvwbz8BWut:-O4u-W9G-KKxQOZppjyZ:b-rf6kdz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/02e67de540a99cd95687dcebe3358a020b7cf7aa/.github/workflows/ios-end-to-end-tests-merge-to-main.yml#L14)):*

I had a small chat with @albin-mullvad on this, I also agree with his conclusion.
AFAICT, as long as the [default permissions](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) for the `GITHUB_TOKEN` are set to restrictive, most jobs should have no permissions. (Please take this with a grain of salt, this is my understanding of the documentation, but I'm not 100% sure whether I got it right)

As long as we don't define any new jobs in this `yml` file, the change should have no impact.

With that being said, I don't have any strong opinions on the changes in this PR, so I'm fine with them.
</blockquote></details>

100% agree that we should not just apply a bunch of global permissions to larger/more complex workflows. For workflows with more than one job, where only one job need permissions, we should only give that single job the needed permissions if practical.

albin-mullvad (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

buggmagnet (approved):
:lgtm:

> Reviewed 27 of 28 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6660#-:-O4uQtrg0YBYtwep5zNe:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6668, opened 2024-08-23
Match no OpenVPN relays when core privacy features are enabled

dlon:
WireGuard will always be preferred, when any of these features is enabled, even when the tunnel protocol is set to auto.

Core privacy features include PQ, multihop, and DAITA.

Close DES-1066.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6668)

### DIFF
--- a/mullvad-relay-selector/src/relay_selector/matcher.rs
+++ b/mullvad-relay-selector/src/relay_selector/matcher.rs
@@ -26,22 +26,22 @@ pub fn filter_matching_relay_list(
 ) -> Vec<Relay> {
     let relays = relay_list.relays();
 
-    let locations = ResolvedLocationConstraint::from_constraint(&query.location, custom_lists);
+    let locations = ResolvedLocationConstraint::from_constraint(query.location(), custom_lists);
     let shortlist = relays
             // Filter on tunnel type
-            .filter(|relay| filter_tunnel_type(&query.tunnel_protocol, relay))
+            .filter(|relay| filter_tunnel_type(&query.tunnel_protocol(), relay))
             // Filter on active relays
             .filter(|relay| filter_on_active(relay))
             // Filter by location
             .filter(|relay| filter_on_location(&locations, relay))
             // Filter by ownership
-            .filter(|relay| filter_on_ownership(&query.ownership, relay))
+            .filter(|relay| filter_on_ownership(&query.ownership(), relay))
             // Filter by providers
-            .filter(|relay| filter_on_providers(&query.providers, relay))
+            .filter(|relay| filter_on_providers(query.providers(), relay))
             // Filter by DAITA support
-            .filter(|relay| filter_on_daita(&query.wireguard_constraints.daita, relay))
+            .filter(|relay| filter_on_daita(&query.wireguard_constraints().daita, relay))
             // Filter by obfuscation support
-            .filter(|relay| filter_on_obfuscation(&query.wireguard_constraints, relay_list, relay));
+            .filter(|relay| filter_on_obfuscation(query.wireguard_constraints(), relay_list, relay));
 
     // The last filtering to be done is on the `include_in_country` attribute found on each
     // relay. When the location constraint is based on country, a relay which has
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ Line wrap the file at 100 chars.                                              Th
 #### Windows
 - Add experimental support for Windows ARM64.
 
+### Changed
+- Never use OpenVPN as a fallback protocol when any of the following features is enabled:
+  multihop, quantum-resistant tunnels, or DAITA.
+
 ### Fixed
 - macOS and Linux: Fix potential crash when disconnecting with DAITA enabled.
 
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -2937,6 +2937,7 @@ fn new_selector_config(settings: &Settings) -> SelectorConfig {
             daita: settings.tunnel_options.wireguard.daita.enabled,
             #[cfg(not(daita))]
             daita: false,
+            quantum_resistant: settings.tunnel_options.wireguard.quantum_resistant,
         },
     };
 
--- a/mullvad-relay-selector/src/error.rs
+++ b/mullvad-relay-selector/src/error.rs
@@ -13,6 +13,9 @@ pub enum Error {
     #[error("Failed to write relay cache file to disk")]
     WriteRelayCache(#[source] std::io::Error),
 
+    #[error("The combination of relay constraints is invalid")]
+    InvalidConstraints,
+
     #[error("No relays matching current constraints")]
     NoRelay,
 
--- a/mullvad-relay-selector/src/relay_selector/mod.rs
+++ b/mullvad-relay-selector/src/relay_selector/mod.rs
@@ -28,6 +28,7 @@ use mullvad_types::{
     },
     relay_list::{Relay, RelayEndpointData, RelayList},
     settings::Settings,
+    wireguard::QuantumResistantState,
     CustomTunnelEndpoint, Intersection,
 };
 use talpid_types::{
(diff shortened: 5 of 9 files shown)


hulthe (changes requested):
> Reviewed 5 of 6 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6668)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`CHANGELOG.md` line 31 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6668#-O5HrZw7Dk2xNPw8_Nh2:-O5HrZw7Dk2xNPw8_Nh3:b3fqk6j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1e440e17d997724dbbee255b0648775e589f370e/CHANGELOG.md#L31)):*
> ```Markdown
> ### Changed
> - OpenVPN is never used as a fallback protocol when any of the following features is enabled:
>   multihop, quantum-resistant tunnels, or DAITA.
> ```

should have an imperative form

Suggestion:
```
Disable use of OpenVPN as a fallback protocol when any of the following features is enabled:
```
___
*[`mullvad-relay-selector/src/relay_selector/matcher.rs` line 168 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6668#-O5HuUMI4owrq1epQ50n:-O5HuUMI4owrq1epQ50o:bc2f8sv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1e440e17d997724dbbee255b0648775e589f370e/mullvad-relay-selector/src/relay_selector/matcher.rs#L168)):*
> ```XML
> 
> /// Returns whether `relay` satisfy the daita constraint posed by `filter`.
> pub fn filter_on_daita(filter: &Constraint<bool>, relay: &Relay) -> bool {
> ```

BTW I'd argue it's more type safe to pass the `&WireguardRelayQuery` instead of the bool constraint since it's such a non-specific type. Feel free to ignore if you disagree though

mullvad-relay-selector/src/relay_selector/matcher.rs
```diff
@@ -38,10 +39,10 @@ pub fn filter_matching_relay_list(
             .filter(|relay| filter_on_ownership(&query.ownership, relay))
             // Filter by providers
             .filter(|relay| filter_on_providers(&query.providers, relay))
-            // Filter by DAITA support
-            .filter(|relay| filter_on_daita(&query.wireguard_constraints.daita, relay))
             // Filter by obfuscation support
-            .filter(|relay| filter_on_obfuscation(&query.wireguard_constraints, relay_list, relay));
+            .filter(|relay| filter_on_obfuscation(&query.wireguard_constraints, relay_list, relay))
+            // Filter on core privacy features
+            .filter(|relay| filter_on_core_privacy_features(&query.wireguard_constraints, relay));
```
Serock3, line 45: I am a little bit concerned about this design deviating from what is the intent of the relay query, since there's much more logic in the filtering step than usual. I think that ideally you should be able to see that a query with only openvpn and a query with e.g. PQ are incompatible when they are intersected, without having to parse the relay list. It may be hard to make e.g. queries with "only openVPN" and "only Wireguard" not representable, but perhaps we could make them impossible to construct? What do you think about modifying the relay query builder and the `impl<'a> From<NormalSelectorConfig<'a>> for RelayQuery ` block so that `tunnel_protocol` is set to `Only(WireGuard)` whenever a privacy feature is enabled, and possibly removing `filter_on_core_privacy_features` entirely?

Serock3, line 45: I am a little bit concerned about this design deviating from what is the intent of the relay query, since there's much more logic in the filtering step than usual. I think that ideally you should be able to see that a query with only openvpn and a query with e.g. PQ are incompatible when they are intersected, without having to parse the relay list. It may be hard to make e.g. queries with "only openVPN" and "only Wireguard" not representable, but perhaps we could make them impossible to construct? What do you think about modifying the relay query builder and the impl<'a> From<NormalSelectorConfig<'a>> for RelayQuery block so that tunnel_protocol is set to Only(WireGuard) whenever a privacy feature is enabled, and possibly removing filter_on_core_privacy_features entirely?

mullvad-relay-selector/src/relay_selector/matcher.rs
```diff
@@ -153,6 +141,44 @@ fn filter_on_obfuscation(
...
+        (
+            Constraint::Only(true),
+            RelayEndpointData::Wireguard(WireguardRelayEndpointData { daita, .. }),
+        ) => *daita,
+        // Filter out all non-WireGuard relays
+        (Constraint::Only(true), _) => false,
+        // Keep all relays when DAITA isn't requested.
+        _ => true,
+    }
+}
```
Serock3, line 180: You can ignore this if you decide to do the above, but if you keep these, maybe it's more readable to use the same matching method for all of them?
```suggestion
// When core privacy features are enabled, only a subset of WireGuard relays will be included
fn filter_on_core_privacy_features(query: &WireguardRelayQuery, relay: &Relay) -> bool {
    filter_on_daita(&query.daita, relay)
        && filter_on_multihop(query.multihop(), relay)
        && filter_on_quantum_resistance(query.quantum_resistant, relay)
}

/// Filter out relays that do not support WireGuard multihop.
fn filter_on_multihop(enabled: bool, relay: &Relay) -> bool {
    // When multihop is enabled, discard all non-wireguard relays
    if enabled {
        matches!(relay.endpoint_data, RelayEndpointData::Wireguard(_))
    } else {
        true
    }
}

/// Filter out relays that do not support quantum resistance.
fn filter_on_quantum_resistance(state: QuantumResistantState, relay: &Relay) -> bool {
    // When PQ is enabled, discard all non-wireguard relays
    if state == QuantumResistantState::On {
        matches!(relay.endpoint_data, RelayEndpointData::Wireguard(_))
    } else {
        true
    }
}

/// Returns whether `relay` satisfy the daita constraint posed by `filter`.
pub fn filter_on_daita(filter: &Constraint<bool>, relay: &Relay) -> bool {
    if matches!(filter, Constraint::Only(true)) {
        matches!(
            &relay.endpoint_data,
            RelayEndpointData::Wireguard(WireguardRelayEndpointData { daita: true, .. })
        )
    } else {
        true
    }
}
```

dlon: Fixed by making it impossible to construct invalid queries (as discussed).

mullvad-relay-selector/src/relay_selector/matcher.rs
dlon: Fixed by making it impossible to construct invalid queries (as discussed).

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6668)* status: 2 of 8 files reviewed, 3 unresolved discussions (waiting on @hulthe and @Serock3)

___
*[`CHANGELOG.md` line 31 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6668#-O5HrZw7Dk2xNPw8_Nh2:-O5JQogV67_N6Wv-cMDm:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1e440e17d997724dbbee255b0648775e589f370e/CHANGELOG.md#L31)):*

should have an imperative form

Suggestion:
```
Disable use of OpenVPN as a fallback protocol when any of the following features is enabled:
```
</blockquote></details>

Done.

Serock3 (approved):

hulthe (commented):
> Reviewed 3 of 6 files at r3, 1 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6668)* status: 6 of 9 files reviewed, 2 unresolved discussions (waiting on @Serock3)

Serock3 (approved):
> Reviewed 1 of 6 files at r1, 1 of 1 files at r2, 2 of 6 files at r3, 2 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6668)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6682, opened 2024-08-26
Fix test_split_tunnel

dlon:
`ping` uses raw sockets rather than `SOCK_DGRAM` by default. This only works as root.

Related test run: https://github.com/mullvad/mullvadvpn-app/actions/runs/10593508501

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6682)

### DIFF
--- a/test/connection-checker/src/net.rs
+++ b/test/connection-checker/src/net.rs
@@ -22,7 +22,7 @@ pub fn send_tcp(opt: &Opt, destination: SocketAddr) -> eyre::Result<()> {
     sock.bind(&socket2::SockAddr::from(bind_addr))
         .wrap_err(eyre!("Failed to bind TCP socket to {bind_addr}"))?;
 
-    let timeout = Duration::from_millis(opt.leak_timeout);
+    let timeout = Duration::from_secs(opt.leak_timeout);
     sock.set_write_timeout(Some(timeout))?;
     sock.set_read_timeout(Some(timeout))?;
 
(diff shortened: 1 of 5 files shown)


hulthe (dismissed):

test/scripts/ssh-setup.sh
```diff
@@ -128,6 +128,9 @@ function install_packages_apt {
     echo "Installing required apt packages"
     apt update
     apt install -yf xvfb wireguard-tools curl
+    if ! which ping &>/dev/null; then
+        apt install -yf iputils-ping || true
```
hulthe, line 132: Why not fail if ping fails to install? Won't some tests end up failing if that's the case?

test/connection-checker/src/net.rs
```diff
@@ -74,3 +75,53 @@ pub fn send_ping(opt: &Opt, destination: IpAddr) -> eyre::Result<()> {
...
+// We use the ping command (which has capabilities/setuid set) to get around that.
+#[cfg(target_os = "linux")]
+pub fn send_ping(opt: &Opt, destination: IpAddr) -> eyre::Result<()> {
+    eprintln!("Leaking ICMP packets to {destination}");
+
+    let mut cmd = std::process::Command::new("ping");
+
+    // NOTE: Rounding up to nearest second, since some versions don't support fractional
+    //       seconds
+    let timeout_sec = ((opt.leak_timeout + 1000 - 1) / 1000).to_string();
```
hulthe, line 105: `ConnCheckerHandle::check_connection` sort of relies on the timing here. Are we sure this won't cause a bug? If there's doubt, maybe we should just change the type of `Opt::leak_timeout` to seconds as well.

test/connection-checker/src/net.rs
```diff
@@ -74,3 +75,53 @@ pub fn send_ping(opt: &Opt, destination: IpAddr) -> eyre::Result<()> {
 
     Ok(())
 }
+
+#[cfg(target_os = "macos")]
+pub fn send_ping(opt: &Opt, destination: IpAddr) -> eyre::Result<()> {
```
hulthe, line 80: maybe add a comment explaining why this function is different from the windows one

test/connection-checker/src/net.rs
dlon: Changed to seconds. :)

test/connection-checker/src/net.rs
dlon: Done!

hulthe (approved):

PR #6683, opened 2024-08-27
Add trixie to list of debian versions to generate deb repositories for

faern:
We already support `testing`. So this is not a big deal. Just that when you install/upgrade to trixie then `lsb_release -cs` will return `trixie` and not `testing`. So if we intend to support Debian versions in testing, we should probably add the names of the distros as soon as they are decided/used by the OS.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6683)


raksooo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6683#-:-O5InLsP7g_hIhrZLFui:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6696, opened 2024-08-28
Fix off-screen element detection issue

mojganii:
Due to the increasing number of items in VPN settings, our current UI test struggles to locate elements that are off-screen. This update resolves the issue by implementing scrolling to find the targeted elements.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696)

### DIFF
--- a/ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsDataSource.swift
+++ b/ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsDataSource.swift
@@ -457,6 +457,7 @@ final class VPNSettingsDataSource: UITableViewDiffableDataSource<
         header.accessibilityIdentifier = .wireGuardPortsCell
         header.titleLabel.text = title
         header.accessibilityCustomActionName = title
+        header.isExpanded = isExpanded(.wireGuardPorts)
         header.infoButtonHandler = { [weak self] in
             if let self {
                 self.delegate?.showInfo(for: .wireGuardPorts)
@@ -500,6 +501,7 @@ final class VPNSettingsDataSource: UITableViewDiffableDataSource<
         header.accessibilityIdentifier = .wireGuardObfuscationCell
         header.titleLabel.text = title
         header.accessibilityCustomActionName = title
+        header.isExpanded = isExpanded(.wireGuardObfuscation)
         header.didCollapseHandler = { [weak self] header in
             guard let self else { return }
 
@@ -529,6 +531,7 @@ final class VPNSettingsDataSource: UITableViewDiffableDataSource<
         header.accessibilityIdentifier = .udpOverTCPPortCell
         header.titleLabel.text = title
         header.accessibilityCustomActionName = title
+        header.isExpanded = isExpanded(.wireGuardObfuscationPort)
         header.didCollapseHandler = { [weak self] header in
             guard let self else { return }
 
@@ -558,6 +561,7 @@ final class VPNSettingsDataSource: UITableViewDiffableDataSource<
         header.accessibilityIdentifier = .quantumResistantTunnelCell
         header.titleLabel.text = title
         header.accessibilityCustomActionName = title
+        header.isExpanded = isExpanded(.quantumResistance)
         header.didCollapseHandler = { [weak self] header in
             guard let self else { return }
 
@@ -592,6 +596,15 @@ final class VPNSettingsDataSource: UITableViewDiffableDataSource<
         }
 
         return nil
+    }
+
+    /*
+      Since we are dequeuing headers, it's crucial to maintain the state of expansion.
+      Using screenshots as a single source of truth to capture the state allows us to determine whether headers are expanded or not.
+     */
+    private func isExpanded(_ section: Section) -> Bool {
+        let snapshot = snapshot()
+        return snapshot.numberOfItems(inSection: section) != 0
     }
 }
 
--- a/ios/MullvadVPNUITests/Pages/VPNSettingsPage.swift
+++ b/ios/MullvadVPNUITests/Pages/VPNSettingsPage.swift
@@ -15,10 +15,11 @@ class VPNSettingsPage: Page {
     }
 
     private func cellExpandButton(_ cellAccessiblityIdentifier: AccessibilityIdentifier) -> XCUIElement {
-        let table = app.tables[AccessibilityIdentifier.vpnSettingsTableView]
-        let matchingCells = table.otherElements.containing(.any, identifier: cellAccessiblityIdentifier.rawValue)
+        let tableView = app.tables[AccessibilityIdentifier.vpnSettingsTableView]
+        let matchingCells = tableView.otherElements[cellAccessiblityIdentifier.rawValue]
         let expandButton = matchingCells.buttons[AccessibilityIdentifier.expandButton]
-
+        let lastCell = tableView.cells.allElementsBoundByIndex.last!
+        tableView.scrollDownToElement(element: lastCell)
         return expandButton
     }
 
--- a/ios/MullvadVPNUITests/XCUIElement+Extensions.swift
+++ b/ios/MullvadVPNUITests/XCUIElement+Extensions.swift
@@ -16,4 +16,25 @@ extension XCUIElement {
         _ = XCTWaiter().wait(for: [expectation], timeout: timeout)
         return !exists
     }
+
+    func scrollDownToElement(element: XCUIElement, maxScrolls: UInt = 5) {
+        var count = 0
+        while !element.isVisible && count < maxScrolls {
+            swipeUp(velocity: .slow)
+            count += 1
+        }
+    }
+
+    func scrollUpToElement(element: XCUIElement, maxScrolls: UInt = 5) {
+        var count = 0
+        while !element.isVisible && count < maxScrolls {
+            swipeDown(velocity: .slow)
+            count += 1
+        }
+    }
+
+    var isVisible: Bool {
+        guard self.exists && !self.frame.isEmpty else { return false }
+        return XCUIApplication().windows.element(boundBy: 0).frame.contains(self.frame)
+    }
 }


buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsDataSource.swift` line 602 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5SC20WEVCERwTMgLPK:-O5SC20WEVCERwTMgLPL:b401cxu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPN/View%20controllers/VPNSettings/VPNSettingsDataSource.swift#L602)):*
> ```Swift
> 
>     private func isExpanded(_ section: Section) -> Bool {
>         let snapshot = snapshot()
> ```

I realize that the only reason the previous solution worked, was that the table view was not big enough to actually start recycling headers, since the only place where the information about being expanded was the header itself, and it's always false by default, had the table been big enough, we would have started recycling expanded headers, and hilarity would have ensued.

That being said, this solution has the same problem. By doing this, we're implicitly relying on the view states to keep the information whether headers are expanded or not, which when they get reused, will not work correctly.

The `VPNSettingsViewModel` should instead keep the information whether headers are expanded here. 
___
*[`ios/MullvadVPNUITests/XCUIElement+Extensions.swift` line 23 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5S8_lZ-CVjhgdNHoib:-O5S8_lZ-CVjhgdNHoic:b-oifqbu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPNUITests/XCUIElement%2BExtensions.swift#L23)):*
This will not always work, `swipeUp` doesn't let you control where the swipe will end, so you might scroll past the element you're trying to reach. 

Same for the `swipeDown` usage below. 

I'm also confused now, because the `tap` documentation says the following : 
> If the element exists within a scrollable view but is offscreen, XCTest will attempt to scroll the element onscreen before performing the tap.

@niklasberglund do you know what could be the problem that we are facing here ?
___
*[`ios/MullvadVPNUITests/XCUIElement+Extensions.swift` line 36 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5S7khl2ogEHWbO7D80:-O5S7khl2ogEHWbO7D81:b-24diaa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPNUITests/XCUIElement%2BExtensions.swift#L36)):*
I don't think we need this, there's already the [`isHittable`](https://developer.apple.com/documentation/xctest/xcuielement/1500561-ishittable property which does what we want 

> `isHittable` returns `true` if the element exists and can be clicked, tapped, or pressed at its current location. It returns `false` if the element does not exist, is offscreen, or is covered by another element.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsDataSource.swift` line 602 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5SC20WEVCERwTMgLPK:-O5lmnZEEZVcw9aRonlL:bqv4i5b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPN/View%20controllers/VPNSettings/VPNSettingsDataSource.swift#L602)):*

I realize that the only reason the previous solution worked, was that the table view was not big enough to actually start recycling headers, since the only place where the information about being expanded was the header itself, and it's always false by default, had the table been big enough, we would have started recycling expanded headers, and hilarity would have ensued.

That being said, this solution has the same problem. By doing this, we're implicitly relying on the view states to keep the information whether headers are expanded or not, which when they get reused, will not work correctly.

The `VPNSettingsViewModel` should instead keep the information whether headers are expanded here.
</blockquote></details>

I think it will be safe to ask the snapshot for this information. It's more or less a view model itself and keeps track of the raw data you're providing it with. If x number of items have been added to the snapshot and you ask it at any later time, you should get x number of items back.

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPNUITests/XCUIElement+Extensions.swift` line 23 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5S8_lZ-CVjhgdNHoib:-O5n8gLB5L_p_tbcLGY6:b-3mpy2y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPNUITests/XCUIElement%2BExtensions.swift#L23)):*
> If the element exists within a scrollable view but is offscreen, XCTest will attempt to scroll the element onscreen before performing the tap.

Oohh really? I was not aware of this.

> `swipeUp` doesn't let you control where the swipe will end, so you might scroll past the element you're trying to reach.

If its scrolling past the element(and this swiping is still necessary) you could specify a slower gesture velocity https://developer.apple.com/documentation/xctest/xcuicoordinate/3752787-swipeup

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPNUITests/XCUIElement+Extensions.swift` line 23 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5S8_lZ-CVjhgdNHoib:-O5qyk2Y21huVieog-1G:b303iew) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPNUITests/XCUIElement%2BExtensions.swift#L23)):*
> If the element exists within a scrollable view but is offscreen, XCTest will attempt to scroll the element onscreen before performing the tap.

But usually it won't exist because of table view optimisation right? It will only exist if its partially visible or maybe a tiny bit off screen? So the automatic scrolling is not as helpful as it sounds

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsDataSource.swift` line 602 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5SC20WEVCERwTMgLPK:-O5rseej3nBtGQUKR9ZD:b-dj0ny1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPN/View%20controllers/VPNSettings/VPNSettingsDataSource.swift#L602)):*

I think it will be safe to ask the snapshot for this information. It's more or less a view model itself and keeps track of the raw data you're providing it with. If x number of items have been added to the snapshot and you ask it at any later time, you should get x number of items back.
</blockquote></details>

@buggmagnet your original option is true.the problems is related to dequeue the header.but there are two solutions here:
1- keep tracking of expanding state in the view model
2- use screenshot to tell us about the state which is the thing @rablador mentioned and I agree to ask screenshot for the state instead of keeping the flag in view model. 
___
*[`ios/MullvadVPNUITests/XCUIElement+Extensions.swift` line 23 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5S8_lZ-CVjhgdNHoib:-O5rtwIpEhJFOT_pTrrh:b-s7w323) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPNUITests/XCUIElement%2BExtensions.swift#L23)):*

> If the element exists within a scrollable view but is offscreen, XCTest will attempt to scroll the element onscreen before performing the tap.

But usually it won't exist because of table view optimisation right? It will only exist if its partially visible or maybe a tiny bit off screen? So the automatic scrolling is not as helpful as it sounds
</blockquote></details>

regarding `tap` method. it's not working about the `UITableViewHeaderFooterView` but when it comes to `UITableViewCell` then there is no need to scroll to find the element.

@niklasberglund do we have a safer way to look for an element or velocity can solve the problem? I was thinking to scroll page by page.
I see what @buggmagnet is referring but I couldn't find a better solution .
___
*[`ios/MullvadVPNUITests/XCUIElement+Extensions.swift` line 36 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5S7khl2ogEHWbO7D80:-O5rrGHhCwO4xBFy5AEY:by515ab) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPNUITests/XCUIElement%2BExtensions.swift#L36)):*

I don't think we need this, there's already the [`isHittable`](https://developer.apple.com/documentation/xctest/xcuielement/1500561-ishittable property which does what we want 

> `isHittable` returns `true` if the element exists and can be clicked, tapped, or pressed at its current location. It returns `false` if the element does not exist, is offscreen, or is covered by another element.
</blockquote></details>

yeah, but `isHittable` relies on `isUserInteractionEnabled`, then if the item is already finable by accessible identifier but it's not clickable then `isHittable` returns false. checking the `exists` flag to recognize if the element is there or not is safer IMHO.

buggmagnet (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @mojganii, @niklasberglund, and @rablador)

___
*[`ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsDataSource.swift` line 602 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5SC20WEVCERwTMgLPK:-O5sBGOjBUJH2FejPlCL:b-70b95i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPN/View%20controllers/VPNSettings/VPNSettingsDataSource.swift#L602)):*

@buggmagnet your original option is true.the problems is related to dequeue the header.but there are two solutions here:
1- keep tracking of expanding state in the view model
2- use screenshot to tell us about the state which is the thing @rablador mentioned and I agree to ask screenshot for the state instead of keeping the flag in view model.
</blockquote></details>

Alright, if you both feel this is an acceptable solution, then let's go with it :thumbsup: 

However, We should add a comment explaining that we are using the snapshot as the source of truth on whether headers are expanded or not.
___
*[`ios/MullvadVPNUITests/XCUIElement+Extensions.swift` line 23 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5S8_lZ-CVjhgdNHoib:-O5sHXcUCJPTs2OUV19A:b-nablcg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPNUITests/XCUIElement%2BExtensions.swift#L23)):*

regarding `tap` method. it's not working about the `UITableViewHeaderFooterView` but when it comes to `UITableViewCell` then there is no need to scroll to find the element.

@niklasberglund do we have a safer way to look for an element or velocity can solve the problem? I was thinking to scroll page by page.
I see what @buggmagnet is referring but I couldn't find a better solution .
</blockquote></details>

We discussed a bit more with @mojganii, and we came to the conclusion that while the solution is not perfect, it is working, and it is an improvement over what we currently have, therefore, we are going with this.

If the tests start failing again in the future, of if this doesn't scale, we will revisit this later.
___
*[`ios/MullvadVPNUITests/XCUIElement+Extensions.swift` line 20 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5sD7_XEivg5Er1HuXV:-O5sD7_YBnDrNj7_clvD:bdnecmw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ecbdea1fec1388250ab0f4639d4efd753b96815c/ios/MullvadVPNUITests/XCUIElement%2BExtensions.swift#L20)):*
> ```Swift
>     }
> 
>     func scrollUpToElement(element: XCUIElement, maxScrolls: UInt = 5) {
> ```

As a side note, shouldn't that be scroll *down* to element, since we're using a swipe *up* motion ? (same for the other method)

niklasberglund (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696)* status: 2 of 3 files reviewed, 4 unresolved discussions (waiting on @buggmagnet, @mojganii, and @rablador)

___
*[`ios/MullvadVPNUITests/XCUIElement+Extensions.swift` line 23 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5S8_lZ-CVjhgdNHoib:-O5sZAUv2Fwl-F0sHEAv:b-ux8ond) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPNUITests/XCUIElement%2BExtensions.swift#L23)):*

We discussed a bit more with @mojganii, and we came to the conclusion that while the solution is not perfect, it is working, and it is an improvement over what we currently have, therefore, we are going with this.

If the tests start failing again in the future, of if this doesn't scale, we will revisit this later.
</blockquote></details>

@mojganii @buggmagnet nice 👍
___
*[`ios/MullvadVPNUITests/XCUIElement+Extensions.swift` line 20 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5sD7_XEivg5Er1HuXV:-O5s_LGz3xrrGQfjZz2b:b-8roea2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ecbdea1fec1388250ab0f4639d4efd753b96815c/ios/MullvadVPNUITests/XCUIElement%2BExtensions.swift#L20)):*

As a side note, shouldn't that be scroll *down* to element, since we're using a swipe *up* motion ? (same for the other method)
</blockquote></details>

Agree with @buggmagnet and wonder if scrolling up is necessary or do we actually only need to scroll down?

rablador (dismissed):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @mojganii)

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/View controllers/VPNSettings/VPNSettingsDataSource.swift` line 602 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5SC20WEVCERwTMgLPK:-O5sLZWh6FQBhgWpxBZC:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/daea3f87fbb1a1985250b0f5804de14ac5e3dfa6/ios/MullvadVPN/View%20controllers/VPNSettings/VPNSettingsDataSource.swift#L602)):*

Alright, if you both feel this is an acceptable solution, then let's go with it :thumbsup: 

However, We should add a comment explaining that we are using the snapshot as the source of truth on whether headers are expanded or not.
</blockquote></details>

Done.
___
*[`ios/MullvadVPNUITests/XCUIElement+Extensions.swift` line 20 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-O5sD7_XEivg5Er1HuXV:-O5wIYzTDl8ct30Ur7fP:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ecbdea1fec1388250ab0f4639d4efd753b96815c/ios/MullvadVPNUITests/XCUIElement%2BExtensions.swift#L20)):*

Agree with @buggmagnet and wonder if scrolling up is necessary or do we actually only need to scroll down?
</blockquote></details>

Done.

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6696#-:-O5wNTNfDqx8nPWXtxZs:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6717, opened 2024-09-02
Fix relays not being filtered correctly when Daita is enabled

rablador:
Current cached relay list has no servers with Daita support, so to avoid possibly empty location lists when Daita is on on new app install we should update the json.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6717)


buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6717#-:-O5n6qaLAsh1aYQHC-xX:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved

PR #6737, opened 2024-09-04
Fix bug in publish-app-to-repositories so stable are published to beta

faern:
This script was only publishing the app to *one* repository. Either stable or beta. This meant that whenever we published a stable app, superseding the previous beta, the stable repos would receive it, but the beta repos would not. So people on the beta repos would be stuck on the beta, which would be older than the stable release.

This change makes it so that `publish-app-to-repositories.sh` always publish the app to the beta repos. And then conditionally, if it's a stable release, *also* publishes it to the stable repositories.

I extracted most of the logic into a function. To make it easier to call it twice.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6737)

### DIFF
--- a/ci/publish-app-to-repositories.sh
+++ b/ci/publish-app-to-repositories.sh
@@ -107,15 +107,28 @@ function notify_repository_service {
     mv "$tmp_notify_file" "$notify_file"
 }
 
-stable_or_beta="stable"
-if [[ $version == *"-beta"* ]]; then
-    stable_or_beta="beta"
+function publish_app_to_repo {
+    # source files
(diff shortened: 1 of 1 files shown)


raksooo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6737#-:-O5x-kWI22m-G2NyfIQ-:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6739, opened 2024-09-04
Fix selected relay label being too long when multihoping with Daita

rablador:
Currently, when we turn on multihop and Daita on, the label "server abc using Daita via server xyz" is too long, and cut in the middle.

This fix may seem unintuitive in that it uses a stackview with label and image rather than the previous custom buttom. However, adding multiline support to button titles requires extra constraints to align text with button bounds, plus opens up a world of headache when trying to freely position the button image. With the stackviews we can move things around freely while still keeping the original look and feel by emulating button pressing behavior.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6739)

### DIFF
--- a/ios/MullvadVPN/View controllers/Tunnel/ConnectionPanelView.swift
+++ b/ios/MullvadVPN/View controllers/Tunnel/ConnectionPanelView.swift
@@ -29,15 +29,15 @@ class ConnectionPanelView: UIView {
 
     var connectedRelayName = "" {
         didSet {
-            collapseButton.accessibilityIdentifier = .relayStatusCollapseButton
-            collapseButton.setTitle(connectedRelayName, for: .normal)
-            collapseButton.accessibilityLabel = NSLocalizedString(
+            collapseView.accessibilityIdentifier = .relayStatusCollapseButton
+            collapseView.title.text = connectedRelayName
+            collapseView.accessibilityLabel = NSLocalizedString(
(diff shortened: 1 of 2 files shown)


buggmagnet (approved):
> With the stackviews we can move things around freely while still keeping the original look and feel by emulating button pressing behavior.
It does not emulate the `touchUpInside` part of a button 🙃
With a button, if you start the press inside the touch area, but then slide your finger away, eventually, the action won't trigger (you can see how the highlight goes away once your finger is far enough)

With this change, once the user has pressed, they cannot cancel that action, they have to commit to it.
Given we do this only to toggle the panel showing on or off, it's not a big deal here.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6739#-:-O601P-W-qNCU2TOguHE:by5c7ow)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

rablador (commented):
Yeah, I know, but I decided to ignore it. As you say, this is a simple non-destructive toggle, so I think it's fine. Especially considering the pros def outweighs the cons.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6739#-:-O60UZZeDA2IA_zqPFNB:b-wboo8o)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6767, opened 2024-09-09
Store the raw JSON payload from the getRelays API call

rablador:
Instead of storing the parsed relays to disk we should store the raw json response. This will help us avoid old structure being read when doing migration to new object properties in our relay related data.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767)

### DIFF
--- a/ios/MullvadREST/ApiHandlers/RESTAPIProxy.swift
+++ b/ios/MullvadREST/ApiHandlers/RESTAPIProxy.swift
@@ -107,13 +107,14 @@ extension REST {
                     switch httpStatus {
                     case let httpStatus where httpStatus.isSuccess:
                         return .decoding {
-                            let serverRelays = try self.responseDecoder.decode(
+                            // Discarding result since we're only interested in knowing that it's parseable.
+                            _ = try self.responseDecoder.decode(
                                 ServerRelaysResponse.self,
                                 from: data
                             )
                             let newEtag = response.value(forHTTPHeaderField: HTTPHeader.etag)
 
-                            return .newContent(newEtag, serverRelays)
+                            return .newContent(newEtag, data)
                         }
 
                     case .notModified where etag != nil:
@@ -284,7 +285,7 @@ extension REST {
 
     public enum ServerRelaysCacheResponse {
         case notModified
-        case newContent(_ etag: String?, _ value: ServerRelaysResponse)
+        case newContent(_ etag: String?, _ rawData: Data)
     }
 
     private struct CreateApplePaymentRequest: Encodable {
--- a/ios/MullvadREST/Relay/CachedRelays.swift
+++ b/ios/MullvadREST/Relay/CachedRelays.swift
@@ -8,7 +8,7 @@
 
 import Foundation
 
-/// A struct that represents the relay cache on disk
+/// A struct that represents the relay cache in memory
 public struct CachedRelays: Codable, Equatable {
     /// E-tag returned by server
     public let etag: String?
--- a/ios/MullvadREST/Relay/IPOverrideWrapper.swift
+++ b/ios/MullvadREST/Relay/IPOverrideWrapper.swift
@@ -18,21 +18,23 @@ public class IPOverrideWrapper: RelayCacheProtocol {
         self.ipOverrideRepository = ipOverrideRepository
     }
 
-    public func read() throws -> CachedRelays {
+    public func read() throws -> StoredRelays {
         let cache = try relayCache.read()
         let relayResponse = apply(overrides: ipOverrideRepository.fetchAll(), to: cache.relays)
+        let rawData = try REST.Coding.makeJSONEncoder().encode(relayResponse)
 
-        return CachedRelays(relays: relayResponse, updatedAt: cache.updatedAt)
+        return try StoredRelays(etag: cache.etag, rawData: rawData, updatedAt: cache.updatedAt)
     }
 
-    public func readPrebundledRelays() throws -> CachedRelays {
+    public func readPrebundledRelays() throws -> StoredRelays {
         let prebundledRelays = try relayCache.readPrebundledRelays()
         let relayResponse = apply(overrides: ipOverrideRepository.fetchAll(), to: prebundledRelays.relays)
+        let rawData = try REST.Coding.makeJSONEncoder().encode(relayResponse)
 
-        return CachedRelays(relays: relayResponse, updatedAt: prebundledRelays.updatedAt)
+        return try StoredRelays(etag: prebundledRelays.etag, rawData: rawData, updatedAt: prebundledRelays.updatedAt)
     }
 
-    public func write(record: CachedRelays) throws {
+    public func write(record: StoredRelays) throws {
         try relayCache.write(record: record)
     }
 
--- a/ios/MullvadREST/Relay/RelayCache.swift
+++ b/ios/MullvadREST/Relay/RelayCache.swift
@@ -12,59 +12,66 @@ import MullvadTypes
 public protocol RelayCacheProtocol {
     /// Reads from a cached list,
     /// which falls back to reading from prebundled relays if there was no cache hit
-    func read() throws -> CachedRelays
+    func read() throws -> StoredRelays
     /// Reads the relays file that were prebundled with the app installation.
     ///
     /// > Warning: Prefer `read()` over this unless there is an explicit need to read
     /// relays from the bundle, because those might contain stale data.
-    func readPrebundledRelays() throws -> CachedRelays
-    func write(record: CachedRelays) throws
+    func readPrebundledRelays() throws -> StoredRelays
+    func write(record: StoredRelays) throws
 }
 
 /// - Warning: `RelayCache` should not be used directly. It should be used through `IPOverrideWrapper` to have
 /// ip overrides applied.
 public final class RelayCache: RelayCacheProtocol {
-    private let fileCache: any FileCacheProtocol<CachedRelays>
+    private let fileURL: URL
+    private let fileCache: any FileCacheProtocol<StoredRelays>
 
     /// Designated initializer
     public init(cacheDirectory: URL) {
-        fileCache = FileCache(fileURL: cacheDirectory.appendingPathComponent("relays.json", isDirectory: false))
+        fileURL = cacheDirectory.appendingPathComponent("relays.json", isDirectory: false)
+        fileCache = FileCache(fileURL: fileURL)
     }
 
     /// Initializer that accepts a custom FileCache implementation. Used in tests.
-    init(fileCache: some FileCacheProtocol<CachedRelays>) {
+    init(fileCache: some FileCacheProtocol<StoredRelays>) {
+        fileURL = FileManager.default.temporaryDirectory.appendingPathComponent("relays.json", isDirectory: false)
         self.fileCache = fileCache
     }
 
-    /// Safely read the cache file from disk using file coordinator and fallback to prebundled
-    /// relays in case if the relay cache file is missing.
-    public func read() throws -> CachedRelays {
+    /// Safely read the cache file from disk using file coordinator and fallback in the following manner:
+    /// 1. If there is a file but it's not decodable, try to parse into the old cache format. If it's still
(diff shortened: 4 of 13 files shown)


pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadREST/Relay/StoredRelays.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-O6LvX01BPgYsjLcyiUk:-O6LvX01BPgYsjLcyiUl:biw3rof) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7017e92c716a228f48e32e152dbae1211ed96ffd/ios/MullvadREST/Relay/StoredRelays.swift#L25)):*
> ```Swift
>     public var relays: REST.ServerRelaysResponse {
>         get throws {
>             try REST.Coding.makeJSONDecoder().decode(REST.ServerRelaysResponse.self, from: rawData)
> ```

How often do we _get_ it? Could we make this lazy as to not decode the whole relay list all of the time?

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPN/RelayCacheTracker/RelayCacheTracker.swift` line 122 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-O6Pj5nJ-h9mCi2DT_1R:-O6Pj5nJ-h9mCi2DT_1S:bufolv3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7017e92c716a228f48e32e152dbae1211ed96ffd/ios/MullvadVPN/RelayCacheTracker/RelayCacheTracker.swift#L122)):*
> ```Swift
>         )
> 
>         let now = Date()
> ```

As per the discussion earlier, I do not believe `updatedAt` should be changed to the current timestamp. Instead, we shouldn't change it at all or change it such that it forces a relay list update.

The reason we have the hotfix is because there will be cases where people cannot reliably reach our API - but we should do our best to fetch a new relay list if  we know that the data on disk is incomplete.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadREST/Relay/StoredRelays.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-O6LvX01BPgYsjLcyiUk:-O6PjgOM5IaeRLTZWOnS:b-sewl67) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7017e92c716a228f48e32e152dbae1211ed96ffd/ios/MullvadREST/Relay/StoredRelays.swift#L25)):*

How often do we _get_ it? Could we make this lazy as to not decode the whole relay list all of the time?
</blockquote></details>

Also, could we not store the `REST.ServerRelaysResponse` as is here? The fact that this is a fallible getter gets on my nerves and inflates the diff size here.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767)* status: 0 of 13 files reviewed, 2 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadREST/Relay/StoredRelays.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-O6LvX01BPgYsjLcyiUk:-O6PkLyP8IJkbmktLhl-:b-ruvnp8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7017e92c716a228f48e32e152dbae1211ed96ffd/ios/MullvadREST/Relay/StoredRelays.swift#L25)):*

Also, could we not store the `REST.ServerRelaysResponse` as is here? The fact that this is a fallible getter gets on my nerves and inflates the diff size here.
</blockquote></details>

The above could be achieved by deserializing the relay list from the raw data and never serializing it to disk in the first place.

buggmagnet (changes requested):
> Reviewed 3 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767)* status: 3 of 13 files reviewed, 3 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadREST/Relay/RelayCache.swift` line 43 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-O6QAXZZ1O9Iu1WY_Ddr:-O6QAXZZ1O9Iu1WY_Dds:bq58bhk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7017e92c716a228f48e32e152dbae1211ed96ffd/ios/MullvadREST/Relay/RelayCache.swift#L43)):*
> ```Swift
>     public func read() throws -> StoredRelays {
>         do {
>             return try fileCache.read()
> ```

The first time we launch the app with this patch, reading from the cache will fail because a `CachedRelays` type was stored, but we're trying to decode a `StoredRelays` type.

In this case, we want to be able to create a `StoredRelays` directly from a `CachedRelays`.
We probably want something along those lines

```swift
    public init(cachedRelays: CachedRelays) throws {
        self.etag = cachedRelays.etag
        self.rawData = try JSONEncoder().encode(cachedRelays.relays)
        self.updatedAt = cachedRelays.updatedAt
    }
```

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767)* status: 3 of 13 files reviewed, 3 unresolved discussions (waiting on @pinkisemils)

___
*[`ios/MullvadREST/Relay/RelayCache.swift` line 43 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-O6QAXZZ1O9Iu1WY_Ddr:-O6QJ9fl8pgC_BDN-LX5:b-d6h3at) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7017e92c716a228f48e32e152dbae1211ed96ffd/ios/MullvadREST/Relay/RelayCache.swift#L43)):*

The first time we launch the app with this patch, reading from the cache will fail because a `CachedRelays` type was stored, but we're trying to decode a `StoredRelays` type.

In this case, we want to be able to create a `StoredRelays` directly from a `CachedRelays`.
We probably want something along those lines

```swift
    public init(cachedRelays: CachedRelays) throws {
        self.etag = cachedRelays.etag
        self.rawData = try JSONEncoder().encode(cachedRelays.relays)
        self.updatedAt = cachedRelays.updatedAt
    }
```
</blockquote></details>

Yes, with fallback on the bundled relays. As also discussed with @pinkisemils, we should do something like the above.
___
*[`ios/MullvadREST/Relay/StoredRelays.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-O6LvX01BPgYsjLcyiUk:-O6QIBA28XqLFmyldCII:b-a722) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7017e92c716a228f48e32e152dbae1211ed96ffd/ios/MullvadREST/Relay/StoredRelays.swift#L25)):*

The above could be achieved by deserializing the relay list from the raw data and never serializing it to disk in the first place.
</blockquote></details>

I'm not sure I follow. So we shouldn't store the raw data after all...?
___
*[`ios/MullvadVPN/RelayCacheTracker/RelayCacheTracker.swift` line 122 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-O6Pj5nJ-h9mCi2DT_1R:-O6QIa3sAOVir4-7qKJB:b-o980dz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7017e92c716a228f48e32e152dbae1211ed96ffd/ios/MullvadVPN/RelayCacheTracker/RelayCacheTracker.swift#L122)):*

As per the discussion earlier, I do not believe `updatedAt` should be changed to the current timestamp. Instead, we shouldn't change it at all or change it such that it forces a relay list update.

The reason we have the hotfix is because there will be cases where people cannot reliably reach our API - but we should do our best to fetch a new relay list if  we know that the data on disk is incomplete.
</blockquote></details>

If we force a new fetch here there would be no need for the hotfix either, since the idea behind the fix is to graft some new properties into old structure without the need for network calls. I think leaving it makes more sense.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767)* status: 3 of 13 files reviewed, 3 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadREST/Relay/StoredRelays.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-O6LvX01BPgYsjLcyiUk:-O6QMrXCEX6S30JzQb2N:beh42ei) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7017e92c716a228f48e32e152dbae1211ed96ffd/ios/MullvadREST/Relay/StoredRelays.swift#L25)):*

I'm not sure I follow. So we shouldn't store the raw data after all...?
</blockquote></details>

We should store `rawData`, but fail the deserialization if we cannot deserialize `rawData` into `ServerRelyResponse` during deserializaiton. We can chat about this in person.
___
*[`ios/MullvadVPN/RelayCacheTracker/RelayCacheTracker.swift` line 122 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-O6Pj5nJ-h9mCi2DT_1R:-O6QNfPq0Hl4XdMJcFai:bxpfybi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7017e92c716a228f48e32e152dbae1211ed96ffd/ios/MullvadVPN/RelayCacheTracker/RelayCacheTracker.swift#L122)):*

If we force a new fetch here there would be no need for the hotfix either, since the idea behind the fix is to graft some new properties into old structure without the need for network calls. I think leaving it makes more sense.
</blockquote></details>

We can force the fetch, but we can never rely on the fetch working. Even if we don't force a new fetch, we shouldn't update the timestamp every time the grafting takes place.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767)* status: 9 of 13 files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadREST/Relay/StoredRelays.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-O6LvX01BPgYsjLcyiUk:-O6RK1nf5Lgocjr8k93J:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7017e92c716a228f48e32e152dbae1211ed96ffd/ios/MullvadREST/Relay/StoredRelays.swift#L25)):*

We discussed this offline, and came to the conclusion taht we can make the `init` failable, create this as a `let` property and do the JSON parsing once when we instantiate this to avoid decoding from raw JSON every time.
</blockquote></details>

Done.

pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767#-:-O6SW1PSA_A17XojQTQp:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

pinkisemils (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6767)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

PR #6768, opened 2024-09-09
Add mullvad-encrypted-dns-proxy crate for API obfuscation

pinkisemils:
I've added a crate that will enable us to use `apisocks5` from our Rust code. It can:
- Resolve configs from arbitrary domain names
- Create proxy configurations form IPv6 addresses
- Use said proxy configurations to proxy data to our API

The `rustls` client configuration looks eerily similar to that of the `mullvad-api`'s client config, but I am not sure if it should be reused. I half-expect this crate to be a dependency of `mullvad-api` in one way or another, but I can also remove the AsyncRead and AsyncWrite impls for the `Forwarder`. I also expect my crusty rust code to not be quite up to spec with the best way to write code in September of 2024, so you needn't be gentle. 

Also, this is a sign to have a healthy discussion as to why this crate _shouldn't_ be named `mullvad-api-socks5-dns-obfuscated-encrypted-port-fowrard-proxy`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ members = [
     "mullvad-jni",
     "mullvad-management-interface",
     "mullvad-nsis",
+    "mullvad-encrypted-dns-proxy",
     "mullvad-paths",
     "mullvad-problem-report",
     "mullvad-relay-selector",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1340,20 +1340,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "07698b8420e2f0d6447a436ba999ec85d8fbf2a398bbd737b82cac4a2e96e512"
 dependencies = [
  "async-trait",
+ "bytes",
  "cfg-if",
  "data-encoding",
  "enum-as-inner",
  "futures-channel",
  "futures-io",
  "futures-util",
+ "h2 0.3.26",
+ "http 0.2.12",
  "idna 0.4.0",
  "ipnet",
  "once_cell",
  "rand 0.8.5",
+ "rustls",
+ "rustls-pemfile",
  "serde",
  "thiserror",
  "tinyvec",
  "tokio",
+ "tokio-rustls",
  "tracing",
  "url",
 ]
@@ -1373,10 +1379,12 @@ dependencies = [
  "parking_lot",
  "rand 0.8.5",
  "resolv-conf",
+ "rustls",
  "serde",
  "smallvec",
  "thiserror",
  "tokio",
+ "tokio-rustls",
  "tracing",
 ]
 
@@ -2379,6 +2387,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "mullvad-encrypted-dns-proxy"
+version = "0.0.0"
+dependencies = [
+ "env_logger 0.10.2",
+ "hickory-resolver",
+ "log",
+ "rustls",
+ "tokio",
+ "webpki-roots",
+]
+
+[[package]]
 name = "mullvad-exclude"
 version = "0.0.0"
 dependencies = [
@@ -2404,6 +2424,7 @@ dependencies = [
  "cbindgen",
  "libc",
  "log",
+ "mullvad-encrypted-dns-proxy",
  "oslog",
  "shadowsocks-service",
  "talpid-tunnel-config-client",
@@ -4852,6 +4873,12 @@ name = "wasm-bindgen-shared"
 version = "0.2.92"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "af190c94f2773fdb3729c55b007a722abb5384da03bc0986df4c289bf5567e96"
+
+[[package]]
+name = "webpki-roots"
+version = "0.25.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5f20c57d8d7db6d3b86154206ae5d8fba62dd39573114de97c2cb0578251f8e1"
 
 [[package]]
 name = "which"
--- a/mullvad-encrypted-dns-proxy/Cargo.toml
+++ b/mullvad-encrypted-dns-proxy/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "mullvad-encrypted-dns-proxy"
+description = "A port forwarding proxy that retrieves its configuration from a AAAA record over DoH"
+authors.workspace = true
+repository.workspace = true
+license.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+tokio = { workspace = true, features = [ "macros" ] }
+log = { workspace = true }
+hickory-resolver = { version = "0.24.1", features = [ "dns-over-https-rustls" ]}
+webpki-roots = "0.25.0"
+rustls = "0.21"
+
+[dev-dependencies]
+env_logger = { workspace = true }
--- a/mullvad-encrypted-dns-proxy/examples/forwarder.rs
+++ b/mullvad-encrypted-dns-proxy/examples/forwarder.rs
@@ -1,0 +1,42 @@
+use std::env::args;
+
+use mullvad_encrypted_dns_proxy::{config_resolver, Forwarder};
+use tokio::net::TcpListener;
+
+/// This can be tested out by using curl:
+/// `curl https://api.mullvad.net:$port/app/v1/relays --resolve api.mullvad.net:$port:$addr`
+///  where $addr and $port are the listening address of the proxy (bind_addr).
+#[tokio::main]
+async fn main() {
+    env_logger::init();
+
+    let bind_addr = args().nth(1).unwrap_or("127.0.0.1:0".to_owned());
+
+    let resolvers = config_resolver::default_resolvers();
+    let configs = config_resolver::resolve_configs(&resolvers, "frakta.eu")
+        .await
+        .expect("Failed to resolve configs");
+
+    let proxy_config = configs
+        .into_iter()
+        .find(|c| c.obfuscation.is_some())
+        .expect("No XOR config");
+    println!("Proxy config in use: {:?}", proxy_config);
+
+    let listener = TcpListener::bind(bind_addr)
+        .await
+        .expect("Failed to bind listener socket");
+
+    let listen_addr = listener
+        .local_addr()
+        .expect("failed to obtain listen address");
+    println!("Listening on {listen_addr}");
+
+    while let Ok((client_conn, client_addr)) = listener.accept().await {
+        println!("Incoming connection from {client_addr}");
+        let connected = Forwarder::connect(&proxy_config)
+            .await
+            .expect("failed to connect to obfuscator");
+        let _ = connected.forward(client_conn).await;
+    }
+}
--- a/mullvad-encrypted-dns-proxy/src/config/mod.rs
+++ b/mullvad-encrypted-dns-proxy/src/config/mod.rs
@@ -1,0 +1,124 @@
+//! Parse and use various proxy configurations as they are retrieved via AAAA records, hopefully
+//! served by DoH resolvers.
+
+use core::fmt;
+use std::net::{Ipv6Addr, SocketAddrV4};
+
+mod plain;
+mod xor;
+
+pub use xor::XorKey;
+
+/// All the errors that can happen during deserialization of a [`ProxyConfig`].
+#[derive(Debug, Eq, PartialEq)]
+pub enum Error {
+    /// The proxy type field has a value this library is not compatible with
+    UnknownProxyType(u16),
+    /// The XorV1 proxy type is deprecated and not supported
+    XorV1Unsupported,
+    /// The port is not valid
+    InvalidPort(u16),
+    /// The key to use for XOR obfuscation was empty (all zeros)
+    EmptyXorKey,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::UnknownProxyType(t) => write!(f, "Unknown type of proxy: {t:#x}"),
+            Self::XorV1Unsupported => write!(f, "XorV1 proxy types are not supported"),
+            Self::InvalidPort(port) => write!(f, "Port {port} is not valid for remote endpoint"),
+            Self::EmptyXorKey => write!(f, "The key material for XOR obfuscation is empty"),
+        }
+    }
+}
+
+impl std::error::Error for Error {}
+
+/// Type of a proxy configuration. Derived from the 2nd hextet of an IPv6 address in network byte
+/// order. E.g. an IPv6 address such as `7f7f:2323::`  would have a proxy type value of `0x2323`.
+#[derive(PartialEq, Debug)]
+enum ProxyType {
+    Plain,
+    XorV1,
+    XorV2,
+}
+
+impl TryFrom<[u8; 2]> for ProxyType {
+    type Error = Error;
+
+    fn try_from(bytes: [u8; 2]) -> Result<Self, Self::Error> {
+        match u16::from_le_bytes(bytes) {
+            0x01 => Ok(Self::Plain),
+            0x02 => Ok(Self::XorV1),
+            0x03 => Ok(Self::XorV2),
+            unknown => Err(Error::UnknownProxyType(unknown)),
+        }
+    }
+}
+
+pub trait Obfuscator: Send {
+    /// Applies obfuscation to a given buffer of bytes. Changes the data in place.
+    fn obfuscate(&mut self, buffer: &mut [u8]);
+}
+
+/// Represents a Mullvad Encrypted DNS proxy configuration. Created by parsing
+/// the config out of an IPv6 address resolved over DoH.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ProxyConfig {
+    /// The remote address to connect to the proxy over. This is the address
+    /// on the internet where the proxy is listening.
+    pub addr: SocketAddrV4,
+    /// If the proxy requires some obfuscation of the data sent to/received from it,
+    /// it's represented by an obfuscation config here.
+    pub obfuscation: Option<ObfuscationConfig>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum ObfuscationConfig {
+    XorV2(xor::XorKey),
+}
+
+impl ObfuscationConfig {
+    /// Instantiate an obfuscator from the given obfuscation config.
+    pub fn create_obfuscator(&self) -> Box<dyn Obfuscator> {
+        match self {
+            Self::XorV2(key) => Box::new(xor::XorObfuscator::new(*key)),
+        }
+    }
+}
+
+impl TryFrom<Ipv6Addr> for ProxyConfig {
+    type Error = Error;
+
+    fn try_from(ip: Ipv6Addr) -> Result<Self, Self::Error> {
+        let data = ip.octets();
+
+        let proxy_type_bytes = <[u8; 2]>::try_from(&data[2..4]).unwrap();
+        let proxy_config_payload = <[u8; 12]>::try_from(&data[4..16]).unwrap();
+
+        let proxy_type = ProxyType::try_from(proxy_type_bytes)?;
+
+        match proxy_type {
+            ProxyType::Plain => plain::parse_plain(proxy_config_payload),
+            ProxyType::XorV1 => Err(Error::XorV1Unsupported),
+            ProxyType::XorV2 => xor::parse_xor(proxy_config_payload),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::net::Ipv6Addr;
+
+    use super::{Error, ProxyConfig};
+
+    #[test]
+    fn wrong_proxy_type() {
+        let addr: Ipv6Addr = "ffff:2345::".parse().unwrap();
+        match ProxyConfig::try_from(addr) {
+            Err(Error::UnknownProxyType(0x4523)) => (),
+            anything_else => panic!("Unexpected proxy config parse result: {anything_else:?}"),
+        }
+    }
+}
--- a/mullvad-encrypted-dns-proxy/src/config/plain.rs
+++ b/mullvad-encrypted-dns-proxy/src/config/plain.rs
@@ -1,0 +1,79 @@
+use std::net::{Ipv4Addr, SocketAddrV4};
+
+/// Parse a proxy config that does not obfuscate. It still can circumvent censorship since it is reaching our
+/// API through a different IP address.
+///
+/// A plain configuration is represented by proxy type [`super::ProxyType::Plain`]. Normally the
+/// input to this function will come from the last 12 bytes of an IPv6 address. A plain
+/// configuration interprets the following bytes from a given IPv6 address:
+/// bytes 2-4 - u16le - proxy type - must be 0x01
+/// bytes 4-8 - [u8; 4] - 4 bytes representing the proxy IPv4 address
+/// bytes 8-10 - u16le - port on which the proxy is listening
+///
+/// Given the above, an IPv6 address `2001:100:b9d5:9a75:3804::` will have the second hexlet
+/// (0x0100) represent the proxy type, the following 2 hexlets (0xb9d5, 0x9a75) - the IPv4 address
+/// of the proxy endpoint, and the final hexlet represents the port for the proxy endpoint - the
+/// remaining bytes can be ignored.
+pub fn parse_plain(data: [u8; 12]) -> Result<super::ProxyConfig, super::Error> {
+    let (ip_bytes, tail) = data.split_first_chunk::<4>().unwrap();
+    let (port_bytes, _tail) = tail.split_first_chunk::<2>().unwrap();
+
+    let ip = Ipv4Addr::from(*ip_bytes);
+    let port = u16::from_le_bytes(*port_bytes);
+    if port == 0 {
+        return Err(super::Error::InvalidPort(0));
+    }
+    let addr = SocketAddrV4::new(ip, port);
+
+    Ok(super::ProxyConfig {
+        addr,
+        obfuscation: None,
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use std::net::{Ipv6Addr, SocketAddrV4};
+
+    use crate::config::{Error, ProxyConfig};
+
+    #[test]
+    fn parsing() {
+        struct Test {
+            input: Ipv6Addr,
+            expected: Result<ProxyConfig, Error>,
+        }
+        let tests = vec![
+            Test {
+                input: "2001:100:7f00:1:3905::".parse::<Ipv6Addr>().unwrap(),
+                expected: Ok(ProxyConfig {
+                    addr: "127.0.0.1:1337".parse::<SocketAddrV4>().unwrap(),
+                    obfuscation: None,
+                }),
+            },
+            Test {
+                input: "2001:100:c0a8:101:bb01::".parse::<Ipv6Addr>().unwrap(),
+                expected: Ok(ProxyConfig {
+                    addr: "192.168.1.1:443".parse::<SocketAddrV4>().unwrap(),
+                    obfuscation: None,
+                }),
+            },
+            Test {
+                input: "2001:100:c0a8:101:bb01:404::".parse::<Ipv6Addr>().unwrap(),
+                expected: Ok(ProxyConfig {
+                    addr: "192.168.1.1:443".parse::<SocketAddrV4>().unwrap(),
+                    obfuscation: None,
+                }),
+            },
+            Test {
+                input: "2001:100:c0a8:101:0000:404::".parse::<Ipv6Addr>().unwrap(),
+                expected: Err(Error::InvalidPort(0)),
+            },
+        ];
+
+        for t in tests {
+            let parsed = ProxyConfig::try_from(t.input);
+            assert_eq!(parsed, t.expected);
+        }
+    }
+}
--- a/mullvad-encrypted-dns-proxy/src/config/xor.rs
+++ b/mullvad-encrypted-dns-proxy/src/config/xor.rs
@@ -1,0 +1,204 @@
+use core::fmt;
+use std::net::{Ipv4Addr, SocketAddrV4};
+
+/// Parse a proxy config that XORs all traffic with the given key.
+///
+/// A Xor configuration is represented by the proxy type `ProxyType::XorV2`. There used to be a `XorV1`, but it
+/// is deprecated and should not be used.
+///
+/// The following bytes of an IPv6 address are interpreted to derive a Xor configuration:
+/// bytes 2-4 - u16le - proxy type - must be 0x03
+/// bytes 4-8 - [u8; 4] - 4 bytes representing the proxy IPv4 address
+/// bytes 8-10 - u16le - port on which the proxy is listening
+/// bytes 10-16 - [u8; 6] - xor key bytes. 0x00 marks a premature end of the key
+/// Given the above, `2001:300:b9d5:9a75:3a04:eafd:1100:ad9e` will have the second hexlet (0x0300)
+/// represent the proxy type, the next 2 hexlets (0xb9d5,0x9a75) represent the IPv4 address for the
+/// proxy endpoint, the next hexlet (`3a04`) represents the port for the proxy endpoint, and
+/// the final 3 hexlets `eafd:1100:ad9e` represent the xor key (0xEA, 0xFD, 0x11).
+pub fn parse_xor(data: [u8; 12]) -> Result<super::ProxyConfig, super::Error> {
+    let (ip_bytes, tail) = data.split_first_chunk::<4>().unwrap();
+    let (port_bytes, key_bytes) = tail.split_first_chunk::<2>().unwrap();
+    let key_bytes = <[u8; 6]>::try_from(key_bytes).unwrap();
+
+    let ip = Ipv4Addr::from(*ip_bytes);
+    let port = u16::from_le_bytes(*port_bytes);
+    if port == 0 {
+        return Err(super::Error::InvalidPort(port));
+    }
+    let addr = SocketAddrV4::new(ip, port);
+
+    let key = XorKey::try_from(key_bytes)?;
+
+    Ok(super::ProxyConfig {
+        addr,
+        obfuscation: Some(super::ObfuscationConfig::XorV2(key)),
+    })
+}
+
+/// A bunch of bytes, representing a "key" Simply meaning a slice of bytes that the data
+/// will be XORed with.
+#[derive(Copy, Clone, Eq, PartialEq, Hash)]
+pub struct XorKey {
+    data: [u8; 6],
+    len: usize,
+}
+
+impl XorKey {
+    /// Return the XOR key material. Will always have at least length 1.
+    pub fn key_data(&self) -> &[u8] {
+        &self.data[0..self.len]
+    }
+}
+
+impl fmt::Debug for XorKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "0x")?;
+        for byte in self.key_data() {
+            write!(f, "{byte:0>2x}")?;
+        }
+        Ok(())
+    }
+}
+
+impl TryFrom<[u8; 6]> for XorKey {
+    type Error = super::Error;
+
+    fn try_from(mut key_bytes: [u8; 6]) -> Result<Self, Self::Error> {
+        let key_len = key_bytes
+            .iter()
+            .position(|b| *b == 0x00)
+            .unwrap_or(key_bytes.len());
+        if key_len == 0 {
+            return Err(super::Error::EmptyXorKey);
+        }
+
+        // Reset bytes after terminating null to zeros.
+        // Allows simpler implementations of Eq and Hash
+        key_bytes[key_len..].fill(0);
+
+        Ok(Self {
+            data: key_bytes,
+            len: key_len,
+        })
+    }
+}
+
+#[derive(Debug)]
+pub struct XorObfuscator {
+    key: XorKey,
+    key_index: usize,
+}
+
+impl XorObfuscator {
+    pub fn new(key: XorKey) -> Self {
+        Self { key, key_index: 0 }
+    }
+}
+
+impl super::Obfuscator for XorObfuscator {
+    fn obfuscate(&mut self, buffer: &mut [u8]) {
+        let key_data = self.key.key_data();
+        for byte in buffer {
+            *byte ^= key_data[self.key_index % key_data.len()];
+            self.key_index = (self.key_index + 1) % key_data.len();
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::net::{Ipv6Addr, SocketAddrV4};
+
+    use crate::config::xor::{XorKey, XorObfuscator};
+    use crate::config::{Error, ObfuscationConfig, Obfuscator, ProxyConfig};
+
+    #[test]
+    fn xor_parsing() {
+        struct Test {
+            input: Ipv6Addr,
+            expected: Result<ProxyConfig, Error>,
+        }
+        let tests = vec![
+            Test {
+                input: "2001:300:7f00:1:3905:0102:304:506"
+                    .parse::<Ipv6Addr>()
+                    .unwrap(),
+                expected: Ok(ProxyConfig {
+                    addr: "127.0.0.1:1337".parse::<SocketAddrV4>().unwrap(),
+                    obfuscation: Some(ObfuscationConfig::XorV2(
+                        XorKey::try_from([0x01, 0x02, 0x03, 0x04, 0x05, 0x06]).unwrap(),
+                    )),
+                }),
+            },
+            Test {
+                input: "2001:300:7f00:1:3905:0100:304:506"
+                    .parse::<Ipv6Addr>()
+                    .unwrap(),
+                expected: Ok(ProxyConfig {
+                    addr: "127.0.0.1:1337".parse::<SocketAddrV4>().unwrap(),
+                    obfuscation: Some(ObfuscationConfig::XorV2(
+                        XorKey::try_from([0x01, 0, 0, 0, 0, 0]).unwrap(),
+                    )),
+                }),
+            },
+            Test {
+                input: "2001:300:c0a8:101:bb01:ff04:204:0"
+                    .parse::<Ipv6Addr>()
+                    .unwrap(),
+                expected: Ok(ProxyConfig {
+                    addr: "192.168.1.1:443".parse::<SocketAddrV4>().unwrap(),
+                    obfuscation: Some(ObfuscationConfig::XorV2(
+                        XorKey::try_from([0xff, 0x04, 0x02, 0x04, 0, 0]).unwrap(),
+                    )),
+                }),
+            },
+        ];
+
+        for t in tests {
+            let parsed = ProxyConfig::try_from(t.input);
+            assert_eq!(parsed, t.expected);
+        }
+    }
+
+    #[test]
+    fn obfuscation() {
+        const INPUT: &[u8] = &[1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+        let mut payload = INPUT.to_vec();
+
+        let xor_key = XorKey::try_from([0xff, 0x04, 0x02, 0x04, 0x00, 0x00]).unwrap();
+
+        let mut xor_obfuscator = XorObfuscator::new(xor_key);
+        let mut xor_deobfuscator = XorObfuscator::new(xor_key);
+
+        xor_obfuscator.obfuscate(&mut payload);
+
+        assert_eq!(
+            payload,
+            &[0xfe, 0x06, 0x01, 0x00, 0xfa, 0x02, 0x05, 0x0c, 0xf6, 0x0e]
+        );
+
+        xor_deobfuscator.obfuscate(&mut payload);
+        assert_eq!(INPUT, payload.as_slice());
+    }
+
+    // Before XOR-v2 there was XOR-v1, which is now deprecated. This test verifies that the old Xor
+    // config does not deserialize.
+    #[test]
+    fn old_xor_addr() {
+        match ProxyConfig::try_from(
+            "2001:200:7f00:1:3905:0102:304:506"
+                .parse::<Ipv6Addr>()
+                .unwrap(),
+        ) {
+            Err(Error::XorV1Unsupported) => (),
+            anything_else => panic!("Unexpected proxy config parse result: {anything_else:?}"),
+        }
+    }
+
+    #[test]
+    fn xor_key_debug_fmt() {
+        let key = XorKey::try_from([0x01, 0xff, 0x31, 0x00, 0x00, 0x00]).unwrap();
+        let key_str = format!("{key:?}");
+        assert_eq!(key_str, "0x01ff31");
+    }
+}
--- a/mullvad-encrypted-dns-proxy/src/config_resolver.rs
+++ b/mullvad-encrypted-dns-proxy/src/config_resolver.rs
@@ -1,0 +1,140 @@
+//! Resolve valid proxy configurations via DoH.
+//!
+use crate::config;
+use core::fmt;
+use hickory_resolver::{config::*, error::ResolveError, TokioAsyncResolver};
+use rustls::ClientConfig;
+use std::{net::IpAddr, sync::Arc};
+
+/// The port to connect to the DoH resolvers over.
+const RESOLVER_PORT: u16 = 443;
+
+pub struct Nameserver {
+    pub name: String,
+    pub addr: Vec<IpAddr>,
+}
+
+#[derive(Debug)]
+pub struct ResolutionError(ResolveError);
+
+impl fmt::Display for ResolutionError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl std::error::Error for ResolutionError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        self.0.source()
+    }
+}
+
+/// Returns a set of well known public DoH resolvers. A sane default in many cases.
+pub fn default_resolvers() -> Vec<Nameserver> {
+    vec![
+        Nameserver {
+            name: "one.one.one.one".to_owned(),
+            addr: vec!["1.1.1.1".parse().unwrap(), "1.0.0.1".parse().unwrap()],
+        },
+        Nameserver {
+            name: "dns.google".to_owned(),
+            addr: vec!["8.8.8.8".parse().unwrap(), "8.8.4.4".parse().unwrap()],
+        },
+        Nameserver {
+            name: "dns.quad9.net".to_owned(),
+            addr: vec![
+                "9.9.9.9".parse().unwrap(),
+                "149.112.112.112".parse().unwrap(),
+            ],
+        },
+    ]
+}
+
+/// Look up the `domain` towards the given `resolvers`, and try to deserialize all the returned
+/// AAAA records into [`ProxyConfig`](config::ProxyConfig)s.
+pub async fn resolve_configs(
+    resolvers: &[Nameserver],
+    domain: &str,
+) -> Result<Vec<config::ProxyConfig>, ResolutionError> {
+    let mut resolver_config = ResolverConfig::new();
+    for resolver in resolvers.iter() {
+        let ns_config_group = NameServerConfigGroup::from_ips_https(
+            &resolver.addr,
+            RESOLVER_PORT,
+            resolver.name.clone(),
+            false,
+        )
+        .into_inner();
+        for ns_config in ns_config_group {
+            resolver_config.add_name_server(ns_config);
+        }
+    }
+
+    resolver_config.set_tls_client_config(Arc::new(client_config_tls12()));
+
+    resolve_config_with_resolverconfig(resolver_config, Default::default(), domain).await
+}
+
+pub async fn resolve_config_with_resolverconfig(
+    resolver_config: ResolverConfig,
+    options: ResolverOpts,
+    domain: &str,
+) -> Result<Vec<config::ProxyConfig>, ResolutionError> {
+    let resolver = TokioAsyncResolver::tokio(resolver_config, options);
+    let lookup = resolver
+        .ipv6_lookup(domain)
+        .await
+        .map_err(ResolutionError)?;
+
+    let addrs = lookup.into_iter().map(|aaaa_record| aaaa_record.0);
+
+    let mut proxy_configs = Vec::new();
+    for addr in addrs {
+        match config::ProxyConfig::try_from(addr) {
+            Ok(proxy_config) => {
+                log::trace!("IPv6 {addr} parsed into proxy config: {proxy_config:?}");
+                proxy_configs.push(proxy_config);
+            }
+            Err(e) => log::error!("IPv6 {addr} fails to parse to a proxy config: {e}"),
+        }
+    }
+
+    Ok(proxy_configs)
+}
+
+fn client_config_tls12() -> ClientConfig {
+    use rustls::RootCertStore;
+    let mut root_store = RootCertStore::empty();
+    root_store.add_trust_anchors(webpki_roots::TLS_SERVER_ROOTS.iter().map(|ta| {
+        rustls::OwnedTrustAnchor::from_subject_spki_name_constraints(
+            ta.subject,
+            ta.spki,
+            ta.name_constraints,
+        )
+    }));
+
+    ClientConfig::builder()
+        .with_safe_default_cipher_suites()
+        .with_safe_default_kx_groups()
+        .with_safe_default_protocol_versions() // this enables TLS 1.2 and 1.3
+        .unwrap()
+        .with_root_certificates(root_store)
+        .with_no_client_auth()
+}
+
+#[cfg(test)]
+#[tokio::test]
+async fn test_resolution() {
+    let nameservers = vec![Nameserver {
+        addr: vec!["1.1.1.1".parse().unwrap()],
+        name: "one.one.one.one".to_owned(),
+    }];
+
+    let _ = resolve_configs(&nameservers, "frakta.eu").await.unwrap();
+}
+
+#[cfg(test)]
+#[test]
+fn default_resolvers_dont_panic() {
+    let _ = default_resolvers();
+}
--- a/mullvad-encrypted-dns-proxy/src/forwarder.rs
+++ b/mullvad-encrypted-dns-proxy/src/forwarder.rs
@@ -1,0 +1,194 @@
+//! Forward TCP traffic over various proxy configurations.
+
+use std::{
+    io,
+    task::{ready, Poll},
+};
+
+use tokio::{
+    io::{AsyncRead, AsyncWrite},
+    net::TcpStream,
+};
+
+use crate::config::Obfuscator;
+
+/// Forwards local traffic to a proxy endpoint, obfuscating it if the proxy config says so.
+///
+/// Obtain [`ProxyConfig`](crate::config::ProxyConfig)s with
+/// [resolve_configs](crate::config_resolver::resolve_configs).
+pub struct Forwarder {
+    read_obfuscator: Option<Box<dyn Obfuscator>>,
+    write_obfuscator: Option<Box<dyn Obfuscator>>,
+    server_connection: TcpStream,
+}
+
+impl Forwarder {
+    /// Create a forwarder that will connect to a given proxy endpoint.
+    pub async fn connect(proxy_config: &crate::config::ProxyConfig) -> io::Result<Self> {
+        let server_connection = TcpStream::connect(proxy_config.addr).await?;
+
+        let (read_obfuscator, write_obfuscator) =
+            if let Some(obfuscation_config) = &proxy_config.obfuscation {
+                (
+                    Some(obfuscation_config.create_obfuscator()),
+                    Some(obfuscation_config.create_obfuscator()),
+                )
+            } else {
+                (None, None)
+            };
+
+        Ok(Self {
+            read_obfuscator,
+            write_obfuscator,
+            server_connection,
+        })
+    }
+
+    /// Forwards traffic from the client stream to the remote proxy, obfuscating and deobfuscating
+    /// it in the process.
+    pub async fn forward(self, client_stream: TcpStream) {
+        let (server_read, server_write) = self.server_connection.into_split();
+        let (client_read, client_write) = client_stream.into_split();
+        let _ = tokio::join!(
+            forward(self.read_obfuscator, client_read, server_write),
+            forward(self.write_obfuscator, server_read, client_write)
+        );
+    }
+}
+
+impl tokio::io::AsyncRead for Forwarder {
+    fn poll_read(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &mut tokio::io::ReadBuf<'_>,
+    ) -> std::task::Poll<std::io::Result<()>> {
+        let socket = std::pin::pin!(&mut self.server_connection);
+        match ready!(socket.poll_read(cx, buf)) {
+            // in this case, we can read and deobfuscate.
+            Ok(()) => {
+                if let Some(read_obfuscator) = &mut self.read_obfuscator {
+                    read_obfuscator.obfuscate(buf.filled_mut());
+                }
+                Poll::Ready(Ok(()))
+            }
+            Err(err) => Poll::Ready(Err(err)),
+        }
+    }
+}
+
+impl tokio::io::AsyncWrite for Forwarder {
+    fn poll_write(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &[u8],
+    ) -> Poll<Result<usize, io::Error>> {
+        let socket = std::pin::pin!(&mut self.server_connection);
+        if let Err(err) = ready!(socket.poll_write_ready(cx)) {
+            return Poll::Ready(Err(err));
+        };
+
+        let mut owned_buf = buf.to_vec();
+        if let Some(write_obfuscator) = &mut self.write_obfuscator {
+            write_obfuscator.obfuscate(&mut owned_buf);
+        }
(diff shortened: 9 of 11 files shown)


pinkisemils:
Should we define the _default_ resolvers (cloudflare, cloud9, google) in this crate? 
What should the default resolvers be?

Cargo.toml
```diff
@@ -18,6 +18,7 @@ members = [
     "mullvad-jni",
     "mullvad-management-interface",
     "mullvad-nsis",
+    "mullvad-obfuscated-dns-proxy",
     "mullvad-paths",
```
Serock3, line 22:
```suggestion
    "mullvad-nsis",
    "mullvad-obfuscated-dns-proxy",
```

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 0 of 12 files reviewed, 1 unresolved discussion (waiting on @Serock3)

___
*[`Cargo.toml` line 22 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751405268:-O6PsY9x-MUmtX3eTt92:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/Cargo.toml#L22)):*


```suggestion
    "mullvad-nsis",
    "mullvad-obfuscated-dns-proxy",
```
</blockquote></details>

Done.

mullvad-obfuscated-dns-proxy/src/config_resolver.rs
```diff
@@ -0,0 +1,108 @@
...
+
+fn client_config_tls12() -> ClientConfig {
+    use rustls::RootCertStore;
+    let mut root_store = RootCertStore::empty();
+    root_store.add_trust_anchors(webpki_roots::TLS_SERVER_ROOTS.iter().map(|ta| {
+        rustls::OwnedTrustAnchor::from_subject_spki_name_constraints(
+            ta.subject,
+            ta.spki,
+            ta.name_constraints,
+        )
```
Serock3, line 83: webpki-roots could be bumbed to version "0.26.5" with this change
```suggestion
        rustls::OwnedTrustAnchor::from_subject_spki_name_constraints(
            ta.subject.as_ref(),
            ta.subject_public_key_info.as_ref().to_owned(),
            ta.name_constraints
                .as_ref()
                .map(|nc| nc.as_ref().to_owned()),
        )
```

mullvad-obfuscated-dns-proxy/src/config_resolver.rs
```diff
@@ -0,0 +1,108 @@
...
+    }));
+
+    ClientConfig::builder()
+        .with_safe_default_cipher_suites()
+        .with_safe_default_kx_groups()
+        .with_safe_default_protocol_versions() // this enables TLS 1.2 and 1.3
+        .unwrap()
```
Serock3, line 90: If you bump the versions of `rustls` to `0.23` and `webpki-roots` to `0.26`, this can be simplified to
```rust
root_store.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
```

Note that it relies on the `TrustAnchor` type being shared, you could also leave `rustls` at  `0.22` with the explicit conversion
```
let root_store = rustls::RootCertStore {
  roots: webpki_roots::TLS_SERVER_ROOTS.iter().cloned().collect(),
};
```

dlon (changes requested):
> Reviewed 5 of 12 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 6 of 12 files reviewed, 12 unresolved discussions (waiting on @pinkisemils and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/Cargo.lock` line 1 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6PntNdAHMLTvirlXOG:-O6PntNdAHMLTvirlXOH:b-h4g0za) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/Cargo.lock#L1)):*
> ```TOML
> # This file is automatically @generated by Cargo.
> ```

Is this supposed to have its own lock file? Is it not part of the same workspace?
___
*[`mullvad-obfuscated-dns-proxy/src/config/plain.rs` line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Pt_rCDyUVCrZIR-0B:-O6Pt_rCDyUVCrZIR-0C:b-t69cc7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/plain.rs#L14)):*
> ```XML
> 
> impl Plain {
>     pub async fn forward(
> ```

Never used?
___
*[`mullvad-obfuscated-dns-proxy/src/config/xor.rs` line 31 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Prii4D3T-35q29nJn:-O6Prii4D3T-35q29nJo:b-r4elt5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/xor.rs#L31)):*
> ```XML
>         let _ = cursor.read_u16::<LittleEndian>().unwrap();
>         let proxy_type = cursor.read_u16::<LittleEndian>().unwrap();
>         if proxy_type != 0x03 {
> ```

Could this not be a magic constant?
___
*[`mullvad-obfuscated-dns-proxy/src/config/xor.rs` line 48 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6PsCKc4aPpEnTREpNh:-O6PsCKc4aPpEnTREpNi:b-a1eu5v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/xor.rs#L48)):*
> ```XML
>             .collect::<Vec<_>>();
>         if xor_key.is_empty() {
>             return Err(Error::EmptyXorKey);
> ```

Could "plain" not be implemented using the same code with a zeroed xor key?
___
*[`mullvad-obfuscated-dns-proxy/src/config/xor.rs` line 137 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6PpkeL608si5jPWlXF:-O6PpkeL608si5jPWlXG:b-sj782f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/xor.rs#L137)):*
> ```XML
>     xor.obfuscate(&mut obfuscated);
>     dexor.obfuscate(&mut obfuscated);
>     assert_eq!(input, obfuscated.as_slice());
> ```

Super-nit: This is not "obfuscated".
___
*[`mullvad-obfuscated-dns-proxy/src/config/xor.rs` line 141 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6PsaRO-BvqLK2cUM0E:-O6PsaRO-BvqLK2cUM0F:b-sq7104) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/xor.rs#L141)):*
> ```XML
> 
> #[test]
> fn test_old_xor_addr() {
> ```

Could you rename or comment to explain the intent of this test?
___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 1 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6PxGHy-CJ2oxp7rBIg:-O6PxGHy-CJ2oxp7rBIh:bcwqgyz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L1)):*
> ```XML
> use std::{io, task::Poll};
> ```

Module is duplicated in `./forwarder`?
___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 31 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6PtkZI6a35iMdySaeI:-O6PtkZI6a35iMdySaeJ:b-i40y86) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L31)):*
> ```XML
>         let (client_read, client_write) = client_stream.into_split();
>         let handle = tokio::spawn(async move {
>             tokio::spawn(forward(self.read_obfuscator, client_read, server_write));
> ```

The double-`spawn` seems unnecessary. Besides, `handle.await` will not wait for the inner task to resolve. If that is intentional (but I'm guessing it's not), why not just *not* await it instead?
___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/forwarder/mod.rs` line 33 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Px88j5xALrQ9eV0lB:-O6Px88j5xALrQ9eV0lC:b-7dha14) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/forwarder/forwarder/mod.rs#L33)):*
> ```XML
>             tokio::spawn(forward(self.read_obfuscator, client_read, server_write));
>         });
>         let _ = forward(self.write_obfuscator, server_read, client_write).await;
> ```

I wonder if this needs to select either task failing. If it's guaranteed to resolve, then ignore this.
___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/forwarder/mod.rs` line 74 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6PvZtc7XJUaxcLfnRp:-O6PvZtc7XJUaxcLfnRq:bcl00p3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/forwarder/forwarder/mod.rs#L74)):*
> ```XML
>             }
>             Poll::Pending => {
>                 return Poll::Pending;
> ```

Nit: Could use `ready!` on `poll_...`.
___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/forwarder/mod.rs` line 112 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Pw-NtDx6CszQf-tD_:-O6Pw-NtDx6CszQf-tDa:b-4b13j1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/forwarder/forwarder/mod.rs#L112)):*
> ```XML
>         let mut bytes_received = &mut buf[..n_bytes_read];
> 
>         obfuscator.obfuscate(&mut bytes_received);
> ```

Nit: It's already borrowed.

mullvad-obfuscated-dns-proxy/src/config/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+                    proxies
+                        .xor
+                        .push(Xor::try_from(ip).map_err(Error::InvalidXor)?);
+                }
+                // this type is ignored.
+                ProxyType::XorV1 => continue,
+            }
+        }
+
+        Ok(proxies)
```
Serock3, line 75: Very much a style preference, but it seems you could remove the `ProxyType` enum entirely and use this
```suggestion
        let plain = ips
            .iter()
            .filter_map(|ip| Plain::try_from(*ip).ok())
            .collect();
        let xor = ips
            .iter()
            .filter_map(|ip| Xor::try_from(*ip).ok())
            .collect();
        Ok(AvailableProxies { plain, xor })
```

dlon (commented):
> Reviewed 6 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: all files reviewed, 15 unresolved discussions (waiting on @pinkisemils and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/src/main.rs` line 23 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Q0XDf8pMzfJGFcNSz:-O6Q0XDg0fZ9xZxVYmdr:b-e79y6f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/main.rs#L23)):*
> ```XML
>         .expect("Failed to bind listener socket");
>     while let Ok((client_conn, _client_addr)) = listener.accept().await {
>         let connected = crate::forwarder::Forwarder::connect(obfuscator.clone())
> ```

Nit: Unnecessary clone

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: all files reviewed, 17 unresolved discussions (waiting on @pinkisemils and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/Cargo.toml` line 2 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Q1Wr5DhZwZskGMrbA:-O6Q1Wr6Ct9xUhHlLYB_:b-jkbnp0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/Cargo.toml#L2)):*
> ```TOML
> [package]
> name = "mullvad-obfuscated-dns-proxy"
> ```

Can we please get a description here as well? We have a lot of crates and it is hard to know which one does what in certain circumstances. This is probably the least self explanatory from just the name.
___
*[`mullvad-obfuscated-dns-proxy/Cargo.toml` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Q1pjv1CKAvmC3hAZz:-O6Q1pjwBYClNo754yHY:b-i1o7i1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/Cargo.toml#L13)):*
> ```TOML
> 
> [dependencies]
> tokio = { version = "1", features = [ "full" ]}
> ```

This is already available as a workspace dependency. Use that.

Can we avoid the `full` feature? That seems more like for development. In production we should only enable what we need.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: all files reviewed, 18 unresolved discussions (waiting on @pinkisemils and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 86 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Q2VjA18pmv9uvB4Nz:-O6Q2VjA18pmv9uvB4O-:b-ba3ixv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L86)):*
> ```XML
> pub trait Obfuscator: Send {
>     fn addr(&self) -> SocketAddrV4;
>     fn obfuscate(&mut self, buffer: &mut [u8]);
> ```

Can we please add documentation for both these trait methods, to explain what the implementer is supposed to do.
___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 87 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Q2Hoz48rXpX9f6FAs:-O6Q2Hoz48rXpX9f6FAt:b-wrisro) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L87)):*
> ```XML
>     fn addr(&self) -> SocketAddrV4;
>     fn obfuscate(&mut self, buffer: &mut [u8]);
>     fn clone(&self) -> Box<dyn Obfuscator>;
> ```

Seems confusing to have a custom clone method that does not come from `Clone`?

faern:
> The rustls client configuration looks eerily similar to that of the mullvad-api's client config, but I am not sure if it should be reused

I think they can be kept separate, to not depend on each other. Unless it's a lot of work to define them two times. Having the crates be independent has value, as they can progress without blocking each other.

mullvad-obfuscated-dns-proxy/src/forwarder/forwarder/mod.rs
```diff
@@ -0,0 +1,116 @@
+use std::{io, task::Poll};
```
Serock3, line 1: This file seems to be an unused duplicate of `src/forwarder.mod`

Serock3 (changes requested):
Duplicate `forwarder` module

mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+        })
+    }
+    pub async fn forward(self, client_stream: TcpStream) {
+        let (server_read, server_write) = self.server_connection.into_split();
+        let (client_read, client_write) = client_stream.into_split();
+        let handle = tokio::spawn(async move {
+            tokio::spawn(forward(self.read_obfuscator, client_read, server_write));
+        });
+        let _ = forward(self.write_obfuscator, server_read, client_write).await;
+        let _ = handle.await;
```
Serock3, line 34: Perhaps a `tokio::join` would be more readable here

mullvad-obfuscated-dns-proxy/src/config/mod.rs
Serock3: I took a look at this again, and I wonder if maybe the simplest implementation would be to also remove `Plain::try_from` and `Xor::try_from` and put all the logic in the `AvailableProxies::try_from` implementation like this.

```rust
#[derive(Debug)]
pub enum Error {
    UnknownType(u16),
    InvalidXor(xor::Error),
    InvalidPlain(plain::Error),
    UnexpectedType(u16),
    EmptyXorKey,
}

impl TryFrom<Vec<Ipv6Addr>> for AvailableProxies {
    type Error = Error;

    fn try_from(ips: Vec<Ipv6Addr>) -> Result<Self, Self::Error> {
        let mut proxies = AvailableProxies {
            plain: vec![],
            xor: vec![],
        };

        for ip in ips {
            let mut data = Cursor::new(ip.octets());
            // skip the first 2 bytes since it's just padding to make the IP look more like a
            // legit IPv6 address.

            data.set_position(2);
            let proxy_type = data
                .read_u16::<LittleEndian>()
                .expect("IPv6 must have at least 16 bytes");
            match proxy_type {
                0x01 => {
                    let mut ipv4_bytes = [0u8; 4];
                    data.read_exact(&mut ipv4_bytes).unwrap();
                    let v4_addr = Ipv4Addr::from(ipv4_bytes);

                    let port = data.read_u16::<LittleEndian>().unwrap();
                    proxies.plain.push(Plain {
                        addr: SocketAddrV4::new(v4_addr, port),
                    });
                }
                0x02 => {
                    let mut ipv4_bytes = [0u8; 4];
                    data.read_exact(&mut ipv4_bytes).unwrap();
                    let v4_addr = Ipv4Addr::from(ipv4_bytes);

                    let port = data.read_u16::<LittleEndian>().unwrap();

                    let mut key_bytes = [0u8; 6];
                    data.read_exact(&mut key_bytes).unwrap();
                    let xor_key = key_bytes
                        .into_iter()
                        .filter(|byte| *byte != 0x00)
                        .collect::<Vec<_>>();
                    if xor_key.is_empty() {
                        return Err(Error::EmptyXorKey);
                    }
                    proxies.xor.push(Xor {
                        addr: SocketAddrV4::new(v4_addr, port),
                        xor_key,
                        key_index: 0,
                    });
                }
                // this type is ignored.
                0x03 => continue,
                unknown => return Err(Error::UnknownType(unknown)),
            }
        }

        Ok(proxies)
    }
}
```

It would remove the duplicate checks of the proxy type byte and some other boilerplate.

mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+};
+
+use crate::config::Obfuscator;
+
+pub struct Forwarder {
+    read_obfuscator: Box<dyn Obfuscator>,
+    write_obfuscator: Box<dyn Obfuscator>,
+    server_connection: TcpStream,
+}
+
```
Serock3, line 15: Why does the `Forwarder` use dynamic dispatch for the `Obfuscator`? Would it not be as simple to use generics instead?

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 4 of 12 files reviewed, 14 unresolved discussions (waiting on @dlon, @faern, and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/Cargo.lock` line 1 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6PntNdAHMLTvirlXOG:-O6PzD1W9viSmHKQ6lpj:brmivp2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/Cargo.lock#L1)):*

Is this supposed to have its own lock file? Is it not part of the same workspace?
</blockquote></details>

Will remove, I first worked on the crate in isolation.
___
*[`mullvad-obfuscated-dns-proxy/Cargo.toml` line 13 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Q1pjv1CKAvmC3hAZz:-O6ui6PtCjVCTFHKiB8L:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/Cargo.toml#L13)):*

This is already available as a workspace dependency. Use that.

Can we avoid the `full` feature? That seems more like for development. In production we should only enable what we need.
</blockquote></details>

Done.
___
*[`mullvad-obfuscated-dns-proxy/src/config_resolver.rs` line 90 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751462096:-O6uiBy6AAQ4hhvNDweY:bt2yt7a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config_resolver.rs#L90)):*

If you bump the versions of `rustls` to `0.23` and `webpki-roots` to `0.26`, this can be simplified to
```rust
let root_store = RootCertStore::from_iter(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
```

Note that it relies on the version of the `TrustAnchor` type matching for both crates, you could also leave `rustls` at  `0.22` with the explicit conversion
```
let root_store = rustls::RootCertStore {
  roots: webpki_roots::TLS_SERVER_ROOTS.iter().cloned().collect(),
};
```
</blockquote></details>

Yes, but `hickory-dns` first needs to bump it. Until they do, we can't use newer versions of `rustls`.
___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 35 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751482881:-O6uiLWJ5dSKuv46h8MW:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L35)):*


```suggestion
        data.set_position(2);
```
</blockquote></details>

Done.
___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 87 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Q2Hoz48rXpX9f6FAs:-O6Q5niE6Kr2PdsqPOz8:b-gy9fp1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L87)):*

Seems confusing to have a custom clone method that does not come from `Clone`?
</blockquote></details>

I was too stupid to have a trait depend on clone and have the trait be object safe.
___
*[`mullvad-obfuscated-dns-proxy/src/config/xor.rs` line 31 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Prii4D3T-35q29nJn:-O6uiV08AsasuPHE-KIk:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/xor.rs#L31)):*

Could this not be a magic constant?
</blockquote></details>

Done.
___
*[`mullvad-obfuscated-dns-proxy/src/config/xor.rs` line 48 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6PsCKc4aPpEnTREpNh:-O6Q5-ao6IFruPmGMQvA:b-hponqz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/xor.rs#L48)):*

Could "plain" not be implemented using the same code with a zeroed xor key?
</blockquote></details>

It could, but it is a distinct proxying type. Given that `2001:100::/96` will always be plain, and `2001:300::/96` will always be xor-v2. 
___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751642577:-O6uiofJ-bzfFgjKbLFq:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L34)):*

Perhaps a `tokio::join` would be more readable here
</blockquote></details>

Done.
___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1760712398:-O6uiY9y4taMNKfewlPr:b-vghsx0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/475893aa4e1a56dd9e9ed1c593eae57aa1e35298/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L15)):*

Why does the `Forwarder` use dynamic dispatch for the `Obfuscator`? Would it not be as simple to use generics instead?
</blockquote></details>

I'd argue that any user of `Forwarder` will then need to know and care about those types, which is not simpler. But this can be rewritten to use specific types instead. I am ambivalent as to which side we pick - I picked the more lazy option here.
___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/forwarder/mod.rs` line 1 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751620947:-O6uirF94R-vebnxjHXQ:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/forwarder/forwarder/mod.rs#L1)):*

This file seems to be an unused duplicate of `src/forwarder.mod`
</blockquote></details>

Done.

dlon (commented):
> Reviewed 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: all files reviewed, 12 unresolved discussions (waiting on @faern, @pinkisemils, and @Serock3)

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 11 of 12 files reviewed, 11 unresolved discussions (waiting on @dlon, @pinkisemils, and @Serock3)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6zvn0QEdzaLogbhJM8:-O6zvn0R1HkSEW-pjzOJ:bpi9zp4) (no related file):*
This crate has virtually no doc comments. I'd like to request a bit more documentation what the library does.

___

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 11 of 12 files reviewed, 12 unresolved discussions (waiting on @dlon, @pinkisemils, and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/Cargo.toml` line 2 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6zw7BV8FwP2MBkcFu8:-O6zw7BW2t5nqk9Q5ze9:bsy1t95) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d703f5085a54c6b3a38695108e3191bfd00c209c/mullvad-obfuscated-dns-proxy/Cargo.toml#L2)):*
> ```TOML
> [package]
> name = "mullvad-obfuscated-dns-proxy"
> ```

Rename the crate to `mullvad-encrypted-dns-proxy` mayhaps? Since that's the name we finally settled on.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 11 of 12 files reviewed, 13 unresolved discussions (waiting on @dlon, @pinkisemils, and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 75 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751495295:-O7-74h0E7dBG_5OFg4Q:b-73zqj0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L75)):*

I took a look at this again, and I wonder if maybe the simplest implementation would be to also remove `Plain::try_from` and `Xor::try_from` and put all the logic in the `AvailableProxies::try_from` implementation like this.

```rust
const PLAIN: u16 = 0x01;
const XORV2: u16 = 0x02;
const XORV1: u16 = 0x03;

#[derive(Debug)]
pub enum Error {
    UnknownType(u16),
    InvalidXor(xor::Error),
    InvalidPlain(plain::Error),
    UnexpectedType(u16),
    EmptyXorKey,
}

impl TryFrom<Vec<Ipv6Addr>> for AvailableProxies {
    type Error = Error;

    fn try_from(ips: Vec<Ipv6Addr>) -> Result<Self, Self::Error> {
        let mut proxies = AvailableProxies {
            plain: vec![],
            xor: vec![],
        };

        for ip in ips {
            let mut data = Cursor::new(ip.octets());
            // skip the first 2 bytes since it's just padding to make the IP look more like a
            // legit IPv6 address.

            data.set_position(2);
            let proxy_type = data
                .read_u16::<LittleEndian>()
                .expect("IPv6 must have at least 16 bytes");
            match proxy_type {
                PLAIN => {
                    let mut ipv4_bytes = [0u8; 4];
                    data.read_exact(&mut ipv4_bytes).unwrap();
                    let v4_addr = Ipv4Addr::from(ipv4_bytes);

                    let port = data.read_u16::<LittleEndian>().unwrap();
                    proxies.plain.push(Plain {
                        addr: SocketAddrV4::new(v4_addr, port),
                    });
                }
                XORV2 => {
                    let mut ipv4_bytes = [0u8; 4];
                    data.read_exact(&mut ipv4_bytes).unwrap();
                    let v4_addr = Ipv4Addr::from(ipv4_bytes);

                    let port = data.read_u16::<LittleEndian>().unwrap();

                    let mut key_bytes = [0u8; 6];
                    data.read_exact(&mut key_bytes).unwrap();
                    let xor_key = key_bytes
                        .into_iter()
                        .filter(|byte| *byte != 0x00)
                        .collect::<Vec<_>>();
                    if xor_key.is_empty() {
                        return Err(Error::EmptyXorKey);
                    }
                    proxies.xor.push(Xor {
                        addr: SocketAddrV4::new(v4_addr, port),
                        xor_key,
                        key_index: 0,
                    });
                }
                // this type is ignored.
                XORV1 => continue,
                unknown => return Err(Error::UnknownType(unknown)),
            }
        }

        Ok(proxies)
    }
}
```

It would remove the duplicate checks of the proxy type byte and some other boilerplate.
</blockquote></details>

I like that we can avoid the extra parsing and error handling of the invalid types. A code structure that allows parsing the type once and then just instantiating the correct obfuscator instance after that would be nicer indeed.
___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 87 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O7-6AOb1y-MZtKK3FNF:-O7-6AOb1y-MZtKK3FNG:bdvbc6j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d703f5085a54c6b3a38695108e3191bfd00c209c/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L87)):*
> ```XML
>     /// Provides the endpoint for the proxy. This address must be connected and all traffic to it
>     /// should first be obfuscated with `Obfuscator::obfuscate`.
>     fn addr(&self) -> SocketAddrV4;
> ```

I'd argue that the address of the *proxy server* is unrelated to the *obfuscation done on the data*. You could also avoid duplicating the socket address field and related code into every implementer if it was lifted out. I imagine something like:

```rust
pub struct Proxy {
    addr: SocketAddrV4,
    obfuscator: Box<dyn Obfuscator>,
}
```
___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 17 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O7-4s0w4LvOJ8mnAwzl:-O7-4s0w4LvOJ8mnAwzm:bsalofi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d703f5085a54c6b3a38695108e3191bfd00c209c/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L17)):*
> ```XML
> 
> impl Forwarder {
>     pub async fn connect(read_obfuscator: Box<dyn Obfuscator>) -> io::Result<Self> {
> ```

The argument name indicates the obfuscator will be used for *reading*. But currently it will be used for both reading and writing (it is cloned into `write_obfuscator`). If the current design is kept, I suggest that the argument name is changed to reflect this. So the API of `Forwarder` is more clear.

Serock3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: all files reviewed, 9 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[`mullvad-obfuscated-dns-proxy/src/config_resolver.rs` line 90 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751462096:-O7-_wEV-3hkNkpeYAkC:b-yz5l3w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config_resolver.rs#L90)):*

Yes, but `hickory-dns` first needs to bump it. Until they do, we can't use newer versions of `rustls`.
</blockquote></details>

So what about just bumping `webpki-roots` and doing the explicit conversion?
___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1760712398:-O737eJTEVmfq3gni_44:b-n3wqs4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/475893aa4e1a56dd9e9ed1c593eae57aa1e35298/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L15)):*

I'd argue that any user of `Forwarder` will then need to know and care about those types, which is not simpler. But this can be rewritten to use specific types instead. I am ambivalent as to which side we pick - I picked the more lazy option here.
</blockquote></details>

Maybe I'm being pedantic, and I'm sure it doesn't really matter in practice, but it bothers me that the `obfuscator.obfuscate(bytes_received);` call  in the `forwarder::forward` function does a v-table lookup for each message it reads. Could it not be a relatively hot part of the code?

Since the `Forwarder::forward` method consumes `self`, would it not be reasonable to combine `Forwarder::connect` and `Forwarder::forward` into a single function, and make it generic? Since also the `AsyncRead` and `AsyncWrite` implementations for the `Forwarder` type are unused, it seems that the entire contents of this file could be replaced with

```rust
pub async fn forward(obfuscator: impl Obfuscator, client_stream: TcpStream) -> io::Result<()> {
    let server_connection = TcpStream::connect(obfuscator.addr()).await?;
    let write_obfuscator = obfuscator.clone();
    let (server_read, server_write) = server_connection.into_split();
    let (client_read, client_write) = client_stream.into_split();
    let ((), ()) = tokio::try_join!(
        forward_inner(obfuscator, client_read, server_write),
        forward_inner(write_obfuscator, server_read, client_write)
    )?;
    Ok(())
}

async fn forward_inner(
    mut obfuscator: impl Obfuscator,
    mut source: impl AsyncRead + Unpin,
    mut sink: impl AsyncWrite + Unpin,
) -> io::Result<()> {
    use tokio::io::{AsyncReadExt, AsyncWriteExt};
    let mut buf = vec![0u8; 1024 * 64];
    while let Ok(n_bytes_read) = AsyncReadExt::read(&mut source, &mut buf).await {
        if n_bytes_read == 0 {
            break;
        }
        let bytes_received = &mut buf[..n_bytes_read];

        obfuscator.obfuscate(bytes_received);
        sink.write_all(bytes_received).await?;
    }
    Ok(())
}
```

(After replacing the return type of `Obfuscator::clone` with `Self`)

Serock3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: all files reviewed, 9 unresolved discussions (waiting on @faern and @pinkisemils)

___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1760712398:-O73KT092-bX_3sbGCAh:bf9k5nm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/475893aa4e1a56dd9e9ed1c593eae57aa1e35298/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L15)):*

Maybe I'm being pedantic, and I'm sure it doesn't really matter in practice, but it bothers me that the `obfuscator.obfuscate(bytes_received);` call  in the `forwarder::forward` function does a v-table lookup for each message it reads. Could it not be a relatively hot part of the code?

Since the `Forwarder::forward` method consumes `self`, would it not be reasonable to combine `Forwarder::connect` and `Forwarder::forward` into a single function, and make it generic? Since also the `AsyncRead` and `AsyncWrite` implementations for the `Forwarder` type are unused, it seems that the entire contents of this file could be replaced with

```rust
pub async fn forward(obfuscator: impl Obfuscator, client_stream: TcpStream) -> io::Result<()> {
    let server_connection = TcpStream::connect(obfuscator.addr()).await?;
    let write_obfuscator = obfuscator.clone();
    let (server_read, server_write) = server_connection.into_split();
    let (client_read, client_write) = client_stream.into_split();
    let ((), ()) = tokio::try_join!(
        forward_inner(obfuscator, client_read, server_write),
        forward_inner(write_obfuscator, server_read, client_write)
    )?;
    Ok(())
}

async fn forward_inner(
    mut obfuscator: impl Obfuscator,
    mut source: impl AsyncRead + Unpin,
    mut sink: impl AsyncWrite + Unpin,
) -> io::Result<()> {
    use tokio::io::{AsyncReadExt, AsyncWriteExt};
    let mut buf = vec![0u8; 1024 * 64];
    while let Ok(n_bytes_read) = AsyncReadExt::read(&mut source, &mut buf).await {
        if n_bytes_read == 0 {
            break;
        }
        let bytes_received = &mut buf[..n_bytes_read];

        obfuscator.obfuscate(bytes_received);
        sink.write_all(bytes_received).await?;
    }
    Ok(())
}
```

(After replacing the return type of `Obfuscator::clone` with `Self`)
</blockquote></details>

Here's a patch for the entire change in case you're interested https://github.com/mullvad/mullvadvpn-app/compare/add-obfuscated-dns-proxy...add-obfuscated-dns-proxy-patch?expand=1

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: all files reviewed, 9 unresolved discussions (waiting on @pinkisemils and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 87 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O7-6AOb1y-MZtKK3FNF:-O73RRoQA1PADEMWSPTX:b-ormoty) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d703f5085a54c6b3a38695108e3191bfd00c209c/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L87)):*

I'd argue that the address of the *proxy server* is unrelated to the *obfuscation done on the data*. You could also avoid duplicating the socket address field and related code into every implementer if it was lifted out. I imagine something like:

```rust
pub struct Proxy {
    addr: SocketAddrV4,
    obfuscator: Box<dyn Obfuscator>,
}
```
</blockquote></details>

Sorry, I meant `obfuscator: Option<Box<dyn Obfuscator>>`. Because the plain does not need one. So instead of there being a plain obfuscator, there can be no obfuscator. And obfuscation only happening if it is `Some(obfuscator)`.
___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1760712398:-O73RbhK62xAZCLT69YX:bhm5nsu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/475893aa4e1a56dd9e9ed1c593eae57aa1e35298/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L15)):*

Here's a patch for the entire change in case you're interested https://github.com/mullvad/mullvadvpn-app/compare/add-obfuscated-dns-proxy...add-obfuscated-dns-proxy-patch?expand=1
</blockquote></details>

Performance is not an issue here. That part feels like premature optimization. Please remember that this will only transport API traffic. So it's a few bytes to kilobytes of metadata only.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: all files reviewed, 9 unresolved discussions (waiting on @faern and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/src/config_resolver.rs` line 90 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751462096:-O73UvBgE2qDJd0WFZry:b2y1o9o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config_resolver.rs#L90)):*

So what about just bumping `webpki-roots` and doing the explicit conversion?
</blockquote></details>

I don't want to spend time here at all, but I expect that the older `rustls` crate might not be able to consume newer versions of `webpki`. 
___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 75 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751495295:-O73I0a99AaQ2izeL27_:bgk6vfv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L75)):*

I like that we can avoid the extra parsing and error handling of the invalid types. A code structure that allows parsing the type once and then just instantiating the correct obfuscator instance after that would be nicer indeed.
</blockquote></details>

I much prefer the standalone nature of those types.
___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 86 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6Q2VjA18pmv9uvB4Nz:-O73TWDk2wtIaD0ppj8-:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L86)):*

Can we please add documentation for both these trait methods, to explain what the implementer is supposed to do.
</blockquote></details>

Done.
___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 87 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O7-6AOb1y-MZtKK3FNF:-O73SxhH19-pXsX6Ojr1:b4ahbza) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d703f5085a54c6b3a38695108e3191bfd00c209c/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L87)):*

Sorry, I meant `obfuscator: Option<Box<dyn Obfuscator>>`. Because the plain does not need one. So instead of there being a plain obfuscator, there can be no obfuscator. And obfuscation only happening if it is `Some(obfuscator)`.
</blockquote></details>

This complicates things, because then the plain case must be a special one. A forwarder needs to know where to connect a TCP socket and it need something to obfuscate the bytes - and this interface provides these details.  Further, we do not know which bytes in the Ipv6 address will be used to specify the socket address, so the obfuscator itself must contain that data.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 1 of 15 files reviewed, 9 unresolved discussions (waiting on @dlon, @pinkisemils, and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 75 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751495295:-O74I2Iv7ly1yuV1glQx:b-oum7vd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L75)):*

I much prefer the standalone nature of those types.
</blockquote></details>

I just feel like the parsing is implemented twice. First it only parses a subset of the fields to tell the program what other, more complete parser to apply to the same data again. I don't care about the performance. But it feels both like duplicated effort and things that must be kept in sync for no reason. As well as extra unwraps and stuff that would otherwise not be needed.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 1 of 15 files reviewed, 9 unresolved discussions (waiting on @dlon, @pinkisemils, and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/Cargo.lock` line 1 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6PntNdAHMLTvirlXOG:-O74IaJs748W_cTMRsHx:b-pws98b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/Cargo.lock#L1)):*

Will remove, I first worked on the crate in isolation.
</blockquote></details>

Ping? I see things are pushed here but this is not removed. This file makes me nervous :D

Serock3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 1 of 15 files reviewed, 9 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1760712398:-O74boNgAeKG6rcQzms1:bh52zcz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/475893aa4e1a56dd9e9ed1c593eae57aa1e35298/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L15)):*

Performance is not an issue here. That part feels like premature optimization. Please remember that this will only transport API traffic. So it's a few bytes to kilobytes of metadata only.
</blockquote></details>

That's fair. I still think removing the `Forwarder`, if possible, makes the code significantly more readable.
___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 75 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751495295:-O74gRix95VVQMdBllet:b-um82gs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L75)):*

I just feel like the parsing is implemented twice. First it only parses a subset of the fields to tell the program what other, more complete parser to apply to the same data again. I don't care about the performance. But it feels both like duplicated effort and things that must be kept in sync for no reason. As well as extra unwraps and stuff that would otherwise not be needed.
</blockquote></details>

What do you think about moving the parsing logic for a slice of the byte sequence to the individual obfuscators? E.g.
```rust
impl From<&[u8; 16]> for Plain {
    fn from(data: &[u8; 16]) -> Self {
        let (v4_bytes, mut port_bytes) = data.split_first_chunk::<4>().unwrap();

        let v4_addr = Ipv4Addr::from(*v4_bytes);
        let port = port_bytes.read_u16::<LittleEndian>().unwrap();

        Plain {
            addr: SocketAddrV4::new(v4_addr, port),
        }
    }
}
```
and
```rust
impl TryFrom<&[u8; 16]> for Xor {
    type Error = Error;

    fn try_from(data: &[u8; 16]) -> Result<Self, Self::Error> {
        let (v4_bytes, mut remaining_bytes) = data.split_first_chunk::<4>().unwrap();

        let v4_addr = Ipv4Addr::from(*v4_bytes);
        let port = remaining_bytes.read_u16::<LittleEndian>().unwrap();

        let xor_key: Vec<u8> = remaining_bytes[..6]
            .iter()
            .copied()
            .filter(|&byte| byte != 0x00)
            .collect();

        if xor_key.is_empty() {
            return Err(Error::EmptyXorKey);
        }
        Ok(Xor {
            addr: SocketAddrV4::new(v4_addr, port),
            xor_key,
            key_index: 0,
        })
    }
}
```

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 1 of 15 files reviewed, 8 unresolved discussions (waiting on @dlon, @faern, and @pinkisemils)

___
*[`mullvad-obfuscated-dns-proxy/src/config_resolver.rs` line 90 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751462096:-O78EBxXBwYtFyHDN9kG:b-lxb0kh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config_resolver.rs#L90)):*

I don't want to spend time here at all, but I expect that the older `rustls` crate might not be able to consume newer versions of `webpki`.
</blockquote></details>

`webpki-roots` v0.26 can consume `rustls` v0.22 and above, but `hickory-resolver` only supports `rustls` v0.21 as of now. There seems to be no way to update on our end.

dlon (approved):
> Reviewed 8 of 14 files at r5, 4 of 5 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: all files reviewed, 8 unresolved discussions (waiting on @faern and @pinkisemils)

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 7 of 15 files reviewed, 8 unresolved discussions (waiting on @dlon, @faern, and @Serock3)

*[a discussion](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6zvn0QEdzaLogbhJM8:-O7DIkX44uQEL2en_DS3:b-896fix) (no related file):*

This crate has virtually no doc comments. I'd like to request a bit more documentation what the library does.
</blockquote></details>

Done.

___
___
*[`mullvad-obfuscated-dns-proxy/Cargo.lock` line 1 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6PntNdAHMLTvirlXOG:-O7DIgHiAMC18AtkHP5h:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/Cargo.lock#L1)):*

Ping? I see things are pushed here but this is not removed. This file makes me nervous :D
</blockquote></details>

Done.
___
*[`mullvad-obfuscated-dns-proxy/Cargo.toml` line 2 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O6zw7BV8FwP2MBkcFu8:-O7DEf2tE92oj66xswDB:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d703f5085a54c6b3a38695108e3191bfd00c209c/mullvad-obfuscated-dns-proxy/Cargo.toml#L2)):*

Rename the crate to `mullvad-encrypted-dns-proxy` mayhaps? Since that's the name we finally settled on.
</blockquote></details>

Done.
___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 75 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1751495295:-O79-Jwv9b179l7UC9sX:b-8g7ufp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7a219d8bc7152e9740bae138ca20a98d84bb2ee/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L75)):*

What do you think about moving the parsing logic for a slice of the byte sequence to the individual obfuscators? E.g.
```rust
impl From<&[u8; 16]> for Plain {
    fn from(data: &[u8; 16]) -> Self {
        let (v4_bytes, mut port_bytes) = data.split_first_chunk::<4>().unwrap();

        let v4_addr = Ipv4Addr::from(*v4_bytes);
        let port = port_bytes.read_u16::<LittleEndian>().unwrap();

        Plain {
            addr: SocketAddrV4::new(v4_addr, port),
        }
    }
}
```
and
```rust
impl TryFrom<&[u8; 16]> for Xor {
    type Error = Error;

    fn try_from(data: &[u8; 16]) -> Result<Self, Self::Error> {
        let (v4_bytes, mut remaining_bytes) = data.split_first_chunk::<4>().unwrap();

        let v4_addr = Ipv4Addr::from(*v4_bytes);
        let port = remaining_bytes.read_u16::<LittleEndian>().unwrap();

        let xor_key: Vec<u8> = remaining_bytes[..6]
            .iter()
            .copied()
            .filter(|&byte| byte != 0x00)
            .collect();

        if xor_key.is_empty() {
            return Err(Error::EmptyXorKey);
        }
        Ok(Xor {
            addr: SocketAddrV4::new(v4_addr, port),
            xor_key,
            key_index: 0,
        })
    }
}
```
</blockquote></details>

That wouldn't be a bad way to go about it - but I would make it `From<[u8; 12]>` since the header can be removed. The first 4 bytes are not significant for each of the specific proxy types.
___
*[`mullvad-obfuscated-dns-proxy/src/config/mod.rs` line 87 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O7-6AOb1y-MZtKK3FNF:-O73WIHIDd8roAFRQCzT:bd43vo8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d703f5085a54c6b3a38695108e3191bfd00c209c/mullvad-obfuscated-dns-proxy/src/config/mod.rs#L87)):*

This complicates things, because then the plain case must be a special one. A forwarder needs to know where to connect a TCP socket and it need something to obfuscate the bytes - and this interface provides these details.  Further, we do not know which bytes in the Ipv6 address will be used to specify the socket address, so the obfuscator itself must contain that data.
</blockquote></details>

I would be open to not calling this trait `Obfuscator` but something like `ProxyConfig`, because that is essentially what it is. But it also implements obfuscation, since it is derived from the proxy configuration.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 6 of 15 files reviewed, 2 unresolved discussions (waiting on @dlon and @Serock3)

___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1760712398:-O7DzPRD-lpADOl0_6xs:b-gxot18) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/475893aa4e1a56dd9e9ed1c593eae57aa1e35298/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L15)):*

That's fair. I still think removing the `Forwarder`, if possible, makes the code significantly more readable.
</blockquote></details>

The `Forwarder` is not removed. But lots of other stuff is changed. So please re-evaluate if you think this feedback is still relevant.

pinkisemils (commented):
> Reviewed 1 of 8 files at r3, 3 of 14 files at r5, 1 of 1 files at r7, 1 of 8 files at r9, 1 of 2 files at r12, 1 of 3 files at r13, 7 of 7 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 5 of 15 files reviewed, 2 unresolved discussions (waiting on @Serock3)

dlon (dismissed):
Nice cleanup.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-:-O7OCt88EqwcVnG2HRbA:b7oy6p4)* status: 5 of 15 files reviewed, 3 unresolved discussions (waiting on @pinkisemils and @Serock3)

___
*[`mullvad-encrypted-dns-proxy/src/config/xor.rs` line 69 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O7OEaicECW24jioIh3F:-O7OEaicECW24jioIh3G:b-im9n85) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a0a35616260c000279f4fba15b9cb8bd37714cfe/mullvad-encrypted-dns-proxy/src/config/xor.rs#L69)):*
> ```XML
>         let key_len = key_bytes
>             .iter()
>             .position(|b| *b == 0x00)
> ```

You're probably aware of this, but if not I just want to point it out. This code differs from the go implementation in that it terminates after the first null byte, whereas the reference code filters out null bytes while appending to the key any remaining bytes that are non-zero.

Serock3 (dismissed):
> Reviewed 1 of 14 files at r5, 1 of 8 files at r9, 1 of 2 files at r11, 5 of 7 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 10 of 15 files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1760712398:-O7EnXdF0Cn9H3y92RDL:b-8qat2v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/475893aa4e1a56dd9e9ed1c593eae57aa1e35298/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L15)):*

The `Forwarder` is not removed. But lots of other stuff is changed. So please re-evaluate if you think this feedback is still relevant.
</blockquote></details>

Unless I am misunderstanding some aspect of the `Forwarder`, it's still intended to be created and then immediately consumed, which I think is more succinctly represented by a function than a type. But I won't block the PR. If you want it merged fast or don't feel like spending more time on it, then go ahead. We can always change it in the future.
___
*[`mullvad-encrypted-dns-proxy/src/config/plain.rs` line 6 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O7EVBtOA-3NTDxp_7PX:-O7EVBtOA-3NTDxp_7PY:b-49up66) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a0a35616260c000279f4fba15b9cb8bd37714cfe/mullvad-encrypted-dns-proxy/src/config/plain.rs#L6)):*
> ```XML
> /// API through a different IP address.
> ///
> /// A plain configuration is represented by proxy type `ProxyType::Plain`. A plain
> ```

Great documentation!

Nit: You could mention what is expected of the function argument, e.g. "For an IPv6 address of plain proxy type, this function parses bytes 4-16 into a proxy config".

Also,  `ProxyType::Plain` could be a doc-link, e.g.



_Code snippet:_
```XML
[`super::ProxyType::Plain`]
```
___
*[`mullvad-encrypted-dns-proxy/src/config/xor.rs` line 106 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O7Eg34g4ydzckUgscCJ:-O7Eg34g4ydzckUgscCK:bz9mgq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a0a35616260c000279f4fba15b9cb8bd37714cfe/mullvad-encrypted-dns-proxy/src/config/xor.rs#L106)):*
> ```XML
>         }
>     }
> }
> ```

Here's an alternative implementation that I experimented with, you can ignore it unless you prefer it.
```rust
#[derive(Debug)]
pub struct XorObfuscator {
    key_iter: Cycle<Take<IntoIter<u8, 6>>>,
}

impl XorObfuscator {
    pub fn new(key: XorKey) -> Self {
        let key_iter = key.data.into_iter().take(key.len).cycle();
        Self { key_iter }
    }
}

impl super::Obfuscator for XorObfuscator {
    fn obfuscate(&mut self, buffer: &mut [u8]) {
        for (byte, key) in buffer.iter_mut().zip(&mut self.key_iter) {
            *byte ^= key;
        }
    }
}
```



_Code quote:_
```XML
#[derive(Debug)]
pub struct XorObfuscator {
    key: XorKey,
    key_index: usize,
}

impl XorObfuscator {
    pub fn new(key: XorKey) -> Self {
        Self { key, key_index: 0 }
    }
}

impl super::Obfuscator for XorObfuscator {
    fn obfuscate(&mut self, buffer: &mut [u8]) {
        let key_data = self.key.key_data();
        for byte in buffer {
            *byte ^= key_data[self.key_index % key_data.len()];
            self.key_index = (self.key_index + 1) % key_data.len();
        }
    }
}
```

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 10 of 15 files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs` line 15 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#gh-1760712398:-O7SoUrI9FOCOXdOg98X:b-tyhus) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/475893aa4e1a56dd9e9ed1c593eae57aa1e35298/mullvad-obfuscated-dns-proxy/src/forwarder/mod.rs#L15)):*

Unless I am misunderstanding some aspect of the `Forwarder`, it's still intended to be created and then immediately consumed, which I think is more succinctly represented by a function than a type. But I won't block the PR. If you want it merged fast or don't feel like spending more time on it, then go ahead. We can always change it in the future.
</blockquote></details>

It can be used for forwarding, but the forwarder itself also implements `AsyncRead` and `AsyncWrite`, so it can be supplied to a `TlsStream` to be used directly. 
___
*[`mullvad-encrypted-dns-proxy/src/config/xor.rs` line 69 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O7OEaicECW24jioIh3F:-O7Snq3W0EJBgqqNMBG2:b4x7h0i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a0a35616260c000279f4fba15b9cb8bd37714cfe/mullvad-encrypted-dns-proxy/src/config/xor.rs#L69)):*

You're probably aware of this, but if not I just want to point it out. This code differs from the go implementation in that it terminates after the first null byte, whereas the reference code filters out null bytes while appending to the key any remaining bytes that are non-zero.
</blockquote></details>

That was XOR V1 behavior, this is XOR V2 behavior. 

https://github.com/mullvad/apisocks5/blob/master/internal/ipv6md/addrportxorv2/addrportxorv2.go#L78-L89

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 10 of 15 files reviewed, 1 unresolved discussion (waiting on @dlon and @pinkisemils)

___
*[`mullvad-encrypted-dns-proxy/src/config/plain.rs` line 6 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768#-O7EVBtOA-3NTDxp_7PX:-O7Sqqzt1VM_lGZm2MAk:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a0a35616260c000279f4fba15b9cb8bd37714cfe/mullvad-encrypted-dns-proxy/src/config/plain.rs#L6)):*

Great documentation!

Nit: You could mention what is expected of the function argument, e.g. "For an IPv6 address of plain proxy type, this function parses bytes 4-16 into a proxy config".

Also,  `ProxyType::Plain` could be a doc-link, e.g.
</blockquote></details>

Done

pinkisemils (commented):
> Reviewed 1 of 1 files at r16.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6768)* status: 10 of 15 files reviewed, all discussions resolved

BionicBison05:
Could this have any implications for setting a DNS-over-HTTPS server as a custom DNS server in the future?

faern:
No it should not. Since we do our Encrypted DNS proxy DoH lookups to specific resolvers hardcoded in this library. What the OS setting is does not affect how it operates.

PR #6813, opened 2024-09-18
Make the packet tunnel enter error state when failing device check

buggmagnet:
This PR makes sure that when a device check error happens on each connection of the Packet Tunnel,  we do enter the error state instead of silently logging the issue and doing nothing.

When this happens, the users will be told via the in app banner that they will have to log out and log in again because their device's integrity couldn't be checked.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6813)


rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6813)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #4163, opened 2022-11-23
Add search to location list and refactor select location view

raksooo:
This PR:
* Refactors everything related to the select location view
* Adds a search bar to the select location view

The refactoring consists of:
* Moving all source files related to the view into `gui/src/renderer/components/select-location`.
* Move all sorting/filtering/modifications of relay list into `gui/src/renderer/components/select-location/RelayListContext.tsx`. This means that `gui/src/main/relay-list.ts` could be removed.
* Simplify the components preparing and rendering the list items.
* Split the `SelectLocation` component up into multiple parts.
* Move to functional components and hooks.
* Probably more

Since most of the code has been changed/rewritten I would suggest reviewing everything in `gui/src/renderer/components/select-location` as new files instead of just the changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/4163)

PR #6832, opened 2024-09-21
Autofocus location search bar when switching tabs

### LINKED ISSUE #6554: [Feature request] Selecting the `Entry` and `Exit` list of servers should focus their search input
### I have checked if others have suggested this already

- [X] I have checked this issue tracker to see if others have reported similar issues.

### Feature description

When selecting the `Entry` and `Exit` list of servers, their search input should be automatically focused.
This would provide a better user experience and a faster search.

### Alternative solutions

None.

### Type of feature

- [ ] Better privacy/anonymity
- [ ] Better at circumventing censorship
- [X] Easier to use
- [ ] Other

### Operating System

- [ ] Android
- [ ] iOS
- [X] Windows
- [X] macOS
- [X] Linux

MrChocolatine:
### What this PR changes

This PR adds an autofocus to the search bar, of the `Selection location` view, when changing tabs.

### Why this is wanted

To contribute and implement the feature requested in #6554 .
_(I know I was told the UI is going to be re-worked: https://github.com/mullvad/mullvadvpn-app/issues/6554#issuecomment-2317435263 , but I thought I could satisfy my curiosity by giving it a shot)_

### How it's implemented

> While I know React, I am not pro-efficient with it as it's not my daily tool. So I hope my implementation is okay-ish and fits your standards.

By notifying when the location type has changed (using `useBoolean()`), while passing the information to the `<SearchBar>` component.

### How to test the change

The autofocus of the search bar is currently not tested, even when it was initially added:
[`d5f216e` (#4163)](https://github.com/mullvad/mullvadvpn-app/pull/4163/commits/d5f216e2d99071958d9d635e9639971255ca61ef)

---

Fixes #6554

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6832)

### DIFF
--- a/gui/src/renderer/components/SearchBar.tsx
+++ b/gui/src/renderer/components/SearchBar.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 
 import { colors } from '../../config.json';
 import { messages } from '../../shared/gettext';
-import { useStyledRef } from '../lib/utilityHooks';
+import { useCombinedRefs, useStyledRef } from '../lib/utilityHooks';
 import { normalText } from './common-styles';
 import ImageView from './ImageView';
 
@@ -67,16 +67,16 @@ export const StyledClearIcon = styled(ImageView)({
 });
 
 interface ISearchBarProps {
+  searchInputRef?: React.Ref<HTMLInputElement>;
   searchTerm: string;
   onSearch: (searchTerm: string) => void;
   className?: string;
   disableAutoFocus?: boolean;
-  hasLocationTypeChanged?: boolean;
-  resetLocationTypeChange?: () => void;
 }
 
 export default function SearchBar(props: ISearchBarProps) {
   const inputRef = useStyledRef<HTMLInputElement>();
+  const combinedRef = useCombinedRefs(inputRef, props.searchInputRef);
 
   const onInput = useCallback(
     (event: React.FormEvent) => {
@@ -95,17 +95,15 @@ export default function SearchBar(props: ISearchBarProps) {
     if (!props.disableAutoFocus) {
       inputRef.current?.focus({ preventScroll: true });
     }
-
-    return () => props.resetLocationTypeChange?.();
-  }, [props.hasLocationTypeChanged]);
+  }, []);
 
   return (
     <StyledSearchContainer className={props.className}>
       <StyledSearchInput
-        ref={inputRef}
         value={props.searchTerm}
         onInput={onInput}
         placeholder={messages.gettext('Search for...')}
+        ref={combinedRef}
       />
       <StyledSearchIcon source="icon-search" width={24} tintColor={colors.white60} />
       {props.searchTerm.length > 0 && (
--- a/gui/src/renderer/components/select-location/SelectLocation.tsx
+++ b/gui/src/renderer/components/select-location/SelectLocation.tsx
@@ -9,7 +9,7 @@ import { daitaFilterActive, filterSpecialLocations } from '../../lib/filter-loca
 import { useHistory } from '../../lib/history';
 import { formatHtml } from '../../lib/html-formatter';
 import { RoutePath } from '../../lib/routes';
-import { useBoolean, useNormalRelaySettings } from '../../lib/utilityHooks';
+import { useNormalRelaySettings, useStyledRef } from '../../lib/utilityHooks';
 import { useSelector } from '../../redux/store';
 import * as Cell from '../cell';
 import { useFilteredProviders } from '../Filter';
@@ -102,14 +102,13 @@ export default function SelectLocation() {
     }
   }, [relaySettingsUpdater, resetScrollPositions, relaySettings]);
 
-  const [hasLocationTypeChanged, notifyLocationTypeChange, resetLocationTypeChange] =
-    useBoolean(false);
+  const searchInputRef = useStyledRef<HTMLInputElement>();
 
   const changeLocationType = useCallback(
     (locationType: LocationType) => {
+      searchInputRef.current?.focus();
       saveScrollPosition();
       setLocationType(locationType);
-      notifyLocationTypeChange();
     },
     [saveScrollPosition],
   );
@@ -243,10 +242,9 @@ export default function SelectLocation() {
               )}
 
               <StyledSearchBar
+                searchInputRef={searchInputRef}
                 searchTerm={searchValue}
                 onSearch={updateSearchTerm}
-                hasLocationTypeChanged={hasLocationTypeChanged}
-                resetLocationTypeChange={resetLocationTypeChange}
               />
             </StyledNavigationBarAttachment>
 


raksooo:
Thank you for providing us with this work! It's definitely a nice addition! I have some concerns about the code though:
1. The `SearchBar` component is used in the Split tunneling view as well and in there the `LocationType`-props doesn't make much sense. I would prefer to not pass anything that specific to that component.
2. The `useBoolean` approach seems a bit more complicated than needed, but should work fine.

I have some suggestions for you if you would like to improve it yourself, or if you want me to do it I can submit another PR for this :) Here is how I would do it:
* Forward a ref into `SearchBar` using `React.forwardRef` and combine it with the `inputRef` using `useCombinedRefs`. You can check [AutoResizingTextInputWithRef](https://github.com/mullvad/mullvadvpn-app/blob/main/gui/src/renderer/components/cell/Input.tsx#L196) for inspiration.
* Keep a ref in the `SelectLocation` component and pass it down to `SearchBar`
* In `changeLocationType` in `SelectLocation` call `.focus()` on the ref.

This would let the `SelectLocation` component call `.focus()` on the input field without exposing any internals of the select location view to the search bar. Let me know if you want me to make these changes :)

MrChocolatine:
> Thank you for providing us with this work! It's definitely a nice addition! I have some concerns about the code though:
> 
> 1. The `SearchBar` component is used in the Split tunneling view as well and in there the `LocationType`-props doesn't make much sense. I would prefer to not pass anything that specific to that component.
> 2. The `useBoolean` approach seems a bit more complicated than needed, but should work fine.
> 
> I have some suggestions for you if you would like to improve it yourself, or if you want me to do it I can submit another PR for this :) Here is how I would do it:
> 
> * Forward a ref into `SearchBar` using `React.forwardRef` and combine it with the `inputRef` using `useCombinedRefs`. You can check [AutoResizingTextInputWithRef](https://github.com/mullvad/mullvadvpn-app/blob/main/gui/src/renderer/components/cell/Input.tsx?rgh-link-date=2024-09-24T12%3A00%3A09Z#L196) for inspiration.
> * Keep a ref in the `SelectLocation` component and pass it down to `SearchBar`
> * In `changeLocationType` in `SelectLocation` call `.focus()` on the ref.
> 
> This would let the `SelectLocation` component call `.focus()` on the input field without exposing any internals of the select location view to the search bar. Let me know if you want me to make these changes :)

Thanks for the feedback @raksooo .

I will try to rework the PR following your suggestions 👍 .

MrChocolatine:
> Thank you for providing us with this work! It's definitely a nice addition! I have some concerns about the code though:
> 
> 1. The `SearchBar` component is used in the Split tunneling view as well and in there the `LocationType`-props doesn't make much sense. I would prefer to not pass anything that specific to that component.
> 2. The `useBoolean` approach seems a bit more complicated than needed, but should work fine.
> 
> I have some suggestions for you if you would like to improve it yourself, or if you want me to do it I can submit another PR for this :) Here is how I would do it:
> 
> * Forward a ref into `SearchBar` using `React.forwardRef` and combine it with the `inputRef` using `useCombinedRefs`. You can check [AutoResizingTextInputWithRef](https://github.com/mullvad/mullvadvpn-app/blob/main/gui/src/renderer/components/cell/Input.tsx?rgh-link-date=2024-09-24T12%3A00%3A09Z#L196) for inspiration.
> * Keep a ref in the `SelectLocation` component and pass it down to `SearchBar`
> * In `changeLocationType` in `SelectLocation` call `.focus()` on the ref.
> 
> This would let the `SelectLocation` component call `.focus()` on the input field without exposing any internals of the select location view to the search bar. Let me know if you want me to make these changes :)

@raksooo , changes applied following your suggestion :v: .

gui/src/renderer/components/SearchBar.tsx
```diff
@@ -71,10 +71,12 @@ interface ISearchBarProps {
   onSearch: (searchTerm: string) => void;
   className?: string;
   disableAutoFocus?: boolean;
+  searchInputRef?: React.Ref<HTMLInputElement>;
 }
 
 export default function SearchBar(props: ISearchBarProps) {
```
MrChocolatine, line 77: In your [previous suggestion](https://github.com/mullvad/mullvadvpn-app/pull/6832#issuecomment-2371063363) @raksooo , you showed me `AutoSizingTextInputWithRef` as an example:

https://github.com/mullvad/mullvadvpn-app/blob/d35866d337647a26dc8ffc2e875212306dd7c0ce/gui/src/renderer/components/cell/Input.tsx#L196

This components consumes a `forwardedRef` parameter, the `ref` object does not come from its `props`.
But here I was only able to pass it via a prop.

gui/src/renderer/components/SearchBar.tsx
```diff
@@ -71,10 +71,12 @@ interface ISearchBarProps {
   onSearch: (searchTerm: string) => void;
   className?: string;
   disableAutoFocus?: boolean;
+  searchInputRef?: React.Ref<HTMLInputElement>;
```
MrChocolatine, line 74: In your [previous suggestion](https://github.com/mullvad/mullvadvpn-app/pull/6832#issuecomment-2371063363) @raksooo , you showed me `AutoSizingTextInputWithRef` as an example:

https://github.com/mullvad/mullvadvpn-app/blob/d35866d337647a26dc8ffc2e875212306dd7c0ce/gui/src/renderer/components/cell/Input.tsx#L196

This components consumes a `forwardedRef` parameter, the `ref` object does not come from its `props`.
But here I was only able to pass it via a prop.

gui/src/renderer/components/SearchBar.tsx
```diff
@@ -67,6 +67,7 @@ export const StyledClearIcon = styled(ImageView)({
 });
 
 interface ISearchBarProps {
+  searchInputRef?: React.Ref<HTMLInputElement>;
```
MrChocolatine, line 70: In your [previous suggestion](https://github.com/mullvad/mullvadvpn-app/pull/6832#issuecomment-2371063363) @raksooo , you showed me `AutoSizingTextInputWithRef` as an example:

https://github.com/mullvad/mullvadvpn-app/blob/d35866d337647a26dc8ffc2e875212306dd7c0ce/gui/src/renderer/components/cell/Input.tsx#L196

This components consumes a `forwardedRef` parameter, the `ref` object does not come from its `props`.
But here I was only able to pass it via a prop.

MrChocolatine:
Strangely tests are failing:

https://github.com/mullvad/mullvadvpn-app/actions/runs/11314889729/job/31465392999#step:11:57
image

while I cannot reproduce locally:

image

gui/src/renderer/components/SearchBar.tsx
raksooo: To use the `ref`-prop you will need to wrap the component in `React.forwardRef`.
```
function SearchBar(props: ISearchBarProps, forwardedRef: React.Ref<HTMLInputElement>) { ... }

export default React.forwardRef(SearchBar);
```

Then you won't have to include it in props and you'll be able to use `SearchBar` like this:
```
<SearchBar ref={searchInputRef} ... />
```

I think `React.forwardRef` won't be needed in future versions of react, but for now we'll need to add it :)

raksooo:
The failing test should be fixed in this PR: https://github.com/mullvad/mullvadvpn-app/pull/6970 :)

MarkusPettersson98:
Hi @MrChocolatine, sorry for the long delay since the last response. We are currently implementing a new design for the select location view, which means that your PR will most likely be very out of date.

PR #6840, opened 2024-09-23
Add fallback when failing to create `wintun` adapter

Serock3:
This is an attempt to fix the elusive "Failed to create Wintun adapter" bug. Our workaround is to open an existing adapter with name "Mullvad" as a fallback. We cannot reproduce the bug, so our only option is to release this patch and investigate if the bug-reports for the issue stop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6840)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6840)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6842, opened 2024-09-23
Fix out of time use case virtual time tests

kl:
The awaitItem() method from Turbine is used which causes delays to be skipped. This causes some tests to behave incorrectly where we want to assert that a given value is emitted after a specific delay. The solution is to instead use the expectMostRecentItem() method when we want to assert that an item has been emitted at a given point in time.

Further reading:

https://medium.com/@hi.enricog/explore-testing-coroutines-with-runtest-by-examples-4c4274a86a83

https://github.com/cashapp/turbine/issues/265

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6842)


Pururun (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6842#-:-O7Y-6UE7fGIhoo6sS_h:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6846, opened 2024-09-23
Use encrypted dns proxy in `mullvad-ios`

MarkusPettersson98:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)

### DIFF
--- a/mullvad-encrypted-dns-proxy/src/config_resolver.rs
+++ b/mullvad-encrypted-dns-proxy/src/config_resolver.rs
@@ -4,10 +4,12 @@ use crate::config;
 use core::fmt;
 use hickory_resolver::{config::*, error::ResolveError, TokioAsyncResolver};
 use rustls::ClientConfig;
-use std::{net::IpAddr, sync::Arc};
+use std::{net::IpAddr, sync::Arc, time::Duration};
+use tokio::time::error::Elapsed;
 
 /// The port to connect to the DoH resolvers over.
 const RESOLVER_PORT: u16 = 443;
+const DEFAULT_TIMEOUT: Duration = std::time::Duration::from_secs(10);
 
 pub struct Nameserver {
     pub name: String,
@@ -15,17 +17,26 @@ pub struct Nameserver {
 }
 
 #[derive(Debug)]
-pub struct ResolutionError(ResolveError);
+pub enum Error {
+    ResolutionError(ResolveError),
+    Timeout(Elapsed),
+}
 
-impl fmt::Display for ResolutionError {
+impl fmt::Display for Error {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        self.0.fmt(f)
+        match self {
+            Error::ResolutionError(err) => err.fmt(f),
+            Error::Timeout(err) => err.fmt(f),
+        }
     }
 }
 
-impl std::error::Error for ResolutionError {
+impl std::error::Error for Error {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-        self.0.source()
+        match self {
+            Self::ResolutionError(ref err) => Some(err),
+            Self::Timeout(ref err) => Some(err),
+        }
     }
 }
 
@@ -50,13 +61,17 @@ pub fn default_resolvers() -> Vec<Nameserver> {
     ]
 }
 
+pub async fn resolve_default_config() -> Result<Vec<config::ProxyConfig>, Error> {
+    resolve_configs(&default_resolvers(), "frakta.eu").await
+}
+
 /// Look up the `domain` towards the given `resolvers`, and try to deserialize all the returned
 /// AAAA records into [`ProxyConfig`](config::ProxyConfig)s.
 pub async fn resolve_configs(
     resolvers: &[Nameserver],
     domain: &str,
-) -> Result<Vec<config::ProxyConfig>, ResolutionError> {
-    let mut resolver_config = ResolverConfig::new();
+) -> Result<Vec<config::ProxyConfig>, Error> {
+    let mut nameservers = ResolverConfig::new();
     for resolver in resolvers.iter() {
         let ns_config_group = NameServerConfigGroup::from_ips_https(
             &resolver.addr,
@@ -66,25 +81,28 @@ pub async fn resolve_configs(
         )
         .into_inner();
         for ns_config in ns_config_group {
-            resolver_config.add_name_server(ns_config);
+            nameservers.add_name_server(ns_config);
         }
     }
 
-    resolver_config.set_tls_client_config(Arc::new(client_config_tls12()));
+    nameservers.set_tls_client_config(Arc::new(client_config_tls12()));
+    let mut resolver_config: ResolverOpts = Default::default();
 
-    resolve_config_with_resolverconfig(resolver_config, Default::default(), domain).await
+    resolver_config.timeout = Duration::from_secs(5);
+    resolve_config_with_resolverconfig(nameservers, resolver_config, domain, DEFAULT_TIMEOUT).await
 }
 
 pub async fn resolve_config_with_resolverconfig(
     resolver_config: ResolverConfig,
     options: ResolverOpts,
     domain: &str,
-) -> Result<Vec<config::ProxyConfig>, ResolutionError> {
+    timeout: Duration,
+) -> Result<Vec<config::ProxyConfig>, Error> {
     let resolver = TokioAsyncResolver::tokio(resolver_config, options);
-    let lookup = resolver
-        .ipv6_lookup(domain)
+    let lookup = tokio::time::timeout(timeout, resolver.ipv6_lookup(domain))
         .await
-        .map_err(ResolutionError)?;
+        .map_err(Error::Timeout)?
+        .map_err(Error::ResolutionError)?;
 
     let addrs = lookup.into_iter().map(|aaaa_record| aaaa_record.0);
 
--- a/mullvad-ios/src/encrypted_dns_proxy.rs
+++ b/mullvad-ios/src/encrypted_dns_proxy.rs
@@ -1,0 +1,221 @@
+use crate::ProxyHandle;
+
+use mullvad_encrypted_dns_proxy::{config::ProxyConfig, config_resolver, Forwarder};
+use std::{
+    collections::HashSet,
+    io, mem,
+    net::{Ipv4Addr, SocketAddr},
+    ptr,
+};
+use tokio::{net::TcpListener, task::JoinHandle};
+
+pub struct EncryptedDnsProxyState {
+    /// Note that we rely on the randomness of the ordering of the items in the hashset to pick a
+    /// random configurations every time.
+    configurations: HashSet<ProxyConfig>,
+    tried_configurations: HashSet<ProxyConfig>,
+}
+
+#[derive(Debug)]
+pub enum Error {
+    /// Failed to initialize tokio runtime.
+    TokioRuntime,
+    /// Failed to bind a local listening socket, the one that will be forwarded through the proxy.
+    BindLocalSocket(io::Error),
+    /// Failed to get local listening address of the local listening socket.
+    GetBindAddr(io::Error),
+    /// Failed to initialize forwarder.
+    Forwarder(io::Error),
+    /// Failed to fetch a proxy configuration over DNS.
+    FetchConfig(config_resolver::Error),
+    /// Failed to initialize with a valid configuration.
+    NoConfigs,
+}
+
+impl From<Error> for i32 {
+    fn from(err: Error) -> Self {
+        match err {
+            Error::TokioRuntime => -1,
+            Error::BindLocalSocket(_) => -2,
+            Error::GetBindAddr(_) => -3,
+            Error::Forwarder(_) => -4,
+            Error::FetchConfig(_) => -5,
+            Error::NoConfigs => -6,
+        }
+    }
+}
+
+impl EncryptedDnsProxyState {
+    async fn start(&mut self) -> Result<ProxyHandle, Error> {
+        self.fetch_configs().await?;
+        let proxy_configuration = self.next_configuration().ok_or(Error::NoConfigs)?;
+
+        let local_socket = Self::bind_local_addr()
+            .await
+            .map_err(Error::BindLocalSocket)?;
+        let bind_addr = local_socket.local_addr().map_err(Error::GetBindAddr)?;
+        let forwarder = Forwarder::connect(&proxy_configuration)
+            .await
+            .map_err(Error::Forwarder)?;
+        let join_handle = Box::new(tokio::spawn(async move {
+            if let Ok((client_conn, _)) = local_socket.accept().await {
+                let _ = forwarder.forward(client_conn).await;
+            }
+        }));
+
+        Ok(ProxyHandle {
+            context: Box::into_raw(join_handle).cast(),
+            port: bind_addr.port(),
+        })
+    }
+
+    async fn bind_local_addr() -> io::Result<TcpListener> {
+        let bind_addr = SocketAddr::new(Ipv4Addr::UNSPECIFIED.into(), 0);
+        TcpListener::bind(bind_addr).await
+    }
+
+    /// Select a config.
+    /// Always select an obfuscated configuration, if there are any left untried. If no obfuscated
+    /// configurations exist, try plain configurations. The order is randomized due to the hash set
+    /// storing the configurations in a random order.
+    fn next_configuration(&mut self) -> Option<ProxyConfig> {
+        if self.should_reset() {
+            self.reset();
+        }
+        // TODO: currently, the randomized order of proxy config retrieval depends on the random
+        // iteration order of a given HashSet instance. Since for now, there will be only 2
+        // different configurations, it barely matters. In the future, we should use `rand`
+        // instead, so that the behavior is explicit and clear.
+
+        // First, try getting an obfuscated configuration, if there exist any.
+        let config = if let Some(obfuscated_config) = self
+            .configurations
+            .difference(&self.tried_configurations)
+            .find(|config| config.obfuscation.is_some())
+            .cloned()
+        {
+            obfuscated_config
+        } else {
+            // If no obfuscated configurations exist, try the next untried configuration.
+            self.configurations
+                .difference(&self.tried_configurations)
+                .next()
+                .cloned()?
+        };
+
+        self.tried_configurations.insert(config.clone());
+        Some(config)
+    }
+
+    /// Fetch a config, but error out only when no existing configuration was there.
+    async fn fetch_configs(&mut self) -> Result<(), Error> {
+        match mullvad_encrypted_dns_proxy::config_resolver::resolve_default_config().await {
+            Ok(new_configs) => {
+                self.configurations = HashSet::from_iter(new_configs.into_iter());
+            }
+            Err(err) => {
+                log::error!("Failed to fetch a new proxy configuration: {err:?}");
+                if self.is_empty() {
+                    return Err(Error::FetchConfig(err));
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn is_empty(&self) -> bool {
+        self.configurations.is_empty()
+    }
+
+    /// Checks if the `tried_configurations` set should be reset.
+    /// It should only be reset if the difference between `configurations` and
+    /// `tried_configurations` is an empty set - in this case all available configurations have
+    /// been tried.
+    fn should_reset(&self) -> bool {
+        self.configurations
+            .difference(&self.tried_configurations)
+            .count()
+            == 0
+    }
+
+    /// Clears the `tried_configurations` set.
+    fn reset(&mut self) {
+        self.tried_configurations.clear();
+    }
+}
+
+/// Initializes a valid pointer to an instance of `EncryptedDnsProxyState`.
+#[no_mangle]
+pub unsafe extern "C" fn encrypted_dns_proxy_init() -> *mut EncryptedDnsProxyState {
+    let state = Box::new(EncryptedDnsProxyState {
+        configurations: Default::default(),
+        tried_configurations: Default::default(),
+    });
+    Box::into_raw(state)
+}
+
+/// This must be called only once to deallocate `EncryptedDnsProxyState`.
+///
+/// # Safety
+/// `ptr` must be a valid, exclusive pointer to `EncryptedDnsProxyState`, initialized
+/// by `encrypted_dns_proxy_init`. This function is not thread safe, and should only be called
+/// once.
+#[no_mangle]
+pub unsafe extern "C" fn encrypted_dns_proxy_free(ptr: *mut EncryptedDnsProxyState) {
+    let _ = unsafe { Box::from_raw(ptr) };
+}
+
+/// # Safety
+/// encrypted_dns_proxy must be a valid, exclusive pointer to `EncryptedDnsProxyState`, initialized
+/// by `encrypted_dns_proxy_init`. This function is not thread safe.
+/// `proxy_handle` must be pointing to a valid memory region for the size of a `ProxyHandle`. This
+/// function is not thread safe, but it can be called repeatedly. Each successful invocation should
+/// clean up the resulting proxy via `[encrypted_dns_proxy_stop]`.
+///
+/// `proxy_handle` will only contain valid values if the return value is zero. It is still valid to
+/// deallocate the memory.
+#[no_mangle]
+pub unsafe extern "C" fn encrypted_dns_proxy_start(
+    encrypted_dns_proxy: *mut EncryptedDnsProxyState,
+    proxy_handle: *mut ProxyHandle,
+) -> i32 {
+    let handle = match crate::mullvad_ios_runtime() {
+        Ok(handle) => handle,
+        Err(err) => {
+            log::error!("Cannot instantiate a tokio runtime: {}", err);
+            return Error::TokioRuntime.into();
+        }
+    };
+
+    let mut encrypted_dns_proxy = unsafe { Box::from_raw(encrypted_dns_proxy) };
+    let proxy_result = handle.block_on(encrypted_dns_proxy.start());
+    mem::forget(encrypted_dns_proxy);
+
+    match proxy_result {
+        Ok(handle) => unsafe { ptr::write(proxy_handle, handle) },
+        Err(err) => {
+            let empty_handle = ProxyHandle {
+                context: ptr::null_mut(),
+                port: 0,
+            };
+            unsafe { ptr::write(proxy_handle, empty_handle) }
+            log::error!("Failed to create a proxy connection: {err:?}");
+            return err.into();
+        }
+    }
+
+    0
+}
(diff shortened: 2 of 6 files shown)


pinkisemils:
We believe this is finally ready for review. With these changes, I am introducing an FFI to be used by the iOS codebase. It also adds a kind of a retry strategy. This was a joint effort between @MarkusPettersson98 and me. 

Whilst in the weeds, we found that the timeout in the resolver config for `hickory-resolver` isn't actually taken into account - we should probably report it as an error to them.

faern (commented):
As discussed in the office, please see this document for the latest spec. I believe the only difference from the code you propose here, is how we pick XOR over plain: https://mullvad.atlassian.net/wiki/spaces/CLIEN/pages/3736502288/Encrypted+DNS+proxy

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-:-O7ZOnh86EZAa7x_WAya:bs5wnty)* status: 0 of 9 files reviewed, all discussions resolved

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`mullvad-encrypted-dns-proxy/src/config/mod.rs` line 79 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7ZQPYp46QL2ZfSGo6l:-O7ZQPYp46QL2ZfSGo6m:b-18zzca) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a155c03e5dabe2da938a07ba80159a937f974e78/mullvad-encrypted-dns-proxy/src/config/mod.rs#L79)):*
> ```XML
>     pub obfuscation: Option<ObfuscationConfig>,
> 
>     pub r#type: ProxyType,
> ```

As discussed in the office: Remove this field and instead derive proxy type (to prefer XOR over plain) via the `obfuscation` field above instead.

pinkisemils (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @dlon and @MarkusPettersson98)

___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 154 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7Z_9RE4iyfBIX0k8JI:-O7_8wgv8ZMmnQotmzVA:b-xdxsd1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6b23ab188667940b462e54fd3e0177ef7d843ea8/mullvad-ios/src/encrypted_dns_proxy.rs#L154)):*

Should `forget` be called here?
</blockquote></details>

No, we should not. This was most certainly not my best judgement call.

MarkusPettersson98 (commented):
> Reviewed 5 of 9 files at r1, 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)* status: 0 of 9 files reviewed, 6 unresolved discussions (waiting on @dlon and @faern)

___
*[`mullvad-encrypted-dns-proxy/src/config/mod.rs` line 79 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7ZQPYp46QL2ZfSGo6l:-O7bpu5wC0YNitU7z7MS:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a155c03e5dabe2da938a07ba80159a937f974e78/mullvad-encrypted-dns-proxy/src/config/mod.rs#L79)):*

As discussed in the office: Remove this field and instead derive proxy type (to prefer XOR over plain) via the `obfuscation` field above instead.
</blockquote></details>

Done
___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 113 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c1jLCAUTbaQDHVMt3:-O7c1jLCAUTbaQDHVMt4:b-9ppqre) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L113)):*
> ```XML
>         self.tried_configurations.insert(config.clone());
>         Some(config)
>     }
> ```

Nit What do you think about appying some DRY? Nice to not need the explicit early return in the middle of the function imho.
```rust
    fn next_configuration(&mut self) -> Option<ProxyConfig> {
        let mut difference = self.configurations.difference(&self.tried_configurations);
        let config = difference
            // prefer obfuscated proxy configurations.
            .find(|config| config.obfuscation.is_some())
            .or_else(|| difference.next())
            .cloned()?;
        self.tried_configurations.insert(config.clone());
        Some(config)
    }
```



_Code quote:_
```XML
    fn next_configuration(&mut self) -> Option<ProxyConfig> {
        if let Some(xor_config) = self
            .configurations
            .difference(&self.tried_configurations)
            // prefer obfuscated proxy configurations.
            .find(|config| config.obfuscation.is_some())
            .cloned()
        {
            self.tried_configurations.insert(xor_config.clone());
            return Some(xor_config);
        }

        let config = self
            .configurations
            .difference(&self.tried_configurations)
            .next()
            .cloned()?;
        self.tried_configurations.insert(config.clone());
        Some(config)
    }
```
___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 159 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7bvcwL0cK1uoz1Yp9g:-O7bvcwL0cK1uoz1Yp9h:bha9b3e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L159)):*
> ```XML
> /// `proxy_handle` will only contain valid values if the return value is zero. It is still valid to
> /// deallocate the memory.
> ///
> ```

Nit: Trailing blank line



_Code quote:_
```XML
/// `proxy_handle` will only contain valid values if the return value is zero. It is still valid to
/// deallocate the memory.
///
```
___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 170 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7bsF9k26aKeRWvKUld:-O7bsF9l44mZ8w2wIQS9:bk0ufsd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L170)):*
> ```XML
>             log::error!("Cannot instantiate a tokio runtime: {}", err);
>             return -1;
>         }
> ```

Do we want to promote `-1` error code to a variant of the `Error` enum defined in this module? Currently, just looking at the conversion from `Error` to an error code / `i32` implies that there is a variant missing
```rust
impl From<Error> for i32 {
    fn from(err: Error) -> Self {
        match err {
            Error::BindLocalSocket(_) => -2,
            Error::GetBindAddr(_) => -3,
            Error::Forwarder(_) => -4,
            Error::FetchConfig(_) => -5,
            Error::NoConfigs => -6,
        }
    }
}
```
Let's say we add `Error::Runtime`, we could get rid of the 'magic' error value from the `encrypted_dns_proxy_start` function
```rust
#[derive(Debug)]
pub enum Error {
    /// Failed to initialize the tokio runtime.
    Runtime,
    /// Failed to bind a local listening socket, the one that will be forwarded through the proxy
    BindLocalSocket(io::Error),
    /// Failed to get local listening address of the local listening socket.
    GetBindAddr(io::Error),
    /// Failed to initialize forwarder.
    Forwarder(io::Error),
    /// Failed to fetch new.
    FetchConfig(config_resolver::Error),
    /// Failed to initialize with a valid configuration.
    NoConfigs,
}

impl From<Error> for i32 {
    fn from(err: Error) -> Self {
        match err {
            Error::Runtime => -1,
            Error::BindLocalSocket(_) => -2,
            Error::GetBindAddr(_) => -3,
            Error::Forwarder(_) => -4,
            Error::FetchConfig(_) => -5,
            Error::NoConfigs => -6,
        }
    }
}

...

pub unsafe extern "C" fn encrypted_dns_proxy_start(
    encrypted_dns_proxy: *mut EncryptedDnsProxyState,
    proxy_handle: *mut ProxyHandle,
) -> i32 {
    let handle = match crate::mullvad_ios_runtime() {
        Ok(handle) => handle,
        Err(err) => {
            log::error!("Cannot instantiate a tokio runtime: {}", err);
            return Error::Runtime.into();
        }
    };
...
```



_Code quote:_
```XML
        Err(err) => {
            log::error!("Cannot instantiate a tokio runtime: {}", err);
            return -1;
        }
```
___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 188 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7bvSDt5fuXW5Edpjlv:-O7bvSDt5fuXW5Edpjlw:bgxskbo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L188)):*
> ```XML
> }
> 
> /// SAFETY:
> ```

Nit:
```
/// # Safety
```



_Code quote:_
```XML
/// SAFETY:
```

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)* status: 0 of 9 files reviewed, 6 unresolved discussions (waiting on @dlon and @MarkusPettersson98)

___
*[`mullvad-encrypted-dns-proxy/src/config/mod.rs` line 40 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c7G0i4qgArE0fGHaF:-O7c7G0jDiUaRyHXeD4I:by5z2w0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-encrypted-dns-proxy/src/config/mod.rs#L40)):*
> ```XML
> /// Type of a proxy configuration. Derived from the 2nd hextet of an IPv6 address in network byte
> /// order. E.g. an IPv6 address such as `7f7f:2323::`  would have a proxy type value of `0x2323`.
> #[derive(PartialEq, Debug, Eq, Hash, Clone)]
> ```

These extra impls can be removed now, when we decided to not expose this type publicly. The crate does not need those impls internally.

faern (commented):
> Reviewed 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)* status: 2 of 9 files reviewed, 6 unresolved discussions (waiting on @dlon and @MarkusPettersson98)

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)* status: 2 of 9 files reviewed, 7 unresolved discussions (waiting on @dlon and @MarkusPettersson98)

___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 113 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c1jLCAUTbaQDHVMt3:-O7c8vCfEuSU9cxf6bye:bnr9ev4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L113)):*

Nit What do you think about appying some DRY? Nice to not need the explicit early return in the middle of the function imho.
```rust
    fn next_configuration(&mut self) -> Option<ProxyConfig> {
        let mut difference = self.configurations.difference(&self.tried_configurations);
        let config = difference
            // prefer obfuscated proxy configurations.
            .find(|config| config.obfuscation.is_some())
            .or_else(|| difference.next())
            .cloned()?;
        self.tried_configurations.insert(config.clone());
        Some(config)
    }
```
</blockquote></details>

Computing the difference fewer times would indeed be nice. Not because of the computational cost, but because it's currently done three times and it's a lot of code and the difference is never given a name. If it's given a name it can be easier to read, and less code as well.

If the insertion into `tried_configurations` can be in a single place that would also be very nice.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)* status: 2 of 9 files reviewed, 8 unresolved discussions (waiting on @dlon and @MarkusPettersson98)

___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 65 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c9O0j1rm2nabZna5C:-O7c9O0j1rm2nabZna5D:bp2lg08) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L65)):*
> ```XML
>         self.fetch_configs().await?;
>         if self.should_reset() {
>             self.reset();
> ```

IMO, the reset is part of the *selection algorithm* which is "owned" by the `next_configuration` method. I feel like it would be more logical if the reset happens there.

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)* status: 2 of 9 files reviewed, 8 unresolved discussions (waiting on @dlon and @faern)

___
*[`mullvad-encrypted-dns-proxy/src/config/mod.rs` line 40 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c7G0i4qgArE0fGHaF:-O7cCTrwB5M87BDTJNm1:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-encrypted-dns-proxy/src/config/mod.rs#L40)):*

These extra impls can be removed now, when we decided to not expose this type publicly. The crate does not need those impls internally.
</blockquote></details>

Done.
___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 16 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c7hR96KorM9jM_yH_:-O7cAzTV-AIMskLFS7x5:b-oic1jk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L16)):*

Emils and I have touched on this topic in the office. I just want to point out that we want the *exact same* proxy selection algorithm on all other platforms. So if you decide to merge this into `mullvad-ios` now it has to be moved later. We are not going to duplicate the entire selection algorithm.
</blockquote></details>

Yes, that makes sense. I also think it makes sense to have it in `mullvad-ios` for now and move it later.

buggmagnet (commented):
> Reviewed 2 of 9 files at r1, 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)* status: 4 of 9 files reviewed, 12 unresolved discussions (waiting on @dlon, @faern, and @MarkusPettersson98)

___
*[`mullvad-encrypted-dns-proxy/src/config_resolver.rs` line 65 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c7i-6Cb2CTx3YmDjK:-O7c7i-6Cb2CTx3YmDjL:b-jdgad7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-encrypted-dns-proxy/src/config_resolver.rs#L65)):*
> ```XML
> 
> pub async fn resolve_default_config() -> Result<Vec<config::ProxyConfig>, Error> {
>     resolve_configs(&default_resolvers(), "frakta.eu").await
> ```

nit:
If we ever want to this be usable in staging  (or other) environments, we should either let the call site (via the FFI) pass the domain to resolve, or use a configuration flag to modify this.

But I'm okay with this for a v1, just food for thought here.
___
*[`mullvad-ios/Cargo.toml` line 13 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c98bZAjaxMZ331B8w:-O7c98bZAjaxMZ331B8x:b88os5f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/Cargo.toml#L13)):*
> ```TOML
> workspace = true
> 
> [target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
> ```

nit:
This shouldn't be necessary right ?
___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 27 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7cAEkg-gZV1BE2gA-P:-O7cAEkg-gZV1BE2gA-Q:b-e7l2ag) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L27)):*
> ```XML
>     /// Failed to initialize forwarder.
>     Forwarder(io::Error),
>     /// Failed to fetch new.
> ```

Failed to fetch a new configuration I imagine ?
___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 77 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7cBBfu0HDZkci7eXYN:-O7cBBfu0HDZkci7eXYO:bql7e4c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L77)):*
> ```XML
>             .await
>             .map_err(Error::Forwarder)?;
>         let join_handle = Box::new(tokio::spawn(async move {
> ```

Will this be created on the shared tokio runtime that was gotten in `encrypted_dns_proxy_start` ?

buggmagnet (commented):
It would be nice to separate the FFI from the implementation from a design point of view, especially if we want to have this shared with other clients. They do not need for example, the shared rust runtime from iOS

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-:-O7cD_mV6ZZAFOfXeaJs:b18jaz5)* status: 4 of 9 files reviewed, 12 unresolved discussions (waiting on @dlon, @faern, and @MarkusPettersson98)

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)* status: 4 of 9 files reviewed, 12 unresolved discussions (waiting on @buggmagnet, @dlon, and @faern)

___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 113 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c1jLCAUTbaQDHVMt3:-O7cCyATBy9n_fCfrJVI:bi12odf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L113)):*

Computing the difference fewer times would indeed be nice. Not because of the computational cost, but because it's currently done three times and it's a lot of code and the difference is never given a name. If it's given a name it can be easier to read, and less code as well.

If the insertion into `tried_configurations` can be in a single place that would also be very nice.
</blockquote></details>

As Emils pointed out to me, `find` may exhaust the iterator if there are no obfuscator configs, thus calling `next` on the iterator at that point will yield `None`. Here is an example to prove that:
```rust
#[test]
fn test_differences() {
    use std::collections::HashSet;
    let mut set = HashSet::new();
    set.insert(1);
    set.insert(2);
    let mut other_set = HashSet::new();
    other_set.insert(1);
    let mut difference = set.difference(&other_set);

    difference.find(|x| **x == 3);
    assert_eq!(difference.next(), Some(2).as_ref())
}
```
It is solvable by recomputing the difference in the closure passed to `or_else`. But we will still potentially recompute it as we do in the current implementation, FYI

MarkusPettersson98 (commented):
We will figure this out when we implement this for other clients. For now, we've decided to keep it as is for simplicity.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-:-O7cGYTx4QSN0BvDLj2w:bz0xiob)* status: 3 of 9 files reviewed, 8 unresolved discussions (waiting on @buggmagnet and @faern)

___
*[`mullvad-encrypted-dns-proxy/src/config_resolver.rs` line 65 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c7i-6Cb2CTx3YmDjK:-O7cFlIn1F02mpkGu54n:b-g7t2fd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-encrypted-dns-proxy/src/config_resolver.rs#L65)):*

nit:
If we ever want to this be usable in staging  (or other) environments, we should either let the call site (via the FFI) pass the domain to resolve, or use a configuration flag to modify this.

But I'm okay with this for a v1, just food for thought here.
</blockquote></details>

That's an interesting idea, but out of scope for this iteration :)
___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 27 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7cAEkg-gZV1BE2gA-P:-O7cFGlt9lK7RloWmr0W:bb77uf3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L27)):*

Failed to fetch a new configuration I imagine ?
</blockquote></details>

Yes!
___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 65 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c9O0j1rm2nabZna5C:-O7cFu2BA7lpjrxBK51G:bygjma1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L65)):*

IMO, the reset is part of the *selection algorithm* which is "owned" by the `next_configuration` method. I feel like it would be more logical if the reset happens there.
</blockquote></details>

Correct, implemented
___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 77 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7cBBfu0HDZkci7eXYN:-O7cFMt80X9mzWXQHAYY:b2nhswy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/src/encrypted_dns_proxy.rs#L77)):*

Will this be created on the shared tokio runtime that was gotten in `encrypted_dns_proxy_start` ?
</blockquote></details>

Yes
___
*[`mullvad-ios/Cargo.toml` line 13 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7c98bZAjaxMZ331B8w:-O7cFdc24TW21utbmtw6:bh95ke4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fbe4c0dffc570f7e4d7875703a573b2d9694b389/mullvad-ios/Cargo.toml#L13)):*

nit:
This shouldn't be necessary right ?
</blockquote></details>

You're right

buggmagnet (commented):
Sounds good to me @MarkusPettersson98  :thumbsup: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-:-O7cNfKl3Ue3w1KzmOE9:b-qcjj2c)* status: 3 of 9 files reviewed, 5 unresolved discussions (waiting on @faern and @MarkusPettersson98)

buggmagnet (commented):
> Reviewed 1 of 1 files at r4, 1 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846)* status: 5 of 9 files reviewed, 5 unresolved discussions (waiting on @faern and @MarkusPettersson98)

___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 195 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-O7cyloJ42O_MDsAgo69:-O7cyloJ42O_MDsAgo6A:b-cfnda) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/196e4a14f830d08a49146f987d205a5a8670a8a2/mullvad-ios/src/encrypted_dns_proxy.rs#L195)):*
> ```XML
> 
> /// # Safety
> /// `proxy_config` must be a valid pointer to a `ProxyHandle` as initialized by
> ```

We should mention thread safety here too

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6846#-:-O7d4MSS7j9AyHgMPSSv:bnfp4nl)* status: 5 of 9 files reviewed, 5 unresolved discussions (waiting on @faern and @MarkusPettersson98)

mullvad-ios/src/encrypted_dns_proxy.rs
```diff
@@ -0,0 +1,217 @@
...
+
+        self.tried_configurations.insert(config.clone());
+        Some(config)
+    }
+
+    /// Fetch a config, but error out only when no existing configuration was there.
+    async fn fetch_configs(&mut self) -> Result<(), Error> {
+        match mullvad_encrypted_dns_proxy::config_resolver::resolve_default_config().await {
+            Ok(new_configs) => {
+                self.configurations = HashSet::from_iter(new_configs.into_iter());
```
faern, line 110: This part is important for the random order promises we make. If `self.configurations` were not re-created but rather the same `HashSet` was cleared and re-filled, then we would be stuck in trying proxies in the same order every time. Maybe this line should document that it's important for that reason?

mullvad-encrypted-dns-proxy/src/config_resolver.rs
```diff
@@ -4,28 +4,39 @@ use crate::config;
...
     }
 }
 
-impl std::error::Error for ResolutionError {
+impl std::error::Error for Error {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-        self.0.source()
+        match self {
+            Self::ResolutionError(ref err) => Some(err),
+            Self::Timeout(ref err) => Some(err),
```
faern, line 38: We can't both return the inner error as our source *and* use the inner errors `Display` implementation up in our `Display` implementation. That makes the error stack have two entries that are identical. It will end up like this:

```
Error: Timeout reached
Caused by: Timeout reached
```

If you use an inner error as your own message in `Display` then in your implementation of `Error::source` you need to return `inner_error.source()` as your source. To cut out this repetition.

PR #6873, opened 2024-09-26
Move pick_random_relay_excluding to helpers module for consistency

faern:
All other helper methods for selecting random relays are there. So I thought it would be easier to read if we have them all in one place. Renaming the function to align more consistently with the others at the same time.

I also added tiny amount of docs. And "fixed" a lifetime nitpick.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6873)

### DIFF
--- a/mullvad-relay-selector/src/relay_selector/helpers.rs
+++ b/mullvad-relay-selector/src/relay_selector/helpers.rs
@@ -29,6 +29,17 @@ pub enum Error {
     NoMatchingPort,
 }
 
+/// Picks a relay at random from `relays`, but don't pick `exclude`.
+pub fn pick_random_relay_excluding<'a>(
+    relays: &'a [Relay],
+    exclude: &'_ Relay,
+) -> Option<&'a Relay> {
+    relays
+        .iter()
+        .filter(|&a| a != exclude)
+        .choose(&mut thread_rng())
+}
+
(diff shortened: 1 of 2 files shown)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6873)* status: 0 of 2 files reviewed, all discussions resolved

___
*[`mullvad-relay-selector/src/relay_selector/helpers.rs` line 35 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6873#-O7hlwSLCntJ80ljQ6u7:-O7hlwSLCntJ80ljQ6u8:b6xkpc0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/cf0278b898b3e538382dfb75239e6dbe655a35b5/mullvad-relay-selector/src/relay_selector/helpers.rs#L35)):*
> ```XML
> pub fn pick_random_relay_excluding<'a>(
>     relays: &'a [Relay],
>     exclude: &'_ Relay,
> ```

Nit, FYI: The relay used for exclusion is not part of the return value. So the `exclude` reference only has to be valid for the lifetime of this function, not for as long as the return value is alive.

hulthe (approved):

PR #6893, opened 2024-10-01
Fix bug when overriding API with only MULLVAD_API_HOST

faern:
I'm trying to run the daemon against stagemole. I found a bug! :partying_face: 

When the override is done with only `MULLVAD_API_HOST` it should look up the IP via DNS. This part works but, it uses the wrong hostname. In this code path we never update `api.host`, resulting it in falling back to the default (`api.mullvad.net`)

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6893)

### DIFF
--- a/mullvad-api/src/lib.rs
+++ b/mullvad-api/src/lib.rs
@@ -199,6 +199,7 @@ impl ApiEndpoint {
                         )
                     })
                     .next();
+                api.host = Some(host);
             }
             (host, Some(address)) => {
                 let addr = address.parse().unwrap_or_else(|_| {


MarkusPettersson98 (approved):
:lgtm:
:see_no_evil: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6893#-:-O87YSAG-xr2ugOI_OV-:b-fmxmqd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6938, opened 2024-10-07
Allow settings migration from packet tunnel

buggmagnet:
This PR adds the ability for the PacketTunnel settings to perform settings migration.
The net benefit of this is that users that have automatic app updates turned on won't have to manually disconnect and reconnect from the VPN when the updates itself, the upgrade should be transparent for them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938)

### DIFF
--- a/ios/MullvadREST/RetryStrategy/RetryStrategy.swift
+++ b/ios/MullvadREST/RetryStrategy/RetryStrategy.swift
@@ -82,9 +82,9 @@ extension REST {
         public static var failedMigrationRecovery = RetryStrategy(
             maxRetryCount: .max,
             delay: .exponentialBackoff(
-                initial: .seconds(10),
-                multiplier: UInt64(2),
-                maxDelay: .minutes(5)
+                initial: .seconds(5),
+                multiplier: UInt64(1),
+                maxDelay: .minutes(1)
             ),
             applyJitter: true
         )
--- a/ios/MullvadREST/RetryStrategy/RetryStrategy.swift
+++ b/ios/MullvadREST/RetryStrategy/RetryStrategy.swift
@@ -78,6 +78,16 @@ extension REST {
             ),
             applyJitter: true
         )
+
+        public static var failedMigrationRecovery = RetryStrategy(
+            maxRetryCount: .max,
+            delay: .exponentialBackoff(
+                initial: .seconds(10),
+                multiplier: UInt64(2),
+                maxDelay: .minutes(5)
+            ),
+            applyJitter: true
+        )
     }
 
     public enum RetryDelay: Equatable {
--- a/ios/MullvadSettings/MigrationManager.swift
+++ b/ios/MullvadSettings/MigrationManager.swift
@@ -23,38 +23,54 @@ public enum SettingsMigrationResult {
 
 public struct MigrationManager {
     private let logger = Logger(label: "MigrationManager")
+    private let cacheDirectory: URL
 
-    public init() {}
+    public init(cacheDirectory: URL) {
+        self.cacheDirectory = cacheDirectory.appendingPathComponent("migrationState.json")
+    }
 
     /// Migrate settings store if needed.
     ///
     /// Reads the current settings, upgrades them to the latest version if needed
     /// and writes back to `store` when settings are updated.
+    ///
+    /// In order to avoid migration happening from both the VPN and the host processes at the same time,
+    /// a non existant file path is used as a lock to synchronize access between the processes.
+    /// This file is accessed by `NSFileCoordinator` in order to prevent multiple processes accessing at the same time.
     /// - Parameters:
     ///   - store: The store to from which settings are read and written to.
-    ///   - proxyFactory: Factory used for migrations that involve API calls.
     ///   - migrationCompleted: Completion handler called with a migration result.
     public func migrateSettings(
         store: SettingsStore,
         migrationCompleted: @escaping (SettingsMigrationResult) -> Void
     ) {
-        let resetStoreHandler = { (result: SettingsMigrationResult) in
-            // Reset store upon failure to migrate settings.
-            if case .failure = result {
-                SettingsManager.resetStore()
-            }
-            migrationCompleted(result)
-        }
+        let fileCoordinator = NSFileCoordinator(filePresenter: nil)
+        var error: NSError?
 
-        do {
-            try upgradeSettingsToLatestVersion(
-                store: store,
-                migrationCompleted: migrationCompleted
-            )
-        } catch .itemNotFound as KeychainError {
-            migrationCompleted(.nothing)
-        } catch {
-            resetStoreHandler(.failure(error))
+        // This will block the calling thread if another process is currently running the same code.
+        // This is intentional to avoid TOCTOU issues, and guaranteeing settings cannot be read
+        // in a half written state.
+        // The resulting effect is that only one process at a time can do settings migrations.
+        // The other process will be blocked, and will have nothing to do as long as settings were successfully upgraded.
+        fileCoordinator.coordinate(writingItemAt: cacheDirectory, error: &error) { _ in
+            let resetStoreHandler = { (result: SettingsMigrationResult) in
+                // Reset store upon failure to migrate settings.
+                if case .failure = result {
+                    SettingsManager.resetStore()
+                }
+                migrationCompleted(result)
+            }
+
+            do {
+                try upgradeSettingsToLatestVersion(
+                    store: store,
+                    migrationCompleted: migrationCompleted
+                )
+            } catch .itemNotFound as KeychainError {
+                migrationCompleted(.nothing)
+            } catch {
+                resetStoreHandler(.failure(error))
+            }
         }
     }
 
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -696,6 +696,7 @@
 		A932D9F32B5EB61100999395 /* HeadRequestTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A932D9F22B5EB61100999395 /* HeadRequestTests.swift */; };
 		A932D9F52B5EBB9D00999395 /* RESTTransportStub.swift in Sources */ = {isa = PBXBuildFile; fileRef = A932D9F42B5EBB9D00999395 /* RESTTransportStub.swift */; };
 		A935594C2B4C2DA900D5D524 /* APIAvailabilityTestRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = A935594B2B4C2DA900D5D524 /* APIAvailabilityTestRequest.swift */; };
+		A939661B2CAE6CE1008128CA /* MigrationManagerMultiProcessUpgradeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A939661A2CAE6CE1008128CA /* MigrationManagerMultiProcessUpgradeTests.swift */; };
 		A94D691A2ABAD66700413DD4 /* WireGuardKitTypes in Frameworks */ = {isa = PBXBuildFile; productRef = 58FE25E22AA72AE9003D1918 /* WireGuardKitTypes */; };
 		A94D691B2ABAD66700413DD4 /* WireGuardKitTypes in Frameworks */ = {isa = PBXBuildFile; productRef = 58FE25E72AA7399D003D1918 /* WireGuardKitTypes */; };
 		A95EEE362B722CD600A8A39B /* TunnelMonitorState.swift in Sources */ = {isa = PBXBuildFile; fileRef = A95EEE352B722CD600A8A39B /* TunnelMonitorState.swift */; };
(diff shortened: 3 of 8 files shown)


pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/MullvadSettings/MigrationManager.swift` line 39 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8b-bGyBLWX2BsAJq9p:-O8b-bGyBLWX2BsAJq9q:b-ngnz3b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/MullvadSettings/MigrationManager.swift#L39)):*
> ```Swift
>     /// In order to avoid migration happening from both the VPN and the host processes at the same time,
>     /// a non existant file path is used as a lock is used to synchronize access between the processes.
>     /// This file is accessed by `NSFileCoordinator` in order to prevent multiple processes accessing at the same time.
> ```

If we can lock a file path, why not use the path to the settings file instead? No need for locking unrelated files.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @pinkisemils)

___
*[`ios/MullvadSettings/MigrationManager.swift` line 39 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8b-bGyBLWX2BsAJq9p:-O8b06Di4oTspFQsoHXl:bysg6cf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/MullvadSettings/MigrationManager.swift#L39)):*

If we can lock a file path, why not use the path to the settings file instead? No need for locking unrelated files.
</blockquote></details>

Because the settings are not stored in a file, but in the keychain itself

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/MullvadSettings/MigrationManager.swift` line 39 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8b-bGyBLWX2BsAJq9p:-O8b0n-Q0KRPIPbrLaO5:b-njaagj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/MullvadSettings/MigrationManager.swift#L39)):*

Because the settings are not stored in a file, but in the keychain itself
</blockquote></details>

You are correct.  But in that case, shouldn't this lock be acquired anytime the settings are read?

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadSettings/MigrationManager.swift` line 39 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8b-bGyBLWX2BsAJq9p:-O8bJJD0EoUQDKXpm5kH:bvcfzjs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/MullvadSettings/MigrationManager.swift#L39)):*

You are correct.  But in that case, shouldn't this lock be acquired anytime the settings are read?
</blockquote></details>

No, we do not need to do that if we ensure the migration is done when the UI process starts. This is sound now. Great work!
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 193 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8bJU2G9t_x-4acPwQ-:-O8bJU2G9t_x-4acPwQ0:b-7k4sds) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L193)):*
> ```Swift
>                         )
>                     DispatchQueue.main.asyncAfter(deadline: .now() + delay.timeInterval) { [unowned self] in
>                         performSettingsMigration()
> ```

If this does not block the startup, what does the packet tunnel read from the settings? Would it not fail to deserialize or read garbage constraints? Does it make sense to have the packet tunnel not block it's startup on this?

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 193 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8bJU2G9t_x-4acPwQ-:-O8bMCTHB23f2amEAw-_:b-30c2fv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L193)):*

Packet tunnel will get killed in 60 seconds if doesn't start anyway.
Our best hope is that we enter the blocked state until the user starts using their phones, at which point we're at the mercy of the retry logic.

Do keep in mind that this scenario is *extremely unlikely*. It would mean the phone both had a reboot AND and an app update before the user unlocked their phone after the reboot. 

I would not worry about those scenarios.
</blockquote></details>

The other scenario is that the phone was such in a bad state that we were not able to upgrade the settings. 
Our upgrade process right now is basically infaillible (since we removed the migration that did networking calls after SettingsV1).

rablador (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadREST/RetryStrategy/RetryStrategy.swift` line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8eu0_87o_JAt5ilLRI:-O8eu0_87o_JAt5ilLRJ:b-j64ttu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/MullvadREST/RetryStrategy/RetryStrategy.swift#L83)):*
> ```Swift
> 
>         public static var failedMigrationRecovery = RetryStrategy(
>             maxRetryCount: .max,
> ```

Are we going to retry this basically forever?
___
*[`ios/MullvadSettings/MigrationManager.swift` line 38 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8evG1E-fa193HbFXPD:-O8evG1E-fa193HbFXPE:b-muivs3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/MullvadSettings/MigrationManager.swift#L38)):*
> ```Swift
>     ///
>     /// In order to avoid migration happening from both the VPN and the host processes at the same time,
>     /// a non existant file path is used as a lock is used to synchronize access between the processes.
> ```

Nit: "is used as a lock is used to"
___
*[`ios/MullvadVPNTests/MullvadTypes/FileCacheTests.swift` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8ex0bk4FcmEp8w1xBe:-O8ex0bk4FcmEp8w1xBf:bwgzvz9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/MullvadVPNTests/MullvadTypes/FileCacheTests.swift#L10)):*
> ```Swift
> 
> import Foundation
> @testable import MullvadMockData
> ```

Added by mistake?
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 193 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8bJU2G9t_x-4acPwQ-:-O8exj1aB-Ji9EnCeT-m:b-oc61z3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L193)):*

The other scenario is that the phone was such in a bad state that we were not able to upgrade the settings. 
Our upgrade process right now is basically infaillible (since we removed the migration that did networking calls after SettingsV1).
</blockquote></details>

So basically, before migration happens, the tunnel is started and (possibly) enters blocked state. Then migration finishes and hopefully succeeds. The tunnel reconnection timer fires at some point and user leaves blocked state?

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938)* status: all files reviewed, 5 unresolved discussions (waiting on @pinkisemils and @rablador)

___
*[`ios/MullvadREST/RetryStrategy/RetryStrategy.swift` line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8eu0_87o_JAt5ilLRI:-O8f69qi8wZn9KtLGFPV:b-bxq1ft) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/MullvadREST/RetryStrategy/RetryStrategy.swift#L83)):*
That's what the ticket asks for:
> The packet tunnel must be able to handle a reboot - when the keychain is locked. It should retry with a backoff.
___
*[`ios/MullvadSettings/MigrationManager.swift` line 38 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8evG1E-fa193HbFXPD:-O8f6ckC9o9EYpDwbfcs:b-1akjx8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/MullvadSettings/MigrationManager.swift#L38)):*

Nit: "is used as a lock is used to"
</blockquote></details>

I reread this 3 times and yet... nice catch ! 
___
*[`ios/MullvadVPNTests/MullvadTypes/FileCacheTests.swift` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8ex0bk4FcmEp8w1xBe:-O8f6lmV6oXRcV8AnJgd:b-amoa76) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/MullvadVPNTests/MullvadTypes/FileCacheTests.swift#L10)):*

Added by mistake?
</blockquote></details>

yes, leftover of another approach I had tried where I wanted to add a test here, will delete. 
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 193 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8bJU2G9t_x-4acPwQ-:-O8f6s3_7w8NymihKsZh:b-gzdkwa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L193)):*
I want to emphasize that the odds of this failing are astronomically low for 2 reasons:
- The `upgradeToNextVersion` function cannot fail. i.e. the only failure mode is keychain settings have been badly corrupted (or we messed up our next `TunnelSettings` by adding fields that do not deserialize), at which point we cannot guarantee the app will work reliably at all. If this fails, upgrading from the UI process will end in the same result, and the user will have to re-install the app in any case
- The scenario in which we cannot read from the settings because the keychain is still locked only happens if the user's device rebooted without the user interacting with the device (since they would otherwise unlock the device to use it). We could actually remove this restriction by using `kSecAttrAccessibleAlways` instead. If we're really worried about other apps reading our keychain entries (or reading them from a backup) there are better solutions around. 

> before migration happens, the tunnel is started and (possibly) enters blocked state.

Migration is not asynchronous, the tunnel won't start before migration is done, therefore we cannot enter the blocked state _because_ we are waiting on migration. It either succeeds and we start the tunnel, or it fails

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadREST/RetryStrategy/RetryStrategy.swift` line 83 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8eu0_87o_JAt5ilLRI:-O8fBm1i1mBLzA9aiUr5:bk3b2kh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/MullvadREST/RetryStrategy/RetryStrategy.swift#L83)):*

That's what the ticket asks for:
> The packet tunnel must be able to handle a reboot - when the keychain is locked. It should retry with a backoff.
</blockquote></details>

Still, forever...?
___
*[`ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift` line 193 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-O8bJU2G9t_x-4acPwQ-:-O8fBLKk13qgNCuT36uA:b-ewbwbv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ff36ba9521c08aa894c974ccab2d0fb26c5f2c27/ios/PacketTunnel/PacketTunnelProvider/PacketTunnelProvider.swift#L193)):*
> Migration is not asynchronous, the tunnel won't start before migration is done

Aha, then all is good I think. Checked the Linear issue just now and one of the acceptace criteria also states that blocked state should not happen.

rablador (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-:-O8fuD7xESNWEu9eXuMN:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

rablador (approved):
> Reviewed 1 of 1 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938)* status: all files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

pinkisemils (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6938#-:-O8kNJ4_4Bf4sjh-3xQz:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #6993, opened 2024-10-16
Fix parsing of build tags

Serock3:
This PR fixes a few issue regarding how the test framework used tags when constructing the path to the build artifact.

* The regex which parsed the build version was broken because it included quotation marks. 
* The test framework never pruned tags, which made it use local tags even when they were outdated.
* The buildserver scripts remove forbidden characters from the "+$tag" suffix , which is now mirrored by the test framework.

I also split the parsing of build version regex from the function that decides if a build is a dev version.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6993)

### DIFF
--- a/.github/workflows/desktop-e2e.yml
+++ b/.github/workflows/desktop-e2e.yml
@@ -158,7 +158,7 @@ jobs:
       - name: Run end-to-end tests
         shell: bash -ieo pipefail {0}
         run: |
-          git fetch --tags --force
+          git fetch --tags --prune-tags --force
           export TEST_FILTERS="${{ github.event.inputs.tests }}"
           ./test/scripts/ci-runtests.sh ${{ matrix.os }}
       - uses: actions/upload-artifact@v3
@@ -234,7 +234,7 @@ jobs:
       - name: Run end-to-end tests
         shell: bash -ieo pipefail {0}
         run: |
-          git fetch --tags --force
+          git fetch --tags --prune-tags --force
           export TEST_FILTERS="${{ github.event.inputs.tests }}"
           ./test/scripts/ci-runtests.sh ${{ matrix.os }}
       - uses: actions/upload-artifact@v3
@@ -311,7 +311,7 @@ jobs:
       - name: Run end-to-end tests
         shell: bash -ieo pipefail {0}
         run: |
-          git fetch --tags --force
+          git fetch --tags --prune-tags --force
           export TEST_FILTERS="${{ github.event.inputs.tests }}"
           ./test/scripts/ci-runtests.sh ${{ matrix.os }}
       - uses: actions/upload-artifact@v3
--- a/test/scripts/test-utils.sh
+++ b/test/scripts/test-utils.sh
@@ -91,12 +91,23 @@ function nice_time {
     echo "\"$*\" completed in $((s/60))m:$((s%60))s"
     return $result
 }
+# Matches $1 with a build version string and sets the following exported variables:
+# - BUILD_VERSION: The version part of the build string (e.g., "2024.3-beta1-dev-").
+# - COMMIT_HASH: The commit hash part of the build string (e.g., "abcdef").
+# - TAG: The tag part of the build string (e.g., "+tag").
+function parse_build_version {
+    if [[ "$1" =~ (^[0-9.]+(-beta[0-9]+)?-dev-)([0-9a-z]+)(\+[0-9a-z|-]+)?$ ]]; then
+        BUILD_VERSION="${BASH_REMATCH[1]}"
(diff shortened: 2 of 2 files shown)


dlon (dismissed):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6993)* status: all files reviewed, 1 unresolved discussion (waiting on @Serock3)

___
*[`test/scripts/test-utils.sh` line 50 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6993#-O9i61241voBuNScjs5u:-O9i61241voBuNScjs5v:b-2nb3cr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ba141694f3fba2555bf110bebaf48d3b39622f81/test/scripts/test-utils.sh#L50)):*
> ```Shell
> if [[ -n "$TAG" && ${CURRENT_VERSION} =~ -dev- ]]; then
>     # Remove disallowed version characters from the tag
>     CURRENT_VERSION+="+${TAG//[^0-9a-z_-]/}"
> ```

Should this really strip uppercase letters? But I guess that's already the case in `buildserver-build.sh` as well, so being consistent makes sense. :shrug:

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6993)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`test/scripts/test-utils.sh` line 50 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6993#-O9i61241voBuNScjs5u:-O9iMIVK2D7doLbmi8ex:bgdp7ch) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/ba141694f3fba2555bf110bebaf48d3b39622f81/test/scripts/test-utils.sh#L50)):*

Should this really strip uppercase letters? But I guess that's already the case in `buildserver-build.sh` as well, so being consistent makes sense. :shrug:
</blockquote></details>

Really weird, but if we don't match the behavior we don't get the correct URLs :shrug: I guess we could change it in both places if you think it would be an improvement.

raksooo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6993)* status: all files reviewed, 1 unresolved discussion (waiting on @Serock3)

___
*[`test/scripts/test-utils.sh` line 99 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6993#-O9ixtwSD14NiX8rupIA:-O9ixtwT0VDv2EtikEjr:b-etdg8h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/46167b45c080da7787f9e1159598c50fc4bdbf66/test/scripts/test-utils.sh#L99)):*
> ```Shell
> # - COMMIT_HASH: The commit hash part of the build string (e.g., "abcdef").
> # - TAG: The tag part of the build string (e.g., "+tag").
> function parse_build_version {
> ```

I find it a bit confusing that this function doesn't return the values but set them as global variables. If my memory isn't failing me it's not possible to return multiple values in a simple way though. I would suggest splitting this function into three different functions that each return one of build version, commit hash and tag.

raksooo (dismissed):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6993)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Serock3 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6993)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @dlon and @raksooo)

___
*[`test/scripts/test-utils.sh` line 99 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6993#-O9ixtwSD14NiX8rupIA:-O9oEauw-8H0TLmoBJyY:bc3v2aw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/46167b45c080da7787f9e1159598c50fc4bdbf66/test/scripts/test-utils.sh#L99)):*

I find it a bit confusing that this function doesn't return the values but set them as global variables. If my memory isn't failing me it's not possible to return multiple values in a simple way though. I would suggest splitting this function into three different functions that each return one of build version, commit hash and tag.
</blockquote></details>

As per our discussion, I will keep the current implementation

raksooo (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/6993)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7003, opened 2024-10-17
Minor cleanup in tunnel config client

faern:
As part of my cleanup week I decided to make some small improvements to the ephemeral peer negotiation code.

My main pet peeve that led me to this was:
```rust
fn post_quantum_secrets(enable: bool) {
    if enable {
        ...; Some(...)
    } else {
        None
    }
}
```
This is a code smell. So I moved the conditional logic out. It creates a few more lines of code, but makes the lower level function cleaner and focused on what its supposed to do.

I also made a few minor tweaks to improve documentation in the crate:
* Document why we run Classic McEliece in a separate thread
* Put higher level function over lower level function
* rename one instance of "post quantum" to "quantum resistant". We are not going to get away from the discrepancy that the field in the gRPC endpoint is called `post_quantum` and that the user facing feature is called `quantum-resistant tunnel`. But out of the two I think "quantum resistant" makes more sense most of the time. "Post quantum" in itself means nothing. At most it refers to the time after when quantum computers have been invented. It talks about nothing else. "Post quantum cryptography" is a term, but you need the word "cryptography" for it to have any meaning. And even then, it does not say much in our context. We want to express whether or not the ephemeral peer should be safe against quantum computers. So using "quantum resistant" is the proper term here.
* Added docs to link the two ephemeral peer MTU fix codes together. Ideally they would be refactored, but I won't do that in this PR.

### DIFF
--- a/talpid-tunnel-config-client/src/classic_mceliece.rs
+++ b/talpid-tunnel-config-client/src/classic_mceliece.rs
@@ -2,9 +2,7 @@ use classic_mceliece_rust::{keypair_boxed, Ciphertext, CRYPTO_CIPHERTEXTBYTES};
 pub use classic_mceliece_rust::{PublicKey, SecretKey, SharedSecret};
 
 /// The `keypair_boxed` function needs just under 1 MiB of stack in debug
-/// builds. Even though it probably works to run it directly on the main
-/// thread on all OSes, we take this precaution and always generate the huge
-/// keys on a separate thread with a large enough stack.
+/// builds.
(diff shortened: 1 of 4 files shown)


dlon (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7003#-:-O9PVVAM1ckhV-afwzAX:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7008, opened 2024-10-18
Move encrypted DNS state to crate shared between iOS and everyeone else

MarkusPettersson98:
Move the Encrypted DNS config selection algorithm to the `mullvad-encrypted-dns-proxy` crate where both `mullvad-ios` and `mullvad-daemon` can consume it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7008)

### DIFF
--- a/mullvad-encrypted-dns-proxy/src/lib.rs
+++ b/mullvad-encrypted-dns-proxy/src/lib.rs
@@ -1,13 +1,12 @@
 //! Mullvad Encrypted DNS proxy is a custom protocol for reaching the Mullvad API over proxies,
-//! with some amont of simple obfuscation applied.
+//! with some amount of simple obfuscation applied.
 //!
 //! The proxy endpoints and what obfuscation they expect is fetched over DNS-over-HTTPS (DoH)
 //! in AAAA records. The AAAA (IPv6) records are then decoded into a proxy config consisting
 //! of a remote endpoint to connect to, and what obfuscation to use.
-//!
+pub use forwarder::Forwarder;
 
 pub mod config;
 pub mod config_resolver;
 mod forwarder;
-
-pub use forwarder::Forwarder;
+pub mod state;
--- a/mullvad-encrypted-dns-proxy/src/state.rs
+++ b/mullvad-encrypted-dns-proxy/src/state.rs
@@ -1,0 +1,96 @@
+//! This module defines a cache for Encrypted DNS proxy configs. The cache contains a method for
+//! fetching new configs as needed.
+
+use std::collections::HashSet;
+
+use crate::config::ProxyConfig;
+use crate::config_resolver::{self, resolve_default_config};
+
+/// Keep track of fetched proxy configurations.
+///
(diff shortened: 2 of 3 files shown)


MarkusPettersson98:
Maybe it makes more sense to name the `state` module `cache` instead? :thinking:

dlon (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7008#-:-O9TvEF71prsOhtBlQEB:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`mullvad-encrypted-dns-proxy/src/state.rs` line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7008#-O9TsqL77mFzErGDZflk:-O9TsqL77mFzErGDZfll:b-ngz3he) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/19b670dcbdfadf6a3e3785a820b61aec534e84d1/mullvad-encrypted-dns-proxy/src/state.rs#L34)):*
> ```XML
>     /// Select a config.
>     /// Always select an obfuscated configuration, if there are any left untried. If no obfuscated
>     /// configurations exist, try plain configurations. The order is randomized due to the hash set
> ```

Nit: The fact that hash sets are used sounds like an implementation detail. Maybe it should not be mentioned here.

dlon (approved):
> Maybe it makes more sense to name the `state` module `cache` instead? 🤔

Not to bikeshed, but I guess that this module has two responsibilities: (1) owning a cache, (2) keeping track of used configs. Maybe "state" is vague enough to encompass both?

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7008#-:-O9TzNTJ0faPI58mdZHg:b-vsmhf0)* status: all files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

MarkusPettersson98 (commented):
Fair enough, point 2 would make 'cache' misleading. 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7008#-:-O9UPgSS4ZQ7SYHUuI_d:bnafur)* status: all files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`mullvad-encrypted-dns-proxy/src/state.rs` line 34 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7008#-O9TsqL77mFzErGDZflk:-O9UPrcF-m9lS5dDyphM:b3xr6fx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bf6d35877e8a774e3e0d9eba1a656eea50326cdf/mullvad-encrypted-dns-proxy/src/state.rs#L34)):*

Nit: The fact that hash sets are used sounds like an implementation detail. Maybe it should not be mentioned here.
</blockquote></details>

I agree that it an implementation detail, but this is something we would like to change eventually. Making it as visible as possible might make us consider changing it sooner rather than later :shrug: ?

PR #7038, opened 2024-10-22
Make the daemon aware if ipv4 and/or ipv6 is available

Pururun:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7038)


Pururun:
Closing this since there is a better solution and most code is outdated.

PR #7053, opened 2024-10-23
Improve DnsDialogViewModel nav arguments

kl:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7053)

### DIFF
--- a/android/app/src/androidTest/kotlin/net/mullvad/mullvadvpn/compose/screen/VpnSettingsScreenTest.kt
+++ b/android/app/src/androidTest/kotlin/net/mullvad/mullvadvpn/compose/screen/VpnSettingsScreenTest.kt
@@ -378,7 +378,7 @@ class VpnSettingsScreenTest {
     fun testClickAddDns() =
         composeExtension.use {
             // Arrange
-            val mockedClickHandler: (Int?, String?) -> Unit = mockk(relaxed = true)
+            val mockedClickHandler: (Int?) -> Unit = mockk(relaxed = true)
             setContentWithTheme {
                 VpnSettingsScreen(
                     state = VpnSettingsUiState.createDefault(isCustomDnsEnabled = true),
@@ -390,7 +390,7 @@ class VpnSettingsScreenTest {
             onNodeWithText("Add a server").performClick()
 
             // Assert
-            verify { mockedClickHandler.invoke(null, null) }
+            verify { mockedClickHandler.invoke(null) }
         }
 
     @Test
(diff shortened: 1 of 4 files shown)


Pururun (dismissed):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7053#-:-O9xo2BaDCJJZ7yM72xV:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Rawa (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7053)* status: all files reviewed, 1 unresolved discussion (waiting on @kl)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/DnsDialogViewModel.kt` line 81 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7053#-OAMEDpVEh7-kXteRsXg:-OAMEDpVEh7-kXteRsXh:b-r161v0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/98375f95b3988216595923c49ecdecb7e7bc4c97/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/DnsDialogViewModel.kt#L81)):*
> ```Kotlin
>                 viewModelScope,
>                 SharingStarted.Lazily,
>                 createViewState(emptyList(), null, false, EMPTY_STRING),
> ```

I'm a bit torn of this, I think it would be nicer if we directly show the correct state.

Right now if you click on an existing entry we would incorrectly show `EMPTY_STRING` first. Maybe it is necessary to introduce a loading state?

Some other remarks that I think we can do better:
- We should review if we can make `MutableStateFlow<Settings?>` more clear. It doesn't seem to add much in it current form ( except as acting like a cache for the first Settings object we saw to avoid an update as we press save.)
- _ipAddressInput being nullable feels wrong, it looks like the user can input a "null" but that wouldn't be valid input from a user. (It is a backing field and `onDnsInputChange` only accepts String, but it gets a bit confusing. Maybe `lateinit` until we get first `Settings`?
- The following line does a lot of heavy lifting and isn't very readable:
```
                val input =
                    inputAddress
                        ?: navArgs.index?.let { index -> dnsList[index].addressString() }
                        ?: EMPTY_STRING
```
E.g that if `navArgs.index` would either be always null or always a value.

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7053)* status: all files reviewed, 1 unresolved discussion (waiting on @kl)

PR #7062, opened 2024-10-24
Add note about Application Guard being deprecated

faern:
As we discussed briefly AFK. Since Microsoft treats Application Guard as deprecated, we see very little value in spending a large effort into fixing it. Instead we should just accept the state of things and recommend users to not use Application Guard.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7062)


raksooo (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7062)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7064, opened 2024-10-24
Update data structure to support new obfuscation selection

rablador:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064)

### DIFF
--- a/ios/MullvadSettings/WireGuardObfuscationSettings.swift
+++ b/ios/MullvadSettings/WireGuardObfuscationSettings.swift
@@ -8,54 +8,158 @@
 
 import Foundation
 
-/// Whether UDP-over-TCP obfuscation is enabled
+/// Whether obfuscation is enabled and which method is used
 ///
 /// `.automatic` means an algorithm will decide whether to use it or not.
 public enum WireGuardObfuscationState: Codable {
-    case automatic
+    @available(*, deprecated, renamed: "udpOverTcp")
     case on
+
+    case automatic
+    case udpOverTcp
+    case shadowsocks
     case off
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        var allKeys = ArraySlice(container.allKeys)
+        guard let key = allKeys.popFirst(), allKeys.isEmpty else {
+            throw DecodingError.typeMismatch(
+                WireGuardObfuscationState.self,
+                DecodingError.Context(
+                    codingPath: container.codingPath,
+                    debugDescription: "Invalid number of keys found, expected one.",
+                    underlyingError: nil
+                )
+            )
+        }
+
+        switch key {
+        case .automatic:
+            self = .automatic
+        case .on, .udpOverTcp:
+            self = .udpOverTcp
+        case .shadowsocks:
+            self = .shadowsocks
+        case .off:
+            self = .off
+        }
+    }
 }
 
-/// The port to select when using UDP-over-TCP obfuscation
-///
-/// `.automatic` means an algorith will decide between using `port80` or `port5001`
-public enum WireGuardObfuscationPort: UInt16, Codable, CaseIterable {
-    case automatic = 0
-    case port80 = 80
-    case port5001 = 5001
+public enum WireGuardObfuscationUdpOverTcpPort: Codable, Equatable, CustomStringConvertible {
+    case automatic
+    case port80
+    case port5001
 
-    /// The `UInt16` representation of the port.
-    /// - Returns: `0` if `.automatic`, `80` or `5001` otherwise.
-    public var portValue: UInt16 {
-        self == .automatic ? 0 : rawValue
-    }
-
-    public init?(rawValue: UInt16) {
-        switch rawValue {
-        case 80:
-            self = .port80
-        case 5001:
-            self = .port5001
-        default: self = .automatic
+    public var portValue: UInt16? {
+        switch self {
+        case .automatic:
+            nil
+        case .port80:
+            80
+        case .port5001:
+            5001
         }
     }
 
-    public init(from decoder: Decoder) throws {
-        let container = try decoder.singleValueContainer()
-        let decodedValue = try? container.decode(UInt16.self)
-
-        let port = WireGuardObfuscationPort.allCases.first(where: { $0.rawValue == decodedValue })
-        self = port ?? .automatic
+    public var description: String {
+        switch self {
+        case .automatic:
+            NSLocalizedString(
+                "WIREGUARD_OBFUSCATION_UDP_TCP_PORT_AUTOMATIC",
+                tableName: "VPNSettings",
+                value: "Automatic",
+                comment: ""
+            )
+        case .port80:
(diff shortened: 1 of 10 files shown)


mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064)* status: 0 of 21 files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadSettings/WireGuardObfuscationSettings.swift` line 128 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064#-OAq_Tls5k8jAJSDrhP7:-OAq_Tls5k8jAJSDrhP8:b-s32ihp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/067219c91010e7f6d63796bc7855e50c0a799ea5/ios/MullvadSettings/WireGuardObfuscationSettings.swift#L128)):*
> ```Swift
> 
>     public let state: WireGuardObfuscationState
>     public let udpOverTcpPort: WireGuardObfuscationUdpOverTcpPort
> ```

are we going to keep the previous state when user is changing obfuscation method?  in this architecture, if we add another state into `WireGuardObfuscationState` that has its own custom configuration then we need to add another property for that here, as well. is it the thing we will?

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064)* status: 0 of 21 files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadSettings/WireGuardObfuscationSettings.swift` line 56 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064#-OAriaWsA_KDeBwtO-JK:-OAvn_30D5cyqH2CqKb8:b-4ncs6g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/067219c91010e7f6d63796bc7855e50c0a799ea5/ios/MullvadSettings/WireGuardObfuscationSettings.swift#L56)):*

This should not be `Optional`, we always know the real value of the port.
</blockquote></details>

We don't, since `.automatic` isn't really equal to eg. 0 we should properly represent that a port has not been explicity set.
___
*[`ios/MullvadSettings/WireGuardObfuscationSettings.swift` line 128 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064#-OAq_Tls5k8jAJSDrhP7:-OAvmra274n16rSsb88g:bcwftrm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/067219c91010e7f6d63796bc7855e50c0a799ea5/ios/MullvadSettings/WireGuardObfuscationSettings.swift#L128)):*

are we going to keep the previous state when user is changing obfuscation method?  in this architecture, if we add another state into `WireGuardObfuscationState` that has its own custom configuration then we need to add another property for that here, as well. is it the thing we will?
</blockquote></details>

I think so. Doing it this way we keep each port setting no matter what other settings the user changes, so that's according to spec I think. Unfortunately we could not - in a good way - combine state and port into a single enum. The reason for that being that we set state and port in different places at different times, which makes it really bothersome to keep track of in a viewmodel.
___
*[`ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift` line 61 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064#-OArjCKY-fbKYvuXFjXc:-OAvowIC-vU4Mi91pN6X:bpv8cqg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/067219c91010e7f6d63796bc7855e50c0a799ea5/ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift#L61)):*

This is the reason why `portValue` shouldn't be nullable. `0` is an invalid port number in this case, and it should be impossible
to have this condition here since if the obfuscation is set to `automatic` it would either be `80` or `5001`
</blockquote></details>

That won't be true for long. This code will change when the relay selector PR goes up, and this part will handle the custom port for shadowsocks as well, which can be anything. I don't think we should be relying on an Int that's implicitly wrong if it happens to be a 0 but correct if it's a 1.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064)* status: 0 of 21 files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift` line 61 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064#-OArjCKY-fbKYvuXFjXc:-OAvwAuh0jgeJXjFsIOb:bqnx9h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/067219c91010e7f6d63796bc7855e50c0a799ea5/ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift#L61)):*

That won't be true for long. This code will change when the relay selector PR goes up, and this part will handle the custom port for shadowsocks as well, which can be anything. I don't think we should be relying on an Int that's implicitly wrong if it happens to be a 0 but correct if it's a 1.
</blockquote></details>

Just to clarify, in the new code there won't be any fallbacks like this.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064)* status: 0 of 21 files reviewed, 2 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadSettings/WireGuardObfuscationSettings.swift` line 56 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064#-OAriaWsA_KDeBwtO-JK:-OAw6RCd3VWldTGe154t:bxplirk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/067219c91010e7f6d63796bc7855e50c0a799ea5/ios/MullvadSettings/WireGuardObfuscationSettings.swift#L56)):*

We don't, since `.automatic` isn't really equal to eg. 0 we should properly represent that a port has not been explicity set.
</blockquote></details>

This was discussed offline.
___
*[`ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift` line 61 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064#-OArjCKY-fbKYvuXFjXc:-OAw6VNwA2XedQ7jQHVU:b-93926e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/067219c91010e7f6d63796bc7855e50c0a799ea5/ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift#L61)):*

Just to clarify, in the new code there won't be any fallbacks like this.
</blockquote></details>

Sounds good then, a temporary solution is fine.

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7064#-:-OAwrAQ_4Wyk99IZJUrB:bnfp4nl)* status: 0 of 21 files reviewed, 1 unresolved discussion (waiting on @mojganii)

PR #7105, opened 2024-10-30
Stop duplication in error source chain

faern:
Don't both print the contained error in the Display impl and also return it in source.

The error, as it was before, would print an error chain like this:
```
Error: Resolution error something something
Caused by: Resolution error something something
```

An error should avoid including its source error in its own `Display` impl. Or, if it is included, it should not return that from the `source` method.

If you want to dive more into how long I have cared about this, you can follow this URL :D https://github.com/rust-lang/api-guidelines/pull/210

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7105)


MarkusPettersson98 (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7105)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7115, opened 2024-11-01
Update documentation around how osv-scanner ignores should be written

faern:
This PR has improvements and clarifications around what we should do when `osv-scanner` reports a vulnerability. The original cause for this PR was that it was unclear if we were allowed to ignore entire packages, and how to write such a `PackageOverrides` entry. Then I also took the opportunity to improve the beginning of the document, taking a broader view of how to handle vulnerability reports overall.

I also slightly changed the order of some paragraphs to make them come in a bit more of a priority order. Priority in the sense that expiry date and reason are the most important fields to specify correctly. The exact format of `id` is not as important to get right.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7115)


raksooo (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7115#-:-OAqVPYg9fk3GKMWs9L7:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7137, opened 2024-11-06
Disable `fault_handler` logging in release builds

hulthe:
Also add a reentrancy-guard to the fault handler

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7137)

PR #7142, opened 2024-11-07
Add changelog entry for "Disable fault handler in release builds"

hulthe:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7142)


faern:
Discoverability is key. So adding a link to the PR this is a follow up to: #7137

PR #7148, opened 2024-11-08
Disable module verifier on all targets

buggmagnet:
This PR disables the [Module Verifier](https://developer.apple.com/documentation/xcode/identifying-and-addressing-framework-module-issues) option that was present in most targets.

As we do not distribute our binaries separately, and we do not add new frameworks left and right that often, we do not need this setting permanently enabled, it adds 30 seconds to each build (even with cache)

We should probably use the setting when we add a new Framework, to be on the safe side, but make sure to disable it anyway.


rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7148)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7155, opened 2024-11-11
Remove old waitForTunnelUp function

Rawa:


albin-mullvad (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7155)* status: all files reviewed, 1 unresolved discussion (waiting on @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/TalpidVpnService.kt` line 141 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7155#-OBPGP5TCpEG7A0kFSe-:-OBPGP5TCpEG7A0kFSe0:bjrlayv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/82f31fb9a647e0736e3e999044126542a25654f2/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/TalpidVpnService.kt#L141)):*
> ```Kotlin
>         val tunFd = vpnInterfaceFd.detachFd()
> 
>         waitForTunnelUp(tunFd, config.routes.any { route -> route.isIpv6 })
> ```

Were we able to find any historical reasons for adding this wait by looking in the git history? Would also be good to clarify in the git commit why we remove it as a future reference in case it causes any issues.



_Code quote:_
```Kotlin
waitForTunnelUp
```

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7155)* status: all files reviewed, 1 unresolved discussion (waiting on @albin-mullvad)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/TalpidVpnService.kt` line 141 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7155#-OBPGP5TCpEG7A0kFSe-:-OBPOcWzBTKjVUOMV41L:b-269c8f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e914beb13e7099162ece01cd534bd7e4c89e91d5/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/TalpidVpnService.kt#L141)):*

Were we able to find any historical reasons for adding this wait by looking in the git history? Would also be good to clarify in the git commit why we remove it as a future reference in case it causes any issues.
</blockquote></details>

:100:% agree that we should try to document why this was removed. Doing so in the commit message ought to be enough

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7155)* status: 2 of 6 files reviewed, 1 unresolved discussion (waiting on @albin-mullvad and @pinkisemils)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/TalpidVpnService.kt` line 141 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7155#-OBPGP5TCpEG7A0kFSe-:-OBVlCUW7AX2645dVPYF:bdse425) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e914beb13e7099162ece01cd534bd7e4c89e91d5/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/TalpidVpnService.kt#L141)):*

:100:% agree that we should try to document why this was removed. Doing so in the commit message ought to be enough
</blockquote></details>

I've added some documentation, but I'll be putting this on hold for a bit since we want to consider doing this in the daemon instead. I talked with @pinkisemils whom proposed that we can setup a socket with `NETLINK_ROUTE` and then get events for when the routing tables changes. This would possible be faster than this implementation that uses `connectivityManager.registerDefaultNetworkCallback()`, which seem to take at least 50ms on a debug build.

Rawa:
Closing in wait for us to plan DROID-1610

PR #7172, opened 2024-11-13
Make unix signal-handler signal-safe

hulthe:
Other changes:
- Re-enable the signal handler in release-builds.
- Disable backtrace printing by default since it's not signal-safe.
- Add `MULLVAD_BACKTRACE_ON_FAULT` env variable to enable backtracing (Yes, that makes it not signal safe! But something we are willing to have as opt-in due to how it can help us debug stuff).
- Remove the alternate signal stack. The reasons for this are:
  - Setting up an alt-stack in a safe way is not trivial, our previous
    attempt was unsound in the presence of stack overflows. It can be
    done safely with mmap, but would require careful review.
  - The alt-stack is thread-local, meaning it would need to be
    initialized on a per-thread basis. We would need to hook into tokio
    and std::thread::spawn to be able to get good coverage, and even
    then there would no good way to ensure that *all* threads have an
    alternate stack, except that...
  - Rust (by default) allocates an alternate stack for every thread.
    Unfortunately, the prescence of Go code in our linked binary
    disables this feature. IMO, we should strive towards not having any
    Go code linked into the daemon for this reason.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,13 @@ See [this](Release.md) for instructions on how to make a new release.
   interface UDS socket to users in the specified group. This means that only users in that group can
   use the CLI and GUI. By default, everyone has access to the socket.
 
+* `MULLVAD_BACKTRACE_ON_FAULT` - When enabled, if the daemon encounters a fault (e.g. `SIGSEGV`),
+  it will log a backtrace to stdout, and to `daemon.log`. By default, this is disabled in
+  release-builds and enabled in debug-builds. Set variable to `1` or `0` to explicitly enable or
+  disable this feature. Logging the backtrace cause heap allocation. Allocation is not signal safe,
+  but here it runs in the signal handler. This in technically undefined behavior and therefore
+  disabled by default. This usually works, but enable at your own risk.
+
 ### Development builds only
 
 * `MULLVAD_API_HOST` - Set the hostname to use in API requests. E.g. `api.mullvad.net`.
--- a/mullvad-daemon/src/exception_logging/mod.rs
+++ b/mullvad-daemon/src/exception_logging/mod.rs
@@ -8,4 +8,4 @@ pub use win::enable;
 mod unix;
 
 #[cfg(unix)]
-pub use unix::enable;
+pub use unix::{enable, set_log_file};
--- a/mullvad-daemon/src/exception_logging/unix.rs
+++ b/mullvad-daemon/src/exception_logging/unix.rs
@@ -1,145 +1,283 @@
-//! Installs signal handlers to catch critical program faults and logs them. See [`enable`].
-//!
-//! NOTE: The signal handlers are disabled in release-builds. See docs on
-//! `handler::logging_fault_handler` for reasoning.
+//! Install signal handlers to catch critical program faults and log them. See [`enable`].
+#![warn(clippy::undocumented_unsafe_blocks)]
 
-pub use handler::enable;
+use libc::siginfo_t;
+use nix::sys::signal::{sigaction, SaFlags, SigAction, SigHandler, SigSet, Signal};
 
-#[cfg(not(debug_assertions))]
-mod handler {
-    /// This is a no-op on release-builds; it does NOT install a signal handler.
-    pub fn enable() {
-        // No fault handler in release-builds.
+use core::fmt;
+use std::{
+    backtrace::Backtrace,
+    env,
+    ffi::{c_int, c_void, CString},
+    os::fd::{FromRawFd, RawFd},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Once, OnceLock,
+    },
+};
+
+/// Write fault to this file.
+static LOG_FILE_PATH: OnceLock<CString> = OnceLock::new();
+
+/// If true, the signal-handler will log a backtrace when triggered.
+///
+/// Default value is `true` for debug-builds, and `false` for release-builds.
+/// This can be overridden by setting the env var [ENABLE_BACKTRACE_VAR] to `1` or `0`.
+///
+/// # Safety
+/// The [Backtrace] implementation we use does not guarantee signal-safety; Invoking it from a
+/// signal handler can be unsound. In practice this usually works fine, but to avoid any risk to
+/// users, we disable backtracing by default in release-builds.
+static ENABLE_BACKTRACE: AtomicBool = AtomicBool::new(cfg!(debug_assertions));
+
+/// Name of the environment variable that sets [ENABLE_BACKTRACE].
+const ENABLE_BACKTRACE_VAR: &str = "MULLVAD_BACKTRACE_ON_FAULT";
+
+/// The signals we install handlers for.
+const FAULT_SIGNALS: [Signal; 5] = [
+    // Access to invalid memory address
+    Signal::SIGBUS,
+    // Floating point exception
+    Signal::SIGFPE,
+    // Illegal instructors
+    Signal::SIGILL,
+    // Invalid memory reference
+    Signal::SIGSEGV,
+    // Bad syscall
+    Signal::SIGSYS,
+];
+
+/// Set the file path used for fault handler logging.
+///
+/// Panics if called more than once.
+pub fn set_log_file(file_path: impl Into<CString>) {
+    if let Err(_file_path) = LOG_FILE_PATH.set(file_path.into()) {
+        panic!("set_log_file may not be called more than once");
     }
 }
 
-#[cfg(debug_assertions)]
-mod handler {
-    use libc::siginfo_t;
-    use nix::sys::signal::{sigaction, SaFlags, SigAction, SigHandler, SigSet, Signal};
+/// Install signal handlers to catch critical program faults, log them, and exit the process.
+pub fn enable() {
+    static INIT_ONCE: Once = Once::new();
 
-    use std::{
-        backtrace::Backtrace,
-        ffi::{c_int, c_void},
-        sync::{
-            atomic::{AtomicBool, Ordering},
-            Once,
+    INIT_ONCE.call_once(|| {
+        if let Ok(override_backtrace) = env::var(ENABLE_BACKTRACE_VAR).map(|v| v != "0") {
+            ENABLE_BACKTRACE.store(override_backtrace, Ordering::Release);
+        }
+
+        // SA_ONSTACK tells the signal handler to use an alternate stack, if one is available.
+        // The purpose of an alternate stack is to have somewhere to execute the signal handler in
+        // the case of a stack overflow. I.e. if an alternate stack hasn't been configured, stack
+        // overflows will silently cause the process to exit with code SIGSEGV.
+        //
+        // `libc::sigaltstack` can be used to set up an alternate stack on the current thread. The
+        // default behaviour of the Rust runtime is to set up alternate stacks for the main thread,
+        // and for every thread spawned using `std::thread`. However, Rust will not do this if any
+        // signal handlers have been configured before the Rust runtime is initialized (note: the
+        // initialization happens before main() is called). For example, if any Go code is linked
+        // into this binary, the Go runtime will probably be initialized first, and will set up
+        // it's own signal handlers.
+        //
+        // Go requires this flag to be set for all signal handlers.
+        // https://github.com/golang/go/blob/d6fb0ab2/src/os/signal/doc.go
+        let sig_handler_flags = SaFlags::SA_ONSTACK;
+
+        let signal_action = SigAction::new(
+            SigHandler::SigAction(fault_handler),
+            sig_handler_flags,
+            SigSet::empty(),
+        );
+
+        for signal in &FAULT_SIGNALS {
+            // SAFETY:
+            // `fault_handler` is signal-safe if ENABLE_BACKTRACE is false.
+            // See docs on ENABLE_BACKTRACE.
+            if let Err(err) = unsafe { sigaction(*signal, &signal_action) } {
+                log::error!("Failed to install signal handler for {}: {}", signal, err);
+            }
+        }
+    });
+}
+
+/// Signal handler to catch signals that are used to indicate unrecoverable errors in the daemon.
+///
+/// # Safety
+/// This function is signal-safe if [ENABLE_BACKTRACE] is false.
+//
+// NOTE: When implementing, make sure to adhere to the rules of signal-safety!
+// For a detailed definition, see https://man7.org/linux/man-pages/man7/signal-safety.7.html
+// The short version is:
+// - This function must be re-entrant.
+// - This function must only call functions that are signal-safe.
+// - The man-page provides a list of posix-functions that are signal-safe. (These can be found
+//   in the `libc`-crate)
+extern "C" fn fault_handler(
+    signum: c_int,
+    _siginfo: *mut siginfo_t,
+    _thread_context_ptr: *mut c_void,
+) {
+    // 128 + signum is the convention set by bash.
+    let signum_code: c_int = signum.saturating_add(0x80);
+
+    let code: c_int = match log_fault_to_file_and_stdout(signum) {
+        // Signal numbers are positive integers.
+        Ok(()) => signum_code,
+
+        // map error to error-codes.
+        Err(err) => match err {
+            FaultHandlerErr::Open => 2,
+            FaultHandlerErr::Write => 3,
+            FaultHandlerErr::FSync => 4,
+            FaultHandlerErr::Reentrancy => 5,
         },
     };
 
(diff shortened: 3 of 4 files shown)


hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[`mullvad-daemon/src/logging.rs` line 121 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OB_YKWB8aaX7WzCOj0u:-OB_YKWB8aaX7WzCOj0v:bb7pa6j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f18c5e16c79371e4cfdc504dd0362f0e4c4f5b29/mullvad-daemon/src/logging.rs#L121)):*
> ```XML
>         #[cfg(unix)]
>         crate::exception_logging::set_log_file(
>             std::ffi::CString::new(log_file.as_os_str().as_encoded_bytes()).unwrap(),
> ```

todo

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: 0 of 4 files reviewed, 2 unresolved discussions

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 113 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OB_Y_CwDzO7D1u4VwZD:-OB_Y_CwDzO7D1u4VwZE:ba3sxwj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f18c5e16c79371e4cfdc504dd0362f0e4c4f5b29/mullvad-daemon/src/exception_logging/unix.rs#L113)):*
> ```XML
> 
>     let code: c_int = match log_fault_to_file_and_stdout(signum) {
>         // Signal numbers are positive integers
> ```

todo: not sure if this is true? The exit code in practice appears to be e.g. `-11` for SIGSEGV.

Also: calling _exit() with a negative code will return `256 - code`, which is strange to say the least.

MarkusPettersson98 (commented):
Very nice work :yellow_heart: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-:-OB_noRi5Tj9IT1FsNd2:bgykzai)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @hulthe)

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 177 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OB_kHN0ECjz6X6LdKAL:-OB_kHN0ECjz6X6LdKAM:bz21db4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f18c5e16c79371e4cfdc504dd0362f0e4c4f5b29/mullvad-daemon/src/exception_logging/unix.rs#L177)):*
> ```XML
>         Err(_) => {
>             // SIGNAL-SAFETY: formatting an i32 is signal-safe.
>             writeln!(w, "Signal handler triggered by unknown signal: {signum}")?;
> ```

Could you convince me since I do not carry the same intuition as you; is formatting an i32 signal-safe because it fits into the size of a pointer? Or is it because it is allocated on the stack? etc



_Code quote:_
```XML
            // SIGNAL-SAFETY: formatting an i32 is signal-safe.
            writeln!(w, "Signal handler triggered by unknown signal: {signum}")?;
```

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @MarkusPettersson98)

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 177 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OB_kHN0ECjz6X6LdKAL:-OB_zU5y2QF30SVozRZa:b-jkphrb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f18c5e16c79371e4cfdc504dd0362f0e4c4f5b29/mullvad-daemon/src/exception_logging/unix.rs#L177)):*

Could you convince me since I do not carry the same intuition as you; is formatting an i32 signal-safe because it fits into the size of a pointer? Or is it because it is allocated on the stack? etc
</blockquote></details>

That's a very good question, and not easy to answer.

There's no concept in Rust of signal-safety, so the only way to be sure is to look at the implementation. In this case, the implementation of [Display for i32](https://doc.rust-lang.org/src/core/fmt/num.rs.html#210). An implementation is, of course, subject to change.

Formally: A function is signal-safe if it does not interact with any global shared mutable state (e.g. an allocator). Less formally: One *could* intuit (without looking at the source code) that property *should* hold for i32::fmt, since it would be pretty crazy for such a function to allocate, or to have any other global state, but there's no guarantees of that.

In summary: I looked at the source of i32::fmt, and it looks safe to me. That might change in the future, but I doubt it will. The alternative is to not print the signal number at all, or to format it ourselves.

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @hulthe)

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 177 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OB_kHN0ECjz6X6LdKAL:-OBa6qaeBKDK6wpcNtk_:bvtq4c9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f18c5e16c79371e4cfdc504dd0362f0e4c4f5b29/mullvad-daemon/src/exception_logging/unix.rs#L177)):*

That's a very good question, and not easy to answer.

There's no concept in Rust of signal-safety, so the only way to be sure is to look at the implementation. In this case, the implementation of [Display for i32](https://doc.rust-lang.org/src/core/fmt/num.rs.html#210). An implementation is, of course, subject to change.

Formally: A function is signal-safe if it does not interact with any global shared mutable state (e.g. an allocator). Less formally: One *could* intuit (without looking at the source code) that property *should* hold for i32::fmt, since it would be pretty crazy for such a function to allocate, or to have any other global state, but there's no guarantees of that.

In summary: I looked at the source of i32::fmt, and it looks safe to me. That might change in the future, but I doubt it will. The alternative is to not print the signal number at all, or to format it ourselves.
</blockquote></details>

Great answer, thanks :+1:

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 113 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OB_Y_CwDzO7D1u4VwZD:-OBaBBqg9yj1iRTlWflJ:b-wq9j4a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f18c5e16c79371e4cfdc504dd0362f0e4c4f5b29/mullvad-daemon/src/exception_logging/unix.rs#L113)):*

todo: not sure if this is true? The exit code in practice appears to be e.g. `-11` for SIGSEGV.

Also: calling _exit() with a negative code will return `256 - code`, which is strange to say the least.
</blockquote></details>

Made the signum-codes positive instead.

MarkusPettersson98 (dismissed):
> Reviewed 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 113 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OB_Y_CwDzO7D1u4VwZD:-OBdYzWm-byBhMqNwFlV:b-dxnmqx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f18c5e16c79371e4cfdc504dd0362f0e4c4f5b29/mullvad-daemon/src/exception_logging/unix.rs#L113)):*

Made the signum-codes positive instead.
</blockquote></details>

:smile: :ok_hand:

pinkisemils (dismissed):
:lgtm:

> Reviewed 1 of 4 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-:-OBkv11f3Jgo9opvRK7F:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @hulthe, @MarkusPettersson98, and @pinkisemils)

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 2 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OC4Uwu60TAWdqQeGc2q:-OC4Uwu60TAWdqQeGc2r:bk7tcn6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d80f9f851caac261e33a7df2f0cd3d807f5bcd6a/mullvad-daemon/src/exception_logging/unix.rs#L2)):*
> ```XML
> //! Install signal handlers to catch critical program faults and log them. See [`enable`].
> #![warn(clippy::undocumented_unsafe_blocks)]
> ```

This can probably be removed? I fail to make it generate any warnings when I remove it, so I suspect it's a leftover from development? Either way, we should try to avoid suppressing linting as much as possible.

MarkusPettersson98 (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: all files reviewed, 1 unresolved discussion (waiting on @hulthe)

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 2 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OC4Uwu60TAWdqQeGc2q:-OC8UEp18ftQb9U7W54C:bi9dc1i) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d80f9f851caac261e33a7df2f0cd3d807f5bcd6a/mullvad-daemon/src/exception_logging/unix.rs#L2)):*

This can probably be removed? I fail to make it generate any warnings when I remove it, so I suspect it's a leftover from development? Either way, we should try to avoid suppressing linting as much as possible.
</blockquote></details>

undocumented_unsafe_blocks is actually disabled by default, this _enables_ that lint.

Related issue: [DES-1459](https://linear.app/mullvad/issue/DES-1459)

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 2 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OC4Uwu60TAWdqQeGc2q:-OC922gE8zB26fTxj2n0:byrhr0o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d80f9f851caac261e33a7df2f0cd3d807f5bcd6a/mullvad-daemon/src/exception_logging/unix.rs#L2)):*

undocumented_unsafe_blocks is actually disabled by default, this _enables_ that lint.

Related issue: [DES-1459](https://linear.app/mullvad/issue/DES-1459)
</blockquote></details>

Aaah. First time someone fine grained adds more stricter linting maybe! So I just assumed you made it weaker :joy: We should ideally have this be warn/deny in the entire workspace. But that's a large body of work that is completely unrelated to this PR :) I endorse stricter linting so please go ahead with this

pinkisemils (approved):
> Reviewed 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @hulthe, @MarkusPettersson98, and @pinkisemils)

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 64 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OCIEzaw2OZBL3O7js-9:-OCIEzaw2OZBL3O7js-A:bvm88ke) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fcb3ac7fc26f21f6721460042e73b617b8d3ba61/mullvad-daemon/src/exception_logging/unix.rs#L64)):*
> ```XML
> 
>     INIT_ONCE.call_once(|| {
>         if let Ok(override_backtrace) = env::var(ENABLE_BACKTRACE_VAR).map(|v| v == "1") {
> ```

All other boolean environment variables that control the daemon has the enable logic ` != "0"`. Can we keep this so it's consistent? There should only be two possible values, but I'd like all boolean control mechanisms to react the same to setting the value to `"sausage"` if we can.

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @faern, @MarkusPettersson98, and @pinkisemils)

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 64 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OCIEzaw2OZBL3O7js-9:-OCIKiao6n9_KROJ307T:bo6vx3s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fcb3ac7fc26f21f6721460042e73b617b8d3ba61/mullvad-daemon/src/exception_logging/unix.rs#L64)):*

All other boolean environment variables that control the daemon has the enable logic ` != "0"`. Can we keep this so it's consistent? There should only be two possible values, but I'd like all boolean control mechanisms to react the same to setting the value to `"sausage"` if we can.
</blockquote></details>

My reasoning is that potentially harmful options should fail safe, and default to false if the input is ambiguous. It shouldn't (imo) be possible that `HARMFUL_OPTION=false` actually _enables_ the harmful option. I'll change it if you disagree though.

Our documentation only defines our env variables in terms of "1" and "0" anyway, so there's no deviation from spec.

MarkusPettersson98 (approved):
> Reviewed 1 of 1 files at r7, 2 of 3 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

hulthe (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @faern and @MarkusPettersson98)

___
*[`mullvad-daemon/src/exception_logging/unix.rs` line 64 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172#-OCIEzaw2OZBL3O7js-9:-OCbt9BG2SNy3IXFW2iP:bh7848x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fcb3ac7fc26f21f6721460042e73b617b8d3ba61/mullvad-daemon/src/exception_logging/unix.rs#L64)):*

My reasoning is that potentially harmful options should fail safe, and default to false if the input is ambiguous. It shouldn't (imo) be possible that `HARMFUL_OPTION=false` actually _enables_ the harmful option. I'll change it if you disagree though.

Our documentation only defines our env variables in terms of "1" and "0" anyway, so there's no deviation from spec.
</blockquote></details>

Went ahead and changed it back

faern (approved):
> Reviewed 2 of 3 files at r12.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7172)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7174, opened 2024-11-13
Ensure leak test are running

Rawa:


Pururun (approved):
:lgtm:

> Reviewed 1 of 3 files at r1, 7 of 7 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7174#-:-OBdru4S6qM3cMcYCvnh:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Pururun (dismissed):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7174#-:-OByIlBiAiHamfo4aZst:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7174)* status: 8 of 9 files reviewed, 4 unresolved discussions (waiting on @albin-mullvad, @niklasberglund, and @Pururun)

___
*[`android/scripts/run-instrumented-tests.sh` line 136 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7174#-OBycdk83cNR717p--Ah:-OBygOWH-BfM4Sdswtyz:bwq4zbo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f33d84b71f058e83a8b4e0bd474af5c6a1968a3/android/scripts/run-instrumented-tests.sh#L136)):*

Should this issue be considered addressed by this PR? https://linear.app/mullvad/issue/DROID-1413/inconsistent-naming-convention-in-properties-files

I haven't researched and verified it but lowercase:ing the names seems like a good convention to me
</blockquote></details>

I talked about this internally with @albin-mullvad when you were away, we also have another issue:
https://linear.app/mullvad/issue/DROID-1430/rework-localproperties-and-e2eproperties
Both this and the other one should be addressed in an other PR. The current implementation aligns with what we have right now. 
___
*[`android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/extension/UiAutomatorExtensions.kt` line 20 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7174#-OByTPi411vLEg8ypL4k:-OBygfyF3Z2I-gZay4us:bot8saq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f33d84b71f058e83a8b4e0bd474af5c6a1968a3/android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/extension/UiAutomatorExtensions.kt#L20)):*

Is this function needed or could `findObjectWithTimeout` be used? It does the same thing but with a `null` check as well
</blockquote></details>

The `findObjectWithTimeout` requires an object to exist otherwise it will error out. This one will check if a item exist without requiring it to exist. In the case where it is used we check if there are Vpn profiles with a timeout. The list could be empty because of 2 reasons:
1. View has not been fully loaded
2. There is no VPN profiles at all
___
*[`android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/rule/ForgetAllVpnAppsInSettingsTestRule.kt` line 30 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7174#-OByay07BOnOblwQi1ZW:-OByhDBwBJnjl2f0Wj9s:b8db42d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f33d84b71f058e83a8b4e0bd474af5c6a1968a3/android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/rule/ForgetAllVpnAppsInSettingsTestRule.kt#L30)):*

I think we should have a `findObjectsWithTimeout` function which do this. We have `findObjectWithTimeout` singular but not plural.
</blockquote></details>

I did that initially but adding a `findObjectsWithTimeout` is not straight forward, because of the cases mentioned above. E.g `findObjectsWithTimeout` can not fail a test because finding 0 items or 5 items are both legitimate cases. Thus until we know better what we want I'd prefer us holding of until we know the use cases more.
___
*[`android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/rule/ForgetAllVpnAppsInSettingsTestRule.kt` line 48 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7174#-OByc23TCpcShm7aP29T:-OByj4tZ1CitbvgjHl8Z:b52mw7m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f33d84b71f058e83a8b4e0bd474af5c6a1968a3/android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/rule/ForgetAllVpnAppsInSettingsTestRule.kt#L48)):*

I wonder if `org.junit.jupiter.api.fail` should be used here or if `error` produces same result. I've been using `fail` for failing with messages elsewhere.
</blockquote></details>

:+1: Yeah, fail would be more accurate I believe, fixed.

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7174)* status: 8 of 9 files reviewed, 4 unresolved discussions (waiting on @albin-mullvad, @niklasberglund, and @Pururun)

___
*[`android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/extension/UiAutomatorExtensions.kt` line 20 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7174#-OByTPi411vLEg8ypL4k:-OBykb32-V8GqNlIQApN:b-t2gd43) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f33d84b71f058e83a8b4e0bd474af5c6a1968a3/android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/extension/UiAutomatorExtensions.kt#L20)):*

The `findObjectWithTimeout` requires an object to exist otherwise it will error out. This one will check if a item exist without requiring it to exist. In the case where it is used we check if there are Vpn profiles with a timeout. The list could be empty because of 2 reasons:
1. View has not been fully loaded
2. There is no VPN profiles at all
</blockquote></details>

Same goes for looking at the "Forget VPN" and "Delete VPN", we shouldn't error out because "Forget VPN" doesn't exist because we might be on a device that has the "Delete VPN" button instead.

niklasberglund (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7174#-:-OBylJlW4ncMUEDH-wkJ:bnfp4nl)* status: 8 of 9 files reviewed, all discussions resolved (waiting on @Pururun)

PR #7176, opened 2024-11-14
Remove the cargo audit CI job, since the same is covered by cargo deny

faern:
We currently check vulnerability databases with three separate tools in CI for Rust code. This makes maintaining the list of ignored vulns messy and complicated. It also wastes CI time for no good reason.

`cargo deny` does everything that `cargo audit` does, and more. So we'll keep that one for now.

We should later check if we can make `cargo deny` not check CVEs and only check for yanking. Since `osv-scanner` already scans the dependencies for RustSec advisories (and more databases)

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7176)


MarkusPettersson98 (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7176)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7199, opened 2024-11-19
Apply the obfuscation port to the entry configuration only

rablador:
Currently, the obfuscation port is applied to the exit configuration at all times. Instead, it should be applied to the entry configuration when using multihop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199)

### DIFF
--- a/ios/MullvadREST/ApiHandlers/ServerRelaysResponse.swift
+++ b/ios/MullvadREST/ApiHandlers/ServerRelaysResponse.swift
@@ -37,7 +37,7 @@ extension REST {
         public var daita: Bool?
 
         public func override(ipv4AddrIn: IPv4Address?) -> Self {
-            return BridgeRelay(
+            BridgeRelay(
                 hostname: hostname,
                 active: active,
                 owned: owned,
@@ -65,7 +65,7 @@ extension REST {
         public let shadowsocksExtraAddrIn: [String]?
 
         public func override(ipv4AddrIn: IPv4Address?, ipv6AddrIn: IPv6Address?) -> Self {
-            return ServerRelay(
+            ServerRelay(
                 hostname: hostname,
                 active: active,
                 owned: owned,
@@ -82,7 +82,7 @@ extension REST {
         }
 
         public func override(daita: Bool) -> Self {
-            return ServerRelay(
+            ServerRelay(
                 hostname: hostname,
                 active: active,
                 owned: owned,
--- a/ios/MullvadREST/Relay/MultihopDecisionFlow.swift
+++ b/ios/MullvadREST/Relay/MultihopDecisionFlow.swift
@@ -22,6 +22,7 @@ protocol MultihopDecisionFlow {
 struct OneToOne: MultihopDecisionFlow {
     let next: MultihopDecisionFlow?
     let relayPicker: RelayPicking
+
     init(next: (any MultihopDecisionFlow)?, relayPicker: RelayPicking) {
         self.next = next
         self.relayPicker = relayPicker
@@ -47,10 +48,11 @@ struct OneToOne: MultihopDecisionFlow {
             throw NoRelaysSatisfyingConstraintsError(.entryEqualsExit)
         }
 
-        let exitMatch = try relayPicker.findBestMatch(from: exitCandidates)
+        let exitMatch = try relayPicker.findBestMatch(from: exitCandidates, obfuscate: false)
         let entryMatch = try relayPicker.findBestMatch(
             from: entryCandidates,
-            closeTo: daitaAutomaticRouting ? exitMatch.location : nil
+            closeTo: daitaAutomaticRouting ? exitMatch.location : nil,
+            obfuscate: true
         )
 
         return SelectedRelays(entry: entryMatch, exit: exitMatch, retryAttempt: relayPicker.connectionAttemptCount)
@@ -95,8 +97,8 @@ struct OneToMany: MultihopDecisionFlow {
                 .pick(entryCandidates: entryCandidates, exitCandidates: exitCandidates, daitaAutomaticRouting: true)
         }
 
-        let entryMatch = try multihopPicker.findBestMatch(from: entryCandidates)
-        let exitMatch = try multihopPicker.exclude(relay: entryMatch, from: exitCandidates)
+        let entryMatch = try multihopPicker.findBestMatch(from: entryCandidates, obfuscate: true)
+        let exitMatch = try multihopPicker.exclude(relay: entryMatch, from: exitCandidates, obfuscate: false)
 
         return SelectedRelays(entry: entryMatch, exit: exitMatch, retryAttempt: relayPicker.connectionAttemptCount)
     }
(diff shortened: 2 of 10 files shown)


buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199)* status: all files reviewed, 1 unresolved discussion (waiting on @rablador)

___
*[`ios/MullvadREST/Relay/RelayPicking.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199#-OC8iS_d8QKR2LRfF0Kl:-OC8iS_d8QKR2LRfF0Km:bub0zj5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7432679c906a86dc70ac25f1833cf73ee99d91db/ios/MullvadREST/Relay/RelayPicking.swift#L25)):*
> ```Swift
>         from candidates: [RelayWithLocation<REST.ServerRelay>],
>         closeTo location: Location? = nil,
>         obfuscate: Bool
> ```

I think this conveys the wrong information, and is misleading.
It's effectively saying "use obfuscation" when instead it should be saying "use the obfuscated port if present" since this doesn't decide whether obfuscation is ultimately used or not. 

How about we rename it `shouldUseObfuscatedPort` ?

rablador (commented):
Added fixes for issue https://linear.app/mullvad/issue/IOS-948/fix-port-selection-for-shadowsocks as well.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199#-:-OCJHZko1GsqsNlpNyYB:b-rxwbaq)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/MullvadREST/Relay/RelayPicking.swift` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199#-OC8iS_d8QKR2LRfF0Kl:-OCJHTc1EtYjLfeFKs24:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7432679c906a86dc70ac25f1833cf73ee99d91db/ios/MullvadREST/Relay/RelayPicking.swift#L25)):*

I think this conveys the wrong information, and is misleading.
It's effectively saying "use obfuscation" when instead it should be saying "use the obfuscated port if present" since this doesn't decide whether obfuscation is ultimately used or not. 

How about we rename it `shouldUseObfuscatedPort` ?
</blockquote></details>

Done.

buggmagnet (commented):
Looks good, but Unit Tests are broken now :melting_face: 

> Reviewed 4 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199#-:-OCXlFu60O8m-gtpB2-Y:b-tc9rrp)* status: 9 of 10 files reviewed, all discussions resolved

rablador (commented):
Fixed!

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199#-:-OCY7wIA0jsEfBJfVDo1:bt2vr0y)* status: 9 of 10 files reviewed, all discussions resolved (waiting on @buggmagnet)

buggmagnet (dismissed):
:lgtm:

> Reviewed 1 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199#-:-OCYlqMuD9CXgZbLIYc-:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

acb-mv (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199)* status: all files reviewed, 4 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadREST/Relay/RelayPicking.swift` line 14 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199#-OCbcdEQDPCK6FPp-BPK:-OCbx3Id2Wp9CH8feOQF:b1gpe3o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b9cfdf5b533366fb2c90b215b4b4ffa0ea37e90b/ios/MullvadREST/Relay/RelayPicking.swift#L14)):*

nit : that wouldn't be better to integrate all properties in another structure. since the it's growing up.
</blockquote></details>

I think it's fine at this point, but maybe in the future. Adding a wrapper adds an extra layer that might make things harder to follow.
___
*[`ios/MullvadREST/Relay/RelayPicking.swift` line 15 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199#-OCbeiEU3IDxx-9APvqB:-OCbwfP6DkObfm8Wqo1x:b-nuiqzy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b9cfdf5b533366fb2c90b215b4b4ffa0ea37e90b/ios/MullvadREST/Relay/RelayPicking.swift#L15)):*

there no need to make it an exposed computed property out of the object. instead of using that property I suggest to use ` obfuscation.relays` directly
</blockquote></details>

I thought it would be clearer to specify a generic `relays` variable. It might not be obvious for a reader that the filtered relays from the obfuscation result is the thing you want.
___
*[`ios/MullvadREST/Relay/RelayPicking.swift` line 54 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199#-OCY91DrARcYpRSxEpan:-OCbweSk7ANwwBF2q6E4:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b9cfdf5b533366fb2c90b215b4b4ffa0ea37e90b/ios/MullvadREST/Relay/RelayPicking.swift#L54)):*

I suggest to add some comment here about the process we are doing when the port is out of the range.
</blockquote></details>

Done.
___
*[`ios/MullvadREST/Relay/RelayPicking.swift` line 89 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199#-OCbe2vI2bdoJZ_yylqI:-OCbv9ka8d6nYOwT04qE:bo062cy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b9cfdf5b533366fb2c90b215b4b4ffa0ea37e90b/ios/MullvadREST/Relay/RelayPicking.swift#L89)):*

what is the reason behind setting that true in single hop?
</blockquote></details>

It's not very obvious here, I agree, but `useObfuscatedPort` means that we should use the obfuscated port if there's one available. In singlehop there's only one (exit) relay, so obfuscation should always be applied in that case. In multihop we only set `useObfuscatedPort` to true if it's the entry relay.

mojganii (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7199)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7212, opened 2024-11-21
Remove split tunnel driver path check

dlon:
The condition was supposed to replace the driver service if the path was wrong, but it never actually worked because: (1) it is negated, (2) the service path contains a prefix `\??\` which is not stripped:

```
expected_syspath == "C:\\Users\\abc\\mullvadvpn-app\\target\\debug\\mullvad-split-tunnel.sys"
get_driver_binpath(...) == Ok("\\??\\C:\\Program Files\\Mullvad VPN\\resources\\mullvad-split-tunnel.sys")
```

PR just removes it since it hasn't caused any issues.


MarkusPettersson98 (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7212)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7233, opened 2024-11-25
Fix log redaction loading issue

mojganii:
this PR take log redaction out of main thread which caused blocking UI when working around it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)

### DIFF
--- a/ios/CHANGELOG.md
+++ b/ios/CHANGELOG.md
@@ -29,6 +29,7 @@ Line wrap the file at 100 chars.                                              Th
 ## [2024.10 - 2024-11-20]
 ### Fixed
 - Removed deadlock when losing connectivity without entering offline state.
+- Improved log reporting.
 
 ## [2024.9 - 2024-11-07]
 ### Added
--- a/ios/MullvadSettings/SettingsManager.swift
+++ b/ios/MullvadSettings/SettingsManager.swift
@@ -47,7 +47,10 @@ public enum SettingsManager {
 
     public static func getLastUsedAccount() throws -> String {
         let data = try store.read(key: .lastUsedAccount)
-        return String(decoding: data, as: UTF8.self)
+        guard let result = String(bytes: data, encoding: .utf8) else {
+            throw StringDecodingError(data: data)
+        }
+        return result
     }
 
     public static func setLastUsedAccount(_ string: String?) throws {
@@ -218,21 +221,5 @@ public struct UnsupportedSettingsVersionError: LocalizedError {
         Stored settings version was not the same as current version, \
         stored version: \(storedVersion), current version: \(currentVersion)
         """
-    }
-}
-
-public struct StringDecodingError: LocalizedError {
-    public let data: Data
-
-    public var errorDescription: String? {
-        "Failed to decode string from data."
-    }
-}
-
-public struct StringEncodingError: LocalizedError {
-    public let string: String
-
-    public var errorDescription: String? {
-        "Failed to encode string into data."
     }
 }
--- a/ios/MullvadTypes/StringConversionError.swift
+++ b/ios/MullvadTypes/StringConversionError.swift
@@ -1,0 +1,31 @@
+//
+//  StringDecodingError.swift
+//  MullvadVPN
+//
+//  Created by Mojgan on 2024-12-02.
+//  Copyright © 2024 Mullvad VPN AB. All rights reserved.
+//
+
+public struct StringDecodingError: LocalizedError {
+    public let data: Data
+
+    public init(data: Data) {
+        self.data = data
+    }
+
+    public var errorDescription: String? {
+        "Failed to decode string from data."
+    }
+}
+
+public struct StringEncodingError: LocalizedError {
+    public let string: String
+
+    public init(string: String) {
+        self.string = string
+    }
+
+    public var errorDescription: String? {
+        "Failed to encode string into data."
+    }
+}
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -871,6 +871,7 @@
 		F0164EC32B4C49D30020268D /* ShadowsocksLoaderStub.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0164EC22B4C49D30020268D /* ShadowsocksLoaderStub.swift */; };
 		F0164ED12B4F2DCB0020268D /* AccessMethodIterator.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0164ED02B4F2DCB0020268D /* AccessMethodIterator.swift */; };
 		F01DAE332C2B032A00521E46 /* RelaySelection.swift in Sources */ = {isa = PBXBuildFile; fileRef = F01DAE322C2B032A00521E46 /* RelaySelection.swift */; };
+		F022EBA62CF0C6AE009484B9 /* ConsolidatedApplicationLog.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5871FB95254ADE4E0051A0A4 /* ConsolidatedApplicationLog.swift */; };
 		F028A56A2A34D4E700C0CAA3 /* RedeemVoucherViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F028A5692A34D4E700C0CAA3 /* RedeemVoucherViewController.swift */; };
 		F028A56C2A34D8E600C0CAA3 /* AddCreditSucceededViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F028A56B2A34D8E600C0CAA3 /* AddCreditSucceededViewController.swift */; };
 		F02F41A02B9723AF00625A4F /* AddLocationsViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = F02F419A2B9723AE00625A4F /* AddLocationsViewController.swift */; };
@@ -946,6 +947,8 @@
 		F09D04C12AF39EA2003D4F89 /* OutgoingConnectionService.swift in Sources */ = {isa = PBXBuildFile; fileRef = F09D04BC2AEBB7C5003D4F89 /* OutgoingConnectionService.swift */; };
 		F0A086902C22D6A700BF83E7 /* TunnelSettingsStrategyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0A0868F2C22D6A700BF83E7 /* TunnelSettingsStrategyTests.swift */; };
 		F0A1638A2C47B77300592300 /* ServerRelaysResponse+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0ACE3342BE51745006D5333 /* ServerRelaysResponse+Stubs.swift */; };
+		F0A7EBB22CEF6C79005BB671 /* ConsolidatedApplicationLogTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0A7EBB12CEF6C79005BB671 /* ConsolidatedApplicationLogTests.swift */; };
+		F0A7EBB62CF092CC005BB671 /* ApplicationConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58BFA5CB22A7CE1F00A6173D /* ApplicationConfiguration.swift */; };
 		F0ACE30D2BE4E478006D5333 /* MullvadMockData.h in Headers */ = {isa = PBXBuildFile; fileRef = F0ACE30A2BE4E478006D5333 /* MullvadMockData.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		F0ACE3102BE4E478006D5333 /* MullvadMockData.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = F0ACE3082BE4E478006D5333 /* MullvadMockData.framework */; };
 		F0ACE3112BE4E478006D5333 /* MullvadMockData.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = F0ACE3082BE4E478006D5333 /* MullvadMockData.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
@@ -964,6 +967,7 @@
 		F0ACE3372BE517F1006D5333 /* ServerRelaysResponse+Stubs.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0ACE3342BE51745006D5333 /* ServerRelaysResponse+Stubs.swift */; };
 		F0ADC3722CD3AD1600A1AD97 /* ChipCollectionView.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0ADC3712CD3AD1600A1AD97 /* ChipCollectionView.swift */; };
 		F0ADC3742CD3C47400A1AD97 /* ChipFlowLayout.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0ADC3732CD3C47400A1AD97 /* ChipFlowLayout.swift */; };
+		F0ADF1CD2CFDFF3100299F09 /* StringConversionError.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0ADF1CC2CFDFF3100299F09 /* StringConversionError.swift */; };
 		F0B0E6972AFE6E7E001DC66B /* XCTest+Async.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0B0E6962AFE6E7E001DC66B /* XCTest+Async.swift */; };
 		F0B894EF2BF751C500817A42 /* RelayWithLocation.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0B894EE2BF751C500817A42 /* RelayWithLocation.swift */; };
 		F0B894F12BF751E300817A42 /* RelayWithDistance.swift in Sources */ = {isa = PBXBuildFile; fileRef = F0B894F02BF751E300817A42 /* RelayWithDistance.swift */; };
@@ -2166,12 +2170,14 @@
 		F09D04BF2AF39D63003D4F89 /* OutgoingConnectionServiceTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OutgoingConnectionServiceTests.swift; sourceTree = "<group>"; };
 		F0A0868F2C22D6A700BF83E7 /* TunnelSettingsStrategyTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TunnelSettingsStrategyTests.swift; sourceTree = "<group>"; };
 		F0A163882C47B46300592300 /* SingleHopEphemeralPeerExchangerTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SingleHopEphemeralPeerExchangerTests.swift; sourceTree = "<group>"; };
+		F0A7EBB12CEF6C79005BB671 /* ConsolidatedApplicationLogTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ConsolidatedApplicationLogTests.swift; sourceTree = "<group>"; };
 		F0ACE3082BE4E478006D5333 /* MullvadMockData.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; includeInIndex = 0; path = MullvadMockData.framework; sourceTree = BUILT_PRODUCTS_DIR; };
 		F0ACE30A2BE4E478006D5333 /* MullvadMockData.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = MullvadMockData.h; sourceTree = "<group>"; };
 		F0ACE32E2BE4EA8B006D5333 /* MockProxyFactory.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MockProxyFactory.swift; sourceTree = "<group>"; };
 		F0ACE3342BE51745006D5333 /* ServerRelaysResponse+Stubs.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = "ServerRelaysResponse+Stubs.swift"; sourceTree = "<group>"; };
 		F0ADC3712CD3AD1600A1AD97 /* ChipCollectionView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ChipCollectionView.swift; sourceTree = "<group>"; };
 		F0ADC3732CD3C47400A1AD97 /* ChipFlowLayout.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ChipFlowLayout.swift; sourceTree = "<group>"; };
+		F0ADF1CC2CFDFF3100299F09 /* StringConversionError.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StringConversionError.swift; sourceTree = "<group>"; };
 		F0B0E6962AFE6E7E001DC66B /* XCTest+Async.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "XCTest+Async.swift"; sourceTree = "<group>"; };
 		F0B894EE2BF751C500817A42 /* RelayWithLocation.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RelayWithLocation.swift; sourceTree = "<group>"; };
 		F0B894F02BF751E300817A42 /* RelayWithDistance.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RelayWithDistance.swift; sourceTree = "<group>"; };
@@ -2425,6 +2431,7 @@
 		440E9EF02BDA93CB00B1FD11 /* MullvadVPN */ = {
 			isa = PBXGroup;
 			children = (
+				F0A7EBB02CEF6C5F005BB671 /* Log */,
 				440E9EF62BDA957300B1FD11 /* Classes */,
 				440E9F002BDA997C00B1FD11 /* Extensions */,
 				440E9EFB2BDA97C600B1FD11 /* GeneralAPIs */,
@@ -2704,6 +2711,7 @@
 				5898D2AF2902A67C00EB5EBA /* RelayLocation.swift */,
 				581DA2722A1E227D0046ED47 /* RESTTypes.swift */,
 				58F1311427E0B2AB007AC5BC /* Result+Extensions.swift */,
+				F0ADF1CC2CFDFF3100299F09 /* StringConversionError.swift */,
 				A91614D02B108D1B00F416EB /* TransportLayer.swift */,
 				58E511E028DDB7F100B0BCDE /* WrappingError.swift */,
 			);
@@ -4205,6 +4213,14 @@
 			path = GeneralAPIs;
 			sourceTree = "<group>";
 		};
+		F0A7EBB02CEF6C5F005BB671 /* Log */ = {
+			isa = PBXGroup;
+			children = (
+				F0A7EBB12CEF6C79005BB671 /* ConsolidatedApplicationLogTests.swift */,
+			);
+			path = Log;
+			sourceTree = "<group>";
+		};
 		F0ACE3092BE4E478006D5333 /* MullvadMockData */ = {
 			isa = PBXGroup;
 			children = (
@@ -5378,6 +5394,7 @@
 				A9A5FA3B2ACB05910083449F /* UIMetrics.swift in Sources */,
 				58B07C182AEFDD6C00A09625 /* StoreTransactionLog.swift in Sources */,
 				A9A5FA382ACB05600083449F /* InputTextFormatter.swift in Sources */,
+				F022EBA62CF0C6AE009484B9 /* ConsolidatedApplicationLog.swift in Sources */,
 				A9A5FA372ACB052D0083449F /* ApplicationTarget.swift in Sources */,
 				A9A5F9E12ACB05160083449F /* AddressCacheTracker.swift in Sources */,
 				A9A5F9E22ACB05160083449F /* BackgroundTask.swift in Sources */,
@@ -5488,6 +5505,7 @@
 				A9A5FA282ACB05160083449F /* WgKeyRotation.swift in Sources */,
 				449872E42B7CB96300094DDC /* TunnelSettingsUpdateTests.swift in Sources */,
 				A9A5FA292ACB05160083449F /* AddressCacheTests.swift in Sources */,
+				F0A7EBB22CEF6C79005BB671 /* ConsolidatedApplicationLogTests.swift in Sources */,
 				A9B6AC182ADE8F4300F7802A /* MigrationManagerTests.swift in Sources */,
 				7A9BE5AB2B909A1700E2A7D0 /* LocationDataSourceProtocol.swift in Sources */,
 				A9A5FA2A2ACB05160083449F /* CoordinatesTests.swift in Sources */,
@@ -5515,6 +5533,7 @@
 				A9A5FA342ACB05160083449F /* StringTests.swift in Sources */,
 				7A52F96C2C17450C00B133B9 /* RelaySelectorWrapperTests.swift in Sources */,
 				A9A5FA352ACB05160083449F /* WgKeyRotationTests.swift in Sources */,
+				F0A7EBB62CF092CC005BB671 /* ApplicationConfiguration.swift in Sources */,
 				7AB4CCB92B69097E006037F5 /* IPOverrideTests.swift in Sources */,
 				A9A5FA362ACB05160083449F /* TunnelManagerTests.swift in Sources */,
 			);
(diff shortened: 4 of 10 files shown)


rablador (changes requested):
> Reviewed 6 of 6 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/View controllers/ProblemReport/ProblemReportInteractor.swift` line 29 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCbqvxbDxnvIGeeEWDF:-OCbqvxbDxnvIGeeEWDG:b-7thij7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e625845ef602cace1c43ed0ed9b20932eb1065d/ios/MullvadVPN/View%20controllers/ProblemReport/ProblemReportInteractor.swift#L29)):*
> ```Swift
> 
>     func loadLogFiles(completion: @escaping () -> Void) {
>         DispatchQueue
> ```

Is this queueu necessary when `consolidatedLog.addLogFiles` already runs in a concurrent queue? Same for the calls to `consolidatedLog.string` below.
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCbnGRG6RRNZnj5gzbw:-OCbnGRG6RRNZnj5gzbx:b-fqfgd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e625845ef602cace1c43ed0ed9b20932eb1065d/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L55)):*
> ```Swift
> 
>     func addLogFiles(fileURLs: [URL], completion: @escaping () -> Void = {}) {
>         logQueue.async(flags: .barrier) {
> ```

Why do we use .barrier?
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 113 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCbod2s5et2cP3l0PqR:-OCbod2s5et2cP3l0PqS:b-prol3p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e625845ef602cace1c43ed0ed9b20932eb1065d/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L113)):*
> ```Swift
>         let redactedPath = redact(string: path)
> 
>         if let lossyString = readFileLossy(path: path, maxBytes: bufferSize) {
> ```

Why getting rid of the gobal var? `.logMaximumFileSize` ensures all log related activities use the same size.

buggmagnet (changes requested):
> Reviewed 2 of 6 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 8 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift` line 28 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCcfKPT3ccYLfFgXZmP:-OCcfKPT3ccYLfFgXZmQ:bfzsx5w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b97d84a542913cd2007395a17eaa9e2ca4aa39c/ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift#L28)):*
> ```Swift
> 
>     override func tearDownWithError() throws {
>         super.tearDown()
> ```

We should call the correct `super` function, `tearDownWithError` here 
___
*[`ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift` line 39 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCcg1WG6_-FV-_9X97K:-OCcg1WG6_-FV-_9X97L:b-ql9ntg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b97d84a542913cd2007395a17eaa9e2ca4aa39c/ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift#L39)):*
> ```Swift
>     func testAddLogFiles() {
>         var string = ""
>         let expectation = self.expectation(description: "Log files added")
> ```

nit
no need for `self` here and below
___
*[`ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift` line 80 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCcwlS09I2CgmucIzY9:-OCcwlS09I2CgmucIzYA:bko9nxw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b97d84a542913cd2007395a17eaa9e2ca4aa39c/ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift#L80)):*
> ```Swift
>     }
> 
>     func testAddLogFilesRaceCondition() {
> ```

We discussed this offline, and decided to remove this test and `testReportedStringRaceCondition` as it doesn't really test against race conditions.
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 133 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCc64zi2DTaU8RNkFbY:-OCc64zj3vXtZeqGQ5mo:b-l7m0e3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/795ce077aa11fbaaa8b9e05e553e42820750df42/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L133)):*
> ```Swift
>             String(decoding: data, as: UTF8.self)
>                 .drop { ch in
>                     // Drop leading replacement characters produced when decoding data
> ```

That comment was useful, we should leave it
___
*[`ios/MullvadVPN/View controllers/ProblemReport/ProblemReportInteractor.swift` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCccF9G-6gxZEU0ncTu:-OCccF9HBW1bVjZjUHEK:b-rp1pcl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b97d84a542913cd2007395a17eaa9e2ca4aa39c/ios/MullvadVPN/View%20controllers/ProblemReport/ProblemReportInteractor.swift#L30)):*
> ```Swift
>     func loadLogFiles(completion: @escaping () -> Void) {
>         DispatchQueue
>             .global()
> ```

We don't need to dispatch on the `global` queue here since we're already doing this in a private concurrent queue.
The same applies to all uses of `global` in this file. 

If we absolutely need to dispatch off the UI thread here, the `ProblemReportInteractor` class can use a private dispatch queue.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: 2 of 7 files reviewed, 11 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 55 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCbnGRG6RRNZnj5gzbw:-OCd8JnD3nsaYnXjLEm1:b-p5f5t8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e625845ef602cace1c43ed0ed9b20932eb1065d/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L55)):*

Why do we use .barrier?
</blockquote></details>

Since the queue in concurrent, we need to be sure manipulation is executing in thread safe manner.
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 113 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCbod2s5et2cP3l0PqR:-OCd8UlXBR9wJ-SAxg-g:b6j52ig) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e625845ef602cace1c43ed0ed9b20932eb1065d/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L113)):*

Why getting rid of the gobal var? `.logMaximumFileSize` ensures all log related activities use the same size.
</blockquote></details>

I'll add it through dependency injection.
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 133 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCc64zi2DTaU8RNkFbY:-OCd8pbh7_tM6ITQmyjL:b-jpuc3t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/795ce077aa11fbaaa8b9e05e553e42820750df42/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L133)):*

That comment was useful, we should leave it
</blockquote></details>

It was unintentional.
___
*[`ios/MullvadVPN/View controllers/ProblemReport/ProblemReportInteractor.swift` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCccF9G-6gxZEU0ncTu:-OCd8yHi6_BHHhFaaX5F:bkajrg6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b97d84a542913cd2007395a17eaa9e2ca4aa39c/ios/MullvadVPN/View%20controllers/ProblemReport/ProblemReportInteractor.swift#L30)):*

We don't need to dispatch on the `global` queue here since we're already doing this in a private concurrent queue.
The same applies to all uses of `global` in this file. 

If we absolutely need to dispatch off the UI thread here, the `ProblemReportInteractor` class can use a private dispatch queue.
</blockquote></details>

We don't need that here. I made mistake.
___
*[`ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift` line 28 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCcfKPT3ccYLfFgXZmP:-OCd96hK7876smmGLqAW:b-kl8g3z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b97d84a542913cd2007395a17eaa9e2ca4aa39c/ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift#L28)):*

We should call the correct `super` function, `tearDownWithError` here
</blockquote></details>

good catch.
___
*[`ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift` line 39 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCcg1WG6_-FV-_9X97K:-OCd9Cqj4iPaP6Xc1wWJ:b-83sjo4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b97d84a542913cd2007395a17eaa9e2ca4aa39c/ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift#L39)):*

nit
no need for `self` here and below
</blockquote></details>

Nice.
___
*[`ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift` line 80 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCcwlS09I2CgmucIzY9:-OCdAEzr8mXbHnMrop7U:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b97d84a542913cd2007395a17eaa9e2ca4aa39c/ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift#L80)):*

We discussed this offline, and decided to remove this test and `testReportedStringRaceCondition` as it doesn't really test against race conditions.
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 36 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OChQF2P2ThvqVm2ahcJ:-OChQF2P2ThvqVm2ahcK:bct5kdo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/243b6b4c9ac9b9991077d3b29e23c0f85f43ba92/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L36)):*
> ```Swift
>     private let logQueue = DispatchQueue(
>         label: "com.mullvad.consolidation.logs.queue",
>         attributes: .concurrent
> ```

since there is I/O Operations and they can sometimes cause heavy process, I rather to user serial queue instead of concurrent one.
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 68 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OChQzvj83hnRNIPZK--:-OChQzvj83hnRNIPZK-0:b9hdkmp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/13c8ebcaddd8bfd2d4c36bc99cc25c518a18e8d6/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L68)):*
> ```Swift
>         logQueue.async(flags: .barrier) {
>             self.logs.append(LogAttachment(label: message, content: redactedError))
>             DispatchQueue.main.async {
> ```

switch to main for calling the completion, since `ConsolidatedApplicationLog` belongs and interacts with the main app.
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 76 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OChQYYXEdf0yB7bSruU:-OChQYYXEdf0yB7bSruV:bh6635w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/13c8ebcaddd8bfd2d4c36bc99cc25c518a18e8d6/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L76)):*
> ```Swift
>     var string: String {
>         var result = ""
>         logQueue.sync {
> ```

Since It's a serial queue anymore then for avoiding deadlock, I removed invoking `write`  function an the sync process.

rablador (commented):
> Reviewed 5 of 5 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 9 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift` line 23 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OChUVqW8y3OeFj-s2Su:-OChUVqX73pzAuL-Te03:b7o4nc0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/243b6b4c9ac9b9991077d3b29e23c0f85f43ba92/ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift#L23)):*
> ```Swift
>             redactCustomStrings: mockRedactStrings,
>             redactContainerPathsForSecurityGroupIdentifiers: mockSecurityGroupIdentifiers,
>             bufferSize: 65_536
> ```

How come we use a different buffer size than the default in prod code?

rablador (dismissed):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 3 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 157 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCmQaow9-SK7hDNFz3s:-OCmQaow9-SK7hDNFz3t:buyqbl9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/13c8ebcaddd8bfd2d4c36bc99cc25c518a18e8d6/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L157)):*
> ```Swift
>         }
> 
>         let data = fileHandle.readData(ofLength: Int(bufferSize))
> ```

This method seems deprecated, we should use its replacement
```swift
public func read(upToCount count: Int) throws -> Data?
```
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 160 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCmQ53uDD56yLelTvyR:-OCmQ53uDD56yLelTvyS:b-n8kzrz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/13c8ebcaddd8bfd2d4c36bc99cc25c518a18e8d6/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L160)):*
> ```Swift
>         let replacementCharacter = Character(UTF8.decode(UTF8.encodedReplacementCharacter))
>         let lossyString = String(
>             String(decoding: data, as: UTF8.self)
> ```

:policewoman: :warning: `Optional Data -> String Conversion Violation: Prefer failable `String(data:encoding:)` initializer when converting `Data` to `String` (optional_data_string_conversion)`
___
*[`ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift` line 38 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCmjvf_44QQ8fQbMngF:-OCmjvf_44QQ8fQbMngG:br92zbe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/13c8ebcaddd8bfd2d4c36bc99cc25c518a18e8d6/ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift#L38)):*
> ```Swift
>     }
> 
>     func testAddLogFiles() {
> ```

Somehow, this test is broken now

rablador (changes requested):
> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 4 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 77 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCo_KSE7Yn3Pn8S5ysA:-OCo_KSE7Yn3Pn8S5ysB:bvxr5zk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e1defbf28b825966f06412a5768bc393f89caa84/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L77)):*
> ```Swift
>         var result = ""
>         logQueue.sync {
>             if !logs.isEmpty {
> ```

This seems to be almost identical to `write()` below. Can we create a new function for the constructiom of `result`/`localOutput`?

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 157 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCmQaow9-SK7hDNFz3s:-OCn1D5u-G_L4QO4hoz9:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/13c8ebcaddd8bfd2d4c36bc99cc25c518a18e8d6/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L157)):*

This method seems deprecated, we should use its replacement
```swift
public func read(upToCount count: Int) throws -> Data?
```
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 160 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCmQ53uDD56yLelTvyR:-OCn1EIR84gX3JXEX3GA:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/13c8ebcaddd8bfd2d4c36bc99cc25c518a18e8d6/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L160)):*

:policewoman: :warning: `Optional Data -> String Conversion Violation: Prefer failable `String(data:encoding:)` initializer when converting `Data` to `String` (optional_data_string_conversion)`
</blockquote></details>

Done.
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 77 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCo_KSE7Yn3Pn8S5ysA:-OD5XHzF7wuJmaMvSonD:b1fumv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e1defbf28b825966f06412a5768bc393f89caa84/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L77)):*

This seems to be almost identical to `write()` below. Can we create a new function for the constructiom of `result`/`localOutput`?
</blockquote></details>

No, that can cause `deadlock`.
___
*[`ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift` line 38 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCmjvf_44QQ8fQbMngF:-OCnDatLEY1vjBSv3LZM:bt2vr1b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/13c8ebcaddd8bfd2d4c36bc99cc25c518a18e8d6/ios/MullvadVPNTests/MullvadVPN/Log/ConsolidatedApplicationLogTests.swift#L38)):*

Somehow, this test is broken now
</blockquote></details>

Fixed.

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 3 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 77 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCo_KSE7Yn3Pn8S5ysA:-OD5aRSE7bIjqfTgrmmJ:b2osp97) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9759cb30da613ee838dcff4d30e6564b1eb0b5c4/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L77)):*

No, that can cause `deadlock`.
</blockquote></details>

I agree with @rablador, we can create such a function

Note that you can also use `dispatchPrecondition(condition: .onQueue(logQueue))` (or the `.notOnQueue` variant) to make assertions about deadlocks and whatnot.

```swift
    func formatHeader() -> String {
        var result = ""
        result += "System information:\n"
        for (key, value) in metadata {
            result += "\(key.rawValue): \(value)\n"
        }
        result += "\n"
        for attachment in logs {
            result += "\(kLogDelimiter)\n"
            result += "\(attachment.label)\n"
            result += "\(kLogDelimiter)\n"
            result += "\(attachment.content)\n\n"
        }

        return result
    }

    var string: String {
        var result = ""
        logQueue.sync {
            if !logs.isEmpty {
                result = formatHeader()
            }
        }
        return result
    }

    func write(to stream: inout some TextOutputStream) {
        logQueue.sync {
            var localOutput = ""
            if !logs.isEmpty {
                localOutput = formatHeader()
                stream.write(localOutput)
            }
        }
    }
```
___
*[`ios/MullvadSettings/SettingsManager.swift` line 50 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OD5MQR05II5kYh5VNdG:-OD5MQR16_9rj4Npnvmz:bbfq156) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9759cb30da613ee838dcff4d30e6564b1eb0b5c4/ios/MullvadSettings/SettingsManager.swift#L50)):*
> ```Swift
>     public static func getLastUsedAccount() throws -> String {
>         let data = try store.read(key: .lastUsedAccount)
>         return String(bytes: data, encoding: .utf8) ?? ""
> ```

We should throw a `StringDecodingError` instead of silently failing and returning an empty string here.
___
*[`ios/MullvadVPN/View controllers/ProblemReport/ProblemReportInteractor.swift` line 24 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OD5UdM57_rJI_k7GChP:-OD5UdM57_rJI_k7GChQ:br1azt2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9759cb30da613ee838dcff4d30e6564b1eb0b5c4/ios/MullvadVPN/View%20controllers/ProblemReport/ProblemReportInteractor.swift#L24)):*
> ```Swift
>         self.tunnelManager = tunnelManager
>         self.consolidatedLog = ConsolidatedApplicationLog(
>             redactCustomStrings: [tunnelManager.deviceState.accountData?.number].compactMap { $0 },
> ```

I realize that this `redactCustomStrings` is not a good design. 
It only does something if the user is logged in when sending logs, otherwise this will be empty (because no `accountData.number` when logged out) and won't redact anything.

Thankfully, we have a `redactAccountNumber` function that uses a regex to redact account numbers, otherwise we'd have a pretty major problem. 

We should delete this false sense of privacy, and make sure the regex we use (for now) is solid instead.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadSettings/SettingsManager.swift` line 50 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OD5MQR05II5kYh5VNdG:-OD5jSvt9n-AA_Cgj6Vp:b5q5tp6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9759cb30da613ee838dcff4d30e6564b1eb0b5c4/ios/MullvadSettings/SettingsManager.swift#L50)):*

We should throw a `StringDecodingError` instead of silently failing and returning an empty string here.
</blockquote></details>

but the function `String(bytes: data, encoding: .utf8)`  doesn't throw any thing and the new one is  the same as what they were.
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 77 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCo_KSE7Yn3Pn8S5ysA:-OD5i2aL0v0v02AO9rSd:baakb38) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9759cb30da613ee838dcff4d30e6564b1eb0b5c4/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L77)):*

I agree with @rablador, we can create such a function

Note that you can also use `dispatchPrecondition(condition: .onQueue(logQueue))` (or the `.notOnQueue` variant) to make assertions about deadlocks and whatnot.

```swift
    func formatHeader() -> String {
        var result = ""
        result += "System information:\n"
        for (key, value) in metadata {
            result += "\(key.rawValue): \(value)\n"
        }
        result += "\n"
        for attachment in logs {
            result += "\(kLogDelimiter)\n"
            result += "\(attachment.label)\n"
            result += "\(kLogDelimiter)\n"
            result += "\(attachment.content)\n\n"
        }

        return result
    }

    var string: String {
        var result = ""
        logQueue.sync {
            if !logs.isEmpty {
                result = formatHeader()
            }
        }
        return result
    }

    func write(to stream: inout some TextOutputStream) {
        logQueue.sync {
            var localOutput = ""
            if !logs.isEmpty {
                localOutput = formatHeader()
                stream.write(localOutput)
            }
        }
    }
```
</blockquote></details>

`dispatchPrecondition(condition: .onQueue(logQueue))` can't prevent from probably deadlock, it's just for checking, I found another way to do that.
___
*[`ios/MullvadVPN/View controllers/ProblemReport/ProblemReportInteractor.swift` line 24 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OD5UdM57_rJI_k7GChP:-OD5k9siERORCtzOQ2-m:buwhpf9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9759cb30da613ee838dcff4d30e6564b1eb0b5c4/ios/MullvadVPN/View%20controllers/ProblemReport/ProblemReportInteractor.swift#L24)):*

I realize that this `redactCustomStrings` is not a good design. 
It only does something if the user is logged in when sending logs, otherwise this will be empty (because no `accountData.number` when logged out) and won't redact anything.

Thankfully, we have a `redactAccountNumber` function that uses a regex to redact account numbers, otherwise we'd have a pretty major problem. 

We should delete this false sense of privacy, and make sure the regex we use (for now) is solid instead.
</blockquote></details>

I know it's good to improve code when there is room to do that but It's beyond the current issue I opened PR for. since it's adding through DI then I changed that to accept nil and mange that in `ConsolidatedApplicationLog`

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadSettings/SettingsManager.swift` line 50 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OD5MQR05II5kYh5VNdG:-OD6BMjL8QlgUK-K6n77:bi23v4z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9759cb30da613ee838dcff4d30e6564b1eb0b5c4/ios/MullvadSettings/SettingsManager.swift#L50)):*

but the function `String(bytes: data, encoding: .utf8)`  doesn't throw any thing and the new one is  the same as what they were.
</blockquote></details>

That `init` can return a `nil` value, and this API is already declared as `throws` so the call site expects to handle errors.
In general, it's bad practice to hide errors like this, because it makes it significantly harder to debug  when things go wrong.
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 77 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCo_KSE7Yn3Pn8S5ysA:-OD6Bj-4Df38I6mevqi6:b-5vfik3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9759cb30da613ee838dcff4d30e6564b1eb0b5c4/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L77)):*

`dispatchPrecondition(condition: .onQueue(logQueue))` can't prevent from probably deadlock, it's just for checking, I found another way to do that.
</blockquote></details>

You're right that `dispatchPrecondition` does not prevent deadlocks, but that isn't the point of using that API.
The point is that whenever you cannot guarantee that a call happens on a specific dispatch queue, this will blow up and give you a stack trace, which will lead to the root cause.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: 5 of 10 files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadSettings/SettingsManager.swift` line 50 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OD5MQR05II5kYh5VNdG:-OD6jQxC32MgfqLfsmEX:b-ilyshw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9759cb30da613ee838dcff4d30e6564b1eb0b5c4/ios/MullvadSettings/SettingsManager.swift#L50)):*

That `init` can return a `nil` value, and this API is already declared as `throws` so the call site expects to handle errors.
In general, it's bad practice to hide errors like this, because it makes it significantly harder to debug  when things go wrong.
</blockquote></details>

I agree to throw exception when it's not able to cast data to intended encoding but I'm just telling it was the thing we had before :)
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 77 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-OCo_KSE7Yn3Pn8S5ysA:-OD6jE-N20OcH2on4gb6:b-8ron0n) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9759cb30da613ee838dcff4d30e6564b1eb0b5c4/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L77)):*

You're right that `dispatchPrecondition` does not prevent deadlocks, but that isn't the point of using that API.
The point is that whenever you cannot guarantee that a call happens on a specific dispatch queue, this will blow up and give you a stack trace, which will lead to the root cause.
</blockquote></details>

I agree but with the current implementation I suppose we don't need that anymore.

rablador (approved):
> Reviewed 5 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet)

buggmagnet (changes requested):
> Reviewed 1 of 5 files at r3, 1 of 3 files at r4, 1 of 5 files at r5, 1 of 2 files at r6, 3 of 5 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233)* status: all files reviewed, 2 unresolved discussions (waiting on @mojganii)

___
*[`ios/Shared/LaunchArguments.swift` line 51 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-ODAz2bO-nPiDxwWdz_4:-ODAz2bO-nPiDxwWdz_5:bevppsn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e7379c3b51cd0801629e0ac2c4267babd5a54d7c/ios/Shared/LaunchArguments.swift#L51)):*
> ```Swift
>         let data = try encoder.encode(self)
>         guard let result = String(bytes: data, encoding: .utf8) else {
>             throw StringDecodingError(data: data)
> ```

That should be a `StringEncodingError` here

buggmagnet (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7233#-:-ODB4aq13-VCxUQnKW0N:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

PR #7255, opened 2024-11-28
Make height prop in Spacing transient

raksooo:
I noticed that the `height` property on `Spacing` was missing a `$`. Transient props in styled components are marked with `$` and won't be passed onto the html tag.

https://styled-components.com/docs/api#transient-props

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7255)


raksooo:
Closing due to @olmoh working on improvements that remove this components

PR #7380, opened 2024-12-19
Ban async-std and old hyper and tokio versions

faern:
In order to somewhat keep our dependency tree, compile times and bugs down, we should avoid certain crates. We already do this, but the list was very short. Here are some more crates that it would be good if we avoided (accidentally?) depending on. As long as there are viable alternatives we should never use these crates.

If we come to a situation where we really have no option, entries here can be discussed of course.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7380)

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -80,8 +80,14 @@ deny = [
     { name = "openssl-sys" },
     { name = "openssl-src" },
     { name = "openssl-probe" },
+    # We are using clap 4, and want to avoid multiple versions
     { name = "clap", version = "2" },
     { name = "clap", version = "3" },
+    # We are using tokio as our asynchronous runtime. Having multiple async runtimes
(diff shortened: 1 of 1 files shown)


MarkusPettersson98 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7380#-:-OETiCiy2qiFvoEMqu4e:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

raksooo (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7380)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7382, opened 2024-12-19
Remove direct winapi dependency from mullvad-daemon exception handling

faern:
We still pull in `winapi` as a transitive dependency via a bunch of crates. But this PR at least tries to remove the last usage of it as a direct dependency. This makes very little difference in practice, but is a step towards only using a single binding for the Windows API.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7382)

### DIFF
--- a/mullvad-daemon/src/exception_logging/win.rs
+++ b/mullvad-daemon/src/exception_logging/win.rs
@@ -211,11 +211,13 @@ unsafe extern "system" fn logging_exception_filter(info_ptr: *const EXCEPTION_PO
 
 #[cfg(target_arch = "aarch64")]
 fn get_context_info(context: &CONTEXT) -> String {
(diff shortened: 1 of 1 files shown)


faern:
I had to make this PR a bit less glorious. To get the `EXCEPTION_EXECUTE_HANDLER` constant, we need to upgrade to `windows-sys 0.59`. But this was not trivial, mostly due to https://github.com/rust-lang/socket2/pull/528 :facepalm: 

I could just vendor the constant. But getting rid of `winapi` is not time critical. So I decided to keep it for now and see what happens with the `socket2` issue.

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7382)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7421, opened 2025-01-07
Fix `test_upgrade_app` fails on Windows

MarkusPettersson98:
This PR fixes the `test_upgrade_app` test on Windows. The `mullvad` binary is seemingly not in PATH after upgrading the app. So, as a workaround we use the absolute path instead. This seems to work just fine!


Serock3 (dismissed):
:lgtm:, as long as the tests pass. Nice drive-by upgrade of the error types too!

Serock3 (approved):

PR #7429, opened 2025-01-08
Use bridge interface IP from `tart`

Serock3:
The IP of the bridge interface was incorrectly assumed to be constant, but has at some point changed. This broke the algorithm  or finding the bridge interface.

Replace the algorithm with one based on the guest IP, which is given by `tart`. As the bridge IP is no longer constant, we also  expose it through the `TEST_CONFIG`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7429)

### DIFF
--- a/test/test-manager/src/main.rs
+++ b/test/test-manager/src/main.rs
@@ -9,6 +9,8 @@ mod summary;
 mod tests;
 mod vm;
 
+#[cfg(target_os = "macos")]
+use std::net::IpAddr;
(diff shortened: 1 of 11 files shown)


dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7429)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7473, opened 2025-01-16
Fix ending up in blocked state when disabling split tunnel

MarkusPettersson98:
Fix daemon ending up in blocked state if the user toggled split tunneling without having granted Full Disk Access to `mullvad-daemon`. This could only ever be accomplished from the CLI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7473)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@ Line wrap the file at 100 chars.                                              Th
 ### Fixed
 - (macOS and Windows only) Add the correct route when using obfuscation with Wireguard.
 
+#### macOS
+- Fix daemon ending up in blocked state if the user toggled split tunneling without having granted
+  Full Disk Access to `mullvad-daemon`. This could only ever be accomplished from the CLI.
+
 
 ## [2025.2] - 2025-01-08
 ### Fixed
--- a/desktop/packages/mullvad-vpn/src/renderer/components/NotificationArea.tsx
+++ b/desktop/packages/mullvad-vpn/src/renderer/components/NotificationArea.tsx
@@ -46,7 +46,7 @@ interface IProps {
 }
 
 export default function NotificationArea(props: IProps) {
-  const { showFullDiskAccessSettings, reconnectTunnel } = useAppContext();
+  const { showFullDiskAccessSettings } = useAppContext();
 
   const account = useSelector((state: IReduxState) => state.account);
   const locale = useSelector((state: IReduxState) => state.userInterface.locale);
@@ -80,8 +80,7 @@ export default function NotificationArea(props: IProps) {
   const disableSplitTunneling = useCallback(async () => {
     setIsModalOpen(false);
     await setSplitTunnelingState(false);
-    await reconnectTunnel();
-  }, [reconnectTunnel, setSplitTunnelingState]);
+  }, [setSplitTunnelingState]);
 
   const notificationProviders: InAppNotificationProvider[] = [
     new ConnectingNotificationProvider({ tunnelState }),
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -1539,11 +1539,36 @@ impl Daemon {
         tx: ResponseTx<(), Error>,
     ) {
         let save_result = match update {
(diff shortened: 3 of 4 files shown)


Serock3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7473)* status: all files reviewed, 1 unresolved discussion (waiting on @MarkusPettersson98)

___
*[`mullvad-daemon/src/lib.rs` line 1558 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7473#-OH37D9287jNStKz8A7A:-OH37D9287jNStKz8A7B:b-1y4yuk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fd88c42eb0e5a95a0f003f8cd342f6b969e06507/mullvad-daemon/src/lib.rs#L1558)):*
> ```XML
>                 if cfg!(target_os = "macos") {
>                     let split_tunnel_will_be_disabled = !state;
>                     if split_tunnel_was_enabled && split_tunnel_will_be_disabled {
> ```

My understanding is that if you enable split tunneling without full disk access (via the CLI), you end up in the error state, not the blocked state. If so, could we add a check for being in the error state to this if-statement? That should make it less likely to trigger unnecessarily.

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7473)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @Serock3)

___
*[`mullvad-daemon/src/lib.rs` line 1558 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7473#-OH37D9287jNStKz8A7A:-OH3CLOq8WBvG794WJZd:br8wbyd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fd88c42eb0e5a95a0f003f8cd342f6b969e06507/mullvad-daemon/src/lib.rs#L1558)):*

My understanding is that if you enable split tunneling without full disk access (via the CLI), you end up in the error state, not the blocked state. If so, could we add a check for being in the error state to this if-statement? That should make it less likely to trigger unnecessarily.
</blockquote></details>

That makes sense! Good catch:)

Serock3 (dismissed):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7473)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`mullvad-daemon/src/lib.rs` line 1550 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7473#-OH3ETAW2JbDPFnlB5ZF:-OH3ETAW2JbDPFnlB5ZG:bacol12) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72665236730d116c7235df6c7e0db1896554a85d/mullvad-daemon/src/lib.rs#L1550)):*
> ```XML
>                     .await
>                     .map_err(Error::SettingsError);
>                 // If Full Disk Access (FDA) is disabled, we may want to reconnect after disabling
> ```

Nit :see_no_evil: Maybe this could be worded in a way that make it a little more clear how you would end up the error state in the first place, before you try to disable full disk access?

E.g. "On macOS, if split tunneling is enabled without Full Disk Access then the user will enter the error state. This is only possible using either using the CLI or by disabling FDA after activating split tunneling. In these cases, we want to reconnect if split tunneling is disabled."

Or something. I'll leave it up to you if you want to change the text.

Serock3 (approved):
Well done! :lgtm: but make sure to test it if you haven't already

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7473#-:-OHC6L-h0K8Qr1YhvhIJ:bxol7yt)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Serock3 (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7473)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7494, opened 2025-01-21
Remove most `default-members` to speed up builds and reduce duplication

faern:
We included almost every crate in `default-members`. All crates except `windows-installer` which could not build without specific inputs as to what installers to include (it built a full universal installer, bundling arch specific installers at build time). This list had a lot of duplication and seems like a bit of a mess to maintain.

Since our workspace also contained virtually everything, a simple `cargo build` built everything, which might not be desired from a build time perspective. This PR reduce the `default-members` to some "sane default" bare minimum. It includes `mullvad-daemon` and `mullvad-cli` because those are the main binaries, and anyone building this project is most likely after those binaries. It also includes `mullvad-version` since we have a ton of invocations to `cargo run --bin mullvad-version`, and that does not work if the crate is not a default crate. We could also update all those invocations to add an extra `-p mullvad-version` to solve the issue. But I felt like the version computation binary was central enough to be included as a default crate :shrug: 

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7494)

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -75,8 +75,8 @@ jobs:
           RUSTFLAGS: --deny warnings
         run: |
           source env.sh
-          time cargo clippy --locked --all-targets --no-default-features
-          time cargo clippy --locked --all-targets --all-features
+          time cargo clippy --workspace --locked --all-targets --no-default-features
+          time cargo clippy --workspace --locked --all-targets --all-features
 
   clippy-check-android:
     name: Clippy linting, Android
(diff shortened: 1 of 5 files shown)


dlon (approved):
:lgtm: :sparkles: 

> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7494#-:-OH8Xet92M-zwZPaeh9E:bdti0a6)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7508, opened 2025-01-23
Add rustc remap prefix helper crate

kl:
Fixes: DROID-1747

This is needed for reproducible builds. This binary helper crate gets the file paths to the user's Cargo and Rustup home dirs, as well as the path to the Mullvad app source dir. These paths are then remapped to fixed values which is needed to make the build reproducible.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7508)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1581,11 +1581,11 @@ dependencies = [
 
 [[package]]
 name = "home"
-version = "0.5.9"
+version = "0.5.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e3d1354bf6b7235cb4a0576c2619fd4ed18183f689b12b006a0ee7329eeff9a5"
+checksum = "589533453244b0995c858700322199b2becb13b627df2851f64a2775d024abcf"
 dependencies = [
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3768,6 +3768,13 @@ name = "regex-syntax"
 version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "adad44e29e4c806119491a7f06f03de4d1af22c3a680dd47f1e6e179439d1f56"
+
+[[package]]
+name = "remap-path-prefix"
+version = "0.1.0"
+dependencies = [
+ "home",
+]
 
 [[package]]
 name = "resolv-conf"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3771,7 +3771,7 @@ checksum = "adad44e29e4c806119491a7f06f03de4d1af22c3a680dd47f1e6e179439d1f56"
 
 [[package]]
 name = "remap-path-prefix"
-version = "0.1.0"
+version = "0.0.0"
 dependencies = [
  "home",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ members = [
     "mullvad-types",
     "mullvad-types/intersection-derive",
     "mullvad-version",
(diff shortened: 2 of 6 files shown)


faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7508)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @kl)

___
*[`Cargo.toml` line 57 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7508#-OHIMICXDl6iUFzzUGHY:-OHIMICXDl6iUFzzUGHZ:b-pjp7nd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/43eeb65c77d8daa85f51f1e1495ac45a4338e611/Cargo.toml#L57)):*
> ```TOML
>     "mullvad-daemon",
>     "mullvad-version",
>     "remap-path-prefix",
> ```

Please don't add a crate to the default members unless it's strictly needed. You can instead run the binary by adding `-p remap-path-prefix` to your invocation of `cargo run`.

We want to keep this list as short as possible both to bring down compile times during development as well as not maintaining two lists of crates.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7508)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @kl)

___
*[`remap-path-prefix/Cargo.toml` line 3 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7508#-OHIMhHx8WiuQyMVwriS:-OHIMhHx8WiuQyMVwriT:bh0cchy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/43eeb65c77d8daa85f51f1e1495ac45a4338e611/remap-path-prefix/Cargo.toml#L3)):*
> ```TOML
> [package]
> name = "remap-path-prefix"
> version = "0.1.0"
> ```

Please do not specify version of crates. Also make sure to inherit all other fields from the workspace. Including linting rules etc. Otherwise this crate risks deviating in code quality and other aspects. See other `Cargo.toml` files for what to include. You can basically just copy paste from them.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7508)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @kl)

___
*[`remap-path-prefix/src/main.rs` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7508#-OHIPu562ukF90nYd1e8:-OHIPu562ukF90nYd1e9:b-x4lsek) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/43eeb65c77d8daa85f51f1e1495ac45a4338e611/remap-path-prefix/src/main.rs#L16)):*
> ```XML
> 
>     let source_dir = env!("CARGO_MANIFEST_DIR")
>         .split(concat!("/", env!("CARGO_PKG_NAME")))
> ```

I can't on the top of my head come up with a better solution, so maybe this is the best we can do. But this has two downsides:
1. It's evaluated at compile time. Mean if the invocation of `cargo run` on this binary were to reuse a previous build built from another directory, this would point to the wrong source dir.
2. It relies on the relative path to this crate relative to the workspace root. I checked, but unfortunately did not find any cargo env set that points to the workspace root dir. That would have been awesome.

Maybe we can mitigate issue 1 with a `build.rs` containing something like `println!("cargo:rerun-if-env-changed=CARGO_MANIFEST_DIR");`

kl:
Superseded by #7512

PR #7522, opened 2025-01-24
Append --locked when building Rust libraries for iOS for releases

faern:
This prevents the build from succeeding if the lockfile is dirty. This stops us from releasing apps that are hard to re-build with the exact same dependencies, or that had malicious stuff injected under some circumstances

I cannot very easily test this out. Sorry for that. So I'm marking the PR as a draft since I'm not sure if this works.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7522)

### DIFF
--- a/ios/build-rust-library.sh
+++ b/ios/build-rust-library.sh
@@ -24,12 +24,12 @@ echo ${FEATURE_FLAGS:+--features "$FEATURE_FLAGS"}
 fi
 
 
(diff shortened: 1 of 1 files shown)


faern:
This PR is superseded by https://github.com/mullvad/mullvadvpn-app/pull/9088

PR #7577, opened 2025-02-03
Track IPv6 connectivity on Android

Pururun:
Exposes the availability of IPv4 and IPv6 to the daemon.

This is done by creating a dummy socket on the uderlying network.

This aligns android with other platforms.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)

### DIFF
--- a/android/CHANGELOG.md
+++ b/android/CHANGELOG.md
@@ -34,6 +34,9 @@ Line wrap the file at 100 chars.                                              Th
 ### Removed
 - Remove Google's resolvers from encrypted DNS proxy.
 
+### Fixed
+- Will no longer try to connect over IPv6 if IPv6 is not available.
+
 
 ## [android/2024.10-beta2] - 2024-12-20
 
--- a/android/app/src/androidTest/kotlin/net/mullvad/mullvadvpn/talpid/util/ConnectivityManagerUtilKtTest.kt
+++ b/android/app/src/androidTest/kotlin/net/mullvad/mullvadvpn/talpid/util/ConnectivityManagerUtilKtTest.kt
@@ -1,11 +1,17 @@
 package net.mullvad.mullvadvpn.talpid.util
 
 import android.net.ConnectivityManager
+import android.net.LinkAddress
+import android.net.LinkProperties
 import android.net.Network
+import android.net.NetworkCapabilities
 import app.cash.turbine.test
 import io.mockk.every
 import io.mockk.mockk
 import io.mockk.mockkStatic
+import io.mockk.verify
+import java.net.Inet4Address
+import java.net.Inet6Address
 import kotlin.test.assertEquals
 import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.seconds
@@ -13,10 +19,11 @@ import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.callbackFlow
 import kotlinx.coroutines.test.runTest
+import net.mullvad.talpid.model.Connectivity
 import net.mullvad.talpid.util.NetworkEvent
+import net.mullvad.talpid.util.UnderlyingConnectivityStatusResolver
+import net.mullvad.talpid.util.defaultNetworkEvents
 import net.mullvad.talpid.util.hasInternetConnectivity
-import net.mullvad.talpid.util.networkEvents
-import net.mullvad.talpid.util.networksWithInternetConnectivity
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 
@@ -31,19 +38,22 @@ class ConnectivityManagerUtilKtTest {
     /** User being online, the listener should emit once with `true` */
     @Test
     fun userIsOnline() = runTest {
-        val network = mockk<Network>()
-        every { connectivityManager.networksWithInternetConnectivity() } returns setOf(network)
-        every { connectivityManager.networkEvents(any()) } returns
+        val network = mockk<Network>(relaxed = true)
+        val linkProperties = mockLinkProperties(true, true)
+        val mockResolver = mockk<UnderlyingConnectivityStatusResolver>()
+        every { connectivityManager.defaultNetworkEvents() } returns
             callbackFlow {
                 delay(100.milliseconds) // Simulate connectivity listener being a bit slow
                 send(NetworkEvent.Available(network))
+                delay(100.milliseconds) // Simulate connectivity listener being a bit slow
+                send(NetworkEvent.LinkPropertiesChanged(network, linkProperties))
                 awaitClose {}
             }
 
-        connectivityManager.hasInternetConnectivity().test {
+        connectivityManager.hasInternetConnectivity(mockResolver).test {
             // Since initial state and listener both return `true` within debounce we only see one
             // event
-            assertEquals(true, awaitItem())
+            assertEquals(Connectivity.Status(true, true), awaitItem())
             expectNoEvents()
         }
     }
@@ -51,12 +61,12 @@ class ConnectivityManagerUtilKtTest {
     /** User being offline, the listener should emit once with `false` */
     @Test
     fun userIsOffline() = runTest {
-        every { connectivityManager.networksWithInternetConnectivity() } returns setOf()
-        every { connectivityManager.networkEvents(any()) } returns callbackFlow { awaitClose {} }
+        val mockResolver = mockk<UnderlyingConnectivityStatusResolver>()
+        every { connectivityManager.defaultNetworkEvents() } returns callbackFlow { awaitClose {} }
 
-        connectivityManager.hasInternetConnectivity().test {
+        connectivityManager.hasInternetConnectivity(mockResolver).test {
             // Initially offline and no network events, so we should get a single `false` event
-            assertEquals(false, awaitItem())
+            assertEquals(Connectivity.Status(false, false), awaitItem())
             expectNoEvents()
         }
     }
@@ -64,19 +74,22 @@ class ConnectivityManagerUtilKtTest {
     /** User starting offline and then turning on a online after a while */
     @Test
     fun initiallyOfflineThenBecomingOnline() = runTest {
-        every { connectivityManager.networksWithInternetConnectivity() } returns emptySet()
-        every { connectivityManager.networkEvents(any()) } returns
+        val network = mockk<Network>()
+        val linkProperties = mockLinkProperties(true, true)
+        val mockResolver = mockk<UnderlyingConnectivityStatusResolver>()
+        every { connectivityManager.defaultNetworkEvents() } returns
             callbackFlow {
                 // Simulate offline for a little while
                 delay(5.seconds)
                 // Then become online
                 send(NetworkEvent.Available(mockk()))
+                send(NetworkEvent.LinkPropertiesChanged(network, linkProperties))
                 awaitClose {}
             }
 
-        connectivityManager.hasInternetConnectivity().test {
-            assertEquals(false, awaitItem())
-            assertEquals(true, awaitItem())
+        connectivityManager.hasInternetConnectivity(mockResolver).test {
+            assertEquals(Connectivity.Status(false, false), awaitItem())
+            assertEquals(Connectivity.Status(true, true), awaitItem())
             expectNoEvents()
         }
     }
@@ -85,46 +98,23 @@ class ConnectivityManagerUtilKtTest {
     @Test
     fun initiallyOnlineAndThenTurningBecomingOffline() = runTest {
         val network = mockk<Network>()
-        every { connectivityManager.networksWithInternetConnectivity() } returns setOf(network)
-        every { connectivityManager.networkEvents(any()) } returns
+        val linkProperties = mockLinkProperties(true, true)
+
+        val mockResolver = mockk<UnderlyingConnectivityStatusResolver>()
+        every { connectivityManager.defaultNetworkEvents() } returns
             callbackFlow {
                 // Starting as online
                 send(NetworkEvent.Available(network))
+                send(NetworkEvent.LinkPropertiesChanged(network, linkProperties))
                 delay(5.seconds)
                 // Then becoming offline
                 send(NetworkEvent.Lost(network))
                 awaitClose {}
             }
 
-        connectivityManager.hasInternetConnectivity().test {
-            assertEquals(true, awaitItem())
-            assertEquals(false, awaitItem())
-            expectNoEvents()
-        }
-    }
-
-    /**
-     * User turning on Airplane mode as our connectivity listener starts so we never get any
-     * onAvailable event from our listener. Initial value will be `true`, followed by no
-     * `networkEvent` and then turning on network again after 5 seconds
-     */
-    @Test
-    fun incorrectInitialValueThenBecomingOnline() = runTest {
-        every { connectivityManager.networksWithInternetConnectivity() } returns setOf(mockk())
-        every { connectivityManager.networkEvents(any()) } returns
-            callbackFlow {
-                delay(5.seconds)
-                send(NetworkEvent.Available(mockk()))
-                awaitClose {}
-            }
-
-        connectivityManager.hasInternetConnectivity().test {
-            // Initial value is connected
-            assertEquals(true, awaitItem())
-            // Debounce time has passed, and we never received any network events, so we are offline
-            assertEquals(false, awaitItem())
-            // Network is back online
-            assertEquals(true, awaitItem())
+        connectivityManager.hasInternetConnectivity(mockResolver).test {
+            assertEquals(Connectivity.Status(true, true), awaitItem())
+            assertEquals(Connectivity.Status(false, false), awaitItem())
             expectNoEvents()
         }
     }
@@ -133,26 +123,34 @@ class ConnectivityManagerUtilKtTest {
     @Test
     fun roamingFromCellularToWifi() = runTest {
         val wifiNetwork = mockk<Network>()
+        val wifiNetworkLinkProperties = mockLinkProperties(true, false)
         val cellularNetwork = mockk<Network>()
+        val cellularNetworkLinkProperties = mockLinkProperties(true, false)
+        val mockResolver = mockk<UnderlyingConnectivityStatusResolver>()
 
-        every { connectivityManager.networksWithInternetConnectivity() } returns
-            setOf(cellularNetwork)
-        every { connectivityManager.networkEvents(any()) } returns
+        every { connectivityManager.defaultNetworkEvents() } returns
             callbackFlow {
                 send(NetworkEvent.Available(cellularNetwork))
+                send(
+                    NetworkEvent.LinkPropertiesChanged(
+                        cellularNetwork,
+                        cellularNetworkLinkProperties,
+                    )
+                )
                 delay(5.seconds)
                 // Turning on WiFi, we'll have duplicate networks until phone decides to turn of
                 // cellular
                 send(NetworkEvent.Available(wifiNetwork))
+                send(NetworkEvent.LinkPropertiesChanged(wifiNetwork, wifiNetworkLinkProperties))
                 delay(30.seconds)
                 // Phone turning off cellular network
                 send(NetworkEvent.Lost(cellularNetwork))
                 awaitClose {}
             }
 
-        connectivityManager.hasInternetConnectivity().test {
+        connectivityManager.hasInternetConnectivity(mockResolver).test {
             // We should always only see us being online
-            assertEquals(true, awaitItem())
+            assertEquals(Connectivity.Status(ipv4 = true, ipv6 = false), awaitItem())
             expectNoEvents()
         }
     }
@@ -161,23 +159,32 @@ class ConnectivityManagerUtilKtTest {
     @Test
     fun roamingFromWifiToCellular() = runTest {
         val wifiNetwork = mockk<Network>()
+        val wifiNetworkLinkProperties = mockLinkProperties(true, false)
         val cellularNetwork = mockk<Network>()
+        val cellularNetworkLinkProperties = mockLinkProperties(true, false)
+        val mockResolver = mockk<UnderlyingConnectivityStatusResolver>()
 
-        every { connectivityManager.networksWithInternetConnectivity() } returns setOf(wifiNetwork)
-        every { connectivityManager.networkEvents(any()) } returns
+        every { connectivityManager.defaultNetworkEvents() } returns
             callbackFlow {
                 send(NetworkEvent.Available(wifiNetwork))
+                send(NetworkEvent.LinkPropertiesChanged(wifiNetwork, wifiNetworkLinkProperties))
                 delay(5.seconds)
                 send(NetworkEvent.Lost(wifiNetwork))
                 // We will have no network for a little time until cellular chip is on.
                 delay(150.milliseconds)
                 send(NetworkEvent.Available(cellularNetwork))
+                send(
+                    NetworkEvent.LinkPropertiesChanged(
+                        cellularNetwork,
+                        cellularNetworkLinkProperties,
+                    )
+                )
                 awaitClose {}
             }
 
-        connectivityManager.hasInternetConnectivity().test {
+        connectivityManager.hasInternetConnectivity(mockResolver).test {
             // We should always only see us being online, small offline state is caught by debounce
-            assertEquals(true, awaitItem())
+            assertEquals(Connectivity.Status(ipv4 = true, ipv6 = false), awaitItem())
             expectNoEvents()
         }
     }
@@ -186,30 +193,114 @@ class ConnectivityManagerUtilKtTest {
     @Test
     fun slowRoamingFromWifiToCellular() = runTest {
         val wifiNetwork = mockk<Network>()
+        val wifiNetworkLinkProperties = mockLinkProperties(false, true)
         val cellularNetwork = mockk<Network>()
+        val cellularNetworkLinkProperties = mockLinkProperties(false, true)
+        val mockResolver = mockk<UnderlyingConnectivityStatusResolver>()
 
-        every { connectivityManager.networksWithInternetConnectivity() } returns setOf(wifiNetwork)
-        every { connectivityManager.networkEvents(any()) } returns
+        every { connectivityManager.defaultNetworkEvents() } returns
             callbackFlow {
                 send(NetworkEvent.Available(wifiNetwork))
+                send(NetworkEvent.LinkPropertiesChanged(wifiNetwork, wifiNetworkLinkProperties))
                 delay(5.seconds)
                 send(NetworkEvent.Lost(wifiNetwork))
                 // We will have no network for a little time until cellular chip is on.
                 delay(500.milliseconds)
                 send(NetworkEvent.Available(cellularNetwork))
+                send(
+                    NetworkEvent.LinkPropertiesChanged(
+                        cellularNetwork,
+                        cellularNetworkLinkProperties,
+                    )
+                )
                 awaitClose {}
             }
 
-        connectivityManager.hasInternetConnectivity().test {
+        connectivityManager.hasInternetConnectivity(mockResolver).test {
             // Wifi is online
-            assertEquals(true, awaitItem())
+            assertEquals(Connectivity.Status(false, true), awaitItem())
             // We didn't get any network within debounce time, so we are offline
-            assertEquals(false, awaitItem())
+            assertEquals(Connectivity.Status(false, false), awaitItem())
             // Cellular network is online
-            assertEquals(true, awaitItem())
+            assertEquals(Connectivity.Status(false, true), awaitItem())
             expectNoEvents()
         }
     }
+
+    /** Switching between networks with different configurations. */
+    @Test
+    fun roamingFromWifiWithIpv6OnlyToWifiWithIpv4Only() = runTest {
+        val ipv6Network = mockk<Network>()
+        val ipv6NetworkLinkProperties = mockLinkProperties(false, true)
+        val ipv4Network = mockk<Network>()
+        val ipv4NetworkLinkProperties = mockLinkProperties(true, false)
+        val mockResolver = mockk<UnderlyingConnectivityStatusResolver>()
+
+        every { connectivityManager.defaultNetworkEvents() } returns
+            callbackFlow {
+                send(NetworkEvent.Available(ipv6Network))
+                send(NetworkEvent.LinkPropertiesChanged(ipv6Network, ipv6NetworkLinkProperties))
+                delay(5.seconds)
+                send(NetworkEvent.Lost(ipv6Network))
+                delay(100.milliseconds)
+                send(NetworkEvent.Available(ipv4Network))
+                send(NetworkEvent.LinkPropertiesChanged(ipv4Network, ipv4NetworkLinkProperties))
+                awaitClose {}
+            }
+
+        connectivityManager.hasInternetConnectivity(mockResolver).test {
+            // Ipv6 network is online
+            assertEquals(Connectivity.Status(false, true), awaitItem())
+            // Ipv4 network is online
+            assertEquals(Connectivity.Status(true, false), awaitItem())
+            expectNoEvents()
+        }
+    }
+
+    /** Vpn network should NOT check link properties but should rather use socket implementation */
+    @Test
+    fun checkVpnNetworkUsingSocketImplementation() = runTest {
+        val vpnNetwork = mockk<Network>()
+        val capabilities = mockk<NetworkCapabilities>()
+        every { capabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_NOT_VPN) } returns
+            false
+        val mockResolver = mockk<UnderlyingConnectivityStatusResolver>()
+        every { mockResolver.currentStatus() } returns Connectivity.Status(true, true)
+
+        every { connectivityManager.defaultNetworkEvents() } returns
+            callbackFlow {
+                send(NetworkEvent.Available(vpnNetwork))
+                send(NetworkEvent.CapabilitiesChanged(vpnNetwork, capabilities))
+                awaitClose {}
+            }
+
+        connectivityManager.hasInternetConnectivity(mockResolver).test {
+            // Network is online
+            assertEquals(Connectivity.Status(true, true), awaitItem())
+        }
+
+        verify(exactly = 1) { mockResolver.currentStatus() }
+    }
+
+    private fun mockLinkProperties(ipv4: Boolean, ipv6: Boolean) =
+        mockk<LinkProperties> {
+            val linkAddresses = buildList {
+                if (ipv4) {
+                    val linkIpv4Address: LinkAddress = mockk()
+                    val ipv4Address: Inet4Address = mockk()
+                    every { linkIpv4Address.address } returns ipv4Address
+                    add(linkIpv4Address)
+                }
+                if (ipv6) {
+                    val linkIpv6Address: LinkAddress = mockk()
+                    val ipv6Address: Inet6Address = mockk()
+                    every { linkIpv6Address.address } returns ipv6Address
+                    add(linkIpv6Address)
+                }
+            }
+
+            every { this@mockk.linkAddresses } returns linkAddresses
+        }
 
     companion object {
         private const val CONNECTIVITY_MANAGER_UTIL_CLASS =
--- a/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt
+++ b/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt
@@ -5,6 +5,8 @@ import android.net.LinkProperties
 import java.net.InetAddress
 import kotlin.collections.ArrayList
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharingStarted
@@ -15,13 +17,22 @@ import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.receiveAsFlow
 import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.plus
+import kotlinx.coroutines.runBlocking
+import net.mullvad.talpid.model.Connectivity
 import net.mullvad.talpid.model.NetworkState
 import net.mullvad.talpid.util.RawNetworkState
+import net.mullvad.talpid.util.UnderlyingConnectivityStatusResolver
+import net.mullvad.talpid.util.activeRawNetworkState
 import net.mullvad.talpid.util.defaultRawNetworkStateFlow
 import net.mullvad.talpid.util.hasInternetConnectivity
+import net.mullvad.talpid.util.resolveConnectivityStatus
 
-class ConnectivityListener(private val connectivityManager: ConnectivityManager) {
-    private lateinit var _isConnected: StateFlow<Boolean>
+class ConnectivityListener(
+    private val connectivityManager: ConnectivityManager,
+    private val resolver: UnderlyingConnectivityStatusResolver,
+) {
+    private lateinit var _isConnected: StateFlow<Connectivity>
     // Used by JNI
     val isConnected
         get() = _isConnected.value
@@ -37,6 +48,7 @@ class ConnectivityListener(private val connectivityManager: ConnectivityManager)
     val currentDnsServers: ArrayList<InetAddress>
         get() = _mutableNetworkState.value?.dnsServers ?: ArrayList()
 
+    @OptIn(FlowPreview::class)
     fun register(scope: CoroutineScope) {
         // Consider implementing retry logic for the flows below, because registering a listener on
         // the default network may fail if the network on Android 11
@@ -53,12 +65,19 @@ class ConnectivityListener(private val connectivityManager: ConnectivityManager)
 
         _isConnected =
             connectivityManager
-                .hasInternetConnectivity()
-                .onEach { notifyConnectivityChange(it) }
+                .hasInternetConnectivity(resolver)
+                .onEach { notifyConnectivityChange(it.ipv4, it.ipv6) }
                 .stateIn(
-                    scope,
+                    scope + Dispatchers.IO,
                     SharingStarted.Eagerly,
-                    true, // Assume we have internet until we know otherwise
+                    // Has to happen on IO to avoid NetworkOnMainThreadException, we actually don't
+                    // send any traffic just open a socket to detect the IP version.
+                    runBlocking(Dispatchers.IO) {
+                        resolveConnectivityStatus(
+                            connectivityManager.activeRawNetworkState(),
+                            resolver,
+                        )
+                    },
                 )
     }
 
@@ -80,7 +99,7 @@ class ConnectivityListener(private val connectivityManager: ConnectivityManager)
             linkProperties?.dnsServersWithoutFallback(),
         )
 
-    private external fun notifyConnectivityChange(isConnected: Boolean)
+    private external fun notifyConnectivityChange(isIPv4: Boolean, isIPv6: Boolean)
 
     private external fun notifyDefaultNetworkChange(networkState: NetworkState?)
 }
--- a/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/TalpidVpnService.kt
+++ b/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/TalpidVpnService.kt
@@ -26,6 +26,7 @@ import net.mullvad.talpid.model.CreateTunResult.OtherAlwaysOnApp
 import net.mullvad.talpid.model.CreateTunResult.OtherLegacyAlwaysOnVpn
 import net.mullvad.talpid.model.TunConfig
 import net.mullvad.talpid.util.TalpidSdkUtils.setMeteredIfSupported
+import net.mullvad.talpid.util.UnderlyingConnectivityStatusResolver
 
 open class TalpidVpnService : LifecycleVpnService() {
     private var activeTunStatus by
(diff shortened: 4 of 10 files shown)


dlon (approved):
It could be mentioned in the changelog that this prevents the app from trying to connect over IPv6 if it is not available.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-:-OIAtX0dE3adFUpTP4s7:bjlgsy2)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIAsuXs3o5CtlfiXoUM:-OIAsuXt0UwVktF1fhoU:b1cr4zp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/98c09bf16c200c65509edd4412af8ec3f838d26b/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L32)):*
> ```Kotlin
>         } catch (_: SocketException) {
>             Logger.e("Socket could not be set up")
>             false
> ```

I'm a bit worried about this failing for some unknown reason and causing the app to think it's offline. But maybe it never will?

Pururun (commented):
Added!

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-:-OIAwD3IC62-tjouWrnN:bqmjwcm)* status: all files reviewed, 1 unresolved discussion

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIAsuXs3o5CtlfiXoUM:-OIAvdYg8MiiGoj_GMJb:b-dcdwyd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/98c09bf16c200c65509edd4412af8ec3f838d26b/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L32)):*

I'm a bit worried about this failing for some unknown reason and causing the app to think it's offline. But maybe it never will?
</blockquote></details>

Good point, since we need to have some kind of connection for this to even be attempted, maybe we should fall back to everything connected if both of IPv4 and IPv6 is not available.
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 49 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIAw5Ax2siGFXsIeRE7:-OIAw5Ax2siGFXsIeRE8:b-aa886o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/98c09bf16c200c65509edd4412af8ec3f838d26b/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L49)):*
> ```Kotlin
> 
>         _isConnected =
>             combine(connectivityManager.defaultNetworkFlow(), hasInternetCapability()) {
> ```

We will update for every default network event here, maybe that is a bit excessive?

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 7 of 9 files reviewed, 1 unresolved discussion (waiting on @dlon)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 32 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIAsuXs3o5CtlfiXoUM:-OIB0IWrBv6SZc9jXRcq:bpksnzb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/98c09bf16c200c65509edd4412af8ec3f838d26b/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L32)):*

Good point, since we need to have some kind of connection for this to even be attempted, maybe we should fall back to everything connected if both of IPv4 and IPv6 is not available.
</blockquote></details>

I added a fallback behaviour in connectivity listener so that we will never return a disconnected state if at least one network with internet is available.

dlon (approved):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 1 unresolved discussion

MarkusPettersson98 (commented):
> Reviewed 3 of 8 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 1 unresolved discussion

hulthe (commented):
> Reviewed 2 of 8 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 4 unresolved discussions (waiting on @Pururun)

___
*[`talpid-core/src/connectivity_listener.rs` line 85 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIBc4ya235x0p8oBpPr:-OIBc4ya235x0p8oBpPs:b-lfhjk7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/talpid-core/src/connectivity_listener.rs#L85)):*
> ```XML
> 
>         Ok(ConnectivityListener {
>             jvm: android_context.clone().jvm,
> ```

nit: unnecessary clone
___
*[`talpid-core/src/connectivity_listener.rs` line 140 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIBd0b52qZj7jmIOkKh:-OIBd0b52qZj7jmIOkKi:b-rczyo5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/talpid-core/src/connectivity_listener.rs#L140)):*
> ```XML
>             .expect("Missing ConnectionStatus.ipv6")
>             .z()
>             .expect("ipv6 is not a boolean");
> ```

Have you considered using `derive(jnix::FromJava)` instead of converting the java object manually?
___
*[`talpid-core/src/connectivity_listener.rs` line 187 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIBbLH3CSty6n709v3u:-OIBbLH3CSty6n709v3v:b-tt2kou) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/talpid-core/src/connectivity_listener.rs#L187)):*
> ```XML
> 
>     let isIPv4 = JNI_TRUE == is_ipv4;
>     let isIPv6 = JNI_TRUE == is_ipv6;
> ```

nit: variables should be snake_case

Rawa (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 7 unresolved discussions (waiting on @Pururun)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 49 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF14PSEOdb8DTsrCvG:-OIF14PSEOdb8DTsrCvH:b-h7am4z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L49)):*
> ```Kotlin
> 
>         _isConnected =
>             combine(connectivityManager.defaultNetworkFlow(), hasInternetCapability()) {
> ```

Is it necessary to combine `hasInternetCapability` and `defaultNetworkFlow`? Or maybe we think it is a bit slow?
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1yZZ7w9TbRSN29a0:-OIF1yZZ7w9TbRSN29a1:b-v3gik5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L24)):*
> ```Kotlin
>         ip: T,
>         protect: (socket: DatagramSocket) -> Unit,
>     ): Boolean {
> ```

Right now we don't separate between fail to fetch if a IP version vs not available, should we? Or is it fine that they mean the same thing?
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 27 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1lJ49_K9IaOj22HN:-OIF1lJ49_K9IaOj22HO:b-jxyp7m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L27)):*
> ```Kotlin
>         val socket = DatagramSocket()
>         return try {
>             protect(socket)
> ```

We should verify that the protect was successful, probably also log if it isn't.

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 7 unresolved discussions (waiting on @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 49 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF14PSEOdb8DTsrCvG:-OIF5-8gE_4u_W141shM:bosuk5d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L49)):*

Is it necessary to combine `hasInternetCapability` and `defaultNetworkFlow`? Or maybe we think it is a bit slow?
</blockquote></details>

What do you mean slow? The reason is that we need to know if the have a internet connection and we want to react to changes in default network so that we can check if network has IPv4 and/or IPv6. But as I noted above we probably only need to react if the underlying default network change, when chanhing from mobile data to wifi or between wifi networks.
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1yZZ7w9TbRSN29a0:-OIF5sqN3iMC9wSQT9to:b-o0nvdy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L24)):*

Right now we don't separate between fail to fetch if a IP version vs not available, should we? Or is it fine that they mean the same thing?
</blockquote></details>

To me it is fine, I have currently not seen any reason to different them.
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 27 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1lJ49_K9IaOj22HN:-OIF64by0c0tBwhxmM6r:b-ty21e3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L27)):*

We should verify that the protect was successful, probably also log if it isn't.
</blockquote></details>

Protect will always fail if a VPN is not set up, so if we want to verify it we need to know if the VPN is set up properly.

dlon (dismissed):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 8 unresolved discussions (waiting on @Pururun and @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 49 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIAw5Ax2siGFXsIeRE7:-OIF3Kcp3vkyLSVwxW_7:bdac20c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/98c09bf16c200c65509edd4412af8ec3f838d26b/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L49)):*

We will update for every default network event here, maybe that is a bit excessive?
</blockquote></details>

If it's not too much work, maybe "debouncing" the events would make sense. See for example `BurstGuard`
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 21 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF2vkg1UxKEBKJl4Qg:-OIF2vkg1UxKEBKJl4Qh:b-zba7fo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L21)):*
> ```Kotlin
>     // If the ip version is not supported on the underlying network it will trigger a socket
>     // exception. If not it should return the local ip address.
>     private inline fun <reified T : InetAddress> isIPAvailable(
> ```

Should we document the reason for this hack? I.e., "there's no other way to figure out which interface is the (non-VPN) default interface".

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 8 unresolved discussions (waiting on @Pururun and @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1yZZ7w9TbRSN29a0:-OIHLAA0AcHJYZxrPTXz:bu5a70c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L24)):*

To me it is fine, I have currently not seen any reason to different them.
</blockquote></details>

It seems a bit risky to return "no connectivity" if the cause is something else. If possible, it seems safer to return `true` unless it fails for that specific reason

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 3 of 10 files reviewed, 5 unresolved discussions (waiting on @dlon, @hulthe, @MarkusPettersson98, and @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 21 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF2vkg1UxKEBKJl4Qg:-OIK7qos7MF8Qgw6luIT:b-d3xen0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L21)):*

Should we document the reason for this hack? I.e., "there's no other way to figure out which interface is the (non-VPN) default interface".
</blockquote></details>

Good idead, added!
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1yZZ7w9TbRSN29a0:-OIK7uLr9-1bDIS1bPhC:bph67d1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L24)):*

It seems a bit risky to return "no connectivity" if the cause is something else. If possible, it seems safer to return `true` unless it fails for that specific reason
</blockquote></details>

I guess it depends on what we expected behaviour. When I was testing it always threw a `SocketException` when IPv6/IPv4 was not available, but I am fine with just returning true if not exception is thrown.
___
*[`talpid-core/src/connectivity_listener.rs` line 85 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIBc4ya235x0p8oBpPr:-OIKChK72OfkXPBxAxfh:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/talpid-core/src/connectivity_listener.rs#L85)):*

nit: unnecessary clone
</blockquote></details>

Done
___
*[`talpid-core/src/connectivity_listener.rs` line 140 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIBd0b52qZj7jmIOkKh:-OIKCgjc5npHNuSRnSvq:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/talpid-core/src/connectivity_listener.rs#L140)):*

Have you considered using `derive(jnix::FromJava)` instead of converting the java object manually?
</blockquote></details>

Done
___
*[`talpid-core/src/connectivity_listener.rs` line 187 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIBbLH3CSty6n709v3u:-OIKCg8lAUN4X2RVJwYH:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/talpid-core/src/connectivity_listener.rs#L187)):*

nit: variables should be snake_case
</blockquote></details>

Done

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 3 of 10 files reviewed, 5 unresolved discussions (waiting on @dlon, @hulthe, @MarkusPettersson98, and @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1yZZ7w9TbRSN29a0:-OIKWD9e2nx70WH11y99:b-osiy12) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L24)):*

I guess it depends on what we expected behaviour. When I was testing it always threw a `SocketException` when IPv6/IPv4 was not available, but I am fine with just returning true if not exception is thrown.
</blockquote></details>

I changed it so that it always assumed that it is working for now if it does not throw an exception.

hulthe (commented):
> Reviewed 5 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 8 of 10 files reviewed, 5 unresolved discussions (waiting on @dlon, @Pururun, and @Rawa)

MarkusPettersson98 (commented):
> Reviewed 3 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 8 of 10 files reviewed, 5 unresolved discussions (waiting on @dlon, @Pururun, and @Rawa)

dlon (approved):
> Reviewed 2 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 4 unresolved discussions (waiting on @Pururun and @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1yZZ7w9TbRSN29a0:-OIM2sml6KpQ8PHQAdVs:btie2gm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L24)):*

I changed it so that it always assumed that it is working for now if it does not throw an exception.
</blockquote></details>

As long as `SocketException` is raised *only* in that case, I guess it's fine

Rawa (changes requested):
> Reviewed 1 of 1 files at r2, 4 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 4 unresolved discussions (waiting on @Pururun)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 49 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF14PSEOdb8DTsrCvG:-OIPcvnd9GP18sXm4NkS:bkoac9l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L49)):*

What do you mean slow? The reason is that we need to know if the have a internet connection and we want to react to changes in default network so that we can check if network has IPv4 and/or IPv6. But as I noted above we probably only need to react if the underlying default network change, when chanhing from mobile data to wifi or between wifi networks.
</blockquote></details>

I mean the `defaultNetworkFlow` could also provide if it has internet or not. But we wouldn't know for sure until the `Capabilities` callback is invoked.
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 27 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1lJ49_K9IaOj22HN:-OIPdeas9q8-kNKBWIuP:b-6dp4v4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L27)):*

Protect will always fail if a VPN is not set up, so if we want to verify it we need to know if the VPN is set up properly.
</blockquote></details>

Yes, but we don't know if it will Succeed just because of a VPN is currently setup, it probably will but there is no guarantee in the interface. If we fail to protect we should handle it gracefully.

```
val protectSuccess = protect(socket)
if(!protectSuccess) {
    // return what ever makes sense if the protect has failed.
}
```

Rawa (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 5 unresolved discussions (waiting on @Pururun)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 29 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIPeKzyB2TrpFtvCM3q:-OIPeKzyB2TrpFtvCM3r:b-8zlnt0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1c6696e3b1ad908fd2967056de2ac3685e3577f5/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L29)):*
> ```Kotlin
>     ): Boolean {
>         val socket = DatagramSocket()
>         return try {
> ```

We should limit the try to be as small as possible. `protect` should be lifted out.

Rawa (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 5 unresolved discussions (waiting on @dlon and @Pururun)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1yZZ7w9TbRSN29a0:-OIPeeHK98no1U7-QYSn:b-grd46f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L24)):*

As long as `SocketException` is raised *only* in that case, I guess it's fine
</blockquote></details>

I'll leave this unresolved until that has been confirmed.

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 3 of 10 files reviewed, 5 unresolved discussions (waiting on @dlon, @hulthe, @MarkusPettersson98, and @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 49 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF14PSEOdb8DTsrCvG:-OImFiPQAF2lhGVbbIyx:b-wr0031) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L49)):*

I mean the `defaultNetworkFlow` could also provide if it has internet or not. But we wouldn't know for sure until the `Capabilities` callback is invoked.
</blockquote></details>

Changed it to use the new `defaultRawNetworkStateFlow()` which I guess is as slow?
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 27 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1lJ49_K9IaOj22HN:-OImG0SA0VfpvhMVzTV2:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L27)):*

Yes, but we don't know if it will Succeed just because of a VPN is currently setup, it probably will but there is no guarantee in the interface. If we fail to protect we should handle it gracefully.

```
val protectSuccess = protect(socket)
if(!protectSuccess) {
    // return what ever makes sense if the protect has failed.
}
```
</blockquote></details>

Done
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 29 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIPeKzyB2TrpFtvCM3q:-OImFxeBEn9m0EG44Xw6:b-yovdpg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1c6696e3b1ad908fd2967056de2ac3685e3577f5/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L29)):*

We should limit the try to be as small as possible. `protect` should be lifted out.
</blockquote></details>

Moved out
___
*[`android/test/e2e/src/main/kotlin/net/mullvad/mullvadvpn/test/e2e/misc/Networking.kt` line 9 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIPevLzEXSPTmCvskps:-OImG2720-L2C7ebqV_W:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1c6696e3b1ad908fd2967056de2ac3685e3577f5/android/test/e2e/src/main/kotlin/net/mullvad/mullvadvpn/test/e2e/misc/Networking.kt#L9)):*

nit: Unnecessary change?
</blockquote></details>

Done

MarkusPettersson98 (commented):
> Reviewed 1 of 1 files at r5, 1 of 6 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 6 of 10 files reviewed, 5 unresolved discussions (waiting on @dlon, @hulthe, @Pururun, and @Rawa)

hulthe (commented):
> Reviewed 2 of 6 files at r6, 1 of 1 files at r7, 1 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 7 of 10 files reviewed, 5 unresolved discussions (waiting on @dlon, @Pururun, and @Rawa)

Rawa (commented):
> Reviewed 1 of 1 files at r5, 2 of 6 files at r6, 1 of 1 files at r7, 2 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 8 of 10 files reviewed, 3 unresolved discussions (waiting on @dlon and @Pururun)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 49 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF14PSEOdb8DTsrCvG:-OK78R-l9ZZAXJM6SizC:by7ui3s) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L49)):*

Changed it to use the new `defaultRawNetworkStateFlow()` which I guess is as slow?
</blockquote></details>

It uses the same mechanism. I was referring to the point that `rawNetworkState` also has the knowledge if we have internet capability as well. But it would only look at the network that is used by default right now.

Rawa (changes requested):
> Reviewed 1 of 7 files at r4, 1 of 6 files at r6, 1 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 9 of 10 files reviewed, 4 unresolved discussions (waiting on @dlon and @Pururun)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/model/Connectivity.kt` line 4 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OK7Ae0D2AaXposKA15u:-OK7Ae0D2AaXposKA15v:bk69yho) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b3dde0888a3306ee38adeb78fdf6530034a610a/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/model/Connectivity.kt#L4)):*
> ```Kotlin
> 
> sealed class Connectivity {
>     data class Status(val ipv4: Boolean, val ipv6: Boolean) : Connectivity()
> ```

Feels like we should name the variables a bit better. Given I have a `Connectivity.Status.ipv4` it is not clear what it means.

Rawa (commented):
> Reviewed 1 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 4 unresolved discussions (waiting on @dlon and @Pururun)

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 4 unresolved discussions (waiting on @dlon and @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 49 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF14PSEOdb8DTsrCvG:-OKQBR-_5Fa6clsZInI4:b-utd1xj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L49)):*

It uses the same mechanism. I was referring to the point that `rawNetworkState` also has the knowledge if we have internet capability as well. But it would only look at the network that is used by default right now.
</blockquote></details>

Yeah it is a bit of a conflict in that we want to look at the default network if we have IPv4 and IPv6, but we don't really want to rely on the default network to check if we have any internet connection as that is somewhat unreliable.
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/model/Connectivity.kt` line 4 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OK7Ae0D2AaXposKA15u:-OKQBkGm97NgvyL_104x:bwm26sb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b3dde0888a3306ee38adeb78fdf6530034a610a/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/model/Connectivity.kt#L4)):*

Feels like we should name the variables a bit better. Given I have a `Connectivity.Status.ipv4` it is not clear what it means.
</blockquote></details>

Changed it to `ipv4Available`

Rawa (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon and @Pururun)

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 4 unresolved discussions (waiting on @dlon and @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 77 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OKS9D__E4NO5aRzCO-U:-OKT6pbK-KqEvMI6jL7q:b-wv6t1u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c8c4d1d83a115315d200ac0ec1dce194ebb2dd2/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L77)):*

Maybe I'm going in circles, but would this still work when roaming? Our default network would be WiFi and suddenly the underlying network may change right?
</blockquote></details>

I assume the default network state will update with new capabilities.
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 167 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OKS9TAU3BttJHJmkLmN:-OKT6yanAgCeQe7Bz-9n:b-98usgo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c8c4d1d83a115315d200ac0ec1dce194ebb2dd2/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L167)):*

Verify that this aligns with the lingo used for Vpn tunnels, I assumed the Underlying network is the network that the VPN uses, not just any network that is not vpn.
</blockquote></details>

Changed it to `isNotVpn` since that is what we are looking for anyway.
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/model/Connectivity.kt` line 4 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OK7Ae0D2AaXposKA15u:-OKT8FBk79QojIUO5k1l:b-k4eswl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8b3dde0888a3306ee38adeb78fdf6530034a610a/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/model/Connectivity.kt#L4)):*

Maybe `hasIpv4` would be even better? `ipv4Available` makes it sound that it is available but not necessarily active.
</blockquote></details>

Just realized the reason why it was named `ipv4` and `ipv6` was because it reflects a underlying type in the daemon. So changing it leads to crash. So will revert.

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 7 of 10 files reviewed, 3 unresolved discussions (waiting on @dlon and @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 49 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIAw5Ax2siGFXsIeRE7:-OKVx3JCE6CCqMuVj--d:b-v64nh4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/98c09bf16c200c65509edd4412af8ec3f838d26b/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L49)):*

If it's not too much work, maybe "debouncing" the events would make sense. See for example `BurstGuard`
</blockquote></details>

Added debounce (for this and other reasons)
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 77 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OKS9D__E4NO5aRzCO-U:-OKVx8Z2EaXGJDK1LjK-:bjyvaws) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c8c4d1d83a115315d200ac0ec1dce194ebb2dd2/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L77)):*

I assume the default network state will update with new capabilities.
</blockquote></details>

Code changed, new versions works fine when roaming with a debounce.
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 167 at r9](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OKS9TAU3BttJHJmkLmN:-OKVxLlE4dQosEhQduN4:b-h327qj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c8c4d1d83a115315d200ac0ec1dce194ebb2dd2/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L167)):*

Changed it to `isNotVpn` since that is what we are looking for anyway.
</blockquote></details>

Changed it again to `isVpn`

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 7 of 10 files reviewed, 3 unresolved discussions (waiting on @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt` line 24 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIF1yZZ7w9TbRSN29a0:-OKWQYYU79X5vHKfdjEm:bjzuigb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/666d22b558535aff046d473386707404ba77a6d9/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IPAvailabilityUtils.kt#L24)):*

I'll leave this unresolved until that has been confirmed.
</blockquote></details>

Tested on 4 devices with Android 11, 14 and 15 and seems to work fine.

kl (changes requested):
> Reviewed 1 of 1 files at r5, 3 of 6 files at r6, 1 of 1 files at r7, 1 of 3 files at r8, 2 of 3 files at r10, 4 of 4 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 6 unresolved discussions (waiting on @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IpUtils.kt` line 24 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OKesLmtAUGO2TMMZ5hi:-OKesLmtAUGO2TMMZ5hj:b2fbfvu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c3dc2c7674ba20eb404492352f95067bba5b69f/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IpUtils.kt#L24)):*
> ```Kotlin
>     // If the ip version is not supported on the underlying network it will trigger a socket
>     // exception. Otherwise we assume it is available.
>     private inline fun <reified T : InetAddress> hasIpVersion(
> ```

Does not need to be `inline` and `reified` IMO.
___
*[`talpid-core/src/connectivity_listener.rs` line 126 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OKexT_A4Ry-2mDu_f4V:-OKexT_A4Ry-2mDu_f4W:b-a4ao1a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c3dc2c7674ba20eb404492352f95067bba5b69f/talpid-core/src/connectivity_listener.rs#L126)):*
> ```XML
>             .expect("Missing isConnected")
>             .l()
>             .expect("isConnected is not an object");
> ```

These two `.expect` calls will panic when the previous code returned an error.

To prevent panic we could do this instead:

```
        let is_connected = env.call_method(
            self.android_listener.as_obj(),
            "isConnected",
            "()Lnet/mullvad/talpid/model/Connectivity;",
            &[],
        );

        let is_connected = match is_connected {
            Ok(JValue::Object(object)) => object,
            value => {
                return Err(Error::InvalidMethodResult(
                    "ConnectivityListener",
                    "isConnected",
                    format!("{:?}", value),
                ))
            }
        };

        Ok(Connectivity::from_java(&env, is_connected))
```
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/ConnectivityManagerUtil.kt` line 138 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OKerrwM9g8IvsuqPZel:-OKerrwM9g8IvsuqPZem:b-gr74mn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c3dc2c7674ba20eb404492352f95067bba5b69f/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/ConnectivityManagerUtil.kt#L138)):*
> ```Kotlin
>         // the underlying network
>         Connectivity.Status(
>             IpUtils.hasIPv4(protect = { protect(it) }),
> ```

Can be simplified to 
```
hasIPv4(protect),
hasIPv6(protect),
```

Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: 5 of 12 files reviewed, 6 unresolved discussions (waiting on @hulthe, @kl, @MarkusPettersson98, and @Rawa)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/ConnectivityManagerUtil.kt` line 138 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OKerrwM9g8IvsuqPZel:-OKfiePX6EhEOOAo57ik:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c3dc2c7674ba20eb404492352f95067bba5b69f/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/ConnectivityManagerUtil.kt#L138)):*

Can be simplified to 
```
hasIPv4(protect),
hasIPv6(protect),
```
</blockquote></details>

Done
___
*[`talpid-core/src/connectivity_listener.rs` line 126 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OKexT_A4Ry-2mDu_f4V:-OKfil560iIlH_4VbjxF:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c3dc2c7674ba20eb404492352f95067bba5b69f/talpid-core/src/connectivity_listener.rs#L126)):*

These two `.expect` calls will panic when the previous code returned an error.

To prevent panic we could do this instead:

```
        let is_connected = env.call_method(
            self.android_listener.as_obj(),
            "isConnected",
            "()Lnet/mullvad/talpid/model/Connectivity;",
            &[],
        );

        let is_connected = match is_connected {
            Ok(JValue::Object(object)) => object,
            value => {
                return Err(Error::InvalidMethodResult(
                    "ConnectivityListener",
                    "isConnected",
                    format!("{:?}", value),
                ))
            }
        };

        Ok(Connectivity::from_java(&env, is_connected))
```
</blockquote></details>

Done.
___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IpUtils.kt` line 24 at r12](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OKesLmtAUGO2TMMZ5hi:-OKfifZs6AqiY1dVpoXe:b-1mbpqt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/6c3dc2c7674ba20eb404492352f95067bba5b69f/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/util/IpUtils.kt#L24)):*

Does not need to be `inline` and `reified` IMO.
</blockquote></details>

Done (when @Rawa has pushed his code)

dlon (approved):
> Reviewed 1 of 6 files at r6, 1 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: all files reviewed, 3 unresolved discussions (waiting on @kl)

___
*[`android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt` line 49 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-OIAw5Ax2siGFXsIeRE7:-OKjlDyY3W1XtHFY9mPt:b-8fx2u7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/98c09bf16c200c65509edd4412af8ec3f838d26b/android/lib/talpid/src/main/kotlin/net/mullvad/talpid/ConnectivityListener.kt#L49)):*

Added debounce (for this and other reasons)
</blockquote></details>

:+1:

Rawa (approved):
> Reviewed 2 of 2 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Rawa (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7577#-:-OKkPS7S2DQ3J-JV6g_P:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7600, opened 2025-02-05
Unify daemon app version types

kl:
Android now has alpha builds which currently crash because the daemon version parsing code does not handle version names with "alpha" in them.

This commit removes `ParsedAppVersion` in favor of `mullvad_version::Version`. This ensures that we have one place in the code base the parses version strings, and also fixes the issue where `ParsedAppVersion` didn't handle alpha versions, which `mullvad_version::Version` currenty supports.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)


albin-mullvad (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @kl)

___
*[`mullvad-types/src/version.rs` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIL7a0w7d8s1zMLD7k9:-OIL7a0x8i2p8altgHqy:bb3u3kr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a51e87ec8c48085e8e95c2af8640234aac7a2c90/mullvad-types/src/version.rs#L16)):*
> ```XML
> // alpha builds, so the code in this file will simply parse alpha versions as beta versions, so
> // as far as the Rust code is concerned there is no difference between -alpha and -beta.
> static BETA_REGEX: LazyLock<Regex> = LazyLock::new(|| {
> ```

Shouldn't we rename this one to something like `PRE_STABLE_REGEX` to match the naming in `mullvad-version`?



_Code quote:_
```XML
BETA_REGEX
```

kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @albin-mullvad)

___
*[`mullvad-types/src/version.rs` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIL7a0w7d8s1zMLD7k9:-OIL8tYY3ycR7hkKtNJb:bev9kk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a51e87ec8c48085e8e95c2af8640234aac7a2c90/mullvad-types/src/version.rs#L16)):*

Shouldn't we rename this one to something like `PRE_STABLE_REGEX` to match the naming in `mullvad-version`?
</blockquote></details>

We could, the reason I didn't is that the version info gets parsed into this struct (which I have made no changes to):

```rust
/// Parses a version string into a type that can be used for comparisons.
#[derive(Eq, PartialEq, Debug, Clone)]
pub enum ParsedAppVersion {
    Stable(u32, u32),
    Beta(u32, u32, u32),
    Dev(u32, u32, Option<u32>, String),
}
```

So the remaining code knows what a "beta" is but has no concept of a an alpha. But we could change the name of the regex variable but it will always only be used to parse a `ParsedAppVersion::Beta`.

I don't have a strong opinion either way. Should I change the name? Another option is to add an `Alpha` variant to `ParsedAppVersion` but then this PR becomes more invloved as then we have to update all places where `ParsedAppVersion` is used in the Rust code.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 1 unresolved discussion (waiting on @albin-mullvad)

___
*[`mullvad-types/src/version.rs` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIL7a0w7d8s1zMLD7k9:-OIM50WDDUKXWSgAeBc-:b-bta6ko) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a51e87ec8c48085e8e95c2af8640234aac7a2c90/mullvad-types/src/version.rs#L16)):*

We could, the reason I didn't is that the version info gets parsed into this struct (which I have made no changes to):

```rust
/// Parses a version string into a type that can be used for comparisons.
#[derive(Eq, PartialEq, Debug, Clone)]
pub enum ParsedAppVersion {
    Stable(u32, u32),
    Beta(u32, u32, u32),
    Dev(u32, u32, Option<u32>, String),
}
```

So the remaining code knows what a "beta" is but has no concept of a an alpha. But we could change the name of the regex variable but it will always only be used to parse a `ParsedAppVersion::Beta`.

I don't have a strong opinion either way. Should I change the name? Another option is to add an `Alpha` variant to `ParsedAppVersion` but then this PR becomes more invloved as then we have to update all places where `ParsedAppVersion` is used in the Rust code.
</blockquote></details>

To be honest, I think the last suggestion sounds like the right one. "Prestable" is probably fine as well, though you'd still have to implement `Ord` somehow. Pretending they're beta versions is not ideal, IMO.

I don't think that it would affect a lot of code outside of this module if you did either. Seems like we only rely on `Ord`

albin-mullvad (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[`mullvad-types/src/version.rs` line 16 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIL7a0w7d8s1zMLD7k9:-OIMbMUU3O1P1chqNoX9:bbsk86k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a51e87ec8c48085e8e95c2af8640234aac7a2c90/mullvad-types/src/version.rs#L16)):*

To be honest, I think the last suggestion sounds like the right one. "Prestable" is probably fine as well, though you'd still have to implement `Ord` somehow. Pretending they're beta versions is not ideal, IMO.

I don't think that it would affect a lot of code outside of this module if you did either. Seems like we only rely on `Ord`
</blockquote></details>

@dlon we discussed it a bit with @faern and decided to try to unify the version structs and logic in `mullvad-version` and `mullvad-types`. So this PR will likely be reworked.

dlon (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 1 unresolved discussion (waiting on @kl)

___
*[`mullvad-version/src/lib.rs` line 159 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIQibdgDiBoMnvP7WgD:-OIQk16k2Q3mjpK98beT:bav4cdo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5060ed68befda6e91ad951f849cafdbdfdb0b057/mullvad-version/src/lib.rs#L159)):*

Is this correct? The beta will be "lower" than any stable version, even if the year is higher, which isn't how it was implemented before. And it assumes that all stable versions are equal.
</blockquote></details>

I think I'm wrong. Could we add tests?

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 3 unresolved discussions (waiting on @kl)

albin-mullvad (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

albin-mullvad (approved):

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`mullvad-version/src/lib.rs` line 159 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIQibdgDiBoMnvP7WgD:-OIRc02OD-Zjl6Qjb-Dg:b-dhlgu0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5060ed68befda6e91ad951f849cafdbdfdb0b057/mullvad-version/src/lib.rs#L159)):*

I think this should be equivalent to the previous Ord implementation. There are tests in the `test\_version\_upgrade\_suggestions` function in `version_check.rs` that indirectly test the Ord implementation. Do you think they are enough or do we want more tests?
</blockquote></details>

I would prefer some more direct tests of the `Ord` implementation, but I think this is good.

faern (changes requested):
> Reviewed 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 3 unresolved discussions (waiting on @kl)

___
*[`mullvad-version/src/lib.rs` line 34 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIUCpDH8gwxTAuwJVlA:-OIUCpDH8gwxTAuwJVlB:bf6soqo) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L34)):*
> ```XML
> 
>     pub fn is_stable(&self) -> bool {
>         !self.is_alpha() && !self.is_beta() && !self.is_dev()
> ```

Slightly nitpicky. But would this not be simpler as `self.pre_stable.is_none() && self.dev.is_none()`? Shorter to read, and stays constant no matter how many prestable types we hypothetically add.
___
*[`mullvad-version/src/lib.rs` line 120 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIUDU67-CneimXnQKMk:-OIUDU67-CneimXnQKMl:b-weisy2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L120)):*
> ```XML
>             .as_str()
>             .parse()
>             .unwrap();
> ```

We should probably return an `Err` instead of panicking if trying to parse a version string where the year (or the other integers) happens to not be integers. Please also add at least one test for this type of parsing error case :pray: 
___
*[`mullvad-version/src/lib.rs` line 149 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIUF1GU3txHK9CRRnPJ:-OIUF1GU3txHK9CRRnPK:b-bj7qa3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L149)):*
> ```XML
> }
> 
> impl Ord for Version {
> ```

Please add unit tests for the ordering. This is not trivial code and lends itself well to being unit tested.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 5 unresolved discussions (waiting on @kl)

___
*[`mullvad-version/src/lib.rs` line 154 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIULjU904Q-rQXcyY68:-OIULjUAACzY-yFTWFVo:b-uy42y1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L154)):*
> ```XML
>             // dev > stable|beta|alpha
>             match (self.is_dev(), other.is_dev()) {
>                 (true, false) => Ordering::Greater,
> ```

Does this not imply that `2025.1-beta1-dev-abc` is greater than `2025.1-dev-abc`? This should probably not be true since `2025.1-beta1 < 2025.1`. Should the dev suffix not only affect the ordering if *all other* components are equal?
___
*[`mullvad-version/src/lib.rs` line 174 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIUMWB58toSeCjyokpO:-OIUMWB58toSeCjyokpP:b-pkond8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L174)):*
> ```XML
>         } else {
>             // both are alpha
>             Ordering::Equal
> ```

In this branch, we must compare the `pre_stable` numbers. Since `2025.1-alpha2` should be `Greater` than `2025.1-alpha1` (but less than `2025.1-beta1`).

kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 4 unresolved discussions (waiting on @faern)

___
*[`mullvad-version/src/lib.rs` line 159 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIQibdgDiBoMnvP7WgD:-OIVRAHV6fxLblurv9kZ:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5060ed68befda6e91ad951f849cafdbdfdb0b057/mullvad-version/src/lib.rs#L159)):*

I would prefer some more direct tests of the `Ord` implementation, but I think this is good.
</blockquote></details>

Done
___
*[`mullvad-version/src/lib.rs` line 34 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIUCpDH8gwxTAuwJVlA:-OIVN9mDCVBc9rMr5zyY:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L34)):*

Slightly nitpicky. But would this not be simpler as `self.pre_stable.is_none() && self.dev.is_none()`? Shorter to read, and stays constant no matter how many prestable types we hypothetically add.
</blockquote></details>

Done
___
*[`mullvad-version/src/lib.rs` line 120 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIUDU67-CneimXnQKMk:-OIVNYdA4yvppV-T-GRM:b-qe2bim) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L120)):*

We should probably return an `Err` instead of panicking if trying to parse a version string where the year (or the other integers) happens to not be integers. Please also add at least one test for this type of parsing error case :pray:
</blockquote></details>

The regex specifies that the year and incremental version must be integers of a fixed length, so if you give it an invalid year it will return an error (not panic) when matching the regex. I changed the tests for this so that it is more apparent that this function doesn't panic.
___
*[`mullvad-version/src/lib.rs` line 149 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIUF1GU3txHK9CRRnPJ:-OIVQgLR5kGPbaldCeTi:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L149)):*

Please add unit tests for the ordering. This is not trivial code and lends itself well to being unit tested.
</blockquote></details>

Done
___
*[`mullvad-version/src/lib.rs` line 154 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIULjU904Q-rQXcyY68:-OIVQzyx7sRdjWx0vS5p:b9n772r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L154)):*

Does this not imply that `2025.1-beta1-dev-abc` is greater than `2025.1-dev-abc`? This should probably not be true since `2025.1-beta1 < 2025.1`. Should the dev suffix not only affect the ordering if *all other* components are equal?
</blockquote></details>

Yes good catch!
I added tests and changed the version struct a little. Now all versions types (stable, beta and alpha) can optionally be `dev` versions, and `dev` is only checked at the very end of the comparison chain.
___
*[`mullvad-version/src/lib.rs` line 174 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIUMWB58toSeCjyokpO:-OIVRD7u1ew7vrplwY47:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L174)):*

In this branch, we must compare the `pre_stable` numbers. Since `2025.1-alpha2` should be `Greater` than `2025.1-alpha1` (but less than `2025.1-beta1`).
</blockquote></details>

Done

Rawa (approved):
> Reviewed 2 of 4 files at r2, 1 of 1 files at r3, 2 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 5 unresolved discussions (waiting on @faern)

___
*[`mullvad-version/src/main.rs` line 78 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIionQI9-Eoq-dNVPGe:-OIionQI9-Eoq-dNVPGf:b-ubm5hj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1648d040ef193fae27cdf7da672af8231c455bb8/mullvad-version/src/main.rs#L78)):*
> ```XML
>     };
> 
>     let decade_and_year = version.year % 2000;
> ```

nit: This will work until 2100 then it will start returning 100, 101 etc. It should be:
```
version.year % 100
```

Rawa (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 5 unresolved discussions (waiting on @faern and @kl)

kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: 3 of 6 files reviewed, 5 unresolved discussions (waiting on @dlon, @faern, and @Rawa)

___
*[`mullvad-version/src/main.rs` line 78 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIionQI9-Eoq-dNVPGe:-OIixRWfDmAwzplIwwBm:bl8angm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1648d040ef193fae27cdf7da672af8231c455bb8/mullvad-version/src/main.rs#L78)):*

nit: This will work until 2100 then it will start returning 100, 101 etc. It should be:
```
version.year % 100
```
</blockquote></details>

Nice catch! :grin:  Fixed.

Rawa (dismissed):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: 2 of 6 files reviewed, 4 unresolved discussions (waiting on @dlon and @faern)

___
*[`mullvad-version/src/main.rs` line 78 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIionQI9-Eoq-dNVPGe:-OIj0tjG-yUZDuYH7w-V:b-8fx2u7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1648d040ef193fae27cdf7da672af8231c455bb8/mullvad-version/src/main.rs#L78)):*

Nice catch! :grin:  Fixed.
</blockquote></details>

:+1:

faern (changes requested):
> Reviewed 1 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 4 unresolved discussions

___
*[`mullvad-version/src/lib.rs` line 34 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIUCpDH8gwxTAuwJVlA:-OIVbkeyB96s8PwyR6hu:b8885s9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L34)):*

Done
</blockquote></details>

Now you don't check `self.dev`. `2025.1-dev-abc` should not be considered stable, right? Or is a dev-version of a "stable" version actually stable? I doubt that's the logic we want, but we have to check the places where this is used.
___
*[`mullvad-version/src/lib.rs` line 120 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIUDU67-CneimXnQKMk:-OIVbyI706JKX-uc6Ryi:b-jp7wk4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L120)):*

The regex specifies that the year and incremental version must be integers of a fixed length, so if you give it an invalid year it will return an error (not panic) when matching the regex. I changed the tests for this so that it is more apparent that this function doesn't panic.
</blockquote></details>

You are correct! My bad. We can safely unwrap here. Great
___
*[`mullvad-version/src/lib.rs` line 154 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIULjU904Q-rQXcyY68:-OIVep_l6uJnGhvXPDUQ:bccpe83) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L154)):*

Yes good catch!
I added tests and changed the version struct a little. Now all versions types (stable, beta and alpha) can optionally be `dev` versions, and `dev` is only checked at the very end of the comparison chain.
</blockquote></details>

Awesome!
___
*[`mullvad-version/src/lib.rs` line 61 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIp_il95crULUz8h-tU:-OIp_il95crULUz8h-tV:b-sb0nls) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3a9037e27a14141cc791344144419d1e397efa1a/mullvad-version/src/lib.rs#L61)):*
> ```XML
>     /// The following order is used: stable > beta > alpha
>     /// If two versions are identical except for one being a dev version
>     /// (ending with the suffix -dev-[SHA]), he dev version is considered greater than
> ```

s/he/the
___
*[`mullvad-version/src/lib.rs` line 64 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIp_q_l6XMzqTSJr3Iu:-OIp_q_l6XMzqTSJr3Iv:b-p355jt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3a9037e27a14141cc791344144419d1e397efa1a/mullvad-version/src/lib.rs#L64)):*
> ```XML
>     /// the non-dev version, but if both are dev versions they are considered equal
>     /// since the dev version SHA is not ordered.
>     pub fn is_later_version_than(&self, other: &Self) -> bool {
> ```

Can't this return `std::cmp::Ordering`? We just need it to not implement `PartialOrd` to not break the rules about `PartialOrd + PartialEq`, we can still use the ordering type? Doing so will likely make it easier to later use methods such as `slice::sort_by` and similar to implement sorting.
___
*[`mullvad-version/src/main.rs` line 78 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIVcmAw-p329dYjQINP:-OIVcmAw-p329dYjQINQ:b-7mw1cg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f26020bd6817276746ee00adce83dcbd06f3f8f/mullvad-version/src/main.rs#L78)):*
> ```XML
>     };
> 
>     let decade_and_year = version.year % 2000;
> ```

Nitpick. Hard to name variable. But `year` here is used as if it refers to the last digit in `2025` only, which is a bit weird IMO. How about `year_last_two_digits`?
___
*[`mullvad-version/src/main.rs` line 107 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIVdPrt1vsoEWSUYL5M:-OIVdPrt1vsoEWSUYL5N:bhca7nf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f26020bd6817276746ee00adce83dcbd06f3f8f/mullvad-version/src/main.rs#L107)):*
> ```XML
> /// On Windows we currently support the following versions: stable, beta and dev.
> fn is_valid_windows_version(version: &Version) -> bool {
>     version.is_stable() || version.is_beta()
> ```

Why were dev versions removed from this check?

faern (commented):
> Reviewed 1 of 4 files at r2, 1 of 1 files at r6, 2 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 4 unresolved discussions (waiting on @kl)

kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: 2 of 6 files reviewed, 2 unresolved discussions (waiting on @faern and @Rawa)

___
*[`mullvad-version/src/lib.rs` line 34 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIUCpDH8gwxTAuwJVlA:-OItED-i-9y6x7g0z-tf:bnzs49h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/37c32c7e026b8abbfa55dc81ebf86da9f953f72b/mullvad-version/src/lib.rs#L34)):*

Now you don't check `self.dev`. `2025.1-dev-abc` should not be considered stable, right? Or is a dev-version of a "stable" version actually stable? I doubt that's the logic we want, but we have to check the places where this is used.
</blockquote></details>

Yes, I changed the logic to be that way (stable-dev is still stable, just as beta-dev is beta and alpha-dev is alpha). This made implementing the ord logic simpler. I have checked everywhere it is used as part of removing the Ord impl.
___
*[`mullvad-version/src/lib.rs` line 61 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIp_il95crULUz8h-tU:-OItGis81gMgddnGwMdg:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3a9037e27a14141cc791344144419d1e397efa1a/mullvad-version/src/lib.rs#L61)):*

s/he/the
</blockquote></details>

Done
___
*[`mullvad-version/src/lib.rs` line 64 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIp_q_l6XMzqTSJr3Iu:-OItGl-jEABlhOply_GU:b-896fjx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3a9037e27a14141cc791344144419d1e397efa1a/mullvad-version/src/lib.rs#L64)):*

Can't this return `std::cmp::Ordering`? We just need it to not implement `PartialOrd` to not break the rules about `PartialOrd + PartialEq`, we can still use the ordering type? Doing so will likely make it easier to later use methods such as `slice::sort_by` and similar to implement sorting.
</blockquote></details>

Done
___
*[`mullvad-version/src/main.rs` line 78 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIVcmAw-p329dYjQINP:-OItH5hW2Z3DVIxOLf_2:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f26020bd6817276746ee00adce83dcbd06f3f8f/mullvad-version/src/main.rs#L78)):*

Nitpick. Hard to name variable. But `year` here is used as if it refers to the last digit in `2025` only, which is a bit weird IMO. How about `year_last_two_digits`?
</blockquote></details>

Done
___
*[`mullvad-version/src/main.rs` line 107 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OIVdPrt1vsoEWSUYL5M:-OItEi858n5rrJTPCfqP:b-v0tlxq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f26020bd6817276746ee00adce83dcbd06f3f8f/mullvad-version/src/main.rs#L107)):*

Why were dev versions removed from this check?
</blockquote></details>

Because the logic got changed so that a stable-dev is now considered stable.

faern (commented):
> Reviewed 2 of 4 files at r9.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @Rawa)

faern (dismissed):
> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: all files reviewed, 1 unresolved discussion

___
*[`mullvad-version/src/lib.rs` line 48 at r10](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600#-OJOaNf95bG6AEvohApJ:-OJOaNf95bG6AEvohApK:b-klo9ik) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bed9178ce44cfecf97bdef195fb5dac48bb3fce6/mullvad-version/src/lib.rs#L48)):*
> ```XML
>     /// dev version (i.e. contains a -dev suffix) and is not a pre-release version (i.e. alpha
>     /// or beta). Example: 2025.10
>     pub fn is_stable(&self) -> bool {
> ```

Nothing calls this except our own unit tests. So even if the implementation looks correct, I'm a bit hesitant to expose an API that is slightly unclear and **no one needs**.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7600)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7636, opened 2025-02-10
Add setup-rust script and git hook

kl:
The setup-rust helper script installs the default targets and components that each platform needs with rustup.

The setup-rust-post-checkout script is a git hook (needs to be copied to .git/hooks/post-checkout, which can be done manually or by running `setup-rust install-hook`) that detects when the rust-toolchain.toml file has changed and installs the default targets/components (and/or extra targets/components that the user has specified in the hook script).

So say we have a new desktop developer and they are setting up their environment. They could run the following:
`scripts/setup-rust desktop && scripts/setup-rust install-hook desktop` after which things Should Just Work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)

### DIFF
--- a/BuildInstructions.md
+++ b/BuildInstructions.md
@@ -9,6 +9,15 @@ on your platform please submit an issue or a pull request.
 ## All platforms
 
 - Get the latest **stable** Rust toolchain via [rustup.rs](https://rustup.rs/).
+  - Install default targets and components needed for desktop
+    ```bash
+    ./scripts/setup-rust desktop
+     ```
+  - (Optional) Run the following to install a git `post-checkout` hook that will automatically
+    run the `setup-rust` script when the Rust version specified in the `rust-toolchain.toml` file changes:
+    ```bash
+    .scripts/setup-rust install-hook
+    ```
 
 - You need Node.js and npm. You can find the exact versions in the `volta` section of
   `desktop/package.json`. The toolchain is managed by volta.
--- a/android/BuildInstructions.md
+++ b/android/BuildInstructions.md
@@ -151,7 +151,13 @@ environment variables:
 
 - Install Android targets
   ```bash
-  rustup target add aarch64-linux-android armv7-linux-androideabi i686-linux-android x86_64-linux-android
+  ./scripts/setup-rust android
+  ```
+
+- (Optional) Run the following to install a git `post-checkout` hook that will automatically
+  run the `setup-rust` script when the Rust version specified in the `rust-toolchain.toml` file changes:
+  ```bash
+  .scripts/setup-rust install-hook
   ```
 
 #### 6. Download wireguard-go-rs submodule
--- a/ios/BuildInstructions.md
+++ b/ios/BuildInstructions.md
@@ -10,8 +10,16 @@ Rust should be installed via [rustup](https://rustup.rs). Once rust is
 installed, do not forget to install the iOS targets:
 
 ```bash
-rustup target install aarch64-apple-ios aarch64-apple-ios-sim
+./scripts/setup-rust ios
 ```
+
+(Optional) Run the following to install a git `post-checkout` hook that will automatically
+run the `setup-rust` script when the Rust version specified in the `rust-toolchain.toml` file changes:
+
+```bash
+.scripts/setup-rust install-hook
+```
+
 You only need to install the ARM simulator target, which matches the current Apple Silicon architecture.
 
 Once both rust and go are installed, ensure they are available in your path.
--- a/scripts/setup-rust
+++ b/scripts/setup-rust
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+#
+# Installs the default toolchains and components for different platforms.
+# To use this script rustup must be installed first.
+
+set -eu
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+cd "$SCRIPT_DIR/.."
+
+source scripts/utils/log
+
+ANDROID_TARGETS="x86_64-linux-android i686-linux-android aarch64-linux-android armv7-linux-androideabi"
+ANDROID_COMPONENTS="rust-analyzer"
+
+DESKTOP_TARGETS="x86_64-pc-windows-msvc x86_64-pc-windows-gnu i686-pc-windows-msvc"
+DESKTOP_COMPONENTS="rust-analyzer"
+
+IOS_TARGETS="aarch64-apple-ios-sim aarch64-apple-ios x86_64-apple-ios"
+IOS_COMPONENTS="rust-analyzer"
+
+function main {
+    if [[ $# -eq 0 ]]; then
+        print_usage
+        exit 1
+    fi
+
+    case "$1" in
+    "android")
+        setup "Android" "$ANDROID_TARGETS" "$ANDROID_COMPONENTS"
+        ;;
+    "desktop")
+        setup "Desktop" "$DESKTOP_TARGETS" "$DESKTOP_COMPONENTS"
+        ;;
+    "ios")
+        setup "Android" "$IOS_TARGETS" "$IOS_COMPONENTS"
+        ;;
+    "install-hook")
+        install_hook
+        ;;
+    "--help")
+        print_usage
+        exit 0
+        ;;
+    *)
+        log_error "Invalid argument: \`$1\`"
+        print_usage
+        exit 1
+        ;;
+    esac
+}
+
+function print_usage {
+    log "Setup default Rust targets and components for different platforms"
+    log ""
+    log "Usage: setup-rust android|desktop|ios|install-hook"
+    log "  android                    Run Android-specific setup"
+    log "  desktop                    Run Desktop-specific setup"
+    log "  ios                        Run iOS-specific setup"
+    log "  install-hook               Copies the setup-rust-post-checkout file to .git/hooks/post-checkout"
+}
+
+function setup {
+    local platform=$1
+    local targets=$2
+    local components=$3
+
+    log "Installing default Rust targets/components"
+    log "platform: $platform"
+    log "targets: $targets"
+    log "components: $components"
+    # shellcheck disable=SC2086
+    rustup target add $targets
+    # shellcheck disable=SC2086
+    rustup component add $components
+}
+
+function install_hook {
+    local hook=$SCRIPT_DIR/../.git/hooks/post-checkout
+    if [[ -f "$hook" ]]; then
+        log_error "$(realpath "$hook") file already exists - will not overwrite"
+        exit 1
+    else
+        cp "$SCRIPT_DIR/setup-rust-post-checkout" "$hook"
+        chmod +x "$hook"
+        log "Hook installed. You must now set the environment variable MULLVAD_SETUP_PLATFORM to "
+        log "\`android\`, \`desktop\` or \`ios\` in your shell environment."
+    fi
+}
+
+# Run script
+main "$@"
--- a/scripts/setup-rust-post-checkout
+++ b/scripts/setup-rust-post-checkout
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+#
+# To install the hook copy this file to `.git/hooks/post-checkout` or run
+# scripts/setup-rust install_hook
+
+set -u
+
+# Set to "android", "desktop", or "ios".
+MULLVAD_SETUP_PLATFORM="${MULLVAD_SETUP_PLATFORM:-}"
+
+SETUP_RUST_SCRIPT="scripts/setup-rust"
+
+if [[ ! -f "$SETUP_RUST_SCRIPT" ]]; then
+    exit 0
+fi
+
+if [[ -z ${MULLVAD_SETUP_PLATFORM+x} ]]; then
+    echo "MULLVAD_SETUP_PLATFORM is not set, must be set to " >&2
+    echo "\`android\`, \`desktop\` or \`ios\`" >&2
+    exit 1
+fi
+
+git diff-tree --exit-code "$1".."$2" --quiet -- rust-toolchain.toml
+
+# Exit code 1 means there was a change, 0 means no change.
+if [[ $? -eq 1 ]]; then
+    $SETUP_RUST_SCRIPT "$MULLVAD_SETUP_PLATFORM"
+fi


albin-mullvad (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: 0 of 2 files reviewed, 3 unresolved discussions

___
*[`scripts/setup-rust` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIoDgJl9POQX24RsNie:-OIoDgJl9POQX24RsNif:bkt58qd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust#L26)):*
> ```
>   log "Options:"
>   log "--print-targets        Only print the targets that would be added"
>   log "--print-components     Only print the components that would be added"
> ```

Do we need to distinguish these? Imo it would simlify things to just use a `--dry-run` working like this: https://unix.stackexchange.com/a/433806



_Code quote:_
```
  log "--print-targets        Only print the targets that would be added"
  log "--print-components     Only print the components that would be added"
```
___
*[`scripts/setup-rust` line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIoCdyI1oWRzjUFnFOt:-OIoCdyI1oWRzjUFnFOu:b-kqw4bq) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust#L30)):*
> ```
> 
> function setup_android {
>   local targets="x86_64-linux-android i686-linux-android aarch64-linux-android armv7-linux-androideabi"
> ```

It would be nice to define the target/component list as variables at the top of the script, e.g. `ANDROID_COMPONENTS="rust-analyzer"` so that it's easy to get an overview of the platform components/targets without digging into the functions.
___
*[`scripts/setup-rust` line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIo9kPh-KO5kg-sKDG7:-OIo9kPh-KO5kg-sKDG8:b-tezqeu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust#L80)):*
> ```
> }
> 
> if [ $# -eq 0 ]; then
> ```

I believe we usually put the input/argument parsing at the top in a `main` function structure that we call from the bottom of the script using a `main "$@"` call. I'm referring to the scripts where attempt a function structure. Apart from keeping this scripts consistent another benefit of doing so is that it's easy/quick to get an overview of the main entry points. Another benefit is to keep the argument parsing and usage print next to each other in order to help visually ensure they are up-to-date with each other. What do think about making such a change? I'll gladly discuss/explain further it outside of the review!

Example:
https://github.com/mullvad/mullvadvpn-app/blob/ab3bd2bcfa26832e014efa7c84bae4c01221a00a/scripts/localization#L11

albin-mullvad (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @kl)

___
*[`scripts/setup-rust-post-checkout` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIo_TJBApMeG-MA8EFR:-OIo_TJBApMeG-MA8EFS:bt4zqu8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust-post-checkout#L17)):*
> ```
> EXTRA_TARGETS=""
> # Any extra components outside of the ones defined in `setup-rust` that should be installed.
> EXTRA_COMPONENTS=""
> ```

Are these not more convenient to define in your shell environment to avoid hook modifications? :thinking: That would also allow for automatic hook updates, where the hook updates itself (or is symlinked).



_Code quote:_
```
# Set to "android", "desktop", or "ios" or leave empty.
# If empty this hook will not call the `setup-rust` script and only install the EXTRA_TARGETS
# and EXTRA_COMPONENTS.
PLATFORM=""
# Any extra targets outside of the ones defined in `setup-rust` that should be installed.
EXTRA_TARGETS=""
# Any extra components outside of the ones defined in `setup-rust` that should be installed.
EXTRA_COMPONENTS=""
```
___
*[`scripts/setup-rust-post-checkout` line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIoaeas1zxXgaPLH2nF:-OIoaeas1zxXgaPLH2nG:b-c4d5kd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust-post-checkout#L27)):*
> ```
> CHANGED=$(git diff "$1" "$2" --stat -- rust-toolchain.toml | wc -l)
> if [ "$CHANGED" -gt 0 ];
> then
> ```

This hook includes more complexity than I expected, such as all the `wc` based calculations. Is there any way we could simplify it?

kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad)

___
*[`scripts/setup-rust` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIoDgJl9POQX24RsNie:-OIovf_j-THKotP2ZIHa:blmjdx2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust#L26)):*

Do we need to distinguish these? Imo it would simlify things to just use a `--dry-run` working like this: https://unix.stackexchange.com/a/433806
</blockquote></details>

They are used by the hook to get the number of default targets/components specified for the given platform. This number is used to check if the already installed target/component count is less than the needed target/component count, and only if it is less we call `rustup target add`. This is done so that we only `rustup target add` when necessary, and not every time for example when switching between two branches that have different versions of Rust specified in `rust-toolchain.toml`
___
*[`scripts/setup-rust` line 30 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIoCdyI1oWRzjUFnFOt:-OIoxtHu6tx5pijSpXW0:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust#L30)):*

It would be nice to define the target/component list as variables at the top of the script, e.g. `ANDROID_COMPONENTS="rust-analyzer"` so that it's easy to get an overview of the platform components/targets without digging into the functions.
</blockquote></details>

Done
___
*[`scripts/setup-rust` line 80 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIo9kPh-KO5kg-sKDG7:-OIoy1dh5Kus9BEandMy:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust#L80)):*

I believe we usually put the input/argument parsing at the top in a `main` function structure that we call from the bottom of the script using a `main "$@"` call. I'm referring to the scripts where attempt a function structure. Apart from keeping this scripts consistent another benefit of doing so is that it's easy/quick to get an overview of the main entry points. Another benefit is to keep the argument parsing and usage print next to each other in order to help visually ensure they are up-to-date with each other. What do think about making such a change? I'll gladly discuss/explain further it outside of the review!

Example:
https://github.com/mullvad/mullvadvpn-app/blob/ab3bd2bcfa26832e014efa7c84bae4c01221a00a/scripts/localization#L11
</blockquote></details>

Done
___
*[`scripts/setup-rust-post-checkout` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIo_TJBApMeG-MA8EFR:-OIpAQ1047IFak5ngIa9:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust-post-checkout#L17)):*

Are these not more convenient to define in your shell environment to avoid hook modifications? :thinking: That would also allow for automatic hook updates, where the hook updates itself (or is symlinked).
</blockquote></details>

Done
___
*[`scripts/setup-rust-post-checkout` line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIoaeas1zxXgaPLH2nF:-OIoy84rBminF6pbRnCL:b9duier) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust-post-checkout#L27)):*

This hook includes more complexity than I expected, such as all the `wc` based calculations. Is there any way we could simplify it?
</blockquote></details>

The calculations are needed in order to prevent calling `rustup target add` more than strictly necessary (e.g. in cases where we already have all the targets/components installed but are switching between two branches that have different toolchain versions)

albin-mullvad (commented):
> Reviewed 1 of 2 files at r1, 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: 1 of 2 files reviewed, 2 unresolved discussions

___
*[`scripts/setup-rust` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIoDgJl9POQX24RsNie:-OIpLdBc9F9V5BDqvXk1:b-y1p29t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust#L26)):*

They are used by the hook to get the number of default targets/components specified for the given platform. This number is used to check if the already installed target/component count is less than the needed target/component count, and only if it is less we call `rustup target add`. This is done so that we only `rustup target add` when necessary, and not every time for example when switching between two branches that have different versions of Rust specified in `rust-toolchain.toml`
</blockquote></details>

Aha, I missed that part :thumbsup: 
___
*[`scripts/setup-rust-post-checkout` line 17 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIo_TJBApMeG-MA8EFR:-OIpNLhdEq6OgyYafmiD:b-83sjoh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust-post-checkout#L17)):*

Done
</blockquote></details>

Nice!
___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIpN0Nq2pLFZukIs2y_:-OIpN0Nq2pLFZukIs2ya:b-s304gx):*
Not sure what the best practice is, but maybe we should recommend symlinking in order to easily keep it up-to-date in case it changes over time? :thinking: Not sure how that would work on Windows though. This comment also affect the `install-hook` function.



_Code quote:_
```
  The setup-rust-post-checkout script is a git hook (needs to be
  copied to .git/hooks/post-checkout, which can be done manually or
  by running `setup-rust install-hook`) that detects when
```

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @kl)

___
*[`scripts/setup-rust-post-checkout` line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIoaeas1zxXgaPLH2nF:-OJ38dCF4ECT_bCYK2yB:bkeqkw8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust-post-checkout#L27)):*

The calculations are needed in order to prevent calling `rustup target add` more than strictly necessary (e.g. in cases where we already have all the targets/components installed but are switching between two branches that have different toolchain versions)
</blockquote></details>

You probably don't want to use the git porcelain commands for this, but rather the plumbing tools. You can do the above without `wc` with:
```
git diff-tree --exit-code $previous_commit..$new_commit -- rust-toolchain.toml
```
This exits with exit code `1` if the file has a diff between the commits. Exit code `0` if it has no change.
___
*[`scripts/setup-rust-post-checkout` line 10 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJ372vxCXzfx_d3LHYR:-OJ372vxCXzfx_d3LHYS:b-c37uqb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b10ee2a6e8e207f0dfe4ab5880e9e3c6278d54e/scripts/setup-rust-post-checkout#L10)):*
> ```
> set -eu
> 
> function main() {
> ```

Nitpick. But our coding guidelines says to not use parenthesis on functions in bash. https://github.com/mullvad/coding-guidelines/blob/main/bash.md#functions

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @kl)

___
*[`scripts/setup-rust-post-checkout` line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIoaeas1zxXgaPLH2nF:-OJ39Z7g2syDqwg_I_ky:bcurob) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust-post-checkout#L27)):*

You probably don't want to use the git porcelain commands for this, but rather the plumbing tools. You can do the above without `wc` with:
```
git diff-tree --exit-code $previous_commit..$new_commit -- rust-toolchain.toml
```
This exits with exit code `1` if the file has a diff between the commits. Exit code `0` if it has no change.
</blockquote></details>

Also slap a `--quiet` on it to stay silent. Then you can work with exit codes only and not look at any text output. And `--quiet` implies `--exit-code`. So only `--quiet` is actually needed.

Plumbing tools are nice for programmatic use since they are more or less guaranteed to never change behavior.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @albin-mullvad and @kl)

___
*[`scripts/setup-rust-post-checkout` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJ39xhL53LK73ZzgSTx:-OJ39xhL53LK73ZzgSTy:bba8udp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b10ee2a6e8e207f0dfe4ab5880e9e3c6278d54e/scripts/setup-rust-post-checkout#L30)):*
> ```
> 
>     if [ "$changed" -gt 0 ];
>     then
> ```

Style nitpicking. But I suggest you study our bash coding guidelines: https://github.com/mullvad/coding-guidelines/blob/main/bash.md#formatting

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: 1 of 2 files reviewed, 5 unresolved discussions (waiting on @albin-mullvad and @kl)

___
*[`scripts/setup-rust-post-checkout` line 50 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJ3B4UA5JrcllFBXFKU:-OJ3B4UA5JrcllFBXFKV:b-n9zwkf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b10ee2a6e8e207f0dfe4ab5880e9e3c6278d54e/scripts/setup-rust-post-checkout#L50)):*
> ```
>         fi
> 
>         if [ "$installed_targets_count" -lt "$required_targets_count" ]; then
> ```

Do we really need to count targets and only install things when we have fewer targets? Adding a target that already exists seems to be an extremely cheap operation already. And we will only trigger these things when `rust-toolchain.toml` changes anyway, which will not be very common. So I think we can affort calling `rustup target add ...` a few more times and reduce complexity of this hook by a lot?

It's also not clear to me how the count of targets is a stable check. You can have lots of targets, but just not the correct ones, and it would prevent this hook from installing the correct targets.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @albin-mullvad and @kl)

___
*[`scripts/setup-rust` line 65 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJ3CpHN1I9vstB0m53y:-OJ3CpHN1I9vstB0m53z:bgvdbvp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b10ee2a6e8e207f0dfe4ab5880e9e3c6278d54e/scripts/setup-rust#L65)):*
> ```
> 
> function setup {
>   case "$4" in
> ```

You are in general using a lot of `local` variable definitions, which is great! I think this function could use some as well. Knowing what `$1`, `$2` etc are by name would make it much more readable.

We frequently start our bash functions by assigning all arguments to `local` variables, to give them reasonable names.

kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @faern)

___
*[`scripts/setup-rust` line 65 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJ3CpHN1I9vstB0m53y:-OJNsUCI4ngVos59HaPW:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b10ee2a6e8e207f0dfe4ab5880e9e3c6278d54e/scripts/setup-rust#L65)):*

You are in general using a lot of `local` variable definitions, which is great! I think this function could use some as well. Knowing what `$1`, `$2` etc are by name would make it much more readable.

We frequently start our bash functions by assigning all arguments to `local` variables, to give them reasonable names.
</blockquote></details>

Done
___
*[`scripts/setup-rust-post-checkout` line 27 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OIoaeas1zxXgaPLH2nF:-OJNtCrB4rVuvgPnHGQP:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/42b0348ac9b31d1dcc44f4132f676fbe504bc054/scripts/setup-rust-post-checkout#L27)):*

Also slap a `--quiet` on it to stay silent. Then you can work with exit codes only and not look at any text output. And `--quiet` implies `--exit-code`. So only `--quiet` is actually needed.

Plumbing tools are nice for programmatic use since they are more or less guaranteed to never change behavior.
</blockquote></details>

Done
___
*[`scripts/setup-rust-post-checkout` line 30 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJ39xhL53LK73ZzgSTx:-OJNtAYd0JvyXLOKE-Ij:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b10ee2a6e8e207f0dfe4ab5880e9e3c6278d54e/scripts/setup-rust-post-checkout#L30)):*

Style nitpicking. But I suggest you study our bash coding guidelines: https://github.com/mullvad/coding-guidelines/blob/main/bash.md#formatting
</blockquote></details>

Done.
___
*[`scripts/setup-rust-post-checkout` line 50 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJ3B4UA5JrcllFBXFKU:-OJNtEx1-rK3NCCzd8Bp:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b10ee2a6e8e207f0dfe4ab5880e9e3c6278d54e/scripts/setup-rust-post-checkout#L50)):*

Do we really need to count targets and only install things when we have fewer targets? Adding a target that already exists seems to be an extremely cheap operation already. And we will only trigger these things when `rust-toolchain.toml` changes anyway, which will not be very common. So I think we can affort calling `rustup target add ...` a few more times and reduce complexity of this hook by a lot?

It's also not clear to me how the count of targets is a stable check. You can have lots of targets, but just not the correct ones, and it would prevent this hook from installing the correct targets.
</blockquote></details>

Done.

kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @albin-mullvad and @faern)

___
*[`scripts/setup-rust-post-checkout` line 10 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJ372vxCXzfx_d3LHYR:-OJNtWWIEQo9PLv9AVgX:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/3b10ee2a6e8e207f0dfe4ab5880e9e3c6278d54e/scripts/setup-rust-post-checkout#L10)):*

Nitpick. But our coding guidelines says to not use parenthesis on functions in bash. https://github.com/mullvad/coding-guidelines/blob/main/bash.md#functions
</blockquote></details>

Done.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: all files reviewed, 3 unresolved discussions

___
*[`scripts/setup-rust` line 4 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJNtL608l-LpLU8eOxf:-OJNtL611l_QMphs2OKU:bn7l4hm) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/550c14d51e9127056bf078b5c9bfcfb6eef4a84e/scripts/setup-rust#L4)):*
> ```
> #
> # Installs the default toolchains and components for different platforms.
> # To use this script rustup must be installed first.
> ```

I think we should mention this script/the hook in some documentation. Currently it's not very discoverable. Probably `BuildInstructions.md` in the root (but that is desktop specific)?
___
*[`scripts/setup-rust` line 66 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJNuJhdEOfDywt_YIYy:-OJNuJhdEOfDywt_YIYz:bhchksu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/550c14d51e9127056bf078b5c9bfcfb6eef4a84e/scripts/setup-rust#L66)):*
> ```
>     local PLATFORM=$1
>     local TARGETS=$2
>     local COMPONENTS=$3
> ```

Very nitpicky, and we are a bit inconsistent in the repo overall, but these should ideally be lowercase.
___
*[`scripts/setup-rust` line 75 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJNv3IkCaL8zrgFe5at:-OJNv3IkCaL8zrgFe5au:bwq6cwh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/550c14d51e9127056bf078b5c9bfcfb6eef4a84e/scripts/setup-rust#L75)):*
> ```
>     rustup target add $2
>     # shellcheck disable=SC2086
>     rustup component add $3
> ```

You have already assigned the arguments to local variables, use them by that name instead of argument number.

faern (commented):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: all files reviewed, 2 unresolved discussions (waiting on @kl)

___
*[`android/BuildInstructions.md` line 154 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJOPapv-yOdn64L1jSS:-OJOPapv-yOdn64L1jST:byuksxt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/74d383604671f32533d714dd935487b7483e0ad5/android/BuildInstructions.md#L154)):*
> ```Markdown
> - Install Android targets
>   ```bash
>   ./scripts/setup-rust android
> ```

This is awesome! :star_struck: Great work. Nice to replace the hardcoded list with a script that can be updated without touching the docs.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`scripts/setup-rust` line 4 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJNtL608l-LpLU8eOxf:-OJOVNSMAD9UkkfzjZIU:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/550c14d51e9127056bf078b5c9bfcfb6eef4a84e/scripts/setup-rust#L4)):*

I think we should mention this script/the hook in some documentation. Currently it's not very discoverable. Probably `BuildInstructions.md` in the root (but that is desktop specific)?
</blockquote></details>

Done.
___
*[`scripts/setup-rust` line 66 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJNuJhdEOfDywt_YIYy:-OJOAEbj2PLQIvF6agaU:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/550c14d51e9127056bf078b5c9bfcfb6eef4a84e/scripts/setup-rust#L66)):*

Very nitpicky, and we are a bit inconsistent in the repo overall, but these should ideally be lowercase.
</blockquote></details>

Done.
___
*[`scripts/setup-rust` line 75 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7636#-OJNv3IkCaL8zrgFe5at:-OJOAfjd3fooUwiRuf-R:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/550c14d51e9127056bf078b5c9bfcfb6eef4a84e/scripts/setup-rust#L75)):*

You have already assigned the arguments to local variables, use them by that name instead of argument number.
</blockquote></details>

Done.

PR #7668, opened 2025-02-13
Add an end to end test for UDP-over-TCP on port 80

buggmagnet:
This PR adds an end to end test that connects to a relay using UDP-over-TCP obfuscation on port80.
The test also verifies using packet capture that the traffic sent to the relay during its course is happening exclusively on port 80 using a TCP connection.

The PR also fixes an innocent mistake in `DetailsView.swift` and makes the chevron icon accessibility tap action actually expand the connection details.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7668)

### DIFF
--- a/ios/MullvadVPN/View controllers/Tunnel/ConnectionView/DetailsView.swift
+++ b/ios/MullvadVPN/View controllers/Tunnel/ConnectionView/DetailsView.swift
@@ -42,7 +42,7 @@ extension ConnectionView {
                             connectionDetailRow(
                                 title: LocalizedStringKey("Out IPv6"),
                                 value: outAddressIpv6,
-                                accessibilityId: .connectionPanelOutAddressRow
+                                accessibilityId: .connectionPanelOutIpv6AddressRow
                             )
                         }
                     }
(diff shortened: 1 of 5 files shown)


buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7668)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @SteffenErn)

___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 199 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7668#-OJN-T5h0KjKvxRJu_nl:-OJN7m2p0A5Xs8vaN2Rn:b-bxj75l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/72684795064d7fc50dc9c741e383ae59e6be7551/ios/MullvadVPNUITests/RelayTests.swift#L199)):*
<details><summary><i>Previously, SteffenErn (Steffen Ernst) wrote…</i></summary><blockquote>

Is it enough to check that one stream on port 80 exists or does it make sense to verify that no other stream on a different port exists?
</blockquote></details>

The way the packet capture API was designed is not immediately obvious. 
There will be 1 `Stream` object exactly, that we capture here by filtering by `connectedToIPAddress` 

As the router that does the capture is used by multiple teams, we only want to bother about the relay our device has connected to.
In a nutshell, this is only stream we care about here.

PR #7682, opened 2025-02-17
Fix capacity value

pronebird:
Fix Vec capacity which is set to 4 but only inserts 2 elements.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7682)


dlon (approved):
LGTM :)

PR #7686, opened 2025-02-18
Update `print-package-versions` script for Linux repositories

faern:
The original issue I fixed here is that the script was not compatible with Fedora 41, which my `fedora:latest` image pointed to. It needed the newer argument format to add repositories.

I then realized the script became very noisy. Seems like Fedora 41 print more stuff to stdout/stderr. So I had to update parts of the silencing code.

I then wanted to prepare for the `mullvad-browser-alpha` package soon being added. So I list and show all packages having `mullvad-` in its name instead of just the vpn and browser.

Lastly, I pinned the container images to fixed releases, to avoid things breaking in the future because `latest` is a moving target.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7686)

### DIFF
--- a/desktop/scripts/release/print-package-versions
+++ b/desktop/scripts/release/print-package-versions
@@ -88,6 +88,6 @@ if [[ $rpm == "true" ]]; then
 
     bash -c "podman run --rm -it fedora:latest sh -c \
         \"dnf install -y 'dnf-command(config-manager)' $silent_stdout; \
-        dnf config-manager --add-repo $repository_server_public_url/rpm/$release_channel/mullvad.repo $silent_stdout; \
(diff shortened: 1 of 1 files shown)


faern:
I found an issue when trying the `3-verify-build` script that calls `print-package-versions`. I fixed it and it should now work.

MarkusPettersson98 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7686#-:-OJqTnA6552JpLWMUs55:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7687, opened 2025-02-18
Add `mullvad-browser-alpha` to our linux repositories

faern:
There are alpha builds of the Mullvad Browser now, and we want them in our Linux repositories. It is only needed in the production environment, and it should be available on both the stable and beta repositories.

This PR is related to #7686, which would be good to have merged also, because it makes it easier to print and debug the repository content.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7687)

### DIFF
--- a/ci/mullvad-browser/download-mullvad-browser.sh
+++ b/ci/mullvad-browser/download-mullvad-browser.sh
@@ -2,9 +2,7 @@
 
 set -eu
 
-# TODO: Uncomment when alpha is to be released
-# BROWSER_RELEASES=("alpha" "stable")
-BROWSER_RELEASES=("stable")
+BROWSER_RELEASES=("stable" "alpha")
 REPOSITORIES=("stable" "beta")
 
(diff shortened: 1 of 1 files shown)


ci/mullvad-browser/download-mullvad-browser.sh
```diff
@@ -4,7 +4,7 @@ set -eu
 
 # TODO: Uncomment when alpha is to be released
```
ruihildt, line 5: As the MB alpha can be added to the repositories at the earliest opportunity, we can remove this line.

faern: Good catch! Clumsy of me to not notice. Fixed

raksooo (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7687)* status: all files reviewed, 1 unresolved discussion (waiting on @ruihildt)

PR #7734, opened 2025-02-28
Make button composable

olmoh:
- Makes button component in component library composable
- Updates existing usages of Button component

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7734)

### DIFF
--- a/desktop/packages/mullvad-vpn/src/renderer/lib/components/button/components/index.ts
+++ b/desktop/packages/mullvad-vpn/src/renderer/lib/components/button/components/index.ts
@@ -1,0 +1,2 @@
+export * from './ButtonIcon';
+export * from './ButtonText';
--- a/desktop/packages/mullvad-vpn/src/renderer/components/InfoButton.tsx
+++ b/desktop/packages/mullvad-vpn/src/renderer/components/InfoButton.tsx
@@ -27,7 +27,7 @@ export default function InfoButton({ title, message, children, ...props }: InfoB
         type={ModalAlertType.info}
         buttons={[
           <Button key="back" onClick={hide}>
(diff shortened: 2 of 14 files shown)


desktop/packages/mullvad-vpn/src/renderer/lib/components/button/components/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from '../ButtonContext';
```
tobias-jarvelov, line 1: **question:** Do we want to re-export `ButtonContext` here? As it is already exported in `components/button/index.ts` I don't see a reason to export it here as well.

**suggestion:** Remove the export of `ButtonContext` from this index file.

tobias-jarvelov (changes requested):

tobias-jarvelov (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7734)* status: all files reviewed, 1 unresolved discussion (waiting on @olmoh)

desktop/packages/mullvad-vpn/src/renderer/lib/components/button/components/index.ts
olmoh: Good catch, I previously had the context in the `components` folder but forgot to remove the export. No need to export it twice so will remove it from the `components` index file.

tobias-jarvelov (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7734)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7737, opened 2025-03-01
Offer to store Account Number in password manager

Rawa:
When users create new account prompt their password manager, if available, to store the password.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7737)


Rawa (commented):
1. Yes, this is intentional, since the account number in closer to a password than a username and a bit of a secret that you don't share. This also means that password managers won't show it w/o user explicitly revealing it like on our account page.
2. That would part of another feature, with Compose 1.8.0 we can offer autofill from keyboard, then if we want to offer to store on login that would be yet another feature.
3. Yes, there is nothing linking the password entry to the website at this moment, I'm a bit uncertain about the integration to make it work with the website (if it is supported at all). But there is something called digital asset links (https://developer.android.com/identity/sign-in/credential-manager#add-support-dal) that allows us to add a link to our webiste as well, if this would be forwarded to the website with the store password request I'm not sure of though.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7737#-:-OKQC3yC-c-aF0SUVGEl:b-b22yx0)* status: 8 of 10 files reviewed, all discussions resolved

Pururun (approved):
1. :+1: 
2. Right, let's wait until then again.
3. I think it is fine with this feature then, just maybe create a new issue if we want to explore that in the future.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7737#-:-OKQvC7t-a1ga2I49Dvs:bq02kni)* status: 3 of 10 files reviewed, all discussions resolved

Pururun (dismissed):
> Reviewed 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7737)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Pururun (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7737)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7773, opened 2025-03-07
Zeroize ed25519 secret when dropped

dlon:
Fix DES-1808

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7773)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1037,6 +1037,7 @@ checksum = "4a3daa8e81a3963a60642bcc1f90a670680bd4a77535faa384e9d1c79d620871"
 dependencies = [
  "curve25519-dalek",
  "ed25519",
+ "rand_core 0.6.4",
  "serde",
  "sha2",
  "subtle",
@@ -2909,6 +2910,7 @@ dependencies = [
  "thiserror 2.0.9",
  "tokio",
  "vec1",
+ "zeroize",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


MarkusPettersson98 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7773)* status: all files reviewed, 1 unresolved discussion

___
*[`mullvad-update/src/format/key.rs` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7773#-OKkbyS96wVVUFGlRKdh:-OKkbyS96wVVUFGlRKdi:bjeewxf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e927235a231589d15057edec0dbb8194da8ec164/mullvad-update/src/format/key.rs#L12)):*
> ```XML
> /// ed25519 secret/signing key
> #[derive(Debug, Clone, PartialEq, zeroize::ZeroizeOnDrop)]
> pub struct SecretKey(ed25519_dalek::SigningKey);
> ```

If we care about zeroizing the signing key, I challenge that we derive `Debug` on it (Also `Clone`, but I understand that `clap` needs it in this case). But `Debug` we could do without imho :blush:



_Code quote:_
```XML
#[derive(Debug, Clone, PartialEq, zeroize::ZeroizeOnDrop)]
pub struct SecretKey(ed25519_dalek::SigningKey);
```

PR #7774, opened 2025-03-07
Use the correct connect function to detect online status

Pururun:
We are using `connect(address, port)` but we should use `connect(address)` this is because the `connect(address, port)` does not throw SocketException if the public address is not reachable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7774)


Rawa (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7774)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7835, opened 2025-03-18
Forbid `implicit_clone` clippy lint

faern:
I stumbled on this lint from the pedantic series of clippy lints. I found it to be quite valid thing to lint about. Sometimes it can just be seen as a stylistic choice between `to_owned` or `clone` maybe. But what the lint wants to communicate is that `to_owned` should be used when you have a `&str` and want a `String`. While if you have a `&String` and need a `String` you should clone it. Being more explicit with the intentions.

I don't think it hurts to enable. Will make the code slightly more consistent. I saw no "false positives" where this just caused annoying noise. But tell me what you think.

There are also two "trophy cases" here. You can see in one commit I'm able to completely get rid of cloning in two places. Not that it's going to affect our performance in any way at all, but nice with less cloning if it's completely unnecessary.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7835)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,7 @@ single_use_lifetimes = "warn"
 [workspace.lints.clippy]
 unused_async = "deny"
 undocumented_unsafe_blocks = "warn"
+implicit_clone = "warn"
 
 [workspace.dependencies]
 hickory-proto = "0.24.3"
--- a/mullvad-cli/src/cmds/proxies.rs
+++ b/mullvad-cli/src/cmds/proxies.rs
@@ -168,8 +168,8 @@ impl ProxyEditParams {
     pub fn merge_shadowsocks(self, shadowsocks: &Shadowsocks) -> Shadowsocks {
         let ip = self.ip.unwrap_or(shadowsocks.endpoint.ip());
         let port = self.port.unwrap_or(shadowsocks.endpoint.port());
-        let password = self.password.unwrap_or(shadowsocks.password.to_owned());
-        let cipher = self.cipher.unwrap_or(shadowsocks.cipher.to_owned());
+        let password = self.password.unwrap_or(shadowsocks.password.clone());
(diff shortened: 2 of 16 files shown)


hulthe (dismissed):

dlon (approved):
:lgtm:

> Reviewed 14 of 14 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7835#-:-OLdFJQ_-v9yY0g7Mdm1:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

hulthe (approved):
> Reviewed 14 of 14 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7835)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Cargo.toml
```diff
@@ -78,6 +78,7 @@ single_use_lifetimes = "warn"
 [workspace.lints.clippy]
 unused_async = "deny"
 undocumented_unsafe_blocks = "warn"
+implicit_clone = "warn"
```
MrChocolatine, line 81: The lint level is set to `warn`, should the PR reflect this more accurately instead of saying "it forbids"?

faern: This only means warn while doing development. Our CI forbids all warnings. So this automatically becomes an error in CI and a warning in local development.

MrChocolatine: 👍

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7835)* status: all files reviewed, 1 unresolved discussion (waiting on @MrChocolatine)

PR #7842, opened 2025-03-19
Use new Button component instead of AppButton

olmoh:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7842)


desktop/packages/mullvad-vpn/src/renderer/lib/hooks/use-exclusive-task.tsx
```diff
@@ -0,0 +1,22 @@
...
+export const useExclusiveTask = (task: () => Promise<void>) => {
+  const isRunningRef = useRef(false);
+  const [isRunning, setIsRunning] = useState(false);
+
+  const runTask = useCallback(async () => {
+    if (isRunningRef.current) {
+      return;
+    }
+    isRunningRef.current = true;
+    setIsRunning(true);
```
tobias-jarvelov, line 12: **question:** Isn't the ref alone enough to track if the task is running? It feels like we make use of the `useState` hook to trigger a re-render, but it doesn't feel like it is necessary to do that her?

desktop/packages/mullvad-vpn/src/renderer/lib/hooks/use-exclusive-task.tsx
```diff
@@ -0,0 +1,22 @@
...
+    try {
+      await task();
+    } finally {
+      isRunningRef.current = false;
+      setIsRunning(false);
+    }
+  }, [task]);
+
+  return [runTask, isRunning] as const;
```
tobias-jarvelov, line 21: **suggestion:** Might be a good idea to wrap this in `useMemo` to ensure we give back a stable reference.

tobias-jarvelov (commented):
Looks okay when reading the code. However I'm gonna test it out locally as well. Don't have time to do it now though so I'll just leave my comments meanwhile.

desktop/packages/mullvad-vpn/src/renderer/lib/hooks/use-exclusive-task.tsx
olmoh: Great suggestion, implemented ✅

tobias-jarvelov (changes requested):
Have tested locally and it looks good. One thing we could fix is the git history, more specifically this commit includes a change to `ButtonGroup`: https://github.com/mullvad/mullvadvpn-app/pull/7842/commits/09bd8b5a740070feeaf595b8b094528d0d3d8cd8

That change should be inlcuded in the commit where we add the `ButtonGroup` component: https://github.com/mullvad/mullvadvpn-app/pull/7842/commits/63529b1c456b0ded51a39278d7a9fb7856bd2db9

tobias-jarvelov (approved):

PR #7848, opened 2025-03-20
Add Dot component

olmoh:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7848)

### DIFF
--- a/desktop/packages/mullvad-vpn/src/renderer/lib/components/dot/Dot.tsx
+++ b/desktop/packages/mullvad-vpn/src/renderer/lib/components/dot/Dot.tsx
@@ -1,0 +1,32 @@
+import styled from 'styled-components';
+
+import { Colors } from '../../foundations';
+
+export interface DotProps {
+  variant?: 'primary' | 'success' | 'warning' | 'error';
+  size?: 'tiny' | 'small' | 'medium';
+}
+
+const StyledDiv = styled.div<{ $size: string; $color: string }>`
+  width: ${({ $size }) => $size};
+  height: ${({ $size }) => $size};
+  border-radius: 50%;
+  background-color: ${({ $color }) => $color};
+`;
+
+const sizes = {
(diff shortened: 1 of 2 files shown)


desktop/packages/mullvad-vpn/src/renderer/lib/components/dot/Dot.tsx
```diff
@@ -0,0 +1,31 @@
...
+  medium: '12px',
+};
+
+const colors = {
+  success: Colors.green,
+  warning: Colors.yellow,
+  error: Colors.red,
+};
+
+export const Dot = ({ variant = 'success', size = 'medium', ...props }: DotProps) => {
```
tobias-jarvelov, line 29: **question:** Do we really want `variant` to be optional and have `success` as the default value? This might be confusing if we have logic which feed this component props which for some reason omits the `variant` we might accidentally render the `success` variant.

**suggestion:** Make the `variant` required and remove the default value.

tobias-jarvelov (changes requested):

desktop/packages/mullvad-vpn/src/renderer/lib/components/dot/Dot.tsx
olmoh: I see your point, I like having some sort of default value when it makes since I think it provides some dx, in the case that the default value actually is something you want to use _most_ of the time, which `success` was not. I looked through the app and found that we have a white version of the dot that we use in some spots. To me that makes more sense to use as a default value and would make it behave more like we do in other cases, such as `Text` and `Icon` which defaults to white.

tobias-jarvelov: Nice, then we can add the white version and make that default, let's do that!

tobias-jarvelov (approved):

PR #7852, opened 2025-03-20
Implement getrelays using mullvad api

rablador:
**Why this needs to be done**

`getRelays` call should use the mullvad-api client, without replacing existing implementation.

**What needs to be done**

The API proxy should be implemented using mullvad-api, allowing us to move to the new implementation without removing the old code. For fetching the relay list, the swift side needs to obtain an etag and then also pass the etag to rust again, so that the etag can be used when requesting a new relay list.

Duplicate methods in the APIQuerying interface should also be removed.

**Acceptance criteria**

There is an APIProxy implementation that uses mullvad-api.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7852)


rablador (commented):
To test, replace:

`REST.APIProxy(configuration: configuration)`

with:

```
MullvadAPIProxy(
    transportProvider: configuration.apiTransportProvider,
)
```

in `ProxyFactory.createAPIProxy()`.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7852#-:-OLnTEoEDKcnut1DCP5J:b-f4z20i)* status: 0 of 18 files reviewed, all discussions resolved

pinkisemils (changes requested):
> Reviewed 2 of 13 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7852)* status: 2 of 19 files reviewed, 1 unresolved discussion

___
*[`ios/MullvadREST/ApiHandlers/RESTResponseHandler.swift` line 101 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7852#-OM7Iq6r6D7J3nX5RoeR:-OM7Iq6r6D7J3nX5RoeS:b-9q2cnl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b5d43502cc2814d2dffdca0132f5ce265b416b6/ios/MullvadREST/ApiHandlers/RESTResponseHandler.swift#L101)):*
> ```Swift
>         RustResponseHandler { (response: ProxyAPIResponse) in
>             guard let data = response.data else {
>                 return .unhandledResponse(nil)
> ```

What about responses where the response type is `()`?

pinkisemils (commented):
> Reviewed 5 of 16 files at r1, 9 of 13 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7852)* status: 16 of 19 files reviewed, 1 unresolved discussion (waiting on @rablador)

buggmagnet (commented):
> Reviewed 2 of 13 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7852)* status: 16 of 19 files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[`ios/MullvadREST/ApiHandlers/RESTResponseHandler.swift` line 104 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7852#-OMBWGlm3wPBKsy3xLUX:-OMBWGln0k5NuELzCjWI:b29swf3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b5d43502cc2814d2dffdca0132f5ce265b416b6/ios/MullvadREST/ApiHandlers/RESTResponseHandler.swift#L104)):*
> ```Swift
>             }
> 
>             return if let decoded = try? decoder.decode(type, from: data) {
> ```

We discussed this offline, and came to the conclusion that we shouldn't use `try?` but instead handle the decoding error that might arise from the decoded to make sure we wouldn't lose debug information when we fail to decode the reply from the server.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7852)* status: all files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/MullvadREST/ApiHandlers/RESTResponseHandler.swift` line 104 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7852#-OMBWGlm3wPBKsy3xLUX:-OMHp-cP54AcHLM8S2GO:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b5d43502cc2814d2dffdca0132f5ce265b416b6/ios/MullvadREST/ApiHandlers/RESTResponseHandler.swift#L104)):*

We discussed this offline, and came to the conclusion that we shouldn't use `try?` but instead handle the decoding error that might arise from the decoded to make sure we wouldn't lose debug information when we fail to decode the reply from the server.
</blockquote></details>

Done.

PR #7854, opened 2025-03-20
Add Animate component

olmoh:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7854)

### DIFF
--- a/desktop/packages/mullvad-vpn/src/renderer/lib/components/animate/Animate.tsx
+++ b/desktop/packages/mullvad-vpn/src/renderer/lib/components/animate/Animate.tsx
@@ -1,0 +1,14 @@
+import { AnimatePresentVertical, AnimatePresentVerticalProps } from './components';
+
+export type AnimateProps = {
+  type: 'present-vertical';
+} & AnimatePresentVerticalProps;
+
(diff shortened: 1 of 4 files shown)


olmoh:
Makes sense to add some sort of duration parameter to the component, but since we don't have any duration tokens specified at the moment I have left that out for now.

desktop/packages/mullvad-vpn/src/renderer/lib/components/animate/Animate.tsx
```diff
@@ -0,0 +1,14 @@
...
+export type AnimateProps = {
+  type: 'present-vertical';
+} & AnimatePresentVerticalProps;
+
+export const Animate = ({ type, ...props }: AnimateProps) => {
+  switch (type) {
+    case 'present-vertical':
+      return <AnimatePresentVertical {...props} />;
+    default:
+      return null;
```
MrChocolatine, line 12: In case you want to exhaust your `switch` and get a type-error when it is not:

```suggestion
      return type satisfies never;    
```

I know it goes against your `return null` but since it should never happen...

tobias-jarvelov (approved):

tobias-jarvelov (approved):

PR #7867, opened 2025-03-21
Replace gulp usage and remove associated dependencies

tobias-jarvelov:
This PR:

- Replaces the last things which gulp orchestrated for building the Electron app.
- Removes gulp, browserify dependencies
  - As well as dependencies only used by them, of which there were a lot.
  - This reduces our dependency count by a third(!): `429` dependencies removed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7867)


desktop/packages/mullvad-vpn/tasks/build-production.js
```diff
@@ -0,0 +1,9 @@
+const { build } = require('./build');
+const { setNodeEnvironment } = require('./utils');
+
+async function buildProduction() {
+  setNodeEnvironment('production');
+  await build();
+}
+
+buildProduction();
```
MrChocolatine, line 9: What `await`-ing these `build*` functions?

```ts
await buildProduction()
```

or with an IIFE (I don't think you would need this since you use a recent version of Node.js):

```ts
(async function() {
  await buildProduction()
})();
```

tobias-jarvelov: Thanks for the suggestion! Top level await unfortunately isn't available here as this is a CJS file, otherwise I would support adding it. Adding an IIFE is quite verbose and makes it harder to read the code, so I would prefer not to. 

However, and correct me if I'm wrong, I don't think there's any difference in the outcome with awaiting the `build*` functions or not? Seeing as we only invoke one async function at the top level.

MrChocolatine: You are right, and I instinctively thought I was reading ESM code but it's not.
My bad.

tobias-jarvelov: No worries! Thanks for adding an extra pair of eyes to the code.

olmoh (approved):

desktop/packages/mullvad-vpn/tasks/build-standalone.js
```diff
@@ -0,0 +1,21 @@
+const path = require('path');
+const { removeRecursively, runCommand, setNodeEnvironment } = require('./utils');
+
+const ROOT_DIR = path.resolve('..');
```
tobias-jarvelov, line 4: **issue:** Do not use `path.resolve` without a known reference path. 

**suggestion:** We should use `__dirname` here to ensure the `ROOT_DIR` is resolved to one level above the current file, otherwise it might be resolved to be relative to the current working directory.

MarkusPettersson98 (approved):
> Reviewed 16 of 19 files at r1, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7867)* status: all files reviewed, 2 unresolved discussions (waiting on @MrChocolatine)

PR #7869, opened 2025-03-24
Implement sendproblemreport using mullvad api ios 1134

mojganii:
this PR introduces the ability to use SendProblemReport from Rust into Swift.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869)

### DIFF
--- a/ios/MullvadMockData/MullvadREST/APIProxy+Stubs.swift
+++ b/ios/MullvadMockData/MullvadREST/APIProxy+Stubs.swift
@@ -12,6 +12,14 @@ import MullvadTypes
 import WireGuardKitTypes
 
 struct APIProxyStub: APIQuerying {
+    func mullvadSendProblemReport(
+        _ body: REST.ProblemReportRequest,
+        retryStrategy: REST.RetryStrategy,
+        completionHandler: @escaping ProxyCompletionHandler<Void>
+    ) -> Cancellable {
+        AnyCancellable()
+    }
+
     func mullvadApiGetAddressList(
         retryStrategy: REST.RetryStrategy,
         completionHandler: @escaping ProxyCompletionHandler<[AnyIPEndpoint]>
--- a/ios/MullvadREST/APIRequest/APIRequest.swift
+++ b/ios/MullvadREST/APIRequest/APIRequest.swift
@@ -8,11 +8,14 @@
 
 public enum APIRequest: Codable, Sendable {
     case getAddressList(_ retryStrategy: REST.RetryStrategy)
+    case sendProblemReport(_ retryStrategy: REST.RetryStrategy, problemReportRequest: REST.ProblemReportRequest)
 
     var retryStrategy: REST.RetryStrategy {
-        switch self {
+        return switch self {
         case let .getAddressList(strategy):
-            return strategy
+            strategy
+        case let .sendProblemReport(strategy, _):
+            strategy
         }
     }
 }
--- a/ios/MullvadREST/ApiHandlers/MullvadApiRequestFactory.swift
+++ b/ios/MullvadREST/ApiHandlers/MullvadApiRequestFactory.swift
@@ -24,20 +24,28 @@ public struct MullvadApiRequestFactory: Sendable {
 
             let rawPointer = Unmanaged.passRetained(pointerClass).toOpaque()
 
-            return switch request {
+            switch request {
             case let .getAddressList(retryStrategy):
-                MullvadApiCancellable(handle: mullvad_api_get_addresses(
-                    apiContext.context,
-                    rawPointer,
-                    retryStrategy.toRustStrategy()
-                ))
+                return MullvadApiCancellable(
+                    handle: mullvad_api_get_addresses(
+                        apiContext.context,
+                        rawPointer,
+                        retryStrategy.toRustStrategy()
+                    )
+                )
             case let .sendProblemReport(retryStrategy, problemReportRequest):
-                MullvadApiCancellable(handle: mullvad_api_send_problem_report(
-                    apiContext.context,
-                    rawPointer,
-                    retryStrategy.toRustStrategy(),
-                    problemReportRequest.toRust()
-                ))
+                let rustRequest = RustProblemReportRequest(from: problemReportRequest)
+                return MullvadApiCancellable(
+                    handle: mullvad_api_send_problem_report(
+                        apiContext.context,
+                        rawPointer,
+                        retryStrategy.toRustStrategy(),
+                        rustRequest.toRust()
+                    ),
+                    deinitializer: {
+                        rustRequest.release()
+                    }
+                )
             }
         }
     }
@@ -45,25 +53,4 @@ public struct MullvadApiRequestFactory: Sendable {
 
 extension REST {
     public typealias MullvadApiRequestHandler = (((MullvadApiResponse) throws -> Void)?) -> MullvadApiCancellable
-}
-
-private extension REST.ProblemReportRequest {
-    func toRust() -> UnsafePointer<SwiftProblemReportRequest> {
-        let structPointer = UnsafeMutablePointer<SwiftProblemReportRequest>.allocate(capacity: 1)
-
-        let addressPointer = address.toUnsafePointer()
-        let messagePointer = message.toUnsafePointer()
-        let logPointer = log.toUnsafePointer()
-
-        structPointer.initialize(to: SwiftProblemReportRequest(
-            address: addressPointer,
-            address_len: UInt(address.utf8.count),
-            message: messagePointer,
-            message_len: UInt(message.utf8.count),
-            log: logPointer,
-            log_len: UInt(log.utf8.count)
-        ))
-
-        return UnsafePointer(structPointer)
-    }
 }
--- a/ios/MullvadREST/ApiHandlers/MullvadApiRequestFactory.swift
+++ b/ios/MullvadREST/ApiHandlers/MullvadApiRequestFactory.swift
@@ -31,6 +31,13 @@ public struct MullvadApiRequestFactory: Sendable {
                     rawPointer,
                     retryStrategy.toRustStrategy()
                 ))
+            case let .sendProblemReport(retryStrategy, problemReportRequest):
+                MullvadApiCancellable(handle: mullvad_api_send_problem_report(
+                    apiContext.context,
+                    rawPointer,
+                    retryStrategy.toRustStrategy(),
+                    problemReportRequest.toRust()
+                ))
             }
         }
     }
@@ -38,4 +45,25 @@ public struct MullvadApiRequestFactory: Sendable {
 
 extension REST {
     public typealias MullvadApiRequestHandler = (((MullvadApiResponse) throws -> Void)?) -> MullvadApiCancellable
+}
+
+private extension REST.ProblemReportRequest {
+    func toRust() -> UnsafePointer<SwiftProblemReportRequest> {
+        let structPointer = UnsafeMutablePointer<SwiftProblemReportRequest>.allocate(capacity: 1)
+
+        let addressPointer = address.toUnsafePointer()
+        let messagePointer = message.toUnsafePointer()
+        let logPointer = log.toUnsafePointer()
+
+        structPointer.initialize(to: SwiftProblemReportRequest(
+            address: addressPointer,
+            address_len: UInt(address.utf8.count),
+            message: messagePointer,
+            message_len: UInt(message.utf8.count),
+            log: logPointer,
+            log_len: UInt(log.utf8.count)
+        ))
+
+        return UnsafePointer(structPointer)
+    }
 }
--- a/ios/MullvadREST/ApiHandlers/MullvadApiRequestFactory.swift
+++ b/ios/MullvadREST/ApiHandlers/MullvadApiRequestFactory.swift
@@ -41,10 +41,7 @@ public struct MullvadApiRequestFactory: Sendable {
                         rawPointer,
                         retryStrategy.toRustStrategy(),
                         rustRequest.toRust()
-                    ),
-                    deinitializer: {
-                        rustRequest.release()
-                    }
+                    )
                 )
             }
         }
--- a/ios/MullvadREST/ApiHandlers/RESTAPIProxy.swift
+++ b/ios/MullvadREST/ApiHandlers/RESTAPIProxy.swift
@@ -18,6 +18,12 @@ public protocol APIQuerying: Sendable {
         completionHandler: @escaping @Sendable ProxyCompletionHandler<[AnyIPEndpoint]>
     ) -> Cancellable
 
+    func mullvadSendProblemReport(
+        _ body: REST.ProblemReportRequest,
+        retryStrategy: REST.RetryStrategy,
+        completionHandler: @escaping @Sendable ProxyCompletionHandler<Void>
+    ) -> Cancellable
+
     func getAddressList(
         retryStrategy: REST.RetryStrategy,
         completionHandler: @escaping @Sendable ProxyCompletionHandler<[AnyIPEndpoint]>
@@ -75,6 +81,28 @@ extension REST {
                 name: "get-api-addrs",
                 dispatchQueue: dispatchQueue,
                 request: .getAddressList(retryStrategy),
+                transportProvider: configuration.apiTransportProvider,
+                responseDecoder: responseDecoder,
+                responseHandler: responseHandler,
+                completionHandler: completionHandler
+            )
+
+            operationQueue.addOperation(networkOperation)
+
+            return networkOperation
+        }
+
+        public func mullvadSendProblemReport(
+            _ body: REST.ProblemReportRequest,
+            retryStrategy: REST.RetryStrategy,
+            completionHandler: @escaping ProxyCompletionHandler<Void>
+        ) -> any Cancellable {
+            let responseHandler = rustEmptyResponseHandler()
+
+            let networkOperation = MullvadApiNetworkOperation(
+                name: "send_problem_report",
+                dispatchQueue: dispatchQueue,
+                request: .sendProblemReport(retryStrategy, problemReportRequest: body),
                 transportProvider: configuration.apiTransportProvider,
                 responseDecoder: responseDecoder,
                 responseHandler: responseHandler,
@@ -358,7 +386,7 @@ extension REST {
         let newExpiry: Date
     }
 
-    public struct ProblemReportRequest: Encodable, Sendable {
+    public struct ProblemReportRequest: Codable, Sendable {
         public let address: String
         public let message: String
         public let log: String
--- a/ios/MullvadREST/Extensions/String+UnsafePointer.swift
+++ b/ios/MullvadREST/Extensions/String+UnsafePointer.swift
@@ -7,10 +7,14 @@
 //
 import Foundation
 extension String {
-    func toUnsafePointer() -> UnsafePointer<UInt8>? {
+    /// Converts the `String` to an `UnsafePointer<UInt8>`. The memory allocated for the pointer
+    /// is not null-terminated, so the caller must ensure that they manage the memory properly.
+    func toUnsafePointer() -> (pointer: UnsafePointer<UInt8>?, size: UInt)? {
         guard let data = self.data(using: .utf8) else { return nil }
+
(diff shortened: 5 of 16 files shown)


pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869)* status: 0 of 14 files reviewed, 1 unresolved discussion

___
*[`mullvad-api/src/lib.rs` line 687 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM74x2w8Y_r4yPUVZIT:-OM74x2w8Y_r4yPUVZIU:b8t8816) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-api/src/lib.rs#L687)):*
> ```XML
>         log: &str,
>         metadata: &BTreeMap<String, String>,
>     ) -> impl Future<Output = Result<rest::Response<Incoming>, rest::Error>> {
> ```

We might not *have* to do this duplication here, since on the iOS side, we are not expecting a response anyway :shrug: ?

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869)* status: 0 of 14 files reviewed, 9 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM78KGk3aVDzdjg12cJ:-OM78KGk3aVDzdjg12cK:b-jqx9xe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L14)):*
> ```Swift
>         let buffer = UnsafeMutablePointer<UInt8>.allocate(capacity: data.count)
>         data.copyBytes(to: buffer, count: data.count)
>         return UnsafePointer(buffer)
> ```

Doesn't this leak unless it's cleaned up properly later? 
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 96 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM76BDR33bvn4to9hay:-OM76BDR33bvn4to9haz:bbksujf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L96)):*
> ```XML
> 
>     let response = retry_future(future_factory, should_retry, retry_strategy.delays()).await?;
>     SwiftMullvadApiResponse::with_body(response).await
> ```

Instead of a `with_body` that expects a response, we could also have a default `SwiftMullvadApiResponse::ok()` in cases where the response body is empty and we don't care about the response status code.
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 160 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM76OVS049ziuilvSsJ:-OM76OVS049ziuilvSsK:b-w06zs7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L160)):*
> ```XML
> 
> #[repr(C)]
> pub struct SwiftMap {
> ```

I think we could call this type something more meaningful - like `ProblemReportMetadata`. But if you feel strongly about leaving it generic, it should be moved to some other module in `mullvad-ios`. 
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 172 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM77RxF-5ZywnzF7nmo:-OM77RxF-5ZywnzF7nmp:b-r8wpef) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L172)):*
> ```XML
> 
>     unsafe fn add(&mut self, key: *const c_char, value: *const c_char) {
>         let key = CStr::from_ptr(key).to_string_lossy().into_owned();
> ```

Instead of doing lossy conversions, this code could return a boolean or an error when the conversion fails so we can log about it.
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 188 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM77mNt81qQ8zt1vBxQ:-OM77mNt81qQ8zt1vBxR:byzk561) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L188)):*
> ```XML
>         inner: Box::into_raw(map),
>     });
>     Box::into_raw(swift_map)
> ```

The double boxing isn't necessary. You can just return a `-> SwiftMap` without having another allocation here. The swift side can handle it being just a struct. This would simplify the dropping too. 
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 202 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM77ap16P9gaRZsfk8a:-OM77ap16P9gaRZsfk8b:b-fksbcr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L202)):*
> ```XML
> #[no_mangle]
> pub extern "C" fn swift_map_get(map: *mut SwiftMap, key: *const c_char) -> *const c_char {
>     if let Some(map) = unsafe { map.as_mut() } {
> ```

Do we need this on the Swift side?
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 214 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM78---7MMRhaAEnA42:-OM78---7MMRhaAEnA43:be37po7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L214)):*
> ```XML
> 
> #[no_mangle]
> pub extern "C" fn swift_map_free(map: *mut SwiftMap) {
> ```

Whilst there should be a function that frees the underlying map, 
___
*[`ios/MullvadREST/Rust/RustProblemReportRequest.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM787BE84wI3_QjQ1uI:-OM787BE84wI3_QjQ1uJ:bgli48u) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/ios/MullvadREST/Rust/RustProblemReportRequest.swift#L14)):*
> ```Swift
> 
>     public init(from request: REST.ProblemReportRequest) {
>         rustPointer = UnsafeMutablePointer<SwiftProblemReportRequest>.allocate(capacity: 1)
> ```

You can initialize the problem report request on the stack - no reason for it to be heap allocated here.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869)* status: all files reviewed, 11 unresolved discussions (waiting on @mojganii and @pinkisemils)

___
*[`mullvad-api/src/lib.rs` line 687 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM74x2w8Y_r4yPUVZIT:-OM7GrSb3MJwtsJqtn38:b44tpx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-api/src/lib.rs#L687)):*

We might not *have* to do this duplication here, since on the iOS side, we are not expecting a response anyway :shrug: ?
</blockquote></details>

👍
___
*[`mullvad-api/src/lib.rs` line 681 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM7GcFG5BSBwUbB1QHI:-OM7GcFG5BSBwUbB1QHJ:bu6igfk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-api/src/lib.rs#L681)):*
> ```XML
>     }
> 
>     pub fn porblem_report_response(
> ```

Typo: "porblem"
___
*[`ios/MullvadREST/ApiHandlers/RESTAPIProxy.swift` line 103 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM7EeGp8x3dv9zb5N9z:-OM7EeGp8x3dv9zb5NA-:b-1yfju2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/ios/MullvadREST/ApiHandlers/RESTAPIProxy.swift#L103)):*
> ```Swift
> 
>             let networkOperation = MullvadApiNetworkOperation(
>                 name: "send_problem_report",
> ```

Should be "-" rather than "_".

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869)* status: all files reviewed, 11 unresolved discussions (waiting on @pinkisemils and @rablador)

___
*[`ios/MullvadREST/ApiHandlers/RESTAPIProxy.swift` line 103 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM7EeGp8x3dv9zb5N9z:-OM7Xjrf8tHG11AEXZOg:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/ios/MullvadREST/ApiHandlers/RESTAPIProxy.swift#L103)):*

Should be "-" rather than "_".
</blockquote></details>

Done
___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM78KGk3aVDzdjg12cJ:-OM7YUNp3Lk7NcM_yeTQ:b-143wxe) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L14)):*

Doesn't this leak unless it's cleaned up properly later?
</blockquote></details>

if we use this function without releasing memory, It can lead to memory leakage. shouldn't the owner take care of it?
___
*[`ios/MullvadREST/Rust/RustProblemReportRequest.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM787BE84wI3_QjQ1uI:-OM7epKHBd9pU6VUFfK4:bzeh452) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/ios/MullvadREST/Rust/RustProblemReportRequest.swift#L14)):*

You can initialize the problem report request on the stack - no reason for it to be heap allocated here.
</blockquote></details>

But when passing to Rust, it's better to have access to the original object, which means using a pointer. Am I right, since the object could be large?
___
*[`mullvad-api/src/lib.rs` line 681 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM7GcFG5BSBwUbB1QHI:-OM7f2bfCngAZ6yPRd2z:bcvexe6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-api/src/lib.rs#L681)):*

Typo: "porblem"
</blockquote></details>

Removed based on duplication code
___
*[`mullvad-api/src/lib.rs` line 687 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM74x2w8Y_r4yPUVZIT:-OM7NJH_3R7ne1SRfaq9:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-api/src/lib.rs#L687)):*

We might not *have* to do this duplication here, since on the iOS side, we are not expecting a response anyway :shrug: ?
</blockquote></details>

Done.
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 96 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM76BDR33bvn4to9hay:-OM7NLM3AniDA5N-n0w-:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L96)):*

Instead of a `with_body` that expects a response, we could also have a default `SwiftMullvadApiResponse::ok()` in cases where the response body is empty and we don't care about the response status code.
</blockquote></details>

Done.
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 160 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM76OVS049ziuilvSsJ:-OM7PQa92FJ5-6sg8srz:bakx7de) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L160)):*

I think we could call this type something more meaningful - like `ProblemReportMetadata`. But if you feel strongly about leaving it generic, it should be moved to some other module in `mullvad-ios`.
</blockquote></details>

I renamed that to `ProblemReportMetadata`.
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 172 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM77RxF-5ZywnzF7nmo:-OMBngzZ0ko073hjbWKl:b-7yl44d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L172)):*

Instead of doing lossy conversions, this code could return a boolean or an error when the conversion fails so we can log about it.
</blockquote></details>

It's returning `Result` now.
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 188 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM77mNt81qQ8zt1vBxQ:-OM7lncMCM5546SOzlav:b-nwbvfy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L188)):*

The double boxing isn't necessary. You can just return a `-> SwiftMap` without having another allocation here. The swift side can handle it being just a struct. This would simplify the dropping too.
</blockquote></details>

but we still need to have access to pointer, don't we? like below



_Code snippet:_
```XML
pub extern "C" fn problem_report_metadata_new() -> *mut ProblemReportMetadata {
    let metadata = Box::new(ProblemReportMetadata {
        inner: Map(BTreeMap::new()),
    });

    Box::into_raw(metadata)
}
```
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 202 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM77ap16P9gaRZsfk8a:-OM7X9n_DNBtxEuIx6-t:ba3dfob) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L202)):*

Do we need this on the Swift side?
</blockquote></details>

Removed
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 214 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM78---7MMRhaAEnA42:-OMBo1xW5s9bw7greGci:b-ijcpwn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L214)):*

Whilst there should be a function that frees the underlying map,
</blockquote></details>

Can you elaborate more? is it referring to add comment for this?

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869)* status: all files reviewed, 7 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadREST/Rust/RustProblemReportRequest.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM787BE84wI3_QjQ1uI:-OMBrxJU-cpTQWWjYHB_:b-m08vyp) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/ios/MullvadREST/Rust/RustProblemReportRequest.swift#L14)):*

But when passing to Rust, it's better to have access to the original object, which means using a pointer. Am I right, since the object could be large?
</blockquote></details>

The original struct will only hold a pointer within it. So there is no need to have a pointer to a pointer. You can pass it to Rust as a struct, by value, or as a pointer via `&params`, where `params` is `SwiftProblemReportRequest` . 
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 188 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM77mNt81qQ8zt1vBxQ:-OMBsdKC1HpZgDLuW92A:b-6qzvrh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L188)):*

but we still need to have access to pointer, don't we? like below
</blockquote></details>

```
pub extern "C" fn problem_report_metadata_new() -> ProblemReportMetadata {
    let metadata = ProblemReportMetadata {
        inner: Box::into_raw(Box::new(Map(BTreeMap::new()))),
    });

    metadata
}
```
Something like that, the inner pointer remains a boxed pointer, but we end up returning a struct.
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 214 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM78---7MMRhaAEnA42:-OMBtCCaETN4t-3kYhbM:bv0ylaf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/mullvad-ios/src/api_client/send_problem_report.rs#L214)):*

Can you elaborate more? is it referring to add comment for this?
</blockquote></details>

This was an incomplete comment, but I think this code would be simpler if we didn't use double-allocation here.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869)* status: all files reviewed, 7 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7869#-OM78KGk3aVDzdjg12cJ:-OMBuukfA1pMbFbz4eIH:bp5l781) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5cd343f69d8e5aafcc379ffb2f556b7e17cb9a14/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L14)):*

if we use this function without releasing memory, It can lead to memory leakage. shouldn't the owner take care of it?
</blockquote></details>

Nope, I was a bit confused about how this pointer was used. I think we should document then that the caller here is responsible for releasing the memory here.
It is also not entirely obvious if the pointer here is pointing to an nul-terminated string here or not.

mojganii:
it's being tracked below here: https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925

PR #7872, opened 2025-03-24
Use ListItem component in settings view

olmoh:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7872)


desktop/packages/mullvad-vpn/src/renderer/components/Settings.tsx
```diff
@@ -102,137 +100,142 @@ export default function Support() {
...
+    <Icon icon="chevron-right" />
+  </NavigationListItem>
+);
+
+const DebugButton = () => (
+  <NavigationListItem to={RoutePath.debug}>
+    <ListItem.Label>Developer tools</ListItem.Label>
+    <Icon icon="chevron-right" />
+  </NavigationListItem>
+);
```
tobias-jarvelov, line 236: **question:** When components have been refactored they have also been converted from using `function` declaration to an `arrow function` expression. Is there a reason for this?

**note:** Using a function or arrow function is pretty much equivalent, but for components the arrow functions will show up as unnamed functions in React Devtools' Components tab.

**suggestion:** Unless there's a good reason to convert them to arrow functions, keep using the `function` declaration to ensure the component names are visible in the React Devtools' Components tab.

desktop/packages/mullvad-vpn/src/renderer/components/Settings.tsx
tobias-jarvelov, line 1: **note:** _This is a general comment for this entire file._

When we remove usage of `Cell` and `SettingsGroup` we also remove access to the functionality they provide through their respective React Context. I can't see that we actually use any of that functionality at this level, but it's something we should be aware of when replacing `Cell` and  `Settings*` components in the future.

tobias-jarvelov (changes requested):

desktop/packages/mullvad-vpn/src/renderer/components/Settings.tsx
olmoh: Just did this out of habit, but since we see some positives to using a function I will do so moving forward.

desktop/packages/mullvad-vpn/src/renderer/components/Settings.tsx
olmoh: Yes, I don't know of any special behavior in this exact view either, but we do have to keep it mind for other ones.

tobias-jarvelov (approved):
Much improved! Nicely done!

PR #7876, opened 2025-03-24
Fix syntax error in Apparmor profile

inventor500:
The executable path must be quoted instead of using an escape character.
See man apparmor.d(5), under "Unquoted Profile Name" - "Rules with embedded spaces or tabs must be quoted."

Before this commit, running `aa-enforce /opt/Mullvad\ VPN/mullvad-gui` (as root) returned an error:
```
ERROR: Syntax Error: Unknown line found in file /etc/apparmor.d/mullvad line 5:
    profile mullvad /opt/Mullvad\ VPN/mullvad-gui { userns,
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7876)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,9 @@ Line wrap the file at 100 chars.                                              Th
 ## [2025.5] - 2025-03-26
 This release is identical to 2025.5-beta1
 
+#### Linux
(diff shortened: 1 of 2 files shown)


dlon (approved):
LGTM, thanks!

PR #7881, opened 2025-03-25
Implement changes to suggestedUpgrade in Electron

tobias-jarvelov:
When the `AppVersionInfo.suggestedUpgrade` property is updated, due to the changes in [DES-1811](https://linear.app/mullvad/issue/DES-1811/create-specification-for-communication-between-electron-and-the-daemon) we need to make some changes in the code to handle the new format.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7881)

### DIFF
--- a/desktop/packages/mullvad-vpn/src/shared/notifications/update-available.ts
+++ b/desktop/packages/mullvad-vpn/src/shared/notifications/update-available.ts
@@ -1,6 +1,7 @@
 import { sprintf } from 'sprintf-js';
 
 import { messages } from '../../shared/gettext';
+import { AppVersionInfoSuggestedUpgrade } from '../daemon-rpc-types';
 import { getDownloadUrl } from '../version';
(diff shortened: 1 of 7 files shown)


desktop/packages/mullvad-vpn/src/shared/notifications/update-available.ts
```diff
@@ -21,8 +22,8 @@ export class UpdateAvailableNotificationProvider
 {
   public constructor(private context: UpdateAvailableNotificationContext) {}
 
-  public mayDisplay() {
-    return this.context.suggestedUpgrade ? true : false;
+  public mayDisplay(): boolean {
```
olmoh, line 25: Does not seem like this is needed?

olmoh (approved):

olmoh (approved):

PR #7885, opened 2025-03-25
Add daemon gRPC integration for in-app upgrade feature

tobias-jarvelov:
Integrates the AppUpgrade gRPC messages with the Daemon to make them available to the Electron app via IPC and Redux.


desktop/packages/mullvad-vpn/src/renderer/redux/app-upgrade/reducers.ts
```diff
@@ -0,0 +1,37 @@
...
+    case 'APP_UPGRADE_SET_ERROR':
+      return {
+        ...state,
+        error: action.error,
+      };
+    case 'APP_UPGRADE_RESET':
+      return {
+        ...state,
+        ...initialState,
+      };
```
tobias-jarvelov, line 33: **issue:** We don't need to spread the existing `state` if we are resetting the entire state.

**suggestion:** Remove `...state` and only spread the `initalState`.

desktop/packages/mullvad-vpn/src/shared/daemon-rpc-types.ts
tobias-jarvelov, line 1: **thought:** The types in this file might have a better home in `src/shared/ipc-types.ts`. Discuss this with Oliver.

desktop/packages/mullvad-vpn/src/shared/daemon-rpc-types.ts
```diff
@@ -173,6 +173,48 @@ export type DaemonEvent =
   | { deviceRemoval: Array<IDevice> }
   | { accessMethodSetting: AccessMethodSetting };
 
+export enum AppUpgradeError {
+  generalError,
+  downloadFailed,
+  verificationFailed,
+}
```
tobias-jarvelov, line 180: **issue:** We need some error for when the Electron app fails to start the installer. It would be reasonable to add an `startInstallerFailed` error to the `AppUpgradeError` enum. But we wouldn't want to do that if these types are in the `daemon-rpc-types.ts` file and thus coupled to the Daemon/gRPC.

**suggestion:** Discuss with Oliver to see if these types can be moved somewhere where we can add more values to the enum without associating the new values with the Daemon/gRPC.

desktop/packages/mullvad-vpn/src/main/grpc-type-convertions.ts
```diff
@@ -700,6 +702,68 @@ function convertFromObfuscationSettings(
...
+  if (downloadStartingData !== undefined) {
+    return { type: 'APP_UPGRADE_EVENT_DOWNLOAD_STARTED' };
+  }
+
+  const downloadProgressData = data.getDownloadProgress();
+  if (downloadProgressData !== undefined) {
+    const [server, progress, timeLeftDuration] = [
+      downloadProgressData.getServer(),
+      downloadProgressData.getProgress(),
+      downloadProgressData.getTimeLeft()!,
```
olmoh, line 727: Does not feel like we should need to use `!` here. Can this be removed by either:

- Update the type, if `timeLeft` never should be `undefined`.
- Handle `timeLeft` being `undefined`, if it can be.

If none of these are applicable I think we should add a comment that explains why we need to do this.

desktop/packages/mullvad-vpn/src/shared/daemon-rpc-types.ts
olmoh: Could we prefix the new types in this file with `Daemon`, so `AppUpgradeEvent` => `DaemonAppUpgradeEvent` and so on. This would match the naming of the other daemon events (`DaemonEvent`) and make it clear that this is related to the daemon.
We can then add another type in the app `AppUpgradeEvent`, which extends `DaemonAppUpgradeEvent` with the fail to start installer error.

olmoh (changes requested):

desktop/packages/mullvad-vpn/src/shared/daemon-rpc-types.ts
tobias-jarvelov: Brilliant, of course we should do that!

desktop/packages/mullvad-vpn/src/main/grpc-type-convertions.ts
tobias-jarvelov: Nice find, it isn't necessary and is simply a rest from earlier testing and debugging. Will remove it.

desktop/packages/mullvad-vpn/src/shared/daemon-rpc-types.ts
tobias-jarvelov: **note:** I moved the `enum` `AppUpgradeError` into a `constants` file in the `src/shared` folder so it would exist in a space where no one "owns it" but itself. We can share it between daemon/gRPC integration code, Electron IPC and the rest of the React app.

desktop/packages/mullvad-vpn/src/shared/app-upgrade.ts
```diff
@@ -0,0 +1,3 @@
+import { DaemonAppUpgradeEventStatus } from './daemon-rpc-types';
+
+export type AppUpgradeEvent = DaemonAppUpgradeEventStatus;
```
tobias-jarvelov, line 3: **note:** I moved the `AppUpgradeEvent` into a separate file here in order for it to be reusable by both the Electron IPC and Redux, as it didn't make sense that either IPC or Redux should own this type.

desktop/packages/mullvad-vpn/src/main/app-upgrade.ts
```diff
@@ -0,0 +1,37 @@
+import { DaemonAppUpgradeEvent } from '../shared/daemon-rpc-types';
+import log from '../shared/logging';
+import { DaemonRpc, SubscriptionListener } from './daemon-rpc';
+import { IpcMainEventChannel } from './ipc-event-channel';
+
+export default class AppUpgrade {
```
tobias-jarvelov, line 6: **note:** This class currently only handles the IPC <-> daemon communication for `AppUpgrade` messages. However, it will also hold the logic to start the installer, but that code will be added in another PR.

desktop/packages/mullvad-vpn/src/shared/ipc-schema.ts
```diff
@@ -33,6 +33,8 @@ interface ILogEntry {
   message: string;
 }
 import { MapData } from '../renderer/lib/3dmap';
+import { AppUpgradeEvent } from './app-upgrade';
```
olmoh, line 36: I think the location the current location of the app upgrade types and constants feel a bit confusing. To me it would make sense if these types were located in the same directory. `shared/types/app-upgrade.ts` maybe?

olmoh (approved):
Nice job, lgtm!

PR #7909, opened 2025-03-27
Add ability to start the upgrade installer from electron

tobias-jarvelov:
Adds support for starting the verified installer. The installer can be started either manually or automatically.

- Manually from the Renderer via IPC: `IpcRendererEventChannel.app.upgradeInstallerStart`.
- Automatically if the window is focused and the verified installer becomes available after the user has initiated an app upgrade.


olmoh (approved):
LGTM

tobias-jarvelov:
Fixes: DES-1867

PR #7917, opened 2025-03-28
Add app upgrade list item for linux in app info view

olmoh:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7917)


desktop/packages/mullvad-vpn/src/renderer/components/views/app-info/components/update-available-list-item/UpdateAvailableListItem.tsx
```diff
@@ -4,23 +4,29 @@ import { messages } from '../../../../../../shared/gettext';
...
+type UpdateAvailableListItemProps = Omit<ListItemProps, 'children'>;
+
+export function UpdateAvailableListItem(props: UpdateAvailableListItemProps) {
   const { suggestedUpgrade } = useVersionSuggestedUpgrade();
   const { isBlocked } = useConnectionIsBlocked();
 
+  const openDownloadUrl = useOpenDownloadUrl();
   const pushAppUpgrade = usePushAppUpgrade();
+  const isLinux = useIsLinux();
+  const onClick = isLinux ? openDownloadUrl : pushAppUpgrade;
```
tobias-jarvelov, line 24: **suggestion:** Move `onClick` to a separate hook if we have to conditionally set what action the click handler should perform.

```js
export const useHandleClick = () => {
  const isLinux = useIsLinux();
  const openDownloadUrl = useOpenDownloadUrl();
  const pushAppUpgrade = usePushAppUpgrade();

  const handleClick = useCallback(() => {
    if (isLinux) {
      openDownloadUrl();
    } else {
      pushAppUpgrade();
    }
  }, [openDownloadUrl, pushAppUpgrade];

  return handleClick;
}

...

const handleClick = useHandleClick();

...

<ListItem.Trigger onClick={onClick}>

```

desktop/packages/mullvad-vpn/src/renderer/components/views/app-info/AppInfoView.tsx
```diff
@@ -1,16 +1,23 @@
...
 import { NavigationScrollbars } from '../../NavigationScrollbars';
 import SettingsHeader, { HeaderTitle } from '../../SettingsHeader';
 import { ChangelogListItem, UpdateAvailableListItem, VersionListItem } from './components';
 import { BetaListItem } from './components/beta-list-item';
 import { useShowUpdateAvailable } from './hooks';
 
+const StyledUpdateAvailableListItem = styled(UpdateAvailableListItem)`
+  margin-bottom: 16px;
+`;
+
```
tobias-jarvelov, line 20: **note:** It feels like we had to add a lot of code, seemingly to just support us wrapping `UpdateAvailableListItem` as a styled component with a 16px bottom margin. This change resulted in that `UpdateAvailableListItem` now has to pass props to `ListItem`, which itself had to spread its props onto its `StyledFlex` component, in order for the `className` prop to be set so that the `StyledUpdateAvailableListItem` can add its static margin.

**question:** Is there a reason for doing this instead of adding another `Flex` to the mix?:

```jsx
<Animate
  present={showUpdateAvailable}
  animations={[{ type: 'fade' }, { type: 'wipe', direction: 'vertical' }]}>
    <Flex $margin={{ bottom: 'medium' }}>
      <UpdateAvailableListItem />
    </Flex>
</Animate>
```

olmoh: Yeah that is much cleaner. 🤦 Had to add the flex grow property to ListItem to make it work as expected but I think that should be there anyway. Great feedback!

desktop/packages/mullvad-vpn/src/renderer/components/views/app-info/components/update-available-list-item/hooks/useHandleClick.tsx
```diff
@@ -0,0 +1,9 @@
+import { usePushAppUpgrade } from '../../../../../../history/hooks';
+import { useIsLinux, useOpenDownloadUrl } from './';
```
tobias-jarvelov, line 2: **issue:** When we import from `'.'` we create a circular import, as the index file also exports this file.

**suggestion:** Update the imports to import from each file separately:

```js
import { useIsLinux } from './useIsLinux';
import { useOpenDownloadUrl } from './useOpenDownloadUrl';
```

tobias-jarvelov (changes requested):
Found one thing. Other than that it looks good. Feel free to merge after fixing.

desktop/packages/mullvad-vpn/src/renderer/components/views/app-info/components/update-available-list-item/hooks/useHandleClick.tsx
olmoh: Nice catch 👍

PR #7925, opened 2025-03-31
Implement sendproblemreport using mullvad api

rablador:
This PR introduces support for using `mullvad-api` when `sendProblemRepost` is revoked. Since mullvad-api is not yet fully implemented, For testing purposes, you can simply replace the relevant snippet in `ProxyFactory` for `createAPIProxy()`.

MullvadAPIProxy(
    transportProvider: configuration.apiTransportProvider,
)

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)


pinkisemils (changes requested):
> Reviewed 10 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 10 of 14 files reviewed, 3 unresolved discussions

___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMfkSBz5lMQbpZf6Qg3:-OMfkSBz5lMQbpZf6Qg4:b-2tz7r5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b9ea13cc3b324525ed142cf96da0b402059ddb9/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L12)):*
> ```Swift
>     /// Converts the `String` to an `UnsafePointer<UInt8>`. The memory allocated for the pointer
>     /// is not null-terminated, so the caller must ensure that they manage the memory properly.
>     func toUnsafePointer() -> (pointer: UnsafePointer<UInt8>?, size: UInt)? {
> ```

I want to raise a discussion in the team as to whether we should use an abstraction like this. I am OK with doing this, 
___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 74 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMfllTqDqtIThMhDvl4:-OMfllTqDqtIThMhDvl5:b-e813nf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b9ea13cc3b324525ed142cf96da0b402059ddb9/mullvad-ios/src/api_client/send_problem_report.rs#L74)):*
> ```XML
> }
> 
> async fn mullvad_api_send_problem_report_inner(
> ```

As Jon's changes are now merged, this can largely be replaced with `do_request`.
___
*[`mullvad-ios/src/api_client/mod.rs` line 14 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMfl2YT17vvVkzCJY0A:-OMfl2YT17vvVkzCJY0B:b-j0y744) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b9ea13cc3b324525ed142cf96da0b402059ddb9/mullvad-ios/src/api_client/mod.rs#L14)):*
> ```XML
> mod response;
> mod retry_strategy;
> mod send_problem_report;
> ```

This could just be called `problem_report`.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 10 of 14 files reviewed, 4 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMfkSBz5lMQbpZf6Qg3:-OMflrRcCRaUIdegF_6t:b-kcdc6a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b9ea13cc3b324525ed142cf96da0b402059ddb9/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L12)):*

I want to raise a discussion in the team as to whether we should use an abstraction like this. I am OK with doing this,
</blockquote></details>

... but I'd also be OK with more ceremony around things like this. We could also choose to end up passing the strings are C strings for the message and the email address, and then we no longer need this particular extension either. But if we want to stuff pointers to Strings in a struct, this is very succinct. 
___
*[`mullvad-ios/src/api_client/response.rs` line 33 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMfn1LC7UsRlOgr1Mtp:-OMfn1LC7UsRlOgr1Mtq:b-e3twa0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b9ea13cc3b324525ed142cf96da0b402059ddb9/mullvad-ios/src/api_client/response.rs#L33)):*
> ```XML
>     }
> 
>     pub async fn ok() -> Result<Self, rest::Error> {
> ```

This function does not have to be asynchronous. The CI is warning you about this too. You can also run clippy locally.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 1 of 17 files reviewed, 5 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 11 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMgBWXe8urC1iP8l7LF:-OMgBWXe8urC1iP8l7LG:b-aur3mk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1fc1f42a9352a7e65b8d5725bfcc5a527a39caf9/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L11)):*
> ```Swift
> extension String {
>     /// Converts the `String` to an `UnsafePointer<UInt8>`. The memory allocated for the pointer
>     /// is not null-terminated, so the caller must ensure that they manage the memory properly.
> ```

The fact that the string data is not null terminated implies that wherever this pointer is passed, it has to be passed along with the size of the string. The fact that the caller must manage the memory here and free this pointer after use is orthogonal to that.

pinkisemils (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 1 of 17 files reviewed, 6 unresolved discussions (waiting on @rablador)

___
*[`ios/MullvadVPNTests/MullvadREST/MullvadApi/RustProblemReportRequestTests.swift` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMgBkNw7GpHtEaGjJre:-OMgBkNw7GpHtEaGjJrf:b-okifkr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1fc1f42a9352a7e65b8d5725bfcc5a527a39caf9/ios/MullvadVPNTests/MullvadREST/MullvadApi/RustProblemReportRequestTests.swift#L34)):*
> ```Swift
>         let rustRequest = RustProblemReportRequest(from: request)
>         let rustStruct = rustRequest.toRust()
>         #expect(rustStruct.meta_data != nil, "Metadata should not be for \(metadata)")
> ```

Should the message not say `"Metadata should not be nil for \(metadata)"` ?

pinkisemils (dismissed):
> Reviewed 2 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 2 of 17 files reviewed, 3 unresolved discussions (waiting on @rablador)

___
*[`mullvad-ios/src/api_client/send_problem_report.rs` line 74 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMfllTqDqtIThMhDvl4:-OMgWk4t1me3H9uvvY6L:b-knupaz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b9ea13cc3b324525ed142cf96da0b402059ddb9/mullvad-ios/src/api_client/send_problem_report.rs#L74)):*

Done
</blockquote></details>

The code shouldn't really be copied, but instead use the function that Jon introduced.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 1 of 17 files reviewed, 4 unresolved discussions (waiting on @mojganii and @pinkisemils)

___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMfkSBz5lMQbpZf6Qg3:-OMpV6tn6Q0iVhAReNlh:b-7yoe8l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b9ea13cc3b324525ed142cf96da0b402059ddb9/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L12)):*

I added another function to it which creates null-terminated pointer for string with `withCString`.  If there is no case that we want to send non null-terminated, then I can remove `toUnsafePointer`
</blockquote></details>

Unless we need to I don't think we should. Less code is good code :)
___
*[`mullvad-ios/src/api_client/mod.rs` line 107 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMpSpUbA5hzzFuH1azD:-OMpSpUbA5hzzFuH1azE:b-6ca5aw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e9a7ecc3fd9d328b3a3f7cbdea9783ce1f556f2/mullvad-ios/src/api_client/mod.rs#L107)):*
> ```XML
> }
> 
> async fn do_request_with_empty_body<F, T>(
> ```

Nit: We could use the existing `do_request`, extract everything but the last part (`SwiftMullvadApiResponse::with\_body(response).await`) and create two new functions: `do_request_with_body` and `do_request_with_empty_body`. Both calling `do_request` and returning the appropriate response.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 1 of 17 files reviewed, 5 unresolved discussions (waiting on @mojganii and @pinkisemils)

___
*[`mullvad-ios/src/api_client/response.rs` line 6 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMq8dN_1xtpCtr2cPmC:-OMq8dN_1xtpCtr2cPmD:bz450l6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e9a7ecc3fd9d328b3a3f7cbdea9783ce1f556f2/mullvad-ios/src/api_client/response.rs#L6)):*
> ```XML
> };
> 
> use mullvad_api::StatusCode;
> ```

This should be moved inside `use mullvad_api` below.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 1 of 17 files reviewed, 3 unresolved discussions (waiting on @pinkisemils and @rablador)

___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMfkSBz5lMQbpZf6Qg3:-OMqIZJF41_6lN8aot-Y:bss4a2b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b9ea13cc3b324525ed142cf96da0b402059ddb9/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L12)):*

Unless we need to I don't think we should. Less code is good code :)
</blockquote></details>

Removed.
___
*[`mullvad-ios/src/api_client/mod.rs` line 107 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMpSpUbA5hzzFuH1azD:-OMqDIoy7KBpDXR7UCTg:b-efhsqk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e9a7ecc3fd9d328b3a3f7cbdea9783ce1f556f2/mullvad-ios/src/api_client/mod.rs#L107)):*

Nit: We could use the existing `do_request`, extract everything but the last part (`SwiftMullvadApiResponse::with\_body(response).await`) and create two new functions: `do_request_with_body` and `do_request_with_empty_body`. Both calling `do_request` and returning the appropriate response.
</blockquote></details>

that was the where Emils and I left for the team to decide. I'll do that.
___
*[`mullvad-ios/src/api_client/response.rs` line 6 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMq8dN_1xtpCtr2cPmC:-OMqHxh11HysWMxJck9x:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e9a7ecc3fd9d328b3a3f7cbdea9783ce1f556f2/mullvad-ios/src/api_client/response.rs#L6)):*

This should be moved inside `use mullvad_api` below.
</blockquote></details>

Done

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 1 of 17 files reviewed, 2 unresolved discussions (waiting on @pinkisemils)

___
*[`mullvad-ios/src/api_client/mod.rs` line 107 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMpSpUbA5hzzFuH1azD:-OMqJQ1k9P_6Exq-o5eM:b7tfllu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e9a7ecc3fd9d328b3a3f7cbdea9783ce1f556f2/mullvad-ios/src/api_client/mod.rs#L107)):*

that was the where Emils and I left for the team to decide. I'll do that.
</blockquote></details>

I realised this isn't completely straightforward to do since param type `T: Future<Output = Result<(), rest::Error>>` is different from same param in `do_request`. Maybe we can let it be.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 5 of 17 files reviewed, 4 unresolved discussions (waiting on @buggmagnet, @pinkisemils, and @rablador)

___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 13 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqKrkG7CENSn9ijw4k:-OMqhHtbCLk75CpGhUBj:b-jtoayy) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e9a7ecc3fd9d328b3a3f7cbdea9783ce1f556f2/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L13)):*

There's no need for any of this, Swift already does the right thing when you pass a `String` type to an FFI and exposes the C buffer underneath. 
You can either consume a `const uint8_t *<pointer_name>` or a `const char *<pointer_name>` in the FFI 

You just need to guarantee that the buffer passed to the FFI is copied at the underlying string buffer will not outlive the instance of the string
</blockquote></details>

In that case I need to dealloc the memory by myself, since `SwiftProblemReportRequest` receives pointer for those properties.
Can you elaborate that with an example?

buggmagnet (changes requested):
> Reviewed 1 of 8 files at r3, 1 of 3 files at r5, 2 of 3 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 8 of 17 files reviewed, 8 unresolved discussions (waiting on @pinkisemils and @rablador)

___
*[`mullvad-ios/src/api_client/problem_report.rs` line 122 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqpU-H4nBy-CFbJDSz:-OMqpU-H4nBy-CFbJDT-:bdqi00k) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/mullvad-ios/src/api_client/problem_report.rs#L122)):*
> ```XML
> 
> impl ProblemReportRequest {
>     unsafe fn from_swift_parameters(request: SwiftProblemReportRequest) -> Option<Self> {
> ```

We should add a `//SAFETY:` comment here to document the `unsafe` usage restrictions
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 175 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqpiuS02NCkekV1IWK:-OMqpiuS02NCkekV1IWL:b-1yg5uw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/mullvad-ios/src/api_client/problem_report.rs#L175)):*
> ```XML
>     }
> 
>     unsafe fn add(&mut self, key: *const c_char, value: *const c_char) -> bool {
> ```

Add `SAFETY: ` comments 
___
*[`ios/MullvadVPNTests/MullvadREST/MullvadApi/RustProblemReportRequestTests.swift` line 51 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMql3bO0JbBeX-Kcnll:-OMql3bO0JbBeX-Kcnlm:b-wyzz6z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/ios/MullvadVPNTests/MullvadREST/MullvadApi/RustProblemReportRequestTests.swift#L51)):*
> ```Swift
>         for byteArray in invalidMetadata {
>             byteArray.withUnsafeBytes { (keyPtr: UnsafeRawBufferPointer) in
>                 byteArray.withUnsafeBytes { (valuePtr: UnsafeRawBufferPointer) in
> ```

Why are we iterating the same `byteArray` twice ? Are the elements supposed to be a key-value pair ?
IMHO this is good that we test these kind of things, but there's a caveat 

- We should force `String` usage instead of manipulating buffers when it comes to user provided data, because `String(init:bytes,encoding:)` will automatically reject invalid UTF-8 sequences (which will remove this entire class of problems)

```swift
  1> import Foundation
  2> String(bytes: [0x80], encoding: .utf8)
$R0: String? = nil
```
- I'm guessing this test is sending bogus data to rust, which is fine. But instead, the rust FFI should have test to cover that it doesn't support invalid input. i.e. We should test Swift APIs in swift, and Rust APIs in Rust. IMHO it doesn't really make sense to test one layer from the other language.
___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 2480 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqeVjT4u0Md_zyGt7-:-OMqeVjT4u0Md_zyGt70:b-etfdh4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/ios/MullvadVPN.xcodeproj/project.pbxproj#L2480)):*
> ```
> 		F0E8E4C42A60499100ED26A3 /* AccountDeletionViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountDeletionViewController.swift; sourceTree = "<group>"; };
> 		F0E8E4C82A604E7400ED26A3 /* AccountDeletionInteractor.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountDeletionInteractor.swift; sourceTree = "<group>"; };
> 		F0EEFB9E2D8D60E1007FE4B3 /* RustProblemReportRequest.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RustProblemReportRequest.swift; sourceTree = "<group>"; };
> ```

We should restrict which files can do FFI calls.
All the Classes / Structs / Objects that do FFI calls should reside in the `MullvadRustRuntime` target 
to avoid having circular dependencies down the line

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 8 of 17 files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @pinkisemils, and @rablador)

___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 2480 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqeVjT4u0Md_zyGt7-:-OMqr0RvD5J7DrkxMI2R:bjakbmt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/ios/MullvadVPN.xcodeproj/project.pbxproj#L2480)):*

We should restrict which files can do FFI calls.
All the Classes / Structs / Objects that do FFI calls should reside in the `MullvadRustRuntime` target 
to avoid having circular dependencies down the line
</blockquote></details>

  `RustProblemReportRequest` depends on `MullvadREST` and `MullvadREST` is dependent to `MullvadRustRuntime`, then if I put `RustProblemReportRequest` in `MullvadRustRuntime` I'll make Circular dependency

buggmagnet (changes requested):
> Reviewed 1 of 3 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 21 files reviewed, 8 unresolved discussions (waiting on @mojganii, @pinkisemils, and @rablador)

___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 13 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqKrkG7CENSn9ijw4k:-OMqvEPvDX2J67JuGkLx:bo7v8q4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e9a7ecc3fd9d328b3a3f7cbdea9783ce1f556f2/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L13)):*
That's the point I'm making, you don't need to think about memory management if you pass a `String` object to Rust directly, Swift will do the "Right thing" :tm: 

> When you call a function that is declared as taking an `UnsafePointer<Type>` argument, you can pass any of the following:
...

> A `String` value, if `Type` is `Int8` or `UInt8`. The string is automatically converted to UTF8 in a zero-terminated buffer, and a pointer to that buffer is passed to the function.
...
> The pointer you pass to the function is only guaranteed to be valid for the duration of the function call. Do not persist the pointer and access it after the function has returned.


In the `unsafe fn add(&mut self, key: *const c_char, value: *const c_char) -> bool` function from Rust, we're already making a copy of the arguments passed in this line
```rust
self.0.insert(key_str.to_owned(), value_str.to_owned());
```

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 21 files reviewed, 8 unresolved discussions (waiting on @mojganii, @pinkisemils, and @rablador)

___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 2480 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqeVjT4u0Md_zyGt7-:-OMqwR1042ly4C2oCcbL:b-oerg4p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/ios/MullvadVPN.xcodeproj/project.pbxproj#L2480)):*

`RustProblemReportRequest` depends on `MullvadREST` and `MullvadREST` is dependent to `MullvadRustRuntime`, then if I put `RustProblemReportRequest` in `MullvadRustRuntime` I'll make Circular dependency
</blockquote></details>

It's only dependent on `MullvadREST` because it's using `ProblemReportRequest` which is declared in there.
If we move `ProblemReportRequest` to `MullvadTypes` instead, we won't have this problem anymore.

The general rule is that `MullvadRustRuntime` should only ever depend on `MullvadTypes` and nothing else.

`MullvadREST` should be safe to import `MullvadRustRuntime`

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 21 files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @pinkisemils, and @rablador)

___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 13 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqKrkG7CENSn9ijw4k:-OMqwET87dRuC5El5Smb:b18brzz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e9a7ecc3fd9d328b3a3f7cbdea9783ce1f556f2/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L13)):*

That's the point I'm making, you don't need to think about memory management if you pass a `String` object to Rust directly, Swift will do the "Right thing" :tm: 

> When you call a function that is declared as taking an `UnsafePointer<Type>` argument, you can pass any of the following:
...

> A `String` value, if `Type` is `Int8` or `UInt8`. The string is automatically converted to UTF8 in a zero-terminated buffer, and a pointer to that buffer is passed to the function.
...
> The pointer you pass to the function is only guaranteed to be valid for the duration of the function call. Do not persist the pointer and access it after the function has returned.


In the `unsafe fn add(&mut self, key: *const c_char, value: *const c_char) -> bool` function from Rust, we're already making a copy of the arguments passed in this line
```rust
self.0.insert(key_str.to_owned(), value_str.to_owned());
```
</blockquote></details>

But I receive this warning when I try pass swift during initialising  :
`Cannot pass 'String' to parameter; argument 'address' must be a pointer that outlives the call to 'init(

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 21 files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @pinkisemils, and @rablador)

___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 2480 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqeVjT4u0Md_zyGt7-:-OMqzZQJBAQaJxDI1j5m:b-plxk61) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/ios/MullvadVPN.xcodeproj/project.pbxproj#L2480)):*

It's only dependent on `MullvadREST` because it's using `ProblemReportRequest` which is declared in there.
If we move `ProblemReportRequest` to `MullvadTypes` instead, we won't have this problem anymore.

The general rule is that `MullvadRustRuntime` should only ever depend on `MullvadTypes` and nothing else.

`MullvadREST` should be safe to import `MullvadRustRuntime`
</blockquote></details>

In that case, It's valid!

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 21 files reviewed, 8 unresolved discussions (waiting on @mojganii, @pinkisemils, and @rablador)

___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 2480 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqeVjT4u0Md_zyGt7-:-OMr-T667aRSfrHHekTu:b-nbcbju) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/ios/MullvadVPN.xcodeproj/project.pbxproj#L2480)):*

In that case, It's valid!
</blockquote></details>

What I mean by that is that I don't want classes in `MullvadREST` to be able to call FFI functions, in order to keep a clean separation of concerns, and avoid leaking abstractions. 

i.e. if something does call an FFI function, it should be abstracted to the rest of the app by an object residing in `MullvadRustRuntime`

This is why it's okay for `MullvadREST` to import `MullvadRustRuntime`

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 21 files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @pinkisemils, and @rablador)

___
*[`ios/MullvadREST/Rust/RustProblemReportRequest.swift` line 24 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqM78K5Fz1WDV7FN3H:-OMqznd56PpbzrvE9IS9:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e9a7ecc3fd9d328b3a3f7cbdea9783ce1f556f2/ios/MullvadREST/Rust/RustProblemReportRequest.swift#L24)):*

You can replace this with (for example)
```swift
        for (key, value) in request.metadata {
            let didAddMetadata = swift_problem_report_meta_data_add(problemReportMetaData, key, value)
            if didAddMetadata {
                logger.error("Failed to add metadata. Key: '\(key)' might be invalid or contain unsupported characters.")
            }
        }
```
</blockquote></details>

Done
___
*[`ios/MullvadVPN.xcodeproj/project.pbxproj` line 2480 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqeVjT4u0Md_zyGt7-:-OMr1KDY4-1SZ-gBQObI:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/ios/MullvadVPN.xcodeproj/project.pbxproj#L2480)):*

What I mean by that is that I don't want classes in `MullvadREST` to be able to call FFI functions, in order to keep a clean separation of concerns, and avoid leaking abstractions. 

i.e. if something does call an FFI function, it should be abstracted to the rest of the app by an object residing in `MullvadRustRuntime`

This is why it's okay for `MullvadREST` to import `MullvadRustRuntime`
</blockquote></details>

Done
___
*[`ios/MullvadVPNTests/MullvadREST/MullvadApi/RustProblemReportRequestTests.swift` line 51 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMql3bO0JbBeX-Kcnll:-OMr1LaZ7fhIFRW4Z3KJ:bnkfu8d) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/ios/MullvadVPNTests/MullvadREST/MullvadApi/RustProblemReportRequestTests.swift#L51)):*

Why are we iterating the same `byteArray` twice ? Are the elements supposed to be a key-value pair ?
IMHO this is good that we test these kind of things, but there's a caveat 

- We should force `String` usage instead of manipulating buffers when it comes to user provided data, because `String(init:bytes,encoding:)` will automatically reject invalid UTF-8 sequences (which will remove this entire class of problems)

```swift
  1> import Foundation
  2> String(bytes: [0x80], encoding: .utf8)
$R0: String? = nil
```
- I'm guessing this test is sending bogus data to rust, which is fine. But instead, the rust FFI should have test to cover that it doesn't support invalid input. i.e. We should test Swift APIs in swift, and Rust APIs in Rust. IMHO it doesn't really make sense to test one layer from the other language.
</blockquote></details>

the best bet is Rust do the test for invalid encoding instead of Swift.  I drop that.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 23 files reviewed, 8 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadREST/Extensions/String+UnsafePointer.swift` line 13 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqKrkG7CENSn9ijw4k:-OMr9sR0DdtJ8_GDZtG2:b-hs3y8a) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5e9a7ecc3fd9d328b3a3f7cbdea9783ce1f556f2/ios/MullvadREST/Extensions/String%2BUnsafePointer.swift#L13)):*

But I receive this warning when I try pass swift during initialising  :
`Cannot pass 'String' to parameter; argument 'address' must be a pointer that outlives the call to 'init(
</blockquote></details>

Since this function returns `SwiftProblemReportRequest`, it's beyond the scope to let Swift create  automatically create a persistent pointer. in the  swift\_problem\_report\_meta\_data\_add works since  making the temporary pointer is fine
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 122 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqpU-H4nBy-CFbJDSz:-OMrHJZj2OoTxzB-GGl_:b4czymi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/mullvad-ios/src/api_client/problem_report.rs#L122)):*

We should add a `//SAFETY:` comment here to document the `unsafe` usage restrictions
</blockquote></details>

I'm not able to see line code
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 175 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMqpiuS02NCkekV1IWK:-OMrHQfS7aqoav7f3pSh:b4czymi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4b54dfcd80b1edf71a26b24fbd29e5f78e18b50b/mullvad-ios/src/api_client/problem_report.rs#L175)):*

Add `SAFETY: ` comments
</blockquote></details>

I'm not able to see line code

buggmagnet (changes requested):
> Reviewed 2 of 13 files at r2, 1 of 7 files at r9, 1 of 5 files at r11, 5 of 15 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 9 of 23 files reviewed, 8 unresolved discussions (waiting on @pinkisemils)

___
*[`ios/MullvadRustRuntime/String+UnsafePointer.swift` line 17 at r14](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMuNQoSDg14IAn556BX:-OMuNQoSDg14IAn556BY:b-26g01g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b8f2e26606006e851d25a4ddb50fac133ab441e3/ios/MullvadRustRuntime/String%2BUnsafePointer.swift#L17)):*
This is dangerous and will at best crash in subtle and hard to debug ways, and at worst, straight up crash. Please read the documentation which explicitly states (emphasis mine) 
>  **Do not store or return the pointer for later use.**

That means that the variable passed in that closure is *only valid during the closure*.
Let's solve this together instead !

buggmagnet (changes requested):
> Reviewed 1 of 15 files at r8, 1 of 13 files at r10, 1 of 5 files at r11, 1 of 10 files at r12, 7 of 15 files at r13, 5 of 6 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 23 of 24 files reviewed, 5 unresolved discussions (waiting on @pinkisemils)

___
*[`ios/MullvadRustRuntime/String+UnsafePointer.swift` line 13 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMvgP5e9UYrxq4c51D1:-OMvgP5fCwCfVtCLB-mt:bbdwjhv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b7bae186e22244874d96b963b77725118ceb03f/ios/MullvadRustRuntime/String%2BUnsafePointer.swift#L13)):*
> ```Swift
> extension String {
>     // Ensure the string is converted to a null-terminated C string
>     // UnsafeMutablePointer provides no automated memory management or alignment guarantees.
> ```

nit
`UnsafePointer`
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 229 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMvgyb74mHXviXAd3Ib:-OMvgyb74mHXviXAd3Ic:b-dnh367) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b7bae186e22244874d96b963b77725118ceb03f/mullvad-ios/src/api_client/problem_report.rs#L229)):*
> ```XML
> pub extern "C" fn swift_problem_report_meta_data_free(mut map: ProblemReportMetadata) {
>     if !map.inner.is_null() {
>         // Safety: `map.inner` must have been allocated with `Box::into_raw()`
> ```

We should replace this with (and similar comments like this one)
```
// SAFETY: `map.inner` must be properly aligned and non-null
```

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 19 of 24 files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`mullvad-ios/src/api_client/problem_report.rs` line 229 at r15](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-OMvgyb74mHXviXAd3Ib:-OMvjJqkDwSe3A8hC2R3:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7b7bae186e22244874d96b963b77725118ceb03f/mullvad-ios/src/api_client/problem_report.rs#L229)):*

We should replace this with (and similar comments like this one)
```
// SAFETY: `map.inner` must be properly aligned and non-null
```
</blockquote></details>

Done

buggmagnet (approved):
:lgtm: Although this only pertains for the Swift part. 
Waiting on our venerable Rustaceans to approve the Rust parts. 

> Reviewed 1 of 10 files at r12, 3 of 5 files at r16, 1 of 2 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-:-ON-aao6BeVL2at9vFJM:b90b7n1)* status: all files reviewed, 4 unresolved discussions (waiting on @pinkisemils)

___
*[`ios/MullvadVPNTests/MullvadREST/MullvadApi/RustProblemReportRequestTests.swift` line 51 at r17](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ON-aTrvA9-fcSWIIA5d:-ON-aTrvA9-fcSWIIA5e:b-rwcq73) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b7fd13375812208b6741a0916b0e92dde87b026/ios/MullvadVPNTests/MullvadREST/MullvadApi/RustProblemReportRequestTests.swift#L51)):*
> ```Swift
>         for byteArray in invalidMetadata {
>             byteArray.withUnsafeBytes { (keyPtr: UnsafeRawBufferPointer) in
>                 byteArray.withUnsafeBytes { (valuePtr: UnsafeRawBufferPointer) in
> ```

Let's remove that 2nd iteration of the same array

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 23 of 24 files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @pinkisemils)

___
*[`ios/MullvadVPNTests/MullvadREST/MullvadApi/RustProblemReportRequestTests.swift` line 51 at r17](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ON-aTrvA9-fcSWIIA5d:-ONEDOgu5bWOHKCq1jyr:biwwwh5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2b7fd13375812208b6741a0916b0e92dde87b026/ios/MullvadVPNTests/MullvadREST/MullvadApi/RustProblemReportRequestTests.swift#L51)):*

Let's remove that 2nd iteration of the same array
</blockquote></details>

for being more readable I replaced dictionary to remove twice iteration which was for having key and value from beginning.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 23 of 24 files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @MarkusPettersson98, and @pinkisemils)

___
*[`mullvad-ios/src/api_client/response.rs` line 62 at r18](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONF9_qLDV_DpRufSFj8:-ONFGcrcEPdnSyRiPEmB:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/adecf43a3087e208b8c2b10fe185f0e712b9b945/mullvad-ios/src/api_client/response.rs#L62)):*

Since this function is infallible, I think the definition should be simplified to

```rust
pub fn ok() -> Self {
    Self {
        success: true,
        error_description: null_mut(),
        body: null_mut(),
        body_size: 0,
        etag: null_mut(),
        status_code: StatusCode::NO_CONTENT.as_u16(),
        server_response_code: null_mut(),
    }
}
```
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 64 at r18](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONF5nYxCzZotX2U5XKN:-ONFE3X65x4b1It6_C77:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/adecf43a3087e208b8c2b10fe185f0e712b9b945/mullvad-ios/src/api_client/problem_report.rs#L64)):*

Nit :pick: Since we want to short-circuit and return early if `ProblemReportRequest::from_swift_parameters(request)` is `None`, we could simplify this expression using a `let-else` expression

```rust
let Some(problem_report_request) = unsafe { ProblemReportRequest::from_swift_parameters(request) } else {
    let err = Error::ApiError(
        rest::StatusCode::BAD_REQUEST,
        "Failed to send problem report: invalid address, message, or log data.".to_string(),
    );
    log::error!("{err:?}");
    completion.finish(SwiftMullvadApiResponse::rest_error(err));
    return SwiftCancelHandle::empty();
};
```
As is done when binding `tokio_handle` a couple of lines above this one :point_up: 

Further reading: https://doc.rust-lang.org/beta/rust-by-example/flow_control/let_else.html
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 134 at r18](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONF7aHW8NYt0DbmrGru:-ONFS7py5rXe6pU9gLSd:b-hpz2o3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/adecf43a3087e208b8c2b10fe185f0e712b9b945/mullvad-ios/src/api_client/problem_report.rs#L134)):*

This `unsafe` block is overly scoped - think about rewriting it slightly to make it easier to vet

```rust
// Safety: `ptr` must be a valid, null-terminated C string.
let cstr = unsafe { CStr::from_ptr(ptr) }
cstr.to_str().map(ToOwned::to_owned).unwrap_or_default()
```
Also, the pointer cast is unnecessary, no?
</blockquote></details>

Casting pointer was unnecessary and removed.
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 212 at r18](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONF9M2EEn6ntdO-Z0vK:-ONFDrUMCWzRrxw2jcmZ:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/adecf43a3087e208b8c2b10fe185f0e712b9b945/mullvad-ios/src/api_client/problem_report.rs#L212)):*

Typo `null-terminated`*
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 224 at r18](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONF9EEh-j2f85-014LH:-ONFFAZD4_cvd4g-S0It:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/adecf43a3087e208b8c2b10fe185f0e712b9b945/mullvad-ios/src/api_client/problem_report.rs#L224)):*

Please add `// Safety`-comments for each `unsafe` block :blush:
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 230 at r18](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONF9TzODtzhODyqI6wF:-ONFFWyF-kjTAS0Ju2f_:bs4nhb1) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/adecf43a3087e208b8c2b10fe185f0e712b9b945/mullvad-ios/src/api_client/problem_report.rs#L230)):*

"has not been" what? Dropped/freed? :smile:
</blockquote></details>

freed indeed :)

buggmagnet (commented):
> Reviewed 1 of 1 files at r18, 3 of 4 files at r19, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: all files reviewed, 5 unresolved discussions (waiting on @MarkusPettersson98 and @pinkisemils)

buggmagnet (dismissed):
:lgtm: (Only pertaining for the iOS parts of this PR)

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-:-ONFkfEb6wc5Wx39Nvfx:bfy5z7n)* status: all files reviewed, 5 unresolved discussions (waiting on @MarkusPettersson98 and @pinkisemils)

dlon (commented):
> Reviewed 2 of 13 files at r2, 1 of 15 files at r8, 1 of 7 files at r9, 1 of 13 files at r10, 2 of 15 files at r13, 1 of 5 files at r16, 2 of 4 files at r19, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 24 files reviewed, 7 unresolved discussions (waiting on @buggmagnet, @MarkusPettersson98, @mojganii, and @pinkisemils)

___
*[`mullvad-ios/src/api_client/problem_report.rs` line 168 at r19](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONGN_G1B30sm_yTDpnN:-ONGN_G279St3-G_dwPm:bom2wt) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7fe4650eb492ed318c48dd15b9c40658ae0a5001/mullvad-ios/src/api_client/problem_report.rs#L168)):*
> ```XML
>     }
> 
>     // SAFETY: key and value must point to null-terminated string
> ```

Prefer doc comments (`///`) and `# Safety` for documenting functions. E.g.

```
/// Add key and value pair to the map
///
/// # Safety
///
/// ...
```

See https://std-dev-guide.rust-lang.org/policy/safety-comments.html
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 170 at r19](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONGN6lxE9TKzQYVjGEe:-ONGN6lxE9TKzQYVjGEf:b-bimpyz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7fe4650eb492ed318c48dd15b9c40658ae0a5001/mullvad-ios/src/api_client/problem_report.rs#L170)):*
> ```XML
>     // SAFETY: key and value must point to null-terminated string
>     unsafe fn add(&mut self, key: *const c_char, value: *const c_char) -> bool {
>         if key.is_null() || value.is_null() {
> ```

Nit: Doesn't matter much, but `assert(!ptr.is_null())` or `debug_assert!(!ptr.is_null())` (if anything) would be more appropriate here, given that the safety contract requires that both values are non-null.

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 24 files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @dlon, @MarkusPettersson98, and @pinkisemils)

___
*[`mullvad-ios/src/api_client/problem_report.rs` line 168 at r19](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONGN_G1B30sm_yTDpnN:-ONJZs2-Cip_wVbzB9zd:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7fe4650eb492ed318c48dd15b9c40658ae0a5001/mullvad-ios/src/api_client/problem_report.rs#L168)):*

Prefer doc comments (`///`) and `# Safety` for documenting functions. E.g.

```
/// Add key and value pair to the map
///
/// # Safety
///
/// ...
```

See https://std-dev-guide.rust-lang.org/policy/safety-comments.html
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 170 at r19](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONGN6lxE9TKzQYVjGEe:-ONJdHSL8efa636MC8ny:b-3cslex) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/7fe4650eb492ed318c48dd15b9c40658ae0a5001/mullvad-ios/src/api_client/problem_report.rs#L170)):*

Nit: Doesn't matter much, but `assert(!ptr.is_null())` or `debug_assert!(!ptr.is_null())` (if anything) would be more appropriate here, given that the safety contract requires that both values are non-null.
</blockquote></details>

indeed!

MarkusPettersson98 (commented):
> Reviewed 1 of 15 files at r8, 1 of 13 files at r10, 2 of 4 files at r19, 1 of 3 files at r20.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 24 files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @dlon, @mojganii, and @pinkisemils)

___
*[`mullvad-ios/src/api_client/problem_report.rs` line 34 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONKHLJn8S217xYKfOta:-ONKHLJn8S217xYKfOtb:b7di83x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b571c4080f3bf1e8bdcee0c8fc3a923ac48d7bc0/mullvad-ios/src/api_client/problem_report.rs#L34)):*
> ```XML
> /// the string properties of `SwiftProblemReportRequest` must be pointers to a null terminated strings.
> ///
> /// This function is not safe to call multiple times with the same `CompletionCookie`.
> ```

Is it not safe to call multiple times, or just not safe to call concurrently with the same `completion_cookie`? If the former, could the doc comment be update to reflect that the ownership of `completion_cookie` is moved to the rust side when calling this function, and that the pointer is no longer safe to use after calling this function?



_Code quote:_
```XML
/// This function is not safe to call multiple times with the same `CompletionCookie`.
```
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 177 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONKEX3J5so1JKVvsrbs:-ONKEX3J5so1JKVvsrbt:b-j4fh6f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b571c4080f3bf1e8bdcee0c8fc3a923ac48d7bc0/mullvad-ios/src/api_client/problem_report.rs#L177)):*
> ```XML
>     /// - `key` must be a null-terminated UTF-8 string, containing LF-separated machines.
>     /// - `value` must be a valid pointer to some valid and aligned pointer-sized memory.
>     /// - The pointer written to `out` is NOT safe to be used concurrently.
> ```

Does this comment belong here? It seems like it has been copied from somewhere and someone forgot to remove it? :blush:



_Code quote:_
```XML
    /// - The pointer written to `out` is NOT safe to be used concurrently.
```
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 225 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONKEIzCCjVU2H6eaghy:-ONKEIzCCjVU2H6eaghz:b-j4fh6f) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b571c4080f3bf1e8bdcee0c8fc3a923ac48d7bc0/mullvad-ios/src/api_client/problem_report.rs#L225)):*
> ```XML
> /// - `key` must be a null-terminated UTF-8 string, containing LF-separated machines.
> /// - `value` must be a valid pointer to some valid and aligned pointer-sized memory.
> /// - The pointer written to `out` is NOT safe to be used concurrently.
> ```

Does this comment belong here? It seems like it has been copied from somewhere and someone forgot to remove it? :blush:



_Code quote:_
```XML
/// - The pointer written to `out` is NOT safe to be used concurrently.
```
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 249 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONKHyDn0j0HSQT6aLHh:-ONKHyDn0j0HSQT6aLHi:bjmnjbs) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b571c4080f3bf1e8bdcee0c8fc3a923ac48d7bc0/mullvad-ios/src/api_client/problem_report.rs#L249)):*
> ```XML
>         unsafe {
>             drop(Box::from_raw(map.inner));
>             map.inner = std::ptr::null_mut();
> ```

`map` is dropped at the end of this function scope, so it seems overkill to assign a null-pointer to `map.inner` :blush:



_Code quote:_
```XML
map.inner = std::ptr::null_mut();
```

acb-mv (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 24 files reviewed, 9 unresolved discussions (waiting on @buggmagnet, @dlon, @MarkusPettersson98, @mojganii, @pinkisemils, and @rablador)

___
*[`ios/MullvadRustRuntime/include/mullvad_rust_runtime.h` line 67 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONOrHc3A4XvhHGEz5sM:-ONOrHc3A4XvhHGEz5sN:bsheghj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b571c4080f3bf1e8bdcee0c8fc3a923ac48d7bc0/ios/MullvadRustRuntime/include/mullvad_rust_runtime.h#L67)):*
> ```objc
>   const char *message;
>   const char *log;
>   struct ProblemReportMetadata meta_data;
> ```

The underscore in `meta_data` looks a little odd to me, given that "metadata" is a well-known word in its own right
___
*[`ios/MullvadRustRuntime/include/mullvad_rust_runtime.h` line 223 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONOreoE1Bx-1H7ZJEbk:-ONOreoE1Bx-1H7ZJEbl:b-8cil3m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b571c4080f3bf1e8bdcee0c8fc3a923ac48d7bc0/ios/MullvadRustRuntime/include/mullvad_rust_runtime.h#L223)):*
> ```objc
>                                         const char *value);
> 
> void swift_problem_report_meta_data_free(struct ProblemReportMetadata map);
> ```

As above: I´d name this `swift_problem_report_metadata_add`and `swift_problem_report_metadata_free`

rablador (commented):
Just thought I should add that @mojganii is the author of this PR, not me. Due to problems with GitHub I has to create a new branch and PR for her.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-:-ONPNjj96B7g2NlZwmeu:b-2klzok)* status: 0 of 24 files reviewed, 9 unresolved discussions (waiting on @buggmagnet, @dlon, @MarkusPettersson98, @mojganii, and @pinkisemils)

buggmagnet (dismissed):
:lgtm:

> Reviewed 2 of 3 files at r20, 7 of 11 files at r21, 5 of 7 files at r22, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-:-ONUMG809F1vn4MSx481:bnfp4nl)* status: 0 of 26 files reviewed, 8 unresolved discussions (waiting on @dlon, @MarkusPettersson98, @mojganii, @pinkisemils, and @rablador)

dlon (changes requested):
> Reviewed 1 of 5 files at r11, 1 of 15 files at r13, 1 of 11 files at r21, 2 of 7 files at r22, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 26 files reviewed, 10 unresolved discussions (waiting on @buggmagnet, @MarkusPettersson98, @mojganii, @pinkisemils, and @rablador)

___
*[`mullvad-ios/src/api_client/problem_report.rs` line 225 at r22](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ON_4F006mZ3WXNgz0jj:-ON_4F006mZ3WXNgz0jk:b5kyfry) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7bcd599ed23ac6a1e4cf9b5d68f30c021f7b0be/mullvad-ios/src/api_client/problem_report.rs#L225)):*
> `self` must be a valid, exclusive pointer to `ProblemReportMetadata`, initialized

Should say `map`. Also, I find it confusing that this is referred to as a pointer, when it's technically not a pointer in the function signature.

> [...] initialized

Is this sentence complete?
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 28 at r22](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ON_2ueE0jF_J5KCF80F:-ON_2ueE0jF_J5KCF80G:b-m84kgh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7bcd599ed23ac6a1e4cf9b5d68f30c021f7b0be/mullvad-ios/src/api_client/problem_report.rs#L28)):*
> ```XML
> /// by calling `mullvad_api_init_new`.
> ///
> /// `completion_cookie` must be pointing to a valid instance of `CompletionCookie`. `CompletionCookie` is
> ```

See analogous docs on other functions. This is not completely correct (e.g., `completion_cookie` is not a pointer to a `CompletionCookie`).
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 178 at r22](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ON_1Npm7PCj5AOUbIN3:-ON_1Npm7PCj5AOUbIN4:b-63o8b9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d7bcd599ed23ac6a1e4cf9b5d68f30c021f7b0be/mullvad-ios/src/api_client/problem_report.rs#L178)):*
> ```XML
>     /// # Safety
>     ///
>     /// `self` must be a valid, exclusive pointer to `Map`, initialized
> ```

`self` does not need to be mentioned here, as it's necessarily valid (or should be). If it's referring to `self.inner`, since it is encapsulated in `Map`, it probably shouldn't be mentioned in the documentation of this public function.

dlon (commented):
> Reviewed 2 of 11 files at r21, 2 of 7 files at r22.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 26 files reviewed, 10 unresolved discussions (waiting on @buggmagnet, @MarkusPettersson98, @mojganii, @pinkisemils, and @rablador)

MarkusPettersson98 (commented):
:lgtm: Rusty bits looks good!

> Reviewed 2 of 11 files at r21, 2 of 7 files at r22.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-:-ONTcmqs0_qSlxAJilrW:b-9doya0)* status: 0 of 26 files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @dlon, @mojganii, @pinkisemils, and @rablador)

MarkusPettersson98 (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 26 files reviewed, 7 unresolved discussions (waiting on @buggmagnet, @dlon, @mojganii, @pinkisemils, and @rablador)

___
*[`mullvad-ios/src/api_client/problem_report.rs` line 36 at r22](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONo13GUDm1ODEvmYoVh:-ONo13GUDm1ODEvmYoVi:bml2cia) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1efaa2ca88f106a21066c9748a8967c981ba3dd8/mullvad-ios/src/api_client/problem_report.rs#L36)):*
> ```XML
> /// This function is not safe to call multiple times with the same `CompletionCookie`.
> ///
> /// Ownership of the `CompletionCookie` is transferred to the Rust side upon calling this function.
> ```

I guess what is referred to here is actually the pointer `completion_cookie`, and not the `CompletionCookie` value (which only exists in Rust-land, and is therefore not relevant to mention in the FFI documentation)?



_Code quote:_
```XML
Ownership of the `CompletionCookie`
```

dlon (changes requested):
> Reviewed 1 of 11 files at r21, 1 of 7 files at r22.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 26 files reviewed, 7 unresolved discussions (waiting on @buggmagnet, @MarkusPettersson98, @mojganii, @pinkisemils, and @rablador)

___
*[`mullvad-ios/src/api_client/problem_report.rs` line 178 at r22](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONp2jtc0JFiOdzPnkek:-ONp2jtc0JFiOdzPnkel:blipe2l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1efaa2ca88f106a21066c9748a8967c981ba3dd8/mullvad-ios/src/api_client/problem_report.rs#L178)):*
> ```XML
>     /// # Safety
>     ///
>     /// `self` must be a valid, exclusive pointer to `Map`, initialized
> ```

`self` does not need to be mentioned here, as it's necessarily valid (or should be).

mojganii (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 26 files reviewed, 7 unresolved discussions (waiting on @buggmagnet, @dlon, @MarkusPettersson98, @pinkisemils, and @rablador)

___
*[`ios/MullvadRustRuntime/include/mullvad_rust_runtime.h` line 223 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONOreoE1Bx-1H7ZJEbk:-ONPx1HL0wFDTRjmC_o5:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b571c4080f3bf1e8bdcee0c8fc3a923ac48d7bc0/ios/MullvadRustRuntime/include/mullvad_rust_runtime.h#L223)):*

As above: I´d name this `swift_problem_report_metadata_add`and `swift_problem_report_metadata_free`
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 34 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONKHLJn8S217xYKfOta:-ONPygso2M5iGf9fq2LR:b-h1xukd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b571c4080f3bf1e8bdcee0c8fc3a923ac48d7bc0/mullvad-ios/src/api_client/problem_report.rs#L34)):*

Is it not safe to call multiple times, or just not safe to call concurrently with the same `completion_cookie`? If the former, could the doc comment be update to reflect that the ownership of `completion_cookie` is moved to the rust side when calling this function, and that the pointer is no longer safe to use after calling this function?
</blockquote></details>

how about?



_Code snippet:_
```XML
/// This function is not safe to call multiple times with the same `CompletionCookie`.
/// 
/// Ownership of the `CompletionCookie` is transferred to the Rust side upon calling this function.
/// After this call, the pointer must be considered invalid and must not be reused or accessed in any way.
/// Calling this function again with the same cookie, either sequentially or concurrently, results in undefined behavior.
```
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 177 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONKEX3J5so1JKVvsrbs:-ONPz-pi-e7xzxYk2aS4:b2wphug) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b571c4080f3bf1e8bdcee0c8fc3a923ac48d7bc0/mullvad-ios/src/api_client/problem_report.rs#L177)):*

Does this comment belong here? It seems like it has been copied from somewhere and someone forgot to remove it? :blush:
</blockquote></details>

my bad, removed.
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 225 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONKEIzCCjVU2H6eaghy:-ONPz5RrDAuBTjtG-Psh:bss4a2b) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b571c4080f3bf1e8bdcee0c8fc3a923ac48d7bc0/mullvad-ios/src/api_client/problem_report.rs#L225)):*

Does this comment belong here? It seems like it has been copied from somewhere and someone forgot to remove it? :blush:
</blockquote></details>

Removed.
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 249 at r20](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONKHyDn0j0HSQT6aLHh:-ONPzEQyA9DDvxVw7nL8:ba3dfob) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b571c4080f3bf1e8bdcee0c8fc3a923ac48d7bc0/mullvad-ios/src/api_client/problem_report.rs#L249)):*

`map` is dropped at the end of this function scope, so it seems overkill to assign a null-pointer to `map.inner` :blush:
</blockquote></details>

Removed
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 36 at r22](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925#-ONo13GUDm1ODEvmYoVh:-ONstk1p1Zf6EHDJQTax:b-mwhlyf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1efaa2ca88f106a21066c9748a8967c981ba3dd8/mullvad-ios/src/api_client/problem_report.rs#L36)):*

I guess what is referred to here is actually the pointer `completion_cookie`, and not the `CompletionCookie` value (which only exists in Rust-land, and is therefore not relevant to mention in the FFI documentation)?
</blockquote></details>

I changed that to follow the function signature we already have had in the rest iOS APIs

dlon (commented):
> Reviewed 1 of 3 files at r23.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 26 files reviewed, 6 unresolved discussions (waiting on @buggmagnet, @MarkusPettersson98, @mojganii, @pinkisemils, and @rablador)

dlon (commented):
> Reviewed 1 of 3 files at r23, 1 of 2 files at r24.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 26 files reviewed, 4 unresolved discussions (waiting on @buggmagnet, @MarkusPettersson98, @mojganii, @pinkisemils, and @rablador)

rablador (commented):
> Dismissed @pinkisemils from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 26 files reviewed, 3 unresolved discussions (waiting on @buggmagnet, @dlon, @MarkusPettersson98, and @mojganii)

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7925)* status: 0 of 26 files reviewed, all discussions resolved (waiting on @buggmagnet, @MarkusPettersson98, and @mojganii)

PR #7929, opened 2025-03-31
Disable pre-generation of PQ keys on Android

Rawa:
By generating 2 PQ keys on start we slow the start up of the android app with out about 20%. On android we don't have the daemon running continously and by default PQ nor Multihop is enabled, thus all our users doing a cold start of the app will be 20% slower. 

Ideally the solution would be to conditionally & dynamically change the buffer size depending on what features are active, e.g PQ gives buffer of 1, PQ and Multihop gives a buffer of 2.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7929)

### DIFF
--- a/talpid-core/src/tunnel_state_machine/mod.rs
+++ b/talpid-core/src/tunnel_state_machine/mod.rs
@@ -25,6 +25,7 @@ use talpid_routing::RouteManagerHandle;
 #[cfg(target_os = "macos")]
 use talpid_tunnel::TunnelMetadata;
 use talpid_tunnel::{tun_provider::TunProvider, TunnelEvent};
+#[cfg(not(target_os = "android"))]
 use talpid_tunnel_config_client::classic_mceliece::spawn_keypair_generator;
 #[cfg(target_os = "macos")]
 use talpid_types::ErrorExt;
@@ -181,6 +182,10 @@ pub async fn spawn(
     });
 
     // Spawn a worker that pre-computes McEliece key pairs for PQ tunnels
+    //
+    // On Android we have a different lifecycle of the daemon and creating new keys on start up
+    // comes at a high cost, thus we let the generator be created lazily.
+    #[cfg(not(target_os = "android"))]
     spawn_keypair_generator();
 
     Ok(TunnelStateMachineHandle {


albin-mullvad (dismissed):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7929#-:-OMgjxvxEdPm5mJxEZDx:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Rawa (commented):
The connection times when using PQ will be slower, taking about 1s from disconnected to connected, comparative to a few 100 ms, when having the app open and pressing connect. In other scenarios e.g pressing connect through tile, notification or booting with always-on VPN it is probably little to no improvement (however not tested). I think the main benefit would be having PQ and then switching relay, settings or reconnecting.

1. This would improve it for PQ but still affect all users' boot performance w/o having any benefit for the default scenario. Since we only generate 1 key it probably will be slightly better.
2. Since the daemon is stopped unless the user is in a connected state and not having the app in the foreground (not running as a daemon on desktop). This would basically help only help on re-connections (given that they use PQ) and first connection would be slower as it is today. It would also complicate the logic, dividing the implementation between desktop & android.
3. I don't think it would help the boot scenario since it takes about 1 second for the app to start on a Pixel 8A, it feels also quite arbitrary, we don't know that 500ms is a good number for all set of devices. This still has the problem of still generating keys when they are not to be used.

Disabling the pre-generation is not an ideal solution. However, we need to keep in mind that we are running on a mobile device where the entire daemon is killed upon the user being disconnected and not having the app in foreground, thus the caching does not add as much value as it does on desktop where the daemon is running continuously.

Goal with this PR is not to eliminate the logic all together long-term, I'd say we still want to do something like this on mobile, but rather unblock us from creating a new release.

 Let's discuss further offline.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7929#-:-OMk1A_e5uNBUo28vqop:bujjynd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Rawa (commented):
After discussion, we've decided to simply start the keypair generator lazily on Android. This means if the users doesn't have PQ on the worker wont be started. Thus eliminating the pre-generation of keys, once a key is requested the worker will then be started automatically.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7929#-:-OMkKGWu4uA5ClZytU-t:baod75y)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @Rawa)

Serock3 (dismissed):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7929)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Pururun (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7929#-:-OMkYLN67rcAUqJyVPai:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7934, opened 2025-04-01
Update ListItem hover and pressed state colors

olmoh:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7934)


tobias-jarvelov (approved):

desktop/packages/mullvad-vpn/src/renderer/lib/components/list-item/components/list-item-item/ListItemItem.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { useBackgroundColor } from './hooks';
+
+export interface ListItemItemProps {
+  children: React.ReactNode;
+}
+
+const StyledDiv = styled.div`
+  ${() => {
+    const backgroundColor = useBackgroundColor();
```
tobias-jarvelov, line 11: **note:** Didn't see this the first time around. This looks kinda funky and even if this works I'd argue we shouldn't use a hook inside a styled component. I can't say that it is an antipattern, but it at least feels like a code smell to me. Perhaps merely on the basis that it isn't a common pattern and the implications of calling a hook inside a styled component isn't known.

**suggestion:** Call the `useListItem` hook inside the `ListItemItem` component instead and pass `level` and `disabled` to the styled component as props, and then resolve the background color inside the arrow function:

```js
const StyledDiv = styled.div<TransientProps<Pick<ListItemContext, 'disabled' | 'level'>>`
  ${({ disabled, level }) => {
  const backgroundColor =   disabled ? levels[level].disabled : levels[level].enabled;

  return css`
    --background-color: ${backgroundColor};
  ...
```

```jsx
export const ListItemItem = ({ children }: ListItemItemProps) => {
  const { disabled, level } = useListItem();
  return <StyledDiv $disabled={disabled} $level={level}>{children}</StyledDiv>;
};
```

desktop/packages/mullvad-vpn/src/renderer/lib/components/list-item/components/list-item-trigger/ListItemTrigger.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { ButtonBase } from '../../../button';
+import { useListItem } from '../../ListItemContext';
+
+// TODO: Colors should be replace with
+// with new color tokens once they are implemented.
+const StyledButton = styled(ButtonBase)<{ $disabled?: boolean }>`
+  display: flex;
+  width: 100%;
+  ${() => {
+    const { disabled } = useListItem();
```
tobias-jarvelov, line 13: **note:** See my other comment regarding hooks in styled components.

tobias-jarvelov (changes requested):

desktop/packages/mullvad-vpn/src/renderer/lib/components/list-item/components/list-item-item/ListItemItem.tsx
olmoh: Great feedback, makes sense and I agree that we should avoid this. Will fix.

desktop/packages/mullvad-vpn/src/renderer/lib/components/list-item/components/list-item-trigger/ListItemTrigger.tsx
```diff
@@ -0,0 +1,44 @@
...
+        }
+      `}
+      &&:focus-visible {
+        outline: 2px solid ${Colors.white};
+        outline-offset: -2px;
+        z-index: 10;
+      }
+    `;
+  }}
+`;
```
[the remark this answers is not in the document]
olmoh: Nice suggestion, looks cleaner so added it 💪

tobias-jarvelov (approved):
Looks good, but we probably want to convert the arrow functions to `function` declarations.

PR #7942, opened 2025-04-02
Mark trusted-metadata-signing-pubkeys as needing signatures to be edited

faern:
This file is extremely important to the installer downloader (and soon in app upgrades). When adding these keys we put strict codeowners permissions on it. But we should also make sure the file is always signed when changed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7942)

### DIFF
--- a/.github/workflows/verify-locked-down-signatures.yml
+++ b/.github/workflows/verify-locked-down-signatures.yml
@@ -32,6 +32,7 @@ on:
       - building/linux-container-image.txt
       - building/android-container-image.txt
       - building/sigstore/**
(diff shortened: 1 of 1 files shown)


albin-mullvad (approved):
:lgtm: :lock: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7942#-:-OMqkBrK81dqmzo4Zhdp:bulrcak)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #7946, opened 2025-04-02
Use mullvad api instead of urlsession in accounts proxy

rablador:
**Why this needs to be done**

All API calls should use the mullvad-api client, without replacing existing implementation.

**What needs to be done**

The account proxy should be implemented using mullvad-api, allowing us to move to the new implementation without removing the old code.

**Acceptance criteria**

There is a AccountsProxy implementation that uses mullvad-api.

**How to test**

Replace impl in `RESTProxyFactory -> createAccountsProxy()` with:
```
MullvadAccountProxy(
    transportProvider: APITransportProvider(requestFactory: MullvadApiRequestFactory(apiContext: apiContext)),
)
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946)

### DIFF
--- a/ios/MullvadMockData/MullvadREST/AccountsProxy+Stubs.swift
+++ b/ios/MullvadMockData/MullvadREST/AccountsProxy+Stubs.swift
@@ -13,25 +13,28 @@ import MullvadTypes
 struct AccountProxyStubError: Error {}
 
 struct AccountsProxyStub: RESTAccountHandling {
-    var createAccountResult: Result<REST.NewAccountData, Error> = .failure(AccountProxyStubError())
+    var createAccountResult: Result<NewAccountData, Error> = .failure(AccountProxyStubError())
     var deleteAccountResult: Result<Void, Error> = .failure(AccountProxyStubError())
     func createAccount(
         retryStrategy: REST.RetryStrategy,
-        completion: @escaping ProxyCompletionHandler<REST.NewAccountData>
+        completion: @escaping ProxyCompletionHandler<NewAccountData>
     ) -> Cancellable {
         completion(createAccountResult)
         return AnyCancellable()
     }
 
-    func getAccountData(accountNumber: String) -> any RESTRequestExecutor<Account> {
-        RESTRequestExecutorStub<Account>(success: {
-            Account(
-                id: accountNumber,
-                expiry: Calendar.current.date(byAdding: .day, value: 38, to: Date())!,
-                maxDevices: 1,
-                canAddDevices: true
-            )
-        })
+    func getAccountData(
+        accountNumber: String,
+        retryStrategy: REST.RetryStrategy,
+        completion: @escaping ProxyCompletionHandler<Account>
+    ) -> Cancellable {
+        completion(.success(Account(
+            id: accountNumber,
+            expiry: Calendar.current.date(byAdding: .day, value: 38, to: Date())!,
+            maxDevices: 1,
+            canAddDevices: true
+        )))
+        return AnyCancellable()
     }
 
     func deleteAccount(
--- a/ios/MullvadMockData/MullvadTypes/NewAccountDataMock.swift
+++ b/ios/MullvadMockData/MullvadTypes/NewAccountDataMock.swift
@@ -1,0 +1,23 @@
+//
+//  NewAccountDataMock.swift
+//  MullvadVPN
+//
+//  Created by Jon Petersson on 2025-04-08.
+//  Copyright © 2025 Mullvad VPN AB. All rights reserved.
+//
+
+import MullvadTypes
+
+extension NewAccountData {
+    public static func mockValue() -> NewAccountData {
+        return NewAccountData(
+            id: UUID().uuidString,
+            expiry: Date().addingTimeInterval(3600),
+            maxPorts: 2,
+            canAddPorts: false,
+            maxDevices: 5,
+            canAddDevices: false,
+            number: "1234567890123456"
+        )
+    }
+}
--- a/ios/MullvadREST/ApiHandlers/RESTAPIProxy.swift
+++ b/ios/MullvadREST/ApiHandlers/RESTAPIProxy.swift
@@ -12,37 +12,6 @@ import MullvadTypes
 import Operations
 import WireGuardKitTypes
 
-public protocol APIQuerying: Sendable {
-    func getAddressList(
-        retryStrategy: REST.RetryStrategy,
-        completionHandler: @escaping @Sendable ProxyCompletionHandler<[AnyIPEndpoint]>
-    ) -> Cancellable
-
-    func getRelays(
-        etag: String?,
-        retryStrategy: REST.RetryStrategy,
-        completionHandler: @escaping @Sendable ProxyCompletionHandler<REST.ServerRelaysCacheResponse>
-    ) -> Cancellable
-
-    func createApplePayment(
-        accountNumber: String,
-        receiptString: Data
-    ) -> any RESTRequestExecutor<REST.CreateApplePaymentResponse>
-
-    func sendProblemReport(
-        _ body: REST.ProblemReportRequest,
-        retryStrategy: REST.RetryStrategy,
-        completionHandler: @escaping @Sendable ProxyCompletionHandler<Void>
-    ) -> Cancellable
-
-    func submitVoucher(
-        voucherCode: String,
-        accountNumber: String,
-        retryStrategy: REST.RetryStrategy,
-        completionHandler: @escaping @Sendable ProxyCompletionHandler<REST.SubmitVoucherResponse>
-    ) -> Cancellable
-}
-
 extension REST {
     public final class APIProxy: Proxy<AuthProxyConfiguration>, APIQuerying, @unchecked Sendable {
         public init(configuration: AuthProxyConfiguration) {
--- a/ios/MullvadREST/ApiHandlers/RESTAccountsProxy.swift
+++ b/ios/MullvadREST/ApiHandlers/RESTAccountsProxy.swift
@@ -9,21 +9,6 @@
 import Foundation
 import MullvadTypes
 
-public protocol RESTAccountHandling: Sendable {
-    func createAccount(
-        retryStrategy: REST.RetryStrategy,
-        completion: @escaping @Sendable ProxyCompletionHandler<REST.NewAccountData>
-    ) -> Cancellable
-
-    func getAccountData(accountNumber: String) -> any RESTRequestExecutor<Account>
-
-    func deleteAccount(
-        accountNumber: String,
-        retryStrategy: REST.RetryStrategy,
-        completion: @escaping ProxyCompletionHandler<Void>
-    ) -> Cancellable
-}
-
 extension REST {
     public final class AccountsProxy: Proxy<AuthProxyConfiguration>, RESTAccountHandling, @unchecked Sendable {
         public init(configuration: AuthProxyConfiguration) {
@@ -64,7 +49,11 @@ extension REST {
             return executor.execute(retryStrategy: retryStrategy, completionHandler: completion)
         }
 
-        public func getAccountData(accountNumber: String) -> any RESTRequestExecutor<Account> {
+        public func getAccountData(
+            accountNumber: String,
+            retryStrategy: REST.RetryStrategy,
+            completion: @escaping ProxyCompletionHandler<Account>
+        ) -> Cancellable {
             let requestHandler = AnyRequestHandler(
                 createURLRequest: { endpoint, authorization in
                     var requestBuilder = try self.requestFactory.createRequestBuilder(
@@ -85,11 +74,13 @@ extension REST {
                 with: responseDecoder
             )
 
-            return makeRequestExecutor(
+            let executor = makeRequestExecutor(
                 name: "get-my-account",
                 requestHandler: requestHandler,
                 responseHandler: responseHandler
             )
+
+            return executor.execute(retryStrategy: retryStrategy, completionHandler: completion)
         }
 
         public func deleteAccount(
@@ -133,29 +124,5 @@ extension REST {
 
             return executor.execute(retryStrategy: retryStrategy, completionHandler: completion)
         }
-    }
-
-    public struct NewAccountData: Decodable, Sendable {
-        public let id: String
-        public let expiry: Date
-        public let maxPorts: Int
-        public let canAddPorts: Bool
-        public let maxDevices: Int
-        public let canAddDevices: Bool
-        public let number: String
-    }
-}
-
-extension REST.NewAccountData {
-    public static func mockValue() -> REST.NewAccountData {
-        return REST.NewAccountData(
-            id: UUID().uuidString,
-            expiry: Date().addingTimeInterval(3600),
-            maxPorts: 2,
-            canAddPorts: false,
-            maxDevices: 5,
-            canAddDevices: false,
-            number: "1234567890123456"
-        )
     }
 }
--- a/ios/MullvadREST/ApiHandlers/ServerRelaysResponse.swift
+++ b/ios/MullvadREST/ApiHandlers/ServerRelaysResponse.swift
@@ -138,7 +138,11 @@ extension REST {
         public let wireguard: ServerWireguardTunnels
         public let bridge: ServerBridges
 
-        public init(locations: [String: ServerLocation], wireguard: ServerWireguardTunnels, bridge: ServerBridges) {
+        public init(
+            locations: [String: ServerLocation],
+            wireguard: ServerWireguardTunnels,
+            bridge: ServerBridges
+        ) {
             self.locations = locations
             self.wireguard = wireguard
             self.bridge = bridge
--- a/ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAPIProxy.swift
+++ b/ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAPIProxy.swift
@@ -11,6 +11,37 @@ import MullvadTypes
 import Operations
 import WireGuardKitTypes
 
+public protocol APIQuerying: Sendable {
+    func getAddressList(
+        retryStrategy: REST.RetryStrategy,
+        completionHandler: @escaping @Sendable ProxyCompletionHandler<[AnyIPEndpoint]>
+    ) -> Cancellable
+
+    func getRelays(
+        etag: String?,
+        retryStrategy: REST.RetryStrategy,
+        completionHandler: @escaping @Sendable ProxyCompletionHandler<REST.ServerRelaysCacheResponse>
+    ) -> Cancellable
+
+    func createApplePayment(
+        accountNumber: String,
+        receiptString: Data
+    ) -> any RESTRequestExecutor<REST.CreateApplePaymentResponse>
+
+    func sendProblemReport(
+        _ body: REST.ProblemReportRequest,
+        retryStrategy: REST.RetryStrategy,
+        completionHandler: @escaping @Sendable ProxyCompletionHandler<Void>
+    ) -> Cancellable
+
+    func submitVoucher(
+        voucherCode: String,
+        accountNumber: String,
+        retryStrategy: REST.RetryStrategy,
+        completionHandler: @escaping @Sendable ProxyCompletionHandler<REST.SubmitVoucherResponse>
+    ) -> Cancellable
+}
+
 extension REST {
     public final class MullvadAPIProxy: APIQuerying, @unchecked Sendable {
         let transportProvider: APITransportProviderProtocol
--- a/ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift
+++ b/ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift
@@ -1,0 +1,125 @@
+//
+//  MullvadAccountProxy.swift
+//  MullvadVPN
+//
+//  Created by Jon Petersson on 2025-03-31.
+//  Copyright © 2025 Mullvad VPN AB. All rights reserved.
+//
+
+import MullvadRustRuntime
+import MullvadTypes
+import Operations
+import WireGuardKitTypes
+
+public protocol RESTAccountHandling: Sendable {
+    func createAccount(
+        retryStrategy: REST.RetryStrategy,
+        completion: @escaping @Sendable ProxyCompletionHandler<NewAccountData>
+    ) -> Cancellable
+
+    func getAccountData(
+        accountNumber: String,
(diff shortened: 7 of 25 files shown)


acb-mv (commented):
> Reviewed 12 of 20 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946)* status: 12 of 20 files reviewed, 1 unresolved discussion

___
*[`ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift` line 22 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-OMvmBWpAz4rinPykR6w:-OMvmBWpAz4rinPykR6x:b-mx5dq6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fe0e929a0596ee26ad835d34130acc7a5b838fcf/ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift#L22)):*
> ```Swift
>     func getAccountData(
>         accountNumber: String,
>         retryStrategy: REST.RetryStrategy,
> ```

Perhaps it would make sense to add a default value of `.default` for `retryStrategy` to avoid needing to add this in cases where the default is good enough?

buggmagnet (dismissed):
:lgtm:

> Reviewed 18 of 20 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-:-OMvue-886yrjKyw-JvX:bnfp4nl)* status: 19 of 20 files reviewed, 1 unresolved discussion (waiting on @rablador)

MarkusPettersson98 (commented):
> Reviewed 1 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946)* status: 19 of 20 files reviewed, 5 unresolved discussions (waiting on @rablador)

___
*[`mullvad-ios/src/api_client/account.rs` line 44 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-OMzqxLoCR-mh4bQCshJ:-OMzqxLoCR-mh4bQCshK:bigc0bv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-ios/src/api_client/account.rs#L44)):*
> ```XML
> 
>     let api_context = api_context.into_rust_context();
>     let retry_strategy = unsafe { retry_strategy.into_rust() };
> ```

Please annotate all use of `unsafe` with `// SAFETY: <why this is safe>` comments :blush:



_Code quote:_
```XML
    let retry_strategy = unsafe { retry_strategy.into_rust() };
```
___
*[`mullvad-api/src/lib.rs` line 514 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-OMzpTyO0XFyU_1TxzLS:-OMzpTyO0XFyU_1TxzLT:bxgwgnf) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-api/src/lib.rs#L514)):*
> ```XML
>             let data = request.await?.deserialize().await?;
>             Ok(data)
>         }
> ```

:pick: In this case, the `let data = ..` binding is overkill since the branch operator `?` doesn't coerce the `Error` returned by `deserialize` into a different error type. Thus, the closure can be simplified to

```rust
async move { request.await?.deserialize().await }
```



_Code quote:_
```XML
        async move {
            let data = request.await?.deserialize().await?;
            Ok(data)
        }
```

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946)* status: 19 of 20 files reviewed, 5 unresolved discussions (waiting on @acb-mv and @MarkusPettersson98)

___
*[`ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift` line 22 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-OMvmBWpAz4rinPykR6w:-ON-QjYrCTCzssW41Acu:b-yohi24) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fe0e929a0596ee26ad835d34130acc7a5b838fcf/ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift#L22)):*

Perhaps it would make sense to add a default value of `.default` for `retryStrategy` to avoid needing to add this in cases where the default is good enough?
</blockquote></details>

Current default actually seems to be `noRetry` for `RequestExecutor`. We could add a default for this, but I think it might give the wrong idea that `default` is used if omitted.
___
*[`mullvad-api/src/lib.rs` line 514 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-OMzpTyO0XFyU_1TxzLS:-ON-Re-X1FwWbXLLdTLk:bcn3o5h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-api/src/lib.rs#L514)):*

:pick: In this case, the `let data = ..` binding is overkill since the branch operator `?` doesn't coerce the `Error` returned by `deserialize` into a different error type. Thus, the closure can be simplified to

```rust
async move { request.await?.deserialize().await }
```
</blockquote></details>

Ah, right!

MarkusPettersson98 (commented):
> Reviewed 3 of 20 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946)* status: all files reviewed, 7 unresolved discussions (waiting on @acb-mv and @rablador)

___
*[`mullvad-ios/src/api_client/response.rs` line 62 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-ON067Pr8Gbs3ht2OgNU:-ON067Pr8Gbs3ht2OgNV:b-4ocd7m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-ios/src/api_client/response.rs#L62)):*
> ```XML
>             server_response_code: null_mut(),
>         })
>     }
> ```

This doesn't seem fallible, so the type signature & implementation can (and should) be simplified

```rust
pub fn ok() -> Self {
    Self {
        success: true,
        error_description: null_mut(),
        body: null_mut(),
        body_size: 0,
        etag: null_mut(),
        status_code: StatusCode::NO_CONTENT.as_u16(),
        server_response_code: null_mut(),
    }
}
```



_Code quote:_
```XML
    pub fn ok() -> Result<Self, rest::Error> {
        Ok(Self {
            success: true,
            error_description: null_mut(),
            body: null_mut(),
            body_size: 0,
            etag: null_mut(),
            status_code: StatusCode::NO_CONTENT.as_u16(),
            server_response_code: null_mut(),
        })
    }
```
___
*[`mullvad-ios/src/api_client/account.rs` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-ON01iRC7TBJh9crcHe3:-ON01iRC7TBJh9crcHe4:b72qb9g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-ios/src/api_client/account.rs#L34)):*
> ```XML
>     completion_cookie: *mut libc::c_void,
>     retry_strategy: SwiftRetryStrategy,
>     account_number: *const u8,
> ```

Prefer to use the portable `c_char` type instead

```rust
account_number: *const std::os::raw::c_char
```

Preferably importing `std::os::raw::c_char` to the current module

```rust
use std::os::raw::c_char;

..

#[no_mangle]
pub unsafe extern "C" fn mullvad_api_get_account(
    api_context: SwiftApiContext,
    completion_cookie: *mut libc::c_void,
    retry_strategy: SwiftRetryStrategy,
    account_number: *const c_char,
) -> SwiftCancelHandle {
```

This comment applies to all places where you want to convert a pointer to a `CStr`. As always, remember to document clearly what invariants the caller is expected to uphold for `CStr::from_ptr` to be safe to call :blush: 



_Code quote:_
```XML
account_number: *const u8,
```

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946)* status: all files reviewed, 4 unresolved discussions (waiting on @acb-mv, @dlon, and @MarkusPettersson98)

___
*[`mullvad-ios/src/api_client/account.rs` line 23 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-OMw893x7vdX4Fz_X5Ee:-ON-ma920mhcca0tAICX:bxqacct) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-ios/src/api_client/account.rs#L23)):*

I think rather than make assumptions about the lifetime of `completion_cookie`, the comment should just explain what the conditions are for safely calling this function.

Also, it's clearly not a pointer to a `CompletionCookie`, since the function wraps the pointer in a `CompletionCookie` :shrug: It seems to be a pointer to a `MullvadApiCompletion`?

Could we simply write that the function takes ownership of `completion_cookie`, that it must be a pointer to a `MullvadApiCompletion` that the function is responsible for freeing it using `mullvad_api_completion_finish`? Or something closer to that

(Same applies to a bunch of other functions.)
</blockquote></details>

Something like this?

```
/// This function takes ownership of `completion_cookie`, which must be pointing to a valid instance of Swift 
/// object `MullvadApiCompletion`. The pointer must be freed by calling `mullvad_api_completion_finish` 
/// when completion finishes (in completion.finish).
```
___
*[`mullvad-ios/src/api_client/account.rs` line 34 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-ON01iRC7TBJh9crcHe3:-ONE83itBWCKgY7GMk2G:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-ios/src/api_client/account.rs#L34)):*

Prefer to use the portable `c_char` type instead

```rust
account_number: *const std::os::raw::c_char
```

Preferably importing `std::os::raw::c_char` to the current module

```rust
use std::os::raw::c_char;

..

#[no_mangle]
pub unsafe extern "C" fn mullvad_api_get_account(
    api_context: SwiftApiContext,
    completion_cookie: *mut libc::c_void,
    retry_strategy: SwiftRetryStrategy,
    account_number: *const c_char,
) -> SwiftCancelHandle {
```

This comment applies to all places where you want to convert a pointer to a `CStr`. As always, remember to document clearly what invariants the caller is expected to uphold for `CStr::from_ptr` to be safe to call :blush:
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/account.rs` line 36 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-OMwKyWx5-0oyDGnMZVu:-ONE6TVkD8QrY7XWSX3F:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-ios/src/api_client/account.rs#L36)):*

Slightly off topic, but there's no way that `completion_handler.finish` is a safe function. You can construct a `CompletionCookie` using an arbitrary pointer. At the very least, `CompletionCookie` should hide the internal pointer and be constructed using an unsafe `new` function. If you did that, then `finish` would be safe.
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/response.rs` line 62 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-ON067Pr8Gbs3ht2OgNU:-ONE8tD97YupSafgxJ34:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-ios/src/api_client/response.rs#L62)):*

This doesn't seem fallible, so the type signature & implementation can (and should) be simplified

```rust
pub fn ok() -> Self {
    Self {
        success: true,
        error_description: null_mut(),
        body: null_mut(),
        body_size: 0,
        etag: null_mut(),
        status_code: StatusCode::NO_CONTENT.as_u16(),
        server_response_code: null_mut(),
    }
}
```
</blockquote></details>

Done

MarkusPettersson98 (commented):
> Reviewed 6 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946)* status: 21 of 22 files reviewed, 4 unresolved discussions (waiting on @acb-mv, @buggmagnet, @dlon, and @rablador)

___
*[`mullvad-ios/src/api_client/account.rs` line 23 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-OMw893x7vdX4Fz_X5Ee:-ONFHRhcEhmqjiemBK06:bwc1s7x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-ios/src/api_client/account.rs#L23)):*

Something like this?

```
/// This function takes ownership of `completion_cookie`, which must be pointing to a valid instance of Swift 
/// object `MullvadApiCompletion`. The pointer must be freed by calling `mullvad_api_completion_finish` 
/// when completion finishes (in completion.finish).
```
</blockquote></details>

If `mullvad_api_completion_finish` frees `completion_cookie`, I think this safety remark makes sense :+1: 
___
*[`mullvad-ios/src/api_client/completion.rs` line 31 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-ONFGQqy5pDGX7X_KwDY:-ONFGQqy5pDGX7X_KwDZ:bcru6ls) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c7d3eb3664d5c8a1be550777cc9dba75e64976f/mullvad-ios/src/api_client/completion.rs#L31)):*
> ```XML
>         Self { inner }
>     }
> }
> ```

Nit this constructor might as well be marked as safe. Since we're not trying to read from the pointer, we shouldn't be able to cause undefined behavior.

Is there a case for validating the pointer / what the pointer is pointing at? If so, then you could have an unsafe constructor which does not validation, and then a safe ditto which performs validation.



_Code quote:_
```XML
impl CompletionCookie {
    pub unsafe fn new(inner: *mut std::ffi::c_void) -> Self {
        Self { inner }
    }
}
```

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946)* status: 21 of 22 files reviewed, 4 unresolved discussions (waiting on @acb-mv, @buggmagnet, @MarkusPettersson98, and @rablador)

___
*[`mullvad-ios/src/api_client/account.rs` line 23 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-OMw893x7vdX4Fz_X5Ee:-ONFP5W9AxnDLhusbi8H:b-eyo05w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-ios/src/api_client/account.rs#L23)):*

If `mullvad_api_completion_finish` frees `completion_cookie`, I think this safety remark makes sense :+1:
</blockquote></details>

Sounds good! I'd change this to "will be freed by  `mullvad_api_completion_finish` ..." instead of "must be freed ..." 

"Must be freed by" implies that the caller is responsible for doing this, IMO
___
*[`mullvad-ios/src/api_client/completion.rs` line 31 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-ONFGQqy5pDGX7X_KwDY:-ONFPwCi9Ef2ySp9Jj-d:bprz6rg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5c7d3eb3664d5c8a1be550777cc9dba75e64976f/mullvad-ios/src/api_client/completion.rs#L31)):*

Nit this constructor might as well be marked as safe. Since we're not trying to read from the pointer, we shouldn't be able to cause undefined behavior.

Is there a case for validating the pointer / what the pointer is pointing at? If so, then you could have an unsafe constructor which does not validation, and then a safe ditto which performs validation.
</blockquote></details>

The case is that `Self::finish` is currently a safe function, but calling `mullvad_api_completion_finish` with an arbitrary pointer isn't safe. So you could call `SwiftCompletionHandler::new(CompletionCookie::new(ptr::null_mut())).finish(...)` and get a segfault without even using `unsafe`.

If `new` is unsafe, and the documentation points out that `inner` must be a valid pointer to a `MullvadApiCompletion` that we're responsible for freeing using `mullvad_api_completion_finish`, then `SwiftCompletionHandler::finish` can be assumed to be safe since the caller has promised this when calling `new`.

Does that make sense?

Regardless, I think this should be documented

buggmagnet (approved):
:lgtm: (only pertaining for the iOS parts)

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-:-ONFlt_MC80MIWBn9Mrs:b-5nxdsa)* status: all files reviewed, 4 unresolved discussions (waiting on @acb-mv, @MarkusPettersson98, and @rablador)

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946)* status: 18 of 22 files reviewed, 6 unresolved discussions (waiting on @acb-mv, @buggmagnet, @MarkusPettersson98, and @mojganii)

___
*[`ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift` line 34 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-ONJkJM65JyT9UpozzN8:-ONK859rAzyU9vOZVQHx:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f1d0319c62dc3f975bbfabff4bd7a59d30517cc5/ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift#L34)):*

I highly recommend to aggregate all rest type structure in `RESTTypes` at `MullvadTypes`. we've already divided them between `MullvadTypes` and `MullvadRest`
</blockquote></details>

Done.
___
*[`ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift` line 115 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-ONJktrN-Ampf2V-O-5u:-ONKEKBV5kwxr2R0kxiT:b-26u8g0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f1d0319c62dc3f975bbfabff4bd7a59d30517cc5/ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift#L115)):*

Can't we wrap this function in another object to make it reusable across all Mullvad API proxies?
</blockquote></details>

I don't think there's much value in that. The only difference from this function and creating a new `MullvadApiNetworkOperation` object is that this function adds it to the queueu.
___
*[`ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift` line 138 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-ONJm2yP-GYGtbga_M__:-ONK86gw96lnD2nyHwIG:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f1d0319c62dc3f975bbfabff4bd7a59d30517cc5/ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift#L138)):*

I know it was here before, but I think this piece of code should be moved to `MullvadMockData`, since we're importing `MullvadMockData` in the tests that use this function.
</blockquote></details>

Done.
___
*[`ios/MullvadRustRuntime/include/mullvad_rust_runtime.h` line 155 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-ONJnl-CDeOL442lRzth:-ONJuUS5-d-e5r2NnXF6:b-p0067p) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f1d0319c62dc3f975bbfabff4bd7a59d30517cc5/ios/MullvadRustRuntime/include/mullvad_rust_runtime.h#L155)):*

I don't see this comment convention in the rust side. if it's agreed to start multiline comment with /**/ then we need to replace it with `//` which is spread in `api-client`.
</blockquote></details>

This header file is automaticallt generated, so we can disregard the commenting style here.

MarkusPettersson98 (dismissed):
:lgtm:

> Reviewed 3 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-:-ONKItkyAOErUrD7XTo9:bnfp4nl)* status: 18 of 27 files reviewed, 5 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @mojganii)

___
*[`mullvad-ios/src/api_client/account.rs` line 44 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-OMzqxLoCR-mh4bQCshJ:-ONKIXyoDf6I8j0vhHtC:b-80tlqx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/062f8ef8955bf18a1fe21761402bf4b51416c9f9/mullvad-ios/src/api_client/account.rs#L44)):*

Can I just point to param doc for that?
</blockquote></details>

Sure!

mojganii (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946)* status: 16 of 27 files reviewed, 2 unresolved discussions (waiting on @acb-mv, @buggmagnet, @MarkusPettersson98, and @rablador)

___
*[`ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift` line 115 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946#-ONJktrN-Ampf2V-O-5u:-ONKheMx5elora6P1sBp:bkxjl69) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f1d0319c62dc3f975bbfabff4bd7a59d30517cc5/ios/MullvadREST/MullvadAPI/APIHandlers/MullvadAccountProxy.swift#L115)):*

I don't think there's much value in that. The only difference from this function and creating a new `MullvadApiNetworkOperation` object is that this function adds it to the queueu.
</blockquote></details>

discussed offline.

MarkusPettersson98 (commented):
> Reviewed 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7946)* status: 18 of 27 files reviewed, 2 unresolved discussions (waiting on @acb-mv, @buggmagnet, and @mojganii)

PR #7960, opened 2025-04-04
Small cleanups in desktop release scripts

faern:
I think this PR is easiest reviewed one commit at a time.

1. Removes the check for the `gh` command in a script that (no longer?) uses it.
2. The `ARTIFACT_DIR` variable was defined in three scripts and they depended on having the same value, but this was not obvious at all. I moved the definition into a config file that we source, and then use that single definition.
3. Fix bug in `5-update-and-publish-metadata` where the rollout percentage was passed on an invalid format to `mullvad-release`
4. Use `xclip -sensitive` to read the ed25519 signing key from the clipboard, instead of expecting the person running the script from manually pasting it. Removes a manual step and helps avoid accidentally pasting the key in the wrong place.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7960)

### DIFF
--- a/desktop/scripts/release/4-make-release
+++ b/desktop/scripts/release/4-make-release
@@ -27,10 +27,10 @@ fi
 
 PRODUCT_VERSION=$1
 
-ARTIFACT_DIR="./artifacts"
+# shellcheck source=desktop/scripts/release/release-config.sh
+source "$SCRIPT_DIR/release-config.sh"
 
-rm -rf $ARTIFACT_DIR
-mkdir -p $ARTIFACT_DIR
+rm -rf "$ARTIFACT_DIR" && mkdir -p "$ARTIFACT_DIR" || exit 1
 
 function publish_release {
     echo ">>> Downloading changelog"
@@ -85,5 +85,5 @@ function publish_release {
     echo "The above URL contains the text \"untagged\", but don't worry it is tagged properly and everything will look correct once it's published."
 }
 
-./download-release-artifacts "$PRODUCT_VERSION"
+./download-release-artifacts "$PRODUCT_VERSION" "$ARTIFACT_DIR"
 publish_release
--- a/desktop/scripts/release/5-update-and-publish-metadata
+++ b/desktop/scripts/release/5-update-and-publish-metadata
@@ -19,16 +19,6 @@ if [ $# -ne 3 ]; then
     exit 1
 fi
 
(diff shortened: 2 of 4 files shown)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7960)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[`desktop/scripts/release/download-release-artifacts` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7960#-ON-WLPf0W3Hswr81_xq:-ON-WLPf0W3Hswr81_xr:bqv4l3r) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/fdb6c648770e680a9045efe2cde60be98175a83d/desktop/scripts/release/download-release-artifacts#L20)):*
> ```
> PRODUCT_VERSION=$1
> # The directory where the artifacts will be downloaded to
> ARTIFACT_DIR=$2
> ```

In these "helper scripts" I prefer to pass in as much config as possible instead of just reading it directly from the config file. The reason for this is that when reading `4-make-release` and `5-update-and-publish-metadata` it becomes easier to see that this script depend on that variable. Because those other scripts delete and create the artifact dir, so it's a shared dependency. And this way makes that more obvious.

dlon (dismissed):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7960#-:-ON-mX3w2b8NQ17c8yHh:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7960)* status: all files reviewed, 1 unresolved discussion

___
*[`desktop/scripts/release/release-config.sh` line 6 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7960#-ON1V0tT1TSfLZw8jRID:-ON1V0tT1TSfLZw8jRIE:b-h1iwos) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/800841dac5d329cc92874cb45563481b29011395/desktop/scripts/release/release-config.sh#L6)):*
> ```Shell
> 
> # Where to download app installers locally during the release process.
> # This value is also hardcoded into the `mullvad-release` binary and
> ```

This is a good point :sweat_smile:

PR #7962, opened 2025-04-04
Bump the version of installer downloader to 0.2.0

faern:
This is for internal testing and dogfooding. First version pointing to and using production keys and infrastructure. The intention is to bump the version to `1.0.0` before we release to actual users. But we'll evaluate that after seeing how well this version works.

Is everything ready for building a "release" or other stuff we need to merge first?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2022,7 +2022,7 @@ dependencies = [
 
 [[package]]
 name = "installer-downloader"
-version = "0.1.0"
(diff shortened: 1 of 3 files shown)


dlon (approved):

dlon:
Found a silly issue in #7965. Please wait for that to be merged first

dlon (approved):

PR #7963, opened 2025-04-04
Update ProblemReport view to integrate better with in app upgrade feature

tobias-jarvelov:
- Makes the modal optional so it can be disabled when navigating to it from the AppUpgrade view.
- Conditionally sets the "Upgrade app" button to navigate to an external URL or the AppUpgrade route, depending on platform.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7963)

### DIFF
--- a/desktop/packages/mullvad-vpn/src/renderer/components/ProblemReport.tsx
+++ b/desktop/packages/mullvad-vpn/src/renderer/components/ProblemReport.tsx
@@ -324,7 +324,16 @@ function OutdatedVersionWarningDialog() {
   const outdatedVersion = useSelector((state) => !!state.version.suggestedUpgrade);
   const pushAppUpgrade = usePushAppUpgrade();
 
-  const [showOutdatedVersionWarning, setShowOutdatedVersionWarning] = useState(outdatedVersion);
+  const { location } = useHistory();
+  const { state } = location;
+  const hasSuppressOutdatedVersionWarning = state?.options?.includes(
+    'suppress-outdated-version-warning',
+  );
+  const showOutdatedVersionWarningInitial = outdatedVersion && !hasSuppressOutdatedVersionWarning;
+
+  const [showOutdatedVersionWarning, setShowOutdatedVersionWarning] = useState(
+    showOutdatedVersionWarningInitial,
+  );
 
   const acknowledgeOutdatedVersion = useCallback(() => {
     setShowOutdatedVersionWarning(false);
--- a/desktop/packages/mullvad-vpn/src/renderer/components/ProblemReport.tsx
+++ b/desktop/packages/mullvad-vpn/src/renderer/components/ProblemReport.tsx
@@ -15,6 +15,8 @@ import {
 import { messages } from '../../shared/gettext';
 import { getDownloadUrl } from '../../shared/version';
 import { useAppContext } from '../context';
+import { usePushAppUpgrade } from '../history/hooks';
+import { useIsPlatformLinux } from '../hooks';
 import useActions from '../lib/actionsHook';
 import { Flex, Icon, Spinner } from '../lib/components';
 import { useHistory } from '../lib/history';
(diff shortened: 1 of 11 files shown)


desktop/packages/mullvad-vpn/src/renderer/components/ProblemReport.tsx
```diff
@@ -320,8 +323,16 @@ function OutdatedVersionWarningDialog() {
   const isOffline = useSelector((state) => state.connection.isBlocked);
   const suggestedIsBeta = useSelector((state) => state.version.suggestedIsBeta ?? false);
   const outdatedVersion = useSelector((state) => !!state.version.suggestedUpgrade);
+  const pushAppUpgrade = usePushAppUpgrade();
 
-  const [showOutdatedVersionWarning, setShowOutdatedVersionWarning] = useState(outdatedVersion);
+  const location = useLocation();
+  const query = new URLSearchParams(location.search);
+  const hasSuppressOutdatedVersionWarning = query.has('suppress-outdated-version-warning');
```
olmoh, line 330: I probably should have commented on this in an earlier PR but didn't think of it until now. Instead of using an url parameter we could make use of the state object of react router.

```ts
import { useCallback } from 'react';

import { useHistory } from '../../lib/history';
import { RoutePath } from '../../lib/routes';

export type PushProblemReportProps = {
  search?: string;
};

export const suppressOutdatedVersionWarning = 'suppress-outdated-version-warning';

export const usePushProblemReport = ({ search }: PushProblemReportProps = {}) => {
  const history = useHistory();

  const pushProblemReport = useCallback(() => {
    history.push(
      {
        pathname: RoutePath.problemReport,
        search,
      },
      {
        options: [suppressOutdatedVersionWarning],
      },
    );
  }, [history, search]);

  return pushProblemReport;
};
```

And then we can check for the option inside of the state.
```ts
const { state } = useLocation();
```

I don't think we use URL parameters in any other cases so this would be more similar to have navigation has been done previously.

olmoh (changes requested):

desktop/packages/mullvad-vpn/src/renderer/components/ProblemReport.tsx
```diff
@@ -11,10 +11,14 @@ import {
   useRef,
   useState,
 } from 'react';
+import { useLocation } from 'react-router';
```
olmoh, line 14: We can use the from lib/history instead.

desktop/packages/mullvad-vpn/src/renderer/components/ProblemReport.tsx
```diff
@@ -320,8 +324,17 @@ function OutdatedVersionWarningDialog() {
   const isOffline = useSelector((state) => state.connection.isBlocked);
   const suggestedIsBeta = useSelector((state) => state.version.suggestedIsBeta ?? false);
   const outdatedVersion = useSelector((state) => !!state.version.suggestedUpgrade);
+  const pushAppUpgrade = usePushAppUpgrade();
 
-  const [showOutdatedVersionWarning, setShowOutdatedVersionWarning] = useState(outdatedVersion);
+  const { state } = useLocation<LocationState>();
+  const hasSuppressOutdatedVersionWarning = state?.options?.includes(
```
olmoh, line 330: This currently does not work as intended, and the dialog will be shown every time, since options won't be set in the state.
If we add it to `createState` in `history.tsx` it should work as expected.
```
  private createState(state?: Partial<LocationState>): LocationState {
    return {
      scrollPosition: state?.scrollPosition ?? [0, 0],
      expandedSections: state?.expandedSections ?? {},
      transition: state?.transition ?? transitions.none,
      options: state?.options,
    };
  }
  ```

olmoh (changes requested):

desktop/packages/mullvad-vpn/src/renderer/components/ProblemReport.tsx
tobias-jarvelov: I'm guessing you meant `useHistory` hook from `'lib/history'`, at least that's what I went with.

desktop/packages/mullvad-vpn/src/renderer/components/ProblemReport.tsx
tobias-jarvelov: Great, thanks. I've updated it now.

olmoh (approved):

PR #7986, opened 2025-04-09
Expose transport selector to mullvad api ios 1042

buggmagnet:
This PR enables using the AccessModeSelector powered by Rust with a couple of caveats:
- Swift still decides when a method is rotated (we will make the switch when we are done converting all the API calls to the Rust client)
- Rust currently has no way to indicate to Swift that it will rotate access methods
- The code assumes that pointer conversion between Swift and Rust never fails, the app wouldn't work otherwise, this was the same assumption that was built in when `MullvadApiContext` was introduced, and this expands on that concept. We can make it safer however.

⚠️ The PR is not mergeable yet because of recent additions to the same area in the codebase, but since theres a lot to review, I decided to open it now, and will deal with the merge conflicts as they come ⚠️ 

## How to test this ?
- Apply the patch at attached at the end of this description
- Make sure that the APIProxy is proxying to call to the rust client (Should already be the case if you are running a DEBUG build)
- Launch the app, change the currently selected API access method
- You should see which access method is selected in the console logs

```patch
diff --git a/ios/MullvadREST/Transport/TransportStrategy.swift b/ios/MullvadREST/Transport/TransportStrategy.swift
index 53c22b36c9..e4eb8960c2 100644
--- a/ios/MullvadREST/Transport/TransportStrategy.swift
+++ b/ios/MullvadREST/Transport/TransportStrategy.swift
@@ -64,6 +64,7 @@ public struct TransportStrategy: Equatable, Sendable {
             provider: accessMethodIterator,
             domainName: REST.encryptedDNSHostname
         )
+        // TODO: Can we use the Access Method database here instead of the access method iterator ?
         self
             .opaqueAccessMethodSettingsWrapper = initAccessMethodSettingsWrapper(
                 methods: connectionModeProviderProxy
diff --git a/mullvad-api/src/access_mode.rs b/mullvad-api/src/access_mode.rs
index ecd90b9d82..894685721e 100644
--- a/mullvad-api/src/access_mode.rs
+++ b/mullvad-api/src/access_mode.rs
@@ -267,6 +267,12 @@ impl<B: AccessMethodResolver + 'static> AccessModeSelector<B> {
         let (index, next) = Self::find_next_active(0, &access_method_settings);
         let initial_connection_mode = Self::resolve_with_default(&next, &mut method_resolver).await;
 
+        let log = format!(
+            "resolved connnection mode : {}",
+            initial_connection_mode.connection_mode
+        );
+        log::error!("{:?}", log);
+
         let (change_tx, change_rx) = mpsc::unbounded();
 
         let api_connection_mode = initial_connection_mode.connection_mode.clone();
diff --git a/mullvad-api/src/proxy.rs b/mullvad-api/src/proxy.rs
index 106449bb30..0010b4137d 100644
--- a/mullvad-api/src/proxy.rs
+++ b/mullvad-api/src/proxy.rs
@@ -1,6 +1,7 @@
 use hyper_util::client::legacy::connect::{Connected, Connection};
 use serde::{Deserialize, Serialize};
 use std::{
+    fmt::{self, Display, Formatter},
     io,
     net::SocketAddr,
     path::Path,
@@ -61,6 +62,19 @@ pub enum ApiConnectionMode {
     Proxied(ProxyConfig),
 }
 
+impl Display for ApiConnectionMode {
+    fn fmt(&self, formatter: &mut Formatter<'_>) -> fmt::Result {
+        let result = match self {
+            Self::Direct => "Direct mode".to_owned(),
+            Self::Proxied(proxy) => {
+                let address = proxy.get_endpoint().address;
+                format!("proxied via {:?}", address).to_owned()
+            }
+        };
+        writeln!(formatter, "{}", result.as_str())
+    }
+}
+
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub enum ProxyConfig {
     Shadowsocks(proxy::Shadowsocks),
diff --git a/mullvad-api/src/rest.rs b/mullvad-api/src/rest.rs
index 94c43392ad..1fdd02167d 100644
--- a/mullvad-api/src/rest.rs
+++ b/mullvad-api/src/rest.rs
@@ -202,6 +202,8 @@ impl<T: ConnectionModeProvider + 'static> RequestService<T> {
                     let Some(new_mode) = new_mode else {
                         break;
                     };
+                    let log = format!("new mode: {}", new_mode);
+                    log::error!("{:?}", log);
                     self.connector_handle.set_connection_mode(new_mode);
                 }
                 command = self.command_rx.next() => {
diff --git a/mullvad-ios/src/api_client/access_method_settings.rs b/mullvad-ios/src/api_client/access_method_settings.rs
index e87fce8759..14e449445f 100644
--- a/mullvad-ios/src/api_client/access_method_settings.rs
+++ b/mullvad-ios/src/api_client/access_method_settings.rs
@@ -152,6 +152,7 @@ impl SwiftAccessMethodSettingsWrapper {
     }
 
     pub unsafe fn into_rust_context(self) -> Box<SwiftAccessMethodSettingsContext> {
+        //TODO: Should we increment self here so that we can reuse this ?
         Box::from_raw(self.0)
     }
 }
diff --git a/mullvad-ios/src/api_client/api.rs b/mullvad-ios/src/api_client/api.rs
index c918dda61f..32f9d50915 100644
--- a/mullvad-ios/src/api_client/api.rs
+++ b/mullvad-ios/src/api_client/api.rs
@@ -39,6 +39,17 @@ pub unsafe extern "C" fn mullvad_api_get_addresses(
 
     let api_context = api_context.into_rust_context();
     let retry_strategy = unsafe { retry_strategy.into_rust() };
+    let api_context_clone = api_context.clone();
+    tokio_handle.clone().spawn(async move {
+        let current_connection_mode_future = api_context_clone.access_mode_handler.get_current();
+        match current_connection_mode_future.await {
+            Ok(connection_mode) => {
+                let log = format!("new mode: {}", connection_mode.connection_mode);
+                log::error!("{:?}", log);
+            }
+            Err(error) => log::error!("{:?}", error),
+        }
+    });
 
     let completion = completion_handler.clone();
     let task = tokio_handle.clone().spawn(async move {
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)


rablador (changes requested):
> Reviewed 48 of 48 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 7 unresolved discussions

___
*[`ios/MullvadVPN/Coordinators/Settings/APIAccess/Add/AddAccessMethodCoordinator.swift` line 11 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONURErB-fJaElQ7zH6H:-ONURErB-fJaElQ7zH6I:b-upe3gd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ffc5a016634e6bd996fbbd084e14448d79dc4/ios/MullvadVPN/Coordinators/Settings/APIAccess/Add/AddAccessMethodCoordinator.swift#L11)):*
> ```Swift
> import Combine
> import MullvadSettings
> import MullvadTypes
> ```

Remove import?
___
*[`ios/MullvadSettings/AccessMethodRepositoryProtocol.swift` line 10 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONUNulS5TKw4kXDCG8o:-ONUNulTAULXUHCaO-oC:b-upe3gd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ffc5a016634e6bd996fbbd084e14448d79dc4/ios/MullvadSettings/AccessMethodRepositoryProtocol.swift#L10)):*
> ```Swift
> 
> import Combine
> import MullvadTypes
> ```

Remove import?
___
*[`ios/MullvadRustRuntime/MullvadApiContext.swift` line 36 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONUF0qU862WOPBYGc8s:-ONUF0qU862WOPBYGc8t:bmyo12l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ffc5a016634e6bd996fbbd084e14448d79dc4/ios/MullvadRustRuntime/MullvadApiContext.swift#L36)):*
> ```Swift
> 
>         if context._0 == nil {
>             throw NSError(domain: "", code: 0)
> ```

What's wrong with `MullvadApiContextError.failedToConstructApiClient`?
___
*[`ios/build-rust-library.sh` line 56 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONU9rRjDCIRr7qk7Whd:-ONU9rRjDCIRr7qk7Whe:bzbwo7m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ffc5a016634e6bd996fbbd084e14448d79dc4/ios/build-rust-library.sh#L56)):*
> ```Shell
>     arm64)
>       if [ $IS_SIMULATOR -eq 0 ]; then
>         # Hardware iOS targets      
> ```

Spaces
___
*[`ios/MullvadRustRuntime/MullvadApiCompletion.swift` line 9 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONUCPMe0fUYjEGED14C:-ONUCPMe0fUYjEGED14D:b-upe3gd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ffc5a016634e6bd996fbbd084e14448d79dc4/ios/MullvadRustRuntime/MullvadApiCompletion.swift#L9)):*
> ```Swift
> //
> 
> import MullvadTypes
> ```

Remove import?
___
*[`ios/MullvadRustRuntime/MullvadConnectionModeProvider.swift` line 20 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONUGAa24O5kBesVNwp1:-ONUGAa24O5kBesVNwp2:b-ixi2q2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ffc5a016634e6bd996fbbd084e14448d79dc4/ios/MullvadRustRuntime/MullvadConnectionModeProvider.swift#L20)):*
> ```Swift
> 
>     // 2. Get the custom access methods
>     let filter: [PersistentProxyConfiguration] = [.direct, .bridges, .encryptedDNS]
> ```

Nit: "defaultMethods"?
___
*[`ios/MullvadRustRuntime/MullvadConnectionModeProvider.swift` line 46 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONUHHdyBFTJeFhBlWcr:-ONUHHdyBFTJeFhBlWcs:b-irbr03) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ffc5a016634e6bd996fbbd084e14448d79dc4/ios/MullvadRustRuntime/MullvadConnectionModeProvider.swift#L46)):*
> ```Swift
>     )
> 
>     // 4. Convert the custom access methods (all takes different parameters)
> ```

Any specific reason we don't first construct these raws on swift side and then add then in one swoop on rust side?

dlon (commented):
> Reviewed 1 of 48 files at r1, 1 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: 44 of 48 files reviewed, 7 unresolved discussions (waiting on @rablador)

___
*[`mullvad-daemon/src/lib.rs` line 735 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONVVRwu7YZYBAuzcN0p:-ONVVRwu7YZYBAuzcN0q:bfhu2kc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/149ba77ac1e42e187f37068e9ebe82e6b8b0ceed/mullvad-daemon/src/lib.rs#L735)):*
> ```XML
>                 config.endpoint.clone(),
>                 #[cfg(not(target_os = "ios"))]
>                 internal_event_tx.to_unbounded_sender(),
> ```

Nit: This crate isn't build on iOS. I think the `cfg` is unnecessary

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: 42 of 48 files reviewed, 7 unresolved discussions (waiting on @dlon and @rablador)

___
*[`mullvad-daemon/src/lib.rs` line 735 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONVVRwu7YZYBAuzcN0p:-ONYi0Zd-LqTTwj_6inu:bgusj6h) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/149ba77ac1e42e187f37068e9ebe82e6b8b0ceed/mullvad-daemon/src/lib.rs#L735)):*

Nit: This crate isn't build on iOS. I think the `cfg` is unnecessary
</blockquote></details>

you're right, my IDE was complaining about it missing arguments because I've configured my LSP to only compile for the iOS targets :see_no_evil:

rablador (commented):
> Reviewed 4 of 4 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 2 unresolved discussions (waiting on @buggmagnet and @dlon)

___
*[`ios/MullvadRustRuntime/MullvadConnectionModeProvider.swift` line 46 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONUHHdyBFTJeFhBlWcr:-ONZvgch6kPpF-ySJmQe:b1wgtmi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ffc5a016634e6bd996fbbd084e14448d79dc4/ios/MullvadRustRuntime/MullvadConnectionModeProvider.swift#L46)):*

I'm not sure what you mean by that, could you clarify ?
</blockquote></details>

We create one raw method at a time, then add through rust FFI. Can't we just create all methods first, then send the whole package once?

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon and @rablador)

___
*[`ios/MullvadRustRuntime/MullvadConnectionModeProvider.swift` line 46 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONUHHdyBFTJeFhBlWcr:-ONn3cZ3AZ3cV5MIO5nT:b-2dx6da) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ffc5a016634e6bd996fbbd084e14448d79dc4/ios/MullvadRustRuntime/MullvadConnectionModeProvider.swift#L46)):*

We create one raw method at a time, then add through rust FFI. Can't we just create all methods first, then send the whole package once?
</blockquote></details>

This was discussed offline, and it was decided to leave it as is.

MarkusPettersson98 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 13 unresolved discussions (waiting on @buggmagnet and @dlon)

___
*[`mullvad-ios/src/api_client/shadowsocks_loader.rs` line 18 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONojD8f8vhhnufku1R-:-ONojD8gEvKarT4V47w1:b-tpw3b0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/shadowsocks_loader.rs#L18)):*
> ```XML
>     pub fn new(context: SwiftShadowsocksLoaderWrapperContext) -> SwiftShadowsocksLoaderWrapper {
>         SwiftShadowsocksLoaderWrapper(Box::into_raw(Box::new(context)))
>     }
> ```

Allocating a `Box` and immediately giving up ownership of it using `Box::into_raw` seems a bit strange to me. Is there a reason why you don't keep the `Box`?



_Code quote:_
```XML
    pub fn new(context: SwiftShadowsocksLoaderWrapperContext) -> SwiftShadowsocksLoaderWrapper {
        SwiftShadowsocksLoaderWrapper(Box::into_raw(Box::new(context)))
    }
```

MarkusPettersson98 (changes requested):
> Reviewed 1 of 48 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 16 unresolved discussions (waiting on @buggmagnet and @dlon)

___
*[`mullvad-ios/src/api_client/access_method_resolver.rs` line 79 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONolr4-4L1n4Df3mdH0:-ONolr4-4L1n4Df3mdH1:b61nfa2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/access_method_resolver.rs#L79)):*
> ```XML
>             },
>             connection_mode,
>         ))
> ```

Purely stylistically: introducing a couple of `let`-bindings could help making this code less dense :blush:

```rust
let allowed_endpoint = {
    let endpoint = connection_mode.get_endpoint().unwrap_or_else(|| {
        Endpoint::from_socket_address(
            self.endpoint.address.unwrap(),
            TransportProtocol::Tcp,
        )
    });
    let clients = AllowedClients::All;
    AllowedEndpoint { endpoint, clients }
};

Some((allowed_endpoint, connection_mode))
```



_Code quote:_
```XML
        Some((
            AllowedEndpoint {
                endpoint: match connection_mode.get_endpoint() {
                    Some(endpoint) => endpoint,
                    None => Endpoint::from_socket_address(
                        self.endpoint.address.unwrap(),
                        TransportProtocol::Tcp,
                    ),
                },
                clients: AllowedClients::All,
            },
            connection_mode,
        ))
```
___
*[`mullvad-ios/src/api_client/access_method_resolver.rs` line 83 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONomO044L1xOHp1Ub86:-ONomO044L1xOHp1Ub87:bb7462z) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/access_method_resolver.rs#L83)):*
> ```XML
> 
>     async fn default_connection_mode(&self) -> AllowedEndpoint {
>         // TODO: Call the iOS Address cache implementation instead of returning the default endpoint
> ```

TODO



_Code quote:_
```XML
        // TODO: Call the iOS Address cache implementation instead of returning the default endpoint
```
___
*[`mullvad-ios/src/api_client/access_method_resolver.rs` line 88 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONomRTf9ldpSnSl4G5a:-ONomRTf9ldpSnSl4G5b:bli9ol3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/access_method_resolver.rs#L88)):*
> ```XML
>             endpoint,
>             clients: AllowedClients::All,
>         };
> ```

:pick: `return` statement is superflous here
```rust
async fn default_connection_mode(&self) -> AllowedEndpoint {
    ..
    AllowedEndpoint {
        endpoint,
        clients: AllowedClients::All,
    }
}
```



_Code quote:_
```XML
        return AllowedEndpoint {
            endpoint,
            clients: AllowedClients::All,
        };
```

MarkusPettersson98 (changes requested):
> Reviewed 1 of 48 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 19 unresolved discussions (waiting on @buggmagnet)

___
*[`mullvad-ios/src/api_client/helpers.rs` line 16 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONonhnw5OHyQJHv0ueE:-ONonhnw5OHyQJHv0ueF:b-ndr51t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/helpers.rs#L16)):*
> ```XML
>         4 => {
>             // SAFETY: addr pointer must point to at least addr_len bytes
>             let bytes = unsafe { std::slice::from_raw_parts(addr, addr_len) };
> ```

`std::slice::from_raw_parts` actually *require* the pointer to be non-null and aligned in addition to the length argument making sense: https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety



_Code quote:_
```XML
std::slice::from_raw_parts(addr, addr_len)
```
___
*[`mullvad-ios/src/api_client/helpers.rs` line 160 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONopbefDI5__vjQu_ZA:-ONopbefDI5__vjQu_ZB:bhru5ty) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/helpers.rs#L160)):*
> ```XML
>             Err(_) => None,
>         }
>     }
> ```

:pick: `auth` does not need to be `mut` if we introduce a `block`-expression!

```rust
let auth = {
    if c_username.is_null() || c_password.is_null() {
        None
    } else {
        let username = convert_c_string(c_username);
        let password = convert_c_string(c_password);
        SocksAuth::new(username, password).ok()
    }
};
```



_Code quote:_
```XML
    let mut auth: Option<SocksAuth> = None;
    if !c_username.is_null() && !c_password.is_null() {
        let username = convert_c_string(c_username);
        let password = convert_c_string(c_password);
        auth = match SocksAuth::new(username, password) {
            Ok(auth) => Some(auth),
            Err(_) => None,
        }
    }
```

dlon (commented):
> Reviewed 1 of 48 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 23 unresolved discussions (waiting on @buggmagnet and @MarkusPettersson98)

___
*[`mullvad-ios/src/api_client/helpers.rs` line 98 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONorxQ04a_Qf8L1u3wg:-ONorxQ04a_Qf8L1u3wh:b-qpysh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/helpers.rs#L98)):*
> ```XML
> 
> impl RustAccessMethodSettingVectorContext {
>     pub unsafe fn push(&mut self, setting: AccessMethodSetting) {
> ```

This is not (or should not be) unsafe
___
*[`mullvad-ios/src/api_client/helpers.rs` line 137 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONoqupdBmw7LfPOYn4O:-ONoqupdBmw7LfPOYn4P:b-syl541) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/helpers.rs#L137)):*
> ```XML
> /// # SAFETY
> /// `address` must be a pointer to at least `address_len` bytes.
> /// `c_username` and `c_password` must be a pointer to null terminated strings
> ```

Or null
___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 82 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONosjBRETC96EU2Ribc:-ONosjBRETC96EU2Ribd:b-49kam2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/access_method_settings.rs#L82)):*
> ```XML
>                 // SAFETY: See `convert_builtin_access_method_setting`
>                 let configuration: Shadowsocks =
>                     unsafe { *Box::from_raw(proxy_configuration as *mut _) };
> ```

Maybe just me, but it feels rather non-obvious which arguments are consumed/freed vs. only read.
___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 129 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONorCLq-oRtIgQ_l9vB:-ONorCLq-oRtIgQ_l9vC:bb8e2sn) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/access_method_settings.rs#L129)):*
> ```XML
>     bridges_method_raw: *const c_void,
>     encrypted_dns_method_raw: *const c_void,
>     custom_methods_raw: RustAccessMethodSettingVector,
> ```

Could we not use a C array here?

```
custom_methods_num: usize,
custom_methods: *const MyAccessMethodType,
```

I guess the reason is that `AccessMethodSetting` isn't `repr(C)`?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 24 unresolved discussions (waiting on @buggmagnet and @MarkusPettersson98)

___
*[`mullvad-ios/src/api_client/mod.rs` line 76 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONoz5S7BI72ltUshG7q:-ONoz5S7BI72ltUshG7r:b-h8a744) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/mod.rs#L76)):*
> ```XML
>     let access_methods = settings_wrapper.into_rust_context().settings;
>     api_context
>         .into_rust_context()
> ```

This naming is inconsistent. Here, `into_rust_context` doesn't actually consume/free `SwiftApiContext`/`ApiContext`. I think you should name it `as_rust_context` to convey that, at least.

Btw, it seems like `ApiContext` is never freed at all.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 18 unresolved discussions (waiting on @dlon and @MarkusPettersson98)

___
*[`mullvad-ios/src/encrypted_dns_proxy.rs` line 22 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONoPq672mbMQVij5DXx:-ONs6uWU-dmTO8jehcqv:b1ej1gg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/encrypted_dns_proxy.rs#L22)):*

Do these fields need to be `pub`? Can `EncryptedDnsProxyState` expose the required methods of `State` as needed? :blush:
</blockquote></details>

Apparently, I did this temporarily, and refactored into not needing it anymore. 
___
*[`mullvad-ios/src/api_client/access_method_resolver.rs` line 79 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONolr4-4L1n4Df3mdH0:-ONs8Du28mGeHKNcF9kT:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/access_method_resolver.rs#L79)):*

Purely stylistically: introducing a couple of `let`-bindings could help making this code less dense :blush:

```rust
let allowed_endpoint = {
    let endpoint = connection_mode.get_endpoint().unwrap_or_else(|| {
        Endpoint::from_socket_address(
            self.endpoint.address.unwrap(),
            TransportProtocol::Tcp,
        )
    });
    let clients = AllowedClients::All;
    AllowedEndpoint { endpoint, clients }
};

Some((allowed_endpoint, connection_mode))
```
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/access_method_resolver.rs` line 83 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONomO044L1xOHp1Ub86:-ONs8FbBDfj4z8h44QpI:b-wm0yt3) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/access_method_resolver.rs#L83)):*

TODO
</blockquote></details>

That's intentional. Since this PR was getting very large, we decided to do that work in an upcoming PR.
___
*[`mullvad-ios/src/api_client/access_method_resolver.rs` line 88 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONomRTf9ldpSnSl4G5a:-ONs8XWy3PYmctcbZG6e:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/access_method_resolver.rs#L88)):*

:pick: `return` statement is superflous here
```rust
async fn default_connection_mode(&self) -> AllowedEndpoint {
    ..
    AllowedEndpoint {
        endpoint,
        clients: AllowedClients::All,
    }
}
```
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 129 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONorCLq-oRtIgQ_l9vB:-ONs8jNK1mZxchgDy07t:b-40iurh) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/access_method_settings.rs#L129)):*

Could we not use a C array here?

```
custom_methods_num: usize,
custom_methods: *const MyAccessMethodType,
```

I guess the reason is that `AccessMethodSetting` isn't `repr(C)`?
</blockquote></details>

We had a discussion about this, I had this original prototype.
```rust
   let custom_methods: Vec<AccessMethodSetting> = match custom_methods.is_null() {
        true => vec![],
        false => {
            let custom_methods = unsafe { from_raw_parts(custom_methods, custom_methods_count) };
            let mut methods_vec: Vec<AccessMethodSetting> =
                Vec::with_capacity(custom_methods_count);
            for index in custom_methods {
                let accessMethodRaw = custom_methods[index]; // :x: the type [c_void] cannot be indexed by &c_void
                let accessMethod = unsafe { *Box::from_raw(accessMethodRaw as *mut _) };
                methods_vec.append(accessMethod);
            }
            methods_vec
        }
    };
```

I think we could do this if we re-define all the structures we use for access methods at the FFI layer (or make ours `#[repr(C)]`) but I didn't want to make that jump and have the same structures defined multiple times, so I opted for this instead.

What do you think ? We could pair up on this if you want ?
___
*[`mullvad-ios/src/api_client/helpers.rs` line 16 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONonhnw5OHyQJHv0ueE:-ONsA1o_74wqBM20FTz4:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/helpers.rs#L16)):*

`std::slice::from_raw_parts` actually *require* the pointer to be non-null and aligned in addition to the length argument making sense: https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/helpers.rs` line 98 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONorxQ04a_Qf8L1u3wg:-ONsDeUtCyHnG-k00FRx:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/helpers.rs#L98)):*

This is not (or should not be) unsafe
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/helpers.rs` line 137 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONoqupdBmw7LfPOYn4O:-ONsDqWTEvFfnby3ZHFQ:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/helpers.rs#L137)):*

Or null
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/helpers.rs` line 160 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONopbefDI5__vjQu_ZA:-ONsE7OG-Ujtmc8qkwyQ:b-lmu3zj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/helpers.rs#L160)):*

:pick: `auth` does not need to be `mut` if we introduce a `block`-expression!

```rust
let auth = {
    if c_username.is_null() || c_password.is_null() {
        None
    } else {
        let username = convert_c_string(c_username);
        let password = convert_c_string(c_password);
        SocksAuth::new(username, password).ok()
    }
};
```
</blockquote></details>

Done 
nice ! 
___
*[`mullvad-ios/src/api_client/mod.rs` line 65 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONoQXmHCQqWhT3gjQZy:-ONsEp2XDgDgK-heIpja:b-871wl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/mod.rs#L65)):*

Should these functions mention that they will block _all_ execution on the current thread? :blush:
</blockquote></details>

I thought about it, and whilst I think ultimately it won't really matter since those calls are not doing anything computationally intensive (i.e. they won't block for that long), it can't hurt to mention that, so yes let's mention this !
___
*[`mullvad-ios/src/api_client/mod.rs` line 76 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONoz5S7BI72ltUshG7q:-ONsJ_hj9-jHGTIpk3KP:biz3oo6) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/mod.rs#L76)):*
`as_rust_context` generates `clippy` warnings because `methods called `as_*` usually take `self` by reference or `self` by mutable reference`

> Btw, it seems like `ApiContext` is never freed at all.
Yes, it's intended to live as long as the process it's running in, so we never need to free it.

We discussed this offline and settled for `rust_context` simply
___
*[`mullvad-ios/src/api_client/mod.rs` line 90 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONoSQHU6ZR6-s0xMCNc:-ONsRHlS-qdrnNDVe0iN:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/mod.rs#L90)):*

`unsafe` without `// SAFETY` comment
</blockquote></details>

Done.
___
*[`mullvad-ios/src/api_client/mod.rs` line 93 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONoSDBH14xXli28ylo5:-ONsRZRM8rv328qrNih1:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/mod.rs#L93)):*

:pick: This early-return could be simplified slightly using a `let-else` expression:

```rust
let Some(id) = Id::from_string(unsafe { convert_c_string(access_method_id) }) else {
    return;
};
```
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/mod.rs` line 95 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONoSWV25QHHF2s7jsIw:-ONsR_WS4_eWei-P5JsE:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/mod.rs#L95)):*

`unsafe` without `// SAFETY` comment
</blockquote></details>

Done
___
*[`mullvad-ios/src/api_client/mod.rs` line 95 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONoSbI_C1_j02bNKByK:-ONsSNj36AfP-8QzneRu:b-vm412w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/mod.rs#L95)):*

Consider untangling the `unsafe` block to make them more visible + easier to document. One suggestion is to introduce some more bindings in the function scope:

```rust
pub unsafe extern "C" fn mullvad_api_use_access_method(
    api_context: SwiftApiContext,
    access_method_id: *const c_char,
) {
    // SAFETY: TODO
    let api_context = unsafe { api_context.into_rust_context() };
    // SAFETY: TODO
    let cstr = unsafe { convert_c_string(access_method_id) };

    // These parts are completely safe rust
    let Some(id) = Id::from_string(cstr) else { return };
    api_context.use_access_method(id);
}
```
This would also "fix" the last `unsafe` block scope being overly broad :blush:
</blockquote></details>

Nice, thanks for that suggestion !

Done
___
*[`mullvad-ios/src/api_client/mod.rs` line 120 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONodkJjCRKyAV5DU8et:-ONsSRk2BDHk7-VF1fG2:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/mod.rs#L120)):*

You know the drill
</blockquote></details>

Done.
___
*[`ios/build-rust-library.sh` line 56 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONPIcjvCRGHlEf4SqgK:-ONsXLID4hTcfQlEtzgF:bw9ry7t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/224ffc5a016634e6bd996fbbd084e14448d79dc4/ios/build-rust-library.sh#L56)):*

Is the added whitespace intended? :blush:
</blockquote></details>

It was not intended :see_no_evil: 
Done

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: 42 of 50 files reviewed, 12 unresolved discussions (waiting on @dlon, @MarkusPettersson98, and @rablador)

___
*[`mullvad-ios/src/api_client/mod.rs` line 149 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONoeQ-X-0heIqZTtUjU:-ONtfuP60yjiqN_ds2zh:bzeamaz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/mod.rs#L149)):*

Isn't it only the `bridge_provider.into_rust_context()` call which is `unsafe`? Consider introducing a temporary `let`-binding to make this `unsafe` block easier to vet:

```rust
let method_resolver = {
    // SAFETY: TODO
    let bridge_provider = unsafe { bridge_provider.into_rust_context() };
    SwiftAccessMethodResolver::new(
        endpoint.clone(),
        domain,
        encrypted_dns_proxy_state,
        *bridge_provider,
    )
};
```

Side note: Why are we creating a `Box<SwiftShadowsocksLoaderWrapperContext>` to `bridge_provider` and immediately dereference the `Box`?
</blockquote></details>

We discussed this offline, and the code was improved as a follow-up with less `Box` dancing, and a Rustier implementation.
___
*[`mullvad-ios/src/api_client/shadowsocks_loader.rs` line 18 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONojD8f8vhhnufku1R-:-ONtgbo-1c0iMs8kbILd:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/shadowsocks_loader.rs#L18)):*

Allocating a `Box` and immediately giving up ownership of it using `Box::into_raw` seems a bit strange to me. Is there a reason why you don't keep the `Box`?
</blockquote></details>

Done.
___
*[`mullvad-ios/src/api_client/shadowsocks_loader.rs` line 55 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONomMVsDCxk_H0pJiWu:-ONth1wi4RqCIv5gqXWV:bxvg3yv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/shadowsocks_loader.rs#L55)):*

Valid for how long? Who's responsible for freeing this?
</blockquote></details>

I edited the comment to reflect more acurately what is the contract with this function.
___
*[`mullvad-ios/src/api_client/shadowsocks_loader.rs` line 60 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONol9cA77OAfc_qmw-c:-ONthyKbDwfHFiNT_2Zs:bmvriui) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/shadowsocks_loader.rs#L60)):*

What is accomplished by wrapping this pointer twice and boxing it?
</blockquote></details>

The excessive boxing has been removed

dlon (commented):
> Reviewed 2 of 48 files at r1, 2 of 4 files at r3, 2 of 2 files at r4, 2 of 8 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: 48 of 50 files reviewed, 9 unresolved discussions (waiting on @buggmagnet and @MarkusPettersson98)

___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 129 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONorCLq-oRtIgQ_l9vB:-ONxhL21ClWFl8aECub4:b-3zaylg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/access_method_settings.rs#L129)):*

We had a discussion about this, I had this original prototype.
```rust
   let custom_methods: Vec<AccessMethodSetting> = match custom_methods.is_null() {
        true => vec![],
        false => {
            let custom_methods = unsafe { from_raw_parts(custom_methods, custom_methods_count) };
            let mut methods_vec: Vec<AccessMethodSetting> =
                Vec::with_capacity(custom_methods_count);
            for index in custom_methods {
                let accessMethodRaw = custom_methods[index]; // :x: the type [c_void] cannot be indexed by &c_void
                let accessMethod = unsafe { *Box::from_raw(accessMethodRaw as *mut _) };
                methods_vec.append(accessMethod);
            }
            methods_vec
        }
    };
```

I think we could do this if we re-define all the structures we use for access methods at the FFI layer (or make ours `#[repr(C)]`) but I didn't want to make that jump and have the same structures defined multiple times, so I opted for this instead.

What do you think ? We could pair up on this if you want ?
</blockquote></details>

Let's talk about it. I think we can make the smaller improvement at least, even without duplicate C types for everything, which might not be worth it.

We don't necessarily need to block the PR on this, unless you want to.
___
*[`mullvad-ios/src/api_client/shadowsocks_loader.rs` line 55 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONomMVsDCxk_H0pJiWu:-ONxhCqI7aF56gP2NzjS:b-8fx2u7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/shadowsocks_loader.rs#L55)):*

I edited the comment to reflect more acurately what is the contract with this function.
</blockquote></details>

:+1: 
___
*[`mullvad-ios/src/api_client/shadowsocks_loader.rs` line 37 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONxgf4J4K43bUcS-SBt:-ONxgf4J4K43bUcS-SBu:b2nz1ay) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/678607a2d08258ed98ae94062bbe7a38b834170e/mullvad-ios/src/api_client/shadowsocks_loader.rs#L37)):*
> ```XML
> #[derive(Debug)]
> #[repr(C)]
> pub struct SwiftShadowsocksLoaderWrapperContext {
> ```

Isn't this struct a bit pointless? I would just move the pointer into `SwiftShadowsocksLoaderWrapper`. As long as you keep the pointer private, I think this would be fine

rablador (commented):
> Reviewed 8 of 8 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 9 unresolved discussions (waiting on @buggmagnet and @MarkusPettersson98)

rablador (approved):
Swifty parts :lgtm:

> Reviewed 5 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-:-ONyp_CN6dh7c1jqDOe1:b-wn6j4v)* status: all files reviewed, 8 unresolved discussions (waiting on @buggmagnet and @MarkusPettersson98)

dlon (commented):
> Reviewed 3 of 5 files at r7.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 13 unresolved discussions (waiting on @buggmagnet and @MarkusPettersson98)

___
*[`mullvad-ios/src/api_client/helpers.rs` line 10 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OO-2z9mDi38EFQF2xyL:-OO-2z9mDi38EFQF2xyM:b-5xssvg) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/84d10a4d73f7b8e5cfd67a46e49b6493fb3dddb7/mullvad-ios/src/api_client/helpers.rs#L10)):*
> ```XML
> /// Constructs a new IP address from a pointer containing bytes representing an IP address.
> ///
> /// SAFETY: `addr` must be a pointer to at least `addr_len` bytes.
> ```

```
# Safety
```
___
*[`mullvad-ios/src/api_client/helpers.rs` line 14 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OO-2rlE5zcI3m3nlSQQ:-OO-2rlE5zcI3m3nlSQR:bb45143) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/84d10a4d73f7b8e5cfd67a46e49b6493fb3dddb7/mullvad-ios/src/api_client/helpers.rs#L14)):*
> ```XML
>     match addr_len {
>         4 => {
>             // SAFETY: `addr` pointer must be non-null, aligned, and point to at least addr_len bytes
> ```

This should be mentioned in the `# Safety` section of the *function* docs. That's because the caller of the function is the one who's being asked to meet the requirements.

Here, you can simply state that addr *is* non-null, aligned, and initialized.
___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 121 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OO--t5kCcwQrv3MmpoT:-OO--t5kCcwQrv3MmpoU:be8bh75) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/84d10a4d73f7b8e5cfd67a46e49b6493fb3dddb7/mullvad-ios/src/api_client/access_method_settings.rs#L121)):*
> ```XML
> /// # SAFETY
> /// `direct_method_raw`, `bridges_method_raw` and `encrypted_dns_method_raw` must be raw pointers
> /// resulting from a call to `convert_builtin_access_method_setting`
> ```

Should mention `custom_methods_raw`
___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 128 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OO--vzE6u6lWAEI5wA9:-OO--vzE6u6lWAEI5wAA:b8r7d7y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/84d10a4d73f7b8e5cfd67a46e49b6493fb3dddb7/mullvad-ios/src/api_client/access_method_settings.rs#L128)):*
> ```XML
>     encrypted_dns_method_raw: *const c_void,
>     custom_methods_raw: *const c_void,
>     count: isize,
> ```

Nit: Maybe `num_custom_methods_raw` or something?
___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 145 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OO-0dsC7C7sOk9Z534q:-OO-0dsC7C7sOk9Z534r:bjo7l4v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/84d10a4d73f7b8e5cfd67a46e49b6493fb3dddb7/mullvad-ios/src/api_client/access_method_settings.rs#L145)):*
> ```XML
> }
> 
> unsafe fn access_methods_from_raw_vector(
> ```

Nit: I'd prefer "array" to vector, simply because the latter can be confused with the `Vec` type
___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 147 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OO-0Nu94xGuduaVPBMa:-OO-0Nu94xGuduaVPBMb:bjp4m6l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/84d10a4d73f7b8e5cfd67a46e49b6493fb3dddb7/mullvad-ios/src/api_client/access_method_settings.rs#L147)):*
> ```XML
> unsafe fn access_methods_from_raw_vector(
>     vector_raw: *const c_void,
>     count: isize,
> ```

An unsigned integer is slightly better here, IMO, even if you have to `as isize` it, since a negative count makes no sense. It's nice to express this in the type. But since it's obvious, it's mostly a matter of principle

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 13 unresolved discussions (waiting on @buggmagnet and @MarkusPettersson98)

dlon (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 14 unresolved discussions (waiting on @buggmagnet and @MarkusPettersson98)

___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 153 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OO-6FAx50rpIqlWq_id:-OO-6FAx50rpIqlWq_ie:bwen1sl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/84d10a4d73f7b8e5cfd67a46e49b6493fb3dddb7/mullvad-ios/src/api_client/access_method_settings.rs#L153)):*
> ```XML
>         .unwrap_or_default();
>     let vector_raw: *mut *mut AccessMethodSetting = vector_raw as _;
>     for i in 0..count {
> ```

Sorry about dragging this out further (and for not realizing this when we wrote the code), but `slice::from_raw_parts` would work here.

```
    vector_raw: *const c_void,
    count: usize,
) -> Vec<AccessMethodSetting> {
    // SAFETY: todo
    let slice = unsafe { slice::from_raw_parts(vector_raw, count) };
    slice.iter().map(|&ptr| {
        // SAFETY: `slice` is a slice of pointers to `AccessMethodSetting`
        unsafe { Box::from_raw(ptr) }
    }).collect()
}
```

This would also turn `count` into a `usize`.

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 14 unresolved discussions (waiting on @buggmagnet and @MarkusPettersson98)

rablador (changes requested):
> Reviewed 39 of 39 files at r8.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: 21 of 63 files reviewed, 12 unresolved discussions (waiting on @buggmagnet, @dlon, and @MarkusPettersson98)

___
*[`ios/MullvadVPN/AddressCacheTracker/AddressCacheTracker.swift` line 122 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OO2rz5-2EpzIrpfg-TU:-OO2rz5-2EpzIrpfg-TV:b2dbiwr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/58d963da0e36309b1aa7e1fe862ed9da5bd2fd35/ios/MullvadVPN/AddressCacheTracker/AddressCacheTracker.swift#L122)):*
> ```Swift
> 
>         return Date()
> //        return _nextScheduleDate()
> ```

Don't forget to revert this.

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: 21 of 63 files reviewed, 8 unresolved discussions (waiting on @dlon, @MarkusPettersson98, and @rablador)

___
*[`mullvad-ios/src/api_client/shadowsocks_loader.rs` line 37 at r6](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONxgf4J4K43bUcS-SBt:-OORGBXyBeZi9CIWSISi:b-hwuwhl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/678607a2d08258ed98ae94062bbe7a38b834170e/mullvad-ios/src/api_client/shadowsocks_loader.rs#L37)):*

Isn't this struct a bit pointless? I would just move the pointer into `SwiftShadowsocksLoaderWrapper`. As long as you keep the pointer private, I think this would be fine
</blockquote></details>

It is a bit pointless, you are correct. However, I want to use this as a reference on "how can we do X" in future implementations where we will consume more Rust code from swift. So I'd like to keep it that way for now.
___
*[`ios/MullvadVPN/AddressCacheTracker/AddressCacheTracker.swift` line 122 at r8](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OO2rz5-2EpzIrpfg-TU:-OORH2DL4ecMGu8vm3On:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/58d963da0e36309b1aa7e1fe862ed9da5bd2fd35/ios/MullvadVPN/AddressCacheTracker/AddressCacheTracker.swift#L122)):*

Don't forget to revert this.
</blockquote></details>

Done.

dlon (commented):
> Reviewed 5 of 31 files at r29, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: 26 of 63 files reviewed, 8 unresolved discussions (waiting on @MarkusPettersson98 and @rablador)

___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 82 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-ONosjBRETC96EU2Ribc:-OOR_Q4L3wu-n3g2lVKv:byp84a7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4807401ceee51aeb23329f1081b971317730e3a7/mullvad-ios/src/api_client/access_method_settings.rs#L82)):*

Done
</blockquote></details>

I think the documentation is incorrect. It only assumes ownership of `proxy_configuration`, no?
___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 121 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OO--t5kCcwQrv3MmpoT:-OORa3wc65vdm88sPPBJ:bt1irzd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/84d10a4d73f7b8e5cfd67a46e49b6493fb3dddb7/mullvad-ios/src/api_client/access_method_settings.rs#L121)):*

Done
</blockquote></details>

It's an array of pointers to `AccessMethodSetting`s, not `Box<AccessMethodSetting>`
___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 128 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OO--vzE6u6lWAEI5wA9:-OORaRdU2wXTACjkQDYj:bd7ael5) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/84d10a4d73f7b8e5cfd67a46e49b6493fb3dddb7/mullvad-ios/src/api_client/access_method_settings.rs#L128)):*

Done
</blockquote></details>

This is still just called "count".
___
*[`mullvad-ios/src/api_client/access_method_settings.rs` line 165 at r29](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OORag0dAC_KkFtMn8Ng:-OORag0dAC_KkFtMn8Nh:boratnb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/efc432b362b972c527933df984b81695140eda36/mullvad-ios/src/api_client/access_method_settings.rs#L165)):*
> ```XML
>         .iter()
>         .map(|&ptr| {
>             // SAFETY: `slice` is a slice of pointers to `Box<AccessMethodSetting>` objects
> ```

Probably my fault, but it's actually a slice of pointers to `AccessMethodSetting` created with `Box::into_raw`

dlon (approved):
> Reviewed 1 of 31 files at r31.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: 22 of 63 files reviewed, 8 unresolved discussions (waiting on @buggmagnet, @MarkusPettersson98, and @rablador)

dlon (commented):
> Reviewed 1 of 31 files at r31, 6 of 31 files at r35, 4 of 48 files at r36, 13 of 46 files at r38.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: 29 of 68 files reviewed, 9 unresolved discussions (waiting on @MarkusPettersson98 and @rablador)

___
*[`mullvad-ios/src/ephemeral_peer_proxy/ios_tcp_connection.rs` line 36 at r38](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OOXz4mSCQBjXnWMnkH_:-OOXz4mSCQBjXnWMnkHa:b73y31t) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d5eea59f641527f8c1e2819478f2f81d43767c8/mullvad-ios/src/ephemeral_peer_proxy/ios_tcp_connection.rs#L36)):*
> ```XML
>     /// # Safety
>     /// This function is safe to call so long as the function pointer is valid for its declared
>     /// signature.
> ```

I think the address must be valid too (or at least a null-terminated string)?
___
*[`mullvad-ios/src/api_client/problem_report.rs` line 119 at r38](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OOXyic3Dlkv0GpIbIZ1:-OOXyic3Dlkv0GpIbIZ2:bkfcaco) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4d5eea59f641527f8c1e2819478f2f81d43767c8/mullvad-ios/src/api_client/problem_report.rs#L119)):*
> ```XML
> }
> 
> // SAFETY: Swift does not write to, or use this type from different threads
> ```

Justification seems a bit weird. You could probably copy the other one.

dlon (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-:-OOY-Eit31lOlqmJMDk-:bnfp4nl)* status: 29 of 68 files reviewed, 9 unresolved discussions (waiting on @buggmagnet, @MarkusPettersson98, and @rablador)

MarkusPettersson98 (dismissed):
> Reviewed 1 of 48 files at r1, 3 of 40 files at r8, 1 of 31 files at r29, 1 of 17 files at r37, 19 of 46 files at r38, 1 of 5 files at r39.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: 62 of 68 files reviewed, 5 unresolved discussions (waiting on @buggmagnet, @dlon, and @rablador)

___
*[`mullvad-ios/src/shadowsocks_proxy/ffi.rs` line 51 at r39](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OOadSRq0TZCMjOCofnC:-OOadSRq0TZCMjOCofnD:b-3mo08x) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/523987fc7276bcba815f2eb4710610edfd5a1760/mullvad-ios/src/shadowsocks_proxy/ffi.rs#L51)):*
> ```XML
>     } else {
>         return -1;
>     };
> ```

:pick: `if-let` is not the appropriate way to write this expression, `let-else` would suit better where you just want to return early in the `else`-case

```rust
let Some(bridge_ip) = parse_ip_addr(addr, addr_len) else {
    return -1;
};
```

The same comment can be applied in other places as well.

Won't block on this, but putting it out there as an fyi :blush: 



_Code quote:_
```XML
    let bridge_ip = if let Some(addr) = { parse_ip_addr(addr, addr_len) } {
        addr
    } else {
        return -1;
    };
```
___
*[`mullvad-ios/src/api_client/shadowsocks_loader.rs` line 56 at r39](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986#-OOacjeBEig-CNm_rDSq:-OOacjeBEig-CNm_rDSr:bn6fmbc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/523987fc7276bcba815f2eb4710610edfd5a1760/mullvad-ios/src/api_client/shadowsocks_loader.rs#L56)):*
> ```XML
>         }
>         // SAFETY: The pointer returned by `swift_get_shadowsocks_bridges` is guaranteed
>         // to point to a valid `Shadowsocks` configuration
> ```

:pick: Also, you did null-check the pointer!



_Code quote:_
```XML
        // SAFETY: The pointer returned by `swift_get_shadowsocks_bridges` is guaranteed
        // to point to a valid `Shadowsocks` configuration
```

rablador (commented):
> Reviewed 1 of 17 files at r37, 2 of 46 files at r38, 5 of 5 files at r39, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

rablador (approved):
> Reviewed 4 of 4 files at r40, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7986)* status: all files reviewed, 5 unresolved discussions (waiting on @buggmagnet)

PR #7991, opened 2025-04-09
Add mocked e2e test for app upgrade view

tobias-jarvelov:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7991)

### DIFF
--- a/desktop/packages/mullvad-vpn/test/e2e/mocked/app-upgrade/app-upgrade.spec.ts
+++ b/desktop/packages/mullvad-vpn/test/e2e/mocked/app-upgrade/app-upgrade.spec.ts
@@ -1,9 +1,9 @@
 import { expect, test } from '@playwright/test';
 import { Page } from 'playwright';
 
-import { AppUpgradeError, AppUpgradeEvent } from '../../../src/shared/app-upgrade';
-import { IAppVersionInfo } from '../../../src/shared/daemon-rpc-types';
-import { MockedTestUtils, startMockedApp } from './mocked-utils';
+import { AppUpgradeError, AppUpgradeEvent } from '../../../../src/shared/app-upgrade';
+import { IAppVersionInfo } from '../../../../src/shared/daemon-rpc-types';
+import { MockedTestUtils, startMockedApp } from '../mocked-utils';
 
 let page: Page;
 let util: MockedTestUtils;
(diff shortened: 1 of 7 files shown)


olmoh:
Some initial thoughts on this:

I think it’d be better if each test handled its own setup and teardown instead of relying on state from previous tests. This would have several benefits, like:

- If one test fails, the others can still pass. Right now, a failure in one test causes the rest to fail since they depend on the state from an earlier test.
- It makes tests independent, which means they can run in parallel.

We could handle this by resetting the state in the `afterEach` function and navigating to the app upgrade page in `beforeEach`. If we don’t want to do this for every test, maybe we could split it up into smaller flows. For instance, a flow like should `automatically start download if focused` could begin from a reset app upgrade view, click download, and then verify that it installs.

Otherwise, it looks good, and I think we’re testing some valuable things.

tobias-jarvelov:
Note: This PR is dependent on the changes in the #7989 PR and it's tests won't run correctly until that PR has been merged to the `in-app-upgrades-on-windows-and-macos` branch.

desktop/packages/mullvad-vpn/test/e2e/mocked/app-upgrade/app-upgrade.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+        ),
+      ).toBeVisible();
+
+      const retryDownloadButton = selectors.retryButton();
+      await expect(retryDownloadButton).toBeVisible();
+
+      const reportProblemButton = selectors.reportProblemButton();
+      await expect(reportProblemButton).toBeVisible();
+    });
+    test('Should handle retrying downloading upgrade', async () => {
```
olmoh, line 136: This test currently fails with the latest changes which keeps the last progress in the redux state.

desktop/packages/mullvad-vpn/test/e2e/mocked/app-upgrade/app-upgrade.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+    test.afterAll(() => helpers.testTeardown());
+
+    test('Should start upgrade when clicking Download & install button', async () => {
+      await helpers.startAppUpgrade();
+      const downloadAndInstallButton = selectors.downloadAndInstallButton();
+      await expect(downloadAndInstallButton).toBeHidden();
+    });
+
+    test('Should show indeterminate download progress after upgrade started', async () => {
+      // TODO: Improve by using aria labels
```
olmoh, line 78: We can remove this TODO.

olmoh (changes requested):
Nice job has gotten a lot cleaner! However one test currently fails with the changes from the base branch.

olmoh (approved):

PR #7995, opened 2025-04-10
Support initial prop and conditionally render Animate

olmoh:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/7995)


tobias-jarvelov (commented):
As we talked about previously we wanted to do something about the number of re-renderers required for the current implementation. I've refactored this code to reduce the number of re-renderers required. Let's go over it together and see what we feel about it.

The refactor is in this branch: `refactor-allow-animate-component-to-skip-initial-animations-des-1997`

PR #8012, opened 2025-04-11
Fix nav drawer wrong focus on Android TV

kl:
When switching locations on Android TV when the app goes back to the connect screen, focus was sometimes incorrectly put on the navigation drawer, which caused it to expand when it shouldn't.

This commit fixes that by detecting when we navigate back to the connect screen from the select location screen, and when that is the case we manually focus the select location button.

We check if we get a FocusDirection.Down event to detect if we navigated back from the select location screen, which isn't optimal, but I couldn't find a better way to detect this. This works because if we return from the account or settings screens we get a FocusDirection.Enter event.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8012)

### DIFF
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/compose/screen/ConnectScreen.kt
@@ -40,8 +40,13 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
+import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.alpha
+import androidx.compose.ui.focus.FocusDirection
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.focus.focusProperties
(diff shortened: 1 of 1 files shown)


Rawa (commented):
Do we know why the result is different from those two screens? Is it because we use a `BackNavigator` with a result? Maybe we can fix the cause of this behavior instead?

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8012#-:-ONZr-eB0N-oianOeeEP:bzihtv0)* status: 0 of 1 files reviewed, all discussions resolved

kl (commented):
Seeing the same behavior when I tested without the back navigator. So it seems the behavoir is 
navigating back from drawer -> FocusDirection.Enter
navigating back from select location -> FocusDirection.Down

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8012#-:-ONnLKT-0Wjdvo9F0aRI:b-52lez5)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Rawa (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8012#-:-ONoXn1h8g7Lmjq4cZqL:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #8024, opened 2025-04-14
Set opt-level=3 globally

hulthe:
I've seen a noticeable performance increase on mullvad-masque-proxy by setting opt-level=3 on it. I suspect that we'd see an improvement for shadowsocks, maybe udp2tcp, and maybe macos split tunneling.

Given the number of dependencies that these crates have, it would be a hassle to set opt-level=3 individually on each dependency that might benefit from it.

Here are size comparisons for some binaries that would be affected:

| bin             | opt-level=s | opt-level=3 | tested on |
| --------------- | ----------- | ----------- | --- |
| mullvad-daemon  | 15.9 MiB    | 22.7 MiB    | linux |
| mullvad-cli     | 5.4 MiB     | 7.3 MiB     | linux |
| mullvad-exclude | 362.3 KiB   | 367.5 KiB   | linux |

With opt-level=3: https://releases.mullvad.net/desktop/builds/2025.5-dev-66a2c9%2Btest-opt-level-3/?C=M&O=D

With opt-level=s: https://releases.mullvad.net/desktop/builds/2025.5-dev-b2b32e/?C=M&O=D

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8024)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -132,12 +132,6 @@ proptest = "1.4"
 insta = { version = "1.42", features = ["yaml"] }
 
 [profile.release]
-opt-level = "s"
+opt-level = 3
 lto = true
 strip = true
-
-# Key generation may take over one minute without optimizations enabled.
-[profile.dev.package."classic-mceliece-rust"]
-opt-level = 3
-[profile.release.package."classic-mceliece-rust"]
-opt-level = 3
--- a/desktop/scripts/pack-universal-win.sh
+++ b/desktop/scripts/pack-universal-win.sh
@@ -57,6 +57,7 @@ if [[ -z ${WIN_X64_INSTALLER-} ]] || [[ -z ${WIN_ARM64_INSTALLER-} ]]; then
     exit 1
 fi
 
+export RUSTFLAGS="-C opt-level=s"
 cargo build "${CARGO_ARGS[@]}" -p windows-installer --target x86_64-pc-windows-msvc
 
 dest="dist/MullvadVPN-${PRODUCT_VERSION}.exe"


faern:
This was changed the other way around recently: #6703. See that PR for some motivation and background.

That PR did not focus too much on speed, since we did not have many such measurements. But it was clear that we got a 20-30% improvement in both compile time and binary size. I feel like we can bounce back and fourth forever unless we clarify what our priorities here are. Both compile time and binary size are important metrics.

This PR does not provide a lot of data about how much performance is gained and how that was measured. Performance is way harder to measure than compile time and binary size. Performance varies a lot with the environment it's tested in.

I think I would prefer an approach where we find out which crates gain the most from `opt-level = 3`, and selectively optimize those, as we have done with `classic-mceliece-rust`. Most likely, not all dependencies of `mullvad-masque-proxy` need more optimization to get the benefits you are seeing. It's more likely that there is one hot code path *somewhere* and optimizing that crate gets us most of the way.

hulthe (commented):
Thanks for providing context!

> Both compile time and binary size are important metrics.

No disagreement here. It's a trade-off and we'll need to have a discussion about it.
Hadn't occurred to me that compile-times would be affected, good that you brought that up.

> we got a 20-30% improvement in binary size

I saw the same results when `cargo build`ing. Interesting (and irrelevant): the difference in the `deb`, `rpm`, `exe`, and `pkg`-bundles don't seem to grow by the same amount of MBs. Compression, I guess.

> This PR does not provide a lot of data

I'm seeing an 8-10% increase of network throughput in masque on our benchmarking servers. I'd be happy to discuss the details of the benchmarks IRL if you're interested.

> It's more likely that there is one hot code path _somewhere_

Potentially, but I haven't found one. Looking at the dependency tree, I can think of a dozen-or-so dependencies that might benefit from opt-level=3. I'm concerned that we'd need to spend a large amount of time digging to find them, and we'd end up with a long list that'd be difficult to maintain. Then we'd need to repeat that work for shadowsocks, boringtun, etc.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8024#-:-ONtirPI8JEWSi9Eu_-_:bv4dvtk)* status: 0 of 2 files reviewed, all discussions resolved

PR #8036, opened 2025-04-15
Replace Classic McEliece with HQC

faern:
This is WIP since our servers do not yet support HQC.

NIST has announced that they standardize the HQC KEM algorithm as their "backup" post-quantum secure key exchange mechanism: https://www.nist.gov/news-events/news/2025/03/nist-selects-hqc-fifth-algorithm-post-quantum-encryption

HQC is not based on the same fundamental mathematical problem (error-correcting codes) as ML-KEM (structured lattices). And as a result we keep the important property that our quantum-resistant tunnels continue being safe until **two** problems have been cracked.

The main benefits here are that HQC uses a lot less CPU (~30x) to compute the keypair, and the public key sent to the server is drastically smaller (~75x). This will probably mitigate a lot of the issues we are currently seeing with clients being unable to negotiate the ephemeral peer while they have quantum-resistant tunnels enabled.

This PR uses the [`pqcrypto-hqc`](https://crates.io/crates/pqcrypto-hqc) crate for the implementation of HQC. This crate in turn is just bindings to the C implementation in the PQClean project. My hopes are that the RustCrypto organization will release a pure Rust implementation later. They very recently squatted the crate name [`hqc`](https://crates.io/crates/hqc).

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8036)


hulthe (commented):
> Reviewed 11 of 12 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8036)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

faern:
@dlon performed some rudimentary benchmarking on this. The goal was see approximately how must faster it could end up being in the real world when you combine the fact that HQC keys both cost way less CPU to generate than CME, and the key material to upload to the server is way smaller. This was measured between two machines on the same local network, but with simulated packet loss and latency. This was simulated by running the ephemeral peer negotiation server and client from our repository on both ends.

```
# Results: PQ with CME

## No loss, no delay
0.80 s

## 1% packet loss, 50ms delay
mean: 3.73 s
min: 1.49 s
max: 5.60 s

## No loss, 500ms delay
10 s

# Results: PQ with HQC

## No loss, no delay
0.187 s

## 1% packet loss, 50ms delay
mean: 0.90 s
min: 0.69 s
max: 1.7 s

## No loss, 500ms delay
5.2 s
```

These measurements are by no means conclusive nor verified to be statistically sound. But they do indicate that HQC is noticeably faster than CME.

hulthe (dismissed):
> Reviewed 2 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8036)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #8037, opened 2025-04-15
Improvements to app upgrade implementation

olmoh:
- Makes progress bar visible most steps of app upgrade flow after initialization
- Animates label and progress bar separately to make animations stutter less
- Removes disable from app upgrade list item when connection is blocked

This also adds a state in app upgrades redux state which currently keeps track of the latest progress percentage number. This is used to prevent some cases where the progress could be set to 0 when toggling the connection on and off.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8037)

### DIFF
--- a/desktop/packages/mullvad-vpn/src/renderer/hooks/useAppUpgradeDownloadProgressValue/useAppUpgradeDownloadProgressValue.ts
+++ b/desktop/packages/mullvad-vpn/src/renderer/hooks/useAppUpgradeDownloadProgressValue/useAppUpgradeDownloadProgressValue.ts
@@ -1,3 +1,4 @@
+import { useAppUpgradeLastProgress } from '../../redux/hooks';
 import { useAppUpgradeEventType } from '../useAppUpgradeEventType';
 import { useHasAppUpgradeError } from '../useHasAppUpgradeError';
 import { useHasAppUpgradeVerifiedInstallerPath } from '../useHasAppUpgradeVerifiedInstallerPath';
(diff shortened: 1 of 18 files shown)


desktop/packages/mullvad-vpn/src/renderer/redux/app-upgrade/reducers.ts
```diff
@@ -5,12 +5,16 @@ export interface AppUpgradeReduxState {
   error?: AppUpgradeError;
   errorCount: number;
   event?: AppUpgradeEvent;
+  state: {
+    progress?: number;
+  };
```
tobias-jarvelov, line 10: **suggestion:** Name this `lastProgress` and put it directly on to root state.

desktop/packages/mullvad-vpn/src/renderer/redux/app-upgrade/reducers.ts
```diff
@@ -22,6 +26,13 @@ export function appUpgradeReducer(
       return {
         ...state,
         event: action.event,
+        state:
+          action.event.type === 'APP_UPGRADE_STATUS_DOWNLOAD_PROGRESS'
+            ? {
+                ...state.state,
+                progress: action.event.progress,
+              }
+            : state.state,
```
tobias-jarvelov, line 35: **suggestion:** Move the logic to `app.tsx` for extracting progress and call a redux action `setLastProgress`.

desktop/packages/mullvad-vpn/src/renderer/hooks/useAppUpgradeDownloadProgressValue/useAppUpgradeDownloadProgressValue.ts
```diff
@@ -29,6 +31,6 @@ export const useAppUpgradeDownloadProgressValue = () => {
     case 'APP_UPGRADE_STATUS_VERIFYING_INSTALLER':
       return DOWNLOAD_COMPLETE_VALUE;
     default:
-      return FALLBACK_VALUE;
+      return progress ?? FALLBACK_VALUE;
   }
```
tobias-jarvelov, line 35: **suggestion:** Put a break here and handle this logic in an if clause after the switch.

PR #8063, opened 2025-04-22
Pre-populate Problem report form with upgrade error in description

tobias-jarvelov:
Pre-populate the Report problem form with a message related to the kind of error encountered.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8063)

### DIFF
--- a/desktop/packages/mullvad-vpn/locales/messages.pot
+++ b/desktop/packages/mullvad-vpn/locales/messages.pot
@@ -672,6 +672,12 @@ msgctxt "app-upgrade-view"
 msgid "About %(seconds)s seconds remaining..."
 msgstr ""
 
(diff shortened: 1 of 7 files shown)


tobias-jarvelov:
@olmoh One thing I am not sure about how to approach is whether we should clear the Problem report field sometime, and also if we should save any info the user has previously written there. Let's think about it and discuss it.

tobias-jarvelov:
Oliver and I discussed this and we don't think this is a good feature. We think that a relevant error should be present in the logs, not pre-populated in the report problem field. This PR and the issue on Linear will be closed. However, I will talk to Carl and see if it is something we still want to do, and in that case the PR and issue on Linear can be re-opened.

PR #8112, opened 2025-05-05
Let users cancel sending a problem report

rablador:
Why this needs to be done

Currently if you are stuck connecting and want to send a problem report, you will also be stuck there with no way of going back to disconnect so that you can actually send the problem report.

What needs to be done

- Add a cancel button to the "Sending…" view. This will most likely be disabled during app log consolidation.
- "Failed to send" does not prompt a simple "Internal error" message. Instead we add text that explains that the user can try again later and their message will still be there, or they can email support directly. This also adds the view app logs button on the failed to send view.

Acceptance criteria

- Problem reports can be cancelled.
- Failed to send view UX has updated copy and added button.
Tests

We should unit test the cancellation behavior.


buggmagnet (changes requested):
> Reviewed 18 of 20 files at r1, 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112)* status: 20 of 21 files reviewed, 2 unresolved discussions

___
*[`ios/MullvadVPN.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` line 0 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OPpA8jwC8I-1vPWTi5D:-OPpA8jwC8I-1vPWTi5E:b-lkhx6v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f75e91db5336b199c0c8ee839e01de923c00281/ios/MullvadVPN.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved#L0)):*
_oh no_
___
*[`ios/MullvadVPN/View controllers/ProblemReport/ProblemReportSubmissionOverlayView.swift` line 338 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OPpF3A34aw_NV7M-iz0:-OPpF3A34aw_NV7M-iz1:bvd4kwj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f75e91db5336b199c0c8ee839e01de923c00281/ios/MullvadVPN/View%20controllers/ProblemReport/ProblemReportSubmissionOverlayView.swift#L338)):*
> ```Swift
> 
>     @objc private func handleEmailLabelTap() {
>         if let url = URL(string: "mailto:\(state.supportEmail)") {
> ```

nit
That's a nice touch !

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112)* status: 20 of 21 files reviewed, 1 unresolved discussion (waiting on @buggmagnet)

___
*[`ios/MullvadVPN.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` line  at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OPpA8jwC8I-1vPWTi5D:-OPpHUDkBi5sy8lF4PFY:bepeo6q) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f75e91db5336b199c0c8ee839e01de923c00281/ios/MullvadVPN.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved#L)):*

_oh no_
</blockquote></details>

Huh? Weird, but I'll fix.

mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112)* status: 19 of 21 files reviewed, 4 unresolved discussions (waiting on @buggmagnet)

___
*[`ios/MullvadVPN/Notifications/Notification Providers/LatestChangesNotificationProvider.swift` line 63 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OQ7nv4H1_EeS8KIYjoO:-OQ7nv4H1_EeS8KIYjoP:bijvrry) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/89613518fbb6a92d398b90875f1470019af56ef8/ios/MullvadVPN/Notifications/Notification%20Providers/LatestChangesNotificationProvider.swift#L63)):*
> ```Swift
>                 comment: ""
>             ),
>             options: MarkdownStylingOptions(
> ```

I highly recommend keeping the closure that returns a `Custom Style` for bold text, instead of adding properties specific to text c

olor or  formatting.
___
*[`ios/MullvadVPN/Extensions/NSAttributedString+Extensions.swift` line 30 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OQ7jgMWAKIit2fBZdi3:-OQ7jgMWAKIit2fBZdi4:bvffwun) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a77dce6a754ba73868f36e04f5c44688e8b544d9/ios/MullvadVPN/Extensions/NSAttributedString%2BExtensions.swift#L30)):*
> ```Swift
>                 attributes[.font] = options.font
>             } else {
>                 attributes[.font] = options.boldFont
> ```

Passing the bold font from the caller doesn't meet the markdown string requirements. When we pass a string containing `**`, we want only the text between `**`, like `**Example**`, to appear bold — not the entire string. The new changes apply the font only to the marked text, not to the whole string. the rest of the text doesn't take the font.

but we can make it better in this way:



_Code snippet:_
```Swift
         if stringIndex % 2 == 0 {
                attributes[.font] = options.font
            } else {
                attributes[.font] = options.font.withWeight(.bold)
                attributes[.foregroundColor] = options.textColor
                attributes.merge(applyEffect?(string) ?? [:], uniquingKeysWith: { $1 })
            }
```
___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 60 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OQ84qPD0o2WR7LBzQop:-OQ84qPD0o2WR7LBzQoq:b44a8ss) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/89613518fbb6a92d398b90875f1470019af56ef8/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L60)):*
> ```Swift
> 
>     func addLogFiles(fileURLs: [URL], completion: (@Sendable (Result<String, Error>) -> Void)? = nil) {
>         let workItem = DispatchWorkItem(flags: .barrier) { [weak self] in
> ```

Since we cancel the `WorkItem` when a cancellation occurs, we need to stop appending to the log, which is dispatched directly on the queue. To handle this, we can create a helper function that checks the task status before modifying the log, and use this function in both `addError` and `addSingleLogFile`.



_Code snippet:_
```Swift
    private func safeAsync(execute: @escaping @Sendable () -> Void) {
        let isCancelled = workItem?.isCancelled ?? false
        guard !isCancelled else { return }

        logQueue.async(flags: .barrier) {
            if !isCancelled {
                execute()
            }
        }
    }
    
    private func addSingleLogFile(_ fileURL: URL) {
        guard fileURL.isFileURL else {
            addError(
                message: fileURL.absoluteString,
                error: "Invalid log file URL: \(fileURL.absoluteString)."
            )
            return
        }

        let path = fileURL.path
        let redactedPath = redact(string: path)

        if let lossyString = readFileLossy(path: path, maxBytes: bufferSize) {
            let redactedString = redact(string: lossyString)
            safeAsync {[weak self] in
                self?.logs.append(LogAttachment(label: redactedPath, content: redactedString))
            }
        } else {
            addError(message: redactedPath, error: "Log file does not exist: \(path).")
        }
    }
    
 private func addError(message: String, error: String, completion: (@Sendable () -> Void)? = nil) {
        let redactedError = redact(string: error)
        safeAsync {[weak self] in
            guard let self else { return }
            self.logs.append(LogAttachment(label: message, content: redactedError))
            DispatchQueue.main.async {
                completion?()
            }
        }
    }
```

mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112)* status: 19 of 21 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/View controllers/ProblemReport/ProblemReportViewController.swift` line 268 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OQ8VkWXAatuywCrLmCH:-OQ8VkWXAatuywCrLmCI:b-mq4qjk) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/89613518fbb6a92d398b90875f1470019af56ef8/ios/MullvadVPN/View%20controllers/ProblemReport/ProblemReportViewController.swift#L268)):*
> ```Swift
>             email: viewModel.email,
>             message: viewModel.message
>         ) { completion in
> ```

please, add `weak` here.due to keep these `Self` strong, the memory retains.

mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112)* status: 19 of 21 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 60 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OQ84qPD0o2WR7LBzQop:-OQ8Wf5U8UOCkb3Sh-M5:binx8ns) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/89613518fbb6a92d398b90875f1470019af56ef8/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L60)):*

Since we cancel the `WorkItem` when a cancellation occurs, we need to stop appending to the log, which is dispatched directly on the queue. To handle this, we can create a helper function that checks the task status before modifying the log, and use this function in both `addError` and `addSingleLogFile`.
</blockquote></details>

And don't forget to capture `self` as `weak` in the closures.

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112)* status: 19 of 21 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 60 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OQ84qPD0o2WR7LBzQop:-OQICGQ761w1VnpAPgIc:b-sg028g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/89613518fbb6a92d398b90875f1470019af56ef8/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L60)):*

And don't forget to capture `self` as `weak` in the closures.
</blockquote></details>

Does that matter here? One or two log file can slip past even if we check the cancellation further down the code paths. The idea is that we stop execution, but we don't care too much about the result of log collection. The one thing I'm concerned about is what happens if the user tries to send again and logs are half added.
___
*[`ios/MullvadVPN/Extensions/NSAttributedString+Extensions.swift` line 30 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OQ7jgMWAKIit2fBZdi3:-OQI5DiM79-9PKQikXwe:b7e770c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a77dce6a754ba73868f36e04f5c44688e8b544d9/ios/MullvadVPN/Extensions/NSAttributedString%2BExtensions.swift#L30)):*

Passing the bold font from the caller doesn't meet the markdown string requirements. When we pass a string containing `**`, we want only the text between `**`, like `**Example**`, to appear bold — not the entire string. The new changes apply the font only to the marked text, not to the whole string. the rest of the text doesn't take the font.

but we can make it better in this way:
</blockquote></details>

That is actuallt the reason why I changed it. It seems weird to me that we apply special styling to a part of a text (between the `**`) and then also implicitly apply parts of that styling to the rest of the string. The styling is part of the `MarkdownStylingOptions` object, which indicates (to me at least) that the styling should only be applied on the specific markdown parts. Instead I added an `attributes` param to make sure that the whole text can have a custom font etc. This way we can even have one font for the whole text and another for the markdown parts.
___
*[`ios/MullvadVPN/Notifications/Notification Providers/LatestChangesNotificationProvider.swift` line 63 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OQ7nv4H1_EeS8KIYjoO:-OQI74LUAh8FmfqniBs6:b5vvp8j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/89613518fbb6a92d398b90875f1470019af56ef8/ios/MullvadVPN/Notifications/Notification%20Providers/LatestChangesNotificationProvider.swift#L63)):*

I highly recommend keeping the closure that returns a `Custom Style` for bold text, instead of adding properties specific to text c

olor or  formatting.
</blockquote></details>

In that case we need to add more markdown types for each thing we want to change. Sometimes we only want a text color change rather than a bold font. My idea was that the constructor would be more flexible to mix and match between colors and font/font styles.
___
*[`ios/MullvadVPN/View controllers/ProblemReport/ProblemReportViewController.swift` line 268 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OQ8VkWXAatuywCrLmCH:-OQI6Znm5M4AOqfVLM6i:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/89613518fbb6a92d398b90875f1470019af56ef8/ios/MullvadVPN/View%20controllers/ProblemReport/ProblemReportViewController.swift#L268)):*

please, add `weak` here.due to keep these `Self` strong, the memory retains.
</blockquote></details>

Done.

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112)* status: 19 of 22 files reviewed, 6 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift` line 60 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-OQbON68927FztYNScpN:-OQbON69-Ric8Np-hBEf:b-5afzn7) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/96c6a90c2a3812c730773a6879772d6100442b4e/ios/MullvadVPN/Classes/ConsolidatedApplicationLog.swift#L60)):*
> ```Swift
> 
>     func addLogFiles(fileURLs: [URL], completion: (@Sendable (Result<String, Error>) -> Void)? = nil) {
>         let workItem = DispatchWorkItem(flags: .barrier) { [weak self] in
> ```

By the way, all these `.barrier` flags don't do anything, `logQueue` is not concurrent, and `DispatchQueues` by default are FIFO serial queues.

buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112)* status: 3 of 22 files reviewed, 5 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/View controllers/ProblemReport/ProblemReportSubmissionOverlayView.swift` line 66 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-ORG7rNZ2DKxdL0CidOE:-ORG7rNZ2DKxdL0CidOF:b-5o3nhl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55b510a3fdb920e7883b290e0d2bad5e26e72c75/ios/MullvadVPN/View%20controllers/ProblemReport/ProblemReportSubmissionOverlayView.swift#L66)):*
> ```Swift
>                         comment: ""
>                     ),
>                     attributes: [.foregroundColor: UIColor.successColor]
> ```

According to the source of truth, this should not be in the green color

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112)* status: 3 of 22 files reviewed, 5 unresolved discussions (waiting on @buggmagnet and @mojganii)

___
*[`ios/MullvadVPN/View controllers/ProblemReport/ProblemReportSubmissionOverlayView.swift` line 66 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8112#-ORG7rNZ2DKxdL0CidOE:-ORGKgX3DHQrxb8-B5Mh:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/55b510a3fdb920e7883b290e0d2bad5e26e72c75/ios/MullvadVPN/View%20controllers/ProblemReport/ProblemReportSubmissionOverlayView.swift#L66)):*

According to the source of truth, this should not be in the green color
</blockquote></details>

Done.

PR #8115, opened 2025-05-05
Add a first draft

buggmagnet:
This PR adds a first attempt draft that codifies by a couple of examples how our iOS code should try to interact with Rust.

It was written with GitHub in mind when it comes to the reading experience, and be viewed [here](https://github.com/mullvad/mullvadvpn-app/blob/codify-swift-rust-interactions-in-a-markdown-file-ios-1140/ios/FFI.md)

Please do send suggestions about what best practices you would like to see (or remove) from this document, I will compile them and write neat examples. 

This is meant to be a document we should keep in mind, and reference for future discussions on how to approach certain classes of problems we encounter in the wonderul FFI world.

### DIFF
--- a/ios/FFI.md
+++ b/ios/FFI.md
@@ -1,0 +1,156 @@
+# Rust and FFI
+
+This document is meant to provide best practices and conventions to follow
+when writing FFI code between Swift and Rust.
+
+## String and Data types
+
+- Whenever possible, try to take advantage of Swift's automatic String conversion
+- When dealing with buffers, a pointer to the start of the buffer, and its length should be passed
+- Swift will most of the time do the right thing by passing a pointer to the start of the buffer
+> [!IMPORTANT]
+> Always remember that the pointers passed to the FFI functions are **only valid for the lifetime of the call**
+> If you need to keep the contents of the pointer around for longer, make sure to copy it.
+
+### Examples
+
+Assuming the following FFI function
+
+```C
+void ffi_function(const char *str);
+void other_ffi_function(const uint8_t *address, uintptr_t address_len);
+```
+
+How to call it from Swift
+
+```swift
+let someString = "hello"
+ffi_function(hello)
+
+...
+
+let data = someString.data(using: .utf8)!
+let dataAsArray = data.map { $0 }
+other_ffi_function(dataAsArray, UInt(data.count))
+
+```
+
+## Opaque types
+
+When dealing with opaque types, keep in mind that only types that are natively representable in C can be passed across the FFI boundary.
+Fortunately, opaque types can be declared in such a way that they can be sent across the FFI to be reused later.
+
+### Example
+
+Here is a Rust type that accepts a pointer to an opaque Swift class sent as a `void *` type to Rust
+
+```rust
+#[repr(C)]
+pub struct LoaderWrapperContext {
+    // This pointer is a reference to a Swift object, and is only ever read by Rust.
+    // It is used to call that Swift object across the FFI
+    loader: *const c_void,
+}
+
+#[repr(C)]
+pub struct SwiftLoaderWrapper(LoaderWrapperContext);
+impl SwiftLoaderWrapper {
+    pub fn new(context: LoaderWrapperContext) -> SwiftLoaderWrapper {
+        SwiftLoaderWrapper(context)
+    }
+}
+```
+
+This allows to have a clean separation of the `unsafe` FFI boundary with a safe API in Rust
+
+```rust
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn init_swift_loader_wrapper(
+    loader: *const c_void,
+) -> SwiftLoaderWrapper {
+    let context = LoaderWrapperContext { loader };
+    SwiftLoaderWrapper::new(context)
+}
+
+impl SwiftLoaderWrapper {
+        pub fn safe_call(&self) -> Option<()> {
+            let context = self.context_ref();
+            Some(context.load_things())
+        }
+
+        fn context_ref(&self) -> &LoaderWrapperContext {
+        &self.0
+    }
+}
+
+impl LoaderWrapperContext {
+    pub fn load_things() -> Option<()> {
+        Some(unsafe { use_loader(self.loader) })
+    }
+}
+```
+
+
+Following the example above, here's how to provide an API from swift (That used from Rust in the previous example)
+
+```swift
+public func initLoaderWrapper(loader: Loader) -> SwiftLoaderWrapper {
+    let rawLoader = Unmanaged.passUnretained(loader).toOpaque()
+    return init_swift_loader_wrapper(rawLoader)
+}
+
+@_cdecl("use_loader")
+func useLoader(rawLoader: UnsafeMutableRawPointer?) {
+    guard let rawLoader else { return } }
+    let loader = Unmanaged<Loader>.fromOpaque(rawLoader).takeUnretainedValue()
+}
+```
+
+### More opaque types examples
+
+### Digesting C style arrays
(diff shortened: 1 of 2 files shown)


ios/FFI.md
```diff
@@ -0,0 +1,156 @@
...
+> Make sure to read the warning notice on `from_raw_parts`
+
+```rust
+unsafe fn generic_array_conversion_example<A>(raw_array: *const c_void, elems: usize) -> Vec<A>
+where
+    A: Sized,
+{
+    let raw_array: *mut *mut A = raw_array as _;
+    // SAFETY: `raw_array` must be aligned, non-null and initialized for `count` reads
+    let slice = unsafe { slice::from_raw_parts(raw_array, elems) };
```
faern, line 127: Here, `slice` has the type `&[*mut A]`. Is this intended? This means that the passed argument `raw_array` is an array of *pointers*, not of values. Is this how you want to be passing arrays from Swift -> Rust? Is there a reason that would be better than passing the array of values directly?

If you already have the objects you want to pass in contiguous memory in Swift, then it just sounds like more work having to create an array of pointers to it.

If the Rust side should be able to parse out values of type `A`, like this example indicate that it have to do. Then why pass in an opaque void pointer? Can't the function that Swift calls into take `data: *const TheType, len: usize` directly instead? Since `TheType` must be defined in Rust anyway.

faern (changes requested):

ios/FFI.md
```diff
@@ -0,0 +1,156 @@
...
+- Calling `Box::into_raw` *consumes* the boxed value.
+    - In other words, the caller is responsible for making sure the memory managed by the `Box` instance that was turned to a raw pointer is taken care of.
+- Calling `Box::from_raw` is the opposite operation and consumes the raw pointer, creating a `Box` in the process, and freeing the allocated memory for the raw pointer passed to it.
+- Try not to move a pointer around after it's been boxed
+- Always make sure to have matching calls to `into_raw` and `from_raw`
+
+### Safety
+
+Raw pointers are always *unsafe* to use in Rust.
+This is why the separation of concerns is important, and you should always thrive
```
faern, line 151: s/thrive/strive ?

ios/FFI.md
```diff
@@ -0,0 +1,156 @@
...
+```rust
+#[repr(C)]
+pub struct LoaderWrapperContext {
+    // This pointer is a reference to a Swift object, and is only ever read by Rust.
+    // It is used to call that Swift object across the FFI
+    loader: *const c_void,
+}
+
+#[repr(C)]
+pub struct SwiftLoaderWrapper(LoaderWrapperContext);
```
faern, line 56: I do not understand why both `SwiftLoaderWrapper` and `LoaderWrapperContext` are needed. I don't see any reason why this could not just be:
```rust
pub struct SwiftLoaderWrapper(*const c_void);

impl SwiftLoaderWrapper {
    /// # Safety
    ///
    /// The pointer must be a pointer to a valid SwiftFooBarThing. The pointer must be valid
    /// for the entire lifetime of the returned `SwiftLoaderWrapper`
    pub unsafe fn new(context: *const c_void) -> SwiftLoaderWrapper {
        SwiftLoaderWrapper(context)
    }

    pub fn load_things() {
        unsafe { use_loader(self.0) }
    }
}
```

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8115)* status: all files reviewed, 4 unresolved discussions (waiting on @buggmagnet and @faern)

ios/FFI.md
buggmagnet: > I do not understand why both `SwiftLoaderWrapper` and `LoaderWrapperContext` are needed.

Technically they are not. However, the point is explained after. 

> This allows to have a clean separation of the \`unsafe\` FFI boundary with a safe API in Rust

We are essentially "hiding" the type that contains the raw C pointer so that the Rust side of the code does not need to think about it. 
There are multiple reasons why we do this :
- A type that contains a `c_void` cannot be automatically `Send` or `Sync`.
- Every call that will touch the raw pointer will likely have to be wrapped in `unsafe` which is cumbersome to write, by creating this layer, we can easily provide a `safe` API which we can also guarantee will work in multi threaded contexts since the wrapper can add thread safety to the context object
- The "context" object might be doing FFI calls as well, so we get to keep all the FFI functionality in one single file. 


> I think it would be nice to define opaque types in the C header as simply `struct S;`
The C header is entirely generated by `cbindgen`. I've looked at the source code since I'm not super happy with the generated interfaces, we can do more advanced declarations when it comes to Swift / ObjC, but that would require some `cbindgen` modifications.

However technically, we can use a `#[repr(C)]` `struct` object in a rust method definition that we expose to the FFI and that _should_ work. As a matter of fact, we already do that in a couple of cases.

ios/FFI.md
buggmagnet: > Here, `slice` has the type `&[*mut A]`. Is this intended?

Yes, but I realise that the _why_ is not obvious at all. 
This is meant to provide an example where we'd have to pass an array of objects created in Rust across the FFI.
In which case we'd need to send an array of _pointers_ to Rust objects.

I will rewrite this to mention that this is a special case, and provide a more generic example when it's just an array of scalar types.

> If you already have the objects you want to pass in contiguous memory in Swift,

Arrays are not necessarily contiguous in memory in Swift, depending how they are created. But that's a good sign that I should add a mention to that fact in this doc.

ios/FFI.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Boxing types
+
+`Box`es are a very convenient way to transport opaque data around.
+Chances are that you already have written some code that interacts with `Box` in one way or another.
+
+Things to keep in mind when working with Box:
+- Calling `Box::into_raw` *consumes* the boxed value.
+    - In other words, the caller is responsible for making sure the memory managed by the `Box` instance that was turned to a raw pointer is taken care of.
+- Calling `Box::from_raw` is the opposite operation and consumes the raw pointer, creating a `Box` in the process, and freeing the allocated memory for the raw pointer passed to it.
```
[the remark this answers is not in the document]
buggmagnet: > `Box::from_raw` does not free the allocated memory.

I actually stole that part from the `from_raw` documentation itself. 

>  Specifically, the Box destructor will call the destructor of T and free the allocated memory.
> `let _ = Box::from_raw(my_ptr);`

I realise that this probably comes naturally to Rustaceans, but is definitely not something Swifties will think of immediately, we usually rely on ARC to do the  "correct thing" :tm:  on our behalf.

We  should improve this, which option would you prefer ? 
- We explicitly add a line to mention that one would have to write `let _ = Box::from_raw(my_ptr);` if one wants to actually free the memory immediately
- We remove wording from "and freeing the allocated...."
- We mention that `my_ptr` will be freed _only_ when the created `Box` type is dropped (by going out of scope for example) and give an example of how that would look like

ios/FFI.md
buggmagnet: Oops, yes !

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8115)* status: all files reviewed, 4 unresolved discussions (waiting on @faern)

rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8115)* status: all files reviewed, 4 unresolved discussions (waiting on @faern)

faern (changes requested):

buggmagnet:
We decided to cancel this piece of work as we already tag people from the desktop team in any new Rust code that we write, and talking to them always beats referring to this piece of documentation when it comes to best practices.

PR #8128, opened 2025-05-06
Fix animations after app upgrade UI refactor

olmoh:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8128)


tobias-jarvelov (approved):
This PR does what it says on the tin, approved. However as we discussed previously we should revisit this when we have standardized our animations/transitions.

PR #8125, opened 2025-05-06
Use local DNS resolver on macOS

MarkusPettersson98:
This PR enables the local DNS resolver (which is already running on some macOS versions) always.

# Remaining work
- [x] Check for collisions for the ip+port we try to bind the local DNS resolver to
- [x] Clean up the `lo0` alias that is created when starting the local DNS resolver
- [x] Fall back to `127.0.0.1:53` in case all other loopback addresses/port combos would be unavailable

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8125)

PR #8134, opened 2025-05-07
Remove synthesized offline state

dlon:
This will usually no longer be needed when #8125 is merged, as the local resolver is always running.


dlon:
This is likely very stale, so I'm closing it.

PR #8136, opened 2025-05-08
Update e2e tests for auto-connect and lan sharing

olmoh:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8136)


desktop/packages/mullvad-vpn/test/e2e/installed/state-dependent/vpn-settings/vpn-settings.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+      }
+    };
+
+    const getAutoStartPath = () => {
+      return path.join(os.homedir(), '.config', 'autostart', 'mullvad-vpn.desktop');
+    };
+
+    const autoStartPathExists = () => {
+      return fileExists(getAutoStartPath());
+    };
```
tobias-jarvelov, line 83: **suggestion:** We could move these helper utilities into the `helpers.ts` file and expose them using a `createHelpers` function, like we did in the `app-upgrade` integration test.

tobias-jarvelov (approved):

desktop/packages/mullvad-vpn/test/e2e/installed/state-dependent/vpn-settings/vpn-settings.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    test.afterEach(async () => {
+      const autoConnectSwitch = await routes.vpnSettings.setAutoConnectSwitch(false);
+      await expect(autoConnectSwitch).toHaveAttribute('aria-checked', 'false');
+
+      if (process.platform === 'linux') {
+        expect(autoStartPathExists()).toBeFalsy();
+      }
+
+      const launchAppOnStartupSwitch = await routes.vpnSettings.setLaunchAppOnStartupSwitch(false);
+      await expect(launchAppOnStartupSwitch).toHaveAttribute('aria-checked', 'false');
```
tobias-jarvelov, line 43: **issue:** I think we should avoid having the `setLaunchAppOnStartupSwitch` return the selector. I know that we use the selector directly afterwards here in the test, but one of the benefits of the Page object model is that the test does not have to know or care what selector is used.

**suggestion:** I think we can fix this by introducing a new function on the `vpnSettings` route object model which can then be used like this:

```js
// vpn-settings.spec.ts
const launchAppOnStartupSwitch = await routes.vpnSettings.setLaunchAppOnStartupSwitch(false);
const enabled = await routes.vpnSettings.getLaunchAppOnStartupSwitchEnabled();
await expect(enabled).toBe(false);
```

```js
// vpn-settings-route-object-model.ts
getLaunchAppOnStartupSwitchValue(): boolean {
  const ariaChecked = lanSwitch.getAttribute('aria-checked');

  return ariaChecked === 'true';
}
```

desktop/packages/mullvad-vpn/test/e2e/route-object-models/main/main-route-object-model.ts
```diff
@@ -1,15 +1,15 @@
...
 import { createSelectors } from './selectors';
 
 export class MainRouteObjectModel {
   readonly page: Page;
-  readonly utils: MockedTestUtils;
+  readonly utils: TestUtils;
   readonly selectors: ReturnType<typeof createSelectors>;
 
-  constructor(page: Page, util: MockedTestUtils) {
+  constructor(page: Page, util: TestUtils) {
```
tobias-jarvelov, line 12: **note:** Oh nuts. I wonder how we will solve this. In the mocked test we want access to the `MockedTestUtils` but in the state dependent tests we only want access to the `TestUtils`. Otherwise we won't be able to mock IPC handles/responses in the mocked tests, and that means that individual tests will have to set it up. Maybe that's okay and we can find something which works but I think we need to look at this together and discuss our options.

olmoh: I tried this out a bit and one way that I think would be pretty clean is to first make the route models generic:
`VpnSettingsRouteObjectModel<T extends TestUtils>`

Then we could implement a type guard function:
```
  isMocked(
    this: VpnSettingsRouteObjectModel<TestUtils>,
  ): this is VpnSettingsRouteObjectModel<MockedTestUtils> {
    return typeof (this.utils as MockedTestUtils).mockIpcHandle === 'function';
  }
  ```
  
  Which we can then use in functions to run mocked specific functionality:
  ```
    async setLanSwitch(enabled: boolean) {
    const lanSwitch = this.getLanSwitch();
    const ariaChecked = await lanSwitch.getAttribute('aria-checked');

    if ((enabled && ariaChecked === 'false') || (!enabled && ariaChecked === 'true')) {
      await lanSwitch.click();
    }

    if (this.isMocked()) {
      await this.utils.mockIpcHandle(...));
    }
  }
```

This gives us the ability to to specific functionality based on if we have a mocked or regular test utils, and we don't have to provide anything "extra" to the route object model.

desktop/packages/mullvad-vpn/test/e2e/route-object-models/vpn-settings/vpn-settings-route-object-model.ts
```diff
@@ -0,0 +1,70 @@
...
+    const lanSwitch = this.getLanSwitch();
+    if (enabled) {
+      if ((await lanSwitch.getAttribute('aria-checked')) === 'false') {
+        await lanSwitch.click();
+      }
+    } else {
+      if ((await lanSwitch.getAttribute('aria-checked')) === 'true') {
+        await lanSwitch.click();
+      }
+    }
```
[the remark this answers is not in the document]
olmoh: Good suggestion, looks way cleaner. Chose the first version :)

PR #8142, opened 2025-05-08
Increase app response time when reconnecting

buggmagnet:
This PR makes the app look more responsive when reconnecting, especially when using Post Quantum keys.
It does so by doing the following: 
- Forcing the reconnect operation to give a status update from the tunnel
- Lower the time interval to poll the tunnel from every 3 seconds to every second when connecting, or reconnecting
- Fix a bug that did prevent the timer to switch to a lower value when the app was started with the tunnel already connected
- Make the simulator being able to fake reconnection, and show it on screen, for 600 milliseconds
- Due to the last fix, this created a crash in Unit Tests where the `SettingsManager.unitTestStore` would be read after the tests are ran, instead, the store is reset and never set to nil to avoid said crash
- `SimulatorTunnelProviderHost` can now introspect when receiving app messages via `onHandleProviderMessage` for advanced testing capabilities

## How to test

You can either comment the extra call to `refreshTunnelStatus()` when restarting the tunnel, or increase the `establishingTunnelStatusPollInterval` to a higher value.

With both fixes, the app feels way more responsive, especially when doing PQ exchange on countries that are far away from you. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142)

### DIFF
--- a/ios/CHANGELOG.md
+++ b/ios/CHANGELOG.md
@@ -27,6 +27,7 @@ Line wrap the file at 100 chars.                                              Th
 
 ### Changed
 - Improve the filter view to display the number of available servers based on selected criteria.
+- Make the app feel more responsive when reconnecting.
 
 ## 2025.2 - 2025-02-08
 ### Added
--- a/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift
+++ b/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift
@@ -26,6 +26,8 @@ final class SimulatorTunnelProviderHost: SimulatorTunnelProviderDelegate, @unche
     private let providerLogger = Logger(label: "SimulatorTunnelProviderHost")
     private let dispatchQueue = DispatchQueue(label: "SimulatorTunnelProviderHostQueue")
 
+    public var onHandleProviderMessage: ((TunnelProviderMessage) -> Void)?
+
     init(
         relaySelector: RelaySelectorProtocol,
         transportProvider: TransportProvider,
@@ -110,6 +112,7 @@ final class SimulatorTunnelProviderHost: SimulatorTunnelProviderDelegate, @unche
         }
     }
 
+    // swiftlint:disable:next function_body_length
     private func handleProviderMessage(
         _ message: TunnelProviderMessage,
         completionHandler: ((Data?) -> Void)?
@@ -143,10 +146,16 @@ final class SimulatorTunnelProviderHost: SimulatorTunnelProviderDelegate, @unche
                 break
             }
 
-            setInternalStateConnected(with: selectedRelays)
+            setInternalStateReconnecting(with: selectedRelays)
             reasserting = false
 
             completionHandler?(nil)
+
+            // The PacketTunnel does not run on the simulator.
+            // Fake a reconnecting state that becomes connected after long enough for the UI to change appropriately.
+            dispatchQueue.asyncAfter(deadline: .now() + .milliseconds(600)) { [weak self] in
+                self?.setInternalStateConnected(with: self?.selectedRelays)
+            }
 
         case let .sendURLRequest(proxyRequest):
             urlRequestProxy.sendRequest(proxyRequest) { response in
@@ -189,6 +198,8 @@ final class SimulatorTunnelProviderHost: SimulatorTunnelProviderDelegate, @unche
         case .privateKeyRotation:
             completionHandler?(nil)
         }
+
+        onHandleProviderMessage?(message)
     }
 
     private func pickRelays() throws -> SelectedRelays {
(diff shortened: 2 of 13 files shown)


rablador (approved):
> Reviewed 12 of 12 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

mojganii (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142)* status: all files reviewed, 3 unresolved discussions

___
*[`ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift` line 154 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142#-OQ3KRuiEO8NS5DR00HG:-OQ3KRuiEO8NS5DR00HH:b-98pmmr) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fe2d6ad45539246a42e0f414762d0a56832b4b/ios/MullvadVPN/SimulatorTunnelProvider/SimulatorTunnelProviderHost.swift#L154)):*
> ```Swift
>             completionHandler?(nil)
> 
>             dispatchQueue.asyncAfter(deadline: .now() + .milliseconds(600)) { [weak self] in
> ```

I highly recommend to add a comment which indicates the fake `connected` state is created after `600 ms` on simulator.
___
*[`ios/MullvadVPNTests/MullvadVPN/TunnelManager/TunnelManagerTests.swift` line 205 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142#-OQ3YuAC86sD6U1yn-NH:-OQ3YuAC86sD6U1yn-NI:b-8399ha) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fe2d6ad45539246a42e0f414762d0a56832b4b/ios/MullvadVPNTests/MullvadVPN/TunnelManager/TunnelManagerTests.swift#L205)):*
> ```Swift
> 
>     /// This test verifies that a refresh tunnel status operation is scheduled whenever the tunnel is being restarted
>     func testReconnectingTunnelRefreshesItsStatus() async throws {
> ```

`Swiftlint` complains about function body length. please, add `// swiftlint:disable:next function\_body\_length` to it
___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 298 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142#-OQ3W-GUC6DzAD95VALH:-OQ3W-GUC6DzAD95VALI:b-9m972y) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/a9fe2d6ad45539246a42e0f414762d0a56832b4b/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L298)):*
> ```Swift
> 
>                 return tunnel.reconnectTunnel(to: selectNewRelay ? .random : .current) { result in
>                     if case let .success(observedState) = result {
> ```

In `TunnelStatus`, we are tracking state using two different properties — one related to the `TunnelProvider` and the other used for state management in the main app process. Ideally, we should handle states similarly to how it's done in `ObservedState`, to reduce the chance of future errors. However, since that's not feasible in the short term, I suggest we at least add clearer comments for the `state` property.



_Code snippet:_
```Swift
/// Internal tunnel state used by the main app to manage transitions and UI updates.
/// Unlike `observedState`, which reflects the actual status reported by the tunnel process,
/// this property tracks the app's current understanding or expectation of the tunnel state.
var state: TunnelState = .disconnected
```

rablador (dismissed):
> Reviewed 4 of 4 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142)* status: all files reviewed, 4 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/TunnelManager/TunnelState.swift` line 28 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142#-OQ7dPXn6xqXMrCGToiY:-OQ7dPXn6xqXMrCGToiZ:bvwwtah) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/bd4a39552d48134cae25e7434077a69f052227b4/ios/MullvadVPN/TunnelManager/TunnelState.swift#L28)):*
> ```Swift
> ///
> /// And so on, this is a non-exhaustive list.
> struct TunnelStatus: Equatable, CustomStringConvertible, Sendable {
> ```

Nice explanations!

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142)* status: all files reviewed, 5 unresolved discussions (waiting on @mojganii)

___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 309 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142#-OQ8UJC1AoHRkyWQYXfI:-OQ8UJC1AoHRkyWQYXfJ:b-a9ncez) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c012bfec194493187de8326ef4471e73974799d/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L309)):*
> ```Swift
>                         )
>                         self._tunnelStatus.observedState = observedState
>                         DispatchQueue.main.async {
> ```

This looks very jarring in non PQ connections, we can do better

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142)* status: 11 of 13 files reviewed, 5 unresolved discussions (waiting on @mojganii and @rablador)

___
*[`ios/MullvadVPN/TunnelManager/TunnelManager.swift` line 309 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142#-OQ8UJC1AoHRkyWQYXfI:-OQ8eBSX6jugSbpYHAZI:b-dxe8g2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4c012bfec194493187de8326ef4471e73974799d/ios/MullvadVPN/TunnelManager/TunnelManager.swift#L309)):*

This looks very jarring in non PQ connections, we can do better
</blockquote></details>

We decided to remove the early dispatch as it was causing jankiness in normal connections where the app would flutter between connecting and reconnecting states twice in a row on fast connections.

Removing this makes the app _slightly_ less responsive (takes about 200ms before registering the change in the UI) but it doesn't look janky, which is still a good upgrade over what we used to have.

mojganii (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8142#-:-OQ8psoU0oKT0rFoMz4K:bnfp4nl)* status: 11 of 13 files reviewed, 3 unresolved discussions (waiting on @buggmagnet and @rablador)

PR #8167, opened 2025-05-13
Add error handling for when verifiedInstallerPath does not exist

tobias-jarvelov:
Checks that the `verifiedInstallerPath` exists before starting the installer. If the path does not exist an `AppUpgrade` error is emitted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8167)


desktop/packages/mullvad-vpn/src/renderer/components/views/app-upgrade/components/download-progress/hooks/useMessage/useMessage.ts
```diff
@@ -15,7 +15,7 @@ export const useMessage = () => {
   const hasAppUpgradeError = useHasAppUpgradeError();
   const hasAppUpgradeVerifiedInstallerPath = useHasAppUpgradeVerifiedInstallerPath();
 
-  if (hasAppUpgradeVerifiedInstallerPath) {
+  if (hasAppUpgradeVerifiedInstallerPath && !appUpgradeEventType) {
     return messages.pgettext('app-upgrade-view', 'Download complete!');
   }
```
olmoh, line 21: Commenting on this line since it is the closes to related code. When checking if we want to show the "Download paused" label, we could check if we are in one of the steps after the installer has been downloaded (verified, starting, etc.), to stop us from displaying the download paused label when no download is ongoing.

olmoh (changes requested):
We currently get an incorrect state when following these steps:

1. Start download
2. Block connection
3. Click retry
4. Will show error view, but should show connection blocked view

The connection blocked view is visible briefly but is replaced by the error view.

olmoh (approved):
Great job, lgtm!

PR #8179, opened 2025-05-14
Make the API cancellation handle safe, and properly clear out pointers

buggmagnet:
This PR modifies the signature of `mullvad_api_cancel_task` and `mullvad_api_cancel_task_drop`
to make them take a reference to a `SwiftCancelHandle`

We need to do this in order to avoid passing a copy of the struct to the FFI which would not clear out correctly the pointer contained in it, hence creating double free issues when cancelling and dropping immediately after a `SwiftCancelHandle`.

Big thanks to @hulthe for helping out on this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8179)

### DIFF
--- a/ios/MullvadRustRuntime/MullvadApiCancellable.swift
+++ b/ios/MullvadRustRuntime/MullvadApiCancellable.swift
@@ -9,17 +9,17 @@
 import MullvadTypes
 
 public class MullvadApiCancellable: Cancellable {
-    private let handle: SwiftCancelHandle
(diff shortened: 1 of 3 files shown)


rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8179)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #8202, opened 2025-05-16
Fix API breaking down when used without connectivity

buggmagnet:
This PR adds a reference to the `AddressCache` in the `MullvadApiContext` object.
This fixes an issue where the tokio runtime would panic in the past on trying to unwrap an Optional value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8202)

### DIFF
--- a/ios/MullvadREST/ApiHandlers/AddressCache.swift
+++ b/ios/MullvadREST/ApiHandlers/AddressCache.swift
@@ -11,7 +11,7 @@ import MullvadLogging
 import MullvadTypes
 
 extension REST {
-    public final class AddressCache: @unchecked Sendable {
+    public final class AddressCache: AddressCacheProviding, @unchecked Sendable {
         /// Logger.
         private let logger = Logger(label: "AddressCache")
 
--- a/ios/MullvadRESTTests/MullvadApiTests.swift
+++ b/ios/MullvadRESTTests/MullvadApiTests.swift
@@ -18,6 +18,7 @@ import XCTest
 
 class MullvadApiTests: XCTestCase {
     let encoder = JSONEncoder()
+    let addressCache = REST.AddressCache(canWriteToCache: false, fileCache: MemoryCache())
 
     func makeApiProxy(port: UInt16) throws -> APIQuerying {
         let shadowsocksLoader = ShadowsocksLoaderStub(configuration: ShadowsocksConfiguration(
@@ -38,7 +39,8 @@ class MullvadApiTests: XCTestCase {
             accessMethodWrapper: initAccessMethodSettingsWrapper(
                 methods: accessMethodsRepository
                     .fetchAll()
-            )
+            ),
+            addressCacheProvider: addressCache
         )
 
         let proxy = REST.MullvadAPIProxy(
--- a/ios/MullvadRustRuntime/MullvadAddressCacheProvider.swift
+++ b/ios/MullvadRustRuntime/MullvadAddressCacheProvider.swift
@@ -1,0 +1,31 @@
+//
+//  MullvadAddressCacheProvider.swift
+//  MullvadRustRuntime
+//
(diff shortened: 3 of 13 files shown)


acb-mv (commented):
> Reviewed 3 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8202)* status: 3 of 13 files reviewed, 2 unresolved discussions

___
*[`ios/MullvadTypes/AddressCacheProviding.swift` line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8202#-OQQ8aPH9MPvROnQMW-X:-OQQ8aPH9MPvROnQMW-Y:bsdht8l) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f78742619aade81c8497daba8de4bf8caa70f62/ios/MullvadTypes/AddressCacheProviding.swift#L15)):*
> ```Swift
> }
> 
> public final class DefaultAddressCacheProvider: AddressCacheProviding, Sendable {
> ```

What is the purpose of this class, which appears to just wrap an `AddressCacheProviding` without changing its behaviour? The name doesn't really clarify it.
___
*[`ios/MullvadRustRuntime/MullvadAddressCacheProvider.swift` line 12 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8202#-OQQ7lWA519mctSKws5P:-OQQ7lWA519mctSKws5Q:b-5i6ken) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f78742619aade81c8497daba8de4bf8caa70f62/ios/MullvadRustRuntime/MullvadAddressCacheProvider.swift#L12)):*
> ```Swift
> import MullvadTypes
> 
> public func iniSwiftAddressCacheWrapper(provider: DefaultAddressCacheProvider) -> SwiftAddressCacheWrapper {
> ```

is `iniSwiftAddressCacheWrapper` meant to be `initSwiftAddressCacheWrapper`?

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8202)* status: 3 of 13 files reviewed, 2 unresolved discussions (waiting on @acb-mv)

___
*[`ios/MullvadTypes/AddressCacheProviding.swift` line 15 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8202#-OQQ8aPH9MPvROnQMW-X:-OQbJOtD6lA3nAPjYGOu:b-dwx07w) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/4f78742619aade81c8497daba8de4bf8caa70f62/ios/MullvadTypes/AddressCacheProviding.swift#L15)):*

What is the purpose of this class, which appears to just wrap an `AddressCacheProviding` without changing its behaviour? The name doesn't really clarify it.
</blockquote></details>

The sole purpose of this is to get around circular dependencies. 
My original intent was to move `AddressCache` from `MullvadREST` to `MullvadTypes`.
However, that creates a circular dependency between `MullvadLogging`, which `AddressCache` consumes, and `MullvadTypes` which is consumed by `MullvadLogging` (Specifically because of the `CustomErrorDescriptionProtocol`) 

And I didn't want this to be a big refactoring.

We used the same trick for the `SwiftShadowsocksBridgeProviding` dependency,  it's just a proxy class that forwards the method to the actual class at runtime but which lives in a different module.

I'm up for alternative names, you're right that it's not very descriptive.

rablador (approved):
> Reviewed 12 of 13 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8202)* status: all files reviewed, 2 unresolved discussions (waiting on @acb-mv and @buggmagnet)

PR #8223, opened 2025-05-20
Fix arg order lost when adding to messages.pot

kl:
Currently when we add new strings from strings.xml in the android app and these gets added to message.pot by the translation script the arguments get added only as "%s" and "%d" and any ordering is lost. This means that if you would add a string to the android app that is something like "%1$s %2$s %1$s" this would be added to messages.pot as msgid "%s %s %s" and it would treat it as the string has 3 different arguments instead of 2.

This PR changes the code so that when the Android strings.xml strings are added to messages.pot the argument index (if it exists) is not removed as it was previously.

So the Android string "%1$s %2$s %1$s" will now result in the exact same string being added to messages.pot.

However, the current normalizaion behavior is kept when it comes to checking if an Android string already exists in messages.pot, so e.g. the string `"Not all our servers are %1$s-enabled. Therefore, we use multihop automatically to enable %1$s with any server."` will still match the current value in the pot file which is: `"Not all our servers are %(daita)s-enabled. Therefore, we use multihop automatically to enable %(daita)s with any server."` and no new string is added to messages.pot.

### DIFF
--- a/android/translations-converter/src/android/plurals.rs
+++ b/android/translations-converter/src/android/plurals.rs
@@ -1,5 +1,8 @@
 use super::string_value::StringValue;
 use serde::{Deserialize, Serialize};
+use std::fs::File;
+use std::io::BufReader;
+use std::path::Path;
 use std::{
     fmt::{self, Display, Formatter},
     ops::{Deref, DerefMut},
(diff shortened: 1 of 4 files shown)


Pururun (dismissed):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8223)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Pururun (approved):
We should just create a linear issue so we can fix the changes strings on crowdin after the translation round is done, otherwise great work :star: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8223#-:-OQmBcZ7Bh6CJWNKsYV2:b-fu57sf)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #8256, opened 2025-05-26
Refactor Link component for better Icon alignment when spanning multiple lines

olmoh:
Changes the Icon to align with the final line of the text in a link.


desktop/packages/mullvad-vpn/src/renderer/lib/components/link/components/LinkIcon.tsx
```diff
@@ -5,8 +5,9 @@ import { Icon, IconProps } from '../../icon';
 type LinkIconProps = IconProps;
 
 export const StyledIcon = styled(Icon)`
-  vertical-align: middle;
-  display: inline-flex;
+  vertical-align: text-bottom;
+  display: inline-block;
+  flex-shrink: 0;
```
tobias-jarvelov, line 10: **question:** It doesn't seem like `flex-shrink` should have any effect, given that the parent is not using flex, and neither is this component. Can we remove it?

tobias-jarvelov (approved):
LGTM

PR #8266, opened 2025-05-28
Add `mullvad tunnel set wireguard --allowed-ips`

Serock3:

### DIFF
--- a/mullvad-cli/src/cmds/tunnel.rs
+++ b/mullvad-cli/src/cmds/tunnel.rs
@@ -3,6 +3,7 @@ use clap::Subcommand;
 use mullvad_management_interface::MullvadProxyClient;
 use mullvad_types::{
     constraints::Constraint,
+    relay_constraints::{RelaySettings, WireguardConstraints},
     wireguard::{QuantumResistantState, RotationInterval, DEFAULT_ROTATION_INTERVAL},
 };
 
@@ -44,6 +45,10 @@ pub enum TunnelOptions {
         /// Configure whether to enable DAITA direct only
         #[arg(long)]
         daita_direct_only: Option<BooleanOption>,
+        /// Specify custom allowed IPs for WireGuard tunnels. Use comma-separated values (e.g., "10.0.0.0/8,192.168.0.0/16")
+        /// Default is all traffic (0.0.0.0/0,::/0)
+        #[arg(long)]
+        allowed_ips: Option<String>,
         /// The key rotation interval. Number of hours, or 'any'
         #[arg(long)]
         rotation_interval: Option<Constraint<RotationInterval>>,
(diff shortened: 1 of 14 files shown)


mullvad-cli/src/cmds/tunnel.rs
```diff
@@ -194,6 +225,19 @@ impl Tunnel {
             println!("Quantum resistant setting has been updated");
         }
 
+        if let Some(allowed_ips_str) = allowed_ips {
+            let ips: Vec<String> = if allowed_ips_str.is_empty() {
+                Vec::new()
+            } else {
+                allowed_ips_str
+                    .split(',')
```
dlon, line 233: Could this be done using a clap parser?

mullvad-daemon/src/lib.rs
```diff
@@ -290,6 +290,8 @@ pub enum DaemonCommand {
     /// Toggle macOS network check leak
     /// Set MTU for wireguard tunnels
     SetWireguardMtu(ResponseTx<(), settings::Error>, Option<u16>),
+    /// Set allowed IPs for wireguard tunnels
+    SetWireguardAllowedIps(ResponseTx<(), settings::Error>, Vec<String>),
```
dlon, line 294: This should probably be a vec of `IpAddr`s

mullvad-types/src/constraints/mod.rs
```diff
@@ -142,3 +142,4 @@ impl_intersection_partialeq!(relay_constraints::Ownership);
 impl_intersection_partialeq!(talpid_types::net::TransportProtocol);
 impl_intersection_partialeq!(talpid_types::net::TunnelType);
 impl_intersection_partialeq!(talpid_types::net::IpVersion);
+impl_intersection_partialeq!(Vec<ipnetwork::IpNetwork>);
```
dlon, line 145: Should we add a note that this doesn't implement intersection for all overlapping networks?

Serock3: Yes. I added a note on all usages of the macro.

mullvad-cli/src/cmds/tunnel.rs
Serock3: Yes

Serock3: I tried but there was no way to have a sane default value. Now I use an empty string to reset to 0.0.0.0/0

Serock3: To elaborate: The IP range is represented as a string in protobuf, which we parse and validate in the daemon so you will still get a proper error message (although it will contain "gRPC error").

If we really want to also parse it directly in the CLI, we must have another way to reset to the default, else user must run the command with `--allowed-ips 0.0.0.0/0`. Perhaps another flag `--reset-allowed-ips` could be an alternative, but how would this interact if both flags are set?

mullvad-types/src/relay_constraints.rs
```diff
@@ -407,10 +408,42 @@ impl fmt::Display for OpenVpnConstraints {
...
 }
 
+/// Returns a vector of IP networks representing all of the internet, 0.0.0.0/0.
+/// This may be used in [`crate::net::wireguard::PeerConfig`] to route all traffic
+/// to the tunnel interface.
+pub fn resolve_allowed_ips(
+    allowed_ips: Constraint<Vec<ipnetwork::IpNetwork>>,
+) -> Vec<ipnetwork::IpNetwork> {
+    let all_of_the_internet = vec![
+        "0.0.0.0/0".parse().expect("Failed to parse ipv6 network"),
```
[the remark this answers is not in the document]
Serock3: Done.

PR #8268, opened 2025-05-28
Remove unnecessary fuse

pronebird:
Hi,

I am not entirely sure but I think this `fuse()` is unnecessary given that `None` is being handled in the loop. Let's see if CI can compile that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8268)

### DIFF
--- a/talpid-routing/src/unix/macos/mod.rs
+++ b/talpid-routing/src/unix/macos/mod.rs
@@ -161,9 +161,7 @@ impl RouteManagerImpl {
         })
     }
 
-    pub(crate) async fn run(mut self, manage_rx: mpsc::UnboundedReceiver<RouteManagerCommand>) {
-        let mut manage_rx = manage_rx.fuse();
-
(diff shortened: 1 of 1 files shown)


MarkusPettersson98 (commented):
The `select!` macro we're using requires the future to implement `FusedFuture`, which `futures::channel::mpsc::UnboundedReceiver` does not do. We invoke `fuse()` to get a `Fuse<UnboundedReceiver>`, which does implement `FusedFuture`.

* https://docs.rs/futures/latest/futures/macro.select.html
* https://docs.rs/futures/latest/futures/future/trait.FusedFuture.html#impl-FusedFuture-for-Fuse%3CFut%3E

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8268#-:-OT7Ungj3uoDeIa6vXwj:b7449yg)* status: 0 of 1 files reviewed, all discussions resolved

PR #8278, opened 2025-05-30
Fix boringtun device not being closed when disconnecting

dlon:
This fixes the issue where reconnecting fails every time, because not all spawned tasks were aborted.

Close DES-2161

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8278)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -444,7 +444,7 @@ dependencies = [
 [[package]]
 name = "boringtun"
 version = "0.6.0"
-source = "git+https://github.com/mullvad/boringtun?rev=a7e11fb46d4a#a7e11fb46d4ae65d4c7bf4efb9617548c072afa0"
+source = "git+https://github.com/mullvad/boringtun?rev=c29681df54e1c10ba06d9ca7ff98722c9dcb5a81#c29681df54e1c10ba06d9ca7ff98722c9dcb5a81"
 dependencies = [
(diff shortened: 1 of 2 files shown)


Serock3 (approved):
:raised_hands: :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8278#-:-ORb-c-xCiYDHJd8pgyi:b-36sk3o)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Serock3 (approved):
Great work! Here's a thought though from looking at the commit in the boringtun fork: Could you not achieve the same thing by selecting on the async blocks directly instead of spawning them and wrapping in `AbortOnDrop`? I guess the drawback would be that the futures could not run in parallel, which may impact performance.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8278#-:-ORb-rIT-oTnRV8nlpwr:bmsb6lj)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

dlon:
> I guess the drawback would be that the futures could not run in parallel, which may impact performance.

Yep, that is why I didn't do this. I don't think it makes the code *much* simpler either, so I just went with the (likely) fastest code.

PR #8307, opened 2025-06-05
Localise access method names on creation

acb-mv:
This localises the one thing (other than changelog) found that was unlocalised: access method names. 
Note that currently these are generated in the code and persisted to device storage, meaning that, when we have multiple languages, they will be fixed in the language the user installed the app whilst using even if they change language. That should perhaps be fixed before we add languages.

### DIFF
--- a/ios/MullvadSettings/AccessMethodRepository.swift
+++ b/ios/MullvadSettings/AccessMethodRepository.swift
@@ -18,6 +18,8 @@ public class AccessMethodRepository: AccessMethodRepositoryProtocol, @unchecked 
 
     private let logger = Logger(label: "AccessMethodRepository")
 
+    // The access method names will be localised on creation time. As they are persisted
+    // to on-device storage, they will not be relocalised if the user changes language.
     private let direct = PersistentAccessMethod(
         id: AccessMethodRepository.directId,
         name: NSLocalizedString(
--- a/ios/MullvadSettings/AccessMethodRepository.swift
+++ b/ios/MullvadSettings/AccessMethodRepository.swift
@@ -20,21 +20,36 @@ public class AccessMethodRepository: AccessMethodRepositoryProtocol, @unchecked 
 
     private let direct = PersistentAccessMethod(
         id: AccessMethodRepository.directId,
-        name: "Direct",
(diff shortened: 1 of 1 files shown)


buggmagnet (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8307)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`ios/MullvadSettings/AccessMethodRepository.swift` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8307#-OSIz--f1PLGF11jJIAm:-OSIz--f1PLGF11jJIAn:brzr11m) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5f2f2c2cf2403dc087009c2619204fc76b0fac19/ios/MullvadSettings/AccessMethodRepository.swift#L26)):*
> ```Swift
>         id: AccessMethodRepository.directId,
>         name: NSLocalizedString(
>             "ACCESS_METHOD_NAME_DIRECT",
> ```

In that case, shouldn't we instead save the raw string and only localize it when it's displayed ?

acb-mv (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8307)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @buggmagnet and @rablador)

___
*[`ios/MullvadSettings/AccessMethodRepository.swift` line 26 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8307#-OSIz--f1PLGF11jJIAm:-OSU07VKDiWdhaNXqwzy:b5fy393) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/5f2f2c2cf2403dc087009c2619204fc76b0fac19/ios/MullvadSettings/AccessMethodRepository.swift#L26)):*

That could be a solution.
</blockquote></details>

I've changed it to do the localisation in the UI layer. The main downside is that the keys aren't as neat (they're now things like `"ACCESS_METHOD_NAME:Mullvad bridges", with the natural-language name appended`), though that should be workable

PR #8308, opened 2025-06-05
Fix typo

MrChocolatine:
This PR simply fixes a typo in the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ supported.
 
 #### Linux
 - The deb package repositores now have static codenames on top of the existing distro version
-  specific codenames. The stable repository always have the "stable" codename,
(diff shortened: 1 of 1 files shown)


Serock3 (dismissed):

Serock3:
Good catch! This needs to be rebased on main before it can be merged, though. Merging main into the `patch-1`-branch doesn't work unfortunately. I can do it for you, but then github will not link it the new hash to this PR so I'll have to close it.

MrChocolatine:
> Good catch! This needs to be rebased on main before it can be merged, though. Merging main into the `patch-1`-branch doesn't work unfortunately. I can do it for you, but then github will not link it the new hash to this PR so I'll have to close it.

@Serock3 , misclicks, I updated without rebase.
The branch should be okay now.

PR #8313, opened 2025-06-06
macOS: use `/opt/homebrew` for zsh completions if available

EricFromCanada:
Following how `fish` completions are installed in `/opt/homebrew`, this does the same for `zsh` completions by editing the macOS installer's postinstall and uninstall scripts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8313)


MarkusPettersson98:
Thanks!

PR #8338, opened 2025-06-15
Enable opt-in to MTE async mode

cameronaaron:
android:memtagMode="async"
to the <application> element in the AndroidManifest.xml.

Why this is wanted

Catch memory safety issues early. Asynchronous MTE allows the system to tag heap allocations and check pointer usage at runtime, helping detect buffer overflows and use-after-free bugs in native code without significant performance overhead.

Improve security. Mullvad relies on native libraries (e.g. for crypto). MTE hardens these components against memory-corruption exploits on supported devices.

Align with platform defaults. Newer Android versions (Android 13+) ship with MTE support; opting into async mode ensures we’re taking advantage of platform improvements.

How it’s implemented

Open app/src/main/AndroidManifest.xml.

In the <application> tag, add the attribute:

<application
    android:name=".MullvadApp"
+   android:memtagMode="async"
    …>
    …
</application>
No additional code changes are required—MTE is managed by the platform once enabled in the manifest.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8338)

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -27,6 +27,8 @@
     <application android:name=".MullvadApplication"
                  android:banner="@mipmap/ic_banner"
                  android:allowBackup="false"
+                 tools:targetApi="31"
+                 android:memtagMode="async"
                  android:fullBackupContent="@xml/full_backup_content"
                  android:dataExtractionRules="@xml/data_extraction_rules"
                  android:icon="@mipmap/ic_launcher"


cameronaaron:
this will also opt into MTE on GOS by default and some android 16 devices

cameronaaron:
I looks like the tests are upset because these apis where introduced in 31 and the current min target is 26 if you notice you do not have any users using 26 you might want to consider raising the min api version to take advantage of more modern security features. If not i'll see if I can find a better solution :) Please let me know!

albin-mullvad:
> I looks like the tests are upset because these apis where introduced in 31 and the current min target is 26 if you notice you do not have any users using 26 you might want to consider raising the min api version to take advantage of more modern security features. If not i'll see if I can find a better solution :) Please let me know!

I believe that's fine. The lint is only complaining about the attribute not having impact on lower API levels. I suggest adding the following as an `application` attribute:

```
// Used to suppress memtagMode not having impact on API levels below 31.
tools:targetApi="31"
```

albin-mullvad:
Regarding the build issue, we'll have a look at fixing that. We switched to a GH secret for our non-sensitive dummy debug keystore a while back, but that [doesn't work well with forks](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow). We do this to have a common keystore available across jobs. We'll probably end up adding a fallback where we generate a dummy keystore on-the-fly in case the secret isn't available. We had that behavior in the past so should be quite trivial to add back.

cameronaaron:
@albin-mullvad I think I made the correct changes I am a little bit of a noob with GIT so please let me know if I did something incorrectly but it should be 2 commits from you on top of their main HEAD now and be ready to merge once the broken tests pass

cc/ @matchboxbananasynergy

albin-mullvad (changes requested):
@cameronaaron thank you for addressing that! It now looks in line with our way of working with git :slightly_smiling_face:   

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8338#-:-OT0ks7c3Kk3JQZRnYkL:b-ixibi8)* status: all files reviewed, 1 unresolved discussion (waiting on @cameronaaron)

___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8338#-OSwpuBF19JQUQwzYMcC:-OT0kgJM9OPK9jBwtltA:b-u9lfr7):*

nit: We have a strict commit message style combined with automatic checks so I suggest a slight tweak to the commit message:
</blockquote></details>

It seems like you updated the PR, but not the commit message subject. Please have another look at this one :pray:

albin-mullvad (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8338)* status: all files reviewed, 1 unresolved discussion (waiting on @cameronaaron)

___
*[`-- commits` line 3 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8338#-OSwpuBF19JQUQwzYMcC:-OT0n4tnDgcyf-1TwB54:bb7u4kd):*

It seems like you updated the PR, but not the commit message subject. Please have another look at this one :pray:
</blockquote></details>

Nvm, I was a bit too quick. Looks like there are two commits. That would generally be OK for us, but now they are more or less named the same. To simplify and cleanup things I suggest squashing the two to a single commit with `Enable opt-in to MTE async mode` as subject.

cameronaaron:
Ok let me try to do this!

cameronaaron:
I think that did it :) Still trying to get better at using Git

albin-mullvad:
Build and tests passed in: https://github.com/mullvad/mullvadvpn-app/actions/runs/15726436524 :heavy_check_mark: 

@cameronaaron thank you for your contribution!

This improvement will likely be included in `2025.6`.

cameronaaron:
Thank you!

PR #8405, opened 2025-07-02
Do not persist blocking firewall rules across a reboot during upgrades

MarkusPettersson98:
This PR adds a third cleanup policy to WinFW that allow blocking rules to stay active until BFE restarts (e.g. on a system reboot) after WinFW has been deinitalized. This is useful to not hardlock a client from accessing the internet if an app upgrade where to fail before removing existing firewall rules, as they may simply reboot to reset firewall rules again. Note that lockdown mode is still respected. This change will only take effect is both lockdown mode and auto start is disabled.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8405)

### DIFF
--- a/talpid-core/src/tunnel_state_machine/mod.rs
+++ b/talpid-core/src/tunnel_state_machine/mod.rs
@@ -94,7 +94,7 @@ pub struct InitialTunnelState {
     pub allow_lan: bool,
     /// Block traffic unless connected to the VPN.
     #[cfg(not(target_os = "android"))]
-    pub block_when_disconnected: bool,
+    pub block_when_disconnected: BlockWhenDisconnected,
     /// DNS configuration to use
     pub dns_config: DnsConfig,
     /// A single endpoint that is allowed to communicate outside the tunnel, i.e.
@@ -200,7 +200,7 @@ pub enum TunnelCommand {
     Dns(crate::dns::DnsConfig, oneshot::Sender<()>),
     /// Enable or disable the block_when_disconnected feature.
     #[cfg(not(target_os = "android"))]
-    BlockWhenDisconnected(bool, oneshot::Sender<()>),
+    BlockWhenDisconnected(BlockWhenDisconnected, oneshot::Sender<()>),
     /// Notify the state machine of the connectivity of the device.
     Connectivity(Connectivity),
     /// Open tunnel connection.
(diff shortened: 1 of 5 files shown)


talpid-core/src/tunnel_state_machine/mod.rs
```diff
@@ -234,6 +234,80 @@ enum EventResult {
...
+/// Semantically equivalent to a boolean value, but is grouped togetether with the persist
+/// parameter on Windows for cohesiveness.
+#[derive(Clone, Copy, Debug)]
+pub struct BlockWhenDisconnected {
+    /// If firewall should apply blocking rules in the disconnected state.
+    enable: bool,
+    /// If blocked state should be persisted across a reboot (restart of BFE)
+    persist: bool,
+}
+
```
Serock3: Nit: This could be an enum
```rust
pub enum BlockWhenDisconnected {
    Disabled,
    Enabled {
        persist: bool
    }
}
```

MarkusPettersson98: Nice! Done :+1:

Serock3 (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8405)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Serock3 (dismissed):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8405)* status: all files reviewed, 1 unresolved discussion

___
*[`windows/winfw/src/winfw/winfw.cpp` line 238 at r7](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8405#-OUJdOOACktWy65wWl68:-OUJdOOACktWy65wWl69:b8cnnms) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/c063abe9ae07333ed663ccd714b49a3d480bfcb9/windows/winfw/src/winfw/winfw.cpp#L238)):*
> ```cpp
> 				// Keep non-persistent filters intact, the intent is just to *not*
> 				// persist any filters across a BFE restart, not muck around with
> 				// muck around with any other filters. We will apply blocking
> ```

Typo: Double "much around"

PR #8425, opened 2025-07-05
Add musl as a target ABI for `wireguard-go-rs`

MarkusPettersson98:
This PR updates the `wireguard-go-rs` build script to not assume target ABI to be glibc when cross-compiling. The current solution is not directly extensible by the user, but it easily could be if we really wanted to. At least we don't break cross-compilation to musl targets though.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8425)

### DIFF
--- a/wireguard-go-rs/build.rs
+++ b/wireguard-go-rs/build.rs
@@ -53,6 +53,14 @@ enum AndroidTarget {
     I686,    // "i686"
 }
 
+#[derive(PartialEq, Eq, Clone, Copy)]
+enum Libc {
+    /// glibc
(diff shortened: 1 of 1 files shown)


Serock3 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8425#-:-OUYjkUf3yUjQOoepV-2:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #8464, opened 2025-07-15
Propagate changes of access method from Rust to Swift

acb-mv:
This implements a mechanism for sending access method change notifications from the Rust API code to Swift and saving the current method. The data sent back is the UUID of the current method. 

It currently crashes due to an EXC_BAD_ACCESS in the Swift callback, which needs to be debugged. Also remaining is the removal of `TransportStrategy`.

### DIFF
--- a/ios/MullvadRustRuntime/MullvadApiContext.swift
+++ b/ios/MullvadRustRuntime/MullvadApiContext.swift
@@ -6,18 +6,28 @@
 //  Copyright © 2025 Mullvad VPN AB. All rights reserved.
 //
 
+import MullvadSettings
 import MullvadTypes
 
-public struct MullvadApiContext: @unchecked Sendable {
+func onAccessChangeCallback(selfPtr: UnsafeRawPointer?, bytes: UnsafePointer<UInt8>?) {
+    guard let selfPtr, let bytes else { return }
+    let context = selfPtr.assumingMemoryBound(to: MullvadApiContext.self).pointee
+
+    let uuid = NSUUID(uuidBytes: bytes) as UUID
+    context.accessMethodChangeListener?.accessMethodChangedTo(uuid)
+}
+
+public class MullvadApiContext: @unchecked Sendable {
     enum MullvadApiContextError: Error {
         case failedToConstructApiClient
     }
 
-    public let context: SwiftApiContext
+    public private(set) var context: SwiftApiContext!
     private let shadowsocksBridgeProvider: SwiftShadowsocksBridgeProviding!
     private let shadowsocksBridgeProviderWrapper: SwiftShadowsocksLoaderWrapper!
     private let addressCacheWrapper: SwiftAddressCacheWrapper!
     private let addressCacheProvider: AddressCacheProviding!
+    public var accessMethodChangeListener: MullvadAccessMethodChangeListening?
 
     public init(
         host: String,
@@ -36,6 +46,7 @@ public struct MullvadApiContext: @unchecked Sendable {
         self.addressCacheProvider = defaultAddressCache
         self.addressCacheWrapper = iniSwiftAddressCacheWrapper(provider: defaultAddressCache)
 
+        context = nil
         context = switch disableTls {
         case true:
             mullvad_api_init_new_tls_disabled(
@@ -44,7 +55,9 @@ public struct MullvadApiContext: @unchecked Sendable {
                 domain,
                 shadowsocksBridgeProviderWrapper,
                 accessMethodWrapper,
-                addressCacheWrapper
+                addressCacheWrapper,
+                onAccessChangeCallback,
+                Unmanaged.passRetained(self).toOpaque()
             )
         case false:
             mullvad_api_init_new(
@@ -53,7 +66,9 @@ public struct MullvadApiContext: @unchecked Sendable {
                 domain,
                 shadowsocksBridgeProviderWrapper,
                 accessMethodWrapper,
-                addressCacheWrapper
+                addressCacheWrapper,
+                onAccessChangeCallback,
+                Unmanaged.passRetained(self).toOpaque()
             )
         }
 
--- a/ios/MullvadRustRuntime/include/mullvad_rust_runtime.h
+++ b/ios/MullvadRustRuntime/include/mullvad_rust_runtime.h
@@ -169,7 +169,10 @@ struct SwiftApiContext mullvad_api_init_new_tls_disabled(const char *host,
                                                          const char *domain,
                                                          struct SwiftShadowsocksLoaderWrapper bridge_provider,
                                                          struct SwiftAccessMethodSettingsWrapper settings_provider,
-                                                         struct SwiftAddressCacheWrapper address_cache);
+                                                         struct SwiftAddressCacheWrapper address_cache,
+                                                         void (*access_method_change_callback)(const void*,
+                                                                                               const uint8_t*),
+                                                         const void *access_method_change_context);
 
 /**
  * # Safety
@@ -190,7 +193,10 @@ struct SwiftApiContext mullvad_api_init_new(const char *host,
                                             const char *domain,
                                             struct SwiftShadowsocksLoaderWrapper bridge_provider,
                                             struct SwiftAccessMethodSettingsWrapper settings_provider,
-                                            struct SwiftAddressCacheWrapper address_cache);
+                                            struct SwiftAddressCacheWrapper address_cache,
+                                            void (*access_method_change_callback)(const void*,
+                                                                                  const uint8_t*),
+                                            const void *access_method_change_context);
 
 /**
  * # Safety
@@ -212,7 +218,10 @@ struct SwiftApiContext mullvad_api_init_inner(const char *host,
                                               bool disable_tls,
                                               struct SwiftShadowsocksLoaderWrapper bridge_provider,
                                               struct SwiftAccessMethodSettingsWrapper settings_provider,
-                                              struct SwiftAddressCacheWrapper address_cache);
+                                              struct SwiftAddressCacheWrapper address_cache,
+                                              void (*access_method_change_callback)(const void*,
+                                                                                    const uint8_t*),
+                                              const void *access_method_change_context);
 
 /**
  * Converts parameters into a `Box<AccessMethodSetting>` raw representation that
--- a/ios/MullvadSettings/AccessMethodRepository.swift
+++ b/ios/MullvadSettings/AccessMethodRepository.swift
@@ -175,3 +175,13 @@ public class AccessMethodRepository: AccessMethodRepositoryProtocol, @unchecked 
         SettingsParser(decoder: JSONDecoder(), encoder: JSONEncoder())
     }
 }
+
+extension AccessMethodRepository: MullvadAccessMethodChangeListening {
+    public func accessMethodChangedTo(_ uuid: UUID) {
+        guard let method = accessMethodsSubject.value.first(where: { $0.id == uuid }) else {
+            logger.warning("Change reported to method with unknown ID: \(uuid)")
+            return
+        }
+        save(method)
+    }
+}
--- a/ios/MullvadSettings/MullvadAccessMethodChangeListening.swift
+++ b/ios/MullvadSettings/MullvadAccessMethodChangeListening.swift
@@ -1,0 +1,12 @@
+//
+//  MullvadAccessMethodChangeListening.swift
+//  MullvadVPN
+//
+//  Created by Andrew Bulhak on 2025-07-03.
+//  Copyright © 2025 Mullvad VPN AB. All rights reserved.
+//
+
+// A protocol that listens for notifications of when the current access method has changed. It receives only the UUID of the new method.
+public protocol MullvadAccessMethodChangeListening: AnyObject {
+    func accessMethodChangedTo(_ uuid: UUID)
+}
--- a/ios/MullvadVPN.xcodeproj/project.pbxproj
+++ b/ios/MullvadVPN.xcodeproj/project.pbxproj
@@ -53,6 +53,7 @@
 		4424CDD32CDBD4A6009D8C9F /* SingleChoiceList.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4424CDD22CDBD4A6009D8C9F /* SingleChoiceList.swift */; };
 		447F3D8A2CDE1853006E3462 /* ShadowsocksObfuscationSettingsViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 447F3D882CDE1852006E3462 /* ShadowsocksObfuscationSettingsViewModel.swift */; };
 		447F3D8B2CDE1853006E3462 /* ShadowsocksObfuscationSettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 447F3D892CDE1853006E3462 /* ShadowsocksObfuscationSettingsView.swift */; };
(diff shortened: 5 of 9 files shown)


MarkusPettersson98 (commented):
> Reviewed 1 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464)* status: 1 of 9 files reviewed, 4 unresolved discussions

___
*[`mullvad-ios/src/api_client/mod.rs` line 92 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OVHPCwg-UaV8pPo9oUF:-OVHPCwg-UaV8pPo9oUG:bcqpzbc) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f57a655ec9fc048b371520d0ebd590e6d1bb29e/mullvad-ios/src/api_client/mod.rs#L92)):*
> ```XML
> struct ForeignPtr {
>     ptr: *const c_void,
> }
> ```

Could we say something about the semantics of this pointer type? Is it fine to *always* dereference it? Is it fine if the inner pointer is `null`?

I suggest adding methods on the `ForeignPtr` type that clearly defines every possible action using this pointer, along with the appropriate safety checks.



_Code quote:_
```XML
struct ForeignPtr {
    ptr: *const c_void,
}
```
___
*[`mullvad-ios/src/api_client/mod.rs` line 95 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OVHNxTF05LsVJcOO2KR:-OVHNxTF05LsVJcOO2KS:b-nxygk4) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f57a655ec9fc048b371520d0ebd590e6d1bb29e/mullvad-ios/src/api_client/mod.rs#L95)):*
> ```XML
> /// allow this to be passed across thread boundaries
> unsafe impl Send for ForeignPtr {}
> unsafe impl Sync for ForeignPtr {}
> ```

I would avoid implementing `Sync` for this type, we don't need multiple threads sharing a reference to any `ForeignPtr`, right?



_Code quote:_
```XML
unsafe impl Sync for ForeignPtr {}
```
___
*[`mullvad-ios/src/api_client/mod.rs` line 285 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OVHOiDdBfuKXF07B2Gq:-OVHOiDdBfuKXF07B2Gr:bz430zz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f57a655ec9fc048b371520d0ebd590e6d1bb29e/mullvad-ios/src/api_client/mod.rs#L285)):*
> ```XML
>             let access_method_change_ctx = access_method_change_ctx;
>             // SAFETY: The callback is expected to be called from the Swift side
>             if let Some(callback) = access_method_change_callback {
> ```

Should we log if there is no registered callback? Seems like a confusing case



_Code quote:_
```XML
if let Some(callback) = access_method_change_callback {
```
___
*[`mullvad-ios/src/api_client/mod.rs` line 290 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OVHPp4A-aapOLwPvIDG:-OVHPp4A-aapOLwPvIDH:b-6e5wgd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f57a655ec9fc048b371520d0ebd590e6d1bb29e/mullvad-ios/src/api_client/mod.rs#L290)):*
> ```XML
>                     let uuid = setting.get_id();
>                     let uuid_bytes = uuid.as_bytes();
>                     // SAFETY: The callback is expected to be safe to call
> ```

And does it require anything from it's arguments? Does it take ownership of them? Is it safe to call multiple times? etc



_Code quote:_
```XML
 // SAFETY: The callback is expected to be safe to call
```

acb-mv (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464)* status: 1 of 9 files reviewed, 4 unresolved discussions (waiting on @MarkusPettersson98)

___
*[`mullvad-ios/src/api_client/mod.rs` line 92 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OVHPCwg-UaV8pPo9oUF:-OW2GtIfCvtqw2gE30VN:b-xvbdwj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f57a655ec9fc048b371520d0ebd590e6d1bb29e/mullvad-ios/src/api_client/mod.rs#L92)):*

Could we say something about the semantics of this pointer type? Is it fine to *always* dereference it? Is it fine if the inner pointer is `null`?

I suggest adding methods on the `ForeignPtr` type that clearly defines every possible action using this pointer, along with the appropriate safety checks.
</blockquote></details>

The only actions of this pointer type are to define it with an opaque pointer value received from a FFI caller and to extract this value to pass to a FFI callback.  The pointer is not touched by Rust code and no assumptions are made about what, if anything, it points to. The one invariant is that the callback should see the same binary pointer value as the initial caller passed.
___
*[`mullvad-ios/src/api_client/mod.rs` line 95 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OVHNxTF05LsVJcOO2KR:-OW2Hg095Z-o1MndV1cv:b-663hvj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f57a655ec9fc048b371520d0ebd590e6d1bb29e/mullvad-ios/src/api_client/mod.rs#L95)):*

I would avoid implementing `Sync` for this type, we don't need multiple threads sharing a reference to any `ForeignPtr`, right?
</blockquote></details>

Thanks; removed.
___
*[`mullvad-ios/src/api_client/mod.rs` line 285 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OVHOiDdBfuKXF07B2Gq:-OW2HxW80q4qJx8r86Yf:br7hgtv) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f57a655ec9fc048b371520d0ebd590e6d1bb29e/mullvad-ios/src/api_client/mod.rs#L285)):*

Should we log if there is no registered callback? Seems like a confusing case
</blockquote></details>

We could do, though this code is used exactly once on the Swift side where we do pass a callback. How would you write the logging `else`? (I'm not familiar with logging in Rust.)

acb-mv:
*[`mullvad-ios/src/api_client/mod.rs` line 290 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OVHPp4A-aapOLwPvIDG:-OW2JNVbEtOvAtg_-7wH:bu5jpio) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f57a655ec9fc048b371520d0ebd590e6d1bb29e/mullvad-ios/src/api_client/mod.rs#L290)):*

And does it require anything from it's arguments? Does it take ownership of them? Is it safe to call multiple times? etc
</blockquote></details>

The context pointer is the same one passed from the caller, and so is already accounted for. The UUID bytes are not expected to be valid beyond the call, and the callback copies them to a new Swift UUID structure.

Serock3 (commented):
> Reviewed 3 of 9 files at r1, 1 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464)* status: all files reviewed, 9 unresolved discussions (waiting on @acb-mv, @MarkusPettersson98, and @rablador)

___
*[`mullvad-api/src/access_mode.rs` line 29 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OX2k2xB7qAhaHeAfqwF:-OX3WXsHEb53jaaxlHGg:b-81dmi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9ea06fb99de5edffaad0b61f88cb4df5f563d7b8/mullvad-api/src/access_mode.rs#L29)):*

What does this do?
</blockquote></details>

Derives a string representation of the type that can be used for debug-printing. If you see `log::info!("Acces method event: {access_method_event:?})`, then that is what it is using.

I can't see where this is used, though. If it isn't then it should be removed to reduce compile times.
___
*[`mullvad-ios/src/api_client/mod.rs` line 285 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OVHOiDdBfuKXF07B2Gq:-OX3gKg-CGfZJTxiYCSI:b8x6dvd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f57a655ec9fc048b371520d0ebd590e6d1bb29e/mullvad-ios/src/api_client/mod.rs#L285)):*

We could do, though this code is used exactly once on the Swift side where we do pass a callback. How would you write the logging `else`? (I'm not familiar with logging in Rust.)
</blockquote></details>

You should put the `if let Some(callback)...` part outside the `tokio::spawn`, so that no task is spawned if the callback is `None`.


Suggestion:



_Code snippet:_
```XML
        if let Some(callback) = access_method_change_callback {
            tokio::spawn(async move {
                let access_method_change_ctx = access_method_change_ctx;
                // SAFETY: The callback is expected to be called from the Swift side
                while let Some((event, _sender)) = rx.next().await {
                    let AccessMethodEvent::New {
                        setting,
                        connection_mode: _,
                        endpoint: _,
                    } = event
                    else {
                        continue;
                    };
                    let uuid = setting.get_id();
                    let uuid_bytes = uuid.as_bytes();
                    // SAFETY: The callback is expected to be safe to call
                    unsafe { callback(access_method_change_ctx.ptr, uuid_bytes.as_ptr()) };
                }
            })
        } else {
            log::warn!(
                "No access method change callback provided, access method changes will not be reported to Swift"
            );
        };
```
___
*[`mullvad-api/src/access_mode.rs` line 406 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OX3cbb_26RnxS_02x3d:-OX3cbb_26RnxS_02x3e:bsf95mb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9ea06fb99de5edffaad0b61f88cb4df5f563d7b8/mullvad-api/src/access_mode.rs#L406)):*
> ```XML
>             connection_mode: resolved.connection_mode.clone(),
>             #[cfg(not(target_os = "android"))]
>             endpoint: resolved.endpoint.clone(),
> ```

This clone can be removed
___
*[`mullvad-ios/src/api_client/mod.rs` line 188 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OX3Z4lH6xy-4ufxBxlK:-OX3Z4lH6xy-4ufxBxlL:b-y1jvj2) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9ea06fb99de5edffaad0b61f88cb4df5f563d7b8/mullvad-ios/src/api_client/mod.rs#L188)):*
> ```XML
>     settings_provider: SwiftAccessMethodSettingsWrapper,
>     address_cache: SwiftAddressCacheWrapper,
>     access_method_change_callback: Option<unsafe extern "C" fn(*const c_void, *const u8)>,
> ```

Might be appropriate to add some documentation to the callback and context, though I'm not sure whether they belong here on somewhere else as they are really to be consumed by the implementation on the rust side.

It should also be noted here that the callback is kept and used in a separate task, so it must be valid for the entire lifetime of the program.
___
*[`mullvad-ios/src/api_client/mod.rs` line 307 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OX3hLShCx9k_JqpyoWa:-OX3hLShCx9k_JqpyoWb:b3n9hmu) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9ea06fb99de5edffaad0b61f88cb4df5f563d7b8/mullvad-ios/src/api_client/mod.rs#L307)):*
> ```XML
>         });
> 
>         // TODO: do something with rx, and somehow let the `AccessMethodEvent`s it
> ```

Don't forget to remove these comments

MarkusPettersson98 (commented):
> Reviewed 1 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464)* status: all files reviewed, 1 unresolved discussion

___
*[`mullvad-ios/src/api_client/mod.rs` line 290 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OVHPp4A-aapOLwPvIDG:-OXNatTI4vc9K8hK7eep:btn2ccx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9f57a655ec9fc048b371520d0ebd590e6d1bb29e/mullvad-ios/src/api_client/mod.rs#L290)):*

The context pointer is the same one passed from the caller, and so is already accounted for. The UUID bytes are not expected to be valid beyond the call, and the callback copies them to a new Swift UUID structure.
</blockquote></details>

Good:)
___
*[`mullvad-ios/src/api_client/mod.rs` line 196 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OXNZ6H26mHLOcgfE-XO:-OXNZ6H26mHLOcgfE-XP:bddc2dj) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e86f6c4260ccb5bb11df2f5173d696ccce4b869a/mullvad-ios/src/api_client/mod.rs#L196)):*
> ```XML
>     settings_provider: SwiftAccessMethodSettingsWrapper,
>     address_cache: SwiftAddressCacheWrapper,
>     access_method_change_callback: Option<unsafe extern "C" fn(*const c_void, *const u8)>,
> ```

nit: A type alias would make this type signature look slightly better

```rust
/// We can even put some nice docs here if we want!
/// E.g. what are these two pointers pointing to? Are they nullable? etc.
pub type AccessMethodChangedCallback = unsafe extern "C" fn(*const c_void, *const u8);

..

access_method_change_callback: Option<AccessMethodChangedCallback>,
```



_Code quote:_
```XML
unsafe extern "C" fn(*const c_void, *const u8)
```
___
*[`mullvad-ios/src/api_client/mod.rs` line 245 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-OXN_QFA0dzE6fYZhqdB:-OXN_QFA0dzE6fYZhqdC:b-nne8hb) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/e86f6c4260ccb5bb11df2f5173d696ccce4b869a/mullvad-ios/src/api_client/mod.rs#L245)):*
> ```XML
>     settings_provider: SwiftAccessMethodSettingsWrapper,
>     address_cache: SwiftAddressCacheWrapper,
>     access_method_change_callback: Option<unsafe extern "C" fn(*const c_void, *const u8)>,
> ```

We could re-use the type alias here!



_Code quote:_
```XML
unsafe extern "C" fn(*const c_void, *const u8)
```

rablador (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

rablador (approved):
@MarkusPettersson98 The current build fails on a windows build step. Do you have any insight in this?

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-:-OXS2y29AE9Cfwd68o2c:b-10wdfg)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

MarkusPettersson98 (commented):
I don't see any Windows build failures currently, so it must have been resolved :shrug: 

Regarding the failing linux build with nightly rust compiler: There seems to have been a regression in the rust compiler itself, so I advise that you ignore that particular error for now.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8464#-:-OXT8HKn3jbIy0vprjV0:b-6z72ds)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #8461, opened 2025-07-15
Fix clippy nightly warnings

MarkusPettersson98:
This PR fixes a bunch of nightly `clippy` warnings

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8461)

PR #8492, opened 2025-07-22
Remove unused code in settings migration

faern:
In a sense a follow-up to #8461. I noticed the addition of `#[allow(unused)]` here and was curious as to why this code was unused and why we should ignore it rather than delete it.

Turns out all these structs were unused already when they were introduced. I checked out the commit that first created `v7.rs` and built it with the latest nightly compiler. It complains about these structs being unused already back then. So this must have been a mistake at the creation of this migration, and not caught since the compiler was not as good at finding those back then.

The comment in the commit message from @MarkusPettersson98 is that we should not touch closed migration code. In general I agree. We should ideally not touch this code. But removing code that sits completely unused is different from refactoring code or changing the logic in the code. I don't think this could break anything. If the types were used for anything at all, the compiler should be complaining. We also have  unit tests for the migration, and they pass.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8492)

### DIFF
--- a/mullvad-daemon/src/migrations/v7.rs
+++ b/mullvad-daemon/src/migrations/v7.rs
@@ -1,11 +1,8 @@
-use std::net::SocketAddr;
-
 use super::{Error, Result};
 use mullvad_types::{
     relay_constraints::{BridgeConstraints, BridgeSettings as NewBridgeSettings, BridgeType},
     settings::SettingsVersion,
 };
-use serde::{Deserialize, Serialize};
 use talpid_types::net::{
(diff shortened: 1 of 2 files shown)


dlon (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8492#-:-OVm6jVR3SedBBofY1Cd:bnfp4nl)* status: all files reviewed, 2 unresolved discussions

PR #8505, opened 2025-07-24
Specify image version

Rawa:


Rawa:
Github actions do only provide latest images for OS versions where the NDK is a moving target 😢

PR #8511, opened 2025-07-28
Check for location cells matching name and tag in e2e tests

Pururun:
Previously it would look for the first location cell and try to expand that one. Since this can in some cases be a recents cell the test would fail as it is not possible to expand a recents cell.

~This PR instead changes this behaviour to instead look for all matching location cells and then try to expand all of them. It will not throw an error if it can find any which is expandable.~

Instead we will add a type of cell tag to all names so that we can check for both name and tag and avoid trying to click on the expand cell of recent location items.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8511)


Pururun:
E2E test run https://github.com/mullvad/mullvadvpn-app/actions/runs/16569235898/job/46858026614

Pururun:
New E2E test run:
https://github.com/mullvad/mullvadvpn-app/actions/runs/16598142893

Rawa (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8511)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #8524, opened 2025-07-31
Only log failures to allow notifications in mock api and e2e tests

Pururun:
Sometimes the allow notification dialog will not show. We should still proceed with the test if that is the case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8524)

### DIFF
--- a/android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/interactor/AppInteractor.kt
+++ b/android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/interactor/AppInteractor.kt
@@ -6,12 +6,14 @@ import android.os.Build
 import androidx.test.uiautomator.By
 import androidx.test.uiautomator.UiDevice
 import androidx.test.uiautomator.Until
+import co.touchlab.kermit.Logger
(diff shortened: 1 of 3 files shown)


Pururun (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8524)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @Rawa)

___
*[`android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/page/PrivacyPage.kt` line 9 at r3](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8524#-OWUhVts5yNhTiSaEwBK:-OWUiWTm90BK4QT2DiOR:b-896fix) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/36ca473e3cef88a7ee3284e71d39dd16cd2b2f31/android/test/common/src/main/kotlin/net/mullvad/mullvadvpn/test/common/page/PrivacyPage.kt#L9)):*

We should remove this allow selector, it was used for the notifications
</blockquote></details>

Done.

PR #8543, opened 2025-08-04
Fix flaky AddTimeViewModel tests

Pururun:
Use `UnconfiedTestDispatcher` as dispatcher for all calls to `viewModelScope.launch` in tests which makes the tests pass every time.

### DIFF
--- a/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/AddTimeViewModel.kt
+++ b/android/app/src/main/kotlin/net/mullvad/mullvadvpn/viewmodel/AddTimeViewModel.kt
@@ -3,6 +3,8 @@ package net.mullvad.mullvadvpn.viewmodel
 import android.app.Activity
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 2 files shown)


Rawa (commented):
Not sure why but wasn't able to reproduce it at first, now once I tried again I was successful. Thanks. :star2: Ah so it is all tied to the dispatcher used by the ViewModel. Not sure what the best practice is here, I looked around and saw that using the follow is also an alternative:
```
Dispatchers.setMain(StandardTestDispatcher())
```

Apparently we already do this is `OutOfTimeUseCaseTest` and `LoginViewModelTest` any reason for choosing this strategy instead? In the former I added it according to the commit history and in the latter you did. It was far to long ago for remembering my reasoning (if any) why this was added to the test. Maybe we should investigate a bit further if this should be a broader pattern we decide on across all viewmodels?

I also tried the `setMain` and it seemed to work fine as well, but I'm not sure what we want here exactly.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8543#-:-OWpZbxs3jfwuwoyYOUY:brmn8eu)* status: 0 of 2 files reviewed, all discussions resolved

Rawa (commented):
Oh, this explains a lot. Yeah seems like the purpose is to do exactly this :sweat_smile: . Can you add some documentation to the `TestCoroutineRule` as well? It is in `common` and is applied everywhere but no doc's :see_no_evil: 

> Reviewed 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8543#-:-OWt--Zw3BLqGBDb37kb:bkr5rnt)* status: 1 of 4 files reviewed, all discussions resolved

PR #8557, opened 2025-08-06
Fix obfuscation not handling .off in release mode

rablador:
When adding .quic to obfuscation the debug mode was changed and obfuscation = .off was missed. This fixes that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8557)

### DIFF
--- a/ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift
+++ b/ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift
@@ -57,18 +57,19 @@ public class ProtocolObfuscator<Obfuscator: TunnelObfuscation>: ProtocolObfuscat
 
         remotePort = endpoint.ipv4Relay.port
 
-        #if DEBUG
         let obfuscationProtocol: TunnelObfuscationProtocol? = switch obfuscationMethod {
         case .udpOverTcp:
             .udpOverTcp
         case .shadowsocks:
             .shadowsocks
+        #if DEBUG
         case .quic:
             if let relayFeatures = relayFeatures?.quic {
                 .quic(hostname: relayFeatures.domain, token: relayFeatures.token)
             } else {
                 nil
             }
+        #endif
         default:
             // This is fine, since ObfuscationMethodSelector.obfuscationMethodBy` above should never
             // return .automatic.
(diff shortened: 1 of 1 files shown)


rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8557)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @mojganii and @SteffenErn)

___
*[`ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift` line 66 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8557#-OWzNBvUClGQmzsSmAF6:-OWzUY3eAMwD-8z1YsjP:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9509d80735014fe6e0d82c4c9a75a954b5aaf923/ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift#L66)):*
<details><summary><i>Previously, SteffenErn (Steffen) wrote…</i></summary><blockquote>

We could wrap the '#if DEBUG ' around this case only and remove the '#else' branch to keep things cleaner
</blockquote></details>

Done
___
*[`ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift` line 90 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8557#-OWzJl4t0qFejwZKLZHW:-OWzVCl909p3LoF6wiW_:b-19buk0) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/9509d80735014fe6e0d82c4c9a75a954b5aaf923/ios/PacketTunnelCore/Actor/ProtocolObfuscator.swift#L90)):*

This object shouldn't be initialized unless `Obfuscation` is turned on in `PacketTunnelActor`, and it should be deinitialized when `Obfuscation` is turned off.
 
 Setting it to `nil` from another object violates the **Single Responsibility Principle**, especially since `tunnelObfuscator` is a reference type. Managing its lifecycle is not the responsibility of other objects and doing so can lead to unpredictable behavior.

 For example, if a user turns off `Obfuscation` and then tries to turn it back on, the system might be in an invalid state — especially if ARC has already deallocated the previous instance. At that point, `tunnelObfuscator` would be `nil`, and obfuscation wouldn't occur as expected.
 
 Since `PacketTunnelActor` owns `tunnelObfuscator`, it should be solely responsible for initializing and deinitializing it.
</blockquote></details>

Are you mistaking `tunnelObfuscator` for `protocolObfuscator` here? `tunnelObfuscator` is created, nilled and owned by this class and not by `PacketTunnelActor`.

PR #8559, opened 2025-08-07
Set relay to current country on first start

kl:
Sets the default relay selection to the current country (as determined by am.i.mullvad.net).

In non-release builds of the Android app we do not bundle a relay list in the APK, and the relay list is fetched when the user logs in.

So one of the following can happen:

1. Geolocation request returns, we have a relay list.
2. Geolocation request returns, we do not yet have a relay list.
3. Relay list request returns, we have a geolocation.
4. Relay list request returns, we do not have a geolocation.

In 1. and 3. we can update the default locaiton. In 2. we have to wait until the relay list is fetched from the api until we can update the default location. 4. is unlikely to happen but could happen if am.i.mullvad is down.

If the user's country doesn't have a relay the closest country that has a relay is used.

fixes: DROID-2064

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559)

### DIFF
--- a/android/CHANGELOG.md
+++ b/android/CHANGELOG.md
@@ -25,6 +25,7 @@ Line wrap the file at 100 chars.                                              Th
 ### Added
 - Add list of recent server selections in the select location view.
 - Add support for devices with 16KB page size.
+- Set default server based on the current location when the app is started for the first time.
 
 
 ## [android/2025.6] - 2025-08-12
--- a/android/test/e2e/src/main/kotlin/net/mullvad/mullvadvpn/test/e2e/DefaultLocationTest.kt
+++ b/android/test/e2e/src/main/kotlin/net/mullvad/mullvadvpn/test/e2e/DefaultLocationTest.kt
@@ -1,0 +1,44 @@
+package net.mullvad.mullvadvpn.test.e2e
+
+import java.io.File
+import net.mullvad.mullvadvpn.test.common.page.ConnectPage
+import net.mullvad.mullvadvpn.test.common.page.LoginPage
+import net.mullvad.mullvadvpn.test.common.page.on
+import net.mullvad.mullvadvpn.test.e2e.misc.AccountTestRule
+import org.json.JSONObject
+import org.junit.jupiter.api.Assertions.assertFalse
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.RegisterExtension
+
+class DefaultLocationTest : EndToEndTest() {
+
+    @RegisterExtension @JvmField val accountTestRule = AccountTestRule()
+
+    @Test
+    fun testUpdateDefaultLocationFlag() {
+        app.launchAndEnsureOnLoginPage()
+
+        // The update_default_location flag should be set to true when first starting the app
+        assert(readUpdateDefaultLocationKeyFromSettings())
+
+        on<LoginPage> {
+            enterAccountNumber(accountTestRule.validAccountNumber)
+            clickLoginButton()
+        }
+
+        on<ConnectPage>()
+
+        // After we have logged in the daemon will have set the new default location so the
+        // flag should be false.
+        assertFalse(readUpdateDefaultLocationKeyFromSettings())
+    }
+
+    private fun readUpdateDefaultLocationKeyFromSettings(): Boolean {
+        val settings = File(targetApplication.filesDir, "settings.json")
+        if (!settings.isFile()) error("settings.json does not exist")
+
+        val text = settings.readText()
+        val json = JSONObject(text)
+        return json.getBoolean("update_default_location")
+    }
+}
--- a/mullvad-daemon/src/lib.rs
+++ b/mullvad-daemon/src/lib.rs
@@ -45,7 +45,9 @@ use mullvad_encrypted_dns_proxy::state::EncryptedDnsProxyState;
 use mullvad_relay_selector::{RelaySelector, SelectorConfig};
 #[cfg(target_os = "android")]
 use mullvad_types::account::{PlayPurchase, PlayPurchasePaymentToken};
-use mullvad_types::relay_constraints::GeographicLocationConstraint;
+use mullvad_types::relay_constraints::{
+    GeographicLocationConstraint, LocationConstraint, RelayConstraints, WireguardConstraints,
+};
 #[cfg(any(windows, target_os = "android", target_os = "macos"))]
 use mullvad_types::settings::SplitApp;
 #[cfg(daita)]
@@ -413,6 +415,9 @@ pub enum DaemonCommand {
     ExportJsonSettings(ResponseTx<String, settings::patch::Error>),
     /// Request the current feature indicators.
     GetFeatureIndicators(oneshot::Sender<FeatureIndicators>),
+    // Updates the default (initial) country selection that the user will see when starting the
+    // app for the first time based on their current geolocation.
+    UpdateDefaultLocationCountry(ResponseTx<(), settings::Error>),
 
     // Debug features
     DisableRelay {
@@ -915,8 +920,13 @@ impl Daemon {
         api::forward_offline_state(api_availability.clone(), offline_state_rx);
 
         let relay_list_listener = management_interface.notifier().clone();
+        let internal_event_tx_clone = internal_event_tx.clone();
         let on_relay_list_update = move |relay_list: &RelayList| {
             relay_list_listener.notify_relay_list(relay_list.clone());
+            let (tx, _) = oneshot::channel();
+            let _ = internal_event_tx_clone.send(InternalDaemonEvent::Command(
+                DaemonCommand::UpdateDefaultLocationCountry(tx),
+            ));
         };
 
         let mut relay_list_updater = RelayListUpdater::spawn(
@@ -1303,6 +1313,13 @@ impl Daemon {
             _ => return,
         };
 
+        if self.settings.update_default_location {
+            let (tx, _) = oneshot::channel();
+            let _ = self.tx.send(InternalDaemonEvent::Command(
+                DaemonCommand::UpdateDefaultLocationCountry(tx),
+            ));
+        }
+
         self.management_interface
             .notifier()
             .notify_new_state(self.tunnel_state.clone());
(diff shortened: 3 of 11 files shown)


kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[`mullvad-daemon/src/lib.rs` line 1919 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OX2_LKV9XXTI8EtdxNN:-OX2wf0E1dT-tZdWKcwG:b-rlsmqd) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d94ca8778d167d08b8ec0fd498ab74475d153f1b/mullvad-daemon/src/lib.rs#L1919)):*

This await feels pointless. We log a warning when dropping the receiver, but otherwise it would be fine to just drop it.

Maybe this function could also take a oneshot like the others, pass it on to `SetRelaySettings`, and respond with `Ok(())` in other cases.
</blockquote></details>

Changed this, please have a look again.
___
*[`mullvad-relay-selector/src/relay_selector/mod.rs` line 531 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OX2arzW4-qC3FbUCwlg:-OX2mLNQ5WOocARQGCSF:becdw2o) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d94ca8778d167d08b8ec0fd498ab74475d153f1b/mullvad-relay-selector/src/relay_selector/mod.rs#L531)):*

Would prefer `selector.get_relays().lookup_country_by_name(country_name)?.code`. I.e., implement the lookup for `RelayList`, not the relay selector
</blockquote></details>

Fixed. Also added a new function `access_relays()` to `RelaySelector` because `get_relays()` will clone the relay list which seems wasteful if we don't need to do that.
___
*[`mullvad-types/src/relay_constraints.rs` line 49 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OX2aTHsEjRjokjNh2Fj:-OX2mF5H3LQAZVLuJCr_:btwqiay) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d94ca8778d167d08b8ec0fd498ab74475d153f1b/mullvad-types/src/relay_constraints.rs#L49)):*

Somewhat cluttered/ad hoc method, in my opinion
</blockquote></details>

Do you want me to inline the call or put it somewhere else?
___
*[`mullvad-types/src/settings/mod.rs` line 75 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OX2ZzeK6VyWFAz-q9SH:-OX2m3AH3ToJGPAsCx4s:bc2aarw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d94ca8778d167d08b8ec0fd498ab74475d153f1b/mullvad-types/src/settings/mod.rs#L75)):*

Will this not override the selected relay for all users with existing settings? Is that a good idea?
</blockquote></details>

Yes good point. I think we need a new migration that sets the default location to Updated for existing users. Or is there another way?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559)* status: all files reviewed, 2 unresolved discussions (waiting on @kl)

___
*[`mullvad-types/src/relay_constraints.rs` line 49 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OX2aTHsEjRjokjNh2Fj:-OX4Q2JV8YDUSoIzgFnx:b-je5kbx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d94ca8778d167d08b8ec0fd498ab74475d153f1b/mullvad-types/src/relay_constraints.rs#L49)):*

Do you want me to inline the call or put it somewhere else?
</blockquote></details>

I think inlining is alright.
___
*[`mullvad-types/src/settings/mod.rs` line 75 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OX2ZzeK6VyWFAz-q9SH:-OX49U928Pfi7x1Y4IzS:bz0x4o9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d94ca8778d167d08b8ec0fd498ab74475d153f1b/mullvad-types/src/settings/mod.rs#L75)):*

Yes good point. I think we need a new migration that sets the default location to Updated for existing users. Or is there another way?
</blockquote></details>

Probably not. Maybe guess that they haven't explicitly changed it if the location is Sweden? Not great, though

kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559)* status: all files reviewed, 1 unresolved discussion

___
*[`mullvad-types/src/relay_constraints.rs` line 49 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OX2aTHsEjRjokjNh2Fj:-OX7sG2M4arn16GQ3iUP:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/d94ca8778d167d08b8ec0fd498ab74475d153f1b/mullvad-types/src/relay_constraints.rs#L49)):*

I think inlining is alright.
</blockquote></details>

Done

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559)* status: all files reviewed, 1 unresolved discussion (waiting on @kl)

kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`mullvad-daemon/src/lib.rs` line 1890 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OXU2_GX5DnRwxbvW64s:-OXXIn-7-9EGbLAFYJKy:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1dc3ca6e501f5e23d32d695d29650484edb865e8/mullvad-daemon/src/lib.rs#L1890)):*

This is a bit racey, but perhaps it does not matter? I.e. it could be false twice if called twice in quick succession.
</blockquote></details>

Done
___
*[`mullvad-daemon/src/lib.rs` line 1920 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OXU3ZJ85h0TStoTokGU:-OXXIpDp3jxQ9lUPbviD:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1dc3ca6e501f5e23d32d695d29650484edb865e8/mullvad-daemon/src/lib.rs#L1920)):*

Ideally, we should cancel this if the user updates the relay settings before this command is executed. There's a race condition here.

Would it be possible to update the settings here directly instead of sending another daemon command, e.g. just call `self.on_set_relay_settings(tx, relay_settings)`? This would also fix the other race above.
</blockquote></details>

Done
___
*[`mullvad-daemon/src/settings/mod.rs` line 165 at r4](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OXU5jx366mGcpjIW3V5:-OXXIqCe09GU0Bc6PIp7:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/1dc3ca6e501f5e23d32d695d29650484edb865e8/mullvad-daemon/src/settings/mod.rs#L165)):*

Maybe it makes more sense to set this in `Self::default_settings`? If the settings are reset or fail to be read, it might also make sense to set this to true.
</blockquote></details>

Done

kl (commented):
Updated changelog.

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-:-OXY7U-j9zvS6m-A652N:bjssqys)* status: all files reviewed, 1 unresolved discussion (waiting on @Pururun)

___
*[`-- commits` line 2 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OXXZuZV8Tzlua75hLKA:-OXY4T0R8Cdgur7kAC7w:b5v5mq2):*

This is a really good explanation of what we do, I wonder if there is somewhere else we should document this?
</blockquote></details>

Yeah not sure, could put it in linear as well?
___
*[`mullvad-daemon/src/lib.rs` line 1920 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OXXVlRD0VDfHNm9jG-D:-OXY4bIG9ZhaJ_YmIGvr:bb6un1j) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dc4eb04e57995fb17b3232325ddaa38099356cd5/mullvad-daemon/src/lib.rs#L1920)):*

Nit: I missed this before, but you probably want to return `Ok(())` on `tx` in the `None` case, right? Since you don't actually use the result at the moment, it does not matter too much.
</blockquote></details>

Done

dlon (dismissed):
> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559)* status: all files reviewed, 1 unresolved discussion (waiting on @Pururun)

___
*[`mullvad-daemon/src/lib.rs` line 1920 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OXXVlRD0VDfHNm9jG-D:-OXYcd7138yJ6tCK4cvb:b-jgcp6g) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dc4eb04e57995fb17b3232325ddaa38099356cd5/mullvad-daemon/src/lib.rs#L1920)):*

Done
</blockquote></details>

Probably before the other return statements as well

kl (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559)* status: all files reviewed, 1 unresolved discussion

___
*[`mullvad-daemon/src/lib.rs` line 1920 at r5](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-OXXVlRD0VDfHNm9jG-D:-OXbX8daCZAhwN32sr6e:b-886gsx) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/dc4eb04e57995fb17b3232325ddaa38099356cd5/mullvad-daemon/src/lib.rs#L1920)):*

Probably before the other return statements as well
</blockquote></details>

Fixed

Pururun (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8559#-:-OXbhwj-3xqYaVM38Nq-:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #8561, opened 2025-08-07
Fix recents can contain invalid multihop entry

kl:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8561)

### DIFF
--- a/mullvad-types/src/settings/mod.rs
+++ b/mullvad-types/src/settings/mod.rs
@@ -139,6 +139,19 @@ impl TryFrom<&RelaySettings> for Recent {
                         .ok_or("Location must be Constraint::Only")?
                         .clone();
 
+                    if matches!(
+                        entry,
+                        LocationConstraint::Location(GeographicLocationConstraint::Hostname(..))
+                    ) && matches!(
+                        location,
+                        LocationConstraint::Location(GeographicLocationConstraint::Hostname(..))
+                    ) && entry == location
+                    {
+                        return Err(
+                            "Multihop recent cannot have identical (country, city, host) triple.",
+                        );
+                    }
+
                     Recent::Multihop {
                         entry,
                         exit: location,


Pururun (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8561#-:-OX2mxgi6cS3MAvgNos9:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

Rawa (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8561)* status: all files reviewed, 1 unresolved discussion

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/usecase/RecentsUseCase.kt` line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8561#-OX2nW-C4Xd4K7row8Tq:-OX2nW-C4Xd4K7row8Tr:b-dhalk9) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f1845e8550d396476528fb36351bcf0b1e7903b2/android/app/src/main/kotlin/net/mullvad/mullvadvpn/usecase/RecentsUseCase.kt#L37)):*
> ```Kotlin
>                         val exit = recent.exit.findItem(exitCustomLists, exitRelayList)
> 
>                         val isValid = entry != null && exit != null && entry != exit
> ```

Should we solve this in daemon instead? Otherwise other platforms will show this entry as well?

Rawa (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8561)* status: all files reviewed, 2 unresolved discussions (waiting on @kl)

___
*[`android/app/src/main/kotlin/net/mullvad/mullvadvpn/usecase/RecentsUseCase.kt` line 37 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8561#-OX2nl9kDCF7dH1zWheD:-OX2nl9kDCF7dH1zWheE:b4gi01e) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/f1845e8550d396476528fb36351bcf0b1e7903b2/android/app/src/main/kotlin/net/mullvad/mullvadvpn/usecase/RecentsUseCase.kt#L37)):*
> ```Kotlin
>                         val exit = recent.exit.findItem(exitCustomLists, exitRelayList)
> 
>                         val isValid = entry != null && exit != null && entry != exit
> ```

Also  fine for it to be same entry & exit if they have more than 1 relay in them. E.g it is ok to do Multihop from `Sweden` to `Sweden`, but a "Multi"-hop `se-got-wg-001`to `se-got-wg-001` would be invalid because it is the same server.

kl:
Should be noted that if you have the same city as entry and exit and the city only has one relay it would not be filtered out with this check. We could check for this as well but that would be a bit more complexity since we need to look into the relay list for that. Looks like we have at least 2 relays for every city so I don't think this should be a problem in practice.

Rawa (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8561)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #8578, opened 2025-08-11
Fix nightly E2E tests

rablador:
This PR fixes the UI tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8578)

### DIFF
--- a/.github/workflows/ios-screenshots-tests.yml
+++ b/.github/workflows/ios-screenshots-tests.yml
@@ -43,6 +43,10 @@ jobs:
           brew update
           brew install zip
 
+      - name: Select Xcode version. Remove this step after 25/8 2025, when Xcode 16.4 will be the default version.
+        run: |
+          sudo xcode-select -s "/Applications/Xcode_16.4.app/Contents/Developer"
+
       - name: Run screenshot tests
         run: |
           set -o pipefail && env NSUnbufferedIO=YES xcodebuild \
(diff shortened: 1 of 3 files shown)


rablador (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8578)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @mojganii)

___
*[`ios/MullvadVPNUITests/Pages/DeviceManagementPage.swift` line 66 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8578#-OXXHtwR4Ml14cpR29E_:-OXXo0oc85eqOO70peGa:b-rbuqo8) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/0e42fc2f6017282631870fc4dba1374b31c68e85/ios/MullvadVPNUITests/Pages/DeviceManagementPage.swift#L66)):*

it used to work with `getAccountWithTime()` to see if there is no value for `PARTNER_API_TOKEN`  then  it falls back to fetch `HAS_TIME_ACCOUNT_NUMBER` that is  a staged account number from `UITests.xcconfig`.  as I'm aware `PARTNER_API_TOKEN` is setup and handles by Github. on CI. I don't think we have access to that from somewhere else. 

if we suppose these tests are `environment-dependent` tests then we should **mock or stub dependencies** locally to run them without needing the real secret..
</blockquote></details>

The tests still use `getAccountWithTime()` and such functions in the main test files. The asserts in this file is only to verify certain things, so that hasn't changed.

mojganii (dismissed):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8578#-:-OXcWfs85mTrqTQVgodq:bnfp4nl)* status: 0 of 14 files reviewed, 4 unresolved discussions (waiting on @rablador)

PR #8591, opened 2025-08-12
Invalidate CDN cache after uploading release artifacts

dlon:
Fixes artifacts not being updated when release tags are replaced.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8591)

### DIFF
--- a/ci/buildserver-upload.sh
+++ b/ci/buildserver-upload.sh
@@ -12,6 +12,14 @@ source "$SCRIPT_DIR/buildserver-config.sh"
 
 cd "$UPLOAD_DIR"
 
+function invalidate_bunny_cdn_cache {
+    curl --request POST \
+        --url "https://api.bunny.net/pullzone/${BUNNYCDN_PULL_ZONE_ID}/purgeCache" \
+        --header "AccessKey: ${BUNNYCDN_API_KEY}" \
+        --header 'content-type: application/json' \
+        --fail-with-body
+}
+
 function rsync_upload {
     local file=$1
     local upload_dir=$2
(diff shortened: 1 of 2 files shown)


ci/buildserver-upload.sh
```diff
@@ -39,10 +50,12 @@ while true; do
 
         if [[ "$platform" == "installer-downloader" ]]; then
             upload_path="desktop/installer-downloader"
+            invalidate_cache="true"
```
faern: Do we really need to invalidate the cache here? The Bunny CDN is not serving anything from this folder.

ci/buildserver-upload.sh
```diff
@@ -23,6 +31,9 @@ function rsync_upload {
 
 while true; do
     sleep 10
+
+    invalidate_cache="false"
```
faern: I would be more detailed here to help avoid confusion about what cache. For example `invalidate_bunny_cdn_cache`. We have our own CDN at `cdn.mullvad.net` so just saying `invalidate_cdn_cache` is not enough to clearly define what is being invalidated.

ci/buildserver-upload.sh
dlon: True. I've removed it.

ci/buildserver-upload.sh
dlon: Done.

faern (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8591)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #8594, opened 2025-08-12
Upgrade to React 19

tobias-jarvelov:
- Upgrade `react`/`react-dom` packages from version `^18.3.x` to `^19.1.x`
- Upgrade `redux` package from version `^7.2.9` to `^9.2.0`
- Upgrade `eslint` react hooks and compiler plugins from experimental to stable versions
- Fix post-upgrade typescript errors
  - Add default value for `useRef` since it is required as of React 19
  - Add `Props` argument for `ReactElement` usage to ensure typescript is aware of the valid HTML attributes available.
- Fix post-upgrade lint errors and warnings
  - Mark knowingly unhandled return values as `void`
  - Disable `exhaustive-deps` rule for hooks which uses a `useEffectEvent` callback as the stable `eslint` version can not be configured to handle that situation.
- Update code and types using deprecated functions/features
  - Remove use of React's `batch` callback as it is a no-op since React 18
  - Replace `String.substr` and `Buffer.slice` with non-deprecated alternatives
  - Replace type `MutableRefObject` with `RefObject`

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8594)


olmoh (changes requested):
We discussed this a bit and agreed that it's confusing to have a hook named `useEffectEvent` whose return function gets passed to a dependency array, which is what the React hook with the same name can be used to avoid.
We will ignore the lint errors on affected dependency arrays for the time being and fix this in a better sometime soon.

desktop/packages/mullvad-vpn/src/renderer/components/SplitTunnelingSettings.tsx
```diff
@@ -126,6 +126,11 @@ function LinuxSplitTunnelingSettings(props: IPlatformSplitTunnelingSettingsProps
     });
   });
 
+  // These lint rules are disabled for now. See the following
+  // issue for potential solutions to enable the rules again:
+  // https://linear.app/mullvad/issue/DES-2381
```
olmoh: I don't think we need to add this link here since not everyone has access to it. In my opinion it would be cleaner to add a identical comment for each case and then refer to that from the issue instead.

olmoh (dismissed):

desktop/packages/mullvad-vpn/src/renderer/components/SplitTunnelingSettings.tsx
tobias-jarvelov: Agree, added a message inline in the code.

olmoh (approved):

PR #8597, opened 2025-08-13
Bump MACOSX_DEPLOYMENT_TARGET to 10.12

faern:
10.12 is the oldest that rustc supports right now. 10.7 was emitting a warning at build time. 10.12 is much older than what we run app tests on or officially supports. But some users run older macOS versions and are happy, so we should not lock them out for no good reason.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8597)


faern:
I have verified that `cargo build` stops emitting these warnings with this PR :)

PR #8607, opened 2025-08-14
Fix multihop entry hop leak

dlon:
Fix DES-2277


hulthe (changes requested):
> Reviewed 5 of 19 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8607)* status: 6 of 19 files reviewed, 3 unresolved discussions

___
*[`talpid-core/src/firewall/windows/winfw/mod.rs` line 113 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8607#-OY0zGq58z0UQt5IDXyQ:-OY0zGq58z0UQt5IDXyR:b-rh3kvw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b674524cf1066eb13fb90e68477453f6bb7c1294/talpid-core/src/firewall/windows/winfw/mod.rs#L113)):*
> ```XML
>         .as_ref()
>         .map(|path| path.as_ptr())
>         .unwrap_or_default();
> ```

I'm assuming this would return a null pointer if None. Maybe being eplicit and using map_or(null, ...) is worthwhile considering we're dealing with raw pointers.

Sanity check: Using a null pointer is correct, no?

Same question for all the null pointer changes below
___
*[`talpid-core/src/firewall/windows/winfw/mod.rs` line 115 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8607#-OY0yowMAypMwTxhaSIV:-OY0yowMAypMwTxhaSIW:b-7rt0bl) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b674524cf1066eb13fb90e68477453f6bb7c1294/talpid-core/src/firewall/windows/winfw/mod.rs#L115)):*
> ```XML
>         .unwrap_or_default();
> 
>     // TODO: Make this actually unreachable
> ```

If only rust had support for non-empty arrays :)
___
*[`talpid-core/src/firewall/windows/winfw/mod.rs` line 198 at r2](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8607#-OY12_fCCUG3RAV60A3Y:-OY12_fCCUG3RAV60A3Z:bmgl1oa) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/b674524cf1066eb13fb90e68477453f6bb7c1294/talpid-core/src/firewall/windows/winfw/mod.rs#L198)):*
> ```XML
>         )
>     };
>     // Must be dropped after pointer deref
> ```

is this for the same reason as all the variables below?

PR #8622, opened 2025-08-18
Fix failure to initialize a QUIC connection

MarkusPettersson98:
This PR puts a bandaid on a weird bug in the masque proxy server. The masque proxy client accepts a minimum `max_udp_payload_size` value of 1200 bytes, but the masque proxy server defines the minimum `max_udp_payload_size` to 1252 (lol).

This PR implements `compute_udp_payload_size` properly, but it also enables a temporary hack where we raise the minimum `max_udp_payload_size` to 1252 to match the server's expected minimum value, which fixes connecting over QUIC on macOS.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8622)


hulthe:
We should be able to push a fix to the server today, if so we can revert the 1252 workaround

PR #8626, opened 2025-08-18
Fix `include_in_country` flag filtering out relays too aggresively

MarkusPettersson98:
This PR fixes an issue where the user would be blocked by selecting a country with very few relays that had the `include_in_country` flag set to `true` and the the relay selector somehow managed to filter out all candidate relays. This would manifest itself in a multihop constraint.

Added a small test case that would demonstrate the relay selector being too dumb to pick a valid multihop circuit. Revert the last commit to see it fail.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8626)

### DIFF
--- a/.github/workflows/git-commit-message-style.yml
+++ b/.github/workflows/git-commit-message-style.yml
@@ -34,4 +34,4 @@ jobs:
           # This action defaults to 50 char subjects, but 72 is fine.
           max-subject-line-length: '72'
           # The action's wordlist is a bit short. Add more accepted verbs
-          additional-verbs: 'tidy, wrap, obfuscate, bias, prohibit, forbid, revert, slim, impl'
+          additional-verbs: 'tidy, wrap, obfuscate, bias, prohibit, forbid, revert, slim, impl, disregard'
--- a/.github/workflows/git-commit-message-style.yml
+++ b/.github/workflows/git-commit-message-style.yml
@@ -34,4 +34,4 @@ jobs:
           # This action defaults to 50 char subjects, but 72 is fine.
           max-subject-line-length: '72'
           # The action's wordlist is a bit short. Add more accepted verbs
-          additional-verbs: 'tidy, wrap, obfuscate, bias, prohibit, forbid, revert, slim, impl, disregard'
+          additional-verbs: 'tidy, wrap, obfuscate, bias, prohibit, forbid, revert, slim, impl, disregard, reproduce'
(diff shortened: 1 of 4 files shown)


Serock3 (approved):
> Reviewed 5 of 5 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8626)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`mullvad-relay-selector/src/relay_selector/matcher.rs` line 25 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8626#-OXxoSdu2TRWsXPmBC1k:-OY0GVhi6CefssNYVRCz:bdwizaz) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/61a20067a4ebf2fedf3019afc6821d2fd7c8111f/mullvad-relay-selector/src/relay_selector/matcher.rs#L25)):*

I did what you suggested, but mapped `save_a_dime => filter_matching_relay_list` from your example, since that function is supposed to be called *everywhere*. It's just in `get_wireguard_multihop_config` that we would like to opt out of filtering on `include_in_country`. Added a new public function `filter_matching_relay_list_by_query` without the semantics of filtering on `include_in_country`. Do you think that's fair?

I'm kind of hesitant to break the symmetry of the type signatures of `filter_matching_relay_list` and `filter_matching_relay_by_query`, I'd rather recompute `locations`. `filter_on_include_in_country` is not a public function, so I don't care about the type of that function as much, so might as well pass it `locations` directly.
</blockquote></details>

That works for me!

I won't block the PR on it, but it's a little unfortunate that `filter_matching_relay_list` and `filter_matching_relay_list_by_query` have such similar names though. Not sure I can think of something better... maybe `filter_matching_relay_list_forced` or `filter_matching_relay_list_include_all`? It would be nice if it was obvious that one is a more niche version of the other.
___
*[`mullvad-relay-selector/tests/relay_selector.rs` line 1747 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8626#-OXxYlWVBhKWnVxm3o5o:-OY0F-zy8pOF7nnceNzc:b-478nap) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/61a20067a4ebf2fedf3019afc6821d2fd7c8111f/mullvad-relay-selector/tests/relay_selector.rs#L1747)):*

I can simplify the representation slightly, but it won't get rid of too much noise :( Initially I wanted to parse the JSON representation of the `RelayList`, but that is also super verbose. Would be good if we opted to pivot to using snapshot testing though!

Did what I could, pls be gentle
</blockquote></details>

That's have to do :)

PR #8630, opened 2025-08-19
Add --skip flag to test-manager

hulthe:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8630)

### DIFF
--- a/test/test-manager/src/main.rs
+++ b/test/test-manager/src/main.rs
@@ -114,6 +114,10 @@ enum Commands {
         #[arg(long)]
         openvpn_certificate: Option<PathBuf>,
 
+        /// Names of tests to not run. Any tests given here will be skipped.
+        #[arg(long)]
+        skip: Vec<String>,
+
         /// Names of tests to run. The order given will be respected. If not set, all tests will be
         /// run.
         test_filters: Vec<String>,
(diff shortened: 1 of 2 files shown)


Serock3 (approved):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8630)* status: all files reviewed, 1 unresolved discussion

___
*[`test/test-manager/src/main.rs` line 119 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8630#-OY0yRd3ABBpeuRxj9gf:-OY0yRd3ABBpeuRxj9gg:b7j3vqw) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/2ae60ad71621b7c6880b094292e9bdfae7d2289e/test/test-manager/src/main.rs#L119)):*
> ```XML
>         /// Names of tests to not run. Any tests given here will be skipped.
>         #[arg(long)]
>         skip: Vec<String>,
> ```

A bit off topic, but the flag name `test_filters` is bad and should be changed to `tests` (or  `run_tests` along with `skip_tests`). The name is legacy from when it used to work differently. I guess you would prefer that to be a separate PR, right?

PR #8641, opened 2025-08-21
Automate remaining parts of TCP-over-UDP tests

raksooo:
This PR implements the remaining parts of the manual TCP-over-UDP tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8641)

### DIFF
--- a/test/test-manager/src/tests/tunnel_state.rs
+++ b/test/test-manager/src/tests/tunnel_state.rs
@@ -4,7 +4,6 @@ use super::{
         self, connect_and_wait, send_guest_probes, unreachable_wireguard_tunnel,
         wait_for_tunnel_state,
     },
-    ui,
 };
 use crate::{
     assert_tunnel_state,
@@ -145,13 +144,6 @@ pub async fn test_disconnected_state(
         detected_probes.all(),
         "did not see (all) outgoing packets to destination: {detected_probes:?}",
     );
-
-    // Test UI view
-    //
-
-    log::info!("UI: Test disconnected state");
-    let ui_result = ui::run_test(&rpc, &["disconnected.spec"]).await.unwrap();
-    assert!(ui_result.success());
 
     Ok(())
 }


desktop/packages/mullvad-vpn/test/e2e/installed/state-dependent/tunnel-state.spec.ts
```diff
@@ -16,117 +16,183 @@ const exec = promisify(execAsync);
 // IN_IP: In ip of the relay passed in `HOSTNAME`
 // CONNECTION_CHECK_URL: Url to the connection check
 
+const HOSTNAME = process.env.HOSTNAME;
+const IN_IP = process.env.IN_IP;
+const CONNECTION_CHECK_URL = process.env.CONNECTION_CHECK_URL;
```
MrChocolatine: What about extracting these values in one go:

```suggestion
const {
  CONNECTION_CHECK_URL,
  HOSTNAME,
  IN_IP,
} = process.env;
```

raksooo: Thanks for the suggestion! Done.

desktop/packages/mullvad-vpn/test/e2e/installed/state-dependent/tunnel-state.spec.ts
```diff
@@ -16,117 +16,181 @@ const exec = promisify(execAsync);
...
+
+      await expectConnected(page);
+
+      await routes.main.expandConnectionPanel();
+
+      const inValue = await routes.main.getInValueText();
+      expect(inValue).toMatch(new RegExp(`${escapeRegExp(IN_IP!)}:(80|5001) TCP`));
+    });
+
+    test('App should show correct port', async () => {
```
olmoh: I personally like structuring tests with multiple inputs like this since I find the test to be a bit more readable after. Not a big deal in this one however since we so few inputs:

```suggestion
    const ports = [80, 5001];
    for (const port of ports) {
      test(`App should show port ${port}`, async () => {
        await gotoUdpOverTcpSettings();
        await routes.udpOverTcpSettings.selectPort(port);
        await routes.udpOverTcpSettings.gotoRoot();
        await routes.main.expandConnectionPanel();

        const inValue = await routes.main.getInValueText();
        expect(inValue).toMatch(`${IN_IP}:${port} TCP`);
      });
    }
```

desktop/packages/mullvad-vpn/test/e2e/installed/state-dependent/tunnel-state.spec.ts
```diff
@@ -16,117 +16,181 @@ const exec = promisify(execAsync);
...
-  await expect(inData).toContainText(new RegExp(':53'));
+  test.afterAll(async () => {
+    await page.close();
+  });
 
-  await exec('mullvad relay set tunnel wireguard --port=51820');
-  await expectConnected(page);
-  await page.getByTestId('connection-panel-chevron').click();
-  await expect(inData).toContainText(new RegExp(':51820'));
+  test('App should show disconnected tunnel state', async () => {
```
olmoh, line 41: This test seems to be a duplicate of the test in `disconnected.spec.ts`, if so we could remove one of them. I would suggest keeping this one to keep all tunnel state tests grouped.

desktop/packages/mullvad-vpn/test/e2e/installed/state-dependent/tunnel-state.spec.ts
```diff
@@ -16,117 +16,181 @@ const exec = promisify(execAsync);
...
+      await gotoWireguardSettings();
+      await routes.wireguardSettings.gotoUdpOverTcpSettings();
+    }
+
+    test.afterAll(async () => {
+      await routes.wireguardSettings.gotoRoot();
+    });
+
+    test('App should show UDP', async () => {
+      await expectConnected(page);
```
olmoh, line 114: I think it's easier to keep the tests clear and easy to understand if they manage their own setup as much as possible. In examples like this i think it would be clearer if the test was not expecting to start from the connected state, since that is set by another test in another describe block. It would be clearer if it actually connected itself, this could apply to other settings as well.

It can be a hassle to reset the state for each state dependant test, but I think a reasonable approach is to let each describe block start from a "reset" state and then let the tests inside the block build on each other. This can be done with setting up `afterAll`, `beforeAll` and others to handle returning to a default state. This would require restructuring the whole test file a bit so it is not required.

desktop/packages/mullvad-vpn/test/e2e/installed/state-dependent/tunnel-state.spec.ts
raksooo: Done!

desktop/packages/mullvad-vpn/test/e2e/installed/state-dependent/tunnel-state.spec.ts
raksooo: As discussed AFK, this test is nice to use manually to verify that running tests works. I've kept the test but removed it from the Rust tests to prevent it being called during automated testing. I've also added a comment in the test to describe why it still exists.

desktop/packages/mullvad-vpn/test/e2e/installed/state-dependent/tunnel-state.spec.ts
raksooo: I've added a call to `mullvad connect` that shouldn't do anything but is nice to understand the state.

desktop/packages/mullvad-vpn/test/e2e/route-object-models/main/main-route-object-model.ts
```diff
@@ -28,4 +28,16 @@ export class MainRouteObjectModel {
...
+    await this.selectors.connectionPanelChevronButton().click();
+  }
+
+  getInValue() {
+    return this.selectors.inValueLabel();
+  }
+
+  async getInValueText() {
+    return this.getInValue().innerText();
+  }
```
tobias-jarvelov: **nitpick:** Linter complains about [require-await](https://typescript-eslint.io/rules/require-await/) rule for this function.

**suggestion:**

A quick fix is to remove the `async` from the function, or add `Promise.resolve()` around the returned result:

```js
  async getInValueText() {
    return Promise.resolve(this.getInValue().innerText());
  }
```

Otherwise we could await the result, assign it to a variable and return the variable.

desktop/packages/mullvad-vpn/test/e2e/route-object-models/main/selectors.ts
```diff
@@ -3,4 +3,6 @@ import { Page } from 'playwright';
 export const createSelectors = (page: Page) => ({
   settingsButton: () => page.locator('button[aria-label="Settings"]'),
   selectLocationButton: () => page.getByLabel('Select location'),
+  connectionPanelChevronButton: () => page.getByTestId('connection-panel-chevron'),
+  inValueLabel: () => page.getByTestId('in-ip'),
```
tobias-jarvelov: **suggestion:** `inIpLabel` in my head would be a more descriptive name, as `Value` is a bit ambiguous and could refer to many things.

desktop/packages/mullvad-vpn/test/e2e/installed/state-dependent/tunnel-state.spec.ts
tobias-jarvelov: I agree with Oliver here that it would be better to show that these tests are dependent on each other by wrapping them in a `describe` block. Although, I don't see this as required for the PR to pass, however I would be up for doing it if you do not mind.

tobias-jarvelov (dismissed):
I have added a couple of comments, however none of them are blocking so I'll mark this PR as approved.

desktop/packages/mullvad-vpn/test/e2e/installed/state-dependent/tunnel-state.spec.ts
raksooo: I've now moved the connect command to `test.beforeAll`.

desktop/packages/mullvad-vpn/test/e2e/route-object-models/main/main-route-object-model.ts
raksooo: Fixed.

desktop/packages/mullvad-vpn/test/e2e/route-object-models/main/selectors.ts
raksooo: Done.

olmoh (dismissed):
New tests are looking clean, nice job!

PR #8654, opened 2025-08-25
Correct QUIC max_udp_payload_size fixme

hulthe:


hulthe:
The fix has been deployed to all relays now

PR #8686, opened 2025-09-01
Dont allow duplicate names for api access methods

tobias-jarvelov:
- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8686)

### DIFF
--- a/desktop/packages/mullvad-vpn/src/renderer/components/EditApiAccessMethod.tsx
+++ b/desktop/packages/mullvad-vpn/src/renderer/components/EditApiAccessMethod.tsx
@@ -21,7 +21,7 @@ import { BackAction } from './KeyboardNavigation';
 import { Layout, SettingsContainer, SettingsContent, SettingsNavigationScrollbars } from './Layout';
 import { ModalAlert, ModalAlertType } from './Modal';
 import { NavigationContainer } from './NavigationContainer';
-import { NamedProxyForm } from './ProxyForm';
+import { NamedProxyForm, ProxyFormButtons, ProxyFormInner, ProxyFormNameField } from './ProxyForm';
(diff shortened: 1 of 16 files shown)


desktop/packages/mullvad-vpn/src/renderer/components/EditApiAccessMethod.tsx
```diff
@@ -108,7 +121,19 @@ function AccessMethodForm() {
                 {id !== undefined && method === undefined ? (
                   <span>Failed to open method</span>
                 ) : (
-                  <NamedProxyForm proxy={method} onSave={onSave} onCancel={pop} />
+                  <NamedProxyForm proxy={method} onSave={onSave} onCancel={pop}>
+                    <ProxyFormNameField
+                      rowProps={{
+                        errorMessage: messages.pgettext(
+                          'api-access-methods-view',
+                          'Please select a name for the access method not already in use.',
```
olmoh, line 129: These error messages are currently to wide, they should align with the right edge of the input, but currently align with the edge of the layout. This was the case before as well but it becomes visible now since none of the other error messages are this long (at least not in english), so I think it's something we can fix in this PR.

olmoh (dismissed):
Nice job, I like this version a lot better than the tightly coupled one before. A small styling thing we can fix otherwise it looks good.

hulthe (commented):
Rust stuff still looks good :dancer: 

> @hulthe reviewed 1 of 6 files at r5, 2 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8686#-:-OaOhHK9BAA37o4TNlCm:b-mjgnq5)* status: 8 of 16 files reviewed, 1 unresolved discussion (waiting on @tobias-jarvelov)

hulthe (approved):

PR #8688, opened 2025-09-01
Add a test that asserts DAITA increases packet size on average

buggmagnet:
This PR adds a test that asserts that, on average, turning on DAITA will lead to an increased packet mean size.

To do this, I did the following : 
- Add a `size` property in the captured packets object we get from our custom capture API.
- Add a test that generates traffic with and without DAITA
- Once the results are in, we compute the mean size of packets and we assert that DAITA traffic should see increased mean packet size. 

If this is too flaky on the long run, we can use the [`t-test` method](https://en.wikipedia.org/wiki/Student%27s_t-test) instead which should indicate that the two data sets are significantly different statistically speaking (i.e. DAITA does have an effect on the traffic size)

The changes for the router API have already been deployed as part of this test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8688)

### DIFF
--- a/ci/ios/test-router/raas/src/capture/parse.rs
+++ b/ci/ios/test-router/raas/src/capture/parse.rs
@@ -58,6 +58,7 @@ pub struct ConnectionId {
 pub struct PacketTransmission {
     from_peer: bool,
     timestamp: u64,
+    size: u32,
 }
 
 #[derive(Default, Debug)]
@@ -108,8 +109,8 @@ impl ParsedConnections {
         }
 
         let transport_protocol = packet.transport_protocol();
-        let Some((source_port, destination_port)) =
-            packet_ports(packet.payload(), transport_protocol)
+        let payload = packet.payload();
+        let Some((source_port, destination_port)) = packet_ports(payload, transport_protocol)
         else {
             log::debug!("Failed to parse an IP packet from {source} to {destination}");
             return;
@@ -137,6 +138,7 @@ impl ParsedConnections {
         let packet_transmission = PacketTransmission {
             from_peer: self.ip_matches_peer(source),
             timestamp,
+            size: u32::try_from(payload.len()).unwrap(),
         };
 
         self.connections
--- a/ios/MullvadVPNUITests/Networking/PacketCapture.swift
+++ b/ios/MullvadVPNUITests/Networking/PacketCapture.swift
@@ -114,11 +114,15 @@ class Packet: Codable, Equatable {
     /// Timestamp in microseconds
     private var timestamp: Int64
 
+    // The size of the packet
+    public let size: Int32
+
     public var date: Date
 
     enum CodingKeys: String, CodingKey {
         case fromPeer = "from_peer"
         case timestamp
+        case size
     }
 
     required init(from decoder: Decoder) throws {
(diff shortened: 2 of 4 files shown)


buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8688)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @acb-mv)

___
*[`ios/MullvadVPNUITests/RelayTests.swift` line 514 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8688#-OZ4hIiF1HSC6GZH_bkf:-OZ8KI65-CQuQaJrSIO0:b-1mtj3c) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8fe2cb1e9bc556247a05c055289c8de349a7063f/ios/MullvadVPNUITests/RelayTests.swift#L514)):*

Given the repetitive nature of this test (with and without DAITA), could it perhaps have the repeated parts factored out and called from the test to give a better overview of what's happening? i.e., lines 524-548 becoming something like `generateAndCaptureTraffic()`?
</blockquote></details>

Probably yes. The `generateTrafficAndDisconnect` is also turning into a frankenstein monstrosity which I'm not super happy with. 
I'll refactor this to be more digestible

rablador (changes requested):
> @rablador reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8688)* status: all files reviewed, 2 unresolved discussions (waiting on @acb-mv)

___
*[`ci/ios/test-router/raas/src/capture/parse.rs` line 141 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8688#-OZ8lidr-soXVA9jkqTF:-OZ8lidr-soXVA9jkqTG:b1019vi) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8fe2cb1e9bc556247a05c055289c8de349a7063f/ci/ios/test-router/raas/src/capture/parse.rs#L141)):*
> ```XML
>             from_peer: self.ip_matches_peer(source),
>             timestamp,
>             size: u32::try_from(payload.len()).unwrap(),
> ```

Safe to unwrap like this?

buggmagnet (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8688)* status: all files reviewed, 2 unresolved discussions (waiting on @acb-mv and @rablador)

___
*[`ci/ios/test-router/raas/src/capture/parse.rs` line 141 at r1](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8688#-OZ8lidr-soXVA9jkqTF:-OZ9GEDL--nSrRdsM1ka:bq6py0v) ([raw file](https://github.com/mullvad/mullvadvpn-app/blob/8fe2cb1e9bc556247a05c055289c8de349a7063f/ci/ios/test-router/raas/src/capture/parse.rs#L141)):*

Safe to unwrap like this?
</blockquote></details>

`payload` here represents a packet, which the size at most would be 64K for a TCP / UDP packet.
Since both `usize` and `u32` cannot contain negative values, and the maximum value of `u32` is way above the maximum size of a packet, this conversion should never fail.

PR #8707, opened 2025-09-04
Fix select location not working in rare cases

Rawa:
If a user upgraded from 2022.2 (where multihop entry was Any by default in settings) select location would be broken because it was unable to and when multihop was enabled and the entry was set as Any.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8707)


Pururun (approved):
:lgtm:

> @Pururun reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/mullvadvpn-app/8707#-:-OZJFMGy1UWhAZFDyxsP:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

